PR #370, opened 2013-08-07
RFC: Pluggable engines

lordhoto:
This is basically an updated version of pull request #204 by Fingolfin. It mainly adapts the changes made there to the new configure features (for example, dependencies). It furthermore slightly changes the attribute names and uses a modified python script.

Let me summarize the basic idea: Instead of having to plug in an engine at the various points in our build system (engines/configure.engines, engines/engines.mk and engines/plugins_table.h) now a engine only has to supply a "engine.desc" file which describes the engines. Running devtools/update-engines.py will pick up all engines in the specified path (should be engines/ run from the top level source dir for now) and then generate the files automatically.

The engine.desc files are some simple JSON (https://www.ietf.org/rfc/rfc4627.txt). The basic layout is like this:
The top level object needs to contain the attributes "description" (which is the human readable engine description for configure) and "status", which is either "stable" (enabled by default) or "unstable" (disabled by default). Furthermore, it can have the following optional attributes:
* games: This contains a human readable description for the supported games in case subengines exist
* subengines: This is an object containing engine descriptions again (in those "games" and "subengines" are not allowed).
* dependencies: This is a list of feature dependencies which are required by the engine.

What's the long term goal behind this?

Our engine count increases continously. In case we really ever decide to split out the core, we have an easy way for users of this core to plug in engines into the core.
Even when we do not split out the core we can still plug in calling devtools/update-engines.py in our build system so that it automatically pulls engines added to engines/. Here, update-engines.py needs to be extended such that it can check whether any changes are present. Then this can be plugged into our Makefile to automatically update that and thus re-run configure. The current design of update-engines.py should be helpful, since it can generate the contents of each build file individually. This should allow us to simply compare file contents, if they differ => regenerate. (create_project would need to be adapted to run this too, which would introduce a python dependency for Windows users, but that should not be too bad).

What needs to be changed before we can merge this?

The script "gen_engines_desc.sh" added in bd7b9904bb98bf2704fe69ff10910ecf2657eefb is only for the transition. I think we should remove this from final history (or at least remove the file after the merge, since it's useless then).


sev-:
@lordhoto, this PR is long time rotten.

What about this: you create a separate PR handling just credits, but preferably use Perl instead of Python, primarily because we're already using it. If you have difficulties with Perl code, please tell me and I'll rewrite your Python script for it.

For the credits file format I think we need to come up with something simpler than JSON, so it could be more or less matching current engine-specific files such as configure.engine.

PR #406, opened 2013-10-05
CONFIGURE: Add option for building on RAM-constrained platforms eg. RPi.

digitall:
If enabled, this option passes flags to the linker to reduce the
RAM required to build. Without this, the linker caches all the
object symbols into memory for build speed, but this can trigger
Out Of Memory on swapless platforms with limited RAM.

### DIFF
--- a/configure
+++ b/configure
@@ -1814,6 +1814,7 @@ EOF
 	# We need at least 256MB for a static build with all engines...
 	if test "$_compile_free_memory" -lt "256" ; then
 		echo insufficient
+		echo "To override this check, pass --disable-ramcheck option to configure."
 		exit 1
 	else
 		echo ok
--- a/configure
+++ b/configure
@@ -128,6 +128,7 @@ _taskbar=yes
 _updates=no
 _libunity=auto
 # Default option behavior yes/no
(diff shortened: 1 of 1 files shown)


fingolfin:
Since this PR doesn't seem likely to be merged: How about closing it, but adding a remark in the README, Wiki etc. that suggests for people on low-mem systems and using GNU ld to use `CXXFLAGS="-Wl,--no-keep-memory" ./configure ... && make` ? (That seems to be what it boils down to, no?)

sev-:
I agree. README, or better Wiki in 'Building ScummVM' section is much better.

lordhoto:
I mention the option on our wiki now at two places:
http://wiki.scummvm.org/index.php/Compiling_ScummVM/GCC#Configuring_ScummVM
and
http://wiki.scummvm.org/index.php/Compiling_ScummVM/RPI#Configuring_ScummVM

I took the liberty of using LDFLAGS instead of CXXFLAGS to store the option since CXXFLAGS are not passed to the linker command.

Closing this for now. If there is any future need to implement this in configure, please open a new PR.

PR #425, opened 2014-01-23
ANDROID: Modernize build system

klusark:
Written originally by @Botje for ResidualVM

This code is essentially what we use for the ResidualVM Android build which will be available on the Play Store when 0.2.0 comes out.

The main issue with the build system is that it used deprecated tools, such as aapt, which are no longer present in the latest SDK versions. This updates the build process to use ant, which is supported by Google.

Currently though plugins are not built. Personally, I don't really see a need for plugins anymore as a full build is only 10mb and devices aren't particularly storage constrained these days. I can work on adding it back if there is interest in this approach.


sev-:
fuzzie, I'm merging this. The whole stuff looks like a good idea. And yes, it is probably the time to try monolithic build.

clone2727:
Removing support for separate engine plugins kills me, really. On my commute every day I only have remote access to my computer where I can build for Android (my phone) and then copy the plugin via SFTP to my phone. I have a limited data plan and cannot constantly copy the whole APK to my phone.

sev-:
Sounds like you're doing it everyday... must be some not public work, as I see only couple of commits to the backend and those look rather cosmetic to me. Since I was not aware of your need, I merged it.

We are in a desperate need of a maintainer for Android port for several years now. So if you like to pick the work up and keep the port up to date, I would not mind seeing this PR or that part of it which kills backwards compatibility reverted.

clone2727:
My work, when I do it, and how often I do, it is not only irrelevant to what I said, but frankly is insulting to me that you'd even look at my commits and say that here in the comments.

I had already discussed my concerns about this to fuzzie in private. Since she's the _de facto_ maintainer (she does the releases now, after all), I expected her to merge it and not have it merged out-of-the-blue.

sev-:
I did not seek any insults at all. I just went to check perhaps I am missing something, not with the goal of undermining your work.

And as you see, I just supported fuzzie's desire to try the monolithic build :).

bluegr:
Well, removing plugins was the right way to go, IMHO. It's easier to handle monolithic builds, both for us and for our users, who get confused by having to add an application and further addons to get a game working.

The whole idea here is to get the port running as easily and efficiently as possible for our users. The fact that one developer has an exotic use case does not justify a reason for using an ancient build system. We can also use TWO different build targets, if that's a real issue: one for the modern monolithic build, and one for the no-longer-needed plugin system

clone2727:
@bluegr That's fine. However, we previously had the monolithic build working and in the store, and removing the plugins did not affect that.

I never said that justified using an ancient build system, because I'm sure they could be hooked up with the new one.

bluegr:
We are currently having issues with the Android Store version. The reason this pull request was added was to avoid having such issues, i.e.:

"The main issue with the build system is that it used deprecated tools, such as aapt, which are no longer present in the latest SDK versions. This updates the build process to use ant, which is supported by Google."

As I said, we can keep the current code in a separate pseudo-platform/build target, something like android_plugins, and we can keep everyone happy. However, if everyone else is using a new build system, we can't cling to a deprecated one. But, if we do need it around, we can keep it separate from our main build system, as mentioned.

clone2727:
@bluegr I understand the deprecated tool warning, but that's *completely* separate from breaking plugins! It doesn't need to be a pseudo-target, we already have the configure switch for dynamic plugins. Your argument here is not counter to mine and you seem to be confusing the issue.

bluegr:
But the plugin code requires further changes to the build target, such as an unpacker, which is not needed for monolithic builds (refer to pull request #560). If we do want to keep all of that, we can keep it *separately* from our main Android build target, which is why I mentioned the pseudo-target that will contain all of this code

clone2727:
@bluegr No, it requires code changes, not changes to the build. And it's all controlled by the DYNAMIC_MODULES macro, which is the way it should be.

bluegr:
So, you are saying that we can keep both the monolithic builds and the plugin-based builds, using the DYNAMIC_MODULES macro?

I suppose that is doable. However, we really should decide what to keep, and what Android versions will be supported.

From what you're saying, we can:
1) Keep all of this plugin code with a DYNAMIC_MODULES define (i.e. revert the commits here and include everything within a DYNAMIC_MODULES define)
2) We will do the same with pull request #560
3) We will keep the main build as monolithic (including the Google Play version), and only use plugins for specific use cases, such as yours

So, is that something that you would be happy with? We can switch to a monolithic build, and keep all of the plugin code optional.

clone2727:
@bluegr What you're saying is what already was in place before these pull requests. Monolithic builds worked before this pull request. This pull request just breaks the building of plugins. I never said to revert this commit, I just didn't want plugins broken with it.

bluegr:
The whole changes to the build system do not work with our plugin system, yet, which is why the plugins were removed with this pull request. So, as it stands, the ant compiler that will now be used for the Android builds will need to support our plugin system again.

Also, does the ant compiler work with our buildbot?

digitall:
@bluegr You seem to be a bit confused about what Ant is... It is not a compiler. It is basically the equivalent of "make" for Java and thus the Android buildchain: https://ant.apache.org/

Native android applications written in Java language are compiled to bytecode (Just that the bytecode is Dalvik, not Java)... The NDK native code is compiled using a C/C++ compiler targetted at the native platform i.e. ARM, MIPS or x86.

Getting the NDK linked together with the rest using Ant and personally I would prefer to keep a Makefile driven system.

digitall:
Looking again at this code, the intent is not to break plugins as far as I can see.

The point here is the APK packaging... before this change, we had one APK for the main application and an APK for each plugin when plugins were enabled, and a single big APK with the monolithic static build when they were disabled.

As a side effect of this change, we now have a single APK in both cases: either containing the main applications and plugins in the first case or the monolithic static build in the second.

What @clone2727 wants is to preserve the first option in the APK packaging for the plugins build i.e. to have separate APKs for each plugin, rather than having to download a large APK with every plugin when he just wants to test with one change. Some users may want this as well since they may want just the main application and a few engines if they have a device with limited storage.

digitall:
Keeping this as an option for the plugins build i.e. the plugins build creates the main application and the plugins... the packaging step in the makefile creates a single "big" apk with main application and all plugins and then a "scummvm-main.apk" and "scummvm-<enginename>.apk" for the main application and each plugin..

I suggest we put the "big" apk as the release one in the Google Play store... but both options are retained for nightly builds on the buildbot. This should not be hard to do, once the required Makefile packaging targets for this are provided... calling setup-apk in the various ways.

sev-:
Guys, I find an argument about "trains and coding" completely unrelated.

First, how much code can you write in trains? But that aside, what prevents from building just the engine which is being worked on? Or any smaller engine if you're working on a common code?

I cannot understand why this is being used for justifying the port to stay on ancient toolchain.

bluegr:
So, let me get this straight...

What is the reason that we need to preserve the plugin system? If it's only for development reasons, can't the developer in question enable a single engine? And why are we talking about 3G here? What does 3G have to do with NORMAL development?

If you want to develop over 3G, connect your device to your PC. Or use WIFI. Or use an emulator, such as BlueStacks or Android-x86. Or something.

Do we really want to keep the whole plugin system because we want to be able to develop over 3G?

clone2727:
This is a development issue, not a user-facing issue. Really, if that's the way you feel, than just do it. I'm one of the developers who still uses the feature, and that's clearly not cared about.

And I never said anything about staying on the ancient toolchain. I have no idea where that came from. That's separate from removing plugin support.

klusark:
It shouldn't be that hard to add plugin support to the new build system, but I don't really have time to look into it. Additionally, if  the unpacker is used, it'll have to be fixed in order to build with Java 7.

digitall:
@klusark Thanks... I hope you can look to restore this functionality at some point. While a single APK and a monolithic static build are the easiest and simplest solutions, they are the largest for both downloading and installation. Removing the ability to do dynamic plugin builds for Android (and to package them in separate APKs) could come back to haunt us.

PR #433, opened 2014-01-28
ANDROID: Update the asset archive code to use AAssets

klusark:
The AAssets API was added in android 2.3, so it also requires using
the android-9 toolchain.

Using this gives a rather large speedboost to starting the ScummVM
application.

This is being upstreamed from ReisdualVM https://github.com/residualvm/residualvm/commit/01459a674fb4875b822508b804b83e351f46938b

### DIFF
--- a/backends/platform/android/asset-archive.cpp
+++ b/backends/platform/android/asset-archive.cpp
@@ -22,8 +22,6 @@
 
 #if defined(__ANDROID__)
 
-#include <jni.h>
-
 #include <sys/types.h>
 #include <unistd.h>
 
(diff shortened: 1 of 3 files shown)


sev-:
This needs to be updated before it could me merged. The pull request got rot.

bluegr:
This looks good to be merged, and it has been approved by fuzzie.

As noted, this will drop support for the 4-year old Android 2.2, but we can live with that.

Note that currently our Android builds are broken, as a result of merging pull request #425. The Android port will likely get fixed with the work done on pull request #560. So, I suggest we merge this and pull request #560 to get the new Android builds working again.

digitall:
@fuzzie : Is this OK to be merged now?

sev-:
If no objections till November 14th, I'll merge this.

fuzzie:
I don't think we have to wait that long - my comment above about consensus is more than a year old! If this breaks any builds then we can fix them afterwards.

PR #459, opened 2014-04-04
Proof-of-Concept: Display unknown version information in GUI

lordhoto:
I wanted to push this out for discussion for a long time, so let's do it now :-P. As the title suggests this is a try at showing unknown version information on detection to the user. This will hopefully raise awareness of the users that even though ScummVM might detect a game he still might have an unkown version (and thus that we would like to get a report about it).

The GUI is pretty a simple hack (and should be replaced by a proper scrollable text field, since otherwise it doesn't really fit in lowres mode). The API extensions to detectGames is also really simple and we might think of making this nicer (maybe we add a getLastDetectionError/Information?).

Attached some screenshots so you get a feeling of how it looks like. These show the dialog which will come up when ScummVM's AdvancedDetector detects a game via fallback.
Hires:
scummvm-hires-unknown
Lowres:
scummvm-lores-unknown

### DIFF
--- a/base/plugins.cpp
+++ b/base/plugins.cpp
@@ -514,7 +514,7 @@ GameDescriptor EngineManager::findGameInLoadedPlugins(const Common::String &game
 	return result;
 }
 
-GameList EngineManager::detectGames(const Common::FSList &fslist) const {
+GameList EngineManager::detectGames(const Common::FSList &fslist, Common::String *error) const {
 	GameList candidates;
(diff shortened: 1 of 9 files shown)


bluegr:
Any update on this? Is there a cross-platform way to either copy the detection data to the clipboard, or open the log file in a text editor? (also refer to my previous reply)

sev-:
Any updates?

PR #508, opened 2014-09-16
COMMON: Add support for endian-safe reading/writing of int64

RichieSams:
The concept is rather self explanatory. There are however, some potential problems:

1. First, I need someone to help me figure out if there is a mips instruction for 64 bit swap, and if so, help me write the asm
2. clone2727 brought up the point that int64 can be disabled in configure. If int64 is disabled, this code will fail to compile. There are a couple choices
  1. Surround this code with #ifdef guards. This isn't the best solution because any engines that use the code will now be restricted to backends that allow int64.
  2. Are there actually backends that don't have support for int64? If not, then we could remove the option to disabled int64 in configure

Let me know if you have any questions and feel free to comment.

### DIFF
--- a/common/endian.h
+++ b/common/endian.h
@@ -49,6 +49,17 @@
 #	error No endianness defined
 #endif
 
+
+#define SWAP_CONSTANT_64(a) \
+	((uint64)((((a) >> 56) & 0x000000FF) | \
+	          (((a) >> 40) & 0x0000FF00) | \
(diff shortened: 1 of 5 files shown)


sev-:
Since the test have been implemented, I'm merging this.

Thank you!

clone2727:
What? Why was this merged? The INT64 define problem has not been addressed!

PR #512, opened 2014-09-22
WME: Multirect 2014

tobiatesan:
2014, prettified, simplified version of https://github.com/scummvm/scummvm/pull/392

Provides the ability to keep track of multiple dirty rects in Wintermute, saving on blitting.

Unfortunately, with lots of rects if becomes useless and we fall back to the usual single rect approach.

### DIFF
--- a/engines/wintermute/base/gfx/osystem/base_render_osystem.cpp
+++ b/engines/wintermute/base/gfx/osystem/base_render_osystem.cpp
@@ -40,6 +40,7 @@
 #include "engines/wintermute/math/math_util.h"
 #include "graphics/transparent_surface.h"
 
+#define DIRTY_RECT_GLOBAL_LIMIT 800
 
 namespace Wintermute {
 
@@ -60,6 +61,7 @@ BaseRenderOSystem::BaseRenderOSystem(BaseGame *inGame) : BaseRenderer(inGame) {
 
 	_dirtyRects = new DirtyRectContainer();
 
+	_gigaRect = nullptr;
 	_disableDirtyRects = false;
 	if (ConfMan.hasKey("dirty_rects")) {
 		_disableDirtyRects = !ConfMan.getBool("dirty_rects");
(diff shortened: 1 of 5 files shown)


sev-:
I would like to have .size() usage cleaned up, and then have it finally merged.

tobiatesan:
@sev- you do know that it's generally useless, though, do you?
(Basically, IIRC, it is useful in the case of a simple, simple game with few rects and a machine that has enough cache to keep the index array in there but which is terribly slow at blitting. It certainly does not seem to do much to help JULIA and the various other problem children.)

sev-:
@tobiatesan Besides JULIA there are a great bunch of other games. Do we really see no advantage?

For the Broken Sword 2.5 which runs in 800x600x16bit, a similar dirty rect feature brought instant speedups.

tobiatesan:
@sev- I don't honestly know.

What I do know is that the one title @somaen proposed as a use case doesn't seem to significantly benefit from it in its current state.

I also know that, 2 years down the line, *I don't like it one bit*.

I think it is broken algorithmically, because it uses a naive algorithm in which, we have discovered, the sheer overhead of handling the rectangles outgrows the cost of doing the actual blitting.

It is however possible that with some clever (but certainly not Turing award-level) mathematics some actual results can be had.

I believe it can be modeled as an optimization problem and attacked with the existing mathematical toolbox. Depending on *how* it is modeled, we may want, for example, to backtrack early once we realize that we have gone down a bad branch of our solution tree, where "bad" means "a zillion rectangles".

This is in fact something I would absolutely *love* to do, but it can't be the product of a couple nights of drunk coding, it requires discipline, thought and concentration (unless you are EWD).
I can't say when I'll be able to work on it, unfortunately I have a whole lot of life getting in the way.

I'd like to hear @wjp's take on this, since I believe he's an actual mathematician and not some lame hack.

Another interesting thing, engineering-wise, could be refactoring it to enable "pluggable" algorithms/strategies.

tobiatesan:
@sev-: I don't know about BS2.5, but thanks for the pointer. I'll definitely have a look to how it's implemented in there and how the whole graphics pipeline differs.

PR #525, opened 2014-10-29
Add override specifiers everywhere

SpecLad:
I used `clang-modernize` to stick an `override` specifier on all member functions overriding base class members. Naturally, the diff is absolutely ginormous, but I think it's worth it, since it improves safety and readability.


fingolfin:
Is there any reason this is still open? To reiterate some points already discussed above, and bring up some more:
* If somebody feels this is important to have, they should bring it up on -devel. 
* So far, no realy convining argument was made as to where this would be better than warnings generated by recent clang and gcc warnings (which also catch override problem)
* this PR does not apply anymore
* the changes were made automatically anyway, using `clang-modernizer`, and hence could be recreated anytime a decision to apply this is made
* this PR seems insufficient anyway, as there is no provision to deal with compilers not supporting `override` (which probably is still the majority in actual live systems).

So, there seems no reason to keep this open?!

lordhoto:
Yes, it seems it was forgotten to close this. If we want this, this should be discussed on -devel as discussed earlier in this PR.

>this PR seems insufficient anyway, as there is no provision to deal with compilers not supporting override (which probably is still the majority in actual live systems).

We have an #define in place for compilers which do not support override (of course they do not start to take advantage of override semantics), so using it is not a problem. Actually, we have code with override in master already.

fingolfin:
Right, my mistake -- the wintermute and prince engines use `override` AFAICT

PR #531, opened 2014-11-12
I18N: Localise language names in GUI

BenCastricum:
This approach is based on placing the language names in a separate .po(t) file. Only for languages which are compatible with the current GUI limitations (Latin-1) a .po file should be created.

### DIFF
--- a/gui/launcher.cpp
+++ b/gui/launcher.cpp
@@ -88,6 +88,524 @@ enum {
 	kCmdSavePathClear = 'PSAC'
 };
 
+<<<<<<< 69d43ce1e23d0e3e8d8a5826169a20a2a0cd4b33
+=======
+/*
+ * TODO: Clean up this ugly design: we subclass EditTextWidget to perform
+ * input validation. It would be much more elegant to use a decorator pattern,
+ * or a validation callback, or something like that.
+ */
+class DomainEditTextWidget : public EditTextWidget {
+public:
+	DomainEditTextWidget(GuiObject *boss, const String &name, const String &text, const char *tooltip = 0)
+		: EditTextWidget(boss, name, text, tooltip) {
+	}
+
+protected:
+	bool tryInsertChar(byte c, int pos) {
+		if (Common::isAlnum(c) || c == '-' || c == '_') {
+			_editString.insertChar(c, pos);
+			return true;
+		}
+		return false;
+	}
+};
+
+/*
+ * A dialog that allows the user to edit a config game entry.
+ * TODO: add widgets for some/all of the following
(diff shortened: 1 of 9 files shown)


sev-:
I think it is not quite correct concept.

The problem is that I will have difficulties with switching back to my language from, say, Japanese. E.g. I have no clue that ロシア語 means "Russian language", while I still will recognize it in English.

BenCastricum:
True, if you change to GUI to a language you can't read, then it's hard to use it. But that's an issue that is not ScummVM specific, anything that supports multiple languages has this problem. Deleting the .ini might be suggested as a solution. Or perhaps ScummVM should get a safe-mode CLI switch?
Do you want me to submit a "safe-mode" patch? Or do you consider this a a blocking issue anyway in which case I'll cancel this pull request. I can't solve the issue you describe when the goal is "localize everything".

rootfather:
I'm a fan of localizing everything. A safe-mode CLI switch would be quite helpful to address the problemn sev- mentiones. However, nearly every program with foreign language support has the problem that you might end up with a screwed user experience if you choose a language you can't read.

sev-:
@BenCastricum a safe mode CLI setting sounds good though it could be not very usable, particularly for the Windows users. Do you have any better ideas on how to avoid a 'language lock-in"?

BenCastricum:
How does "hold shift on startup" sound as a way to activate safe mode (which is what my email client uses too)?  This might be doable with ScummVM as well. Of course this doesn't work on keyboardless setups. But then again, it already is a corner case situation what we are trying to solve here.

rootfather:
Sev, I don't see why a specific CLI command should be problematic for Windows users as long as we cover this in the README? Would be the easiest way to prevent a lock-in and should be easier to implement than a safe-key that has to be hold on program startup.

BenCastricum:
Actually the situation sev describes will never happen with this proposal since we only support ascii. GUI / Port limitations prevent the use of non-ascii game descriptions.

gui/launcher.cpp
```diff
@@ -88,6 +88,524 @@ enum {
 	kCmdSavePathClear = 'PSAC'
 };
 
+<<<<<<< 69d43ce1e23d0e3e8d8a5826169a20a2a0cd4b33
```
sev-: Ahem. This should be cleaned up.

sev-:
I am still not in favour of this. Perhaps a separate project for making our GUI UTF-8 could help making it properly. At this moment, however, the approach has too many drawbacks to be merged.

BenCastricum:
Ok, no problem.

PR #536, opened 2014-11-25
BACKENDS: Add zip filesystem node

skristiansson:
This exposes a zip archive as a file system according to
the ScummVM common API.

The benefit of this is that generic code can use compressed
zip archives transparently without having to special case for
it.

To elaborate a bit more on why I'm doing this - it's a part of the Windows Phone 8.1 port I'm working on
and there's certain restrictions what an app is allowed to gain access to on the file system.
In particular, to be able to read files from the SD-card, the application needs to register a set of
file extensions that it is allowed to read.
With the wide plethora of different types of files that can exist in the game data (to not mention files
without extension), the most viable option I saw was to add support of containing game data in an archive.

To give an idea how the zip-fs backend is thought to be used, I pushed a proof-of-concept commit
that make the posix backend use it here:
https://github.com/skristiansson/scummvm/commit/f808db9dd3cffd4d6a4e3c3d22d5feab813e5286

I realize that pulling in code without current users might be undesirable, but this was self-contained
enough from the rest of the wp 8.1 port, so I thought I'd take a chance to get some input on this while
working on the rest.

### DIFF
--- a/backends/fs/zip/zip-fs.cpp
+++ b/backends/fs/zip/zip-fs.cpp
@@ -1,0 +1,126 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "common/system.h"
+#include "common/archive.h"
+#include "common/unzip.h"
+#include "backends/fs/fs-factory.h"
+#include "backends/fs/zip/zip-fs.h"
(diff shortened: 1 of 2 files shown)


somaen:
How about the simplest solution of all? Adding .scummvm or something like that as a file suffix to ALL the gamefiles, and handling the addition of that in the backend? That would mean only one file-extension to register, where the backend can do the rest. The tedious process of renaming all files from i.e. monster.sou -> monster.sou.scummvm could easily be made into a tool for endusers. Less overall technicality, and less changes that would be technically problematic for us.

Hades32:
Another alternative would be to use the user's OneDrive. After you download a folder into your local app data directory you can access every file regardless of the file extension.

----- Ursprüngliche Nachricht -----
Von: "Einar Johan Trøan Sømåen" <notifications@github.com>
Gesendet: ‎10.‎01.‎2015 17:43
An: "scummvm/scummvm" <scummvm@noreply.github.com>
Cc: "Martin Rauscher" <hades32@gmail.com>
Betreff: Re: [scummvm] BACKENDS: Add zip filesystem node (#536)

How about the simplest solution of all? Adding .scummvm or something like that as a file suffix to ALL the gamefiles, and handling the addition of that in the backend? That would mean only one file-extension to register, where the backend can do the rest. The tedious process of renaming all files from i.e. monster.sou -> monster.sou.scummvm could easily be made into a tool for endusers. Less overall technicality, and less changes that would be technically problematic for us.

Hades32:
There might be an alternative way via folder pickers:
http://stackoverflow.com/a/27482747

skristiansson:
Both OneDrive and file picker suggestions are worth investigating (especially the onedrive support, the file picker suggestion has a slight problem that you have to the view 'native gui', not sure how easily achievable that is under the current circumstances).

Meanwhile I've got a proof-of-concept romfs implementation done in my windows phone 8.1 branch:
https://github.com/skristiansson/scummvm/tree/wp81/backends/fs/romfs

It works fairly well, a lot of games I've tried with it works fine, but there's some problems with broken sword (more specifically, seems to be the libvorbis file seek callbacks that make it act up) that I need to investigate.

PR #540, opened 2014-12-01
SDL: add support for SDL2

skristiansson:
As the title says, this set of patches adds initial support for SDL2.
The commit messages for each patch should contain an explanation to deficiencies
in the related areas, but one thing that is not mentioned is that this does not contain support
for explicitly using OpenGL (although it might be the actual backend SDL2 is using to render).

Also, some of the changes might be considered a bit of hacks (the compatibility defines for an instance),
so I'm certainly up for discussion on how to improve on cleanliness.

I've tested the changes on a normal Linux machine, and as far as my tests goes, the changes are
at a state where they are usable (albeit experimental).

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -75,17 +75,21 @@ SdlEventSource::~SdlEventSource() {
 }
 
 int SdlEventSource::mapKey(SDLKey key, SDLMod mod, Uint16 unicode) {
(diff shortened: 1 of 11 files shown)


lordhoto:
Superseded by #572. Thanks for your work.

I added a credits entry for you. Please check whether you think you are fine with that. It is in the "Code contributions" you can, for example, find here: http://scummvm.sourceforge.net/credits/

skristiansson:
I'm fine with that, and thanks for picking this up and finishing it off.

PR #544, opened 2014-12-06
Symbian : add Theora support

fedor4ever:
Symbian use fixed-point Theora decoder unlike other backends.


sev-:
It is not clear to me either why there is __SYMBIAN32__, thus the whole patch could be remade into "Add Theora support to systems which use libtremor".

fedor4ever:
sev, I read wiki - wiki.scummvm.org/index.php/Platforms - only Symbian use Tremor. Because I made patch for Symbian only. If common varian for system with tremor preferrable - I remade patch

PR #545, opened 2014-12-06
SDL: explicitly set unicode parameter to 0 on SDL_KEYUP events

skristiansson:
As per SDL documentation, unicode character codes are only
generated on keydown events, not on keyup events.
http://www.libsdl.org/release/SDL-1.2.15/docs/html/sdlenableunicode.html

This make it explicit to make it clear from the code that the
unicode character will *not* be provided on keyup events.


lordhoto:
Superseded by #572. Thanks for your work.

PR #554, opened 2014-12-30
Day of the Tentacle easter egg

eriktorbjorn:
Here is one possible way of supporting the Maniac Mansion easter egg in Day of the Tentacle, by adding a mechanism where instead of returning to the launcher dialog, it automatically launches another game. Optionally with a save game slot to load on startup.

So to start Maniac Mansion it first creates a temporary savegame, using the hitherto unused (I think) save slot 100, adds Maniac Mansion and itself to the "chained games" list and forces a return to the launcher. Manually returning to the launcher from Maniac Mansion resumes Day of the Tentacle at the temporary savegame.

It detects Maniac Mansion by looking if the ScummVM config file contains any game that is in a "Maniac" sub-directory of the current game. Since this may not always work, it also checks for an "easter_egg" config setting, though there is no GUI for it at the moment, and no error handling. (I.e. no checks if the game exists, and setting the game to Day of the Tentacle itself is probably not such a bright idea, for instance...)

Quitting still quits the entire ScummVM. I couldn't think of any sensible way to do otherwise.

### DIFF
--- a/base/main.cpp
+++ b/base/main.cpp
@@ -529,43 +529,21 @@ extern "C" int scummvm_main(int argc, const char * const argv[]) {
 			// at. At the time of writing, this is used for the Maniac Mansion
 			// easter egg in Day of the Tentacle.
 
-			Common::String chainedGames, nextGame, saveSlot;
+			Common::String chainedGame;
+			int saveSlot = -1;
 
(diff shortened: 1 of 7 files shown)


sev-:
This is really great stuff and properly done.

Merging.

PR #557, opened 2014-12-30
Replace 'Pluto' lua persistance code with re-written, platform-agnostic code

RichieSams:
The old code (pluto), made a lot of assumptions when reading and writing data. One of the biggest was direct serialization of doubles. This code directly address those problems.

Also, I have re-written the code to follow our code formatting guidelines as well as be better documented.

Problems addressed:

1. Doubles
2. Many offsets were serialized as a size_t. Now they are truncated to a uint32. (See [lua_persist.cpp, Line 520](https://github.com/RichieSams/scummvm/blob/67114c3e7eafdd6ba8c7bc799a40f789acc1efa3/engines/sword25/util/lua_persist.cpp#L520)) As said in the comment, this is fine, because if truncation is happening, we have bigger problems.
3. Indexes are stored in the refTable as actual uint32s rather than faked using pointers (Similar to the fixes in 2ec3adfad531b7e1f698c338545fb9817f99e5b1)


sev-:
Okay, merging.

PR #558, opened 2015-01-04
SDL: Warn if the selected joystick does not exist

klusark:


sev-:
What joystick #0 is supposed to do? Issue a warning?

sev-:
ping?

klusark:
joystick_num = 0 tells sdl to use the 1st joystick.

digitall:
This patch is actually fixing the issue where the reported active joystick is always the Joystick #0 data and there is no validation that the joystick number passed actually exists... But the patch is incorrect as the test for the joystick number is off by one...

digitall:
I think this needs a bit more work... in the warning case, odd things could happen as _joystick is not opened... and the line 58 change is spurious as > -1, is the same as >= 0.... unless I am missing something.

digitall:
Ah ignore my note about the off by one, just that I would reverse the test on line 65 to be "if (joystick_num < SDL_NumJoysticks())" as clearer.

digitall:
OK... this all looks fine to me. It basically fixes the debug output to give the right answer i.e. correctly report the active joystick information, rather than Joystick #0 even when another one is in use, and adds a warning when the joystick number is out of range.

This should be merged.

PR #560, opened 2015-01-06
ANDROID: Remove unpacker and fix support for non arm

klusark:
This could be split into two commits, but I can't get it to compile at all with the unpacker code, so I figured I'd just leave it as one.

Since we don't have support for multiple libs, we don't need the unpacker.


clone2727:
FWIW, I already expressed my disappointment/distress at the removal of plugins in pull #425, so I'm against the further removal of the code here and want to see the full support restored instead.

bluegr:
This is the correct way to go, IMHO: we don't want to use a plugin system if we don't need it, just for an exotic use case. I've also replied in pull #425 regarding this

clone2727:
@bluegr It's not used by users now, no, but having a developer on the project making active use of it should be a good enough of a reason to keep it in the code.

sev-:
@clone2727 Do you have time to make it (a) up to modern toolchain (b) reacting to --enable-plugins?

If not, I rather get this PR accepted and make daily builds working over again.

clone2727:
No, I don't have time for it (which is my want for it kept in the first place). I already said in the other one that you can just merge it, since it's obvious no one else cares.

sev-:
Fine, merging.

bluegr:
I don't believe that nobody cares... once we get the plugin system working with the new ant system, we can reintroduce this code

PR #563, opened 2015-01-11
SCUMM: language fixes for HE games

BenCastricum:
Using http://jradventures.forumotion.com/t14-the-uk-english-dubs as source for determining which games are both available in en_us and en_gb I came up with these patches. I also marked the games which still need some more checking but which I don't have access to.

### DIFF
--- a/devtools/scumm-md5.txt
+++ b/devtools/scumm-md5.txt
@@ -493,7 +493,7 @@ freddi	Freddi Fish 1: The Case of the Missing Kelp Seeds
 	cf8ef3a1fb483c5c4b1c584d1167b2c4	-1	de	Windows	HE 73	-	-	Oncer
 	1f2e62b5a9c50589fc342285a6bb3a27	-1	he	Windows	HE 73	-	-	e_orz
 	507bb360688dc4180fdf0d7597352a69	26402	se	Windows	HE 73	-	-	Sven Arvidsson
-	df047cc4792150f601290357566d36a6	-1	us	All	HE 90	Updated	-	khalek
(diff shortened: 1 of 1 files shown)


bluegr:
Shouldn't the rest of the files be updated as well?

BenCastricum:
You're right, I added the update for scumm-md5.h

devtools/scumm-md5.txt
```diff
@@ -489,7 +489,7 @@ freddi	Freddi Fish 1: The Case of the Missing Kelp Seeds
 	cf8ef3a1fb483c5c4b1c584d1167b2c4	-1	de	Windows	HE 73	-	-	Oncer
 	1f2e62b5a9c50589fc342285a6bb3a27	-1	he	Windows	HE 73	-	-	e_orz
 	507bb360688dc4180fdf0d7597352a69	26402	se	Windows	HE 73	-	-	Sven Arvidsson
-	df047cc4792150f601290357566d36a6	-1	us	All	HE 90	Updated	-	khalek
+	df047cc4792150f601290357566d36a6	-1	en	All	HE 90	Updated	-	khalek
```
sev-: I don't understand this, are you sure that it is not US, but also GB? Do you have both? And same thing regarding further similar cases.

BenCastricum: I do not have both (unfortunately), and even then it's possible that multiple versions exists of the same language.

My assumption is that for these type of games MD5 clashes are nearly impossible. There is no registered clash yet. But if there actually is a clash, then there is a higher probability of it being reported if we report incorrect information, instead of generic. If we get a report I will gladly try to detect both versions properly.

sev-:
Okay, let's merge it and see if there are clashes.

PR #564, opened 2015-01-11
SCI: fix definition of GK1 police-sleep patch

mcb30:


bluegr:
Thanks, these slipped through. Merging

PR #568, opened 2015-01-23
ZVISION: Fix sound bug #6767 by making pan_track code similar to origina...

Marisa-Chan:


bluegr:
Looks good at a first glance, thanks! I won't be able to check this until Sunday night, though.

bluegr:
I've just playtested this, and the bug is indeed fixed. I can't spot any obvious regressions, so I'm merging this.

Thanks!

PR #569, opened 2015-01-23
MORTEVIELLE: some display bugs fixed

chrilith:
Hi,

This is a patch for the following problems:

- The boxes are 1 pixel bigger in the DOS game. I fixed the drawBox() method to match original code.
- Also, alerts are missing the first character. So I fixed the show() method changing base string read index.

Thanks!


digitall:
Looks fairly simple changes.

@Strangerke : OK to merge with you?

PR #571, opened 2015-01-25
Small adjustments to match original code 

chrilith:
Hi,

Here are some missing drawBox() adjustments/ They should all match original code now. Also I changed DISK_ACESS_DELAY to 1000msecs which seems to be far enough to read such a short message on screen.

Thanks!


chrilith:
Eh!? Don't know where this 241f459 item come from, sorry...

PR #572, opened 2015-01-25
SDL: Add experimental support for SDL2

lordhoto:
This is basically a cleaned up and extended version of pull request #540. Unlike #540, this rather tries to directly implement compatibility wrappers for removed/changed functionality. This allows us to get rid of defining deprecated functionality with asserts (which is bad because doing so might allow SDL ports to still compile but then fail on runtime).

Other differences compared to #540:
- Working mouse wheel support
- Does not break the OpenGL backend when ScummVM was compiled with SDL1.2
- Our OpenGL SDL backend builds (thus no --disable-opengl is needed when compiling with SDL2) and seems to work fine
- Window icon setup also working in theory for SDL2 (somehow it is distorted when I tested it on X11 though...)
- Window title/icon gets kept when switching from SurfaceSDL <-> OpenGLSDL
- SDL2 can be explicitly enabled by setting SDL_CONFIG when running configure instead of --enable-sdl2

This is only minimally tested.

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -173,7 +173,9 @@ void SdlEventSource::handleKbdMouse() {
 				_km.y_down_count = 1;
 			}
 
-			SDL_WarpMouse((Uint16)_km.x, (Uint16)_km.y);
+			if (_graphicsManager) {
+				_graphicsManager->warpMouseInWindow((Uint16)_km.x, (Uint16)_km.y);
+			}
 		}
 	}
 }
@@ -429,7 +431,9 @@ bool SdlEventSource::handleKeyDown(SDL_Event &ev, Common::Event &event) {
 
 	// Ctrl-m toggles mouse capture
 	if (event.kbd.hasFlags(Common::KBD_CTRL) && ev.key.keysym.sym == 'm') {
-		toggleMouseGrab();
+		if (_graphicsManager) {
+			_graphicsManager->toggleMouseGrab();
+		}
 		return false;
 	}
 
@@ -750,13 +754,6 @@ bool SdlEventSource::remapKey(SDL_Event &ev, Common::Event &event) {
 	}
 #endif
 	return false;
-}
-
-void SdlEventSource::toggleMouseGrab() {
-	if (SDL_WM_GrabInput(SDL_GRAB_QUERY) == SDL_GRAB_OFF)
-		SDL_WM_GrabInput(SDL_GRAB_ON);
-	else
-		SDL_WM_GrabInput(SDL_GRAB_OFF);
 }
 
 void SdlEventSource::resetKeyboadEmulation(int16 x_max, int16 y_max) {
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -183,7 +183,7 @@ void SdlEventSource::handleKbdMouse() {
 			}
 
 			if (_graphicsManager) {
-				_graphicsManager->warpMouseInWindow((Uint16)_km.x, (Uint16)_km.y);
+				_graphicsManager->getWindow()->warpMouseInWindow((Uint16)_km.x, (Uint16)_km.y);
 			}
 		}
 	}
@@ -471,7 +471,7 @@ bool SdlEventSource::handleKeyDown(SDL_Event &ev, Common::Event &event) {
 	// Ctrl-m toggles mouse capture
 	if (event.kbd.hasFlags(Common::KBD_CTRL) && ev.key.keysym.sym == 'm') {
 		if (_graphicsManager) {
-			_graphicsManager->toggleMouseGrab();
+			_graphicsManager->getWindow()->toggleMouseGrab();
 		}
 		return false;
 	}
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -63,8 +63,14 @@ SdlEventSource::SdlEventSource()
 
 		// Enable joystick
 		if (SDL_NumJoysticks() > joystick_num) {
-			debug("Using joystick: %s", SDL_JoystickName(joystick_num));
 			_joystick = SDL_JoystickOpen(joystick_num);
+			debug("Using joystick: %s",
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+			      SDL_JoystickName(_joystick)
+#else
+			      SDL_JoystickName(joystick_num)
+#endif
+			     );
 		} else {
 			warning("Invalid joystick: %d", joystick_num);
 		}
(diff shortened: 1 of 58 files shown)


backends/events/sdl/sdl-events.cpp
```diff
@@ -516,7 +559,7 @@ bool SdlEventSource::handleKeyUp(SDL_Event &ev, Common::Event &event) {
 
 	event.type = Common::EVENT_KEYUP;
 	event.kbd.keycode = SDLToOSystemKeycode(ev.key.keysym.sym);
-	event.kbd.ascii = mapKey(ev.key.keysym.sym, (SDLMod)ev.key.keysym.mod, (Uint16)ev.key.keysym.unicode);
+	event.kbd.ascii = mapKey(ev.key.keysym.sym, (SDLMod)ev.key.keysym.mod, 0);
```
sev-: Aren't we changing behaviour here?

aquadran:
I see you didn't tested with other platforms except linux. there are still few changes needed for windows SDL. and PS3 port need updates in events. also doublebuffersdl-mixer.cpp need small ifdef for SDL2

backends/graphics/dinguxsdl/dinguxsdl-graphics.cpp
```diff
@@ -122,7 +122,7 @@ void DINGUXSdlGraphicsManager::initSize(uint w, uint h) {
 	if (w > 320 || h > 240) {
 		setGraphicsMode(GFX_HALF);
 		setGraphicsModeIntern();
-		_eventSource->toggleMouseGrab();
```
sev-: Looks fishy to me. Is event logger still works with this patch?

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -300,6 +342,58 @@ bool OpenGLSdlGraphicsManager::setupMode(uint width, uint height) {
...
 	}
 
+	// This is pretty confusing since RGBA8888 talks about the memory
+	// layout here. This is a different logical layout depending on
+	// whether we run on little endian or big endian. However, we can
+	// only safely assume that RGBA8888 in memory layout is supported.
+	// Thus, we chose this one.
+	const Graphics::PixelFormat rgba8888 =
+#ifdef SCUMM_LITTLE_ENDIAN
+	                                       Graphics::PixelFormat(4, 8, 8, 8, 8, 0, 8, 16, 24);
```
sev-: Interesting how it worked before...

backends/mixer/sdl/sdl-mixer.cpp
```diff
@@ -57,11 +57,13 @@ void SdlMixerManager::init() {
 		error("Could not initialize SDL: %s", SDL_GetError());
 	}
 
+#if !SDL_VERSION_ATLEAST(2, 0, 0)
```
sev-: What we get with SDL2.0 here?

backends/events/sdl/sdl-events.cpp
lordhoto: Please, see the discussion about that on #540 and #545. Here's a tl;dr to save you some time:
* Since the unicode value for key up events is always 0 there's no differences. We still set some ASCII values based on key codes.

configure
```diff
@@ -170,7 +171,7 @@ _stagingpath="staging"
 _win32path="c:/scummvm"
 _amigaospath="Games:ScummVM"
 _staticlibpath=
-_sdlconfig=sdl-config
+_sdlconfig="$SDL_CONFIG"
```
sev-: And what happens if this variable is not set?

backends/module.mk
```diff
@@ -74,9 +74,11 @@ MODULE_OBJS += \
 
 # SDL 1.3 removed audio CD support
 ifndef USE_SDL13
+ifndef USE_SDL2
```
sev-: I am not sure. Are we getting stubs defined?

backends/graphics/dinguxsdl/dinguxsdl-graphics.cpp
lordhoto: You mean the event recorder? I don't think it ever worked with the Dingux backend, did it? But I can't say I can test Dingux anyway... But I don't see how this should have any effect since the event recorder doesn't do anything in the old SdlEventSource::toggleMouseGrab function. Maybe you can explain this in depth?

backends/module.mk
lordhoto: It did not even try to use it because starting with SDL1.3 (a prerelease used by our PS3 port) we didn't include it anymore. Instead it used the default audio cd implementation (i.e. emulation only). I guarded the implementation with some #ifs now so even trying to including the code which would try to use it will make it fail at unit compile time.

configure
lordhoto: It will try sdl-config, sdl11-config, sdl12-config like before. See line 357.

lordhoto:
@aquadran Yes, the Win32 port seems to require some changes to its icon setup code. I need to check whether my cross compiler setup still works to see if I can fix it.
The PS3 port we use seems to use some SDL 1.3 pre-release. I don't think all ports which rely on a specific SDL version will auto work with this. But then again, that's why we have our porters to work on that...

bgK:
Yes, please do not bother too much with the PS3 port. I will switch it to using native SDL2 API instead of the current SDL1.2 wrapper if this gets merged.

@lordhoto you may want to include this commit if maintaining the version 1.3 differences bothers you:
https://github.com/bgK/scummvm/commit/81e1dab89279351373dea1c0fb18c5c657f012d3

aquadran:
@lordhoto I'll post link to my past effort porting to SDL2, so windows and PS3 porter have some reference of needed changes: https://github.com/aquadran/residualvm/commit/c3160e5f6701d1784f0d35fccf79e7cf49cc5ea2

lordhoto:
I finally updated the pull request so the Win32 build with SDL2 works. I only started the binary with wine, so I couldn't test whether icon setup/taskbar things really work as expected.

@bgK I think it makes more sense if you open a pull request for updating the PS3 port after this is merged.

Tea23:
FWIW I built this out of curiosity and found myself playing through Curse of Monkey Island. Am watching the credits roll by now - had no major issues. Was nice to be able to run ScummVM without the weirdness SDL1 has w/ Linux + multiple monitors.

Thanks a lot for this, lordhoto. Hope it's in the main branch soon :)

bluegr:
Many thanks for this lordhoto! A welcome addition :)

I've just tried to build SDL2 locally with MSVC, and there are a lot of errors caused by a forbidden setjmp() instruction...

I believe that this is caused by an inclusion of windows.h in the code.

I am not sure what's wrong: from your comments, it seems that when SDL_syswm.h is included, it also includes windows.h. I've tried inserting a FORBIDDEN_SYMBOL_EXCEPTION_setjmp before the inclusion of SDL_syswm.h to no avail.

There is a safeguard in common/forbidden.h:

    // mingw-w64 uses [set|long]jmp in system headers
    #ifndef __MINGW64__
    #ifndef FORBIDDEN_SYMBOL_EXCEPTION_setjmp
    #undef setjmp
    #define setjmp(a)	FORBIDDEN_SYMBOL_REPLACEMENT
    #endif

So, since compilation works for you with mingw32, perhaps the safeguard should be extended when compiling with MSVC as well?

PR #573, opened 2015-01-26
AMIGAOS: Re-add a DosList feature

raziel-:
To keep consistency with the original code i re-added the DosList feature dol_Port (which was taken out due to a compiler error using SDK 53.24 and was named wrong anyway).
It offers a message port for the DosList.


digitall:
Looks fine and limited to AmigaOS, and as @raziel- is the porter... Merging.

@raziel- Keep an eye on the buildbot, as if this fails, I may have ask you about upgrading the SDK?

PR #575, opened 2015-01-27
Added Raspberry Pi native 2D API support (dispmanx)

vanfanel:
Added the following objects:

-backends/platform/sdl/raspberrypi/raspberrypi.o
-backends/platform/sdl/raspberrypi/raspberrypi-main.o
-backends/graphics/dispmanxsdl/dispmanxsdl-graphics.o

Hardware scaling using dispmanx is activated by passing "--enable-dispmanx" to the configure script.

Cross-compilation instructions (passing "--host=raspberrypi" to the configure script) are available in http://wiki.scummvm.org/index.php/Compiling_ScummVM/RPI

Also, a double-buffered display using dispmanx resources is provided, thus avoiding tearing in games which is specially evident in scrolling scenes when plain and buggy SDL on FBDEV is used (which was the default on the Raspberry Pi).


sev-:
Could you please reword commits in accordance with our commit guidelines (http://wiki.scummvm.org/index.php/Commit_Guidelines). Particularly add subsytem prefix.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+#if defined(RASPBERRYPI)
+
+#include "backends/graphics/dispmanxsdl/dispmanxsdl-graphics.h"
+#include "graphics/scaler/aspect.h"
+#include "common/mutex.h"
+#include "common/textconsole.h"
+
+#include <bcm_host.h>
+
+#define numpages 2
```
sev-: defines are either kConstantName or CONSTANT_NAME

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.h
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef BACKENDS_GRAPHICS_SDL_DISPMANX_H
+#define BACKENDS_GRAPHICS_SDL_DISPMANX_H
+
+#include "backends/graphics/surfacesdl/surfacesdl-graphics.h"
+
+struct dispvarsStruct;
+struct dispmanxPage;
+
+typedef uint32_t DISPMANX_UPDATE_HANDLE_T;
```
sev-: all uppercase are reserved for constants. Types like this should be are all lowercase.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+   	DISPMANX_RESOURCE_HANDLE_T resource;
+};
+
+DispmanXSdlGraphicsManager::DispmanXSdlGraphicsManager(SdlEventSource *sdlEventSource)
+ : SurfaceSdlGraphicsManager(sdlEventSource) {
+	_dispvars = new dispvarsStruct;
+	DispmanXInit();
+}
+
+DispmanXSdlGraphicsManager::~DispmanXSdlGraphicsManager()
```
sev-: Please move the opening brace to previous line in accordance with our code formatting conventions

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	DispmanXInit();
+}
+
+DispmanXSdlGraphicsManager::~DispmanXSdlGraphicsManager() 
+{
+	DispmanXVideoQuit();
+	delete _dispvars;
+}
+
+void DispmanXSdlGraphicsManager::DispmanXInit () {
```
sev-: Remove space in front of '()'

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	_dispvars->screen = 0;
+	_dispvars->vcImagePtr = 0;
+	_dispvars->pages = (struct dispmanxPage*)calloc(numpages, sizeof(struct dispmanxPage));
```
sev-: space before '*'

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	pthread_mutex_init(&_dispvars->pendingMutex, NULL);
+	pthread_cond_init(&_dispvars->vsyncCondition, NULL);
+	
+	// Before we call any vc_* function, we need to call this one.
+	bcm_host_init();
+	
+	_dispvars->display = vc_dispmanx_display_open(_dispvars->screen);
+}
+
+void DispmanXSdlGraphicsManager::DispmanXSetup (int width, int height, int bpp) {
```
sev-: Same problem with space in front of the opening brace.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	DispmanXFreeResources();
+	vc_dispmanx_display_get_info(_dispvars->display, &(_dispvars->amode));
+
+        _dispvars->pitch = width * (bpp/8);
```
sev-: wrong indentation, surround division sign with spaces.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+		vc_dispmanx_rect_set(&(_dispvars->dstRect), 0, 0, 
+	   		_dispvars->amode.width, _dispvars->amode.height);
+	}
+
+	// We configure the rects now	
+	vc_dispmanx_rect_set(&(_dispvars->bmpRect), 0, 0, width, height);	
+	vc_dispmanx_rect_set(&(_dispvars->srcRect), 0, 0, width << 16, height << 16);	
+
+	int i;
+	for (i = 0; i < numpages; i++)
```
sev-: you may write "for (int i = 0;...)" which is more C++ style

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	
+	// Before we call any vc_* function, we need to call this one.
+	bcm_host_init();
+	
+	_dispvars->display = vc_dispmanx_display_open(_dispvars->screen);
+}
+
+void DispmanXSdlGraphicsManager::DispmanXSetup (int width, int height, int bpp) {
+	DispmanXFreeResources();
+	vc_dispmanx_display_get_info(_dispvars->display, &(_dispvars->amode));
```
sev-: Braces after '&' are not needed here and below, just as you don't have it on lines 87 and 88.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	
+	_dispvars->element = vc_dispmanx_element_add(_dispvars->update, _dispvars->display, 0,
+		&(_dispvars->dstRect), _dispvars->pages[0].resource, &(_dispvars->srcRect), 
+		DISPMANX_PROTECTION_NONE, _dispvars->alpha, 0, (DISPMANX_TRANSFORM_T)0);
+	
+	vc_dispmanx_update_submit_sync(_dispvars->update);		
+}
+
+void DispmanXVSyncCallback (DISPMANX_UPDATE_HANDLE_T u, void * arg){
+	struct dispvarsStruct *_dispvars = (struct dispvarsStruct*) arg;
```
sev-: Correct formatting is "(struct dispvarsStruct *)arg"

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	_dispvars->update = vc_dispmanx_update_start(0);
+	
+	_dispvars->element = vc_dispmanx_element_add(_dispvars->update, _dispvars->display, 0,
+		&(_dispvars->dstRect), _dispvars->pages[0].resource, &(_dispvars->srcRect), 
+		DISPMANX_PROTECTION_NONE, _dispvars->alpha, 0, (DISPMANX_TRANSFORM_T)0);
+	
+	vc_dispmanx_update_submit_sync(_dispvars->update);		
+}
+
+void DispmanXVSyncCallback (DISPMANX_UPDATE_HANDLE_T u, void * arg){
```
sev-: extra space before '*'

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	pthread_mutex_lock(&_dispvars->pendingMutex);
+	_dispvars->pageflipPending = true;	
+	pthread_mutex_unlock(&_dispvars->pendingMutex);
+}
+
+void DispmanXSdlGraphicsManager::DispmanXFreeResources(void) {
+	int i;
+	_dispvars->update = vc_dispmanx_update_start(0);
+		
+    	for (i = 0; i < numpages; i++)
```
sev-: same thing with 'int i'. Move it to "for (int i = 0..."

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.h
```diff
@@ -0,0 +1,52 @@
...
+	DispmanXSdlGraphicsManager(SdlEventSource *sdlEventSource);
+	~DispmanXSdlGraphicsManager(); 
+	bool loadGFXMode();
+	void internUpdateScreen();
+	bool handleScalerHotkeys(Common::KeyCode key);
+	void setFullscreenMode(bool enable);
+	void setAspectRatioCorrection(bool enable);
+protected:	
+	// Raspberry Pi Dispmanx API
+	void DispmanXSetup(int width, int height, int bpp);
```
sev-: please rename them to start from lowercase 'd' to avoid confusion with the class names according to our naming conventions.

sev-:
I haven't reviewed all of it, but there are lots of incosistency with our code formatting guidelines and naming conventions.

Also did you run 'testbed' engine with your backend? This is the way to check correctness of your implementation.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.h
lordhoto: ype names should actually be camel case with an uppercase letter at the start (i.e. see "Type names" at http://wiki.scummvm.org/index.php/Code_Formatting_Conventions#Naming). Hope this helps resolve some confusion and prevents unnecessary work.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+	_videoMode.hardwareWidth = _videoMode.screenWidth * _videoMode.scaleFactor;
+	_videoMode.hardwareHeight = _videoMode.screenHeight * _videoMode.scaleFactor;
+
+	//
+	// Create the surface that contains the 8 bit game data
+	//
+#ifdef USE_RGB_COLOR
+	_screen = SDL_CreateRGBSurface(SDL_SWSURFACE, _videoMode.screenWidth, _videoMode.screenHeight,
+						_screenFormat.bytesPerPixel << 3,
+						((1 << _screenFormat.rBits()) - 1) << _screenFormat.rShift ,
```
sev-: why spaces before commas?

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+		0, 0, 0, 0);
+	
+	_dispvars->pixmem = _hwscreen->pixels;
+
+	// We draw on a RAM surface, but we make this call just to get SDL input initialized. 
+	// Even if we don't use the returned SDL_Surface *, we still need to use the right dimensions
+	// for mouse pointer adjustment to work correctly.
+	SDL_SetVideoMode(_videoMode.screenWidth, _videoMode.screenHeight, 16, SDL_FULLSCREEN);
+
+#ifdef USE_RGB_COLOR
```
sev-: Are you going to support non-RGB COLOR? Is there sense in it on your platform?

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,558 @@
...
+		internUpdateScreen();
+		return true;
+	}
+
+	return true;
+}
+
+void DispmanXSdlGraphicsManager::setFullscreenMode(bool enable) {
+	_videoMode.fullscreen = enable;
+	return;
```
sev-: What does this return statement do here?

configure
```diff
@@ -2980,6 +2986,33 @@ case $_backend in
...
+		cat > $TMPC << EOF
+#include <bcm_host.h>
+
+		int main(int argc, char *argv[]) {
+	bcm_host_init();
+}
+EOF
+		cc_check $DISPMANX_CXXFLAGS $DISPMANX_LIBS && _use_dispmanx=yes
+		if test "$_use_dispmanx" = "yes"; then
+			echo "Activating Raspberry Pi DispmanX graphics backend"
```
sev-: I suggest to have here 'echon' and echo 'done' when successful or error message on failure

sev-:
Okay, I finished reviewing.

What is missing:

1. Target for building binary packages, so it could be put on the platform
2. Platform-specific README which will mention which tools are required, how to run the resulting binaries and probably point to the Wiki with compilation instructions which you've created.

wjp:
My personal main concern is the significant amount of duplication of non-trivial code from surfacesdl.

vanfanel:
This backend has been updated: I'd like it to be merged as it's a great benefit for Raspberry Pi users, greatly improving visuals and performance.
Code duplication from the surfacesdl class has been drastically reducted since all scaling is done by dispmanx, not SDL: only some intermediate SDL software surfaces are used before the final rendering to screen, as in any other sdl-based scummvm graphics backend.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,585 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+//Needed for Raspberry Pi header incussion
```
fingolfin: Typo: "inclusion"; also, a space after `//` should be inserted

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,585 @@
...
+void DispmanXSdlGraphicsManager::internUpdateScreen() {
+	SDL_Surface *srcSurf, *origSurf;
+	int height, width;
+
+	// If the shake position changed, fill the dirty area with blackness
+	if (_currentShakePos != _newShakePos ||
+		(_mouseNeedsRedraw && _mouseBackup.y <= _currentShakePos)) {
+		SDL_Rect blackrect = {0, 0, (Uint16)(_videoMode.screenWidth * _videoMode.scaleFactor), (Uint16)(_newShakePos * _videoMode.scaleFactor)};
+
+		if (_videoMode.aspectRatioCorrection && !_overlayVisible)
```
fingolfin: That seems wrong -- either `dispvars->aspectRatioCorrection` should be used, or this code is dead, no?

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.h
```diff
@@ -0,0 +1,52 @@
...
+	void clearOverlay();
+protected:	
+	// Raspberry Pi Dispmanx API
+	void DispmanXSetup(int width, int height);
+	void DispmanXInit();
+	void DispmanXUpdate();
+	struct dispmanxPage *DispmanXGetFreePage();
+	void DispmanXFreeResources();
+	void DispmanXVideoQuit();
+	struct dispvarsStruct *_dispvars;
```
fingolfin: The `struct` here is superfluous in C++, putting it in there is a sign of C code. The same applies to several other places in the code.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,585 @@
...
+#endif
+		internUpdateScreen();
+		return true;
+	}
+
+	return true;
+}
+
+void DispmanXSdlGraphicsManager::setFullscreenMode(bool enable) {
+	_videoMode.fullscreen = enable;
```
fingolfin: Does this code actually support fullscreen mode? If not, this method should instead ignore the enable value, and the `hasFeature` method should be overwritten suitably.

Of course, on the long run, properly refactoring `SurfaceSdlGraphicsManager` would be the proper solution, but that's a bigger task.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,585 @@
...
+	_numDirtyRects = 0;
+	_forceFull = false;
+	_mouseNeedsRedraw = false;
+}
+
+bool DispmanXSdlGraphicsManager::handleScalerHotkeys(Common::KeyCode key) {
+
+	// Ctrl-Alt-a toggles aspect ratio correction
+	if (key == 'a') {
+		beginGFXTransaction();
```
fingolfin: Since you are bypassing `_videoMode`, the GFX transaction code won't work properly here anyway. So I am not quite sure if this code is correct... it very well might be, but somebody should perhaps carefully think through it (I didn't, I just glanced at this code and wondered).

backends/platform/sdl/raspberrypi/README.RASPBERRYPI
```diff
@@ -0,0 +1,100 @@
...
+==============================================================================
+
+Notes
+============
+
+This version of ScummVM is specially tailored to use DispmanX, the native 2D
+API on the Raspberry Pi. The idea is that scaling and drawing on a double
+buffer with a non-blocking vsync wait is all done using the on-board VideoCore
+hardware, thus using only a small fraction of the CPU ScummVM uses when ran
+on a clunky, software-scaled and desynced X11 enviroment using the X11 API.
```
fingolfin: `environment`

backends/platform/sdl/raspberrypi/README.RASPBERRYPI
```diff
@@ -0,0 +1,100 @@
...
+./configure ./configure --backend=raspberrypi -disable-debug --enable-release 
+--enable-optimizations --disable-mt32emu --disable-flac --disable-mad --disable-vorbis 
+--disable-tremor --disable-fluidsynth --disable-taskbar --disable-timidity --disable-alsa
+
+make
+
+¡¡It will be an SLOW process, taking several hours to complete, unless you
+are running distcc against a fast compilation server!!
+
+2) If we wandt to build by cross-compiling on a GNU/Linux X86-based computer,
```
fingolfin: `want` ?

backends/platform/sdl/raspberrypi/README.RASPBERRYPI
```diff
@@ -0,0 +1,100 @@
...
+class PC or in our Raspberry Pi:
+
+1) Building on the Raspberry Pi itself, although possible, is an SLOW task for the
+little computer unless you use distributed gcc (or distcc for short).
+
+Local compilation would simply consist of the "standard" GNU/Linux building process:
+
+cd <sources_dir>
+
+./configure ./configure --backend=raspberrypi -disable-debug --enable-release
```
fingolfin: Two `./configure` is one too many :)

configure
```diff
@@ -1048,6 +1052,8 @@ for ac_option in $@; do
 	--disable-libunity)       _libunity=no    ;;
 	--enable-opengl)          _opengl=yes     ;;
 	--disable-opengl)         _opengl=no      ;;
+	--enable-dispmanx)	  _dispmanx=yes   ;;
+	--disable-dispmanx)	  _dispmanx=no    ;;
```
fingolfin: These two lines are using tabs where they should be using spaces (matching the surrounding code). Hmm, I guess the Code Formatting Conventions in the ScummVM Wiki should mention that (at least I just didn't find it). But as  a rule of thumb, whenever submitting code for some project, it helps to adapt your code changes to "blend in" with the rest. One trick to notice such problems with tabs is to temporarily set your local tab width to 20 or so, then problematic spots stand out. Another is to grep for places where a tab occurs next to a regular space.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,585 @@
...
+#include "common/textconsole.h"
+
+#include <bcm_host.h>
+
+struct dispvarsStruct { 
+	DISPMANX_DISPLAY_HANDLE_T display;
+	DISPMANX_UPDATE_HANDLE_T update;
+	DISPMANX_ELEMENT_HANDLE_T element;
+	VC_IMAGE_TYPE_T pixFormat;
+	VC_DISPMANX_ALPHA_T alpha;
```
fingolfin: The above line (and multiple others) has trailing whitespace, which stands out in the output of `git diff`. Many projects have code formatting conventions that prohibit this; alas, I just looked at the ScummVM CFC and it doesn't seem to require this. So I guess removing these is not required, but I thought I should mention it anyways, just in case.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.h
```diff
@@ -0,0 +1,52 @@
...
+	void setAspectRatioCorrection(bool enable);
+	void clearOverlay();
+protected:	
+	// Raspberry Pi Dispmanx API
+	void DispmanXSetup(int width, int height);
+	void DispmanXInit();
+	void DispmanXUpdate();
+	struct dispmanxPage *DispmanXGetFreePage();
+	void DispmanXFreeResources();
+	void DispmanXVideoQuit();
```
fingolfin: Method names should start with a lowercase letter.

configure
```diff
@@ -1294,6 +1300,12 @@ arm-riscos)
 	_host_os=riscos
 	_host_cpu=arm
 	;;
+raspberrypi)
+	_host_os=linux
+	_host_cpu=arm
+	_host_alias=arm-linux-gnueabihf
+	_backend=raspberrypi
```
fingolfin: That's the wrong place to set _backend (if you look around, you'll see that no other case does it here. Instead, it should be set in the big case selection starting somewhere around line 2485.

configure
```diff
@@ -2980,6 +2992,8 @@ case $_backend in
 		LIBS="$LIBS -lpng"
 		LIBS="$LIBS -Wl,-Map,mapfile.txt"
 		;;
+	raspberrypi)
+		;;
 	samsungtv)
```
fingolfin: This is where you should set DEFINES, INCLUDES and LIBS (again: take a look at the surrounding code). Also, make sure to imitate what the surrouding code does to add to these vars: It adds one entry per line. E.g. it does
```
		LDFLAGS="$LDFLAGS -shared"
		LDFLAGS="$LDFLAGS -fpic"
``
This way, when a flag is added or removed, it shows up nicely in diffs.

File: configure [hunk 31]
```diff
@@ -2980,6 +2992,8 @@ case $_backend in
 		LIBS="$LIBS -lpng"
 		LIBS="$LIBS -Wl,-Map,mapfile.txt"
 		;;
+	raspberrypi)
+		;;
```
fingolfin: This is where you should set DEFINES, INCLUDES and LIBS (again: take a look at the surrounding code). Also, make sure to imitate what the surrouding code does to add to these vars: It adds one entry per line. E.g. it does
```
		LDFLAGS="$LDFLAGS -shared"
		LDFLAGS="$LDFLAGS -fpic"
```
This way, when a flag is added or removed, it shows up nicely in diffs. It also makes it easier to recognize similarities between backends and refactor them suitably.

configure
```diff
@@ -3115,6 +3129,22 @@ case $_backend in
...
+# In raspberry Pi, we don't use find_sdlconfig since we could be crosscompiling, but still we use SDL
+#
+case $_backend in
+	raspberrypi)
+		INCLUDES="$INCLUDES -I$RPI_ROOTDIR/usr/include/SDL -D_GNU_SOURCE=1 -D_REENTRANT"
+		LIBS="$LIBS -L$RPI_ROOTDIR/usr/lib -L$RPI_ROOTDIR/usr/lib/arm-linux-gnueabihf -lSDL"
+		DEFINES="$DEFINES -DSDL_BACKEND"
+		add_line_to_config_mk "SDL_BACKEND = 1"
+		_16bit=yes
+		_savegame_timestamp=no
```
fingolfin: Don't set `_savegame_timestamp` here -- set it in the case just before this, which already disables timestampes for `gph | n64`.

configure
```diff
@@ -3115,6 +3129,22 @@ case $_backend in
 		;;
 esac
 
+#
+# In raspberry Pi, we don't use find_sdlconfig since we could be crosscompiling, but still we use SDL
+#
```
fingolfin: I'd argue that in a properly setup cross compiling environment, the configure script should actually be able to use `sdl-config` -- namely, it should "see" the sdl-config for the cross compiling SDL, and get the right linker flags from it.

I think that's what several existing backends do, e.g. I assume (but do not know for sure) that the OpenPandora and SamsungTV backends, which do use find_sdlconfig, are cross-compiled.

But if that is impossible for you for some reason, you should at least imitate the other backends, and setup the require DEFINES and `add_line_to_config_mk`  the way they do it -- in this case, webos and wince (search for `SDL_BACKEND` in this file to see what I mean)

configure
```diff
@@ -4049,6 +4079,31 @@ fi
 define_in_config_if_yes "$_opengl" "USE_OPENGL"
 define_in_config_if_yes "$_opengles" "USE_GLES"
 
+# Check if Raspberry Pi dispmanx has been activated. 
+if test "$_dispmanx" = "yes" ; then
+		echocheck "DispmanX graphics"
+		_use_dispmanx=no
+		DISPMANX_CXXFLAGS="-I$RPI_ROOTDIR/opt/vc/include -I$RPI_ROOTDIR/opt/vc/include/interface/vmcs_host/linux/ -I$RPI_ROOTDIR/opt/vc/include/interface/vcos/pthreads -mfpu=vfp -mfloat-abi=hard -I$RPI_ROOTDIR/opt/rpi_root/usr/include/SDL"
+		DISPMANX_LIBS="--sysroot=$RPI_ROOTDIR -L$RPI_ROOTDIR/usr/lib -L$RPI_ROOTDIR/opt/vc/lib -lbcm_host -lvcos -lvchiq_arm"
+		cat > $TMPC << EOF
```
fingolfin: You use `RPI_ROOTDIR` here and elsewhere in a cruicial role, but your README does not mention it. That seems weird.

configure
```diff
@@ -4049,6 +4079,31 @@ fi
...
+EOF
+		cc_check $DISPMANX_CXXFLAGS $DISPMANX_LIBS && _use_dispmanx=yes
+		if test "$_use_dispmanx" = "yes"; then
+			CXXFLAGS="$CXXFLAGS $DISPMANX_CXXFLAGS"
+			LIBS="$LIBS $DISPMANX_LIBS"
+			MODULES="$MODULES backends/platform/sdl"
+			DEFINES="$DEFINES -DRASPBERRYPI"
+			add_line_to_config_mk 'RASPBERRYPI = 1'
+			_opengl=no
+			_opengles=no
```
fingolfin: Is this really necessary? If yes, then it should be done *before* ScummVM runs tests for OpenGL, and reports their results (i.e. way earlier).

fingolfin: In fact, it seems that the result of this detection step affect which backend should be used -- if this check fails, your custom raspberrypi backend does not build. Yet you hard code that `_backend` value earlier (I already commented on that). Thus if dispmanx cannot be found, should't you either error out, or fallback to the regular SDL backend? Otherwise, the user will end up with a broken build environment with no visible hint that something is wrong, except for a tiny message that dispmanx was not found.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
vanfanel: This code ONLY supports fullscreen, framebuffer-like graphics, so this is ok as it is, I think.

fingolfin: Actually, kFeatureFullscreenMode indicates that the backend supports *switching* between windowed and fullscreen mode. If the backend only supports one or the other, then `hasFeature(kFeatureFullscreenMode)` should return false. At least that's how I read the documentation in `common/system.h`. Then again, a quick grep shows that nothing in ScummVM seems to call `hasFeature(kFeatureFullscreenMode)` (I thought it might disable the "fullscreen" option in the options dialog, but that doesn't seem to be the case?). So it probably doesn't matter much beyond doing what is "formally right". :-)

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
vanfanel: You're right, the setFeatureState() call is enough here.

configure
vanfanel: That's because the README points users and devs that want to crossbuild to the wiki instructions. I can include the wiki instructions for crosscompiling in the README, but I think it's a specialized building method so I thought it would be out of it's scope.

fingolfin: Fair enough (although personally I like it when such things are in an README, so that I can lookup stuff even when offline; but since I am hardly ever offline these days, I guess I don't really mind either way ;-). Not that it would matter much if I did mind ;-).

fingolfin:
Oh and one more remark: The placement of the backend in `backends/platform/sdl/raspberrypi` is incorrect, it should be `backends/platform/raspberrypi`, as it specifies a custom `_backend` value in `configure`. And in line 3057 of `configure`, it says:
```
MODULES="$MODULES backends/platform/$_backend"
```
This one actually had me confused a bit, as I wondered what the difference between the two kinds of SDL based backends is -- on the one hand, the AmigaOS / Mac OS X / POSIX / PS3 / Win32 variants (which use `_backend=sdl` and are contained in subdirs of the `sdl` dir), and on the other hand dingux, gph, maemo, openpandora, .... (which use custom `_backend` values, and live outside the `sdl` dir). Well, what I just described is the only pattern I could really see. Otherwise, it seems somewhat arbitrary in which "category" a backend is placed... Maybe those in `sdl` are meant to be those backends which only contain minimal modifications -- but then e.g. the SamsungTV and linuxmoto backends probably would fit in there, too? Perhaps this should eventually be addressed by a refactoring of the ScummVM backends...

But until then, I'd say it would seem like a good idea to stick to the existing pattern.

vanfanel:
@fingolfin: I've updated with all your corrections. However, note there isn't a custom _backend anymore, so the backends/platform/sdl/raspberrypi route is correct now considering your previous note.

configure
```diff
@@ -2537,6 +2548,17 @@ if test -n "$_host"; then
 			_seq_midi=no
 			_port_mk="backends/platform/dingux/dingux.mk"
 			;;
+		raspberrypi)
+			# We should have setup raspberrypi-sdl-config with modified prefix as part of 
+			# the cross-building enviroment.
```
fingolfin: That statement is rather ambiguous (and contains a typos ;-). Perhaps you meant to say something like this? `For proper cross-compilation, we assume that the cross-building environment provides a raspberrypi-sdl-config executable with modified prefix.` or so?

But even then it is not completely clear to me whether this is appropriate. E.g. is it customary that such an executable und that name is provided? At least in the past, most cross environments I dealt with just called it `sdl-config` (and it seems that on the ScummVM buildbot, this is still the case, see e.g. http://buildbot.scummvm.org/builders/master-gp2xwiz/builds/6166/steps/configure/logs/stdio )

But if that's really as the comment suggestings, then `find_sdlconfig` won't find `raspberrypi-sdl-config`. For that, you'd have to do something similar to this (taken from `configure`):
```
		*mingw32*)
			_sdlconfig=$_host-sdl-config
```

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,589 @@
...
+	_dispvars->element = vc_dispmanx_element_add(
+		_dispvars->update,_dispvars->display, 0,
+		&_dispvars->dstRect, 0,
+		&_dispvars->srcRect, DISPMANX_PROTECTION_NONE,
+		&_dispvars->alpha, 0, (DISPMANX_TRANSFORM_T)0);
+
+	vc_dispmanx_update_submit_sync(_dispvars->update);
+}
+
+void dispmanXVSyncCallback (DISPMANX_UPDATE_HANDLE_T u, void *arg) {
```
fingolfin: Minor: There shouldn't be a space before `(`.

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,589 @@
...
+	_dispvars->pageflipPending++;
+	pthread_mutex_unlock(&_dispvars->pendingMutex);
+}
+
+struct dispmanxPage *DispmanXSdlGraphicsManager::dispmanXGetFreePage(void) {
+	struct dispmanxPage *page = NULL;
+
+	while (!page)
+	{
+		// Try to find a free page
```
fingolfin: Minor: opening brace shouldn't be on separate line according to ScummVM CFC. This occurs multiple times later on, I'll not bother you with pointing out each single one :).

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+	_screen = SDL_CreateRGBSurface(SDL_SWSURFACE, _videoMode.screenWidth, _videoMode.screenHeight,
+		_screenFormat.bytesPerPixel << 3,
+		((1 << _screenFormat.rBits()) - 1) << _screenFormat.rShift ,
+		((1 << _screenFormat.gBits()) - 1) << _screenFormat.gShift ,
+		((1 << _screenFormat.bBits()) - 1) << _screenFormat.bShift ,
+		((1 << _screenFormat.aBits()) - 1) << _screenFormat.aShift );
+		if (_screen == NULL)
+			error("allocating _screen failed");
+	// Avoid having SDL_SRCALPHA set even if we supplied an alpha-channel in the format.
```
fingolfin: Something is wrong with the indention here; the `if` statement seems to be one level too deep?

fingolfin:
Overall this looks pretty good to me now, but I'll point out a few more things (mostly minor).

backends/graphics/dispmanxsdl/dispmanxsdl-graphics.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+bool DispmanXSdlGraphicsManager::hasFeature(OSystem::Feature f) {
+	if (f == OSystem::kFeatureFullscreenMode) {
+		return false;
+	}
+}
+
+void DispmanXSdlGraphicsManager::setFullscreenMode(bool enable) {
+	_videoMode.fullscreen = enable;
+}
```
fingolfin: Since fullscreen mode is always one, I would expect this function to do nothing, i.e. ignore the enable/disable value.

However, that might be wrong -- I am not sure whether the ScummVM Options dialog in the GUI properly deals with that case?!

fingolfin:
Regarding code duplication: It is true that there is still quite some, most notably in `internUpdateScreen` and `loadGFXMode`. While this is true, one should point out that several existing SDL based backends do the same or worse. For example, some places that duplicate code from `internUpdateScreen`:
```
$ git grep "If the shake position changed, fill the dirty"
backends/graphics/dinguxsdl/dinguxsdl-graphics.cpp:     // If the shake position changed, fill the dirty area with blackness
backends/graphics/gph/gph-graphics.cpp: // If the shake position changed, fill the dirty area with blackness
backends/graphics/linuxmotosdl/linuxmotosdl-graphics.cpp:       // If the shake position changed, fill the dirty area with blackness
backends/graphics/surfacesdl/surfacesdl-graphics.cpp:   // If the shake position changed, fill the dirty area with blackness
backends/graphics/wincesdl/wincesdl-graphics.cpp:       // If the shake position changed, fill the dirty area with blackness
```
So it would seem unfair to hold this backend back for something that lots of other backends do. At the same time, it might be a good idea to carefully look at this problem, and start to refactor things a bit to make it less bad.

vanfanel:
I updated again with more corrections as sugested by fingolfin.

djwillis:
I'll try and do some tests but this is in a better state then some ports we have in tree (including my own) and it would be nice to get this merged into mainline and offer an optimised RPi port in the next major release.

The wider question of reducing duplication and refactoring some of the backend code is a good one but may be best taken up outside the pull request.

fingolfin:
@djwillis I totally agree, for what little it may be worth :).

vanfanel:
@fingolfin: any news on the merging of this? Seemed to be ready.

wjp:
Can you explain the relationship between the raspberrypi backend and dispmanx support?

I'm very confused by what configure is doing. E.g., detecting dispmanx seems independent of the raspberrypi backend, but then when it finds dispmanx, it enables a RASPBERRYPI define, which among other things seems to affect raspberrypi-main.cpp.

configure
```diff
@@ -944,6 +946,8 @@ Optional Libraries:
   --with-opengl-prefix=DIR Prefix where OpenGL (ES) is installed (optional)
   --disable-opengl         disable OpenGL (ES) support [autodetect]
 
+  --with-dispmanx          enable dispmanx (Raspberry Pi native 2D API) rendering
```
lordhoto: This switch doesn't seem to be handled? I.e. it should be --enable-dispmanx? (How does this flag affect anything btw.? Wouldn't dispmanx still only be used when the backend/host is raspberrypi?)

configure
```diff
@@ -1048,7 +1052,9 @@ for ac_option in $@; do
 	--disable-libunity)       _libunity=no    ;;
 	--enable-opengl)          _opengl=yes     ;;
 	--disable-opengl)         _opengl=no      ;;
-	--enable-bink)            _bink=yes       ;;
+        --enable-dispmanx)        _dispmanx=yes   ;;
```
lordhoto: Formatting in this and the following two lines is broken? Did you use tabs instead of spaces or something like that?

configure
```diff
@@ -4049,6 +4074,28 @@ fi
...
+		int main(int argc, char *argv[]) {
+	bcm_host_init();
+}
+EOF
+		cc_check $DISPMANX_CXXFLAGS $DISPMANX_LIBS && _use_dispmanx=yes
+		if test "$_use_dispmanx" = "yes"; then
+			CXXFLAGS="$CXXFLAGS $DISPMANX_CXXFLAGS"
+			LIBS="$LIBS $DISPMANX_LIBS"
+			DEFINES="$DEFINES -DRASPBERRYPI"
+			add_line_to_config_mk 'RASPBERRYPI = 1'
```
lordhoto: Is this enabling the whole RaspberryPI backend? If so that looks really strange and should rather be decided based on the _backend setting, i.e. --backend.

configure
```diff
@@ -2537,6 +2548,17 @@ if test -n "$_host"; then
 			_seq_midi=no
 			_port_mk="backends/platform/dingux/dingux.mk"
 			;;
+		raspberrypi)
+			# For proper cross-compilation, we assume that the cross-building environment provides
```
lordhoto: For cross compilation we usually pick up something like host-sdl-config which should have the appropriate prefix, which does not necessarily need to concide with $RPI_ROOT/usr. Please check that this works as expected.

lordhoto:
It seems that compilation on a RPi is broken at the moment. We have someone in IRC reporting that (see http://logs.scummvm.org/log.php?log=scummvm.log.23Aug2015&format=html at [12:59]).

I have the feeling that --backend=raspberrypi does not properly handle that it is still an SDL derived backend and thus sdl-config is not picked up/checked. This makes compilation fail for that guy. This needs to be fixed.

configure
vanfanel: No more dispmanx backend. The backend is clearly the SDL one. Please, forget about backend: only "_host" has sense in this scenario.

configure
vanfanel: This was removed. Sorry, of course it made no sense. It's handled by --enable-dispmanx as you supposed.

vanfanel:
I have updated the configure script and building instructions on the howto. It should work way better now for cross-compiling in every system if the instructions are followed.

vanfanel:
This is ready for merging I think. I know I will have to do some additional adjustments to the configure script after merging, but I need it to be merged to continue. It cross-builds fine if instructions are followed.

rootfather:
Any news on this?

sev-:
I think it is ready now. If there are no objections, I'll merge this within a week, e.g. deadline for feedback is Nov 14th.

rootfather:
Great news!

Furthermore, having an official port of ScummVM for the Raspberry Pi series would be a great addition to the project. Vanafel, I'm counting on you! :)

wjp:
@vanfanel Could you merge master into this branch, so we can look at fixing up the last small conflicts? (In particular, whitespace problems in configure and the use of the recently introduced append_var)

wjp:
Thanks

vanfanel:
@sev- 
Can you explain me what do you mean by adding myself to the Credits in this patch?
What file should I add my name to?

criezy:
@vanfanel 
For credits you edit devtools/credits.pl. I guess you will want to add your name in the "Backend Teams" section (starting around line 800).
Then you can update other files by running 'make credits'

vanfanel:
@criezy , @sev- : Added my name to credits.pl
I guess I should setup the automated building process on the buildbot. Please indicate the next steps I should take.

PR #578, opened 2015-01-29
MORTEVIELLE: added DOS font support

chrilith:
Hi,

This is addition to a previous pull request. A DOS font has been aded for proper starting screen rendering.
The related screen has been updated to use this new font.

Thanks!
Chris


sev-:
If it is a real standard IBM DOS font, I would rather add that to the common code for reusability

urukgit:
sev- has a point. That would be really useful! :)

chrilith:
Hi sev,

I believe this font data is accurate enough, checked against http://www.dafont.com/fr/perfect-dos-vga-437.font font which seems to have a pixel perfect implementation for VGA (EGA seems to be a bit different in size looking at the Mortevielle title screen). The DOSBox version is a bit different with larger M,W etc.

I indeed thought about adding it to the common code but don't know where to start. Maybe a common font renderer would be nice to have for DOS/Atari/Amiga/Apple II... but don't remember enough about ScummVM code for now to add a proper renderer based on graphic surfaces format :).

sev-:
See graphics/fonts

chrilith:
OK, done. I'll review my code and send another pull request then with the DOS font managed by the FontMan.

Thanks for the tip.

chrilith:
New pull request #580 with common DOS font support.

bluegr:
Since there's an updated pull request, I'm closing this one

PR #581, opened 2015-01-30
ANDROID: Switch minimum & recommended API Level to 9 (Android 2.3).

digitall:
Also, add maximum API Level of 18 (Android 4.3) "Jelly Bean".

This is not used by Android so will still allow manual installation of
APKs to devices with newer versions, but is used by the Google Play
Store to filter versions offered to users.

This will reduce complaints from users with newer versions of Android.

### DIFF
--- a/dists/android/AndroidManifest.xml
+++ b/dists/android/AndroidManifest.xml
@@ -8,10 +8,11 @@
 		android:installLocation="preferExternal"
 		android:sharedUserId="org.scummvm.scummvm">
 
-	<!-- This version works on Android 1.5 (SDK 3) and newer, but we
-			want Android 2.2 (SDK 8) defaults and features. -->
-	<uses-sdk android:minSdkVersion="3"
-			android:targetSdkVersion="8"/>
+	<!-- This version works on Android 2.3 (API Level 9) onwards,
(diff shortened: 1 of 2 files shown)


digitall:
@fuzzie Happy to merge? If so, could you update the Google Play Store released versions or do we need to do a v1.7.1 for that?

clone2727:
Why would we set the maximum version?

digitall:
@clone2727 "This is not used by Android so will still allow manual installation of
APKs to devices with newer versions, but is used by the Google Play
Store to filter versions offered to users."

Because the newer versions fail to work correctly or are at least untested on newer versions of Android leading to unhappy and complaining users. See http://forums.scummvm.org/viewforum.php?f=17

This only affects Google Play store releases as previously indicated, and we will increment the maxSdkVersion as newer versions of Android vecome more common, stable and tested with ScummVM.

clone2727:
@digitall I fixed the bug and cherry picked it to the 1.7 branch. The maximum version should not be set.

digitall:
Why not?

lordhoto:
Closing this since this is targetting an out-dated release branch.

PR #582, opened 2015-01-30
ANDROID: Switch minimum & recommended API Level to 9 (Android 2.3).

digitall:
Also, add maximum API Level of 18 (Android 4.3) "Jelly Bean".

This is not used by Android so will still allow manual installation of
APKs to devices with newer versions, but is used by the Google Play
Store to filter versions offered to users.

This will reduce complaints from users with newer versions of Android.

### DIFF
--- a/dists/android/AndroidManifest.xml
+++ b/dists/android/AndroidManifest.xml
@@ -7,10 +7,11 @@
 		android:versionName="1.8.0git"
 		android:sharedUserId="org.scummvm.scummvm">
 
-	<!-- This version works on Android 1.5 (SDK 3) and newer, but we
-			want Android 2.2 (SDK 8) defaults and features. -->
-	<uses-sdk android:minSdkVersion="3"
-			android:targetSdkVersion="8"/>
+	<!-- This version works on Android 2.3 (API Level 9) onwards,
+	     but not correctly on Android 4.4 (API Level 19) onwards. -->
+	<uses-sdk android:minSdkVersion="9"
+			android:targetSdkVersion="9"
+			android:maxSdkVersion="18" />
 
 	<application
 			android:label="@string/app_name"
--- a/dists/android/AndroidManifest.xml.in
+++ b/dists/android/AndroidManifest.xml.in
@@ -7,10 +7,11 @@
 		android:versionName="@VERSION@"
 		android:sharedUserId="org.scummvm.scummvm">
 
-	<!-- This version works on Android 1.5 (SDK 3) and newer, but we
-			want Android 2.2 (SDK 8) defaults and features. -->
-	<uses-sdk android:minSdkVersion="3"
-			android:targetSdkVersion="8"/>
+	<!-- This version works on Android 2.3 (API Level 9) onwards,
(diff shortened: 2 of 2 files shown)


digitall:
@fuzzie Happy to merge?

fuzzie:
On Fri, Jan 30, 2015 at 04:45:43AM -0800, David Turner wrote:
> @fuzzie Happy to merge?

Doesn't this break functionality like the menu button? Why is it needed?

fuzzie:
Sorry, wrote that in a hurry on the train. Targetting Android 2.3 is fine afaik. What in master is broken on Android >4.3? Forcing us to push new apks (and annoying all the users, since ScummVM is not tiny) for each new Android version seems like a bad idea.

digitall:
@fuzzie : To deal with the second point first "Forcing us to push new apks (and annoying all the users, since ScummVM is not tiny) for each new Android version seems like a bad idea." 

Yes, it does, but I don't think that would be necessary.

I'm not sure how exactly the Play Store submission works, but my point with this is that we set sdkMaxVersion to the latest revision of Android which we have tested with and support when the version of ScummVM is released i.e. v1.7.1, v1.8.0 etc.

Since maxSdkVersion does not actually stop the APKs running on newer versions of Android, any user with a later device can always use the APKs from the download page and side load them... or buildbot APKs.

The point is that we had no way to test with future versions of Android i.e. v5.0 when we were releasing v1.7.0 and it is broken on that, and this could occur again in the future...

By setting this and updating for future releases, this will prevent the build being offered in the Google Play Store as "stable" when it is in fact totally untested on their Android version.

digitall:
@fuzzie : As for your first point, we have a fair number of complaints and reports of breakage on Android v4.4 (Kitkat) - API Level 19 and 20 and Android v5.0 (Lollipop) - API Level 21 on the forums. See:
http://forums.scummvm.org/viewtopic.php?t=13732
http://forums.scummvm.org/viewtopic.php?t=13762
http://forums.scummvm.org/viewtopic.php?t=13748

This is related to the changes in SD Card File Handling / Permission amongst other things...

My intent is just to prevent future releases from having these kind of issues... It is one thing if the user wants to install a development build from buildbot or APKs from download, that indicates a certain level of technical capability and willingless to deal with bugs... but offering a version to normal users in the Play Store gets us complaints.

digitall:
https://developer.android.com/guide/topics/manifest/uses-sdk-element.html covers this.

It does say of sdkMaxVersion :
Warning: Declaring this attribute is not recommended. First, there is no need to set the attribute as means of  blocking deployment of your application onto new versions of the Android platform as they are released. By design, new versions of the platform are fully backward-compatible.

It is just a pity that the fully backward-compatible bit seems to be wishful thinking on Google's part... hence why this has not been removed.

sev-:
I have no clue about Android development, so need to rely on your expertise for this and previous patch.

fuzzie:
I still think we shouldn't set sdkMaxVersion. The other part is fine, but I guess unnecessary for now?

sev-:
This is now being handled by Android-SDL port.

PR #583, opened 2015-01-30
Update AndroidManifest as requested by digitall

widar:
Don't ask.


digitall:
@widar : Thanks, but this is already covered by Pull Request #581 which I opened earlier :
https://github.com/scummvm/scummvm/pull/581

If you want to comment on the discussion there, feel free.

PR #584, opened 2015-02-01
COMMON: Add getRevTab to FFT

bgK:
This change to the FFT class is needed by the WMA decoder from xoreos we intend to merge in ResidualVM in PR https://github.com/residualvm/residualvm/pull/1128.
The common code in ResidualVM is updated from time to time from ScummVM. The intent behind this PR is not to introduce differences for the FFT class between ScummVM and ResidualVM to keep the update process simple for this class.
The changes are simple and should not prove difficult to maintain.

### DIFF
--- a/common/fft.cpp
+++ b/common/fft.cpp
@@ -61,6 +61,10 @@ FFT::~FFT() {
 	delete[] _tmpBuf;
 }
 
+const uint16 *FFT::getRevTab() const {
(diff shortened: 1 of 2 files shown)


sev-:
15 minutes before the merge. Fascinating

clone2727:
@sev- Is there a problem?

PR #585, opened 2015-02-10
Add: Pause / Break functionality like in the original scumm and SCI executables

andoruB:
First of all I apologise in advance if I do anything silly, my first time making a pull request.  

As I've explained in this [forum post](http://forums.scummvm.org/viewtopic.php?p=81195), I could make use of the Pause / Break functionality that was present in the original SCUMM and SCI games, so that when you would press the Pause key (reported as key 19 in ScummVM here), the game would be halted, but the background music would still play.
In the meantime, I'll try eriktorbjorn's suggestion :)


lordhoto:
It looks like this doesn't contain any changes and tries to merge branch-1-7-0 changes into master? Closing this.

PR #587, opened 2015-02-23
WINTERMUTE: Added mappings for VKeyCodes->ScummVM KeyCodes

jammm:
Fix bug #6654 (white chamber - some keys don't work)

Wintermute games on ScummVM used SDL keycodes for keyboard mapping, wheras
the game scripts only accept VKeyCodes of Windows. Therefore an initial
set of mappings are added and the debug room is now loading properly, when
HOME is pressed (Rest of the keys need to be tested in-game)

### DIFF
--- a/engines/wintermute/base/base_keyboard_state.cpp
+++ b/engines/wintermute/base/base_keyboard_state.cpp
@@ -262,41 +262,92 @@ bool BaseKeyboardState::isCurrentPrintable() const {
 }
 
 //////////////////////////////////////////////////////////////////////////
+enum VKeyCodes {
+	kVkReturn     = 13,
+
+	kVkEscape     = 27,
+
+	kVkSpace      = 32,
+	kVkEnd        = 35,
+	kVkHome       = 36,
+	kVkLeft       = 37,
+	kVkUp         = 38,
+	kVkRight      = 39,
+	kVkDown       = 40,
+
+	kVkF1         = 112,
+	kVkF2         = 113,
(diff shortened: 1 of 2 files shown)


engines/wintermute/base/base_keyboard_state.cpp
```diff
@@ -262,41 +262,91 @@ bool BaseKeyboardState::isCurrentPrintable() const {
...
+
+//////////////////////////////////////////////////////////////////////////
 uint32 BaseKeyboardState::keyCodeToVKey(Common::Event *event) {
+	// todo
 	if (event->type != Common::EVENT_KEYDOWN) {
 		return 0;
 	}
 
 	switch (event->kbd.keycode) {
-	case Common::KEYCODE_KP_ENTER:
```
somaen: Are we not handling the keypad-enter key any more?

somaen:
Looks fine, although I'm not sure about removing the handling of KEYCODE_KP_ENTER, as I assume there was a reason why that was the only handled KEYCODE earlier (sure it was handled wrongly, but...)

engines/wintermute/base/base_keyboard_state.cpp
jammm: Doesn't seem to be used so far, but hopefully someone will raise an issue about it if it's required.

jammm:
I've added the handling of the enter key, I could add the keypad enter key if you want.

jammm:
Added the keypad-enter key handling.

PR #589, opened 2015-03-01
DRASCULA: fix crash/graphic glitch at castle towers

norbertbajko:
Fix bug #5903 DRASCULA-IT: Crash/graphic glitch at castle towers
http://sourceforge.net/p/scummvm/bugs/5903/

The problem was that if you looked at the tower Hacker's head started flying on the screen with the subtitle and the game crashed.
Now I fixed it: the game is not crash, there is no flying head and the subtitle is fixed on the screen upper-center.

withoutheadpluscenter

### DIFF
--- a/engines/drascula/actors.cpp
+++ b/engines/drascula/actors.cpp
@@ -396,6 +396,13 @@ void DrasculaEngine::increaseFrameNum() {
 		curHeight = (int)newHeight;
 		curWidth = (int)newWidth;
 	}
+
+	if ((currentChapter == 5) && (_roomNumber == 45)) {
+		curY = 0;
+		curX = 0;
+		curHeight = 0;
+		curWidth = 0;
+	}
 }
 
 void DrasculaEngine::walkDown() {
--- a/engines/drascula/actors.cpp
+++ b/engines/drascula/actors.cpp
@@ -396,7 +396,10 @@ void DrasculaEngine::increaseFrameNum() {
 		curHeight = (int)newHeight;
 		curWidth = (int)newWidth;
 	}
-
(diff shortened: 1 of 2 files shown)


wjp:
Thank you for taking a look at this.

Can you explain why your patch is the correct way of solving this problem?

norbertbajko:
I fixed the character's coordinates because that caused the crash. The game thought the character is sill moving and when the character went out the screen an array was indexed with a negative index. Because that the array get corrupted and a memcpy function crashed the game. I talked with Sev a lot about the head problem and we came to the conclusion the logical solution is hiding the unnecessary head. And finally I think upper-center is the best place for the subtitles without Hacker's head.

wjp:
How did this scene behave in the original Drascula game in DOS?

norbertbajko:
I played it to the towers and it has the bug too.

dos

wjp:
Ok, thanks.

Could you add some source code comments explaining that too?

norbertbajko:
What do you think of exactly?

wjp:
Some comments that explain what the code and checks you're adding are doing, and which bug they solve.

I realise the drascula engine is pretty poorly commented, but it's never too late to start.

norbertbajko:
I wrote some comments in the code to my changes, is it enough?

wjp:
Yes, I think so. Thanks for the rapid responses and changes.

bluegr:
I just tried your changes, and they are indeed the correct way to solve this issue. Thanks for your work! Merging

norbertbajko:
My pleasure

PR #590, opened 2015-04-07
SCUMM: Add support for ScummC games

AlbanBedel:
I know that this is a bit controversial so I'd like to make a few things clear:

* ScummC is only a compiler and doesn't require any data from LEC games
* ScummC can not be used to modify existing games
* Yes, there is only one demo and no real game to speak of

So I really doubt that adding those few lines to the games detection tables is going to run the ScummVM project in any legal trouble. As the patch is really trivial, and has a very low maintenance impact, I would really appreciate if it could be merged.

### DIFF
--- a/engines/scumm/detection_tables.h
+++ b/engines/scumm/detection_tables.h
@@ -71,6 +71,7 @@ static const PlainGameDescriptor gameDescriptions[] = {
 	{ "monkey2", "Monkey Island 2: LeChuck's Revenge" },
 	{ "pass", "Passport to Adventure" },
 	{ "samnmax", "Sam & Max Hit the Road" },
+	{ "scummc6", "ScummC games for v6" },
 	{ "tentacle", "Day of the Tentacle" },
 	{ "zak", "Zak McKracken and the Alien Mindbenders" },
 	{ "indyloom", "Indiana Jones and the Last Crusade & Loom" },
@@ -80,6 +81,7 @@ static const PlainGameDescriptor gameDescriptions[] = {
 #ifdef ENABLE_SCUMM_7_8
 	{ "ft", "Full Throttle" },
 	{ "dig", "The Dig" },
+	{ "scummc7", "ScummC games for v7" },
 	{ "comi", "The Curse of Monkey Island" },
 #endif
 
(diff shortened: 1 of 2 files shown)


clone2727:
If there's only one demo, why is this adding generic entries? And, I believe in the past it ran as the DOTT target, so why is this needed for that one game?

This was rejected last time (http://sourceforge.net/p/scummvm/patches/826/), so why do you think anything has changed to accept it now?

AlbanBedel:
Currently there is only one demo available, however creating games is the point, so generic entries are needed. I can also add an entry for the OpenQuest demo if you like.

Yes it is possible to run ScummC games as DOTT, however some game specific hacks are then enabled.

Last time it was argued that "endorsing" ScummC would constitute a legal risk for ScummVM. I doubted that back then, and nothing ever happened. Now after nearly 8 years I thought that might be worth another try. I understand that there is some fear around this subject, but I can only re-iterate that no property from LEC is involved. Moreover it isn't suited to modify existing games, so it wouldn't help in infringing on LEC IP either.

JenniBee:
There's actually two ScummC demos (OpenQuest and Road Demo), and one ScummGEN demo (Indiana Jones and the Call of Thunder) available.  All of them compile by default to match the DOTT target, so that they can be found and ran with ScummVM, but that can be turned off in the makefile.  I've gotten the ScummGEN toolchain working myself (I never could get ScummC to compile on my system), compiled the Indiana Jones demo and toyed around a bit with it to get it to display an animation I made.

This pull request is not necessary to add into ScummVM, since both toolchains are set up to match the DOTT target by default, but it would be nice to not have to set the flags to compile the game match DOTT in ScummVM for it to run out of the box (especially since, when a game is distributed it will display as Day of the Tentacle when added to ScummVM, which could cause some unnecessary confusion to those less knowledgeable in things like editing display names for entries in ScummVM).

bluegr:
Since it's just a generic detection entry for these games, without any game-specific changes, it should be fine to merge. IMHO, I don't believe there are IP issues from a generic detection entry

sev-:
No, that will not go. Adding explicit support for ScummC would mean that we do support SCUMM authoring tools which we don't.

There is nothing really good or unique in SCUMM engine, moreover it is quite limited. As before I suggest to look into more modern tools such as AGS or Wintermute.

I'm closing this.

AlbanBedel:
I still don't understand why having these file auto detected would mean any way that ScummVM endorse the tools used to generate them.

But could we at least find an agreement to allow running SCUMM games without any game specific hack enabled in the engine? New games entries seemed like the logical choice technically, but if we can get the same in another way that would be "politically acceptable", I'll be fine with it.

Generic game entries, without any mention of ScummC, would be fine too. At least that would be fair to other SCUMM compiler projects.

clone2727:
@sev-: Why is it OK that we support AGI and SCI fan games then?

PR #591, opened 2015-04-09
CREATE_PROJECT: Fix XCode-project generation.

somaen:
Having poked at this for a while, this fixes most of the details relevant to getting XCode-project generation up and going, I finally got to the point that a working XCode-project can be generated.

I modified the current code to use groups instead of folder-references, as that was what I'd been using in my own hand-made projects anyhow, and then poked at the project file until it stopped crashing on me.

Current limitations:
* HQ-scalers must be disabled (haven't resolved that issue yet)
* Tremor has been disabled (I have no idea why it was forced on earlier
* Tony fails to build as it includes "utils.h" directly, expecting to get "engines/utils.h"
* iOS-targets are completely disabled for the moment (they would require a more complex approach to dividing up the build-phase, since some OS X files should be removed and some should be added).
* No magic handling of frameworks has been added (everything is assumed to be available, and MacPorts is assumed as the source of anything not available from the SDKs)
* Currently Frameworks and Products are placed in the same root-group as everything else, I plan to rework this to create a root-level group for "Sources", and move the source-files into that, for a cleaner project-layout.

Other than that, a "plain"-project (i.e. all engines that are default-enabled) compiles, and runs fine (including tony if "engines/utils.h" is used instead of "utils.h"), with the mentioned caveat about --disable-hq-scalers.

Even with those limitations, I think that the current state is a great improvement on the former "no XCode-project-creation is possible"-situation, and the issues mentioned above can probably be resolved by manually editing the project files until they are resolved. 

The commit history isn't the cleanest, as I changed approach midway, after realizing that the engines/ stuff should be merged into everything else in a cleaner way than what it became at that point. (Thus I'm open to cleaning up any ugliness in the history if needed).

Any feedback is much appreciated.


Templier:
This is great. I never had the time to finish work on this and left it in a completely broken state.

I had started a rewrite using CMake code as a reference. This is available in my fork in the [create_project-xcode](https://github.com/Templier/scummvm/tree/create_project-xcode/devtools/create_project) branch. It was not as far along as the code you started from but it allowed for a cleaner codebase (no more macros and the horrible XCodeProvider::writeSetting() ). It might be worth having a look to see if you can salvage anything.

bluegr:
Note: The inclusion of util.h in tony has been fixed in commit b6aa297fc8

sofakng:
Does this work to create iOS Xcode project files?  It looks like it only supports OS X builds?

bluegr:
@sofakng: As mentioned in the pull request description:

    iOS-targets are completely disabled for the moment (they would require a more complex approach 
    to dividing up the build-phase, since some OS X files should be removed and some should be 
    added).

bluegr:
I've just tested this, great work!

I've added build instructions in our wiki, using MacPorts:
http://wiki.scummvm.org/index.php/Compiling_ScummVM/Mac_OS_X

It would be nice to disable "Allow debugging when using document Versions Browser" to the generated project, if possible.

Unless someone objects, I'll merge this on Sunday, 19th of July 2015

bluegr:
Also, one more thing: I was not able to create a usable project file in dists/macosx, as the root path of the generated project file seemed to be unaffected by command line. Thus, I had to copy the generated project file in ScummVM's root folder

devtools/create_project/xcode.cpp
```diff
@@ -54,39 +62,172 @@ namespace CreateProjectTool {
...
+}
+
+XCodeProvider::Group *XCodeProvider::Group::getChildGroup(const std::string &name) {
+	std::map<std::string, Group*>::iterator it = _childGroups.find(name);
+	assert(it != _childGroups.end());
+	return it->second;
+}
+
+XCodeProvider::Group *XCodeProvider::touchGroupsForPath(const std::string &path) {
+	if (_rootSourceGroup == nullptr) {
```
bluegr: I had to change the nullptr to NULL to get create_project to compile with the default options (C99). Since we don't really use C11 functionality for create_project, I don't believe there's any advantage of using nullptr here

bluegr:
I was able to enable the HQ scalers by disabling NASM. The Assembly versions of these scalers are for x86, so they won't work for x64. Thus, I've substituted --disable-hqscalers with --disable-nasm in the instructions

fingolfin:
Somewhat off-topic, and not very important, but: the name is `Xcode`, not `XCode`.

bluegr:
As promised, I'm merging this. Some small changes will be done in-tree

sofakng:
Anybody going to try working on the iOS build targets? :/

PR #592, opened 2015-04-22
SCI: fix errors in constants

sciguy000:
fixes some palette issues in SQ6


sciguy000:
fixes SCI21 but breaks SCI11, cancelling

PR #593, opened 2015-05-02
SWORD25: Delay next movie frame by videoDecoder.getTimeToNextFrame()

chkr-private:
This fixes the problem that the movies are played too fast
and get out of sync with additionally played speech samples.

This can be tested by watching the intro movies: In the plane, George should start talking about Nico once he is shuffling his papers on the tray. Without the bug fix, the movie plays on my system with twice the speed and he starts talking much later. Additionally, the audio track of the movie gets out of sync with the video, too.

### DIFF
--- a/engines/sword25/fmv/movieplayer.cpp
+++ b/engines/sword25/fmv/movieplayer.cpp
@@ -123,7 +123,7 @@ void MoviePlayer::update() {
 		if (_decoder.endOfVideo()) {
 			// Movie complete, so unload the movie
 			unloadMovie();
-		} else {
+		} else if (_decoder.needsUpdate()) {
 			const Graphics::Surface *s = _decoder.decodeNextFrame();
 			if (s) {
 				// Transfer the next frame


digitall:
This looks like a fairly simple and self-contained change. @sev- , @clone2727 : Happy to merge?

clone2727:
@digitall I have nothing to do with sword25, nor can I test.

digitall:
@clone2727: *shrug* OK. I asked you for an opinion as primary author of the VideoDecoder code and since Broken Sword 2.5 is freeware from http://brokensword25.com/, it should be possible for you to take a look and test if you have time, but up to you.

wjp:
At first glance, the movie play loop goes via scripts. Don't they do any timing?

wjp:
Ah, no, the scripts don't do any timing, but since they do still do quite a bit of other things, I don't think we can let the movieplayer sleep the full time until the next frame directly after rendering.

chkr-private:
@wjp: Next proposal: only draw the next frame when _decoder.needsUpdate() returns true. Works fine for me, too. Please see the updated patch.

bluegr:
That last change looks to be the best and least invasive solution, as the decoder handles the frame rate on its own

bluegr:
This is the correct way to play video files, according to the video decoder's usage.

Here's an example from the ZVision engine:
https://github.com/scummvm/scummvm/blob/master/engines/zvision/video/video.cpp#L116

I'm merging this, since it does fix the videos. As the sword25 engine is still not stable, we can worry about any possible regressions at a later stage.

PR #594, opened 2015-05-04
AGI: Use correct volume for PCjr output

criezy:
This looks like a regression from commit 24bb5da: AGI: Add a layer of abstraction between the sound chip and the two players

I am quite unfamiliar with this code, so I prefer to get some feedback before pushing to master.

### DIFF
--- a/engines/agi/sound_pcjr.cpp
+++ b/engines/agi/sound_pcjr.cpp
@@ -207,7 +207,7 @@ int SoundGenPCJr::volumeCalc(SndGenChan *chan) {
 				chan->attenuationCopy = attenuation;
 
 				attenuation &= 0x0F;
-				attenuation += _mixer->getVolumeForSoundType(Audio::Mixer::kSFXSoundType) / 17;
+				attenuation += (16 - _mixer->getVolumeForSoundType(Audio::Mixer::kSFXSoundType) / 17);
 				if (attenuation > 0x0F)
 					attenuation = 0x0F;
 			}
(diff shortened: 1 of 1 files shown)


criezy:
And just on cue I see a bug report was created a few minutes ago.
Bug #6858 AGI: KQ1 PCjr sound lacks volume envelopes

criezy:
Note on the second commit to fix an issue with the volume: I am actually not convince that applying the volume to the attenuation value is what we want to do. We have a hardcoded table of 16 amplitudes values for the signal and what selecting a low music volume does for example is increasing the attenuation so that only the few lowest amplitudes from the table are used. So the music sounds clipped because we only use a few different values.

I am wondering if instead we should not use the volume to scale the amplitudes in the table and thus always use 16 different amplitudes. This could be done by changing fillSquare() and fillNoise() so that instead of using directly volTable[t->atten] it would use
int16 amplitude = volTable[t->atten] * _mixer->getVolumeForSoundType(Audio::Mixer::kSFXSoundType) / 256;
(and obviously we would also need to remove the effect of the volume on the attenuation in volumeCalc)

Any thoughts on that?

criezy:
While I am at it, here are two bits of code from this same file that look suspicious to me.

The first one starts at line 116.
```cpp
    if (_vm->getVersion() >= 0x3000)
        _dissolveMethod = 3;
    else if (_vm->getVersion() >= 0x2900)
        _dissolveMethod = 2;
    else
        _dissolveMethod = 0;
    
    _dissolveMethod = 3;
```

The last line was added by sev in commit ceb2909 (AGI: Split all sound generators into separate modules.). Was that maybe some debug code he forgot to remove before commit?

The second thing that looks a bit strange:
Line 128: `_mixer->playStream(Audio::Mixer::kMusicSoundType, &_soundHandle, this, -1, Audio::Mixer::kMaxChannelVolume, 0, DisposeAfterUse::NO, true);`
Line 210: `attenuation += (16 - _mixer->getVolumeForSoundType(Audio::Mixer::kSFXSoundType) / 17);`

Shouldn't we use kMusicSoundType (or kSFXSoundType) both times?

Depending on the answers I can make more changes, squash all this and push to master.

criezy:
After looking at this code once more today I have pushed a slightly modified version to master.

PR #597, opened 2015-05-16
Add support for Visual Studio 2015 to create_project

Templier:
This adds support for the new version of Visual Studio. 

Note that an updated set of libraries is required for proper linking and they are not backward compatible with VS 2013. You can grab pre-built libraries from the [scummvm-libs](https://github.com/Templier/scummvm-libs) repository. I can transfer that repository to the organization if desired or let the maintainers update the scummvm_vs_libs.zip package instead.

Tested:
 - Visual Studio 2015 RC in debug and release (win32 & x64)
 - Visual Studio 2013 (using the current library pack)

Tasks to complete before this can be merged:
 - [ ] Do not ignore new warnings (C4267 and C4458)
 - [ ] Recreate the scummvm-libs repository to compile for both VS2013 and VS2015
 - [ ] Update create_project to read 2015 and pre-2015 libraries from different folders
 - [ ] Write script to automate building all libraries
 - [ ] Document process to add/update a library in the repository (using git subtree)
 - [ ] Update wiki with new process to get pre-built binaries for Visual Studio
 - [ ] Transfer ownership of repository to ScummVM organization
 - [ ] Update script to allow building MinGW libraries too (optional)

### DIFF
--- a/common/scummsys.h
+++ b/common/scummsys.h
@@ -46,7 +46,7 @@
 
 	#if defined(WIN32)
 
-		#ifdef _MSC_VER
+		#if defined(_MSC_VER) && _MSC_VER <= 1800
 
 		// FIXME: The placement of the workaround functions for MSVC below
 		// require us to include stdio.h and stdarg.h for MSVC here. This
--- a/devtools/README
+++ b/devtools/README
@@ -72,8 +72,8 @@ create_mort (Strangerke)
 
 create_project (LordHoto, Littleboy)
 --------------
-    Creates project files for Visual Studio 2008, 2010, 2012, 2013, Xcode and
-    Code::Blocks out of the configure / Makefile based build system.
+    Creates project files for Visual Studio 2008, 2010, 2012, 2013, 2015,
+    Xcode and Code::Blocks out of the configure / Makefile based build system.
     It also offers a way to enable or disable certain engines and the use
     of external libraries similar to configure. Run the tool without
     any arguments for further help.
--- a/devtools/create_project/create_project.cpp
+++ b/devtools/create_project/create_project.cpp
@@ -124,7 +124,7 @@ int main(int argc, char *argv[]) {
 	setup.features = getAllFeatures();
 
 	ProjectType projectType = kProjectNone;
-	int msvcVersion = 9;
+	int msvcVersion = 12;
 	bool useSDL2 = false;
 
 	// Parse command line arguments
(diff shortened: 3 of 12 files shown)


dreammaster:
Speak of the devil. Just the other day I provided Strangerke with an updated scummvm_libs that contained the extra libraries we now use (x86 only, not x64) for VS 2013 and earlier, as well as the necessary DLLs. Although he hadn't had a chance to update it yet.

I'll download a copy of your lib and integrate it into my update package, and then pass it on for updating on Sourceforge.

dreammaster:
See my [updated scummvm_lib](https://drive.google.com/file/d/0B7K7KNNR87NPV1BYM3ZTRkd0ekk/view?usp=sharing) for the merging of the two versions. I've cut out the src/ folder to keep the size down, and made a separate lib_2015 for your VS 2015 versions of the libraries, to keep them separate from the existing ones.

dreammaster:
Well, compilation doesn't break for me on my existing MSVC2013, but I don't have MSVC2015 installed to actually test the create_project changes. Overall, the changes seem reasonable. Although I'm a wary of the new warning suppressions that this adds: 4267 - possible loss of data in type conversion, and 4458 - variable declaration hides class member.

The latter is something we would want to keep showing as a warning, since it's something that should be fixed across the code base. As for the former, I wasn't aware that the current code-base was generating a "ton" of warnings. Certainly I don't get many warnings when I do a clean build in VS 2013. As with the other warning, it may be better not to set it, and see if anyone is interested in fixing the all the code that generates such errors.

dreammaster:
Also, it may be just co-incidence, but since I tried compiling your revised create_project, I started getting link errors about a 'win_utf8_io' library compiling ScummVM. It's kind of hard to find information on it.. am I correct in thinking that that's a secondary lib file produced by libFlac? I tried using the lib file from your folder, and it seems to work fine in my VS2013.

Templier:
Warnings: You are right, these should not be ignored by default.

win_utf8_io.lib: This is indeed a new dependency introduced by libflac. I thought it would skip it when trying to link in VS2013 but that doesn't seem to be the case.

It looks like I need to revise some of the work before it can be merged. I think the proper way is to redo the scummvm-libs repository to also compile the VS2013 libraries rather than modifying the projects to compile for 2015 instead. This will ensure both sets of libs are in sync. I have edited the pull request description with the TODO list.

dreammaster:
Deprecated in favor of new pull request

PR #600, opened 2015-07-01
Add support for OPL output through ALSA

clone2727:
As it says on the tin, this allows for output to a real OPL device through ALSA.

This required refactoring out all users of the OPL code to not rely on the OPL class being an AudioStream. This wasn't particularly tricky, except for gob which changed timer callback frequency while running.

I do not have every game available for testing, but I did a bunch:

Tested:
- Player_AD
- MidiDriver_ADLIB
- SCI
- Cine
- Sky
- Queen
- Gob (both drivers)

Not Tested:
- Cruise
- Kyra
- Parallaction
- TsAGE
- MADS
- Sherlock
- AGOS
- MidiDriver_Miles_AdLib

They need testing regardless of real vs. emulated OPL because of the aforementioned rewrites.

Much thanks to @waltervn, who wrote all the ALSA-based code and adapting the engines that needed volume control handled without using the mixer. And also thanks to @lordhoto who wrote some fixes to the OPL selection code.

In the future, hopefully we'll also be able to support real OPL devices on Windows as well, but that's out-of-scope for this PR.


DrMcCoy:
For the record: the gob changes look good to me.

dreammaster:
Cruise, MADS, and TsAGE look okay to me. Sherlock seems to be okay, although it shortly crashes in the MIDIParser (in master as well as this pull request, so it's unrelated to these changes).

lordhoto:
I gave Kyra1, Kyra2, EoB I+II, and LoL a short try and it seems it still works. Changes to Kyra look fine too.

wjp:
Works for me with hardware OPL (ymf724 via alsa), and looks good. Very nice!

One small nitpick is that alsa.cpp probably doesn't belong in the audio/softsynth/opl dir. Maybe audio/alsa_opl.cpp or alsa_fm.cpp instead? Theoretically it should probably be somewhere in backends/, but we can always move it there later if we get other implementations of hardware opl, maybe.

clone2727:
Thanks for testing!

@wjp: I agree on the file location; I can move it tonight. Although, I'd rather just have this under backends/ at this point rather than audio/. Moving that out also leaves adlib.cpp and eas.[cpp|h] also erroneously being under softsynth, but those can just be moved too (the former should be done in this PR, but not the latter).

clone2727:
@wjp Ended up just moving it to audio/. Having it in backends/ introduced a cyclical dependency, that isn't worth working around right now.

If there's no complaints, I'll merge tomorrow or the day after.

wjp:
Thanks

clone2727:
Thanks to those who tested!

PR #601, opened 2015-07-06
Z-Vision polish

elQuotho:
Mostly minor style and documentation corrections, with the exception of one commit which moves all of what I saw as static data declarations for the detection engine into a `detection_tables.h` file, which seems to be the standard destination once the detection data becomes unwieldy.

Feel free to pick and choose which to take aboard, each commit should be self-contained. This is also my first time working with C++, so apologies for any rookie mistakes. I've tested Win32 compilation and Z-Vision game detection for ZGI and Nemesis on Windows 8.1, 64-bit.


RichieSams:
The changes look ok by me. Nice job on the detection move!

bluegr:
At a first glance, the changes look OK. Will check this out tomorrow, when I get back home

elQuotho:
Thanks guys, I'll fix that up shortly.

elQuotho:
I've made the changes in `credits.pl` as requested, generated new `AUTHOR` and `credits.h`, and replaced a couple instances of ZVision I had missed by foolishly relying on Visual Studio's project-file search instead of searching all source files.

I've also squashed the fixes into the previous commit, so they're now part of commit fd8ac7e.

bluegr:
Great work, thanks! Merging

PR #604, opened 2015-07-22
CONFIGURE: add and use append_var function

fingolfin:
Not sure if there is any interest in this ("never change a running system"), but here it is anyway:

This PR adds a new function `append_var` to the `configure` script, which allows replacing constructs like
```
LDFLAGS="$LDFLAGS -static-libgcc -static-libstdc++"
```
by
```
append_var LDFLAGS "-static-libgcc -static-libstdc++"
```
or even
```
append_var LDFLAGS -static-libgcc -static-libstdc++
```
The goal is to improve readability and reduce "code duplication", and thus to reduce the risk of introducing certain bugs, like these two made-up ones:
```
LDFLAGS="$LDFLAG -static-libgcc -static-libstdc++"
CXXFLAGS="$CPPFLAGS -diag-error 10006,10148"
```

I also made sure constructs like this keep working correctly (i.e. the variable `srcdir` is not expanded prematurely):
```
append_var INCLUDES '-I$(srcdir)/backends/platform/dc'
```
Most of the changes were made with two simple regex plus search&replace (for the "obviously correct" conversions), the remaining ones were found using a third (less restrictive) regex and manually modified. I also removed three instances of
```
ASFLAGS="$ASFLAGS"
```

Finally, I originally called the function `append`, but since there are already `set_var` and `get_var`, I thought I should match them. I'd be happy to rename it to anything else you might prefer.

### DIFF
--- a/configure
+++ b/configure
@@ -69,6 +69,16 @@ get_var() {
 	eval echo \$${1}
 }
 
+append_var() {
+    VAR=${1}
+    shift
+    if eval test -z \"\$${VAR}\" ; then
+        eval ${VAR}='$@'
+    else
+        eval ${VAR}=\"\$${VAR} \"'$@'
+    fi
+}
+
 # Add an engine: id name build subengines base-games dependencies
 add_engine() {
 	_engines="${_engines} ${1}"
@@ -1172,7 +1182,7 @@ for ac_option in $@; do
 		_debug_build=no
 		;;
 	--enable-Werror)
-		CXXFLAGS="$CXXFLAGS -Werror"
+		append_var CXXFLAGS "-Werror"
 		;;
 	--enable-release-mode)
 		_release_build=yes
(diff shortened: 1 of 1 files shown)


bluegr:
Nice work! +1 from me

digitall:
This change looks very reasonable and reduces chances of future errors. +1 from me as well. Thanks @fingolfin for this.

Any objections from Core Team to merging this as-is? @sev- ? @lordhoto ? @wjp ?

wjp:
Haven't tested, but the concept looks good to me.

lordhoto:
I didn't test it either, but I think it makes sense to add this.

fingolfin:
So, what exactly needs to happen for this to be merged?

wjp:
Me being back home and testing it, mainly.

I've now tested this on a few different shells, and it seems to work fine. Thanks!

PR #605, opened 2015-07-25
SHERLOCK: added yet another german version for scalpel

mgerhardy:


bluegr:
Where did you get this version from?

mgerhardy:
If only I could answer this. I found it on one of my old disk-to-cdrom collections. But I throw away all my boxes (I know.... silly from todays standpoint)

dreammaster:
Thanks as well. What kind of CD collection is it? Is this an official game compilation disc? If so, the publisher details should be on it, I imagine.

mgerhardy:
No, it's a private one where I "rescued" my box floopy discs. All I know is that it must have been a box version on 3,5" floppy discs. Otherwise it should not have been on that cdrom.

PR #606, opened 2015-07-26
I18N: Czech localization update

Tsbook:
Update Czech documentation and localization up to the latest changes.

Also fix typos that cropped up during editing.


criezy:
Thank you. I have now merged this commit after editing it to revert the change to the Language field of the po file header, which I suspect was not done on purpose. This is used in the GUI so it is better to keep a user friendly name in that field (but poedit always changes it to the country and language code so this is a common issue with most pull request).

PR #607, opened 2015-07-27
SHERLOCK: some small cleanups

mgerhardy:
I would like to get into the codebase of scummvm. Therefore I thought it might be useful to start with some small cleanups. Here they are. Is stuff like this wanted?

### DIFF
--- a/engines/sherlock/inventory.cpp
+++ b/engines/sherlock/inventory.cpp
@@ -94,8 +94,8 @@ void Inventory::loadGraphics() {
 	for (int idx = _invIndex; (idx < _holdings) && (idx - _invIndex) < (int)_invShapes.size(); ++idx) {
 		// Get the name of the item to be displayed, figure out its accompanying
 		// .VGS file with its picture, and then load it
-		int invNum = findInv((*this)[idx]._name);
-		Common::String filename = Common::String::format("item%02d.vgs", invNum + 1);
+		const int invNum = findInv((*this)[idx]._name);
+		const Common::String &filename = Common::String::format("item%02d.vgs", invNum + 1);
 
 		if (!IS_3DO) {
 			// PC
@@ -239,7 +239,6 @@ void Inventory::synchronize(Serializer &s) {
 
 	for (uint idx = 0; idx < size(); ++idx) {
 		(*this)[idx].synchronize(s);
-
 	}
 }
 
--- a/engines/sherlock/inventory.cpp
+++ b/engines/sherlock/inventory.cpp
@@ -142,30 +142,29 @@ int Inventory::putItemInInventory(Object &obj) {
 		_vm->setFlags(obj._pickupFlag);
 
 	for (int useNum = 0; useNum < USE_COUNT; ++useNum) {
-		if (obj._use[useNum]._target.equalsIgnoreCase("*PICKUP*")) {
-			pickupFound = true;
+		if (!obj._use[useNum]._target.equalsIgnoreCase("*PICKUP*"))
+			continue;
+		pickupFound = true;
 
-			for (int namesNum = 0; namesNum < NAMES_COUNT; ++namesNum) {
-				for (uint bgNum = 0; bgNum < scene._bgShapes.size(); ++bgNum) {
-					Object &bgObj = scene._bgShapes[bgNum];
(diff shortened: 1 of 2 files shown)


mgerhardy:
Yes, it's part of the c++ standard - see http://stackoverflow.com/questions/12953127/what-are-copy-elision-and-return-value-optimization

mgerhardy:
cyclic complexity is something that is a matter of style. But RVO is a real optimization. But as I said, this is just an example for some kind of improvements I would like to start on to get in touch with the source code.
If the cyclic complexity reduction is something that is against your style, I can remove the commit.

clone2727:
That's not copy elision. Copy elision is the compiler putting the result of constructing an object into another without copying, not taking reference to one.

lordhoto:
Actually, this reply from the thread you linked explains that RVO applies to the old code before you changed it: http://stackoverflow.com/a/12953150 So, I am a bit confused right now, why you talk about RVO in your commit message, even though RVO by the compiler applied to the code?

mgerhardy:
You are of course all right - but:
    const Type &ref = foo.bar();
will also work if your compiler doesn't support RVO properly - you can see this when you use gcc with -fno-elide-constructors. Not sure if this is something that is useful or needed in the context of scummvm. Not sure what your target compilers are.

    #include <stdio.h>
    
    class Foo {
    private:
        const char *_str;
    public:
        Foo(const char* str) : _str(str) {
            printf("ctor: %s\n", _str);
        }
        Foo(const Foo& other) : _str(other._str) {
            printf("copy: %s\n", _str);
        }
        ~Foo() {
            printf("dtor: %s\n", _str);
        }
    
        const char *Str() const { return _str; }
    };
    
    class Bar {
    private:
        const char *_str;
    public:
        Bar(const char* str) : _str(str) {}
        Foo getFoo() { return Foo(_str); }
    };
    
    int main (int argc, char *argv[]) {
        {
            Bar b("test1");
            Foo foo = b.getFoo();
            (void)foo;
        }
        {
            Bar b2("test2");
            const Foo& foo2 = b2.getFoo();
            (void)foo2;
        }
        printf("after\n");
        return 0;
    }

compile with
    g++ -fno-elide-constructors foo.cpp

(hopefully I did not do anything silly)

but still, it's also most likely just a style issue. Because I would suspect all of the compilers supporting RVO properly.

bluegr:
But that's not what's happening here. Here, you have:

    class Foo {
    ....
    **static** String bar(const char *baz) { ... }
    }

and you're doing:

    String &foobad = Foo::bar("...");

There isn't any new instance of a class here... this is a **static** class member which you're referring to

mgerhardy:
In your case you missed the const - this won't even compile (with sane compilers). It also doesn't matter whether it's static or not. See this example with the same g++ parameter as above:

#include <stdio.h>
    
    class Foo {
    private:
        const char *_str;
    public:
        Foo(const char* str) : _str(str) {
            printf("ctor: %s\n", _str);
        }
        Foo(const Foo& other) : _str(other._str) {
            printf("copy: %s\n", _str);
        }
        ~Foo() {
            printf("dtor: %s\n", _str);
        }
    
        const char *Str() const { return _str; }
    };
    
    static Foo getFoo(const char *str) { return Foo(str); }
    
    int main (int argc, char *argv[]) {
        {
            Foo foo = getFoo("test1");
            (void)foo;
        }
        {
            const Foo& foo2 = getFoo("test2");
            (void)foo2;
        }
        printf("after\n");
        return 0;
    }

mgerhardy:
but anyway - maybe we should just close this. This was also more or less meant as a warmup pull request while digging the code.

lordhoto:
If we assume that the compiler does not handle RVO correctly, we would need to at least adapt all performance critical paths to keep a reference to the returned temporary object around. It seems kind of arbitrary to do it in these places.(?)

At any rate, I would suggest not to mix up changes like you did in the first commit, i.e. adding the const refrences and just generally making variables const (the latter is a style case in most cases you added I think, so not sure whether it's worth the bikesheding...).

dreammaster:
Even though this didn't work out, thanks for making the effort anyway. Hopefully we can look forward to more contributions from you in the future.

PR #608, opened 2015-07-27
BUILD: Added Travis-CI support

mgerhardy:


mgerhardy:
See https://travis-ci.org/mgerhardy/scummvm for a test.

Someone of scummvm has to login to travis-ci.org and set the flag on the main scummvm repo. After that is done, travis-ci will start a build for every commit.

mgerhardy:
If there is interest in these kind of integration, I could also add https://ci.appveyor.com support.

janisozaur:
why no `make test`?

mgerhardy:
I can add whatever is needed - that's how I currently build (and I'm just starting with the code) - if there is something wrong, please let me know, than I will extend it. Will check out the test target - thanks

mgerhardy:
added test and devtools target

bluegr:
Nice work, thanks!

There is functionality to include automated Coverity scans in Travis-CI too, but that could be added afterwards. A project admin will need to have a look at this

fuzzie:
This seems a bit premature since there are no successful builds for the test setup yet :) But builds of pull requests and builds on OS X might be interesting indeed.

mgerhardy:
It failed after I changed the branch to master - I have my local tests done in travis and had the branch sticked to that one. But it worked fine for me (not sure about devtools and test

djwillis:
It strikes me that getting this in is a good idea, it sits nicely alongside the buildbot builds and also gives us a build solution for getting good basic coverage on at least a 'stock' build for pull requests. 

All told,it should be a really worthwhile addition and is little effort to put in.

mgerhardy:
While we are on it - another nice addon is https://linthub.io/

DrMcCoy:
Not your linthub is such a good idea. [For C++, it seems to use cpplint](http://docs.linthub.io/user/configuration/), which is [Google's checker](https://google-styleguide.googlecode.com/svn/trunk/cpplint/cpplint.py), operating from their style and code guidelines. They don't really align with ScummVM's guidelines.

clone2727:
I really hate having a file in the tree that's not related to the actual application but rather a configuration for another. But I'm told there's no other way...

The README.md doesn't belong here. The text inside has nothing to do with being a README and it would conflict with our normal README.

djwillis:
Just to follow up on @clone2727 's comment.

I actually agree that there is no need for the README.md markdown code in this PR. I missed that in my earlier comment.

The .travis.yml would be entirely sufficient and if we wanted to show the Travis CI build status we could choose to do that inside the website or the build subsite with some suitable HTML. I don't believe it is a requirement of the service to show the build status in your README or anywhere else for that matter.

@mgerhardy would you consider dropping the markdown part of the pull?

mgerhardy:
Done. Removed the README.md. Just for reference:

    [Build Status](https://travis-ci.org/scummvm/scummvm)

mgerhardy:
Anything that is missing from my side to accept this PR?

fingolfin:
It seems like this would be nice to have -- and perhaps it could even be set to use `--enable-Werror` at some point? Then perhaps, assuming people regularly check the output, also things like this could be found automatically and quickly (in particular, even without me coming back every coupld months with fixes for warnings given by clang ;-):
```
../../engines/sherlock/tattoo/tattoo_user_interface.cpp:221:34: error: array index 768 is past the end of the array (which contains 256 elements) [-Werror,-Warray-bounds]
        Common::copy(&_lookupTable[0], &_lookupTable[PALETTE_SIZE], &lookupTable[0]);
```

sev-:
Looks good to me. We probably need to install Travis-CI to our server machine.

bluegr:
We don't need to install Travis-CI on our server. Travis-CI runs on its own infrastructure, and checks pre-determined github repositories. All we need to do is tell it to check our repository, and place a .travis.yml file (which is done by this pull request). The rest is done automagically.

Check here: https://travis-ci.org/getting_started

fingolfin:
Any reason this is still not merged, after several months?

lordhoto:
> Any reason this is still not merged, after several months?

I guess nobody with access to scummvm/scummvm set up Travis-CI yet. @wjp: Any chance you can look into this?

.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+  - gcc
+  - clang
+
+os:
+  - linux
+
+script:
+  - ./configure --enable-all-engines
+  - make
+  - make test
```
lordhoto: Is this a good idea? With the given configure set up our tests will always fail (to link due to EventRecorder cross dependencies). Maybe we should pass `--disable-eventrecorder` to `configure` in this setup.

janisozaur:
Any news on this?

sev-:
OK, merging, adding --disable-eventrecorder and looking into the Travis CI setup

PR #611, opened 2015-08-26
SHERLOCK : fix compilation in Symbian

fedor4ever:
In file included from D:\Symbian\Projects\SDL\scummvm\engines\sherlock\scalpel\scalpel.cpp:29:
D:\Symbian\Projects\SDL\scummvm\engines\sherlock/scalpel/tsage/logo.h:191: macro `remove' used without args
ERROR: cpp.EXE failure
make: *** [MAKEFILESCUMMVM_SHERLOCK] Error 33

### DIFF
--- a/engines/sherlock/scalpel/tsage/logo.h
+++ b/engines/sherlock/scalpel/tsage/logo.h
@@ -34,6 +34,10 @@
 #include "sherlock/scalpel/tsage/resources.h"
 #include "sherlock/screen.h"
 
+#ifdef __SYMBIAN32__
+#undef remove
+#endif //__SYMBIAN32__
+
 namespace Sherlock {
 namespace Scalpel {
 


wjp:
It might be nicer to undefine this closer to where it's getting defined in the first place. Where is the relevant system header getting included?

fedor4ever:
Relevant system header is libc\sys\unistd.h
It's definition is :
 #define remove(x)       unlink(x)

SHERLOCK engine ancestor - TSAGE - has another solution:
#ifndef REMOVE_UNDEFINED
#define remove(x)       unlink(x)
#define wremove(x)       wunlink(x)
#endif //REMOVE_UNDEFINED
And then #define REMOVE_UNDEFINED to 1 per engine if required.
Its not possible completely disable because scummvm itself use remove(x) for removing files.

I try for problem engines to undef via gcc option -Uremove, via portdefs.h - no way

wjp:
Sure, the system's remove() is used, but only in exceptional spots (maybe even only one? default-saves.cpp), and we can add something similar to FORBIDDEN_SYMBOL_EXCEPTION_* for usage like that.

wjp:
To elaborate: we explicitly want to allow the name remove to be used for methods, while we _do not_ want to allow arbitrary code to use the system's remove. (See common/forbidden.h)

wjp:
And, wait, do you mean you _change_ the system's unistd.h file?!

fedor4ever:
Yes, I do and it is very bad. *__#ifndef REMOVE_UNDEFINED__* hide this macro from some engines only. Infrastructure code has access to this macro.

fedor4ever:
I add code to common/forbidden.h :
'''c++
#ifndef FORBIDDEN_SYMBOL_EXCEPTION_unistd_h
	#ifndef FORBIDDEN_SYMBOL_EXCEPTION_remove
	#undef remove
	#define remove(a)	FORBIDDEN_SYMBOL_REPLACEMENT
	#endif
#endif

and this to symbian portdefs.h:
#define FORBIDDEN_SYMBOL_EXCEPTION_remove

__This doesn't work.__

wjp:
Most likely because the way unistd.h is included doesn't go via forbidden.h? Or does something else go wrong?

wjp:
Oh, and your patch to forbidden.h isn't right. It shouldn't #define remove(a) to FORBIDDEN_SYMBOL_REPLACEMENT, because engines _should_ be allowed to use it.

fedor4ever:
> It shouldn't #define remove(a) to FORBIDDEN_SYMBOL_REPLACEMENT

Any ideas?

wjp:
What I'd suggest:

In portdefs.h, after including unistd.h, add

    #ifndef USE_SYSTEM_RENAME
    #undef rename
    #endif 

Then, in all places that use the system rename (which is maybe only default-saves.cpp), add a #define USE_SYSTEM_RENAME before including scummsys.h. (With a comment why this is necessary.)

If that works, we should think about the name USE_SYSTEM_RENAME and if we want to change that.

fedor4ever:
I have strange linker error. I try this later on some old and truly buildable version...

fedor4ever:
It works now. uff, 1.5 year bug near end...

PR #613, opened 2015-09-03
SCUMM: Do not reverse stereo for digital iMuse sounds

clone2727:
After comparing with the original Windows executable for The Curse of Monkey Island, I found that we're reversing the digital iMuse sounds when we shouldn't be.

@aquadran added it in 98d9069, but as far as I can tell, it shouldn't actually be here. @fingolfin added comments in fe8719d and 49af58a questioning it, but were both removed without comment.

I have recorded samples proving it too, but no place to upload them with decent bandwidth. Let me know if you want a copy.

### DIFF
--- a/engines/scumm/imuse_digi/dimuse.cpp
+++ b/engines/scumm/imuse_digi/dimuse.cpp
@@ -201,8 +201,7 @@ void IMuseDigital::saveOrLoad(Serializer *ser) {
 
 			track->stream = Audio::makeQueuingAudioStream(freq, (track->mixerFlags & kFlagStereo) != 0);
 
-			_mixer->playStream(track->getType(), &track->mixChanHandle, track->stream, -1, track->getVol(), track->getPan(),
-							DisposeAfterUse::YES, false, (track->mixerFlags & kFlagStereo) != 0);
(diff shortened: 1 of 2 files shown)


aquadran:
When I work on digital iMuse I'm sure it was needed. Maybe at that time sound was reversed after iMuse output in Mixer code. Note, you compare The Curse of Monkey Island, but the code was for The Dig at that time, maybe check this game. I think original game exe had reverse stereo audio option too.
I wonder how it's for Grim Fandango, I'm sure I would notice at the time of porting to ResidualVM.
Unfortunately I don't remember details, it was a long time ago :)

clone2727:
No, we're definitely wrong for The Dig too. Here are the waveform comparisons:
dig_dosbox_space
dig_scummvm_space

clone2727:
The original DOS digital iMuse implementation did have a reverse stereo option, but I had it to 0/off.

aquadran:
I guess you are right, but I'll also look into soon.

clone2727:
In the meantime, I'll merge it this weekend since it matches the original's output now.

aquadran:
ok

clone2727:
Committed to master as 1d45e08.

PR #616, opened 2015-09-17
CINE: Show splash screen in CD version of Future Wars

eriktorbjorn:
This simply shows the Sony splash screen included with the CD version of Future Wars. I think it does all the necessary sanity checking and cleanups, though some might find it a bit hackish.

### DIFF
--- a/engines/cine/cine.cpp
+++ b/engines/cine/cine.cpp
@@ -22,10 +22,14 @@
 
 #include "common/config-manager.h"
 #include "common/debug-channels.h"
+#include "common/events.h"
 
 #include "engines/util.h"
 
 #include "graphics/cursorman.h"
+#include "graphics/palette.h"
+
+#include "image/iff.h"
 
 #include "cine/cine.h"
 #include "cine/bg_list.h"
@@ -89,6 +93,10 @@ void CineEngine::syncSoundSettings() {
 }
 
 Common::Error CineEngine::run() {
+	if (g_cine->getGameType() == GType_FW && (g_cine->getFeatures() & GF_CD)) {
+		showSplashScreen();
+	}
+
 	// Initialize backend
 	initGraphics(320, 200, false);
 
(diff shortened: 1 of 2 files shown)


digitall:
Code looks good to me...

bluegr:
So the resolution of the splash screen itself is 640x480, whereas the game itself runs at 320x200? Odd.

Other than that, the changes look fine to me

eriktorbjorn:
Yeah, I don't know much about the image format but *file* identifies it as "IFF data, ILBM interleaved image, 640 x 480".

bluegr:
Just tested this and it works great. As usual, thank you for your work!

Merging

Kirben:
The original game switched to 640 x 480 resolution (16 colors) for the splash screen, so the resolution is definitely correct.

Unfortunately this patch breaks scaling though, the game isn't set to the current scaler after the splash screen.

eriktorbjorn:
Odd, it seemed to work for me. At least no worse than _Lost in Time_ and _DreamWeb_, which also switch between resolutions during their intros. Could you elaborate?

Kirben:
When the default scaling (2x) is been used, Future Wars (CD) reverts back to smaller 320 x 200 size, after the splash screen is shown.

eriktorbjorn:
That doesn't happen for me. There are the settings from my .scummvmrc's [scummvm] section that I assume are relevant:

gfx_mode=2x
fullscreen=false
aspect_ratio=false

There are no graphics-related settings in the [fw-cd-us] section.

For me, when the game starts the splash screen is shown unscaled. After it's gone, the game runs with the 2x scaler.

Kirben:
Try again, with no gfx_mode setting in the config file.

eriktorbjorn:
Ah. But that will happen with both _Lost in Time_ and _DreamWeb_ as well, so I don't really have any suggestions for how to deal with that. (I don't remember off-hand if there are any other games that ScummVM supports that changes resolution like that.)

PR #617, opened 2015-09-28
CREATE_PROJECT: Add support for Visual Studio 2015

dreammaster:
This is a modified version of Templier's previous pull request for adding support for Visual Studio 2015. It only introduces a single new warning 4267 to deal with the numerous spurious warnings of conversion of 'size_t' to 'type'. It also fixes extra warnings that were being generated due to exception handling.

I've also taken the liberty of recompiling all the required libraries, since the ones Templier provided were causing link warnings due to differences in the runtime libraries. Compiled libraries can be  located here:
Visual Studio 2015 libraries and includes:
http://sourceforge.net/projects/scummvm/files/build/scummvm_libs_2015.zip/download

Similar libraries for Visual Studio 2013 and earlier:
http://sourceforge.net/projects/scummvm/files/build/scummvm_libs_2013.zip/download

### DIFF
--- a/devtools/create_project/msbuild.cpp
+++ b/devtools/create_project/msbuild.cpp
@@ -55,7 +55,14 @@ int MSBuildProvider::getVisualStudioVersion() {
 	if (_version == 12)
 		return 2013;
 
+	if (_version == 14)
+		return 14;
+
 	error("Unsupported version passed to getVisualStudioVersion");
+}
+
+int MSBuildProvider::getSolutionVersion() {
+	return (_version < 14) ? _version + 1 : _version;
 }
 
 namespace {
@@ -116,7 +123,7 @@ void MSBuildProvider::createProjectFile(const std::string &name, const std::stri
 	// Shared configuration
 	project << "\t<Import Project=\"$(VCTargetsPath)\\Microsoft.Cpp.Default.props\" />\n";
 
-    std::string version = "v" + toString(_version) + "0";
+	std::string version = "v" + toString(_version) + "0";
 	std::string llvm = "LLVM-vs" + toString(getVisualStudioVersion());
 
 	outputConfigurationType(setup, project, name, "Release|Win32", version);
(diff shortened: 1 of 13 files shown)


dreammaster:
Added in files missing from my original commit.

devtools/create_project/msbuild.cpp
```diff
@@ -362,7 +383,7 @@ void MSBuildProvider::outputGlobalPropFile(const BuildSetup &setup, std::ofstrea
 	              "\t\t\t<DisableSpecificWarnings>" << warnings << ";%(DisableSpecificWarnings)</DisableSpecificWarnings>\n"
 	              "\t\t\t<AdditionalIncludeDirectories>$(" << LIBS_DEFINE << ")\\include;.;" << prefix << ";" << prefix << "\\engines;" << (setup.tests ? prefix + "\\test\\cxxtest;" : "") << "$(TargetDir);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>\n"
 	              "\t\t\t<PreprocessorDefinitions>" << definesList << "%(PreprocessorDefinitions)</PreprocessorDefinitions>\n"
-	              "\t\t\t<ExceptionHandling>" << ((setup.devTools || setup.tests) ? "Sync" : "") << "</ExceptionHandling>\n";
+	              "\t\t\t<ExceptionHandling>" << ((setup.devTools || setup.tests || _version == 14) ? "Sync" : "") << "</ExceptionHandling>\n";
```
clone2727: Exceptions should not be enabled. If they're to avoid a warning, the warning should be disabled instead.

devtools/create_project/msbuild.cpp
```diff
@@ -305,6 +319,13 @@ void MSBuildProvider::outputProjectSettings(std::ofstream &project, const std::s
 		for (StringList::const_iterator i = setup.libraries.begin(); i != setup.libraries.end(); ++i)
 			libraries += *i + ".lib;";
 
+		if (_version == 14)
+		{
```
clone2727: Formatting

dreammaster:
Looks like simply turning off the exception warnings worked just as well.

PR #619, opened 2015-10-03
I18N: Updated german translation

rootfather:
Hi folks,

I update the german .po file to address the changes that where made a few weeks ago. Really nothing spectacular, just a very few, very minor strings. I also fixed a string in the recorder dialog which was too long for the GUI.

I also updated the german NEWS file to match the current revision of the original file.

Btw: Where in the GUI are the files predictivedialog.cpp and onscreendialog.cpp included? I couldn't find the matching strings in the GUI using today's daily build.

Please note that I found two more missing entries in the .po file:
* the recorder dialog has a string "Unknown Author" - no proper reference in the .po
* "Subtitle speed" text using +/- while playing a game is also missing in the .po

Best regards
rootfather

### DIFF
--- a/doc/de/Neues
+++ b/doc/de/Neues
@@ -17,7 +17,7 @@ Sie auf Englisch unter:
 
  AGI:
    - Es ist nun möglich, die Maus-Unterstützung zu deaktivieren (außer bei
-Amiga-Versionen und Fan-Spielen, die eine Maus benötigen).
+     Amiga-Versionen und Fan-Spielen, die eine Maus benötigen).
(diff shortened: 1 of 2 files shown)


rootfather:
Hello again, @criezy, do you want to check this pull request and maybe merge this changes? There are only a few additions to some missing strings and a few additions to the NEWS file, nothing spectacular at all.

Maybe Simon @SimSaw wants to review this too?

Best regards
rootfather

criezy:
I have now merged this pull request.

PR #620, opened 2015-10-05
NEVERHOOD: Don't crash when loading savegames with a high slot ID

salty-horse:
If you have a list of savegames with a gap in the middle:
  neverhood-win.000
  neverhood-win.001
  neverhood-win.003

loading a savegame after the gap will cause a crash.
This is caused by a confusion in the code between ScummVM's slot (3 in this case),
and the index in the engine's savegame array.

### DIFF
--- a/engines/neverhood/menumodule.cpp
+++ b/engines/neverhood/menumodule.cpp
@@ -93,6 +93,10 @@ void MenuModule::setLoadgameInfo(uint index) {
 	_savegameSlot = (*_savegameList)[index].slotNum;
 }
 
+void MenuModule::setLoadgameSlot(int slot) {
+	_savegameSlot = slot;
+}
+
 void MenuModule::setSavegameInfo(const Common::String &description, uint index, bool newSavegame) {
 	_savegameDescription = description;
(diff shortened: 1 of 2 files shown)


eriktorbjorn:
Just to clarify, this crash seems to only happen (at least for me) when using ScummVM's save/load dialogs, not when using the original ones.

eriktorbjorn:
Surprisingly, this also seems to fix a similar problem when deleting games using the ScummVM dialog. I guess they don't use setDeletegameInfo(). (Again, this already works for me with the original dialogs.)

It does seem a bit unfortunate to have two different functions that do the same thing in subtly different ways - if it's just used once, the call to the original function could be adjusted instead - but I don't know... It's working, as far as I can tell.

salty-horse:
Yes, I was using ScummVM's save/load system. I forgot you could enable the original game's.

bluegr:
Looks good, thanks! I'll test your changes today or tomorrow

salty-horse:
BTW, the commit message can be more accurately renamed to "Don't crash when loading savegames after a gap in the slot ID"

bluegr:
Nice work! Just tested this, and it works as expected, fixing this issue.

Thanks for your work! Merging

PR #621, opened 2015-10-11
NEVERHOOD: Possible fix for bad car behaviour

eriktorbjorn:
This is a possible fix for a glitch in The Neverhood that in some cases causes the car to move up even though you're clicking on a spot below it. I'm not sure if it's the correct fix, but I couldn't find any developer familiar with the engine to discuss it with. And even if it's the correct fix, it's possible that the same reasoning also applies to other parts of the code.

There is an overly long commit message that explains the details, as I understand them, but the short version is that a message is sent to the car with the coordinate to travel to, but it only has an X coordinate. The Y coordinate is always zero.

### DIFF
--- a/engines/neverhood/modules/module2700.cpp
+++ b/engines/neverhood/modules/module2700.cpp
@@ -761,7 +761,7 @@ void Scene2702::moveCarToPoint(NPoint pt) {
 	_tracks.findTrackPoint(pt, minMatchTrackIndex, minMatchDistance, _dataResource);
 	if (minMatchTrackIndex >= 0 && minMatchTrackIndex != _currTrackIndex) {
 		_newTrackIndex = minMatchTrackIndex;
-		_newTrackDestX = pt.x;
+		_newTrackDest = pt;
 		if (_isUpperTrack) {
 			if (_currTrackIndex == 0)
 				sendMessage(_asCar, 0x2003, _trackPoints->size() - 1);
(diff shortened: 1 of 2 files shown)


salty-horse:
The related bug number is buried in the commit message. For reference, it's [#6932](http://sourceforge.net/p/scummvm/bugs/6932/).

johndoe123:
I just had a look at the disasm and it seems your fix is correct.

eriktorbjorn:
Thanks for looking! Did you check all the calls to that function, or just the two I had changed?

johndoe123:
I just checked more, there's a variable _newTrackDestX in Scene2702 and Scene2706 which should be of type NPoint, it seems it should keep both x/y values and not just the x value.
Can you add that fix as well? Thanks!

eriktorbjorn:
Ok, I've made the change, I think. I didn't notice any difference at all in behavior, though: Those two scenes worked fine for me both before and after. (They're at or nearby to the left of the second savegame i made for the bug report.)

bluegr:
Looks good to me as well, nice work! :)

bluegr:
The changes are correct, and I've just tested them. I didn't have any real issues with the car's behavior, but it does work a bit better when exiting these two scenes from the top and bottom. We'll have to wait from the original bug submitter to see if the bug he has experienced has actually been fixed.

Merging, and updating the bug report

PR #348, opened 2013-06-30
AUDIO: Add support for sample rates >65kHz.

fuzzie:
This is needed for 96kHz ogg files in the AGS game The Blackwell Legacy (1.7).

The ARM code isn't updated by this yet.

PR #625, opened 2015-11-09
AUDIO: Add support for sample rates >65kHz.

digitall:
This is an updated version of PR #348 with corrections for ARM ASM.


sev-:
Thanks, merging.

PR #628, opened 2015-11-25
NEWS: Update German NEWS file to address all changes made until 9dbaf8d

rootfather:
I just updated the few stringss that changed until commit 9dbaf8d. Nothing fancy at all, just the announcement of the latest supported games and the addition of the Raspberry Pi port.


sev-:
Great!

PR #630, opened 2015-12-02
IOS: Fixes the iOS port

bSr43:
This pull request brings the compatibility with iOS 7+, the new iPhone 6/6S/6+/6S+ resolution, iTunes file sharing, and fixes many issues with the current iOS version.

It also updates the create_project tool, so that it creates valid Xcode projects for both OS X, and iOS.

Compilation instructions are given in the **backends/platform/iphone/README.md** file.


bluegr:
At a first glance, I have to say this is amazing work! Very well done :)

Plenty of changes to the IOS target, and updates to the create_project too, as well. Kudos!

bSr43:
Thank you :)

backends/platform/iphone/iphone_common.h
```diff
@@ -52,6 +53,11 @@ enum UIViewSwipeDirection {
 	kUIViewSwipeRight = 8
 };
 
+enum UIViewTapDescription {
+	kUIViewTapSingle = 1,
+	kUIViewTapDouble = 2,
```
lordhoto: This has an extra "," at the end which is only legal in C++11, which we do not use.

backends/platform/iphone/iphone_video.mm
```diff
@@ -102,8 +105,10 @@ - (void)createContext {
 	CAEAGLLayer *eaglLayer = (CAEAGLLayer *)self.layer;
 
 	eaglLayer.opaque = YES;
-	eaglLayer.drawableProperties = [NSDictionary dictionaryWithObjectsAndKeys:
-	                                [NSNumber numberWithBool:FALSE], kEAGLDrawablePropertyRetainedBacking, kEAGLColorFormatRGB565, kEAGLDrawablePropertyColorFormat, nil];
+	eaglLayer.drawableProperties = @{
+	                                 kEAGLDrawablePropertyRetainedBacking: @NO,
```
lordhoto: I'm getting the following compiler error on this line on buildbot:
```
backends/platform/iphone/iphone_video.mm:109:74: error: unexpected type name 'BOOL': expected expression
                                         kEAGLDrawablePropertyRetainedBacking: @NO,
                                                                                ^
usr/include/objc/objc.h:50:26: note: expanded from macro 'NO'
#define NO              (BOOL)0
                         ^
```

bSr43: This is the « new » Objective-C 2.0 syntax. Which compiler are you using? Anyway, I’ll switch back to the old notation, there is no need to use the new notation, expect for the ease of reading :)

backends/platform/iphone/iphone_video.mm
```diff
@@ -158,31 +166,74 @@ - (void)createContext {
 	}
 }
 
+- (void)setupGestureRecognizers {
+	UISwipeGestureRecognizer *swipeRight = [[UISwipeGestureRecognizer alloc] initWithTarget:self action:@selector(twoFingersSwipeRight:)];
```
lordhoto: Looks like these swipe gesture recognizers have not been available in older iOS versions (we use SDK version 3.1.2 to support all generations of devices):
```
backends/platform/iphone/iphone_video.mm:170:2: error: unknown type name 'UISwipeGestureRecognizer'
        UISwipeGestureRecognizer *swipeRight = [[UISwipeGestureRecognizer alloc] initWithTarget:self action:@selector(twoFingersSwipeRight:)];     
        ^
```

backends/platform/iphone/iphone_video.mm
```diff
@@ -158,31 +166,74 @@ - (void)createContext {
...
+
+	[self addGestureRecognizer:swipeRight];
+	[self addGestureRecognizer:swipeLeft];
+	[self addGestureRecognizer:swipeUp];
+	[self addGestureRecognizer:swipeDown];
+	[self addGestureRecognizer:doubleTapTwoFingers];
+}
+
+- (CGFloat)optimalScale {
+	CGFloat screenScale = [[UIScreen mainScreen] scale];
```
lordhoto:
```
backends/platform/iphone/iphone_video.mm:198:47: warning: 'UIScreen' may not respond to 'scale'
        CGFloat screenScale = [[UIScreen mainScreen] scale];
                               ~~~~~~~~~~~~~~~~~~~~~ ^
```

I am pretty sure our old code had special code to detect this case. I think it's best to add it again.

backends/platform/iphone/iphone_video.mm
```diff
@@ -158,31 +166,74 @@ - (void)createContext {
...
+	if (screenScale < 2) return screenScale;
+
+	if (UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad) {
```
lordhoto: I guess there's more SDK version fun here:
```
platform/iphone/iphone_video.mm:201:6: error: use of undeclared identifier 'UI_USER_INTERFACE_IDIOM'
        if (UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad) {
            ^
platform/iphone/iphone_video.mm:201:35: error: use of undeclared identifier 'UIUserInterfaceIdiomPad'
        if (UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad) {
                                         ^
```

backends/platform/iphone/iphone_video.mm
bSr43: This is something I feared, indeed… The major issue is that the handling of screen orientation has changed with iOS 7, and writing something which works on both cases is very complex.

backends/platform/iphone/iphone_video.mm
```diff
@@ -158,31 +166,74 @@ - (void)createContext {
...
+		return 1;
+	}
+
+	CGSize screenSize;
+	UIScreen *mainScreen = [UIScreen mainScreen];
+	if ([mainScreen respondsToSelector:@selector(nativeBounds)]) {
+		screenSize = [mainScreen nativeBounds].size;
```
lordhoto:
```
backends/platform/iphone/iphone_video.mm:208:28: warning: instance method '-nativeBounds' not found (return type defaults to 'id') [-Wobjc-method-access]
                screenSize = [mainScreen nativeBounds].size;                                                                
                                         ^~~~~~~~~~~~                                                                       
backends/platform/iphone/iphone_video.mm:208:42: error: property 'size' not found on object of type 'id'         
                screenSize = [mainScreen nativeBounds].size;
                                                       ^
```

This time you check whether the selector is actually available. But it doesn't know about the proper type here. Not sure how to resolve this, this is a bit different from the code we used for the scale selector handling...

backends/platform/iphone/iphone_video.mm
bSr43: Does your toolchain can be downloaded somewhere?
At this time, this port mostly dropped support for very old iOS versions. But if I can try to compile with the same toolchain than yours, I could try to merge the old support, with the new one.

backends/platform/iphone/iphone_video.mm
```diff
@@ -657,62 +688,58 @@ - (bool)getMouseCoords:(CGPoint)point eventX:(int *)x eventY:(int *)y {
...
 
 - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
-	NSSet *allTouches = [event allTouches];
 	int x, y;
 
-	switch ([allTouches count]) {
-	case 1: {
-		UITouch *touch = [touches anyObject];
-		CGPoint point = [touch locationInView:self];
+	NSSet<UITouch *> *allTouches = [event allTouches];
```
lordhoto: This one looks really odd to me, but I don't have the time to investigate this further right now:
```
backends/platform/iphone/iphone_video.mm:717:16: error: expected '>'
        NSSet<UITouch *> *allTouches = [event allTouches];
                      ^
backends/platform/iphone/iphone_video.mm:717:17: error: expected unqualified-id
        NSSet<UITouch *> *allTouches = [event allTouches];
                       ^
```

backends/platform/iphone/iphone_video.mm
bSr43: It comes from the SDK you are using. In fact, the latest SDK defines this selector, so the compiler knows how to deal with it. That being said, if you want to target old devices, you don’t need to use an old SDK, you just need to select the proper deployment target.

Another thing, while I’m thinking about all this, the support for the new iPhone 6 resolution requires the use of the new image asset catalog, which is something that old devices cannot deal with. It means that I think we should have two backends, for the old devices, and for the new ones.

backends/platform/iphone/iphone_video.mm
bSr43: the « <UITouch *> » can be removed, this is only a new syntax element, introduced last year in clang, in order to provide information about the objects in the collection.

lordhoto:
I sadly can't test the changes because this breaks compilation for me. It looks like it uses a lot of newer SDK features, which seem to indicate the port is breaking old devices. Which is something we want to avoid.

What is missing before I can give any further comments is:
* Make code compile with old SDK versions again (preferably 3.1.2, so we can still support all device generations out there).
* Make code compile with our clang 3.3 based toolchain. It seems there are some new syntax features used, which are not supported by this compiler.
* Test code on an actual old iOS device. I can do this, if I get the changes to compile.

If you need any instructions on how to get the toolchain to build or the SDK we use. Please drop me an email. It's simply my nickname followed by the crazy a symbol followed by scummvm.org.

bSr43:
I sent you an email :)

bluegr:
Sadly, this now has merge conflicts, probably because of the recent changes to xcode. Could you please fix them?

bSr43:
It is merged…

bluegr:
This does look OK to be merged now, since it does work with older IOS SDKs. Once the conflicts are resolved, is it in a good state to merge?

Note that there has been some duplicated effort in pull request #645 (though not as much as what has been done in this pull request). It would be nice if this was merged first, and the changes in the other pull request were added afterwards.

So, for now, please resolve the conflicts when you can :) Thanks!

bSr43:
it's done…

lordhoto:
There are still a few issues we need to tackle before this is good to go.

The iOS7 code uses our software scalers. While this is nice in theory, it is bad in practice. This limits the port to 16bit color precision, i.e. preventing it from coping with engines like Wintermute, and Sword25.
Since iOS7+ devices support OpenGL ES 2, we can think of using GLSL based filter implementations in the future (I am working on this in general for ScummVM right now, so there is no need to handle this in this pull request or iOS specific).
I plan to get rid of the iPhone specific OpenGL implementation and replace it with our generic OpenGLGraphicsManager. There are similar plans for our Android port. This will reduce the amount of code we need to maintain, while giving all our ports a similar amount of features. Introducing this scaler feature, and then removing it in the near future is confusing to the users (if we want to merge this before the upcoming 1.8.0 release), thus it is a showstopper for this PR.

Right now, the iOS7 code introduces a lot of code duplication by copying over all iPhone backend files and modifying them. This has all the well known down-sides: chances that we need to update the same code twice, etc. It is much preferable to have it all based on one common code, i.e. we could have a OSystem_IOSBase, and two subclasses for pre-iOS7 and iOS7+, residing in backends/platform/iphone and the build setup selects which one to use.

There are some odd changes to non-iOS-port specific files. Some seem supsicious. For example, `dists/macosx/Info.plist` was changed, but not the `dists/macosx/Info.plist.in` file. Addtionally, the file seems to refer to variables, it is unclear to me how our configure+make based build system assure that these are replaced properly. I will also need to look into changes to `common/` code, I'll get back to that in the next few days.

bSr43:
Regarding the scalers, it could be disabled, indeed (even if I love the current feature on my iPad, but I understand the problem :-))

For the code duplication, I completely agree with you, but maybe you should see this PR only as a first working version, which will be improved later. It required a lot of efforts to make it work smoothly on new iOS devices, but I haven't seen any real interest in merging my PR during the last weeks, although I tried to implement and modify all that have been discussed by email, resulting in lots of new modifications (including this duplication, BTW). I even wrote this script to compile a toolchain for your build system, so that you can easily integrate :-/

About the template files, they was not used by the configure script, even before I started to make changes. I think that those ".in" files could be safely removed.

lordhoto:
> For the code duplication, I completely agree with you, but maybe you should see this PR only as a first working version, which will be improved later.

Whoops, sorry looks like I didn't adjust all of my comment before posting... yes, we can refactor this later. This is why I didn't talk about it being a showstopper.

> About the template files, they was not used by the configure script, even before I started to make changes. I think that those ".in" files could be safely removed.

They are used by our configure+make based build system. The targets `bundle` (for Mac OS X) and `iphonebundle` for iPhone use them to create the bundle files. This is used for release builds and thus should be working.

The `.in` version is required to automatically update our ScummVM version information (i.e. 1.8.0git etc.) whenever it changes. Btw. this reminds me: You need to add the ios7 file the `subs_files` "array" in `devtools/update_version.pl`.

bSr43:
OK, I'll take a look tomorrow (it's midnight here, in France :-)). It'll be easy to update the ".in" files, and I think that I'll surround all the scalers code with a disabled macro. I'll also try to evaluate the amount of work for refactoring of the duplicated code, but if that's not a showstopper for you, maybe I'll create a new PR later :-)

audio/softsynth/fluidsynth.cpp
```diff
@@ -179,7 +184,16 @@ int MidiDriver_FluidSynth::open() {
 
 	const char *soundfont = ConfMan.get("soundfont").c_str();
 
+#ifdef IPHONE_OFFICIAL
+	char *soundfont_fullpath[PATH_MAX];
+	const char *document_path = iOS7_getDocumentsDir();
+	strcpy((char *) soundfont_fullpath, document_path);
+	strcat((char *) soundfont_fullpath, soundfont);
```
lordhoto: I suggest to use a `Common::String` as buffer here. Just for future reference: `strcpy`/`strcat` are pretty unsafe, we usually use `Common::strlcpy` and `Common::strlcat` instead, which are implementing OpenBSD's `strlcpy` and `strlcat`.

bSr43:
The modifications are done!

This explains why I thought that the `.in` files were not used anymore: I haven't seen the `update-version.pl` script!

audio/softsynth/fluidsynth.cpp
```diff
@@ -179,7 +184,16 @@ int MidiDriver_FluidSynth::open() {
 
 	const char *soundfont = ConfMan.get("soundfont").c_str();
 
+#ifdef IPHONE_OFFICIAL
```
lordhoto: This whole block needs a `FIXME/HACK` comment. It indicates that the chroot FS implementation breaks functionality in unexpected ways. In the future we need to get rid of this special code for iOS.

audio/softsynth/fluidsynth.cpp
```diff
@@ -179,7 +184,16 @@ int MidiDriver_FluidSynth::open() {
...
+	char *soundfont_fullpath[PATH_MAX];
+	const char *document_path = iOS7_getDocumentsDir();
```
lordhoto: Looks like `iPHONE_OFFICIAL` is a valid define for pre-iOS7 backend too. Thus, this is probably breaking build for that.

audio/softsynth/fluidsynth.cpp
bSr43: You're right. I changed it.

audio/softsynth/fluidsynth.cpp
bSr43: You're right. I changed it.

backends/fs/chroot/chroot-fs-factory.h
```diff
@@ -0,0 +1,41 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#ifndef CHROOT_FS_FACTORY_H
+#define CHROOT_FS_FACTORY_H
+
+#include "backends/fs/fs-factory.h"
+
+class ChRootFilesystemFactory : public FilesystemFactory {
```
lordhoto: Please add a `FIXME` comment which indicates that using this factory in your backend breaks features silently. Instances are, for example, the FluidSynth code, and the POSIX plugin code.

backends/fs/chroot/chroot-fs.h
```diff
@@ -0,0 +1,54 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#ifndef CHROOT_FS_H
```
lordhoto: `BACKENDS_FS_CHROOT_CHROOT_FS_H`, i.e. we use the file path (most of the time).

backends/fs/chroot/chroot-fs-factory.h
```diff
@@ -0,0 +1,41 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#ifndef CHROOT_FS_FACTORY_H
```
lordhoto: `BACKENDS_FS_CHROOT_CHROOT_FS_FACTORY_H`, i.e. we use the file path (most of the time).

common/str.h
```diff
@@ -217,6 +218,11 @@ class String {
 	 */
 	void trim();
 
+    /**
+     * Add a path component
+     */
+    String stringByAppendingPathComponent(String component, char sep = '/') const;
```
lordhoto: This is definitiely nothing which should be in `Common::String` itself, it is a helper for filenames not a generic string helper. Since you only seem to use it in one place, I suggest to move it there (i.e. the chroot FS code).

But I think you don't even need this. As long as you can assure that `_root` is not empty, you can simply compose paths with: `Common::normalizePath(_root + '/' + component, '/');` in your case.

gui/module.mk
```diff
@@ -32,6 +32,13 @@ MODULE_OBJS := \
 	widgets/scrollbar.o \
 	widgets/tab.o
 
+# Even if it seems redundant, please keep
+# these directives in that order!
+# This is needed by the "create_project" tool, for the OS X / iOS Xcode project
```
lordhoto: I don't understand how this works. create_project seems to set both `IPHONE` and `MACOSX` when going over our files. Now, the comment in [here](https://github.com/bSr43/scummvm/blob/ios-fix/devtools/create_project/create_project.cpp#L343-L347) suggests that create_project selects based on the suffix. But, if IPHONE is defined only `browser.o` is ever picked up?

At any rate, this deserves a `HACK` (and probably `FIXME`) indication.

ports.mk
```diff
@@ -92,11 +92,139 @@ endif
 	cp $(srcdir)/dists/iphone/icon-72.png $(bundle_name)/
 	cp $(srcdir)/dists/iphone/Default.png $(bundle_name)/
 
+ios7bundle: ios7
+	mkdir -p $(bundle_name)
+	awk 'BEGIN {s=0}\
```
lordhoto: What's the reason we do this here, instead of keeping it in the `Info.plist` file already?

ports.mk
```diff
@@ -172,6 +300,13 @@ iphone: $(OBJS)
 		-framework CoreFoundation -framework QuartzCore -framework Foundation \
 		-framework AudioToolbox -framework CoreAudio -lobjc -lz
 
+ios7: $(OBJS)
```
lordhoto: Since this is the same as `iphone` I suggest to use `iphone` for simplicity reasons.

backends/fs/chroot/chroot-fs.cpp
```diff
@@ -0,0 +1,99 @@
...
+bool ChRootFilesystemNode::isReadable() const {
+	return _realNode->isReadable();
+}
+
+bool ChRootFilesystemNode::isWritable() const {
+	return _realNode->isWritable();
+}
+
+AbstractFSNode *ChRootFilesystemNode::getChild(const Common::String &n) const {
+	return new ChRootFilesystemNode(_root, (POSIXFilesystemNode *) _realNode->getChild(n));
```
lordhoto: Formatting, we don't use a space between `(POSIXFilesystemNode *)` and `_realNode->getChild(n)`.

common/fs.h
```diff
@@ -58,10 +58,11 @@ class FSList : public Array<FSNode> {};
 class FSNode : public ArchiveMember {
 private:
 	SharedPtr<AbstractFSNode>	_realNode;
-	FSNode(AbstractFSNode *realNode);
 
 public:
-	/**
+    FSNode(AbstractFSNode *realNode);
```
lordhoto: Formatting, we use tabs for indentation.

common/fs.h
```diff
@@ -58,10 +58,11 @@ class FSList : public Array<FSNode> {};
...
+
+    /**
```
lordhoto: Also, here, use a tab instead of spaces.

common/fs.h
lordhoto: Actually, since this is public now, we need to specify that FSNode takes the ownership of the pointer and will delete it at some point. Please add a comment for that.

ports.mk
bSr43: The reason why I had to edit the file here is related to how Xcode handles this file, due to the usage of Asset Catalog. If we put the keys directly in the Info.plist file, these are removed by Xcode each time we compile the project, and it changes the file on disk.

devtools/create_project/xcode.cpp
```diff
@@ -954,10 +1031,14 @@ void XcodeProvider::setupDefines(const BuildSetup &setup) {
...
+	REMOVE_DEFINE(_defines, "MACOSX");
+	REMOVE_DEFINE(_defines, "IPHONE");
+	REMOVE_DEFINE(_defines, "IPHONE_OFFICIAL");
+	REMOVE_DEFINE(_defines, "IPHONE_OFFICIAL_IOS7");
+	REMOVE_DEFINE(_defines, "SDL_BACKEND");
 	ADD_DEFINE(_defines, "CONFIG_H");
-	ADD_DEFINE(_defines, "SCUMM_NEED_ALIGNMENT");
-	ADD_DEFINE(_defines, "SCUMM_LITTLE_ENDIAN");
 	ADD_DEFINE(_defines, "UNIX");
+	ADD_DEFINE(_defines, "XCODE");
```
lordhoto: Why do we need an XCODE define? I can't seem to find any use of it.

devtools/create_project/xcode.cpp
```diff
@@ -1033,7 +1131,7 @@ std::string XcodeProvider::writeProperty(const std::string &variable, Property &
 
 		output += writeSetting(setting->first, setting->second);
 
-		// The combination of kSettingsAsList, and kSettingsSingleItem should use "," and not ";" (i.e children
+		// The combination of SettingsAsList, and kSettingsSingleItem should use "," and not ";" (i.e children
```
lordhoto: Looks like the "k" was removed for no reason?

devtools/create_project/xcode.h
```diff
@@ -150,7 +152,7 @@ class XcodeProvider : public ProjectProvider {
 	struct Object {
 	public:
 		std::string _id;                // Unique identifier for this object
-		std::string _name;              // Name (may not be unique - for ex. configuration entries)
+ 		std::string _name;              // Name (may not be unique - for ex. configuration entries)
```
lordhoto: The addition of the space(?) here looks wrong.

lordhoto:
How did you create the splash screen images? The orange used in them looks completely different from our standard orange. It's much lighter. We will need to fix this.

The chroot FS thing is a nice idea. Sadly, it seems there are multiple issues with it due to our FSNode API :-/. We'll definitely need to revamp this/bring this up to speed in the future.
However, the whole situation makes me wonder: as far as I understand both builds are not compatible when it comes to configuration files. Now, the funny thing is: `IPHONE_OFFICIAL` is the version, which is not our official release version. I think we should rename the define to something like `USE_IOS_SANDBOX`, which gets rid of the ambiguousness (i.e. official can be seen as talking about the official Apple way of doing things or as an indiciation that this is an official build).

I think this is a fine initial state for the port now. However, before merge I would really like to hear back about the oddity in `gui/module.mk`, I am not confident that `browser_osx.o` is picked up by create_project (except there's major hackery going on in the XCode project file generation, which might very well be the case).

bSr43:
Regarding the color, I have built the shade with Pixelmator, but I'm clearly not a graphist, indeed. Maybe it has something to do with color calibration, or something like that.

I think that renaming the `IPHONE_OFFICIAL` with `USE_IOS_SANDBOX` sound great, indeed!

Regarding the last point, in `gui/module.mk`, if you build the project and take a look at the generated Xcode project, you'll notice that the `browser_osx.cpp` is excluded from compilation for the iOS target, but not the OS X target. As I mentioned in the updated comment, the problem is that the `create_project` tools starts by collecting ALL the files according to the selected backend (Xcode in this case), but the Xcode backend actually generates two distinct Xcode target. So, I need to get the `browser_osx.cpp` file in any case, and filter it out later for the Xcode target. Because both macros are defined during the source gathering, the file is guaranteed to be taken.

bSr43:
this is a lot of hack, but this is mainly due to the design of the create_tool project not suited for a multi-target project generation, but this is so convenient to have both target in the same Xcode project! And I had to maintain the compatibility with your build system…

dists/macosx/Info.plist
```diff
@@ -4,6 +4,18 @@
...
+	<key>CFBundleDisplayName</key>
+	<string>ScummVM</string>
+	<key>CFBundleExecutable</key>
+	<string>scummvm</string>
+	<key>CFBundleGetInfoString</key>
+	<string>1.8.0git, Copyright 2001-2016 The ScummVM Team</string>
+	<key>CFBundleIconFile</key>
+	<string>scummvm.icns</string>
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
```
lordhoto: Looks like the `in` file hardcodes this to `org.scummvm.scummvm`. This makes this gone next time we change our source version. Always make changes to the in file and use `./devtools/update_version.pl` to re-generate the files.

Additionally: the `bundle` target in `ports.mk` has not been updated to replace the variable. This needs to be done.

devtools/create_project/xcode.cpp
```diff
@@ -26,26 +26,33 @@
...
+
+#define CONTAINS_DEFINE(defines, name) \
+	(std::find(defines.begin(), defines.end(), name) != defines.end())
+
 #define ADD_SETTING(config, key, value) \
 	config._settings[key] = Setting(value, "", kSettingsNoQuote);
 
 #define ADD_SETTING_ORDER(config, key, value, order) \
-	config._settings[key] = Setting(value, "", kSettingsNoQuote, 0, order);
+	config.settings[key] = Setting(value, "", SettingsNoQuote, 0, order);
```
lordhoto: The removal of the `k` prefix looks wrong.

backends/fs/chroot/chroot-fs.cpp
```diff
@@ -0,0 +1,112 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "chroot-fs.h"
```
lordhoto: This should be `backends/fs/chroot/chroot-fs.h`.

lordhoto: This file does not compile right now because it does include `backends/fs/posix/posix-fs.h`, which includes system headers. You will need to add:
```
// Re-enable some forbidden symbols to avoid clashes with stat.h and unistd.h.
// Also with clock() in sys/time.h in some Mac OS X SDKs.
#define FORBIDDEN_SYMBOL_EXCEPTION_time_h
#define FORBIDDEN_SYMBOL_EXCEPTION_unistd_h
#define FORBIDDEN_SYMBOL_EXCEPTION_mkdir
#define FORBIDDEN_SYMBOL_EXCEPTION_getenv
#define FORBIDDEN_SYMBOL_EXCEPTION_exit		//Needed for IRIX's unistd.h
```
before including anything in this file (taken from `backends/fs/posix/posix-fs.cpp`).

lordhoto:
We need to fix the history, i.e. commit 37a655cd1969eb9e30ed51b434cc2c7bb52cec7d in specific. This breaks compilation on GNU/Linux amd64 with g++ for me. It also breaks compilation on non-POSIX systems. If you need git help, you can ask wjp or me. If we are at history fixing, we can also drop commits 9450f1ebcc2887aea03da56c21258db8d167affd and 9450f1ebcc2887aea03da56c21258db8d167affd.

Regarding `gui/module.mk`: Yeah, this is a bug in `create_project`. It only works by accident. We should add a comment that this needs to be fixed in the future.

dists/macosx/Info.plist
bSr43: Yes, now that I know about the `update-version.pl` tool, I'll take care of these mistakes.

By the way, the latest Xcode requires the Info.plist to not hardcore the bundle identifier, but rather use the macro which will be replaced by a macro preprocessor during the packaging of the application. So, I decided to do the same for OS X as iOS, i.e., let the `$(PRODUCT_BUNDLE_IDENTIFIER)` macro in the template, and add a `sed` command in the `bundle` target of the makefile.

bSr43:
Just before rewritting history, could you please tell me if the current repository at 040141940591e474b5882d6cfaa1aae90d794ecc compiles correctly?

lordhoto:
> Just before rewritting history, could you please tell me if the current repository at 040141940591e474b5882d6cfaa1aae90d794ecc compiles correctly?

Compiles fine on GNU/Linux amd64 and with MinGW64 here.

It seems I missed that the chroot files are only compiled when POSIX is set, thus the definition check isn't strictly necessary. I stll suggest you to do the same in `backends/fs/chroot-fs-factory.cpp` (following the style by the POSIX files).

Not sure how much history rewriting you want to do, but personally I would fixup all the comment additions, formatting changes, etc. too.

bSr43:
The history has been rewritten. It should be OK now.

lordhoto:
Can we get the rename of `IPHONE_OFFICIAL` in? I'll take a last look at it tomorrow and merge it if nobody else comments till then (should be around noon CET).

bSr43:
Yes, I'll do that in a few hours, hopefully before noon!

bSr43:
@lordhoto, it should be OK now.

lordhoto:
Great! Thanks!

bluegr:
Just wanted to say: excellent work by both @bSr43 and @lordhoto. Thanks guys! :)

PR #632, opened 2015-12-04
Fixes GLES1 rendering. Removes redundant dispmanx renderer. Fixes Rpi building against SDL2.

vanfanel:
This pull request does the following:

-Fixes the GLES1 context detection and initialization in both the configure script and the openglsdl class by specifying the GLES version we need, because SDL2 creates a GLES2 context by default and that leads to failing GLES1 functions (INVALID_ENUM or INVALID_OPERATION errors were produced).

-Removes the dispmanx renderer, which is not needed anymore since SDL2 on the Raspberry Pi uses dispmanx/gles2 anyway, so no need for a specific renderer that bloats the codebase.

-Specifies SDL2 renderer and CXX11 usage on the configure script when host is raspberry pi: CXX11 is not detected properly but the official Raspberry Pi cross-compiler for X86 supports it.
SDL2 guarantees we will use it on this platform, which is very well supported while SDL1 is bit-rotten and has caused problems like black screens on exit, etc.

-Modifies README.RASPBERRYPI accordingly.


vanfanel:
@lordhoto : I have reverted the c++11 enforcement. It was not necessary, you're right. I had not realized c++11 detection depended on _use_cxx11. Sorry.

As for SDL2, yes: it's actually preferred and not forced because find_sdlconfig is still used as you said.
Using SDL2 on the Pi has became fundamental. The SDL1 code is unmantained, reports on black screens on the Pi appeared last years and they are supposedly to be fixed (https://github.com/raspberrypi/firmware/issues/354 for example), but there are still issues and I still get occasional black screens on SDL1 apps: since SDL1 is considered legacy an is unmantained, fixing those is becoming a nonsense effort.
Also, SDL2 does a perfect job on managing dispmanx/gles rendering, which eliminates the need of custom renderer classes for both plain SDL2 (which uses dispmanx/GLES2 under the hood if an accelerated context is created) and SDL2-driven OpenGL_ES renderer.

The only problem with SDL2 is that plain SDL2 backend on both Raspbian/Pi and X86 is that mouse cursor can go out of the screen area in fullscreen mode, but I hope you can fix it very easily. I tried but failed to understand how to "limit" the mouse cursor coordinates to the screen area. 
Other than that, preferring SDL2 on the Pi is a very good option right now for the aforementioned reasons.

lordhoto:
Manually merged. Thanks.

PR #634, opened 2015-12-09
SWORD25: Fix typo

jwilk:


salty-horse:
The word before that, "vertecies", also seems to be wrong. Isn't it "vertices"?

tramboi:
Yes, and the phrase is weird anyway, hence my proposition to remove the comment altogether.

wjp:
The translation done in 53a9d2d0a1dab1119dc1cc12886321fa72743061 is just wrong.

PR #636, opened 2015-12-14
LAB: Engine

sev-:
This is a pull request for the game The Labyrinth of Time. This game is currently available on www.wyrmkeep.com and on GOG.com. The game should be completable: it was completable a month ago and we did regularly regression testing. All the work has been based on sources kindly provided by Wyrmkeep.

The DOS and the Windows versions are supported by this engine.

### DIFF
--- a/engines/lab/anim.cpp
+++ b/engines/lab/anim.cpp
@@ -130,7 +130,7 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 			return;
 		}
 
-		_vm->_music->updateMusic();
+		_vm->updateMusicAndEvents();
 		_header = _diffFile->readUint32LE();
 		_size = _diffFile->readUint32LE();
 
@@ -191,7 +191,7 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 		case 31:
 			if (_waitForEffect) {
 				while (_vm->_music->isSoundEffectActive()) {
-					_vm->_music->updateMusic();
+					_vm->updateMusicAndEvents();
 					_vm->waitTOF();
 				}
 			}
@@ -211,7 +211,7 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 
 				if (_waitForEffect) {
 					while (_vm->_music->isSoundEffectActive()) {
-						_vm->_music->updateMusic();
+						_vm->updateMusicAndEvents();
 						_vm->waitTOF();
 
 						if (_vm->_graphics->_dispBitMap->_drawOnScreen)
@@ -251,7 +251,7 @@ void Anim::stopDiffEnd() {
 
 	_stopPlayingEnd = true;
 	while (_isPlaying) {
-		_vm->_music->updateMusic();
+		_vm->updateMusicAndEvents();
 		diffNextFrame();
 	}
 }
--- a/engines/lab/anim.cpp
+++ b/engines/lab/anim.cpp
@@ -68,7 +68,6 @@ Anim::Anim(LabEngine *vm) : _vm(vm) {
 	_doBlack = false;
 	_diffWidth = 0;
 	_diffHeight = 0;
-	_drawBitMap = _vm->_graphics->_dispBitMap;
 
 	for (int i = 0; i < 3 * 256; i++)
 		_diffPalette[i] = 0;
@@ -79,13 +78,14 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 		// Already done.
 		return;
 
-	if (_vm->_graphics->_dispBitMap->_drawOnScreen)
-		_vm->_graphics->_dispBitMap->_planes[0] = _vm->_graphics->getCurrentDrawingBuffer();
+	BitMap *disp = _vm->_graphics->_dispBitMap;
+	if (disp->_drawOnScreen)
+		disp->_planes[0] = _vm->_graphics->getCurrentDrawingBuffer();
 
-	_vm->_graphics->_dispBitMap->_planes[1] = _vm->_graphics->_dispBitMap->_planes[0] + 0x10000;
-	_vm->_graphics->_dispBitMap->_planes[2] = _vm->_graphics->_dispBitMap->_planes[1] + 0x10000;
-	_vm->_graphics->_dispBitMap->_planes[3] = _vm->_graphics->_dispBitMap->_planes[2] + 0x10000;
-	_vm->_graphics->_dispBitMap->_planes[4] = _vm->_graphics->_dispBitMap->_planes[3] + 0x10000;
+	disp->_planes[1] = disp->_planes[0] + 0x10000;
+	disp->_planes[2] = disp->_planes[1] + 0x10000;
+	disp->_planes[3] = disp->_planes[2] + 0x10000;
+	disp->_planes[4] = disp->_planes[3] + 0x10000;
 
 	_vm->_event->mouseHide();
 
@@ -123,7 +123,7 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 			_isAnim = (_frameNum >= 3) && (!_playOnce);
 			_curBit = 0;
 
-			if (_vm->_graphics->_dispBitMap->_drawOnScreen)
+			if (disp->_drawOnScreen)
 				_vm->_graphics->screenUpdate();
 
 			// done with the next frame.
@@ -145,14 +145,14 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 		case 10:
 			if (onlyDiffData)
 				warning("Boom");
-			_diffFile->read(_drawBitMap->_planes[_curBit], _size);
+			_diffFile->read(disp->_planes[_curBit], _size);
 			_curBit++;
 			break;
 
 		case 11:
 			curPos = _diffFile->pos();
 			_diffFile->skip(4);
-			_vm->_utils->runLengthDecode(_drawBitMap->_planes[_curBit], _diffFile);
+			_vm->_utils->runLengthDecode(disp->_planes[_curBit], _diffFile);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
@@ -160,21 +160,21 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 		case 12:
 			curPos = _diffFile->pos();
 			_diffFile->skip(4);
-			_vm->_utils->verticalRunLengthDecode(_drawBitMap->_planes[_curBit], _diffFile, _drawBitMap->_bytesPerRow);
+			_vm->_utils->verticalRunLengthDecode(disp->_planes[_curBit], _diffFile, disp->_bytesPerRow);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
 
 		case 20:
 			curPos = _diffFile->pos();
-			_vm->_utils->unDiff(_drawBitMap->_planes[_curBit], _vm->_graphics->_dispBitMap->_planes[_curBit], _diffFile, _drawBitMap->_bytesPerRow, false);
+			_vm->_utils->unDiff(disp->_planes[_curBit], disp->_planes[_curBit], _diffFile, disp->_bytesPerRow, false);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
 
 		case 21:
 			curPos = _diffFile->pos();
-			_vm->_utils->unDiff(_drawBitMap->_planes[_curBit], _vm->_graphics->_dispBitMap->_planes[_curBit], _diffFile, _drawBitMap->_bytesPerRow, true);
+			_vm->_utils->unDiff(disp->_planes[_curBit], disp->_planes[_curBit], _diffFile, disp->_bytesPerRow, true);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
@@ -214,7 +214,7 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 						_vm->updateMusicAndEvents();
 						_vm->waitTOF();
 
-						if (_vm->_graphics->_dispBitMap->_drawOnScreen)
+						if (disp->_drawOnScreen)
 							didTOF = true;
 					}
 				}
--- a/engines/lab/anim.cpp
+++ b/engines/lab/anim.cpp
@@ -321,6 +321,8 @@ void Anim::readDiff(Common::File *diffFile, bool playOnce, bool onlyDiffData) {
 
 		if ((uint32)(_numChunks * 0x10000) < (uint32)(((int32)_diffWidth) * _diffHeight))
 			_numChunks++;
+
+		assert (_numChunks < 16);
 	} else
 		return;
 
--- a/engines/lab/anim.cpp
+++ b/engines/lab/anim.cpp
@@ -68,7 +68,7 @@ Anim::Anim(LabEngine *vm) : _vm(vm) {
 	_doBlack = false;
 	_diffWidth = 0;
 	_diffHeight = 0;
-	DrawBitMap = _vm->_graphics->_dispBitMap;
+	_drawBitMap = _vm->_graphics->_dispBitMap;
 
 	for (int i = 0; i < 3 * 256; i++)
 		_diffPalette[i] = 0;
@@ -145,14 +145,14 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 		case 10:
 			if (onlyDiffData)
 				warning("Boom");
-			_diffFile->read(DrawBitMap->_planes[_curBit], _size);
+			_diffFile->read(_drawBitMap->_planes[_curBit], _size);
 			_curBit++;
 			break;
 
 		case 11:
 			curPos = _diffFile->pos();
 			_diffFile->skip(4);
-			_vm->_utils->runLengthDecode(DrawBitMap->_planes[_curBit], _diffFile);
+			_vm->_utils->runLengthDecode(_drawBitMap->_planes[_curBit], _diffFile);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
@@ -160,21 +160,21 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 		case 12:
 			curPos = _diffFile->pos();
 			_diffFile->skip(4);
-			_vm->_utils->verticalRunLengthDecode(DrawBitMap->_planes[_curBit], _diffFile, DrawBitMap->_bytesPerRow);
+			_vm->_utils->verticalRunLengthDecode(_drawBitMap->_planes[_curBit], _diffFile, _drawBitMap->_bytesPerRow);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
 
 		case 20:
 			curPos = _diffFile->pos();
-			_vm->_utils->unDiff(DrawBitMap->_planes[_curBit], _vm->_graphics->_dispBitMap->_planes[_curBit], _diffFile, DrawBitMap->_bytesPerRow, false);
+			_vm->_utils->unDiff(_drawBitMap->_planes[_curBit], _vm->_graphics->_dispBitMap->_planes[_curBit], _diffFile, _drawBitMap->_bytesPerRow, false);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
 
 		case 21:
 			curPos = _diffFile->pos();
-			_vm->_utils->unDiff(DrawBitMap->_planes[_curBit], _vm->_graphics->_dispBitMap->_planes[_curBit], _diffFile, DrawBitMap->_bytesPerRow, true);
+			_vm->_utils->unDiff(_drawBitMap->_planes[_curBit], _vm->_graphics->_dispBitMap->_planes[_curBit], _diffFile, _drawBitMap->_bytesPerRow, true);
 			_curBit++;
 			_diffFile->seek(curPos + _size, SEEK_SET);
 			break;
--- a/engines/lab/anim.cpp
+++ b/engines/lab/anim.cpp
@@ -39,7 +39,7 @@
 namespace Lab {
 
 Anim::Anim(LabEngine *vm) : _vm(vm) {
-	_header = 0;
+	_lastBlockHeader = 0;
 	_curBit = 0;
 	_numChunks = 1;
 	_headerdata._width = 0;
@@ -74,7 +74,7 @@ Anim::Anim(LabEngine *vm) : _vm(vm) {
 }
 
 void Anim::diffNextFrame(bool onlyDiffData) {
-	if (_header == 65535)
+	if (_lastBlockHeader == 65535)
 		// Already done.
 		return;
 
@@ -131,12 +131,12 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 		}
 
 		_vm->updateMusicAndEvents();
-		_header = _diffFile->readUint32LE();
+		_lastBlockHeader = _diffFile->readUint32LE();
 		_size = _diffFile->readUint32LE();
 
 		uint32 curPos = 0;
 
-		switch (_header) {
+		switch (_lastBlockHeader) {
 		case 8:
 			_diffFile->read(_diffPalette, _size);
 			_isPal = true;
@@ -259,7 +259,6 @@ void Anim::stopDiffEnd() {
 void Anim::readDiff(Common::File *diffFile, bool playOnce, bool onlyDiffData) {
 	_playOnce = playOnce;
 	_delayMicros = 0;
-	_header = 0;
 	_curBit = 0;
 	_frameNum = 0;
 	_numChunks = 1;
@@ -275,65 +274,66 @@ void Anim::readDiff(Common::File *diffFile, bool playOnce, bool onlyDiffData) {
 	_diffFile = diffFile;
 
 	_continuous = false;
-	uint32 signature = _diffFile->readUint32BE();
-	_header = _diffFile->readUint32LE();
+	uint32 signature1 = _diffFile->readUint32BE();
+	uint32 signature2 = _diffFile->readUint32LE();
 
-	if ((signature != MKTAG('D', 'I', 'F', 'F')) || (_header != 1219009121L)) {
+	if ((signature1 != MKTAG('D', 'I', 'F', 'F')) || (signature2 != 1219009121L)) {
 		_isPlaying = false;
 		return;
 	}
 
-	_header = _diffFile->readUint32LE();
+	uint32 signature3 = _diffFile->readUint32LE();
 	_size = _diffFile->readUint32LE();
 
-	if (_header == 0) {
-		// sizeof(headerdata) != 18, but the padding might be at the end
-		// 2 bytes, version, unused.
-		_diffFile->skip(2);
-		_headerdata._width = _diffFile->readUint16LE();
-		_headerdata._height = _diffFile->readUint16LE();
-		// 1 byte, depth, unused
-		_diffFile->skip(1);
-		_headerdata._fps = _diffFile->readByte();
-
-		// HACK: The original game defines a 1 second delay when changing screens, which is
-		// very annoying. We first removed the delay, but it looked wrong when changing screens
-		// as it was possible to see that something was displayed, without being able to tell
-		// what it was. A shorter delay (150ms) makes it acceptable during gameplay and
-		// readable. The big question is: do we need that message?
-		g_system->delayMillis(150);
-
-		if (_headerdata._fps == 1)
-			_headerdata._fps = 0;
-
-		// 4 + 2 bytes, buffer size and machine, unused
-		_diffFile->skip(6);
-		_headerdata._flags = _diffFile->readUint32LE();
-
-		_diffFile->skip(_size - 18);
-
-		_continuous = CONTINUOUS & _headerdata._flags;
-		_diffWidth = _headerdata._width;
-		_diffHeight = _headerdata._height;
-		_vm->_utils->setBytesPerRow(_diffWidth);
-
-		_numChunks = (((int32)_diffWidth) * _diffHeight) / 0x10000;
-
-		if ((uint32)(_numChunks * 0x10000) < (uint32)(((int32)_diffWidth) * _diffHeight))
-			_numChunks++;
-
-		assert (_numChunks < 16);
-	} else
+	if (signature3 != 0)
 		return;
 
-	for (_header = 0; _header < 8; _header++)
-		_rawDiffBM._planes[_header] = nullptr;
+	// sizeof(headerdata) != 18, but the padding might be at the end
+	// 2 bytes, version, unused.
+	_diffFile->skip(2);
+	_headerdata._width = _diffFile->readUint16LE();
+	_headerdata._height = _diffFile->readUint16LE();
+	// 1 byte, depth, unused
+	_diffFile->skip(1);
+	_headerdata._fps = _diffFile->readByte();
+
+	// HACK: The original game defines a 1 second delay when changing screens, which is
+	// very annoying. We first removed the delay, but it looked wrong when changing screens
+	// as it was possible to see that something was displayed, without being able to tell
+	// what it was. A shorter delay (150ms) makes it acceptable during gameplay and
+	// readable. The big question is: do we need that message?
+	g_system->delayMillis(150);
+
+	if (_headerdata._fps == 1)
+		_headerdata._fps = 0;
+
+	// 4 + 2 bytes, buffer size and machine, unused
+	_diffFile->skip(6);
+	_headerdata._flags = _diffFile->readUint32LE();
+
+	_diffFile->skip(_size - 18);
+
+	_continuous = CONTINUOUS & _headerdata._flags;
+	_diffWidth = _headerdata._width;
+	_diffHeight = _headerdata._height;
+	_vm->_utils->setBytesPerRow(_diffWidth);
+
+	_numChunks = (((int32)_diffWidth) * _diffHeight) / 0x10000;
+
+	if ((uint32)(_numChunks * 0x10000) < (uint32)(((int32)_diffWidth) * _diffHeight))
+		_numChunks++;
+
+	assert(_numChunks < 16);
+
+	for (int i = 0; i < 8; i++)
+		_rawDiffBM._planes[i] = nullptr;
 
 	if (_headerdata._fps)
 		_delayMicros = 1000 / _headerdata._fps;
 
+	_lastBlockHeader = signature3;
 	if (_playOnce) {
-		while (_header != 65535)
+		while (_lastBlockHeader != 65535)
 			diffNextFrame(onlyDiffData);
 	} else
 		diffNextFrame(onlyDiffData);
--- a/engines/lab/anim.cpp
+++ b/engines/lab/anim.cpp
@@ -95,9 +95,9 @@ void Anim::diffNextFrame(bool onlyDiffData) {
 
 			if (!onlyDiffData) {
 				if (_headerdata._fps) {
-					uint32 targetMillis = g_system->getMillis() + _delayMicros;
-					while (g_system->getMillis() < targetMillis)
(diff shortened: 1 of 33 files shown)


engines/lab/resource.cpp
```diff
@@ -0,0 +1,344 @@
...
+char *Resource::readString(Common::File *file) {
+	byte size = file->readByte();
+	if (!size)
+		return NULL;
+	char *str = new char[size];
+	char *c = str;
+	for (int i = 0; i < size; i++) {
+		*c = file->readByte();
+		// Decrypt char
+		*c++ = (i < size - 1) ? *c - 95 : '\0';
```
eriktorbjorn: GCC warns that this "operation on 'c' may be undefined".

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1206 @@
...
+	_dispBitMap->_rows         = (maxHeight > 0) ? maxHeight : _screenHeight;
+	_dispBitMap->_drawOnScreen = (memoryBuffer == nullptr);
+	if (memoryBuffer)
+		_dispBitMap->_planes[0] = memoryBuffer;
+
+	_vm->_anim->readDiff(_curBitmap, playOnce, onlyDiffData);
+}
+
+void DisplayMan::freePict() {
+	delete _curBitmap;
```
eriktorbjorn: Should be delete[], I think. Valgrind complains.

engines/lab/resource.cpp
eriktorbjorn: Actually, this seems to be why the game crashes for me, right at the end of the intro. If I change it to

*c = (i < size - 1) ? *c - 95 : '\0';
c++;

It works for me. Without that, I get some message about opening a file where the name is all garbage characters.

digitall:
There are still two remaining compiler warnings:
engines/lab/engine.cpp: In member function ‘void Lab::LabEngine::processMainButton(Lab::CloseDataPtr, uint16&, uint16&, uint16&, bool&, uint16, uint16&)’:
engines/lab/engine.cpp:818:6: warning: parameter ‘wrkClosePtr’ set but not used [-Wunused-but-set-parameter]
 void LabEngine::processMainButton(CloseDataPtr wrkClosePtr, uint16 &curInv, uint16 &lastInv, uint16 &oldDirection, bool &forceDraw, uint16 buttonId, uint16 &actionMode) {
      ^
engines/lab/engine.cpp: In member function ‘void Lab::LabEngine::performAction(Lab::CloseDataPtr, uint16, Common::Point, uint16&)’:
engines/lab/engine.cpp:1083:6: warning: parameter ‘wrkClosePtr’ set but not used [-Wunused-but-set-parameter]
 void LabEngine::performAction(CloseDataPtr wrkClosePtr, uint16 actionMode, Common::Point curPos, uint16 &curInv) {

However, I think these are actually indicating a real latent issue. I think the code is intended to assign nullptr to the wrkClosePtr to "delete" them by assigning NULL. However, this is not happening as the assignment is done to the local function parameter copy... so will be "set but not used" as it does not affect the original passed pointer which is still holding an outdated value and is dangerous if indirected? I might be reading this wrong, but any thoughts?

Strangerke:
processMainButton and performAction are called by fromCrumbs(). 
At the time they are called, wrkClosePtr is always a null pointer: they are called at lines 652, 670 and 674 but wrkClosePtr is only set at line 709.

I'm removing those useless parameters.

dreammaster:
I've had a chance to give a quick look-over, and it looks good. Great work :). Just a few minor comments I'd make:

* Very minor formatting at the end of lab.cpp and eventman.cpp could use an extra space before the closing brace of the namespace.

* As far as I'm aware, the preference for method comments is to put them all in the header files, not in the CPP files.

* Looking at LabEngine::processJournal, it looks like you're doing the loop without any calls to delayMilli(). Unless it's buried inside one of the method calls and I missed it. If not, it's good practice to some kind of delay in those kinds of loops so you don't throttle the CPU.

bluegr:
@dreammaster: Thanks :)

The comments will need to be moved to header files, indeed.

There is an issue with that inner loop in processJournal(), indeed. There seems to be a similar issue with processMonitor() as well.

These will require a bit of testing, so I'll have a look at them tomorrow or Monday.

engines/lab/resource.cpp
```diff
@@ -0,0 +1,430 @@
...
+				head = action;
+			if (prev)
+				prev->_nextAction = action;
+			action->_actionType = file->readSint16LE();
+			action->_param1 = file->readSint16LE();
+			action->_param2 = file->readSint16LE();
+			action->_param3 = file->readSint16LE();
+
+			if (action->_actionType == SHOWMESSAGES) {
+				char **messages = (char **)malloc(action->_param1 * 4);
```
eriktorbjorn: This seems to assume that a pointer is 4 bytes. Valgrind warns about invalid writes in the loop below on my 64-bit system.

engines/lab/image.cpp
```diff
@@ -0,0 +1,148 @@
...
+	uint32 size = _width * _height;
+	if (size & 1)
+		size++;
+
+	_imageData = new byte[size];
+	s->read(_imageData, size);
+}
+
+Image::~Image() {
+	delete _imageData;
```
eriktorbjorn: This should be "delete[] _imageData;"

engines/lab/resource.cpp
wjp: And it seems the code below also doesn't handle the 0 terminator (in both cases).

engines/lab/special.cpp
```diff
@@ -0,0 +1,482 @@
...
+		charsDrawn += _graphics->flowTextToMem(_journalBackImage, _journalFont, -2, 2, 0, false, false, false, true, _utils->vgaRectScale(52, 32, 152, 148), curText);
+	}
+
+	_music->updateMusic();
+	curText = (char *)(_journalText.c_str() + charsDrawn);
+	_lastPage = (*curText == 0);
+	_graphics->flowTextToMem(_journalBackImage, _journalFont, -2, 2, 0, false, false, false, true, _utils->vgaRectScale(171, 32, 271, 148), curText);
+
+	curText = (char *)(_journalText.c_str() + charsDrawn);
+	_lastPage = (*curText == 0);
```
wjp: These lines look suspicious. charsDrawn wasn't updated since the last time curText and _lastPage were set, so this doesn't do anything?

(Also, why all the casts to char* ?)

wjp:
All the Common::String function arguments in this engine seem to be passed by value instead of by const-reference. Why?

engines/lab/map.cpp
```diff
@@ -0,0 +1,577 @@
...
+	default:
+		return;
+	}
+
+	if (drawMarkFl)
+		_imgMapX[_direction]->drawImage(drawX, drawY);
+}
+
+bool LabEngine::floorVisited(uint16 floorNum) {
+	for (uint16 i = 1; i <= _maxRooms; i++) {
```
eriktorbjorn: This looks strange: _maps[] is allocated with "_maps = new MapData[_maxRooms];" and here we access elements 1 through _maxRooms. It may be ok to ignore element 0, but element _maxRooms is outside of the allocated range.

engines/lab/engine.cpp
```diff
@@ -0,0 +1,1204 @@
...
+			gotMessage = true;
+			_followingCrumbs = false;
+			if (!fromCrumbs(curMsg->_msgClass, curMsg->_code, curMsg->_qualifier, curMsg->_mouse, curInv, curMsg, forceDraw, curMsg->_code, actionMode))
+				break;
+		}
+	}
+
+	delete _conditions;
+	delete _roomsFound;
+	delete[] _rooms;
```
eriktorbjorn: The LabEngine destructor uses, and deletes, _rooms, so doing it here first is bad. Perhaps all these deletes should be moved to the destructor?

engines/lab/eventman.cpp
```diff
@@ -0,0 +1,301 @@
...
+			}
+			break;
+		case Common::EVENT_QUIT:
+		case Common::EVENT_RTL:
+		default:
+			break;
+		}
+
+		g_system->copyRectToScreen(_vm->_graphics->_displayBuffer, _vm->_graphics->_screenWidth, 0, 0, _vm->_graphics->_screenWidth, _vm->_graphics->_screenHeight);
+		g_system->updateScreen();
```
wjp: Updating the full screen after every single received event is a very bad idea.

wjp:
I've tried testing it (with the GOG version of the game), but in its current state (current git version), I'm sorry to say it's just not playable. I'm getting random crashes, a sluggish interface, and after a few minutes I seem to be stuck because the elevator in the hotel at the beginning can go up, but not back down.

Looking at the history, the commits in the last few days seem pretty intrusive and dangerous, so quite possibly these are recent regressions.

engines/lab/resource.cpp
wjp: I think it would be much better to just get rid of the double function of the _data field of Action. Why not just give it both string and list-of-strings fields, and use the right one depending on the actionType? Then you get type safety and automatic memory management for free.

wjp:
I tried to test some older revisions to see if these problems I mentioned are indeed regressions, but all revisions I tried seem to just crash with the following message:

openDataFile: Couldn't open <garbage>(<garbage>)!

wjp:
eriktorbjorn points out that ea1466fe94b320b301baacfdd27228ffa5e2b86d is the reason older revisions are crashing for me.

wjp:
The last few commits by @Strangerke fixed the elevator bug for me.

sev-:
I have no objections for merging this and continuing work in-tree.

raziel-:
Was it part fo the deal to only support the commercial versions, or do the freely available Amiga version differ too much engine-wise?

Strangerke:
We are indeed supporting the commercial version right now. The Amiga version is the original one, and the file structure and formats are different (maybe even some hardcoded logic). We'll need to reverse the Amiga version to understand the differences.

wjp:
The code looks decent from a quick scan, but I'm still rather concerned that the game basically wasn't working properly when I tried it this weekend. Maybe it was just very bad timing on my part, but it didn't make a good first impression at all.

Did anyone test it recently (after the flurry of refactoring the last few weeks)? If so, what was your impression?

Any comments from the engine authors on overall completeness and stability?

(NB. The main event loop is still broken as I commented earlier. Please fix that.)

engines/lab/eventman.cpp
bluegr: True, that will need to be refactored (and tested)

bluegr:
@wjp: I've changed all of the string handling code to use Common::String quite recently. Unfortunately, I did not have the time to change the associated parameters so that they are passed by reference, as I wanted to perform more tests on the engine before doing that. Strings are modified in several spots, so I didn't want to break anything, plus the change was quite big already. 

The original code was quite messy. Among its flaws, it did not have a clean way of handling memory, resulting in all sorts of memory allocation issues. Glad you like the current code :) The original one wasn't decent. At all...

I tried to check the game after big changes. A lot of regressions have been fixed already, as you can see from the commit history. The game will most certainly need to be tested again, as there have been a lot of changes and major refactoring since I last played it from start to end.

@raziel-: Regarding the Amiga version, it's using a different font format, plus it's missing the two files that contain the inventory buttons ("control" and "inv"). I can't find any replacements, apart from some GIF files that were used in an electronic manual. The hardcoded logic should be mostly the same, but the palette handling and game fonts are two major changes. At the moment, the only thing playable from the Amiga version is the music.

wjp:
I suggest you finish the current round of refactoring, then take the time to test it properly, and let us know when it's ready for review again, so we can take a fresh look.

And lest I forget to mention it: I also think we should definitely fix up history so that commits from the last few months run ok when built with gcc. (re. ea1466fe94b320b301baacfdd27228ffa5e2b86d)

sev-:
I am really wonder, why we're making a new requirement for an engine merge: for it to be in a perfect state.

Why not continue the work in-tree?

bluegr:
@wjp: The current round of refactoring is almost done. The fact that you found regressions in over 300 commits is to be expected. The game needs to be playtested. But the point of the pull request is to test that the engine is in a state that it can be merged in-tree, not that it's bug-free.

wjp:
I'm not making a new requirement. I'm just making a suggestion.

lordhoto:
> And lest I forget to mention it: I also think we should definitely fix up history so that commits from the last few months run ok when built with gcc. (re. ea1466fe94b320b301baacfdd27228ffa5e2b86d)

+1, this can be incredibly helpful for future bug fixing/bisection

wjp:
To clarify my earlier comment: I'd much prefer to review the engine when it's not in the middle of heavy refactoring.

lordhoto:
> To clarify my earlier comment: I'd much prefer to review the engine when it's not in the middle of heavy refactoring.

Same here. I started looking at it, but if you are going to refactor this while I'm still going through it, I would rather stop and come back when all that is done. Otherwise this gets confusing and even more time consuming really fast.

Strangerke:
I'm not refactoring anything right now. I'm about to start a playthough to see if there are regressions (typically the journal code that wjp pointed at).

wjp:
Ok, great.

Strangerke:
Before you ask, I have checked the original code: those assignments were also useless. It's not an error.

engines/lab/lab.h
```diff
@@ -0,0 +1,473 @@
...
+
+struct CrumbData {
+	uint16 _roomNum;
+	uint16 _direction;
+};
+
+#define MAX_CRUMBS          128
+
+typedef CloseData *CloseDataPtr;
+typedef Common::List<Rule *> RuleList;
```
wjp: It looks like all related code will become much cleaner if `RuleList` is changed to a `Common::List<Rule>` instead.

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1052 @@
...
+			if (nheight <= (height / 8))
+				by = 1;
+			else if (nheight <= (height / 4))
+				by = 2;
+			else if (nheight <= (height / 2))
+				by = 3;
+		}
+	}
+
+	delete[] mem;
```
wjp: `mem` has changed since it was allocated, so this doesn't look right.

wjp: And actually it seems like the destructor of `img` will free that memory too.

engines/lab/music.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+	_musicOn = false;
+	_loopSoundEffect = false;
+	_queuingAudioStream = nullptr;
+	_lastMusicRoom = 1;
+	_doReset = true;
+	_waitTillFinished = false;
+}
+
+void Music::updateMusic() {
```
wjp: `updateMusic` is very poorly named, since it does event processing and screen updating as well. It might be more clear to move this function to the main engine class, and then keep only the actual music updating in the `Music` class.

engines/lab/anim.cpp
```diff
@@ -0,0 +1,340 @@
...
+		return;
+
+	if (_vm->_graphics->_dispBitMap->_drawOnScreen)
+		_vm->_graphics->_dispBitMap->_planes[0] = _vm->_graphics->getCurrentDrawingBuffer();
+
+	_vm->_graphics->_dispBitMap->_planes[1] = _vm->_graphics->_dispBitMap->_planes[0] + 0x10000;
+	_vm->_graphics->_dispBitMap->_planes[2] = _vm->_graphics->_dispBitMap->_planes[1] + 0x10000;
+	_vm->_graphics->_dispBitMap->_planes[3] = _vm->_graphics->_dispBitMap->_planes[2] + 0x10000;
+	_vm->_graphics->_dispBitMap->_planes[4] = _vm->_graphics->_dispBitMap->_planes[3] + 0x10000;
+
```
wjp: This function refers to `_vm->_graphics->_dispBitMap` and to `DrawBitMap` (which is actually a member var of Anim despite the missing underscore), but as far as I can tell, these are always the same thing.

wjp: Also, this function is the only place where these `_planes` are ever used, so that `_planes` array can probably just be moved to a local variable in this function?

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1052 @@
...
+	uint16 height = _vm->_utils->vgaScaleY(149) + _vm->_utils->svgaCord(2);
+
+	while (_vm->_music->isSoundEffectActive()) {
+		_vm->_music->updateMusic();
+		_vm->waitTOF();
+	}
+
+	readPict(filename, true, true);
+	setPalette(_vm->_anim->_diffPalette, 256);
+	byte *mem = _vm->_anim->_rawDiffBM._planes[0];
```
wjp: I can't seem to figure out where `_rawDiffBM._planes[0]` is ever set to anything non-null.

engines/lab/anim.cpp
```diff
@@ -0,0 +1,340 @@
...
+		switch (_header) {
+		case 8:
+			_diffFile->read(_diffPalette, _size);
+			_isPal = true;
+			break;
+
+		case 10:
+			if (onlyDiffData)
+				warning("Boom");
```
lordhoto: Is this something which should never happen? Or why is there such a (to me rather meaningless) warning in this case?

engines/lab/anim.h
```diff
@@ -0,0 +1,105 @@
...
+	uint16 _width;
+	uint16 _height;
+	char _fps;
+	uint32 _flags;
+};
+
+struct BitMap {
+	uint16 _bytesPerRow;
+	bool _drawOnScreen;
+	byte *_planes[16];
```
lordhoto: I don't get why this has 16 indices. The code in anim.cpp only ever seems to write actual values to the first 5 indices. The code accessing indicies seems to depend on user data without any checks, thus potentially reading out of bounds anyway. And the code at the end of `readDiff` seems to reset the first 8 indices to nullptr again.

engines/lab/anim.cpp
```diff
@@ -0,0 +1,340 @@
...
+		_vm->_utils->setBytesPerRow(_diffWidth);
+
+		_numChunks = (((int32)_diffWidth) * _diffHeight) / 0x10000;
+
+		if ((uint32)(_numChunks * 0x10000) < (uint32)(((int32)_diffWidth) * _diffHeight))
+			_numChunks++;
+	} else
+		return;
+
+	for (_header = 0; _header < 8; _header++)
```
lordhoto: Not sure why `this->_header` is used as an index here. Is it to conveniently reset `_header` for the use a few lines below?

engines/lab/anim.h
wjp: This `_planes` array can probably just be removed. (See my comments in anim.cpp about it.)

engines/lab/anim.h
```diff
@@ -0,0 +1,105 @@
...
+};
+
+class Anim {
+private:
+	LabEngine *_vm;
+
+	uint32 _header;
```
lordhoto: This variable seems to be used in a lot of places for a lot of different purposes. I suggest to split this and try to localize it as much as possible (and give it more meaningful names). I somewhat tried to gave up on trying to understand what's going with `_header`...

engines/lab/anim.h
```diff
@@ -0,0 +1,105 @@
...
+
+/*
+ * This code is based on Labyrinth of Time code with assistance of
+ *
+ * Copyright (c) 1993 Terra Nova Development
+ * Copyright (c) 2004 The Wyrmkeep Entertainment Co.
+ *
+ */
+
+#ifndef LAB_DIFF_H
```
lordhoto: An oversight from file renaming: This should be `LAB_ANIM_H`.

engines/lab/anim.h
```diff
@@ -0,0 +1,105 @@
...
+	uint16 _frameNum;
+	bool _playOnce;
+	Common::File *_diffFile;
+	uint32 _diffFileStart;
+	uint32 _size;
+	bool _stopPlayingEnd;
+	uint16 _sampleSpeed;
+	uint32 _diffWidth;
+	uint32 _diffHeight;
+	BitMap *DrawBitMap;
```
lordhoto: Naming doesn't conform to our guidelines. (Also mentioned somewhere else on use).

engines/lab/anim.h
lordhoto: I think the local version still needs to figure out the right size (and related implications). :-)

wjp:
I went through most of the code, and overall things look pretty good.

Just the graphics code seems much more complicated than it needs to be, with various buffers and bitmaps and not-quite-planes being mixed and interchanged in confusing ways. But that can be cleaned up I'm sure.

I'm just waiting to hear if the game is still completable.

wjp:
(Oh, and there's still the event loop, which makes the whole interface painfully sluggish for me.)

sev-:
@wjp, @lordhoto: Are there any objections for this work being continued in-tree? We're trying to follow our established rules of waiting, at least, a week for objections, but it seems that we're falling into the usual review process that could always be performed.

The engine is not-yet-ready for the full-blown testing, but as I remember, this was never a requirement for a merge.

wjp:
The game being completable is a requirement.

wjp:
... but, assuming for the moment that it is indeed completable, no, I have no objections. I would like to do that history fix-up before merging, though.

sev-:
@wjp could you please refer me to any prior discussion that game must be completable for its engine being merged in? Even IRC logs would suffice.

sev-:
For the reference: http://wiki.scummvm.org/index.php/HOWTO-Engine_Inclusion#What_needs_to_be_done_before_you_can_ask_for_your_engine_to_be_merged_into_ScummVM.3F

wjp:
Oh, I always go by the list on the wiki at http://wiki.scummvm.org/index.php/HOWTO-Engine_Inclusion:
- The game supported by your engine needs to be completable on at least i386/amd64

sev-:
Weird. I somehow missed that.

Thus, Fullpipe, Prince, Wintermute, LastExpress, Sword25, Tsage and bunch of other engines must went against this rule, moreover, some of them are not completable even now. Should they be removed?

The only completable engines which were included as I remember, were Tony (by legal reasons) and JMP (because @clone2727 works alone).

I need to go back to the discussions, to refresh my memory.

But even if this is the case, the game was completable, it is just some of the recent refactoring which could introduce bugs.

wjp:
That's a good point. It is a bit strange. In any case, I'd still feel a lot more comfortable knowing there are no other unexpected issues from the heavy refactoring the past week.

lordhoto:
> Thus, Fullpipe, Prince, Wintermute, LastExpress, Sword25, Tsage and bunch of other engines must went against this rule, moreover, some of them are not completable even now. Should they be removed?

LastExpress, Sword25, and tSage (PR#11) were merged before we established these guidelines.
Wintermute had completable games on merge according to PR #267.
Prince also had its Polish version completable according to PR #516.
That makes, as far as I can see the situation, FullPipe the only exception for the completable bit here. About it being in a state to be easily included in one of the next releases is somewhat a future promise, which we can't verify at merge time. And one of the reasons why merge time is a perfect opportunity to do a review pass to get some idea whether this is achievable or not. Then again, we might have made mistakes here in the past. But nobody is perfect. On the other hand, and much more importantly, we can learn from these mistakes and do better the next time.

I wouldn't call to remove engines which are not completable because (I think) we are usually keep the status quo for pre-dating decisions.

> @wjp, @lordhoto: Are there any objections for this work being continued in-tree? We're trying to follow our established rules of waiting, at least, a week for objections, but it seems that we're falling into the usual review process that could always be performed.

I can't give an informed statement about whether I have objections or not. I am still looking through the code, the discussion, and also talked a bit with wjp about things which seemed confusing to me. I will try to give some statement on my view before Christmas Eve.

At any rate, IMHO to make a informed decision (a small) review pass is necessary. This is why we allow us some time before rushing to a merge (again IMHO).

sev-:
@LordHoto Those rules were intended to make sure we do not have abandoned engines which we have to exclude.

What is going on now is that we're just killing the initiative with self-imposed bureucracy. This is against the spirit of these rules.

I see nothing could change before Christmas. The only thing you can do is to perform a quality code review and provide your suggestions on how certain things could be changed. But we know well that the devs involved in the engine have a solid track of delivery, and they actually were so eager to continue working on the engine, so it even became an impediment to the review. So, in any case your valuable input will be taken into account and any sensible suggestions applied. Why not in-tree?

wjp:
I beg to differ. The last time we had this discussion (for Access), there was no continued review at all post-merge (but please correct me if I'm wrong). So why not wait until that's done?

I don't see what initiative is being killed. As you say, development is continuing just fine.

sev-:
@wjp, OK, no problem and no immediate hurry. But I hope it will not hang unmerged for another week.

engines/lab/detection.cpp
```diff
@@ -0,0 +1,252 @@
...
+		_maxScanDepth = 4;
+		_directoryGlobs = directoryGlobs;
+		_flags = kADFlagUseExtraAsHint;
+	}
+
+	virtual const char *getName() const {
+		return "Lab";
+	}
+
+	virtual const char *getOriginalCopyright() const {
```
lordhoto: Should we also have a copyright for the initial release here, i.e. mention Terra Nova Development?

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1022 @@
...
+
+	writeColorRegs(vgaPal, 0, 16);
+	_vm->waitTOF();
+}
+
+void DisplayMan::writeColorRegs(byte *buf, uint16 first, uint16 numReg) {
+	byte tmp[256 * 3];
+
+	for (int i = 0; i < 256 * 3; i++)
+		tmp[i] = buf[i] * 4;
```
lordhoto: I would strongly suggest to adjust the VGA component values the following way: `tmp[i] = (buf[i] << 2) | (buf[i] >> 4);` That gives more accurate color results, i.e. white will actually be white etc.

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1022 @@
...
+
+void DisplayMan::setAmigaPal(uint16 *pal, uint16 numColors) {
+	byte vgaPal[16 * 3];
+	uint16 vgaIdx = 0;
+
+	if (numColors > 16)
+		numColors = 16;
+
+	for (uint16 i = 0; i < numColors; i++) {
+		vgaPal[vgaIdx++] = (byte)(((pal[i] & 0xf00) >> 8) << 2);
```
lordhoto: I also suggest to properly fill the low bits here to provide more accurate color results.

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1022 @@
...
+		vgaPal[vgaIdx++] = (byte)(((pal[i] & 0x0f0) >> 4) << 2);
+		vgaPal[vgaIdx++] = (byte)(((pal[i] & 0x00f)) << 2);
+	}
+
+	writeColorRegs(vgaPal, 0, 16);
```
lordhoto: For safety 16 should probably be replaced by `numColors` to match the number of colors actually setup. This might not be an issue because `setAmigaPal` is always called with 16 right now though.

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1022 @@
...
+	int h = y2 - y1 + 1;
+
+	if (x1 + w > _screenWidth)
+		w = _screenWidth - x1;
+
+	if (y1 + h > _screenHeight)
+		h = _screenHeight - y1;
+
+	if ((w > 0) && (h > 0)) {
+		char *d = (char *)getCurrentDrawingBuffer() + y1 * _screenWidth + x1;
```
lordhoto: I would suggest byte over char here to match the idea that the colors should be in [0, 255].

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1022 @@
...
+}
+
+byte *DisplayMan::getCurrentDrawingBuffer() {
+	if (_currentDisplayBuffer)
+		return _currentDisplayBuffer;
+
+	return _displayBuffer;
+}
+
+void DisplayMan::overlayRect(uint16 penColor, uint16 x1, uint16 y1, uint16 x2, uint16 y2) {
```
lordhoto: My understanding is that this tries to draw all filled pieces of a checkerboard pattern (no promises on that though). I think we should either adapt the function name or make the code more obvious as to what it is doing.

engines/lab/engine.cpp
```diff
@@ -0,0 +1,1190 @@
...
+	kButtonSaveLoad,
+	kButtonUseItem,
+	kButtonLookAtItem,
+	kButtonPrevItem,
+	kButtonNextItem,
+	kButtonBreadCrumbs,
+	kButtonFollowCrumbs
+};
+
+static char initcolors[] = { '\x00', '\x00', '\x00', '\x30',
```
lordhoto: Should be "initColors", I think. One might also consider make the base type `byte` and to use 0x## to make it more clear that this is no character data.

engines/lab/engine.cpp
```diff
@@ -0,0 +1,1190 @@
...
+				_graphics->drawMessage(_inventory[curInv]._name.c_str(), false);
+		}
+	} else
+		drawDirection(closePtr);
+
+	_lastTooLong = _graphics->_lastMessageLong;
+}
+
+void LabEngine::freeScreens() {
+	for (uint16 i = 0; i < 20; i++) {
```
lordhoto: In general I would suggest to use "int" for everything which does not need to have its range bounded specifically.

engines/lab/interface.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+uint16 EventManager::makeButtonKeyEquiv(uint16 key) {
+	if (Common::isAlnum(key))
+		key = tolower(key);
+
+	return key;
+}
+
+Button *EventManager::checkNumButtonHit(ButtonList *buttonList, uint16 key) {
+	uint16 gkey = key - '0';
```
lordhoto: `key` seems to come from a `Common::KeyCode` (inside `EventManager::getMsg`). Doing arithmetic like it is an ASCII character is something you should avoid at any cost.
Overall, I suggest to use proper types to avoid these issues from coming up. Here's a short list of suggestions:
1) Make `createButton` take a `Common::KeyCode` for `key`, adjust `Button`'s `_keyEquiv` to be a `Common::KeyCode` too, and finally make all callers of `createButton` pass key codes instead of ASCII values or similar (which is the case right now).
2) I think this allows to get rid of `makeButtonKeyEquiv` too. Right now it treats its input as ASCII even though it's always a `Common::KeyCode`. Thus, the `tolower` etc. is unnecessary because key codes work different from ASCII codes.
3) Make this function take a `Common::KeyCode` for `key` and adapt it properly, i.e. the `gkey` type etc. too.
4) Adapt all code using `VKEY_UPARROW` etc. to use `Common::KeyCode` values too.
5) Make `IntuiMessage::_code` a `Common::KeyCode` too to reflect that it stores values of that type. This might not be trivial because it seems there is an inflow from a `Button`'s `_buttonId` too. Thus, some more tickeling down or other efforts to increase (type) safety might be required.

engines/lab/intro.h
```diff
@@ -0,0 +1,69 @@
...
+	/**
+	 * Goes through, and responds to all the intuition messages currently in the
+	 * message queue.
+	 */
+	void introEatMessages();
+
+	/**
+	 * Reads in a picture.
+	 */
+	void doPictText(const Common::String Filename, TextFont *msgFont, bool isScreen);
```
lordhoto: Random spot to mention it: naming guidelines, we use camelCase not CamelCase. Applies a few lines further below too.

engines/lab/lab.cpp
```diff
@@ -0,0 +1,252 @@
...
+void LabEngine::drawStaticMessage(byte index) {
+	_graphics->drawMessage(_resource->getStaticText((StaticText)index), false);
+}
+
+void LabEngine::changeVolume(int delta) {
+	warning("STUB: changeVolume()");
+}
+
+void LabEngine::waitTOF() {
+	g_system->copyRectToScreen(_graphics->_displayBuffer, _graphics->_screenWidth, 0, 0, _graphics->_screenWidth, _graphics->_screenHeight);
```
lordhoto: `Engine` keeps a `_system` pointer for accessing `OSystem` functionality. Thus, there's no need to use the global object here.

engines/lab/resource.cpp
```diff
@@ -0,0 +1,417 @@
...
+	} while (cond);
+
+	return list;
+}
+
+RuleList *Resource::readRule(Common::File *file) {
+	char c;
+	RuleList *rules = new RuleList();
+
+	do {
```
lordhoto: I think it would be simpler to write this with a single `while` loop without the need for any `if`:
```
while ((c = file->readSByte()) == 1) {
//  [...]
}
```

engines/lab/resource.cpp
```diff
@@ -0,0 +1,417 @@
...
+	ruleList = nullptr;
+}
+
+Action *Resource::readAction(Common::File *file) {
+	char c;
+	Action *action = nullptr;
+	Action *prev = nullptr;
+	Action *head = nullptr;
+
+	do {
```
lordhoto: Can be simplified as outlined above too.

engines/lab/utils.cpp
```diff
@@ -0,0 +1,272 @@
...
+		else
+			error("Unexpected compression scheme %d", bufType);
+	}
+}
+
+void Utils::setBytesPerRow(int num) {
+	_dataBytesPerRow = num;
+}
+
+uint16 Utils::getRandom(uint16 max) {
```
lordhoto: If I understand this function correctly `max` isn't actually the maximum value this function shall return but rather the maximum value +1. Otherwise the call to our random functionality seems wrong and the if is completely confusing to me.

lordhoto:
I took a short look now and this looks good to go. However, the code is really confusing at times :-/.

Depending on how much people care about the code here's some random things I noticed:
* Generally I think there's a lot of fixed values which could either be replaced by named constants or ARRAYSIZE (in the case of loops setting values in arrays).
* Some integer constants have long size specifiers. This might be a left-over from old days from the original source.
* There seems to be a lot of uses of fixed type values for no appareant reason. Generally, I would try to use int/uint as much as possible. On arithmetic the smaller types get promoted to those anyway. And generally the compiler and the processor are more happy about int/uint.
* The graphics code seems to use `char` in a few places. I would personally use `byte` or `uint8` here, because usually pixels/color components are in the value range 0-255. And sign of char depends on the compiler.
* Since memory management seems to have been an issue at points, it might make sense to have structs keeping memory have constructors/destructors (and maybe prohibit assignments).

engines/lab/dispman.cpp
```diff
@@ -0,0 +1,1022 @@
...
+}
+
+int DisplayMan::flowTextToMem(Image *destIm, TextFont *font, int16 spacing, byte penColor,
+			byte backPen, bool fillBack, bool centerh, bool centerv, bool output, Common::Rect textRect,
+			const char *str) {
+	byte *saveDisplayBuffer = _currentDisplayBuffer;
+	uint32 bytesPerPage = _screenBytesPerPage;
+
+	_currentDisplayBuffer = destIm->_imageData;
+	_screenBytesPerPage = (uint32)destIm->_width * (int32)destIm->_height;
```
lordhoto: This just looks odd to cast the operands to different types. In general I saw a lot of arithmetics using casts, making me wonder about the types used in general. So, if someone has too much time on their hands, one could try to clean that up a lot.

engines/lab/detection.cpp
bluegr: Done

engines/lab/dispman.cpp
bluegr: Fixed

sev-:
@wjp Did you have a chance to fix the history? This is currently the only thing which prevents it from the merge.

wjp:
I can fix it up and merge it manually right now.

sev-:
OK

wjp:
Merged manually, with history fix-ups.

engines/lab/dispman.cpp
Strangerke: It can't be removed, it's in fact some missing code in diffNextFrame()... So it's yet to be fixed

engines/lab/anim.cpp
Strangerke: It's the place where rawBitmap has to be set properly ;)

PR #637, opened 2015-12-15
PS3: Update to SDL2

bgK:
The PlayStation 3 port currently uses an old development version of SDL2 with an added PS3 driver. At that point of development, SDL2 still offered a compatibility layer with the SDL 1.2 API.

Thanks to the recent addition of native SDL2 support in ScummVM, it has been possible to update the PS3 port to a newer SDL version based on SDL 2.0.3, removing the SDL 1.2 compatibility layer.

The new SDL version is maintained at: https://bitbucket.org/bgK/sdl_psl1ght

This PR contains:
* The update of the PS3 port to use SDL2
* The removal of the old SDL 1.3 bits which were only used by the PS3 port
* A fix to ensure the sound sample format provided by SDL is the same as the one expected by ScummVM

What's left to be done before this PR can be merged:
- [x] Install the new SDL2 library on the PS3 toolchain on BuildBot. Help is required here, since I don't have access to the server.


digitall:
With some help from @bgK , I have installed the required SDL2 library in the /opt/toolchains/ps3/portlibs alongside the older SDL1.2 library. This was installed from https://bitbucket.org/bgK/sdl_psl1ght/src/ee51abe8b917b351f765f76b3df764c29dda173a?at=psl1ght-2.0.3 at that version and required use of my ps3-cross-compile script to set the environment correctly, before using the script.sh in that tree to do the configure step, then "make", "make install" as root with correct ps3-cross-compile environment sourced was required. Just a note for future reference.

PR #638, opened 2015-12-20
Improve compiler version checks

wjp:
This PR does two related things:

It disables gcc mode, even if ICC is set to gcc-compatibility mode, since it doesn't accept all of the gcc command line options that we use.

It also adds custom version checking code for clang, since previously we output the version of gcc that clang is pretending to be.

I've tested this only on Linux, with gcc 4.8.3, gcc 4.9.3, clang 3.4.2, clang 3.7.0, icc 11.1, icc 15.0.2.

### DIFF
--- a/configure
+++ b/configure
@@ -1686,33 +1686,39 @@ LD=$CXX
 #
 echocheck "compiler version"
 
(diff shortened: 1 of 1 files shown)


lordhoto:
Looks nice. :+1:

PR #639, opened 2015-12-21
GRAPHICS: TrueType font creation from character height

bgK:
Loading a TrueType font currenly requires specifying the point size. The point size not always a very handy metric. This pull request adds an extra parameter "sizeMode" to the font loading method allowing to give the size as the desired height in pixels.

The matching between point sizes and heights is done according to the metrics tables embedded in fonts. The matching algorithm is the same as the one used in Wine (though no code used here comes directly from Wine).

ZVision is changed to use the new size mode, removing a minor hack.
This code is likely to be used by @clone2727's Buried engine. It will also be used in some other engine soon to land in ResidualVM.

### DIFF
--- a/engines/wintermute/base/font/base_font_truetype.cpp
+++ b/engines/wintermute/base/font/base_font_truetype.cpp
@@ -581,7 +581,7 @@ bool BaseFontTT::initFont() {
 	}
 
 	if (file) {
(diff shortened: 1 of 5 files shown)


lordhoto:
Looks fine except some minor formatting issues. I'll merge this when those are fixed.

bgK:
The formatting issues should be resolved now.

lordhoto:
Great! Thanks.

PR #640, opened 2015-12-24
I18N: Czech localization update

Tsbook:
This change contains the latest updates for localization. Also incorporated slight modification to the README file.


criezy:
Merged. Thank you.

PR #642, opened 2015-12-31
Add BS 2.5 to German NEWS file

rootfather:
This request announces the support for BS 2.5 to the German NEWS file. I also updated the .gitignore to ignore all .opendb files created by Visual Studio 2015 during compilation.


criezy:
I cherry-pick the last commit only (db168b4).
I have no idea if the first commit is useful since I don't have Windows. The best is to open a separate pull request for it if it should be merged and somebody who knows better than I do will take care of it.
More generally, please use feature branches and separate pull request for unrelated commits.

PR #645, opened 2016-01-04
CREATE_PROJECT: Make Xcode generator work out-of-the-box again (for Mac OS)

csnover:
This PR gets the Xcode generator for Mac OS working without errors again. I also made some progress on Xcode generator for iOS but it’s still quite busted. I could use help from anyone familiar with the iphone make build to know what files and flags need to be provided for the iOS target, if anyone cares enough to help with this. It’s not a high priority for me, I just needed the Mac OS one working.

### DIFF
--- a/devtools/create_project/xcode.cpp
+++ b/devtools/create_project/xcode.cpp
@@ -35,7 +35,6 @@ namespace CreateProjectTool {
 #ifdef ENABLE_IOS
 #define IOS_TARGET 0
 #define OSX_TARGET 1
-#define SIM_TARGET 2
 #else
 #define OSX_TARGET 0
 #endif
@@ -259,9 +258,6 @@ void XcodeProvider::createWorkspace(const BuildSetup &setup) {
 	_targets.push_back(PROJECT_DESCRIPTION "-iPhone");
 #endif
 	_targets.push_back(PROJECT_DESCRIPTION "-OS X");
-#ifdef ENABLE_IOS
-	_targets.push_back(PROJECT_DESCRIPTION "-Simulator");
-#endif
 	setupCopyFilesBuildPhase();
 	setupFrameworksBuildPhase();
 	setupNativeTarget();
@@ -500,42 +496,6 @@ void XcodeProvider::setupFrameworksBuildPhase() {
 	framework_OSX->_properties["files"] = osx_files;
 
 	_frameworksBuildPhase.add(framework_OSX);
-#ifdef ENABLE_IOS
-	//////////////////////////////////////////////////////////////////////////
-	// Simulator
-	Object *framework_simulator = new Object(this, "PBXFrameworksBuildPhase_" + _targets[SIM_TARGET], "PBXFrameworksBuildPhase", "PBXFrameworksBuildPhase", "", "Frameworks");
-
-	framework_simulator->addProperty("buildActionMask", "2147483647", "", kSettingsNoValue);
-	framework_simulator->addProperty("runOnlyForDeploymentPostprocessing", "0", "", kSettingsNoValue);
-
-	// List of frameworks
-	Property simulator_files;
-	simulator_files._hasOrder = true;
-	simulator_files._flags = kSettingsAsList;
-
-	ValueList frameworks_simulator;
-	frameworks_simulator.push_back("CoreAudio.framework");
-	frameworks_simulator.push_back("CoreFoundation.framework");
-	frameworks_simulator.push_back("Foundation.framework");
-	frameworks_simulator.push_back("UIKit.framework");
-	frameworks_simulator.push_back("AudioToolbox.framework");
-	frameworks_simulator.push_back("QuartzCore.framework");
-	frameworks_simulator.push_back("OpenGLES.framework");
-
-	order = 0;
-	for (ValueList::iterator framework = frameworks_simulator.begin(); framework != frameworks_simulator.end(); framework++) {
-		std::string id = "Frameworks_" + *framework + "_simulator";
-		std::string comment = *framework + " in Frameworks";
-
-		ADD_SETTING_ORDER_NOVALUE(simulator_files, getHash(id), comment, order++);
-		ADD_BUILD_FILE(id, *framework, getHash(*framework), comment);
-		addFileReference(*framework, *framework, properties[*framework]);
-	}
-
-	framework_simulator->_properties["files"] = simulator_files;
-
-	_frameworksBuildPhase.add(framework_simulator);
-#endif
 }
 
 void XcodeProvider::setupNativeTarget() {
@@ -610,9 +570,6 @@ void XcodeProvider::setupProject() {
 	targets._settings[getHash("PBXNativeTarget_" + _targets[IOS_TARGET])] = Setting("", _targets[IOS_TARGET], kSettingsNoValue, 0, 0);
 #endif
 	targets._settings[getHash("PBXNativeTarget_" + _targets[OSX_TARGET])] = Setting("", _targets[OSX_TARGET], kSettingsNoValue, 0, 1);
-#ifdef ENABLE_IOS
-	targets._settings[getHash("PBXNativeTarget_" + _targets[SIM_TARGET])] = Setting("", _targets[SIM_TARGET], kSettingsNoValue, 0, 2);
-#endif
 	project->_properties["targets"] = targets;
 #ifndef ENABLE_IOS
 	// Force list even when there is only a single target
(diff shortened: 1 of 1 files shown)


lordhoto:
PR #630 implements proper support for iOS targets AFAIK. So, there might be some duplicate effort here.

devtools/create_project/xcode.cpp
```diff
@@ -69,6 +69,31 @@ namespace CreateProjectTool {
...
+}
+
+const char * const DIRECTORIES[] = {
+	/* Homebrew */ "/usr/local/",
+	/* MacPorts */ "/opt/local/",
+	/* Fink     */ "/sw/"
+};
+
+std::string findPath(const char * const lib) {
+	for (int i = 0; i < 3; ++i) {
```
lordhoto: We strongly prefer ARRAYSIZE instead of magic constants for this use case.

devtools/create_project/xcode.cpp
```diff
@@ -69,6 +69,31 @@ namespace CreateProjectTool {
 	_buildFile._flags = kSettingsSingleItem; \
 }
 
+/**
+ * Checks whether a file or directory exists.
+ */
+bool exists(const char * const fileName) {
```
lordhoto: We use `const char *const` for this type. Applies to other places too.

devtools/create_project/xcode.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <fstream>
 #include <algorithm>
+#include <sys/stat.h>
```
lordhoto: I'm pretty sure MSVC doesn't feature these POSIX interfaces. Our create_project tool is aimed to be buildable under all platforms we support.

That said, there is probably a bigger issue: We usually generate project files for release sources beforehand. AFAICT this Xcode output depends on the location of static libraries during project file generation. Is there any way we can avoid this and have it pick up the locations when the user opens the project files?

devtools/create_project/xcode.cpp
csnover: You are OK with just `#include "../../common/util.h"` to get this macro?

devtools/create_project/xcode.cpp
csnover: I am by no means an expert on Xcode project files, but I don’t think so? My understanding is Xcode project files are just plist files and have no capability for specifying events like “when this project file is opened, run this command/code/whatever”. The closest thing would be an external build target set up as a prerequisite for the actual build target (I think CMake does this), but I am not sure if is safe to modify the project build file in the middle of a build (I don’t think Xcode will pick up the changes until the next build), so it wouldn’t provide a very good user experience. Detecting here doesn’t seem any different than running autoconf to generate Makefiles with the right paths though?

Some other options I can think of are:

1. Unconditionally specify all the directories in search paths. This will cause warning to be emitted at build time for the ones that don’t exist (which is why I didn’t just do that to begin with). The existing code was also using an explicit path for the static libraries, I’m not sure offhand if a relative filename + libraries search path is good enough? (I would think so but I really was more concerned about just fixing the defines-issue than making this generator excellent.)
2. Add `--prefix` flag to the generator that will override the path search if it is specified.

csnover:
> PR #630 implements proper support for iOS targets AFAIK. So, there might be some duplicate effort here.

Yes, seems so, though not a ton—that PR includes lots of extra hard work and effort to properly build to iOS; such effort is not represented here. :O

devtools/create_project/xcode.cpp
csnover: I guess I also don’t know how to get cross-platform code for finding a directory into create_project. I tried to include common/fs.cpp but it ends up requiring dependencies of irrelevant things like GUI::EventRecorder because it uses flags for the entire build. I don’t know about how to make make do what I want it to do, and I don’t even know if this approach would be acceptable anyway.

devtools/create_project/xcode.cpp
lordhoto: Initially I wanted to disadvise usage of code from common etc. in tools, but it seems other tools do that too (*sigh*). You can either do `#include common/util.h` like other devtools. But it's a bit tricky because we don't actually link against any functionality from ScummVM itself. Thus, this works just because there's no dependencies. Or you just `#define` your own ARRAYSIZE like some other devtools like to do.

devtools/create_project/xcode.cpp
lordhoto: > 1. Unconditionally specify all the directories in search paths.

Does this even work? It looks like it's for static library linking, which takes a full path. I doubt we can simply supply every library and the non-existent ones only cause warnings.

> 2. Add --prefix flag to the generator that will override the path search if it is specified.

Might work, still doesn't help when we pre-generate project files AFAICT.

> I guess I also don’t know how to get cross-platform code for finding a directory into create_project. I tried to include common/fs.cpp but it ends up requiring dependencies of irrelevant things like GUI::EventRecorder because it uses flags for the entire build. I don’t know about how to make make do what I want it to do, and I don’t even know if this approach would be acceptable anyway

Yeah, you will need to implement it yourself in devtools/ code. ScummVM main code cannot be linked to devtools binaries.

lordhoto:
> Yes, seems so, though not a ton—that PR includes lots of extra hard work and effort to properly build to iOS; such effort is not represented here. :O

I just wanted to point it out before you spent a lot of time on this and in the end it might just get replaced soonish anyway.

devtools/create_project/xcode.cpp
csnover: > Does this even work? It looks like it's for static library linking, which takes a full path. I doubt we can simply supply every library and the non-existent ones only cause warnings.

It would work for the paths which are used for `-I` and `-L`. However, the linker will emit warnings for non-existing directories passed to `-L`, which is annoying. I was trying to avoid these.

Library linking for ScummVM OS X uses “normal” `-l` via `OTHER_LDFLAGS` (`scummvmOSX_LdFlags`) instead of putting them into Xcode’s special Linked Frameworks and Libraries section with `DEF_LOCALLIB_STATIC`. The `DEF_LOCALLIB_STATIC` is used only for iOS, and iOS is busted anyway, so, uh. …

> Might work, still doesn't help when we pre-generate project files AFAICT.

Can you tell me more about what the scenario is where these files are pre-generated to be given to other users? To me, at the moment, this feels like asking for a way to give people pre-generated Makefiles without running autoconf on the build machine, which I don’t think is possible either (unless the two machines are the same, of course)?

> Yeah, you will need to implement it yourself in devtools/ code. ScummVM main code cannot be linked to devtools binaries.

So if the runtime directory detection were maintained, I would need to implement a Win32 branch? Are there other platforms that need to be considered?

Practically, and perhaps naïvely, I would think just wrapping in `#ifndef _WIN32` would be good enough; I have no idea why anyone would reasonably try to generate Xcode projects on Windows (and the only practical reason why I could think of someone doing it on a platform other than Mac OS would be for unit testing the Xcode generator). Is this unreasonable?

devtools/create_project/xcode.cpp
csnover: While I don’t like the code duplication, I will just add it to this file since it is a one-liner.

csnover:
> I just wanted to point it out before you spent a lot of time on this and in the end it might just get replaced soonish anyway.

Understood. Since this is primarily for my own education I am happy to do some work here as long as the contribution is felt to be worthwhile by everyone else.

devtools/create_project/xcode.cpp
csnover: Sorry about that. Habits!

csnover:
I updated the PR with:

1. Remove whitespace in `const char * const`
2. Added a branch `#ifdef _WIN32` to avoid POSIX APIs on that platform (this will end up assuming that the project file is being built for a Mac OS user that puts their libs in `/usr/local`, which is reasonable enough for someone generating Xcode projects on Windows…)
3. Add `ARRAYSIZE` and use it instead of magic

There is still the outstanding discussion about pre-generation, but other than that I believe all feedback is addressed and this could be landed as-is.

lordhoto:
PR #630 landed now. You'll need to check whether/if your fixes are still applicable and update the PR accordingly.

Our use case for pre-made project file generation has historically been:
We pre-generate project files to be included in source tarballs for release versions. These are setup in a way that only supported engines are enabled etc., thus matching the feature set you get by official release binaries.
For MSCV we always had pre-built libraries which you just pop in to compile using them. A good idea would be to have the same of Mac OS X to be used with the XCode project files. Supporting the various third-party package managers for OS X seems rather unrealistic for pre-generated project files. But supporting them for use while development is a good idea.

csnover:
I’ll rebase against master and let you know when this is ready for another review. I think that #630 took care of the most significant issues preventing any successful use of Xcode, but the `/opt` hard-coding and several other things in the list from c5dc6822930e1d2079a12f1c5e8e201f089d4bd9 are still there.

It seems like the simplest solution for pre-generated project files would be a command-line flag and then hardcode the paths to "lib" which will be `scummvm/lib/lib` for the libraries and `scummvm/lib/include` for the headers I guess. But doesn’t the official Mac OS build just use autoconf/make like the rest?

lordhoto:
> It seems like the simplest solution for pre-generated project files would be a command-line flag and then hardcode the paths to "lib" which will be `scummvm/lib/lib` for the libraries and `scummvm/lib/include` for the headers I guess. But doesn’t the official Mac OS build just use autoconf/make like the rest?

I think hardcoding the path to something relative to the project files would be best, i.e. that you can easily drop the libraries next to the project files.

Our official Mac OS X build uses our configure/make build system. That's actually the same for our official Win32 build. Also, we don't use autoconf.

sev-:
What is the status of it? Now it doesn't apply cleanly anymore.

csnover:
The iOS-fixing PR did made some of the same changes as this PR, so that’s why there is a conflict now. This PR still contains some additional changes, but I’ve been more focused on SCI32 development since I have a working Xcode environment now so haven’t updated it. If you’d like, it can be closed and I can open a new one later if/when I loop back to this again.

sev-:
I just was hoping to squeeze it into 1.8.0...

sev-:
Any update?

csnover:
It’s not ready and I’m not working on this right now. I will reopen this later when it is ready.

PR #646, opened 2016-01-04
DEVTOOLS: Ignore *.opendb files generated by VS 2015

rootfather:
This commit adds all temporary .opendb files to the .gitignore file.

.opendb files are generated by Visual Studio 2015 during compilation. They are hidden in the file system and seem to be temporary.


wjp:
Thanks

PR #647, opened 2016-01-07
IOS: Updates the splash screens

bSr43:
This updates the iOS media files, in order to match the official media.


lordhoto:
Looks great.

Can you also make a pull request to [scummvm-media](https://github.com/scummvm/scummvm-media) with an SVG which is used to generate these files? Please make sure that the versions in this PR are what you get when using the SVG file. If you need to update the PNG files afterwards, I suggest to replace this commit. Having all these big binary blobs in git isn't too nice :-).

bSr43:
You are right about the binary blobs :)
I'm feeling stupid, because I have thrown the SVG I used to generate the PNG. That being said, I wrote a new little Bash script which generates the SVG, and then convert them using ImageMagick. Would you be interested in this script as well? :)

lordhoto:
> Would you be interested in this script as well? :)

Sure, that sounds great to have in scummvm-media :-).

lordhoto:
Are these files generated from scummvm-media now? If so, I'll merge this.

bSr43:
Yes, they are!

lordhoto:
Great thanks!

PR #648, opened 2016-01-10
SCI: Fixes compilation when SCI32 is disabled

bSr43:
This fixes a compilation issue introduced by the recent refactoring of the SCI engine.


wjp:
Oops, and thanks! I fixed it slightly differently in 22796e559fd252fc77082200f2d32fec0df3ce3b to be more consistent with the other variables like that.

PR #649, opened 2016-01-12
SCI32: kPalVary, kPalette fade support, kPalette findColor support, improvements to game time management

csnover:
I verified this does not break builds without `ENABLE_SCI32` this time! :), and tested against most of the rooms in SQ6 that call kPalVary.

Fades of the Sierra logo and SQ6 title screen are working correctly, but fades between rooms requires at least an implementation of `kSetShowStyle`, so I will probably do that next. kPalVary also appears to be working (in most SQ6 rooms it is used to lighten/darken Roger when he walks near light sources), but there are some lingering palette colour issues; based on a brief conversation with @lskovlun it sounds like Remap probably needs to be implemented for those to be fixed, but it is possible there are still a couple bugs in kPalVary. SQ6 only uses subops 0 (2, 3, and 5 arguments), 1 (1 and 2 arguments), and 3 so it would be great if someone could look through scripts for some other SCI2/3 games to see where they use the other subops.

This PR also includes SCI32 implementation of kPalette findColor from disassembly, and two improvements to game time management:

1. Anywhere that was trying to calculate a tick time goes through `g_sci->getTickCount()` now;
2. Save games now store game time in ticks, in order to avoid time moving backwards for parts of the engine that rely on it never doing that (like kPalVary and kPalCycle).

### DIFF
--- a/engines/sci/engine/kernel_tables.h
+++ b/engines/sci/engine/kernel_tables.h
@@ -697,7 +697,6 @@ static SciKernelMapEntry s_kernelMap[] = {
 	//   (inclusive) are set to 0
 	{ MAP_CALL(SetPalStyleRange),   SIG_EVERYWHERE,           "ii",                   NULL,            NULL },
 
-	// MorphOn - used by SQ6, script 900, the datacorder reprogramming puzzle (from room 270)
 	{ MAP_CALL(MorphOn),            SIG_EVERYWHERE,           "",                     NULL,            NULL },
 
 	// SCI3 Kernel Functions
--- a/engines/sci/engine/kernel_tables.h
+++ b/engines/sci/engine/kernel_tables.h
@@ -695,15 +695,16 @@ static SciKernelMapEntry s_kernelMap[] = {
 	// MovePlaneItems - used by SQ6 to scroll through the inventory via the up/down buttons
 	// SetPalStyleRange - 2 integer parameters, start and end. All styles from start-end
 	//   (inclusive) are set to 0
-	{ MAP_CALL(SetPalStyleRange),  SIG_EVERYWHERE,            "ii",                   NULL,            NULL },
+	{ MAP_CALL(SetPalStyleRange),   SIG_EVERYWHERE,           "ii",                   NULL,            NULL },
 
 	// MorphOn - used by SQ6, script 900, the datacorder reprogramming puzzle (from room 270)
+	{ MAP_CALL(MorphOn),            SIG_EVERYWHERE,           "",                     NULL,            NULL },
 
 	// SCI3 Kernel Functions
-	{ MAP_CALL(PlayDuck),         SIG_EVERYWHERE,           "(.*)",                  NULL,            NULL },
+	{ MAP_CALL(PlayDuck),           SIG_EVERYWHERE,           "(.*)",                 NULL,            NULL },
 #endif
 
-	{ NULL, NULL,                  SIG_EVERYWHERE,           NULL,                    NULL,            NULL }
+	{ NULL, NULL,                   SIG_EVERYWHERE,           NULL,                   NULL,            NULL }
 };
 
 /** Default kernel name table. */
--- a/engines/sci/engine/kernel_tables.h
+++ b/engines/sci/engine/kernel_tables.h
@@ -201,31 +201,42 @@ static const SciKernelMapSubEntry kGraph_subops[] = {
 
 //    version,         subId, function-mapping,                    signature,              workarounds
 static const SciKernelMapSubEntry kPalVary_subops[] = {
-	{ SIG_SINCE_SCI21,     0, MAP_CALL(PalVaryInit),               "ii(i)(i)(i)",          NULL },
-	{ SIG_SCIALL,          0, MAP_CALL(PalVaryInit),               "ii(i)(i)",             NULL },
-	{ SIG_SCIALL,          1, MAP_CALL(PalVaryReverse),            "(i)(i)(i)",            NULL },
-	{ SIG_SCIALL,          2, MAP_CALL(PalVaryGetCurrentStep),     "",                     NULL },
-	{ SIG_SCIALL,          3, MAP_CALL(PalVaryDeinit),             "",                     NULL },
-	{ SIG_SCIALL,          4, MAP_CALL(PalVaryChangeTarget),       "i",                    NULL },
-	{ SIG_SCIALL,          5, MAP_CALL(PalVaryChangeTicks),        "i",                    NULL },
-	{ SIG_SCIALL,          6, MAP_CALL(PalVaryPauseResume),        "i",                    NULL },
+	{ SIG_SCI16,           0, MAP_CALL(PalVaryInit),               "ii(i)(i)",             NULL },
+	{ SIG_SCI16,           1, MAP_CALL(PalVaryReverse),            "(i)(i)(i)",            NULL },
+	{ SIG_SCI16,           2, MAP_CALL(PalVaryGetCurrentStep),     "",                     NULL },
+	{ SIG_SCI16,           3, MAP_CALL(PalVaryDeinit),             "",                     NULL },
+	{ SIG_SCI16,           4, MAP_CALL(PalVaryChangeTarget),       "i",                    NULL },
+	{ SIG_SCI16,           5, MAP_CALL(PalVaryChangeTicks),        "i",                    NULL },
+	{ SIG_SCI16,           6, MAP_CALL(PalVaryPauseResume),        "i",                    NULL },
 #ifdef ENABLE_SCI32
-	{ SIG_SCI32,           8, MAP_CALL(PalVaryUnknown),            "i",                    NULL },
-	{ SIG_SCI32,           9, MAP_CALL(PalVaryUnknown2),           "i",                    NULL },
+	{ SIG_SCI32,           0, MAP_CALL(PalVarySetVary),            "i(i)(i)(ii)",          NULL },
+	{ SIG_SCI32,           1, MAP_CALL(PalVarySetPercent),         "(i)(i)",               NULL },
+	{ SIG_SCI32,           2, MAP_CALL(PalVaryGetPercent),         "",                     NULL },
+	{ SIG_SCI32,           3, MAP_CALL(PalVaryOff),                "",                     NULL },
+	{ SIG_SCI32,           4, MAP_CALL(PalVaryMergeTarget),        "i",                    NULL },
+	{ SIG_SCI32,           5, MAP_CALL(PalVarySetTime),            "i",                    NULL },
+	{ SIG_SCI32,           6, MAP_CALL(PalVaryPauseResume),        "i",                    NULL },
+	{ SIG_SCI32,           7, MAP_CALL(PalVarySetTarget),          "i",                    NULL },
+	{ SIG_SCI32,           8, MAP_CALL(PalVarySetStart),           "i",                    NULL },
+	{ SIG_SCI32,           9, MAP_CALL(PalVaryMergeStart),         "i",                    NULL },
 #endif
 	SCI_SUBOPENTRY_TERMINATOR
 };
 
 //    version,         subId, function-mapping,                    signature,              workarounds
 static const SciKernelMapSubEntry kPalette_subops[] = {
-	{ SIG_SCIALL,          1, MAP_CALL(PaletteSetFromResource),    "i(i)",                 NULL },
-	{ SIG_SCIALL,          2, MAP_CALL(PaletteSetFlag),            "iii",                  NULL },
-	{ SIG_SCIALL,          3, MAP_CALL(PaletteUnsetFlag),          "iii",                  kPaletteUnsetFlag_workarounds },
-	{ SIG_SCIALL,          4, MAP_CALL(PaletteSetIntensity),       "iii(i)",               NULL },
-	{ SIG_SCIALL,          5, MAP_CALL(PaletteFindColor),          "iii",                  NULL },
-	{ SIG_SCIALL,          6, MAP_CALL(PaletteAnimate),            "i*",                   NULL },
-	{ SIG_SCIALL,          7, MAP_CALL(PaletteSave),               "",                     NULL },
-	{ SIG_SCIALL,          8, MAP_CALL(PaletteRestore),            "[r0]",                 NULL },
+	{ SIG_SCIALL,         1, MAP_CALL(PaletteSetFromResource),    "i(i)",                 NULL },
+	{ SIG_SCI16,          2, MAP_CALL(PaletteSetFlag),            "iii",                  NULL },
+	{ SIG_SCI16,          3, MAP_CALL(PaletteUnsetFlag),          "iii",                  kPaletteUnsetFlag_workarounds },
+#ifdef ENABLE_SCI32
+	{ SIG_SCI32,          2, MAP_CALL(PaletteSetFade),            "iii",                  NULL },
+	{ SIG_SCI32,          3, MAP_CALL(PaletteFindColor),          "iii",                  NULL },
+#endif
+	{ SIG_SCI16,          4, MAP_CALL(PaletteSetIntensity),       "iii(i)",               NULL },
+	{ SIG_SCI16,          5, MAP_CALL(PaletteFindColor),          "iii",                  NULL },
+	{ SIG_SCI16,          6, MAP_CALL(PaletteAnimate),            "i*",                   NULL },
+	{ SIG_SCI16,          7, MAP_CALL(PaletteSave),               "",                     NULL },
+	{ SIG_SCI16,          8, MAP_CALL(PaletteRestore),            "[r0]",                 NULL },
 	SCI_SUBOPENTRY_TERMINATOR
 };
 
--- a/engines/sci/engine/savegame.cpp
+++ b/engines/sci/engine/savegame.cpp
@@ -273,7 +273,11 @@ static void sync_SavegameMetadata(Common::Serializer &s, SavegameMetadata &obj) 
 		if (s.getVersion() >= 26)
 			s.syncAsUint32LE(obj.playTime);
 	} else {
-		obj.playTime = g_engine->getTotalPlayTime() / 1000;
+		if (s.getVersion() >= 34) {
+			obj.playTime = g_sci->getTickCount();
+		} else {
+			obj.playTime = g_engine->getTotalPlayTime() / 1000;
+		}
 		s.syncAsUint32LE(obj.playTime);
 	}
 }
@@ -1031,7 +1035,11 @@ void gamestate_restore(EngineState *s, Common::SeekableReadStream *fh) {
 	// Time state:
 	s->lastWaitTime = g_system->getMillis();
 	s->_screenUpdateTime = g_system->getMillis();
-	g_engine->setTotalPlayTime(meta.playTime * 1000);
+	if (meta.version >= 34) {
+		g_sci->setTickCount(meta.playTime);
+	} else {
+		g_engine->setTotalPlayTime(meta.playTime * 1000);
+	}
 
 	if (g_sci->_gfxPorts)
 		g_sci->_gfxPorts->saveLoadWithSerializer(ser);
--- a/engines/sci/engine/savegame.cpp
+++ b/engines/sci/engine/savegame.cpp
@@ -47,6 +47,7 @@
 #include "sci/sound/music.h"
 
 #ifdef ENABLE_SCI32
+#include "sci/graphics/palette32.h"
 #include "sci/graphics/frameout.h"
 #endif
 
@@ -304,6 +305,7 @@ void EngineState::saveLoadWithSerializer(Common::Serializer &s) {
 	_segMan->saveLoadWithSerializer(s);
 
 	g_sci->_soundCmd->syncPlayList(s);
+	// NOTE: This will be GfxPalette32 for SCI32 engine games
 	g_sci->_gfxPalette16->saveLoadWithSerializer(s);
 }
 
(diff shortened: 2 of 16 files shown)


engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+
+void GfxPalette32::updateForFrame() {
+	applyAll();
+	_needsUpdate = false;
+	// _frameOut.remap.remapAllTables(_nextPalette != _sysPalette);
+}
+
+void GfxPalette32::updateHardware() {
+	if (_sysPalette == _nextPalette) {
+		debug("Skipping hardware update because palettes are identical");
```
csnover: This debug message is here because this has never actually happened to me so I don’t know if it is just that literally every single room I have been in changes palettes on every frame (likely, given how animated the backgrounds are in SQ6) or there is some bug preventing the palettes from matching.

engines/sci/graphics/palette32.cpp
```diff
@@ -82,9 +218,262 @@ byte GfxPalette32::matchClutColor(uint16 color) {
...
+		} else {
+			_varyDirection = 0;
+			_varyTargetPercent = _varyPercent = percent;
+		}
+	}
+}
+
+// TODO: This gets called *a lot* in at least the first scene
+// of SQ6. Optimisation would not be the worst idea in the world.
+void GfxPalette32::kernelPalVarySet(const GuiResourceId paletteId, const int16 percent, const int time, const int16 fromColor, const int16 toColor) {
```
csnover: I didn’t instrument the code to check for performance problems, SCI engine does not optimise this (it retrieves and generates the palettes inside of kPalVary), but it would be pretty trivial to avoid the extra resource fetch and palette construction just by tracking the last resource ID (GfxPalette16 code does track this; I was trying to keep the implementation as literal to the SCI engine code as possible at least until it was reviewed once).

engines/sci/graphics/palette32.cpp
```diff
@@ -82,9 +218,262 @@ byte GfxPalette32::matchClutColor(uint16 color) {
...
+
+int16 GfxPalette32::getVaryPercent() const {
+	return _varyPercent < 0 ? -_varyPercent : _varyPercent;
+}
+
+void GfxPalette32::varyOff() {
+	_varyNumTimesPaused = 0;
+	_varyPercent = _varyTargetPercent = 0;
+	_varyFromColor = 0;
+	_varyToColor = 255;
```
csnover: I am a little mystified by this being set to 255 instead of 256 since scripts in SQ6 use 256 as their top value. Possibly a bug in the original engine.

engines/sci/graphics/palette32.cpp
```diff
@@ -223,10 +571,6 @@ int16 GfxPalette32::setCycle(const uint16 fromColor, const uint16 toColor, const
 	cycler->numTimesPaused = 0;
 
 	setCycleMap(fromColor, numColorsToCycle);
-
-	// TODO: Validate that this is the correct return value according
-	// to disassembly
-	return 0;
```
csnover: The correct value was void. :)

engines/sci/graphics/palette32.cpp
csnover: Actually, note to self: double-check this, it could be just the fade code that does 1–256 and all the rest of it does 0–255.

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+
+void GfxPalette32::updateHardware() {
+	if (_sysPalette == _nextPalette) {
+		debug("Skipping hardware update because palettes are identical");
+		return;
+	}
+
+	byte bpal[3 * 256];
+
+	// TODO: Copy assignment operator for Palette??
```
csnover: This can be ignored :) SCI32 does not seem to have any intensity or mapping stuff like I guess existed in earlier engines, so it is just a little inefficient to do the default copy since it brings along these two extra tables. The amount of data is so small in either case it really is unlikely to matter in any practical sense.

engines/sci/engine/kernel_tables.h
```diff
@@ -695,15 +706,16 @@ static SciKernelMapEntry s_kernelMap[] = {
 	// MovePlaneItems - used by SQ6 to scroll through the inventory via the up/down buttons
 	// SetPalStyleRange - 2 integer parameters, start and end. All styles from start-end
 	//   (inclusive) are set to 0
-	{ MAP_CALL(SetPalStyleRange),  SIG_EVERYWHERE,            "ii",                   NULL,            NULL },
+	{ MAP_CALL(SetPalStyleRange),   SIG_EVERYWHERE,           "ii",                   NULL,            NULL },
```
bluegr: Spacing

engines/sci/graphics/palette32.cpp
lskovlun: return abs(_varyPercent);

engines/sci/engine/savegame.cpp
```diff
@@ -721,6 +723,94 @@ void GfxPalette::saveLoadWithSerializer(Common::Serializer &s) {
...
+	saveLoadOptionalPalette32(s, &_varyTargetPalette);
+
+	s.syncAsSint16LE(_varyFromColor);
+	s.syncAsSint16LE(_varyToColor);
+	s.syncAsUint32LE(_varyLastTick);
+	s.syncAsSint32LE(_varyTime);
+	s.syncAsSint16LE(_varyDirection);
+	s.syncAsSint16LE(_varyPercent);
+	s.syncAsSint16LE(_varyTargetPercent);
+	s.syncAsUint16LE(_varyNumTimesPaused);
```
bluegr: Hm. Do we really need to save all of that information in saved games? Usually, palette animations happen during in-game animations, so the users can't save while these are active. Do you have a case where there are active PalVary effects that need to be stored in saved games?

engines/sci/engine/savegame.h
```diff
@@ -37,6 +37,7 @@ struct EngineState;
  *
  * Version - new/changed feature
  * =============================
+ *      34 - SCI32 palettes
```
bluegr: Same question as above

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+	_varyFromColor(0), _varyToColor(255), _varyNumTimesPaused(0),
+	_varyPercent(_varyTargetPercent), _varyLastTick(0),
+	// Palette versioning
+	_version(1), _needsUpdate(false) {
+		// TODO: Sci::Palette should have a default constructor that does
+		// all of this
+		for (int color = 0; color < ARRAYSIZE(_sysPalette.colors); color++) {
+			_sourcePalette.colors[color] = _nextPalette.colors[color] = _sysPalette.colors[color];
+			// TODO: Intensity and mapping are not used by SCI32 but are used by
+			// other parts of the graphic engine right now. Replace those parts.
```
bluegr: I don't understand this comment. Intensity is used by older SCI games, called by kPaletteSetIntensity. Mapping is used by older SCI games, which merged the palette of sprites with the palette of the actual scene. This functionality is not used by SCI32 at all, but it is used by the rest of the SCI engine, so I don't understand why the logic in those parts should be replaced...

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+
+		for (int i = 0; i < ARRAYSIZE(_fadeTable); ++i) {
+			_fadeTable[i] = 100;
+		}
+
+		// TODO: In SCI engine, GfxPalette (PaletteMgr) constructor handles
+		// loading the default palette, but ScummVM already loads it from
+		// somewhere else, so should ScummVM be updated to match SCI or should
+		// it be left the way it is already?
+		assert(kernelSetFromResource(999, true));
```
bluegr: Well, we initialize the default palette in SciEngine::initGraphics() via setDefault(), which occurs during engine initialization. We can just do that in the constructor. 

This was done in commit dc8b826f5ea77a0173c5ad64d6508f9ec7ed798a, in order to fix the palette during MT-32 emulation, where some MT-32 related messages can be set when starting a game. So, we can just leave it the way it is already, as the end result is the same,

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+	varyOff();
 	cycleAllOff();
 }
 
+void GfxPalette32::submit(Palette &palette) {
+	// TODO: This code won’t do much unless/until submit can
+	// receive HunkPalettes in addition to SOLPalettes. In SCI engine,
+	// SOLPalettes do not have timestamp/version fields, only
+	// HunkPalettes do, but in ScummVM there is only one type of palette.
+	if (palette.timestamp == _version) {
```
bluegr: Is there a reason why we would want to introduce another type of palette? If not, we can just remove this TODO. Otherwise, we might want to differentiate the two (but personally, I don't see the reason why we would want to do that...)

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+		return;
+	}
+
+	// TODO: Verify this is doing a copy.
+	Palette oldSourcePalette(_sourcePalette);
+
+	// TODO: In SCI32 this is actually calling SOLPalette::merge,
+	// so there should probably be a merge function on Sci::Palette
+	// instead of doing it inline here like this.
+	for (int i = 0; i < ARRAYSIZE(_sourcePalette.colors); ++i) {
```
bluegr: Oh, do SCI32 games use merging again? This functionality was removed in late SCI1 - SCI1.1. Do you have an example where palette merging takes place?

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+
+	if (!_needsUpdate && _sourcePalette != oldSourcePalette) {
+		++_version;
+		_needsUpdate = true;
+	}
+
+	// TODO: Technically this information is supposed to be
+	// persisted through a HunkPalette object; right now it
+	// would just be lost once the temporary palette was destroyed.
+	// palette.timestamp = _version;
```
bluegr: Hunk segments were not supposed to be stored in SCI saved games. Has this behavior been changed in SCI32?

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+}
+
+override bool GfxPalette32::kernelSetFromResource(GuiResourceId resourceId, bool force) {
+	// TODO: In SCI32, palettes that come from resources come in as
+	// HunkPalette objects, not SOLPalette objects. The HunkPalettes
+	// have some extra persistence stuff associated with them, such that
+	// when they are passed to GfxPalette32::submit, they would get the
+	// version number of GfxPalette32 assigned to them.
```
bluegr: Don't we use timestamps for versioning?

engines/sci/graphics/palette32.cpp
bluegr: OK :) So, just update the comment here

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+
+	// TODO: Copy assignment operator for Palette??
+	for (int i = 0; i < ARRAYSIZE(_sysPalette.colors); ++i) {
+		_sysPalette.colors[i] = _nextPalette.colors[i];
+
+		// NOTE: If the brightness option in the user configuration file is set,
+		// SCI engine adjusts palette brightnesses here by mapping RGB values to values
+		// in some hard-coded brightness tables. There is no reason on modern hardware
+		// to implement this, unless it is discovered that some game uses a non-standard
+		// brightness setting by default
```
bluegr: Perhaps we could implement that as an option in scummvm.ini? Though, I personally don't think it's worth the effort, so we could just leave it as-is

engines/sci/graphics/palette32.cpp
```diff
@@ -21,20 +21,156 @@
...
+		if (_sysPalette.colors[i].used) {
+			bpal[i * 3    ] = _sysPalette.colors[i].r;
+			bpal[i * 3 + 1] = _sysPalette.colors[i].g;
+			bpal[i * 3 + 2] = _sysPalette.colors[i].b;
+		}
+	}
+
+	// TODO: SCI engine does not actually invalidate the screen directly like this;
+	// should we really be doing it here?
+	_screen->_picNotValid = true;
```
bluegr: I think that this flag doesn't do anything for SCI32. I could be wrong, though (it's been awhile, and I can't look at the places where this is used right now)

engines/sci/graphics/palette32.cpp
```diff
@@ -82,9 +218,262 @@ byte GfxPalette32::matchClutColor(uint16 color) {
...
+}
+
+// TODO: This gets called *a lot* in at least the first scene
+// of SQ6. Optimisation would not be the worst idea in the world.
+void GfxPalette32::kernelPalVarySet(const GuiResourceId paletteId, const int16 percent, const int time, const int16 fromColor, const int16 toColor) {
+	Palette palette = getPaletteFromResourceInternal(paletteId);
+	setVary(&palette, percent, time, fromColor, toColor);
+}
+
+void GfxPalette32::kernelPalVaryMergeTarget(GuiResourceId paletteId) {
```
bluegr: Is this actually used in SCI32? Does it do palette merging, or is this something that was just added for completeness sake?

engines/sci/graphics/palette32.cpp
bluegr: Fair enough, this can be reviewed and optimized after the code has been merged

engines/sci/graphics/palette32.cpp
```diff
@@ -82,9 +218,262 @@ byte GfxPalette32::matchClutColor(uint16 color) {
...
+}
+
+void GfxPalette32::setVaryPercent(const int16 percent, const int time, const int16 fromColor, const int16 fromColorAlternate) {
+	if (_varyTargetPalette != nullptr) {
+		setVaryTimeInternal(percent, time);
+	}
+
+	// This looks like a mistake in the actual SCI engine (both SQ6 and Lighthouse);
+	// the values are always hardcoded to -1 in kPalVary, so this code can never
+	// actually be executed
```
bluegr: If that's the case, we should disable this code as well

engines/sci/graphics/palette32.cpp
bluegr: The rest of the code uses 0-255 if I'm not mistaken

engines/sci/graphics/palette32.cpp
```diff
@@ -82,9 +218,262 @@ byte GfxPalette32::matchClutColor(uint16 color) {
...
+
+void GfxPalette32::setStart(const Palette *const palette) {
+	if (_varyStartPalette != nullptr) {
+		delete _varyStartPalette;
+	}
+
+	_varyStartPalette = new Palette(*palette);
+}
+
+void GfxPalette32::mergeStart(const Palette *const palette) {
```
bluegr: Again, is SCI32 actually doing palette merging?

engines/sci/graphics/palette32.cpp
bluegr: :thumbsup:

engines/sci/graphics/palette32.cpp
```diff
@@ -312,27 +656,79 @@ void GfxPalette32::cycleAllOff() {
...
+	}
+
+	assert(toColor <= ARRAYSIZE(_fadeTable));
+
+	for (int i = fromColor; i < toColor; i++)
 		_fadeTable[i] = percent;
+
+	// TODO: This is necessary to get palette fading to work, but it seems like
+	// it should not be necessary.
+	g_system->updateScreen();
```
bluegr: A call to updateScreen() is needed to update the actual palette. 

However, theoretically, the screen should be updated when something is drawn on it. So, technically, this shouldn't be necessary indeed.

engines/sci/graphics/palette32.h
```diff
@@ -90,22 +118,122 @@ namespace Sci {
...
-		 * The fade table records the expected intensity level of each pixel in the palette that will be displayed on
-		 * the next frame.
+		 * The fade table records the expected intensity level of each pixel
+		 * in the palette that will be displayed on the next frame.
 		 */
 		byte _fadeTable[256];
+
+		/**
+		 * TODO: Documentation
+		 */
```
bluegr: This is documented in loadClut() and matchClutColor().

loadClut() will load a color lookup table from a clu file and set the palette found in the file. This is to be used with Phantasmagoria 2.
matchClutColor() matches a color in RGB565 format to a palette index based on the loaded CLUT

engines/sci/graphics/palette32.cpp
csnover: Everywhere! :) All of the palette submissions—both from `CelObj::SubmitPalette` and from `kPalVary`—go through `PaletteMgr::submit` function which (1) optionally converts `HunkPalette` to `SOLPalette` (it’s an overloaded method) and then (2) calls `SOLPalette::merge` which copies just the palette entries marked as "used" from the submitted palette onto the `PaletteMgr::sourcePalette`.

engines/sci/graphics/palette32.cpp
csnover: There aren’t hunk segments stored in save games—at least, not from `PaletteMgr::SaveRestore`. The persistence seems like it is probably just in-memory… perhaps the way Sierra chose to optimise palette resource loads because of how multiple palettes get merged into a single source palette?

The code that does it seemed pretty crazy at the time I looked at it, so I didn’t fully RE to figure out exactly how it was working. There is some pointer at `HunkPalette+0` which is set in `HunkPalette::Init` (this value always seems to be the last allocated handle from `MemoryMgr` immediately after a `ResourceMgr::load` call to load a palette), plus a value at `HunkPalette+4` which comes from `[HunkPalette+0]+0a`. When `HunkPalette::SetVersion` is called and `HunkPalette+4` is not zero, some arithmetic is performed to find the version number again (`HunkPalette+0 + 0x0d + 2 * HunkPalette+4 + 0x12`), and it’s compared to the `HunkPalette::version` field. `MsgMgr::Fatal` is called if they don’t match, otherwise the `newVersion` argument is put to both `HunkPalette::version` and that computed memory handle offset.

engines/sci/graphics/palette32.cpp
csnover: It seemed to be necessary to get palette updates to apply to the screen, but it may not be necessary now that SCI16 GfxPalette code is not being used to display the palette updates. I will double check what happens without it.

engines/sci/graphics/palette32.cpp
csnover: SQ6 does not use this opcode, so from that perspective it is just for completeness. However, there was a comment that QFG4 uses kPalVaryMergeStart though, so it is probably just a matter of going through all the SCI32 games to find one that uses this.

engines/sci/graphics/palette32.cpp
csnover: Yes :)

engines/sci/graphics/palette32.cpp
csnover: SQ6 title screen only performs palette operations, no draws. If a draw call needs to occur for the screen to be updated that is probably why `_screen->_picNotValid` needed to be set for some things to “work” (that was the screen I used to test kPalCycle), and also why `updateScreen` needs to be called here (maybe `kFrameOut` is never called during these kPalette setFade calls). I’ll take a deeper look.

engines/sci/engine/savegame.cpp
csnover: In SQ6 kPalVary is used to do things like change the brightness of your character as you walk around. It is not just used during animations. I did check the disassembly and everything listed here is similarly persisted by SQ6 SCI engine (actually it even persists the `_needsUpdate` flag).

engines/sci/graphics/palette32.cpp
csnover: Sorry, I guess this was a really bad/confusing comment. Because GfxPalette32 inherits from GfxPalette16 there is possibly still some code that relies on intensity and mapping being set to these defaults (I don’t think this is true any more, but I won’t know for sure until I have audited everything). However, SCI32 SOLPalettes don’t have intensity or mapping information (this data is now stored primarily on PaletteMgr) and so the engine shouldn’t ever be using this information for SCI32 games. This was supposed to be a note to myself that any part of the graphics engine that is using intensity/mapping needs to be changed to *not* do that for SCI32 games.

engines/sci/graphics/palette32.cpp
csnover: OK, I will remove it from here and just leave a note about why it is somewhere else, for anyone that might come across it in the future doing RE on another SCI32 engine version. Thanks for the information.

engines/sci/graphics/palette32.cpp
csnover: I feel like the answer is “probably not”, but it may end up being important depending upon that weirdo version setting that the engine does with the HunkPalettes. If ScummVM resource manager doesn’t persist hunk resources at runtime at all then the version/timestamp field might be worthless, since it seems like that was the only way it was used in the SCI32 engine.

I also question the fact that the SCI32 engine has a separate `PaletteMgr::nextPalette` that it copies to `PaletteMgr::currentPalette` (named `sysPalette` in ScummVM) for display. There is an accessor function to get `PaletteMgr::currentPalette` but I haven’t seen a call to it in any SCI code yet. If @wjp has some time I will probably ask to see if his indirect call tracing ever saw such a call. If there is one, I suspect it may be buried somewhere in remap.

engines/sci/graphics/palette32.cpp
csnover: Yes, though ScummVM had previously only stored the timestamp on Palettes, not also on PaletteMgr. As is probably apparent by now, the palette versioning in SCI32 is a bit of a mystery to me. I was having a small mental breakdown over whether to try to follow the timestamp-based approach in ScummVM or stay true to the original engine code and use a counter.

engines/sci/graphics/palette32.cpp
csnover: `kFrameOut` is called as expected on each frame, so I guess it is just that no draws are happening during that time. As such I will change `updateHardware` to call `g_sci->getEventManager()->updateScreen()`, which keeps the extra updates in check and matches more precisely how the original engine worked (and will prevent any other similar drawing errors later on).

engines/sci/graphics/palette32.h
csnover: Thanks, I’ve added that information in here.

engines/sci/graphics/palette32.cpp
csnover: FWIW it looks like the `ResourceMgr::Load` method in SCI engine returns a numeric handle into MemoryMgr::handles that then is used to access the resource data. The address of that handle is what gets assigned to the pointer at +0 of the HunkPalette object. At +0a is the initial version number of the palette. In all the palettes I checked this value is 1.

csnover:
This should be ready for final review/landing now.

bluegr:
Great, thanks for all your work once again! Merging :)

PR #651, opened 2016-01-25
I18N: Update German translation

rootfather:
This updates the German language file. Two SCI related strings are
added, one string that was not entirely clear is updated and fixed.
de_DE.po is now based on scummvm.pot as of commit
94207fbfb7a9090280c93f199a006e400aef6c24.

### DIFF
--- a/po/de_DE.po
+++ b/po/de_DE.po
@@ -1994,7 +1994,7 @@ msgstr ", öffentliches Verzeichnis nicht eingebunden"
 
 #: backends/platform/wii/options.cpp:174
 msgid "Network down"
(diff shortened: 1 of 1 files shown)


criezy:
In this pull request you forgot to remove the fuzzy tag for the strings you translated ("Enable high resolution graphics" and "Enable high resolution graphics/content"). Therefore there are still not used by ScummVM.

rootfather:
Sorry, fix will come in later today. I really have to fix my PoEdit setup,
doing everything manually in an editor just feels plain wrong. :)

PR #652, opened 2016-01-28
ALL: Removing code duplication via new managed surface & screen classes

dreammaster:
This pull request is my attempt to deal with the issue of code duplication between engines. As I work on new engines, I've found myself copying more and more code relating to managing surfaces, blitting with and without transparency, scaling of images, and screen updates with dirty rects handling. With this pull, I've consolidated all the code into classes in the graphics/ folder, and converted over several of the existing engines to use them.

Let me go over the classes involved:

Graphics::Surface
=============
This is no longer modified in any way.

Graphics::ManagedSurface
======================
This is a new class which has the bulk of the new surface functionality. It  now holds an internal Graphics::Surface as a field, rather than deriving from it. The class includes:
- Automatic managing of the surface's data in memory. So if you call create() for a managed surface, it will automatically be freed when the surface is.
- Blitting methods. Adds several variations of "blitFrom", and "transBlitFrom" for blitting with transparency. For scaling of transparent sprites, one variation has a bounding rect for both the source and destination, allowing any sub-section of the source sprite to be scaled to fill any arbitrary area.
- Support for sub-surfaces.. creating surfaces which represent a portion of a parent surface, and automatically translate any dirty rects for drawing to the parent surface in it's co-ordinates.

Graphics::Screen
=============
This class descends from Graphics::ManagedSurface, and provides basic screen management.in addition to a screen surface
- It implements addDirtyRect to add any dirty areas to an internal list of screen areas to be updated
- Implements an update method to update all the affected areas in the dirty rects list to the screen
- Automatically merges overlapping dirty rect areas prior to doing screen updates for better performance.
- It will also likely have generalised palette fading in/out routines in the near future.

Engines
======
The following engines have been updated/simplified to use the new ManagedSurface and Screen classes: Access, MADS, Sherlock, TsAGE, and Voyeur

I've done basic testing on all four engines, and everything still looks okay. I've also played Ringworld (TsAGE) all the way through to doubly ensure nothing is broken. Play-throughs of games for the other engines to follow.


lordhoto:
Right now, I just want to point out that changing Graphics::Surface like this breaks our ARM assembly code.

PaulGilbert:
I'm presuming you mean that adding the virtual method disrupts hardcoded field offsets in the assembly, since the class is no  longer POD? Hopefully this would be an easy matter of adjusting offsets.. in my model, it really benefits from being present in Graphics::Surface, since it meant that I didn't have to duplicate all the drawing methods like hLine, vLine, etc. in ManagedSurface, and ManagedSurface objects can be passed into existing code that expects Graphics:Surface objects.

wjp:
I'm having a lot of trouble reviewing this due to the fact that the total combined diff is huge, and the individual commits are useless to look at since many of them are changed later. Could you clean up the history?

dreammaster:
Okay, I'll look into flattening the commit history this weekend, per wjp's suggestion.

tramboi:
Maybe you could keep Graphics::Surface as the current POD with the same
layout and wrap it in a new class of yours with virtual dispatch (hardest
issue : finding a good name :p), so that the ARM code has not to be fixed
(it might be tricky not to forget anything).

2016-01-28 15:27 GMT+01:00 PaulGilbert <notifications@github.com>:

dreammaster:
Okay, I think the flattening of the commit history is all sorted out. I decided to keep the commit size down by separating the changes into a few commits, one for each engine, and another for all the changes in graphics/

wjp:
Thanks!

bluegr:
Update: this no longer merges cleanly

dreammaster:
I'll update it. I guess, given the issue with ARM, I'll also see about further refactoring ManagedSurface to contain an internal vanilla Graphics::Surface rather than the current virtual additions.

dreammaster:
Okay, I've refactored ManagedSurface to not derive from Graphics::Surface anymore, and reverted Graphics::Surface back to it's original POD state (and fixed all the resulting engine errors that occurred due to the change). I've squashed the commit history again, so the prior commits changing surface.cpp & surface.h are gone. As far as I'm aware, there are no remaining issues that need to still be addressed.

sev-:
I like it. I already envision how I could reuse it in couple of my engines.

sev-:
The code is clean, the APIs are good.

The only thing is that I would add a bit more explanation in the Screen class, particularly about its purpose.

PaulGilbert:
I've added some extra comments and fixed the issues wjp pointed out. Is it okay to merge now?

dreammaster:
Okay, I've fixed the issues wjp pointed out, and added some extra comments. Is it okay to merge it now?

sev-:
No objections from me.

graphics/managed_surface.h
```diff
@@ -303,7 +324,7 @@ class ManagedSurface {
 	 * Draw a thick line.
 	 */
 	void drawThickLine(int x0, int y0, int x1, int y1, int penX, int penY, uint32 color) {
-		_innerSurface.drawThickLine(x0, y0, x1, y1, penX, penY, color);
+		_innerSurface.drawThickLine(x0, y0, x1 + penX, y1 + penY, penX, penY, color);
```
wjp: I think you meant to change the addDirtyRect call here rather than the draw call.

dreammaster:
Darn it. You're right. :(. I'll fix it tongiht.

RichieSams:
I like this PR a lot. These helper classes could help me clean up the text and overlay code in ZVision.

dreammaster:
The drawThickLine addDirtyRect call is now fixed

wjp:
Thanks!

dreammaster:
Unless there are any objections, I presume it's okay to merge it in this weekend?

wjp:
No objections; nice work, and thanks for the clarifications.

PR #654, opened 2016-01-29
AGI: graphics rewrite + cleanup

m-kiewitz:
- graphics code fully rewritten
- Apple IIgs font support
- Amiga Topaz support
- Word parser rewritten
- menu code rewritten
- removed forced 2 second delay on all room changes
  replaced with heuristic to detect situations, where it's required
- lots of naming cleanup
- new console commands show_map, screenobj, vmvars and vmflags
- all sorts of hacks/workarounds removed
- added SCI wait mouse cursor
- added Apple IIgs mouse cursor
- added Atari ST mouse cursor
- added Amiga/Apple IIgs transition
- added Atari ST transition
- user can select another render mode and
  for example use Apple IIgs palette+transition for PC versions
- inventory screen rewritten
- SetSimple command now properly implemented
- PreAGI Mickey: Sierra logo now shown
- saved games: now saving controller key mapping
  also saving automatic save data (SetSimple command)
- fixed invalid memory access when saving games (31 bytes were saved
  using Common::String c_ptr()

Special Thanks to:
- fuzzie for helping out with the Apple IIgs font + valgrind
- eriktorbjorn for helping out with valgrind
- LordHoto for figuring out the code, that caused invalid memory
  access in the original code, when saving a game
- sev for helping out with reversing the Amiga transition

currently missing:
- mouse support for menu
- mouse support for system dialogs
- predictive dialog support
(it should be possible to add at least 1.+3. within a short time)


engines/agi/agi.cpp
```diff
@@ -520,30 +274,419 @@ AgiBase::~AgiBase() {
 	delete _rnd;
 
 	delete _sound;
+
+	if (_fontDataAllocated) {
```
sev-: The check is not really needed, free(NULL) is safe these days.

engines/agi/agi.cpp
```diff
@@ -520,30 +274,419 @@ AgiBase::~AgiBase() {
...
+
+// We load the Mickey Mouse font from MICKEY.EXE
+void AgiBase::loadFontMickey() {
+	Common::File interpreterFile;
+	int32 interpreterFileSize = 0;
+	byte *fontData = nullptr;
+
+	if (!interpreterFile.open("mickey.exe")) {
+		// Continue, if file not found
+		return;
```
sev-: I would issue a warning here

engines/agi/agi.cpp
```diff
@@ -520,30 +274,419 @@ AgiBase::~AgiBase() {
...
-		if (tmpMode != Common::kRenderDefault)
-			_renderMode = tmpMode;
+	if (getFeatures() & (GF_AGI256 | GF_AGI256_2)) {
+		// If current game is AGI256, switch (force) to VGA render mode
+		_renderMode = RENDERMODE_VGA;
 	}
 }
 
+void AgiBase::initFont() {
+	// We are currently using the custom font for all fanmade games
```
sev-: I recommend to move all these font-related methods to font.cpp file (leaving them in AgiBase class, though).

engines/agi/agi.cpp
```diff
@@ -520,30 +274,419 @@ AgiBase::~AgiBase() {
...
+
+	uint16 strikeRowOffset = 0;
+	uint16 strikeCurOffset = 0;
+
+	byte *fontData = nullptr;
+
+	if (!fontFile.open("agifont")) {
+		// Continue,
+		// This also happens when the user selected Apple IIgs as render for the palette for non-AppleIIgs games
+		warning("could not open agifont for Apple IIgs font");
```
sev-: Capitalize the sentence

engines/agi/agi.cpp
```diff
@@ -698,6 +838,30 @@ void AgiEngine::initialize() {
...
+	_picture->showPic();
+	_text->statusDraw();
+	_text->promptRedraw();
+}
+
+// Adjust a given coordinate to the local game screen
+// Used on mouse cursor coordinates before passing them to scripts
+void AgiEngine::adjustPosToGameScreen(int16 &x, int16 &y) {
+	x = x / 2; // 320 -> 160
+	y = y - 8; // remove status bar line
```
m-kiewitz: Will fix this, just noticed that I adjusted in a hardcoded manner and not according to when the play screen actually starts. Just a note that I already know about it.

engines/agi/keyboard.cpp
```diff
@@ -61,21 +62,230 @@ const uint8 scancodeTable[26] = {
...
-void AgiEngine::cleanInput() {
-	while (_game.numEgoWords)
-		free(_game.egoWords[--_game.numEgoWords].word);
-}
+	while (_eventMan->pollEvent(event)) {
+		switch (event.type) {
+		case Common::EVENT_PREDICTIVE_DIALOG: {
+			GUI::PredictiveDialog _predictiveDialog;
+			_predictiveDialog.runModal();
+#if 0
```
sev-: This functionality has to be restored

engines/agi/agi.cpp
```diff
@@ -710,22 +874,25 @@ AgiEngine::~AgiEngine() {
...
 	delete _gfx;
+	delete _words;
 	delete _console;
 }
 
 Common::Error AgiBase::init() {
 
 	// Initialize backend
-	initGraphics(320, 200, false);
+	//initGraphics(320, 200, false);
```
sev-: When looking at it locally, it looks weird. If the screen is inited elsewhere, why keep the call commented and not just remove it?

engines/agi/checks.cpp
```diff
@@ -21,159 +21,156 @@
...
 
 namespace Agi {
 
-int AgiEngine::checkPosition(VtEntry *v) {
-	debugC(4, kDebugLevelSprites, "check position @ %d, %d", v->xPos, v->yPos);
+bool AgiEngine::checkPosition(ScreenObjEntry *screenObj) {
+	bool result = true; // position is fine
+	debugC(4, kDebugLevelSprites, "check position @ %d, %d", screenObj->xPos, screenObj->yPos);
 
-	if (v->xPos < 0 ||
```
sev-: Why this check was removed?

m-kiewitz: It wasn't?
"if (screenObj->xPos < 0)" is done.

engines/agi/console.cpp
```diff
@@ -412,6 +417,182 @@ bool Console::Cmd_BT(int argc, const char **argv) {
...
+			if (screenObj->motionType == kMotionMoveObj) {
+				debugPrintf("motion: move obj\n");
+			} else {
+				debugPrintf("motion: ego\n");
+			}
+			debugPrintf("x: %d, y: %d, stepSize: %d, flag: %x\n", screenObj->move_x, screenObj->move_y, screenObj->move_stepSize, screenObj->move_flag);
+			break;
+		}
+	}
+#if 0
```
sev-: Why is it here?

engines/agi/cycle.cpp
```diff
@@ -402,61 +460,62 @@ int AgiEngine::runGame() {
...
 			break;
-		case Common::kRenderHercG:
-		case Common::kRenderHercA:
-			setvar(vMonitor, kAgiMonitorHercules);
+		case RENDERMODE_HERCULES:
+			setVar(VM_VAR_MONITOR, kAgiMonitorHercules);
 			break;
 		// Don't know if Amiga AGI games use a different value than kAgiMonitorEga
 		// for vMonitor so I just use kAgiMonitorEga for them (As was done before too).
-		case Common::kRenderAmiga:
```
sev-: Actually we now have all these render modes in Common code. Please consider switching to them

engines/agi/font.h
```diff
@@ -25,6 +25,181 @@
...
 namespace Agi {
 
+// Arrow to the right character, used for original saved game dialogs
+// Needs to get patched into at least the Apple IIgs font, because the font didn't support
+// that character and original AGI on Apple IIgs used Apple II menus for saving/restoring.
+static const uint8 fontData_ArrowRightCharacter[8] = {
+	0x00, 0x18, 0x0C, 0xFE, 0x0C, 0x18, 0x00, 0x00,
+};
+
+// topaz data, same as in engines\parallaction\staticres.cpp
```
sev-: Ouch. engines/parallaction/staticres.cpp  Use proper slashes.

engines/agi/keyboard.cpp
```diff
@@ -97,54 +307,128 @@ int AgiEngine::doPollKeyboard() {
...
+		}
+	}
+
 
 	// AGI 3.149 games, The Black Cauldron and King's Quest 4 need KEY_ESCAPE to use menus
 	// Games with the GF_ESCPAUSE flag need KEY_ESCAPE to pause the game
-	if (key == 0 ||
-		(key == KEY_ESCAPE && getVersion() != 0x3149 && getGameID() != GID_BC && getGameID() != GID_KQ4 && !(getFeatures() & GF_ESCPAUSE)) )
-		return false;
+	//		(key == KEY_ESCAPE && getVersion() != 0x3149 && getGameID() != GID_BC && getGameID() != GID_KQ4 && !(getFeatures() & GF_ESCPAUSE)) )
```
sev-: Will it be removed after confirmation the games work, or what?

engines/agi/op_test.cpp
```diff
@@ -221,7 +221,7 @@ uint8 AgiEngine::testCompareStrings(uint8 s1, uint8 s2) {
 			break;
 
 		default:
-			ms1[j++] = toupper(ms1[k]);
```
sev-: this looks suspicious. Why tolower instead of toupper?

engines/agi/keyboard.cpp
m-kiewitz: yes, i kept it in for now. But the original approach wasn't that good anyway. I can't test Black Cauldron because I don't own it. Menu works with the new method in KQ4.

engines/agi/preagi.cpp
```diff
@@ -55,32 +56,22 @@ PreAgiEngine::PreAgiEngine(OSystem *syst, const AGIGameDescription *gameDesc) :
...
-		_fontData = fontData_IBM;
-	}
-
 	_gfx->initMachine();
 
 	_game.gameFlags = 0;
 
-	_game.colorFg = 15;
-	_game.colorBg = 0;
+	//_game._vm->_text->charAttrib_Set(15, 0);
```
sev-: Why is this commented out, same question for the similar change below.

engines/agi/op_test.cpp
m-kiewitz: because that's what sierra did.

engines/agi/preagi.cpp
m-kiewitz: Because it's not needed at the moment. The preAGI code draws text by itself and doesn't use the TextMgr class for now (at least as far as I can see). I'm not sure if PreAGI should get changed accordingly or not.

engines/agi/cycle.cpp
m-kiewitz: I can change that it uses the Common render mode enums, but the internal render mode is not the same as the current configuration render mode. Those are not the same (don't have to be the same). For example for AGI256 games, we enforce VGA as internal render mode.

Back when I created that code, Apple IIgs + Atari ST were not available in the enum.

sev-:
I finished reviewing the PR. Great stuff!

bluegr:
Nice work Martin! :)
Will try to review this tonight

PR #655, opened 2016-01-29
AMIGAOS: Update amigaos.mk to automatically create an AmigaGuide file…

raziel-:
… from README.

Missing files RM2AG.rx and convertRM.sed will follow

### DIFF
--- a/backends/platform/sdl/amigaos/amigaos.mk
+++ b/backends/platform/sdl/amigaos/amigaos.mk
@@ -10,4 +10,8 @@ amigaosdist: $(EXECUTABLE)
 ifdef DIST_FILES_ENGINEDATA
 	cp $(DIST_FILES_ENGINEDATA) $(AMIGAOSPATH)/extras/
 endif
+	cat README | sed -f convertRM.sed > README.conv
+	rx RM2AG.rx README.conv
+	rm README.conv
+	cp README.guide $(AMIGAOSPATH)
 	cp $(DIST_FILES_DOCS) $(AMIGAOSPATH)
--- a/backends/platform/sdl/amigaos/amigaos.mk
+++ b/backends/platform/sdl/amigaos/amigaos.mk
@@ -10,9 +10,9 @@ amigaosdist: $(EXECUTABLE)
 ifdef DIST_FILES_ENGINEDATA
 	cp $(DIST_FILES_ENGINEDATA) $(AMIGAOSPATH)/extras/
 endif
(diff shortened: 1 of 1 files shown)


lordhoto:
I think our buildbot uses the `amigaosdist` target. This means we need to assure `rx` is installed and usable on buildbot.

raziel-:
Changes done.
Thanks for the hints @lordhoto

raziel-:
@lordhoto 

wrt the "rx" command...i could get rid of it completely if i can be sure that protection flags don't get lost on uploaded files.

I'd need to add +s (script) to the rexx file and add the "rx" command to the file's program launcher.

Never tried if github eats those flags

raziel-:
...and it would need an icon (.info) file which holds the program launcher information

sev-:
Thanks for the fixing, merging

raziel-:
@sev- 

Thank you

lordhoto:
Merging this caused the packaging step of AmigaOS builds on our buildbot [to fail because no 'rx' tool is present](http://buildbot.scummvm.org/builders/master-amigaos4/builds/5813/steps/package/logs/stdio).

polluks:
How about <code>**-**rx dists/amiga/RM2AG.rx README.conv</code>?
A German saying: "Keine Arme, keine Kekse"

raziel-:
@lordhoto 
@polluks 

I can live with completely removing that line if it intercepts any build process.
I wrote my own script which automatically updates the sources and builds some of the projects i (try to) maintain one after the other and can easily put that line in there.

As i said on irc it "rx" doesn't work from inside the makefile anyway.

lordhoto:
@raziel-: Just to check whether I got this right: does this mean all the changes from this pull request can be removed because they don't work when running `make` anyway?

raziel-:
@lordhoto 
Well, the "cat" line does what it should but arexx never starts to run because i get a "Program not found" error when it's run through the makefile...means, the automated process i tried to achieve is not performed to the end, the readme will only be converted once by cat, but never gets a conversion to an AmigaGuide file by arexx.
A guide therefore won't ever be copied either.

I tried and it doesn't even work when i add the arexx line to my local script after the makefile installed all the files and finished, i get the same error
I need to wait and let the whole local script finish and then perform the arexx command manually to make it work.
Not sure why it was working when i was testing, seems i missed checking it from within a makefile.

So, yes, i guess it wouldn't make sense keeping a partially working PR in tree.

raziel-:
Well, i can make it work locally (after some more tests) if i get rid of ($srcdir)/ in front of the rx script path

PR #656, opened 2016-01-29
POSIX: Follow XDG Base Directory Specification for User's Files

lordhoto:
This patch makes our POSIX ports use locations for configuration, log, and save games conforming to the XDG Base Directory Specification version 0.8 (right now the latest version, which can be found [here](http://standards.freedesktop.org/basedir-spec/basedir-spec-0.8.html)).

The new locations are as follows:
* Configuration file is now placed at `$XDG_CONFIG_HOME/scummvm/scummvm.ini`
* New default savegame directory is at `$XDG_DATA_HOME/scummvm/saves/`
* The log file is now placed at `$XDG_CACHE_HOME/scummvm/scummvm.log`

We still use the old locations for the configuration file and the savegame directory if those exist from a previous ScummVM installation.

This 'fixes' bug [#6036 POSIX: Use XDG dirs instead of HOME](https://sourceforge.net/p/scummvm/bugs/6036/). There is some discussion on the bug report about using a more easily visible location for save files. If we want that, we should do go for it now instead of having to worry about it again later.

### DIFF
--- a/backends/fs/posix/posix-fs.cpp
+++ b/backends/fs/posix/posix-fs.cpp
@@ -38,6 +38,7 @@
 #include <sys/stat.h>
 #include <dirent.h>
 #include <stdio.h>
+#include <errno.h>
 
 #ifdef __OS2__
 #define INCL_DOS
@@ -250,5 +251,68 @@ Common::SeekableReadStream *POSIXFilesystemNode::createReadStream() {
 Common::WriteStream *POSIXFilesystemNode::createWriteStream() {
 	return StdioStream::makeFromPath(getPath(), true);
 }
+
+namespace Posix {
+
+bool assureDirectoryExists(const Common::String &dir, const char *prefix) {
+	struct stat sb;
+
+	// Check whether the prefix exists if one is supplied.
+	if (prefix) {
+		if (stat(prefix, &sb) != 0) {
+			return false;
+		} else if (!S_ISDIR(sb.st_mode)) {
+			return false;
+		}
+	}
+
+	// Obtain absolute path.
+	Common::String path;
+	if (prefix) {
+		path = prefix;
+		path += '/';
+		path += dir;
+	} else {
+		path = dir;
+	}
+
+	path = Common::normalizePath(path, '/');
+
+	const Common::String::iterator end = path.end();
+	Common::String::iterator cur = path.begin();
+	if (*cur == '/')
+		++cur;
+
+	do {
+		if (cur + 1 != end) {
+			if (*cur != '/') {
+				continue;
+			}
+
+			// It is kind of ugly and against the purpose of Common::String to
+			// insert 0s inside, but this is just for a local string and
+			// simplifies the code a lot.
+			*cur = '\0';
+		}
+
+		if (mkdir(path.c_str(), 0755) != 0) {
+			if (errno == EEXIST) {
+				if (stat(path.c_str(), &sb) != 0) {
+					return false;
+				} else if (!S_ISDIR(sb.st_mode)) {
+					return false;
+				}
+			} else {
+				return false;
+			}
+		}
+
+		*cur = '/';
+	} while (cur++ != end);
+
+	return true;
+}
+
+} // End of namespace Posix
 
 #endif //#if defined(POSIX)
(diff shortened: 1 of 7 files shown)


backends/fs/posix/posix-fs.cpp
```diff
@@ -251,4 +252,67 @@ Common::WriteStream *POSIXFilesystemNode::createWriteStream() {
...
+			return false;
+		}
+	}
+
+	// Obtain absolute path.
+	Common::String path;
+	if (prefix) {
+		path = prefix;
+		path += '/';
+		path += dir;
```
sev-: Any catches for '../../../'? Or these should be fine?

backends/platform/sdl/posix/posix.cpp
```diff
@@ -82,11 +81,54 @@ bool OSystem_POSIX::hasFeature(Feature f) {
...
-	// to the HOME directory of the user.
-	const char *home = getenv("HOME");
-	if (home != NULL && (strlen(home) + 1 + _baseConfigName.size()) < MAXPATHLEN) {
-		configFile = Common::String::format("%s/%s", home, _baseConfigName.c_str());
+	Common::String prefix;
+#ifdef MACOSX
+	prefix = getenv("HOME");
+#elif !defined(SAMSUNGTV)
+	const char *envVar;
+	// Our old configuration file path for POSIX systems was ~/.scummvmrc.
```
sev-: Shouldn't we try to autoupgrade the file location?

backends/platform/sdl/posix/posix.h
```diff
@@ -28,7 +28,7 @@
 class OSystem_POSIX : public OSystem_SDL {
 public:
 	// Let the subclasses be able to change _baseConfigName in the constructor
-	OSystem_POSIX(Common::String baseConfigName = ".scummvmrc");
+	OSystem_POSIX(Common::String baseConfigName = "scummvm.ini");
```
sev-: That looks to me a bit ugly. We provide here the config file name, but suddenly in OSystem constructor, we have .scummvmrc hardcoded.

From the other hand, I see now way how to address it cleanly.

sev-:
In general, I had a good impression, and think that this could make it to 1.8.0

backends/platform/sdl/posix/posix.cpp
lordhoto: That is an option. However, I think it is annoying if you work with older versions of ScummVM for testing. You would end up having to maintain two configuration files. Thus, I am always using `~/.scummvmrc` if that exists.

If we go for auto-upgrading never the less, we will need to make sure we don't override the new one if it exists. Otherwise things can easily break when `~/.scummvmrc` is present (or gets created again by running an older version of ScummVM).

backends/platform/sdl/posix/posix.h
lordhoto: I want to get rid of this parameter anyway. Platforms having a different config file path should simply override `getDefaultConfigFileName`. SamsungTV really abused this by what looks like supplying an absolute path....

backends/fs/posix/posix-fs.cpp
lordhoto: Works fine, `mkdir` uses the CWD as base if the path is not absolute.

backends/platform/sdl/posix/posix.cpp
criezy: I would suggest waiting one version before adding autoupgrade so that the previous version also works with the new location when this happens. I am thinking about users who will find a regression when playing a game (that can happen!) and while waiting for 1.8.1 (or 1.9.0) release will switch back to the previous version to play that game. Of course that will not work for users witching to an older versions but hopefully would catch some (most?) of the cases of users downgrading. If we want to be ultra-conservative we can wait a bit more.

backends/platform/sdl/posix/posix.h
sev-: Do you want to do it before the merge or after?

lordhoto: After since it's general clean up and not directly related to this change.

sev-:
OK, please update the patch (now it has conflicts) and merge.

PR #657, opened 2016-01-29
Handle Alt-x internally in SCUMM.

lordhoto:
This moves handling of Alt-x to quit ScummVM to the SCUMM engine. cbd867329e018d7eca12b3a8842e52b8db9f494d suggests which is why it was added in the first place.

Alt-x is used in AGI's Leisure Suit Larry in the Land of the Lounge Lizards to skip the age protection. This pull request assures that this works on our ports formerly using Alt-x to quit ScummVM.

The README is updated accordingly and the changes are also mentioned in the NEWS file.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -65,6 +65,7 @@ For a more comprehensive changelog of the latest experimental code, see:
    - Implemented original Maniac Mansion v0-v1 walking code.
    - It is now possible to play Maniac Mansion from within Day of the
      Tentacle, with a few caveats. See README for details.
+   - Alt-x can now be used to quit SCUMM games on all platforms.
 
  Tinsel:
    - Improved AdLib music support in Discworld 1.
--- a/NEWS
+++ b/NEWS
@@ -19,6 +19,10 @@ For a more comprehensive changelog of the latest experimental code, see:
  General:
    - Updated Munt MT-32 emulation code to version 1.5.0.
 
+ SDL:
+   - Alt-x no longer quits ScummVM on platforms exhibiting this behavior
+     before. A notable example is our Windows port.
+
  3 Skulls of the Toltecs:
    - Improved AdLib music support.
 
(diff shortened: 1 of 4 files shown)


sev-:
I agree with this changes. However, I think it must be mentioned in a news post.

raziel-:
Vaguely related, not sure, but it's a long standing FR
https://sourceforge.net/p/scummvm/feature-requests/395/

If not related at all, i'm sorry for the noise

lordhoto:
> However, I think it must be mentioned in a news post.

To me that seems excessive, we cover it in our NEWS and README file.

> Vaguely related, not sure, but it's a long standing FR
https://sourceforge.net/p/scummvm/feature-requests/395/

The debugger hotkey is per engine. Most/all engines seem to use Ctrl-D nowadays though. So while the issue you see is related, this would be rather AGI internal, except for the fact that Ctrl-D is somewhat of a standard.

raziel-:
ok, thank you @lordhoto 

Completely off-topic now, sorry.

Would it be possible to use the alternative debugger hotkey (mentioned in the linked FR) and let CTRL-D bring up the dispatcher instead (just for that one single game)?

lordhoto:
> Would it be possible to use the alternative debugger hotkey (mentioned in the linked FR) and let CTRL-D bring up the dispatcher instead (just for that one single game)?

I think it would more make sense to not use Ctrl-D at all in AGI then. Otherwise that gets confusing fast. I don't know wheter "~" is used by AGI, but in general that seems a good choice for the debugger. It's also pretty well known from all the Quake engine games.

lskovlun:
For what it's worth, SCI does not use Ctrl-D, but requires Ctrl-Shift-D.

sev-:
Yes, I think putting it to ~ in SCI makes all sense

raziel-:
@sev- 

I guess you meant AGI?
The FR i linked to was about the AGI Amiga version.

See @lskovlun remark about SCI

PR #658, opened 2016-02-01
WAGE: New Engine

sev-:
Here is my PR for the WAGE engine, a WorldBuider (https://en.wikipedia.org/wiki/World_Builder) games. It was a nice and fun side project of me. There are over 100 games built on this engine.

I've tested few smaller games and they're completable, although a more thorough testing is required.

TODO:
- Sound (very simple SFX, sometimes looped, no music)
- About dialog
- Shortcuts from Actions menu
- Save/load
- Improvements in the input. It still doesn't match the original
- Copy/paste in the text console window
- Add more games to detection
- Window closing (I do not plan to implement window resizing like in original)
- General bugfixing

Also, there are some uncertainties and open questions in the game logic, primarily because there are not many games using those features. The whole engine is based on WAGE engine written in Java by Alex Svitkine (https://github.com/asvitkine/wage-engine) which in turn is mainly a guesswork. I plan to try to chase the original WorldBuilder author in a hope that he shares the sources with us.

In comparison to the Alex's engine, I strive to be pixel perfect and implement many features by referring to the original games running in the vMac emulator.

The engine optionally uses the original Mac fonts, there is a Mac-only script to downloads System 7.0.1 image from Apple (the OS has been put as freeware), extracts it and puts into wage.dat file. It is not clear to me whether we're allowed to redistribute the fonts. As a fallback, the engine uses built-in ScummVM fonts, and that is naturally not pixel-perfect.

The majority of these games were either freeware or shareware, so I plan to prepare 2 archives: originals and extracted games in the MacBinary format ready for ScummVM consumption.


bluegr:
Nice work, kudos! :)

I'll be able to check this out tomorrow

devtools/create_wage/create_wage.sh
```diff
@@ -0,0 +1,119 @@
...
+if test ! -f /Volumes/Fonts/Chicago; then
+	echo "Failed to attach fonts.dmg"
+	exit
+fi
+
+echo done
+
+echo_n "Copying fonts..."
+
+for i in Athens Cairo Chicago Courier Geneva Helvetica London "Los Angeles" Monaco "New York" Palatino "San Francisco" Symbol Times Venice
```
bluegr: Do we actually need all of these fonts? Cairo is the default and most probably necessary, but what about the others?

Can we use Liberation fonts as fallback, instead of the ScummVM built-in ones?
https://fedorahosted.org/liberation-fonts/

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+ */
+
+#include "wage/wage.h"
+#include "wage/entities.h"
+#include "wage/randomhat.h"
+#include "wage/world.h"
+
+namespace Wage {
+
+Obj *WageEngine::getOffer() {
```
bluegr: Um, what is an "offer" here? It's not really apparent. Could you please add a comment to explain what this does?

engines/wage/randomhat.h
```diff
@@ -0,0 +1,77 @@
...
+
+enum {
+	kTokWeapons = -400,
+	kTokMagic = -300,
+	kTokRun = -200,
+	kTokOffer = -100,
+	kTokNone = -100000
+};
+
+class RandomHat {
```
bluegr: I don't understand what this does... It seems to generate random values for weapon, magic etc usage based on tokens? Could you please add a short description?

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+
+Chr *WageEngine::getMonster() {
+	if (_monster != NULL && _monster->_currentScene != _world->_player->_currentScene) {
+		_monster = NULL;
+	}
+	return _monster;
+}
+
+void WageEngine::encounter(Chr *player, Chr *chr) {
+	char buf[512];
```
bluegr: Perhaps it would be better to use Common::String here, along with Common::String::format() instead of snprintf() below?

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+			hat.addTokens(kTokOffer, npc->_losingOffer + 1);
+	}
+
+	ObjList *objs = &npc->_currentScene->_objs;
+	if (npc->_inventory.size() < npc->_maximumCarriedObjects) {
+		int cnt = 0;
+		for (ObjList::const_iterator it = objs->begin(); it != objs->end(); ++it, ++cnt) {
+			if ((*it)->_type != Obj::IMMOBILE_OBJECT) {
+				// TODO: I'm not sure what the chance should be here.
+				hat.addTokens(cnt, 123);
```
bluegr: Does this actually happen? Should we use a warning here when this case occur?

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+			}
+		}
+	}
+
+	int token = hat.drawToken();
+	switch (token) {
+		case kTokWeapons:
+			// TODO: I think the monster should choose the "best" weapon.
+			performAttack(npc, player, weapons->operator[](_rnd->getRandomNumber(weapons->size() - 1)));
```
bluegr: Again, does this actually happen? Is there a test case? (same with kTokMagic below)

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+}
+
+const char *targets[] = { "head", "chest", "side" };
+
+void WageEngine::performAttack(Chr *attacker, Chr *victim, Obj *weapon) {
+	if (_world->_weaponMenuDisabled)
+		return;
+
+	// TODO: verify that a player not aiming will always target the chest??
+	int targetIndex = -1;
```
bluegr: I assume that this comment refers to the case where targetIndex is -1, i.e. when weapon->_type is Obj::MAGICAL_OBJECT, and chance < attacker->_physicalAccuracy. Should there be an error message in this case? Or will out of memory access occur?

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+			prependGenderSpecificPronoun(attacker->_gender), weapon->_name.c_str(),
+			victim->getDefiniteArticle(true), victim->_name.c_str());
+		appendText(buf);
+	}
+
+	playSound(weapon->_sound);
+
+	bool usesDecremented = false;
+	int chance = _rnd->getRandomNumber(255);
+	// TODO: what about obj accuracy
```
bluegr: This comment is a bit cryptic. What does object accuracy refer to?

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+}
+
+bool WageEngine::attackHit(Chr *attacker, Chr *victim, Obj *weapon, int targetIndex) {
+	bool receivedHitTextPrinted = false;
+	char buf[512];
+
+	if (targetIndex != -1) {
+		Obj *armor = victim->_armor[targetIndex];
+		if (armor != NULL) {
+			// TODO: Absorb some damage.
```
bluegr: Test case? Do we need some warning here?

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+					victim->_context._userVariables[SPIR_HIT_BAS];
+			snprintf(buf, 512, "%s%s's condition appears to be %s.",
+				victim->getDefiniteArticle(true), victim->_name.c_str(),
+				getPercentMessage(physicalPercent));
+			appendText(buf);
+		}
+	}
+
+	if (causesSpiritualDamage) {
+		/* TODO */
```
bluegr: Please add a warning here, it'll help us track a test case

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+	for (int i = 0; i < 4; i++, dir--)
+		if ((validMoves & (1 << i)) != 0) {
+			if (dir == 1) {
+				dir = i;
+				break;
+			}
+		}
+
+	char buf[256];
+	snprintf(buf, 256, "%s%s runs %s.", chr->getDefiniteArticle(true), chr->_name.c_str(), directionsS[dir]);
```
bluegr: Again, we could use Common::String here, together with Common::String::format()

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+		} else if (type == Chr::SHIELD_ARMOR) {
+			snprintf(buf, 256, "You are now wearing the %s.", obj->_name.c_str());
+			appendText(buf);
+		} else if (type == Chr::MAGIC_ARMOR) {
+			snprintf(buf, 256, "You are now wearing the %s.", obj->_name.c_str());
+			appendText(buf);
+		} else {
+			snprintf(buf, 256, "You now have the %s.", obj->_name.c_str());
+			appendText(buf);
+		}
```
bluegr: This is a bit redundant, and could be simplified as:

    switch (type) {
    case Chr::HEAD_ARMOR:
    case Chr::BODY_ARMOR:
    case Chr::SHIELD_ARMOR:
    case Chr::MAGIC_ARMOR:
        snprintf(buf, 256, "You are now wearing the %s.", obj->_name.c_str());
        break;
    default:
        snprintf(buf, 256, "You now have the %s.", obj->_name.c_str());
        break;
    }

    appendText(buf);

engines/wage/design.cpp
```diff
@@ -0,0 +1,892 @@
...
+
+	if (_surface == NULL) {
+		_surface = new Graphics::Surface;
+		_surface->create(_bounds->width(), _bounds->height(), Graphics::PixelFormat::createFormatCLUT8());
+		_surface->fillRect(r, kColorGreen);
+
+		needRender = true;
+	}
+
+#if 0
```
bluegr: Why is this bit disabled? Are the dialog graphics not drawn completely?

It seems that the main difference is the line thickness?

engines/wage/detection.cpp
```diff
@@ -0,0 +1,240 @@
...
+	return _gameDescription->filesDescriptions[0].fileName;
+}
+
+}
+
+static const PlainGameDescriptor wageGames[] = {
+	{"afm", "Another Fine Mess"},
+	{"amot", "A Mess O' Trouble"},
+	{"cantitoe", "Camp Cantitoe"},
+	{"scepters", "Enchanced Scepters"},
```
bluegr: Typo: Enchanted Scepters

engines/wage/detection.cpp
```diff
@@ -0,0 +1,240 @@
...
+	{"wage", "WAGE"},
+	{0, 0}
```
bluegr: Just a note here: AFAIK, there were a lot of games using this engine, about 50+. A reminder that they should be added in the detector as well - these are just a small subset

engines/wage/gui.cpp
```diff
@@ -0,0 +1,724 @@
...
+		active = !_sceneIsActive;
+		scrollable = true;
+		closeable = !_sceneIsActive;
+		closeBoxPressed = false;
+		drawTitle = false;
+		break;
+	}
+
+	drawBox(g, x, y, size, size);
+	drawBox(g, x+width-size-1, y, size, size);
```
bluegr: Formatting (applies to this and all the drawBox lines below)

engines/wage/gui.cpp
```diff
@@ -0,0 +1,724 @@
...
+			for (int yy = 0; yy < ARROW_H; yy++) {
+				for (int xx = 0; xx < ARROW_W; xx++) {
+					if (arrowPixels[yy][xx] != 0) {
+						g->hLine(x1+xx, y1+yy, x1+xx, kColorBlack);
+					} else {
+						g->hLine(x1+xx, y1+yy, x1+xx, kColorWhite);
+					}
+				}
+			}
+			fillRect(g, x + width - 13, y + size + ARROW_H, 8, height - 2*size - 1 - ARROW_H*2);
```
bluegr: Formatting

engines/wage/gui.cpp
```diff
@@ -0,0 +1,724 @@
...
+		fillRect(g, x + 5, y + size, 8, height - 2*size - 1);
+		if (!scrollable) {
+			fillRect(g, x + width - 13, y + size, 8, height - 2*size - 1);
+		} else {
+			int x1 = x + width - 15;
+			int y1 = y + size + 1;
+			for (int yy = 0; yy < ARROW_H; yy++) {
+				for (int xx = 0; xx < ARROW_W; xx++) {
+					if (arrowPixels[yy][xx] != 0) {
+						g->hLine(x1+xx, y1+yy, x1+xx, kColorBlack);
```
bluegr: Formatting

engines/wage/gui.cpp
```diff
@@ -0,0 +1,724 @@
...
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 0, 3, 0, 3, 3, 3, 3, 3, 3,
+	0, 0, 3, 3, 3, 0, 0, 3, 3, 3, 3,
+};
+
+static void cursor_timer_handler(void *refCon) {
```
bluegr: Use CamelCase here, like in the other parts

engines/wage/gui.cpp
```diff
@@ -0,0 +1,724 @@
...
+}
+
+void Gui::startMarking(int x, int y) {
+	warning("x: %d y: %d", x, y);
+	_selectionStartY = calcTextY(y);
+	_selectionStartX = calcTextX(x, _selectionStartY);
+
+	_inTextSelection = true;
+
+	warning("x: %d y: %d", _selectionStartX, _selectionStartY);
```
bluegr: I assume that the warnings here are left because this function isn't implemented completely? Please, add a TODO if this is the case

engines/wage/gui.cpp
```diff
@@ -0,0 +1,724 @@
...
+}
+
+void Gui::updateTextSelection(int x, int y) {
+	warning("x: %d y: %d", x, y);
```
bluegr: Same as startMarking()

engines/wage/combat.cpp
```diff
@@ -0,0 +1,915 @@
...
+		appendText(obj->_clickMessage.c_str());
+	}
+}
+
+bool WageEngine::handleMoveCommand(Directions dir, const char *dirName) {
+	Scene *playerScene = _world->_player->_currentScene;
+	const char *msg = playerScene->_messages[dir].c_str();
+
+	warning("Dir: %s  msg: %s", dirName, msg);
```
bluegr: Leftover debug?

engines/wage/script.cpp
```diff
@@ -0,0 +1,1127 @@
...
+
+		debug(9, "evalClickEquality(%s, %s, %d)", lhs->_value.designed->_name.c_str(), rhs->_value.designed->_name.c_str(), partialMatch);
+		debug(9, "l: %d r: %d (ch: %d ob: %d)", lhs->_type, rhs->_type, CHR, OBJ);
+		debug(9, "class: %d", lhs->_value.inputClick->_classType);
+
+		if (lhs->_value.inputClick->_classType == CHR || lhs->_value.inputClick->_classType == OBJ) {
+			Common::String name = lhs->_value.designed->_name;
+			name.toLowercase();
+
+			warning("%s <> %s", name.c_str(), str.c_str());
```
bluegr: Leftover debug?

engines/wage/wage.cpp
```diff
@@ -0,0 +1,502 @@
...
+	for (uint i = 0; i < _world->_orderedChrs.size(); i++) {
+		Chr *chr = _world->_orderedChrs[i];
+		if (!chr->_initialScene.equalsIgnoreCase(STORAGESCENE)) {
+			Common::String key = chr->_initialScene;
+			key.toLowercase();
+			if (_world->_scenes.contains(key) && _world->_scenes[key] != NULL) {
+				_world->move(chr, _world->_scenes[key]);
+
+				if (chr->_playerCharacter)
+					warning("Initial scene: %s", key.c_str());
```
bluegr: Leftover debug?

engines/wage/world.cpp
```diff
@@ -0,0 +1,510 @@
...
+		_soundLibrary1 = readPascalString(res);
+		_soundLibrary2 = readPascalString(res);
+
+		delete res;
+	}
+
+	Common::String *message;
+	if ((message = loadStringFromDITL(resMan, 2910, 1)) != NULL) {
+		message->trim();
+		warning("_gameOverMessage: %s", message->c_str());
```
bluegr: Leftover debug? (same with the other three warnings below)

bluegr:
Overall, this looks to be in a good state, kudos! :) I've left some comments here and there. Some further testing with the rest of the games would be nice, too (there are over 50 of them)

devtools/create_wage/create_wage.sh
sev-: Yes, fonts are referred in the game data. I saw at least 3 fonts used, and the game authors are free to use whatever they want.

As of the Liberation fonts, they will not be pixel-perfect, thus, in my opinion, not much better than the currently used built-in ones.

engines/wage/combat.cpp
sev-: Part of the gameplay. There is such command 'offer' when the hero may offer something to an NPC.

engines/wage/randomhat.h
sev-: Draws random number from classes with probability. Another part of the gameplay.

engines/wage/combat.cpp
sev-: Will do.

engines/wage/combat.cpp
sev-: It does. Just since the engine based on guesswork, actual number of token is not known.

engines/wage/combat.cpp
sev-: It always happens. The comment is related to the unknown logic of weapon choice. Should it be random or the monster could pick the best of its weapons...

engines/wage/combat.cpp
sev-: It is a todo to clarify the actual engine behavior.

engines/wage/combat.cpp
sev-: Obj::_accuracy variable. An attribute of a weapon object.

engines/wage/combat.cpp
sev-: It always happens, but not implemented as the logic is not known.

engines/wage/combat.cpp
sev-: Done.

engines/wage/combat.cpp
sev-: Not really. ToDo is to check the actual messages for these armor types.

engines/wage/design.cpp
sev-: That is for debugging purposes. Will be removed eventually.

engines/wage/detection.cpp
sev-: Fixed

engines/wage/detection.cpp
sev-: Not really. I do not want to pollute the global namespace with minor games. See ADGF_GENERIC define. Rest of the games use 'wage' target.

engines/wage/gui.cpp
sev-: Fixed.

engines/wage/gui.cpp
sev-: Fixed.

engines/wage/gui.cpp
sev-: Fixed.

engines/wage/gui.cpp
sev-: Done.

engines/wage/gui.cpp
sev-: I am working on it right now. These commits were made this afternoon.

engines/wage/gui.cpp
sev-: Same reply :)

engines/wage/combat.cpp
sev-: Right. Killed it.

engines/wage/script.cpp
sev-: Right

engines/wage/wage.cpp
sev-: Actually it is needed, so I put it under debug()

engines/wage/world.cpp
sev-: Hid them too

engines/wage/.gitignore
```diff
@@ -0,0 +1,3 @@
+.deps
```
wjp: This file (.gitignore) shouldn't be here. These things are handled by the main .gitignore.

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+			appendText(buf);
+		}
+	}
+
+	decrementUses(magicalObject);
+}
+
+static const int directionsX[] = { 0, 0, 1, -1 };
+static const int directionsY[] = { -1, 1, 0, 0 };
+static const char *directionsS[] = { "north", "south", "east", "west" };
```
wjp: static const char *const

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+	int directions = 0;
+	Scene *currentScene = npc->_currentScene;
+	for (int dir = 0; dir < 4; dir++) {
+		if (!currentScene->_blocked[dir]) {
+			int destX = currentScene->_worldX + directionsX[dir];
+			int destY = currentScene->_worldY + directionsY[dir];
+
+			Scene *scene = _world->getSceneAt(destX, destY);
+
+			if (scene != NULL && scene->_chrs.size() == 0) {
```
wjp: _chrs.empty(), since that's a Common::List

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+	int delta = maxHp - curHp;
+
+	if (delta > 0) {
+		int bonus = (int)(delta / (8 + _rnd->getRandomNumber(2)));
+		player->_context._statVariables[PHYS_HIT_CUR] += bonus;
+	}
+}
+
+void WageEngine::takeObj(Obj *obj) {
+  if ((int)_world->_player->_inventory.size() >= _world->_player->_maximumCarriedObjects) {
```
wjp: (Minor: wrong indent)

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+			snprintf(buf, 256, "You now have the %s.", obj->_name.c_str());
+			appendText(buf);
+		}
+		appendText(obj->_clickMessage.c_str());
+	}
+}
+
+bool WageEngine::handleMoveCommand(Directions dir, const char *dirName) {
+	Scene *playerScene = _world->_player->_currentScene;
+	const char *msg = playerScene->_messages[dir].c_str();
```
wjp: Might be cleaner to move the c_str() down to the appendText calls to avoid the strlen, but probably unimportant.

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+bool WageEngine::handleInventoryCommand() {
+	Chr *player = _world->_player;
+	ObjArray objs;
+
+	for (ObjArray::const_iterator it = player->_inventory.begin(); it != player->_inventory.end(); ++it)
+		if (!player->isWearing(*it))
+			objs.push_back(*it);
+
+	if (!objs.size()) {
+		appendText((char *)"Your pack is empty.");
```
wjp: Unnecessary cast.

wjp: (There are a few more of these.)

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+	} else {
+		Common::String res("Your pack contains ");
+		appendObjNames(res, objs);
+		appendText(res.c_str());
+	}
+
+	return true;
+}
+
+static const char *armorMessages[] = {
```
wjp: static const char *const

engines/wage/design.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+void FloodFill::addSeed(int x, int y) {
+	byte *p;
+
+	if (x >= 0 && x < _w && y >= 0 && y < _h) {
+		if (!_visited[y * _w + x] && *(p = (byte *)_surface->getBasePtr(x, y)) == _color1) {
+			_visited[y * _w + x] = 1;
+			*p = _color2;
+
+			Common::Point *pt = new Common::Point(x, y);
```
wjp: I don't think using a List of Point pointers makes sense here compared to simply using a List of Point objects. (The object is only 4 bytes.)

engines/wage/design.h
```diff
@@ -0,0 +1,115 @@
...
+
+	void setBounds(Common::Rect *bounds) {
+		_bounds = bounds;
+	}
+
+	Common::Rect *getBounds() {
+		return _bounds;
+	}
+
+    void paint(Graphics::Surface *canvas, Patterns &patterns, int x, int y);
```
wjp: Minor: there are a few wrong indents in this header)

engines/wage/detection.cpp
```diff
@@ -0,0 +1,240 @@
...
+	return desc != 0;
+}
+
+SaveStateList WageMetaEngine::listSaves(const char *target) const {
+	const uint32 WAGEflag = MKTAG('W','A','G','E');
+	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
+	Common::StringArray filenames;
+	char saveDesc[31];
+	Common::String pattern = target;
+	pattern += ".???";
```
wjp: You can use @lordhoto's new ".###' pattern here, probably.

engines/wage/design.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+	if (p->fillType > p->patterns->size())
+		return;
+
+	byte *pat = p->patterns->operator[](p->fillType - 1);
+
+	if (p->thickness == 1) {
+		if (x >= 0 && x < p->surface->w && y >= 0 && y < p->surface->h) {
+			*((byte *)p->surface->getBasePtr(x, y)) =
+				(pat[y % 8] & (1 << (7 - x % 8))) ?
```
wjp: These %8's will produce suboptimal code if optimizations are disabled since x and y are signed.

engines/wage/entities.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+	for (ChrList::const_iterator it = _chrs.begin(); it != _chrs.end(); ++it) {
+		debug(2, "paining Chr: %s", (*it)->_name.c_str());
+		(*it)->_design->paint(surface, ((WageEngine *)g_engine)->_world->_patterns, x, y);
+	}
+}
+
+// Source: Apple IIGS Technical Note #41, "Font Family Numbers"
+// http://apple2.boldt.ca/?page=til/tn.iigs.041
+const char *fontNames[] = {
```
wjp: const char *const (and maybe static)

engines/wage/gui.cpp
```diff
@@ -0,0 +1,856 @@
...
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 3, 0, 3, 3, 3, 3, 3, 3, 3,
+	3, 3, 0, 3, 0, 3, 3, 3, 3, 3, 3,
+	0, 0, 3, 3, 3, 0, 0, 3, 3, 3, 3,
+};
+
+static void cursorTimerHandler(void *refCon) {
```
wjp: This function scares me a bit. It seems to have some race conditions since setting cursoroff/cursorstate isn't protected by a mutex, and I'm not convinced it's safe to call copyRectToScreen from the timer thread.

And, should it always draw the cursor, even if _cursorOff?

engines/wage/menu.cpp
```diff
@@ -0,0 +1,563 @@
...
+	MenuItem(const char *n) : name(n) {}
+};
+
+struct MenuData {
+	int menunum;
+	const char *title;
+	int action;
+	byte shortcut;
+	bool enabled;
+} menuSubItems[] = {
```
wjp: static const?

engines/wage/script.h
```diff
@@ -0,0 +1,178 @@
...
+			Scene *scene;
+			int16 number;
+			Common::String *string;
+			Designed *inputClick;
+		} _value;
+		OperandType _type;
+		Common::String _str;
+
+		Operand(Obj *value, OperandType type) {
+			_value.obj = value;
```
wjp: Would asserts that the type matches be appropriate here?

engines/wage/script.cpp
```diff
@@ -0,0 +1,1126 @@
...
+	BLOCK_END,
+	STATEMENT,
+	OPERATOR,
+	OPCODE
+};
+
+struct Mapping {
+	const char *cmd;
+	int type;
+} mapping[] = {
```
wjp: static const?

engines/wage/util.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+Common::String readPascalString(Common::SeekableReadStream *in) {
+	Common::String s;
+	char *buf;
+	int len;
+	int i;
+
+	len = in->readSByte();
+	if (len < 0)
+		len += 256;
```
wjp: Isn't this just readByte(), then?

engines/wage/menu.cpp
```diff
@@ -0,0 +1,563 @@
...
+		while(i < string.size() && string[i] != ';') // Read token
+			item += string[i++];
+
+		if (item == "(-") {
+			menu->subitems.push_back(new MenuSubItem(NULL, 0));
+		} else {
+			bool enabled = true;
+			int style = 0;
+			char shortcut = 0;
+			char *shortptr = strrchr(item.c_str(), '/');
```
wjp: This fails to compile because strrchr expects a char* (not const char*).

engines/wage/world.cpp
```diff
@@ -0,0 +1,510 @@
...
+const char *World::getAboutMenuItemName() {
+	static char menu[256];
+
+	*menu = '\0';
+
+	if (_aboutMenuItemName.size() == 0) {
+		sprintf(menu, "About %s...", _name.c_str());
+	} else { // Replace '@' with name
+		const char *str = _aboutMenuItemName.c_str();
+		char *pos = strchr(str, '@');
```
wjp: This also doesn't compile due to strchr expecting a char*.

wjp:
Overall this looks very clean; nice work.

Only two real issues for me: I added line comments to two places that make it fail to build for me, and I don't know if the cursor handling in a timer proc is safe/allowed.

engines/wage/.gitignore
sev-: Right, it was from the time when the engine lived in a separate directory. Removed.

engines/wage/combat.cpp
sev-: fixed

engines/wage/combat.cpp
sev-: Yep. I changed the method footprint, and forgot about these. Fixed.

engines/wage/design.h
sev-: Fixed.

engines/wage/design.cpp
sev-: Right, copied it to units

engines/wage/entities.cpp
sev-: Fixed

engines/wage/gui.cpp
sev-: Heh, I had a race condition there :). See Gui::drawInput()

engines/wage/menu.cpp
sev-: Right

engines/wage/script.h
sev-: What do you mean? That is overloaded constructor, and there are no type casts in the callers.

engines/wage/util.cpp
sev-: I'm not sure, need to test.

engines/wage/menu.cpp
sev-: Which platform? I have the following prototype:
char *strrchr(const char *s, int c);

engines/wage/script.h
wjp: I was thinking of (presumably invalid?) calls like

     new Operand(Common::String("x"), INT);

engines/wage/gui.cpp
wjp: Yeah, I saw that call in there. I don't really understand how that makes the logic thread-safe, though.

Also, I asked @lordhoto about calling copyRectToScreen from a thread, and he pointed out the comment around system.h line 400, that says calls to OSystem graphics functions aren't guaranteed to be thread-safe.

engines/wage/menu.cpp
wjp: Ah, in fact I seem to have two:

    char *strchr (char *__s, int __c)
    const char *strchr (const char *__s, int __c)

(and same for strrchr). This is glibc 2.21

sev-:
Guys, I'd like to merge this and continue in-tree.

dreammaster:
For what it's worth from me, there's been adequate time, discussion, and
fixes already. And the engine is in a good enough state for merging in from
what little I've looked at it. So I'd say go for it.

On Sat, Feb 13, 2016 at 5:44 PM, Eugene Sandulenko <notifications@github.com

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+			if ((*it)->_type != Obj::IMMOBILE_OBJECT) {
+				// TODO: I'm not sure what the chance should be here.
+				hat.addTokens(cnt, 123);
+			}
+		}
+	}
+
+	int token = hat.drawToken();
+	switch (token) {
+		case kTokWeapons:
```
lordhoto: (Minor) We don't increase indent level for cases.

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+						performTake(npc, *it);
+				break;
+			}
+	}
+
+	delete weapons;
+	delete magics;
+}
+
+const char *targets[] = { "head", "chest", "side" };
```
lordhoto: Looks like these are never modified, thus `static const char *const` seems more appropriate.

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+	if (items != NULL) {
+		appendText(items->c_str());
+
+		delete items;
+	}
+
+	return true;
+}
+
+Common::String *WageEngine::getGroundItemsList(Scene *scene) {
```
lordhoto: It looks like this is used in two places. One right above, which checks for `NULL` as possible return value, and one in `WageEngine::attackHit` which does not care about it possibly returning `NULL`. Since it looks like when this returns non-`NULL` is always empty it might be more sensible to return a `Common::String` and perform `empty` checks. That also avoids manual memory management.

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+
+	if (objs.size()) {
+		Common::String *res = new Common::String("On the ground you see ");
+		appendObjNames(*res, objs);
+		return res;
+	}
+	return NULL;
+}
+
+void WageEngine::appendObjNames(Common::String &str, ObjArray &objs) {
```
lordhoto: This doesn't seem to modify contents of `objs`, thus I suggest making it `const`. I noticed `getIndefiniteArticle`, takes a non-`const` reference to `Common::String`, but since it doesn't change its contents either, it can easily be modified to take a `const` reference too, allowing this method (and maybe others) to use `const` in more places too.

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+}
+
+Common::String *WageEngine::getGroundItemsList(Scene *scene) {
+	ObjArray objs;
+
+	for (ObjList::const_iterator it = scene->_objs.begin(); it != scene->_objs.end(); ++it)
+		if ((*it)->_type != Obj::IMMOBILE_OBJECT)
+			objs.push_back(*it);
+
+	if (objs.size()) {
```
lordhoto: (Minor) I would suggest to always use `!container.empty()` instead of `container.size()` that way you have a consistent check for _is not empty_ which also doesn't do unnecessary computation for types like lists.

engines/wage/combat.cpp
```diff
@@ -0,0 +1,914 @@
...
+
+				return true;
+			}
+		}
+	}
+
+	return false;
+}
+
+bool WageEngine::tryAttack(Obj *weapon, Common::String &input) {
```
lordhoto: (Minor) This can also take `const` versions of both parameters. Might even make sense to make `weapon` a reference, but as far as I understand it, the only caller has a pointer anyway, so this isn't too important. But in general using (const) references can clarify that the parameter should always be non-`NULL`.

engines/wage/design.cpp
```diff
@@ -0,0 +1,539 @@
...
+ *
+ */
+
+#include "graphics/primitives.h"
+#include "wage/wage.h"
+#include "wage/design.h"
+
+namespace Wage {
+
+struct plotData {
```
lordhoto: (Minor) `PlotData` according to our naming guidelines.

engines/wage/design.cpp
```diff
@@ -0,0 +1,539 @@
...
+	_len = data->readUint16BE() - 2;
+	_data = (byte *)malloc(_len);
+	data->read(_data, _len);
+
+	_surface = NULL;
+}
+
+Design::~Design() {
+	free(_data);
+	if (_surface && _surface->getPixels())
```
lordhoto: (Minor) `Surface::free` also works when `getPixels` returns `NULL`, thus the second part of the check is not necessary.

engines/wage/design.cpp
```diff
@@ -0,0 +1,539 @@
...
+	Common::Array<int> xcoords;
+	Common::Array<int> ycoords;
+
+	numBytes -= 6;
+
+	while (numBytes > 0) {
+		int y2 = y1;
+		int x2 = x1;
+		int b = in.readSByte();
+		if ((b & 0xff) == 0x80) {
```
lordhoto: (Minor) This can be simplified to `b == -128` (same applies to below).

engines/wage/design.cpp
```diff
@@ -0,0 +1,539 @@
...
+		ycoords.push_back(y1);
+		x1 = x2;
+		y1 = y2;
+	}
+	xcoords.push_back(x1);
+	ycoords.push_back(y1);
+
+	int npoints = xcoords.size();
+	int *xpoints = (int *)calloc(npoints, sizeof(int));
+	int *ypoints = (int *)calloc(npoints, sizeof(int));
```
lordhoto: It looks like `xpoints` and `ypoints` are never modified during their lifetime. Thus, I think we can easily remove their creation and use `xcoords` and `ycoords` contents directly.

engines/wage/design.cpp
```diff
@@ -0,0 +1,539 @@
...
+					y++;
+					x = 0;
+					break;
+				}
+			}
+		}
+	}
+
+	while (numBytes--)
+		in.readByte();
```
lordhoto: It looks like `drawBitmap` is always used on a `SeekableReadStream`, this would allow to make `in` of such type and thus allow us to use `skip`/`seek` here.

engines/wage/design.cpp
lordhoto: Yup, same thoughts here :-).

engines/wage/design.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+#include "graphics/primitives.h"
+#include "wage/wage.h"
+#include "wage/design.h"
+
+namespace Wage {
+
+struct plotData {
+	Graphics::Surface *surface;
+	Patterns *patterns;
```
lordhoto: To me it looks we can make this `const Patterns &` and simplify a lot of code, i.e. use `[]` directly etc.

engines/wage/detection.cpp
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	return saveList;
+}
+
+int WageMetaEngine::getMaximumSaveSlot() const { return 999; }
+
+void WageMetaEngine::removeSaveState(const char *target, int slot) const {
+	char fileName[MAXPATHLEN];
+	sprintf(fileName, "%s.%03d", target, slot);
```
lordhoto: I suggest to use `Common::String::format` and let the savefile manager worry about length.

engines/wage/entities.h
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	void paint(Graphics::Surface *screen, int x, int y);
+
+	const char *getFontName();
+};
+
+class Sound {
+public:
+  Sound(Common::String name, Common::SeekableReadStream *data) : _name(name), _data(data) {}
```
lordhoto: (minor) this doesn't use tab for indentation

engines/wage/entities.h
```diff
@@ -0,0 +1,345 @@
...
+	~Sound() { }
```
lordhoto: (minor) superflous explicit destructor

engines/wage/gui.cpp
lordhoto: In general I would suggest to not call *any* graphics functionality from a timer.
`updateScreen` and `setMouseCursor` are *completely* unsafe to call from a timer for OpenGL based outputs. This is luckily not done here.
The other ones aren't thread-safe, but usually work with OpenGL output nowadays due to delayed updates.
I think we should update our `OSystem` documentation to reflect this.

engines/wage/randomhat.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include "common/random.h"
+
+#include "common/hashmap.h"
+#include "wage/randomhat.h"
+
+namespace Wage {
+
+void RandomHat::addTokens(int type, int count) {
+	_tokens.setVal(type, _tokens.getVal(type, 0) + count);
```
lordhoto: So, here I was to suggest we can simplify this to `_tokens[type] += count;` because we default initialize values. And then I saw that we don't on PS2 because of internal compiler errors happening when we do.... _sigh_

engines/wage/script.h
lordhoto: It might also make sense to remove  the `type` parameter when the `type` can be uniquely determined by the `value` parameter's type. (And then make those single parameter constructors `explicit` to prevent any silent conversions.)

engines/wage/script.h
```diff
@@ -0,0 +1,178 @@
...
+			_value.inputClick = value;
+			_type = type;
+		}
+
+		~Operand() {
+			if (_type == STRING)
+				delete _value.string;
+		}
+
+		Common::String toString() {
```
lordhoto: I sould suggest to move the implementation to `script.cpp`.

engines/wage/script.h
```diff
@@ -0,0 +1,178 @@
...
+			char buf[128];
+
+			if (_value.obj == NULL)
```
lordhoto: This check looks odd, what is the purpose? It looks like none of the cases below would check for a `NULL` pointer and thus crash anyway?

engines/wage/script.h
```diff
@@ -0,0 +1,178 @@
...
+			Obj *obj;
+			Chr *chr;
+			Designed *designed;
+			Scene *scene;
+			int16 number;
+			Common::String *string;
+			Designed *inputClick;
+		} _value;
+		OperandType _type;
+		Common::String _str;
```
lordhoto: I don't get the purpose of this member, it looks like it's only used locally in `toString`. Since `toString` doesn't return a const reference, there doesn't seem to be any need to have this around though. Thus I think we can simply remove it.

engines/wage/script.h
```diff
@@ -0,0 +1,178 @@
...
+		Common::String toString() {
+			char buf[128];
+
+			if (_value.obj == NULL)
+				_str = "";
+
+			switch(_type) {
+			case NUMBER:
+				_str = snprintf(buf, 128, "%d", _value.number);
+				return _str;
```
lordhoto: These two lines might be simplified to `return Common::String::format("%d", _value.number);`

engines/wage/util.cpp
lordhoto: Yup, this code has the same result as doing `readByte()` in the first place.

engines/wage/util.cpp
```diff
@@ -0,0 +1,126 @@
...
+	case 'i': case 'I':
+	case 'o': case 'O':
+	case 'u': case 'U':
+		return "an ";
+	}
+	return "a ";
+}
+
+enum {
+	GENDER_HE = 0,
```
lordhoto: (Minor) since this talks about gender it's more natural to use male, female, neuter here.

engines/wage/wage.cpp
```diff
@@ -0,0 +1,505 @@
...
+	_world->_commandsMenu = menu;
+
+	_gui->regenCommandsMenu();
+}
+
+void WageEngine::appendText(const char *str) {
+	_gui->appendText(str);
+
+	if (_inputText.size())
+		_inputText = "";
```
lordhoto: This can be simplified to `_inputText.clear()` and the `size` (again really a `!empty` check) seems superflous.

engines/wage/wage.h
```diff
@@ -0,0 +1,234 @@
...
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+#ifndef WAGE_H
+#define WAGE_H
```
lordhoto: `WAGE_WAGE_H`

engines/wage/world.cpp
```diff
@@ -0,0 +1,510 @@
...
+			}
+		}
+
+		delete res;
+	}
+
+	return NULL;
+}
+
+bool InvComparator(Obj *l, Obj *r) {
```
lordhoto: This is a function thus `invComparator` according to our guidelines. It can be made `static` too.

engines/wage/world.cpp
```diff
@@ -0,0 +1,510 @@
...
+	Designed *from = obj->removeFromCharOrScene();
+	obj->_currentOwner = chr;
+	chr->_inventory.push_back(obj);
+
+	Common::sort(chr->_inventory.begin(), chr->_inventory.end(), InvComparator);
+
+	_engine->onMove(obj, from, chr);
+}
+
+bool ObjComparator(Obj *o1, Obj *o2) {
```
lordhoto: Same as for `InvComparator`. Parameters can be made `const` too (which I forgot to mention earlier).

engines/wage/world.cpp
```diff
@@ -0,0 +1,510 @@
...
+	obj->_currentScene = scene;
+	scene->_objs.push_back(obj);
+
+	if (!skipSort)
+		Common::sort(scene->_objs.begin(), scene->_objs.end(), ObjComparator);
+
+	_engine->onMove(obj, from, scene);
+}
+
+bool ChrComparator(Chr *l, Chr *r) {
```
lordhoto: Same as for `ObjComparator`.

graphics/primitives.cpp
```diff
@@ -79,4 +96,344 @@ void drawThickLine(int x0, int y0, int x1, int y1, int penX, int penY, int color
...
+
+		for (i = 0; i < npoints; i++) {
+			if ((polyY[i] < pixelY && polyY[j] >= pixelY) || (polyY[j] < pixelY && polyY[i] >= pixelY)) {
+				nodeX[nodes++] = (int)(polyX[i] + (double)(pixelY - polyY[i]) / (double)(polyY[j]-polyY[i]) *
+														(double)(polyX[j] - polyX[i]) + 0.5);
+			}
+			j = i;
+		}
+
+		//  Sort the nodes, via a simple “Bubble” sort.
```
lordhoto: Not sure if it's worth it but one could simply use `Common::sort` here.

graphics/primitives.cpp
```diff
@@ -79,4 +96,344 @@ void drawThickLine(int x0, int y0, int x1, int y1, int penX, int penY, int color
...
+
+	//  Loop through the rows of the image.
+	for (int pixelY = bbox.top; pixelY < bbox.bottom; pixelY++) {
+		//  Build a list of nodes.
+		int nodes = 0;
+		j = npoints - 1;
+
+		for (i = 0; i < npoints; i++) {
+			if ((polyY[i] < pixelY && polyY[j] >= pixelY) || (polyY[j] < pixelY && polyY[i] >= pixelY)) {
+				nodeX[nodes++] = (int)(polyX[i] + (double)(pixelY - polyY[i]) / (double)(polyY[j]-polyY[i]) *
```
lordhoto: We might want to replace the floating point math here.

engines/wage/gui-console.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+	if (!str.contains('\n')) {
+		_out.push_back(str);
+		flowText(str);
+		return;
+	}
+
+	// Okay, we got new lines, need to split it
+	// and push substrings individually
+	Common::String tmp = "";
```
lordhoto: There's no need to initialize it with `""`.

engines/wage/gui-console.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+	for (uint i = 0; i < str.size(); i++) {
+		if (str[i] == '\n') {
+			_out.push_back(tmp);
+			flowText(tmp);
+			tmp = "";
```
lordhoto: `tmp.clear()`

engines/wage/menu.cpp
```diff
@@ -0,0 +1,563 @@
...
+					if (item[j] == '(') {
+						item.deleteChar(j);
+						break;
+					}
+			}
+
+			menu->subitems.push_back(new MenuSubItem(item.c_str(), kMenuActionCommand, style, shortcut, enabled));
+		}
+
+		item = "";
```
lordhoto: `item.clear()`

engines/wage/wage.cpp
```diff
@@ -0,0 +1,505 @@
...
+					_gui->drawInput();
+				}
+				break;
+
+			case Common::KEYCODE_RETURN:
+				if (_inputText.empty())
+					break;
+
+				processTurn(&_inputText, NULL);
+				_inputText = "";
```
lordhoto: `_inputText.clear()`

engines/wage/world.cpp
```diff
@@ -0,0 +1,510 @@
...
+				_patterns.push_back(pattern);
+			}
+		}
+		delete res;
+	}
+
+	res = resMan->getResource(MKTAG('M','E','N','U'), 2001);
+	if (res != NULL) {
+		Common::StringArray *menu = readMenu(res);
+		_aboutMenuItemName = "";
```
lordhoto: `_aboutMenuItemName.clear()`

engines/wage/world.cpp
lordhoto: We need to use `const char *` for `pos` here. Otherwise the compiler will try to pick the version of `strchr` taking `char *` resulting in a compilation error.

engines/wage/menu.cpp
lordhoto: We need to use `const char *` for `shortptr` here. Otherwise the compiler will try to pick the version of `strrchr` taking `char *` resulting in a compilation error.

Minor side remark: this should probably be named `shortPtr` according to our guidelines.

lordhoto:
I think overall it looks in a decent state. Good work. We'll need to fix the compilation errors before merge though. We might even want to fixup history here to help bisection in the future. @wjp any thoughts here?

I agree with @bluegr that it would be nice to have more comments. But we don't really have that in many other engines either, so I don't really see this as a hard requirement.

Some random thoughts which would be nice to see resolved later:
* There's a lot of pointer uses which seem unnecessary. Avoiding pointers when unecessary but using either values or const references instead could simplify a lot of code.
* Adding `const` to more places would definitely be nice too.
* I think we should strive for full names instead of abbreviations to ease readability. For example, `Chr` could be made `Character`.
* A lot of places use `size()` of containers for _is empty_ or _is not empty_ checks. I really suggest using `empty()` which conveys the intention of what you are trying to do much better and also has the benefit of being O(1) for our containers.
* There're a lot of warnings about signed/unsigned comparison in loops using `int` for their iterator and comparing it to `container.size()`

engines/wage/entities.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+	_context._statVariables[SPIR_STR_BAS] = _context._statVariables[SPIR_STR_CUR] = _spiritualStength;
+	_context._statVariables[SPIR_HIT_BAS] = _context._statVariables[SPIR_HIT_CUR] = _spiritialHp;
+	_context._statVariables[SPIR_ARM_BAS] = _context._statVariables[SPIR_ARM_CUR] = _naturalArmor;
+	_context._statVariables[SPIR_ACC_BAS] = _context._statVariables[SPIR_ACC_CUR] = _physicalAccuracy;
+
+	_context._statVariables[PHYS_SPE_BAS] = _context._statVariables[PHYS_SPE_CUR] = _runningSpeed;
+}
+
+ObjArray *Chr::getWeapons(bool includeMagic) {
```
lordhoto: We can simply return an `ObjArray` instead of a pointer here. This will simplify caller code by a lot.
RVO will likely be applied and avoid an unnecessary copy. And even if not, the copy should be reasonably small to not cause any issues (and if it does, we can worry about it later).

engines/wage/entities.cpp
```diff
@@ -0,0 +1,500 @@
...
+				list->push_back(_inventory[i]);
+			break;
+		default:
+			break;
+		}
+
+	return list;
+}
+
+ObjArray *Chr::getMagicalObjects() {
```
lordhoto: Same as for `Chr::getWeapons` a few lines above.

engines/wage/menu.cpp
sev-: Done and renamed

engines/wage/util.cpp
sev-: Fixed and tested.

engines/wage/gui.cpp
sev-: I rewrote it for drawing in the main thread.

engines/wage/script.h
sev-: Well, I control what I do with Operands, and that piece of code is feature complete. I'd rather not touch it.

engines/wage/world.cpp
sev-: Fixed

engines/wage/combat.cpp
sev-: Fixed.

engines/wage/combat.cpp
sev-: Fixed in both cases.

engines/wage/combat.cpp
sev-: Fixed.

engines/wage/world.cpp
sev-: Done both.

engines/wage/world.cpp
sev-: Done

engines/wage/world.cpp
sev-: Renamed

graphics/primitives.cpp
sev-: Right, fixed.

engines/wage/gui-console.cpp
sev-: Fixed

engines/wage/gui-console.cpp
sev-: Done

engines/wage/menu.cpp
sev-: Done

engines/wage/wage.cpp
sev-: Done

engines/wage/world.cpp
sev-: Done

engines/wage/wage.cpp
sev-: Fixed

engines/wage/entities.h
sev-: This is a TODO. Sound is not yet implemented

engines/wage/wage.h
sev-: Fixed

engines/wage/util.cpp
sev-: Renamed, thanks for suggestion

engines/wage/script.h
sev-: Fixed

engines/wage/script.h
sev-: This is a remnant from the original Java code. Removed

engines/wage/script.h
sev-: This is a leftover from Java code. Removed

engines/wage/script.h
sev-: Moved.

engines/wage/entities.h
sev-: Fixed

engines/wage/combat.cpp
sev-: Fixed.

engines/wage/combat.cpp
sev-: Sigh. Adjusted everywhere.

engines/wage/design.cpp
sev-: Fixed

engines/wage/design.cpp
sev-: Fixed

engines/wage/design.cpp
sev-: Fixed

engines/wage/design.cpp
sev-: Fixed

graphics/primitives.cpp
sev-: I was getting nasty rounding errors with integers and even floats.

engines/wage/randomhat.cpp
sev-: Sigh

engines/wage/detection.cpp
sev-: Fixed.

engines/wage/detection.cpp
sev-: Where is that? I do not see '#' handling in String::matchString()

engines/wage/design.cpp
sev-: In this case I will not be able to pass it to Graphics::drawPolygonScan()

sev-:
> We'll need to fix the compilation errors before merge though.

Seems to be fixed

> We might even want to fixup history here to help bisection in the future. @wjp any thoughts here?

I am not sure what you're referring to here

> I agree with @bluegr that it would be nice to have more comments. But we don't really have that in many other engines either, so I don't really see this as a hard requirement.

There was no goal to have the engine absolutely clear to everyone (as in every engine we have). Moreover, some of the confusion was related to the engine internals, not to the code itself

> There's a lot of pointer uses which seem unnecessary. Avoiding pointers when unecessary but using either values or const references instead could simplify a lot of code.

I addressed few cases which you referred.

> Adding const to more places would definitely be nice too.

That was addressed

> I think we should strive for full names instead of abbreviations to ease readability. For example, Chr could be made Character.

Wouldn't that just bloat the code?

> A lot of places use size() of containers for is empty or is not empty checks. I really suggest using empty() which conveys the intention of what you are trying to do much better and also has the benefit of being O(1) for our containers.

Fixed everywhere (I hope)

> There're a lot of warnings about signed/unsigned comparison in loops using int for their iterator and comparing it to container.size()

I am not getting those on Mac, will try on Linux which is less forgiving. But in any case I expect more errors/warnings to be uncovered once the buildbot deals with it.

engines/wage/detection.cpp
wjp: 06641f29a7bdcda280b0291f215193f055c83969

I suppose it was added after you branched for wage, so you'll probably have to merge in a recent master for this.

sev-:
Fixed int/uint mess

sev-:
Thanks for the review, guys, I'll merge this in an hour or so, once I have enough time to fix buildbot if needed.

engines/wage/design.cpp
lordhoto: You can easily do `xcoords.begin()` and `ycoords.begin()` to pass them to `Graphics::drawPolygonScan`

lordhoto:
>> We might even want to fixup history here to help bisection in the future. @wjp any thoughts here?
> I am not sure what you're referring to here

To the compilation errors due to `strrchr` and `strchr`. If we fix the when they were introduced by rewriting the history, we can easily do bisections over all intermediate revisions.

> Wouldn't that just bloat the code?

I don't see why it would.

sev-:
Fixed history. Merging now.

PR #659, opened 2016-02-02
SWORD25: Use more appropriate function for growing the stack

wjp:
The call to lua_growstack in unpersistThread would unconditionally at
least double the size of the stack. This caused memory usage to grow
exponentially (literally) with the number of serialized threads.

Bugs #6977, #6999.


RichieSams:
This is a good fix. lua_check stack will automatically grow the stack accordingly. http://pgl.yoyo.org/luai/i/lua_checkstack

So there is no need to do it ourselves. This is the only use of lua_grow stack. So, we can remove the function declaration and implementation.

wjp:
Done. Thanks for reviewing!

PR #661, opened 2016-02-06
I18N: Use German release titles in NEWS and README

rootfather:
This changes the release titles of supported games from the English original titles to the German ones.

I hope that I'm not too late for getting this merged in the 1.8.0 branch.


rootfather:
Small addition: Should I open another pull request to include this changes to the master branch? And what about the version information in the NEWS file (namely "1.8.0 (??.??.????)") - is that something you change JANITORIAL as soon as you tag the 1.8.0 release or is that something the translation teams have to care about themselves?

Please note that this is the first time I experience a new ScummVM release *that* close :)

criezy:
No need to open a separate pull request for master. I will cherry-pick the two commits once I have merged this one.

As for the second question, it depends if we remember to change the date everywhere. We might forget to do it in this translated NEWS file so checking would not do any arm.

PR #662, opened 2016-02-06
I18N: Update Polish translation

mziab:
Updated the Polish translation for ScummVM 1.8.0 and added proper plurals to the few strings that needed them.


sev-:
Thanks, merged

PR #665, opened 2016-02-13
IOS: Adds support for iPad

bSr43:
These keys are automatically added by Xcode when one compiles a regular Xcode project, but are missing when compiling from the command line.


bluegr:
Great work, thanks!
For reference:
https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/iPhoneOSKeys.html#//apple_ref/doc/uid/TP40009252-SW11

Merging

PR #666, opened 2016-02-13
SHERLOCK: New MusicPlayer class

dreammaster:
This is my attempt to refactor the existing MIDI playback code in the Music class to use a MidiPlayer (derived as a new MusicPlayer class) so that we can properly have music match the music_volume level. However, currently I'm not getting any music output.  Hopefully someone can assist in pointing out what minor little mistake I've done wrong and get this refactoring working correctly.

### DIFF
--- a/engines/sherlock/music.cpp
+++ b/engines/sherlock/music.cpp
@@ -218,12 +218,10 @@ void MidiParser_SH::unloadMusic() {
 /*----------------------------------------------------------------*/
 
 MusicPlayer::MusicPlayer(SherlockEngine *vm) : Audio::MidiPlayer(),
-		_vm(vm), _musicType(MT_NULL), _midiMusicData(nullptr) {
+		_vm(vm), _musicType(MT_NULL) {
 	MidiDriver::DeviceHandle dev;
 
 	if (IS_SERRATED_SCALPEL) {
-		// Serrated Scalpel: used an internal Electronic Arts .MUS music engine
-		_parser = new MidiParser_SH();
 		dev = MidiDriver::detectDevice(MDT_MIDI | MDT_ADLIB | MDT_PREFER_MT32);
 		_musicType = MidiDriver::getMusicType(dev);
 
@@ -244,8 +242,6 @@ MusicPlayer::MusicPlayer(SherlockEngine *vm) : Audio::MidiPlayer(),
 			break;
 		}
 	} else {
-		// Rose Tattooo: seems to use Miles Audio 3
-		_parser = MidiParser::createParser_XMIDI();
 		dev = MidiDriver::detectDevice(MDT_MIDI | MDT_ADLIB | MDT_PREFER_GM);
 		_musicType = MidiDriver::getMusicType(dev);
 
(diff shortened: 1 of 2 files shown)


eriktorbjorn:
I do get music with some audio drivers, but not AdLib. Is that what you meant? Either way, there seems to be a bug with memory being freed twice when closing the game, and I also got a crash when the music was about to loop. Perhaps the same crash?

eriktorbjorn:
Maybe there is no music because allocateChannel() and getPercussionChannel() in MidiDriver_SH_AdLib are defined to always return NULL? That probably means that MidiPlayer::sendToChannel() won't ever be able to send the command to a channel

dreammaster:
Hmmm.. if that's the case, why does the existing music (which ignores volume) working at all? But irrespective, if you're right, maybe some implementation can be copied from one of the other engines to finally fix the issue.

eriktorbjorn:
If you mean why the MT-32 music still works in Serrated Scalpel, that MidiDriver does implement allocateChannel() and getPercussionChannel(). I'm still not sure if it's the real reason, though.

raziel-:
Just to add my findings to this PR

Bug: https://bugs.scummvm.org/ticket/10309

bonki:
@eriktorbjorn is correct insofar as that `allocateChannel()` and `getPercussionChannel()` are the immediate reason as to why it won't produce any sound.

The underlying problem and the reason why the current (old) code works is, I believe, the following:

In the old code `MidiParser::sendToDriver()` calls `MidiDriver_SH_AdLib::send()`, probably due to `_midiParser` being passed to `_midiDriver->setTimerCallback()`? I haven't really checked.

The new code passes the new MusicPlayer instance instead which results in `MidiParser::sendToDriver()` calling `MidiPlayer::send()` instead (because `send()` it is not overridden in `MusicPlayer`) which, in turn, calls `MidiPlayer::sendToChannel()` which is where `getPercussionChannel()` and `allocateChannel()` come into play.

If you override `send()` and have it call `_driver->send()` it will follow the old code path and produce sound. This is most probably not the correct fix but at least explains why it currently works and doesn't work with the new MusicPlayer and hopefully provides you with enough information so you can properly fix it in the correct way.

Disclaimer: I haven't really looked much at `audio/midi{parser,player}.cpp` and don't know how any of these work (yet).

@eriktorbjorn is also right about the double free - `MidiPlayer::stop()` already frees `_midiData` for you and because it does so using `free()` `_midiData` is expected to be `malloc()`'d instead which probably means there's no point in overriding `unloadMusic()` (I think).

TL;DL: [This](https://github.com/bonki/scummvm/commit/938bf57bc6def76837ec43ce78ae9f0fcdc808dc) produces sound and fixes the crash.

sev-:
@eriktorbjorn shall we rebase it and finally merge?

dreammaster:
As it stands, the pull request still isn't feature complete, unfortunately. bonki's information sounds valuable, but at the time even with the replies I was unsure about how to complete a proper implementation. And since the engine currently works for music except for not obeying music volume, it wasn't a high priority to determine what a proper solution would be

Macleykun:
> As it stands, the pull request still isn't feature complete, unfortunately. bonki's information sounds valuable, but at the time even with the replies I was unsure about how to complete a proper implementation. And since the engine currently works for music except for not obeying music volume, it wasn't a high priority to determine what a proper solution would be

Want to give this PR another shot?

dreammaster:
Not personally.. my time is currently full up with other things. But as before, anyone else is welcome to do so.

PR #668, opened 2016-02-14
I18N: Updated Danish translation for 18.0

SteveNew:


criezy:
Thank you.
I manually merged the commit so that I could first revert the change to the Language field in the file (I guess changing "Dansk" to "da" was not done on purpose ;) ) and at the same time fix the commit message (to use "1.8.0" instead of "18.0").

PR #669, opened 2016-02-14
AMIGAOS: Fix amigaos.mk to make the rexx script work on local builds

raziel-:
Arexx can't workread makefile variables in paths.
It will error out with a "Program not found!" error.
Removing makefiles pre-defined "($srcdir)/" part makes it work on my platform for local native builds.

This is up for discussion of course, as it will still break buildbot's install procedure if i understand correctly.

### DIFF
--- a/backends/platform/sdl/amigaos/amigaos.mk
+++ b/backends/platform/sdl/amigaos/amigaos.mk
@@ -11,8 +11,13 @@ ifdef DIST_FILES_ENGINEDATA
 	cp $(DIST_FILES_ENGINEDATA) $(AMIGAOSPATH)/extras/
 endif
 	cat ${srcdir}/README | sed -f ${srcdir}/dists/amiga/convertRM.sed > README.conv
-	rx dists/amiga/RM2AG.rx README.conv
+	// AREXX doesn't understand makefile variables in it's execution path, i.e. ${srcdir}.
+	// It will break with a Program not found error. Therefore copying the script to cwd
+	// and removing it again once it finished.
+	cp ${srcdir}/dists/amiga/RM2AG.rx ${srcdir}
+	rx RM2AG.rx README.conv
 	cp README.guide $(AMIGAOSPATH)
+	rm RM2AG.rx
 	rm README.conv
 	rm README.guide
 	cp $(DIST_FILES_DOCS) $(AMIGAOSPATH)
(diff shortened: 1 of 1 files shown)


lordhoto:
While you are at it: Can you try whether adding "OPTIONS AREXX_BIFS AREXX_SEMANTICS" before "PARSE ARG readme_txt" in `dists/amiga/RM2AG.rx` still makes the script work for you?

raziel-:
Unfortunately not.

+++ Error 35 in line 10: Extraneous characters
Command returned 10/35: Extraneous characters

lordhoto:
*sigh* :-/

Is it possible to rewrite it using standard Rexx (streams), i.e. not rely on any ARexx specific functionality, and have it still work with ARexx?

raziel-:
Everything is possible, but it would probably break in any way on my platform due to the outdated rexx parser.
rx is from December 2006 thus based on an even older rexx revision, don't know which exactly, but when it comes to updated stuff i'm locked out.

...and btw, i don't know how to rewrite it using streams

lordhoto:
Does using `rexx -v` work to obtain what interpreter version you have?

lordhoto:
Anyway, back to the issue at hand:
I think it would be best if you copy `RM2AG.rx` to the cwd and use it from there and remove it afterwards. That way it will work for out of tree and in tree builds. And please add a comment why this is needed inside the `amigaos.mk` file. Otherwise someone might just try to simplify this in the future and we're stuck with brokenness again.

raziel-:
No, rx - v tells me that
5.System:> rx -v
Command returned 5/1: Program not found

I have to use AmigaOS's "version" command
5.System:> version rx full
rx 52.1 (09.12.2006)

raziel-:
What does cwd mean?

You mean like this?

cp ($srcdir)/dists/amiga/RM2AG.rx ($srcdir)
rx RM2AG.rx README.conv
cp ($srcdir)README.guide $(AMIGAOSPATH)
rm ($srcdir)RM2AG
rm ($srcdir)README.conv
rm ($srcdir)README.guide

raziel-:
cwd = commands working dir?

I added some changes that still work locally.
If i should add more info to the comment, please let me know (Especially about the builbot error which i have no idea whats going on)

lordhoto:
current working directory, sorry.

You'll have to copy it to `.` instead of `${srcdir}`.

raziel-:
Ah, thanks

Well, ${srcdir} gets translated to "." whithin the makefile script.
Should i still change the code to explicitely copy it to "."?

lordhoto:
> Should i still change the code to explicitely copy it to "."?

Yes, `srcdir` is only `.` when you build from the root of the source tree.

lordhoto:
Merged this a bit cleaned up as 89e9a36af546921afa7b870430e0405b0aa87048.

lordhoto:
Thanks!

PR #671, opened 2016-02-15
I18N: Fix encoding declaration in Czech PO

jwilk:
The file incorrectly declared the encoding as ISO-8859-2, when it was actually encoded in UTF-8 since 524e28a89bdb63b4585cec71fa738798040be1c4.

This bug was found by [i18nspector](http://jwilk.net/software/i18nspector).


lordhoto:
This is a more serious issue. The file should not be UTF-8 encoded at all. The commit you link only seems to change ASCII characters, which makes me wonder why it should be UTF-8 encoded since that commit.

jwilk:
Hmm. Normally encoding doesn't matter as long as it's portable and the declaration matches reality, because gettext converts encodings on the fly.
But then I don't really know anything about how scummvm works...
If you don't want UTF-8, I'd happy to convert the PO file back to ISO-8859-2.

lordhoto:
Well, we don't use gettext inside ScummVM. :-). If you can convert them to ISO-8859-2 and check that everything is fine, that would be great.

jwilk:
On a second thought, I can't convert it to ISO-8859-2, because of this message:
```
msgid "Subtitle speed: "
msgstr "Rychlost titulkù:"
```
The letter `ù` (U+00F9, LATIN SMALL LETTER U WITH GRAVE) is not representable in ISO-8859-2.
I think this is a typo, and the letter `ů` (U+016F, LATIN SMALL LETTER U WITH RING ABOVE) should have been used instead, just like in other translations of this phrase from the same file. But I don't speak Czech myself, so that's just a guess. @Tsbook, can you have a look at it?

Tsbook:
Right, that definitely is a typo. Although I have no idea how it got there, as this character is not used anywhere in Czech language. Anyway, the correct character is "ů", as you said.

criezy:
Thank you all. I fixed the typo and changed the encoding back to ISO-8859-2.

PR #673, opened 2016-02-15
I18N: Fix language code for Swedish

jwilk:
The language code for Swedish is `sv`, not `se`.


lordhoto:
Shouldn't this also rename the po file to `sv_SE`?

jwilk:
Yes, that's what I meant to do. Oops. Fixed now.

criezy:
Thank you.

PR #674, opened 2016-02-16
I18N: Update Basque translation

mikeliturbe:
This updates the Basque language file, translating the untranslated strings and correcting a few previously translated strings.


criezy:
Thank you. I slightly changed the commit message (to use 'I18N:' prefix as per our conventions) and merged the commit.

PR #675, opened 2016-02-17
SCI: SCI32 accurate graphics renderer

csnover:
Before anything, I have to give a special huge thanks to @wjp who did pretty much all of the work reversing the SCI calclists code, helped debug after the initial implementation was written, and found many of my most egregious bugs. Also thanks to everyone in IRC that has been following along and giving suggestions, advice, and information during the work: @m-kiewitz, @lordhoto, @lskovlun, @sev-, and some other folks who I can’t seem to find to list on GitHub. Thanks you all!! :+1: 

This PR introduces SCI32 accurate renderer architecture. Many cel drawing methods still need to be implemented, but I already have some code from @wjp to fill out most of this, so my next goal will be to finish that up so the rendering code is all reasonably solid, then continue working on text bitmap rendering and ScrollWindow, which are incomplete and known regressions from the old hacked code. So far probably like 75% or more of cel drawing already works since only a few draw methods are used most frequently. Depending on how people feel about it I may just start doing this work directly in master; let me know your thoughts. Eventually I will even get back to kSetShowStyle which is what I was trying to implement in the first place!

The only thing I am terribly nervous about here is the change to `Common::Array`, since obviously that is going to be exposed everywhere. I think it’s working, but I could use some extra eyes since it is entirely plausible that I made a really dumb mistake (I would have added unit tests… ¬_¬) and it is only a few lines.

Hopefully everything is reasonably clear and easy to review, I don’t expect a full review here from anyone since there’s a lot of known-still-busted stuff and it’s a lot of code, but I did just want to give everyone the opportunity to at least +1/-1 on the commit breakdown before I just start dumping code into master. I pinky swear I compiled in C++11 and C++98 modes and once again with SCI32 disabled and I checked EcoQuest and it’s great and oh god it’s going to break on MSVC anyway isn’t it, oh well. 🙉

Thanks,

### DIFF
--- a/common/array.h
+++ b/common/array.h
@@ -141,6 +141,12 @@ public:
 		insert_aux(_storage + idx, array.begin(), array.end());
 	}
 
+	/**
+	 * Inserts element before pos.
+	 */
+	void insert(iterator pos, const T &element) {
+		insert_aux(pos, &element, &element + 1);
+	}
 
 	T remove_at(size_type idx) {
 		assert(idx < _size);
@@ -185,6 +191,14 @@ public:
 		_storage = 0;
 		_size = 0;
 		_capacity = 0;
+	}
+
+	iterator erase(iterator pos) {
+		copy(pos + 1, _storage + _size, pos);
+		_size--;
+		// We also need to destroy the last object properly here.
+		_storage[_size].~T();
+		return pos;
 	}
 
 	bool empty() const {
--- a/engines/sci/engine/kgraphics32.cpp
+++ b/engines/sci/engine/kgraphics32.cpp
@@ -80,21 +80,18 @@ reg_t kCantBeHere32(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kAddScreenItem(EngineState *s, int argc, reg_t *argv) {
-	if (g_sci->_gfxFrameout->findScreenItem(argv[0]) == NULL)
-		g_sci->_gfxFrameout->kernelAddScreenItem(argv[0]);
-	else
-		g_sci->_gfxFrameout->kernelUpdateScreenItem(argv[0]);
-	return s->r_acc;
+	g_sci->_gfxFrameout->kernelAddScreenItem(argv[0]);
+	return NULL_REG;
 }
 
 reg_t kUpdateScreenItem(EngineState *s, int argc, reg_t *argv) {
 	g_sci->_gfxFrameout->kernelUpdateScreenItem(argv[0]);
-	return s->r_acc;
+	return NULL_REG;
 }
 
 reg_t kDeleteScreenItem(EngineState *s, int argc, reg_t *argv) {
 	g_sci->_gfxFrameout->kernelDeleteScreenItem(argv[0]);
-	return s->r_acc;
+	return NULL_REG;
 }
 
 reg_t kAddPlane(EngineState *s, int argc, reg_t *argv) {
@@ -115,11 +112,12 @@ reg_t kUpdatePlane(EngineState *s, int argc, reg_t *argv) {
 reg_t kAddPicAt(EngineState *s, int argc, reg_t *argv) {
 	reg_t planeObj = argv[0];
 	GuiResourceId pictureId = argv[1].toUint16();
-	int16 pictureX = argv[2].toSint16();
-	int16 pictureY = argv[3].toSint16();
+	int16 x = argv[2].toSint16();
+	int16 y = argv[3].toSint16();
+	bool mirrorX = argc > 4 ? argv[4].toSint16() : false;
 
-	g_sci->_gfxFrameout->kernelAddPicAt(planeObj, pictureId, pictureX, pictureY);
-	return s->r_acc;
+	g_sci->_gfxFrameout->kernelAddPicAt(planeObj, pictureId, x, y, mirrorX);
+	return NULL_REG;
 }
 
 reg_t kGetHighPlanePri(EngineState *s, int argc, reg_t *argv) {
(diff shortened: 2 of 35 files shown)


common/array.h
```diff
@@ -141,6 +145,12 @@ class Array {
 		insert_aux(_storage + idx, array.begin(), array.end());
 	}
 
+	/**
+	 * Inserts element before pos.
+	 */
```
bluegr: Shouldn't that be *after* pos?

csnover: It’s meant to be before, since that matches std::vector and Common::List and presumably insert_at. Is it not doing a before-insertion?

lordhoto: `insert_aux` is inserting an element before `pos`. So, this does behave like `std::vector::insert` when it comes to insertion position.

lordhoto:
> I would have added unit tests…

So, why didn't you? :-P We have some in `test/common/array.h` already. You can run them via `make test` (but you'll need to pass `--disable-eventrecorder` right now to make that work`).

engines/sci/engine/kgraphics32.cpp
```diff
@@ -206,41 +174,53 @@ reg_t kIsOnMe(EngineState *s, int argc, reg_t *argv) {
...
+	int16 backColor = readSelectorValue(segMan, object, SELECTOR(back));
+	int16 skipColor = readSelectorValue(segMan, object, SELECTOR(skip));
+	GuiResourceId fontId = (GuiResourceId)readSelectorValue(segMan, object, SELECTOR(font));
+	int16 borderColor = readSelectorValue(segMan, object, SELECTOR(borderColor));
+	int16 dimmed = readSelectorValue(segMan, object, SELECTOR(dimmed));
+
+	Common::Rect rect(
+		readSelectorValue(segMan, object, SELECTOR(textLeft)),
+		readSelectorValue(segMan, object, SELECTOR(textTop)),
+		readSelectorValue(segMan, object, SELECTOR(textRight)),
```
lordhoto: I remember reading about rect boundaries of (S)SCI to be handled different from ours in `Common::Rect`. Is this relevant here? It might make sense to have a comment at the places such things are releveant or not relevant (with some justification). That hopefully makes this a lot less confusing.

lordhoto:
The changes for COMMON look fine. But just don't ask me about the SCI bits.

csnover:
> So, why didn't you? :-P We have some in test/common/array.h already.

Man, out of sight, out of mind. The talking about tests in the IRC channel made me think there were none at all and create_project didn’t add the test directory to the project tree when it created the project. This is great news and I will be updating that commit with tests now!

engines/sci/engine/kgraphics32.cpp
csnover: Yes, unfortunately, I need to review all the places where rects come in since they all still mostly need their bottom-right corners extended by a pixel to be the correct size. Thanks for the reminder!

csnover:
OK I updated the PR with some tests for the new iterator methods and as long as I was doing that I also collapsed the warning-fixing commits. Any other concerns/objections before I land this?

engines/sci/graphics/lists32.h
```diff
@@ -0,0 +1,192 @@
...
+ * StablePointerArray.
+ */
+template <class T, uint N>
+class StablePointerArray {
+	uint _size;
+	T *_items[N];
+
+public:
+	typedef T **iterator;
+	typedef T *const *const_iterator;
```
bluegr: Shouldn't that be const **const_iterator?

csnover: Nope! `const T **const_iterator` (or `T const **const_iterator`) would be pointer to pointer to const T which is not correct. The pointer to T is what is const so this definition is what it is supposed to be.

engines/sci/graphics/palette32.cpp
```diff
@@ -206,17 +209,25 @@ int16 GfxPalette32::matchColor(const byte r, const byte g, const byte b, const i
 	return bestIndex;
 }
 
-void GfxPalette32::updateForFrame() {
+bool GfxPalette32::updateForFrame() {
 	applyAll();
 	_versionUpdated = false;
 	// TODO: Implement remapping
```
bluegr: Remapping was implemented via RE, and it is working mostly OK. Are there any showstoppers for re-enabling the current code?

engines/sci/graphics/plane32.h
```diff
@@ -0,0 +1,465 @@
...
+	// TODO: Are these ever actually used?
+	int _field_34, _field_38; // probably a point or ratio
+	int _field_3C, _field_40; // probably a point or ratio
+
+	/**
+	 * Converts the dimensions of the game rect used by
+	 * scripts to the dimensions of the plane rect used to
+	 * render content to the screen. Coordinates with
+	 * remainders are rounded up to the next whole pixel.
+	 */
```
bluegr: Isn't that the same as GfxCoordAdjuster32::kernelLocalToGlobal?

csnover: No, kLocalToGlobal takes the position of a coordinate relative to a plane and changes it to a coordinate relative to the top-left corner of the screen always in game script coordinates. This converts the game script coordinate rect of a plane into its screen coordinate rect.

engines/sci/graphics/palette32.cpp
csnover: As far as I know, remapping was not reversed for SCI32 (though it is for SCI16) and, based on my glancing over the two, is not implemented exactly the same way as in SCI16. I could try to make the SCI16 implementation work with changes, though it sounded like @m-kiewitz was thinking keeping SCI32 graphics code separate was the best idea since the SCI16 code already has a lot of complexity.

Most notably, it seems that the SCI16 remap code only supports one remap at a time, but SCI32 lets developers set up to 9 separate remaps, and each separate remap has its own independent state about what percent, the range, the type, its own remap map, et cetera. It also operates during the draw routines which are totally different in SCI32, which would make the old remapping code difficult to just re-enable. So, I sorry, I know I am saying “no” a lot but I think the answer here is again “no”. :(

engines/sci/graphics/text32.cpp
```diff
@@ -378,12 +588,8 @@ int16 GfxText32::Size(Common::Rect &rect, const char *text, GuiResourceId fontId
...
 	}
 
-	// Adjust the width/height if we're using an upscaled font
-	// for the scripts
-	if (_screen->fontIsUpscaled()) {
-		rect.right = rect.right * _screen->getWidth() / _screen->getDisplayWidth();
-		rect.bottom = rect.bottom * _screen->getHeight() / _screen->getDisplayHeight();
-	}
+	rect.right = rect.right * scriptWidth / _scaledWidth;
+	rect.bottom = rect.bottom * scriptHeight / _scaledHeight;
```
lordhoto: From my limited SCI knowledge I know that the original used different Rect styles (inclusive originally vs exclusive for us). When we do scaling like this, the resulting rect grows slower for downscaling and faster for upscaling in our scheme (AFAICT right now). So, it might be worth considering to use Sierra style scaling and then patch up coordiantes.

engines/sci/graphics/helpers.h
```diff
@@ -118,6 +126,90 @@ struct Window : public Port, public Common::Serializable {
...
+#ifdef ENABLE_SCI32
+/**
+ * Multiplies a rectangle by two ratios with default
+ * rounding. Modifies the rect directly.
+ */
+inline void mul(Common::Rect &rect, const Common::Rational &ratioX, const Common::Rational &ratioY) {
+	rect.left = (rect.left * ratioX).toInt();
+	rect.top = (rect.top * ratioY).toInt();
+	rect.right = (rect.right * ratioX).toInt();
+	rect.bottom = (rect.bottom * ratioY).toInt();
```
lordhoto: This scaling might also be affected by my other comment about scaling in the presence of inclusive vs. exclusive bottom-right.

csnover:
Per IRC: please do continue reviewing and let me know if you notice any problems, but I am going to go ahead and land this as well as changes from the sci32-draw-templating branch. There is at least one known memory problem with GfxText32 so I am focusing on that code for the next little while.

With regards to some of the rect stuff that @lordhoto brought up, it sounds like we’ll probably try working on the rect multiplication helpers to make those multiplications act more in line with the actual engine while still using Common::Rects internally.

csnover:
This is landed now!

PR #676, opened 2016-02-20
I18N: Update Polish translation.

mziab:
Updated Polish translation to include missing strings.


criezy:
Thank you. I have merged the commit to both master and the release branch for 1.8.0.

PR #682, opened 2016-02-25
ALL: Make SaveFileManager case insensitive.

lordhoto:
This pull request switches our SaveFileManager to offer case insensitive file access. This is intended to feature the same logical behavior on all our supported platforms. That is regardless of whether the platform has a case sensitive FS or a case insensitive one, we always expose a case insensitive one on a logical level.

The state this puts us into is:
- SaveFileManager API documentation clearly states it treats savefile names case insensitively
- DefaultSaveFileManager is offering case insensitive access on all platforms

The DefaultSaveFileManager change comes at the cost of us caching the directory contents (to avoid going through all files and finding one matching case insensitively on every operation), this means changes to the savepath outside ScummVM after it has been cached inside ScummVM will be missed.

What has been fixed:
- Formerly DefaultSaveFileManager was having inconsistent behavior: `listSaves` was case insensitive `openForLoading`/`openForSaving`/`removeSavefile`/`copySavefile`/`renameSavefile` were case sensitive if the underlying system was case sensitive. Now everything is case insensitive on all platforms.

What has been tested:
- Case insensitive savefile access on Linux with AGI as test engine

What has not been tested so far:
- Tizen changes
- Changing savepath

Backends which (might) need to be updated to support case insenstive access (from a quick grep based on the fact of them not using DefaultSaveFileManager or a derived class):
- DC
- N64
- DS
- PS2

### DIFF
--- a/backends/platform/tizen/system.cpp
+++ b/backends/platform/tizen/system.cpp
@@ -81,36 +81,41 @@ struct TizenSaveFileManager : public DefaultSaveFileManager {
 };
 
 bool TizenSaveFileManager::removeSavefile(const Common::String &filename) {
-	Common::String savePathName = getSavePath();
+	// Assure the savefile name cache is up-to-date.
+	assureCached(getSavePath());
+	if (getError().getCode() != Common::kNoError)
+		return false;
 
-	checkPath(Common::FSNode(savePathName));
-	if (getError().getCode() != Common::kNoError) {
+	// Obtain node if exists.
+	SaveFileCache::const_iterator file = _saveFileCache.find(filename);
+	if (file == _saveFileCache.end()) {
+		return false;
+	} else {
+		const Common::FSNode fileNode = file->_value;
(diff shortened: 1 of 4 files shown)


sev-:
As I understand, the purpose of this is to be consistent for our FSNode case-insensitivity, right?

I still wonder, are there any advantages beyond symmetric API behaviour (which is fine and enough too, btw).

wjp:
If I understand it correctly, the SaveFileManager was inconsistent within itself too. (e.g., listSaves was case-insensitive, but openForLoading was case-sensitive.)

lordhoto:
> As I understand, the purpose of this is to be consistent for our FSNode case-insensitivity, right?

The purpose is to make SaveFileManager behave the same on all platforms, yeah. This is in line with other engine facing APIs like `Common::File` or the `Common::Archive` API. They all offer case insentive access. `Common::FSNode` is not really relevant here since `FSNode` is really a low-level abstraction API, where system specific behavior is intended to work properly.

PR #683, opened 2016-02-25
I18N: Update Norwegian translations (Bokmaal/Nynorsk)

somaen:
Here are some updates to the norwegian translations.

I haven't had a chance to test any of these in-GUI yet, but I tried to keep it safe where possible (width-wise). Some things are still left untranslated, as I'm not 100% sure about their context.


criezy:
Thanks I added the commit to both master and branch-1-8.

In the future you might want to exclude the translation-dat file, or at least have it in a separate commit. As it is a binary file I am not sure merging would be easy if we were to modify that file (e.g. push an update to another language) in the repository between the moment you created your branch and the moment we decide to merge your pull request.

Actually in the future feel free to push directly to the main repo. Unless you have a doubt on something there is no need to open a pull request.

PR #688, opened 2016-03-02
COMMON: [GSoC] String::replace and Common::replace functionality added.

blorente:
### String::replace:
Inplemented some useful replace functions, with interfaces mimicking those of [the C++ std:: library] (http://en.cppreference.com/w/cpp/string/basic_string/replace). Unit tests are added for each of the new functions. These are `not` implemented using Common::replace, since that would unnecessarily elevate the algorithinc complexity and cache faults by a lot,

### Common::replace:
An all-purpouse algorithm for instance replacement in containers. It follows directly the implementation suggested in [the wiki page](http://wiki.scummvm.org/index.php/TODO#Algorithm_.2F_String).
Also includes unit tests with Common::List<> and Common::String.

### DIFF
--- a/common/algorithm.h
+++ b/common/algorithm.h
@@ -270,5 +270,26 @@ T gcd(T a, T b) {
 #pragma warning(pop)
 #endif
 
+/**
+ * Replacement algorithm for iterables.
+ *
+ * Replaces all occurrences of "original" in [begin, end) with occurrences of "replaced".
+ * 
+ * @param[in, out] begin: First element to be examined.
+ * @param[in] end: Last element in the seubsection. Not examined.
+ * @param[in] original: Elements to be replaced.
+ * @param[in] replaced: Element to replace occurrences of "original".
+ *
+ * @note Usage examples and unit tests may be found in "test/common/algorithm.h"
+ */
+template<class It, class Dat>
+void replace(It begin, It end, const Dat &original, const Dat &replaced) {
+	for (; begin != end; ++begin) {
+        if (*begin == original) {
+            *begin = replaced;
+        }
+    }
+}
+
 } // End of namespace Common
 #endif
--- a/common/str.cpp
+++ b/common/str.cpp
@@ -75,7 +75,7 @@ void String::initWithCStr(const char *str, uint32 len) {
 }
 
 String::String(const String &str)
-    : _size(str._size) {
+	: _size(str._size) {
 	if (str.isStorageIntern()) {
 		// String in internal storage: just copy it
 		memcpy(_storage, str._storage, _builtinCapacity);
@@ -91,7 +91,7 @@ String::String(const String &str)
 }
 
 String::String(char c)
-    : _size(0), _str(_storage) {
+	: _size(0), _str(_storage) {
 
 	_storage[0] = c;
 	_storage[1] = 0;
@@ -132,24 +132,19 @@ void String::ensureCapacity(uint32 new_size, bool keep_old) {
 	if (!isShared && new_size < curCapacity)
 		return;
 
-	if (isShared && new_size < _builtinCapacity) {
-		// We share the storage, but there is enough internal storage: Use that.
-		newStorage = _storage;
-		newCapacity = _builtinCapacity;
-	} else {
-		// We need to allocate storage on the heap!
+	// We need to allocate storage on the heap!
 
-		// Compute a suitable new capacity limit
-		// If the current capacity is sufficient we use the same capacity
-		if (new_size < curCapacity)
-			newCapacity = curCapacity;
-		else
-			newCapacity = MAX(curCapacity * 2, computeCapacity(new_size+1));
+	// Compute a suitable new capacity limit
+	// If the current capacity is sufficient we use the same capacity
+	if (new_size < curCapacity)
+		newCapacity = curCapacity;
+	else
+		newCapacity = MAX(curCapacity * 2, computeCapacity(new_size+1));
 
-		// Allocate new storage
-		newStorage = new char[newCapacity];
-		assert(newStorage);
-	}
+	// Allocate new storage
+	newStorage = new char[newCapacity];
+	assert(newStorage);
+	
 
 	// Copy old data if needed, elsewise reset the new storage.
 	if (keep_old) {
@@ -442,6 +437,58 @@ void String::trim() {
 
 uint String::hash() const {
 	return hashit(c_str());
+}
+
+void String::replace(uint32 pos, uint32 count, const String &str) {
+	replace(pos, count, str, 0, str._size);
+}
+
+void String::replace(uint32 pos, uint32 count, const char *str) {
+	replace(pos, count, str, 0, strlen(str));
+}
+
+void String::replace(iterator begin, iterator end, const String &str) {
+	replace(begin - _str, end - begin, str._str, 0, str._size);
+}
+
+void String::replace(iterator begin, iterator end, const char *str) {
+	replace(begin - _str, end - begin, str, 0, strlen(str));
+}
+
+void String::replace(uint32 posOri, uint32 countOri, const String &str,
+					 uint32 posDest, uint32 countDest) {
+	replace(posOri, countOri, str._str, posDest, countDest);
+}
+
+void String::replace(uint32 posOri, uint32 countOri, const char *str,
+					 uint32 posDest, uint32 countDest) {
+
+	ensureCapacity(_size + countDest - countOri, true);
+
+	// Prepare string for the replaced text.
+	if (countOri < countDest) {
+		uint32 offset = countDest - countOri; ///< Offset to copy the characters
+		uint32 newSize = _size + offset;
+		_size = newSize;
+
+		// Push the old characters to the end of the string
+		for (uint32 i = _size; i >= posOri + countDest; i--) 
+			_str[i] = _str[i - offset];
+
+	} else if (countOri > countDest){
+		uint32 offset = countOri - countDest; ///< Number of positions that we have to pull back
+
+		// Pull the remainder string back
+		for (uint32 i = posOri + countDest; i < _size; i++) 
+			_str[i] = _str[i + offset];	
+
+		_size -= offset; 
+	}
+
+	// Copy the replaced part of the string
+	for (uint32 i = 0; i < countDest; i++) 
+		_str[posOri + i] = str[posDest + i];
+
 }
 
 // static
--- a/common/str.h
+++ b/common/str.h
@@ -46,6 +46,17 @@ namespace Common {
 class String {
 public:
 	static const uint32 npos = 0xFFFFFFFF;
+
+	typedef char          value_type;
+	/**
+	 * Unsigned version of the underlying type. This can be used to cast
+	 * individual string characters to bigger integer types without sign
+	 * extension happening.
+	 */
+	typedef unsigned char unsigned_type;
+	typedef char *        iterator;
+	typedef const char *  const_iterator;
+
 protected:
 	/**
 	 * The size of the internal storage. Increasing this means less heap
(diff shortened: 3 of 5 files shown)


RichieSams:
Could you rebase the commit messages to follow our commit guidelines? 

test/common/algorithm.h
```diff
@@ -25,6 +25,26 @@ class AlgorithmTestSuite : public CxxTest::TestSuite {
...
 
+	template<typename It>
+	bool checkEqual(It one_first, It one_last, It other_first) {
+		if (one_first == one_last) 
+			return true;
+
+		// Check whether two containers have the same items in the same order,
+		// starting from some iterators one_first and other_first
+		// 
+		// It iterates trough the containers, comparing the elements one by one.
```
RichieSams: s/trough/through

common/str.h
```diff
@@ -224,6 +235,23 @@ class String {
...
+	 * Various replace functions to suit different parameters.
+	 * 
+	 * For now, they are unable to change the size of the string.
+	 * Thus, they do not insert, they replace the characters.
+	 */
+	// Replace 'count' bits, starting from 'pos' with str[pos, pos+count)
+	void replace(uint32 pos, uint32 count, const String& str);
+	// The same as above, but accepts a C-like array of characters	
+	void replace(uint32 pos, uint32 count, const char* str);
+	// Replace the space between the iterators with str[begin, end)
```
RichieSams: IMO, this is inaccurate. It should be something along the lines of:

// Replace the characters in [begin, end) with str[0, end - begin)

common/str.h
```diff
@@ -224,6 +235,23 @@ class String {
...
+	void replace(iterator begin, iterator end, const String& str);
+	// Replace _str[pos_ori, pos_ori+count_ori) with 
+	// str._str[pos_dest, pos_dest+count_dest)
```
RichieSams: This is also slightly inaccurate. count_dest is never used in the implementation

blorente:
@RichieSams I am sorry about the commit madness, Usually I only use git with small teams and no conventions, so the rebase was new to me.

RichieSams:
No worries! It looks great now

common/algorithm.h
```diff
@@ -270,5 +270,18 @@ T gcd(T a, T b) {
 #pragma warning(pop)
 #endif
 
+/**
+ * Replacement algorithm for iterables.
+ * Replaces the occurrences of original in [begin, end) with occurrences of replaced 
+ */
+template<class It, class Dat>
+void replace(It begin, It end, const Dat& original, const Dat& replaced) {
```
criezy: Our formatting convention for references is to have the space before the & sign and not after.
See: http://wiki.scummvm.org/index.php/Code_Formatting_Conventions#Whitespaces

common/str.cpp
```diff
@@ -444,6 +444,29 @@ uint String::hash() const {
 	return hashit(c_str());
 }
 
+void String::replace(uint pos, uint count, const String& str) {
```
criezy: Same remark as in algorithm.h: the space goes before the & and not after.
And the same goes for pointers: the space goes before the *.
I am not going to comment again on that but obviously they should all be fixed ;)

common/str.cpp
```diff
@@ -444,6 +444,29 @@ uint String::hash() const {
...
+	replace(pos, count, str._str, 0, count);
+}
+
+void String::replace(uint pos, uint count, const char* str) {
+	replace(pos, count, str, 0, count);
+}
+
+void String::replace(iterator begin, iterator end, const String& str) {
```
criezy: Maybe in this function it would be good to assert that str is long enough to avoid accessing characters beyond the end of the string.

common/str.h
```diff
@@ -224,6 +235,23 @@ class String {
 	uint hash() const;
 
 	/**
+	 * Various replace functions to suit different parameters.
+	 * 
+	 * For now, they are unable to change the size of the string.
+	 * Thus, they do not insert, they replace the characters.
+	 */
+	// Replace 'count' bits, starting from 'pos' with str[pos, pos+count)
```
criezy: I assumes you mean bytes here not bits.

test/common/algorithm.h
```diff
@@ -25,6 +25,26 @@ class AlgorithmTestSuite : public CxxTest::TestSuite {
...
+		// starting from some iterators one_first and other_first
+		// 
+		// It iterates through the containers, comparing the elements one by one.
+		// If it finds a discrepancy, it returns false. Otherwise, it returns true.
+
+		for (; one_first != one_last; ++one_first) {
+			if (*one_first != *other_first) {				
+				return false;
+			} 
+			++other_first;
```
criezy: Note in case you didn't know (and I saw it in str.cpp also but didn't comment): you can have both variable increment in the for.
for (; one_first != one_last; ++one_first, ++other_first) {

blorente: I didn't, thank you.

test/common/str.h
```diff
@@ -419,4 +419,24 @@ class StringTestSuite : public CxxTest::TestSuite
...
+		TS_ASSERT_EQUALS(testString, Common::String("Temporary my ass "));
+
+		// Using iterators (also a terribly useless program as a test).
+		testString.replace(testString.begin(), testString.end(), "Temporary my ass.");
+		TS_ASSERT_EQUALS(testString, Common::String("Temporary my ass."));
+
+		// Now with substrings.		
+		testString.replace(13, 4, "Temporary my butt", 13);
+		TS_ASSERT_EQUALS(testString, Common::String("Temporary my butt"));
+	}
```
uruk: I especially like these test cases. :)

sev-: Just to clarify: of course, this must be removed.

criezy:
Thanks. It looks good now.

criezy:
It would be good to squash those commits.
Once this is done I think this is good to be merged.

wjp:
I think we'll have to talk more about the exact interfaces and if we want all of them, since they're not quite consistent yet.

blorente:
I hope I didn't screw up in the squash. 
Of course, I am eager to discuss the interfaces. I jsut tried to copy and guess what you wanted from the std example given, but the more information I have the better.

common/algorithm.h
```diff
@@ -270,5 +270,18 @@ T gcd(T a, T b) {
 #pragma warning(pop)
 #endif
 
+/**
+ * Replacement algorithm for iterables.
+ * Replaces the occurrences of original in [begin, end) with occurrences of replaced 
+ */
+template<class It, class Dat>
+void replace(It begin, It end, const Dat &original, const Dat &replaced) {
+	for (; begin != end; ++begin) {
```
somaen: Wouldn't this be slightly more idiomatic if it was for(It it = begin; it != end; ++it), instead of modifying begin?

common/str.h
```diff
@@ -224,6 +235,23 @@ class String {
 	uint hash() const;
 
 	/**
+	 * Various replace functions to suit different parameters.
+	 * 
+	 * For now, they are unable to change the size of the string.
+	 * Thus, they do not insert, they replace the characters.
+	 */
+	// Replace 'count' bytes, starting from 'pos' with str[pos, pos+count)
```
somaen: Could these comments be Doyxgenified?

test/common/algorithm.h
```diff
@@ -25,6 +25,24 @@ class AlgorithmTestSuite : public CxxTest::TestSuite {
 		return true;
 	}
 
+	template<typename It>
+	bool checkEqual(It one_first, It one_last, It other_first) {
```
somaen: A doxygen comment above the template would be good.

common/algorithm.h
lordhoto: Actually how it's implemented right now is pretty common.

wjp:
Sorry for the delay in responding. Your replace functions currently don't match the versions in the C++ STL. For example, consider:

	std::string s = "abcd";
	s.replace(s.begin(), s.begin() + 1, "xx");

This results in `xxbcd`.

wjp:
Thanks, this looks better.

But here's a failing extra test:

        Common::String s1 = "TestTestTestTestTestTestTestTestTestTestTest";
        Common::String s2(s1);
        TS_ASSERT_EQUALS(s1, "TestTestTestTestTestTestTestTestTestTestTest");
        TS_ASSERT_EQUALS(s2, "TestTestTestTestTestTestTestTestTestTestTest");
        s1.replace(0, 4, "TEST");
        TS_ASSERT_EQUALS(s1, "TESTTestTestTestTestTestTestTestTestTestTest");
        TS_ASSERT_EQUALS(s2, "TestTestTestTestTestTestTestTestTestTestTest");

blorente:
My bad, I didn't think of making the string unique. It should be fixed now, all the tests are passed.

lordhoto:
@wjp: so what's the status content wise now? is this ready?

I think we should squash the commits. The first one also seems to have a different author than the other ones.

blorente:
@lordhoto: So I can't speak for the content (that's for you devs to decide) but whenever you want I can squash the commits. The second one (from a different person) is just there because I used MinGW git to commit once instead of my usual Commander shell. It's still me, but I think I can change it before sqashing, if necessary.

common/str.h
```diff
@@ -222,6 +233,38 @@ class String {
...
+     * @param posOri Same as pos
+     * @param countOri Same as count
+     * @param posDest Initial position to read str from.
+     * @param countDest Number of chars to read from str. npos by default.
+     */ 
+    // Replace 'count' bytes, starting from 'pos' with str[pos, pos+count).
+	void replace(uint32 pos, uint32 count, const String &str);
+	// The same as above, but accepts a C-like array of characters	
+	void replace(uint32 pos, uint32 count, const char *str);
+	// Replace the characters in [begin, end) with str._str[0, end - begin)
```
wjp: The comments here no longer match the implementation.

common/str.cpp
```diff
@@ -444,6 +444,60 @@ uint String::hash() const {
...
+
+void String::replace(uint32 posOri, uint32 countOri, const String &str,
+					 uint32 posDest, uint32 countDest) {
+	replace(posOri, countOri, str._str, posDest, countDest);
+}
+
+void String::replace(uint32 posOri, uint32 countOri, const char *str,
+					 uint32 posDest, uint32 countDest) {
+
+	assert(strlen(str) >= posDest + countDest);
```
wjp: This assert adds the requirement that str is null-terminated, which I'm not sure is desired.

common/str.cpp
```diff
@@ -444,6 +444,60 @@ uint String::hash() const {
...
+
+    ensureCapacity(_size + countDest - countOri, true);
```
wjp: (Minor: wrong indentation)

common/str.h
```diff
@@ -222,6 +233,38 @@ class String {
...
+     *
+     * @param pos Starting position for the replace in the original string.
+     * @param count Number of chars to replace from the original string.
+     * @param str Source of the new chars.
+     * @param posOri Same as pos
+     * @param countOri Same as count
+     * @param posDest Initial position to read str from.
+     * @param countDest Number of chars to read from str. npos by default.
+     */ 
+    // Replace 'count' bytes, starting from 'pos' with str[pos, pos+count).
```
wjp: (Minor: wrong indentation)

test/common/algorithm.h
```diff
@@ -25,6 +25,33 @@ class AlgorithmTestSuite : public CxxTest::TestSuite {
 		return true;
 	}
 
+
+    /**
```
wjp: (Minor: wrong indentation)

test/common/algorithm.h
```diff
@@ -25,6 +25,33 @@ class AlgorithmTestSuite : public CxxTest::TestSuite {
...
+     * Auxiliary function to check the equality of two generic collections (A and B).
```
wjp: That is not what it does, since it doesn't check the sizes are equal.

wjp:
One more failing test:

        Common::String s3 = "TestTestTestTestTestTestTestTestTestTestTest";
        Common::String s4(s3);
        s3.replace(0, 32, "");
        TS_ASSERT_EQUALS(s3, "TestTestTest");
        TS_ASSERT_EQUALS(s4, "TestTestTestTestTestTestTestTestTestTestTest");

But this is probably exposing a bug in String::ensureCapacity, where the `keep_old` case is interacting badly with the `// We share the storage, but there is enough internal storage: Use that.` branch.

blorente:
About the failing test, I think that the lines you referenced are indeed the cause, since the comment of the method says that no matter what, ensurecapacity makes it secure to write on, but that branch does not unshare it. Removing that case completely makes the test run, and every other test too.

The following commit demonstrates it.

blorente:
So, what's the status on this? Do you think it's ready to squash->merge?

sev-:
@blorente I think yes. Please squash it.

blorente:
I believe it is done :) I hope I didn't break anything

sev-:
Thanks!

sev-:
Argh! I did not notice that you left super-long commit log message.

Now it is merged. Sigh.

PR #689, opened 2016-03-04
SWORD25: Fix "Resource not released" warnings [GSoC]

Tkachov:
(Look https://sourceforge.net/p/scummvm/bugs/6980/)

The warning messages were appearing because when PRECACHE_RESOURCES is
not defined, ResourceManager's `requestResource()` was used to manually
"cache" them. When you do that, ResourceManager "locks" the resource
once, so you must `release()` it later.

Resources were not released, and warnings appeared. When you
`release()` resource, it still is loaded ("cached"). Resource's
`getLockCount()` is 0, though, so it might be unloaded by
ResourceManager if there is a lot of resources.

### DIFF
--- a/engines/sword25/gfx/animationresource.cpp
+++ b/engines/sword25/gfx/animationresource.cpp
@@ -211,8 +211,9 @@ bool AnimationResource::precacheAllFrames() const {
 			error("Could not precache \"%s\".", (*iter).fileName.c_str());
 			return false;
 		}
-#else
-		Kernel::getInstance()->getResourceManager()->requestResource((*iter).fileName);
+#else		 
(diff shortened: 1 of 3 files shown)


RichieSams:
Great catch!

Super minor nitpick. We like to use present tense for commit messages.

So it would be "SWORD25: Fix "Resource not released" warnings [GSoC]"

Tkachov:
Oh, missed that. I've edited the title.

bluegr:
Great work, well done! :)

Tkachov:
Thanks! =)

bluegr:
For reference: that was a bug introduced when I removed the precaching mechanism with commit a2f960017931ad871c37bf1d9c197418aa8afca9

sev-:
Good catch!

PR #691, opened 2016-03-05
WINTERMUTE: Return Common::String in getGameTargetName()

tobiatesan:


sev-:
And what is the advantage of this? Is there other usages of the result which require Common::String as input?

tobiatesan:
Just for coherence with the "sister methods" Common::String getGameId() and Common::Language getLanguage().

tobiatesan:
(Also _targetName is stored, internally, like a Common::String already, just like _gameId and _language.)

engines/wintermute/base/base_file_manager.cpp
```diff
@@ -318,7 +318,7 @@ Common::SeekableReadStream *BaseFileManager::openPkgFile(const Common::String &f
 
 bool BaseFileManager::hasFile(const Common::String &filename) {
 	if (scumm_strnicmp(filename.c_str(), "savegame:", 9) == 0) {
-		BasePersistenceManager pm(BaseEngine::instance().getGameTargetName());
+		BasePersistenceManager pm(BaseEngine::instance().getGameTargetName().c_str());
```
somaen: Might as well update the BasePersistenceManager-constructor to take Common::String& then.

tobiatesan: I guess why not.
Since Common::String inits to "" I believe the condition != "" to be equivalent to the nullptr-check.

Where I was at it I update initSave as well, to match initLoad.

sev-:
Now it all makes sense. Thanks

PR #695, opened 2016-03-05
WINTERMUTE: Add workaround for bug #7067

tobiatesan:
Carol Reed 5 (the full game, data.dcp hash
27b3efc018ade5ee8f4adf08b4e3c0dd) appears to refer to the absolute path
"f:\dokument\spel 5\demo\data\scenes\credits\op_cred_00\op_cred_00.jpg"

PR #696, opened 2016-03-05
WINTERMUTE: Rewrite absolute prefix workarounds block with loop

tobiatesan:
This one adds a commit on top of #695, it's the result of a few words exchanged with @somaen.

It replaces the large if(){} block which only grew larger in #695 with a loop.

See if it turns out useful.

### DIFF
--- a/engines/wintermute/base/file/base_disk_file.cpp
+++ b/engines/wintermute/base/file/base_disk_file.cpp
@@ -113,15 +113,28 @@ Common::SeekableReadStream *openDiskFile(const Common::String &filename) {
 	Common::String fixedFilename = filename;
 	correctSlashes(fixedFilename);
 
-	// Absolute path: TODO: Add specific fallbacks here.
+	// HACK: There are a few games around which mistakenly refer to absolute paths in the scripts.
+	// The original interpreter on Windows usually simply ignores them when it can't find them.
+	// We try to turn the known ones into relative paths.
 	if (fixedFilename.contains(':')) {
-		if (fixedFilename.hasPrefix("c:/windows/fonts/")) { // East Side Story refers to "c:\windows\fonts\framd.ttf"
-			fixedFilename = filename.c_str() + 14;
-		} else if (fixedFilename.hasPrefix("c:/carol6/svn/data/")) {	// Carol Reed 6: Black Circle refers to "c:\carol6\svn\data\sprites\system\help.png"
-			fixedFilename = fixedFilename.c_str() + 19;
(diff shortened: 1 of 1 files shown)


engines/wintermute/base/file/base_disk_file.cpp
```diff
@@ -115,11 +115,24 @@ Common::SeekableReadStream *openDiskFile(const Common::String &filename) {
 
 	// Absolute path: TODO: Add specific fallbacks here.
 	if (fixedFilename.contains(':')) {
-		if (fixedFilename.hasPrefix("c:/windows/fonts/")) { // East Side Story refers to "c:\windows\fonts\framd.ttf"
-			fixedFilename = filename.c_str() + 14;
-		} else if (fixedFilename.hasPrefix("c:/carol6/svn/data/")) {	// Carol Reed 6: Black Circle refers to "c:\carol6\svn\data\sprites\system\help.png"
-			fixedFilename = fixedFilename.c_str() + 19;
-		} else {
+		// We try a few workarounds for known broken absolute paths in game scripts...
+		const char* const knownPrefixes[] = {
```
sev-: Is there a sense to move this to the very top of the file? If so, then we also need to elaborate a bit more in the comment.

tobiatesan: I'm agnostic on the issue.

tobiatesan: On a second thought, I think it's slightly more readable right where it is.
I tried moving it and... I liked it slightly less.
If you strongly feel about the issue I don't object either way.

However, I added a // HACK comment to improve discoverability of... the hack.

tobiatesan:
By the way, about that whole block.
It seems to me that the original interpreter simply disregards hardcoded absolute paths and goes on with its business.
@somaen, I was wondering if there is a particular reason why we want to error out at all?
Not that I'm questioning the choice (it is a perfectly legitimate implementation for what is essentially undefined behaviour :) ), just curious.

sev-:
My question got dropped from the conversation due to an outdated commit.

Will it make sense to move that list on the very top of the file, making it easier to expand in the future?

tobiatesan:
Sorry about that @sev-.
So did my answer:  I think it's slightly more readable right where it is.
I tried moving it and... I liked it slightly less.
I am pretty agnostic on the issue, though. If you strongly feel about the issue I don't object either way.

sev-:
No, it was an open question. I'm merging this.

PR #697, opened 2016-03-06
RASPBERRYPI: remove some default configure options

joolswills:
Remove the turning off of the following options when building for the
Raspberry Pi, as they may be useful for users - eg timestamps for
internet connected Pis, or people wanting to use the event recorder, or
scalers. Update the README.RASPBERRYPI to reflect these changes.
- _savegame_timestamp
- _eventrec
- _build_scalers
- _build_hq_scalers

### DIFF
--- a/backends/platform/sdl/raspberrypi/README.RASPBERRYPI
+++ b/backends/platform/sdl/raspberrypi/README.RASPBERRYPI
@@ -69,9 +69,4 @@ You can find concise instructions for this on the ScummVM wiki:
 
 http://wiki.scummvm.org/index.php/Compiling_ScummVM/RPI 
 
-The configure script is disabling scalers because we prefer dispmanx for that, which 
-makes scalers unnecessary on a CPU limited platform like this, timestamps because most people
-doesn't have an RTC on the Raspberry Pi, and event recorder to save SD card write cycles.
-All these are automatically disabled when we crosscompile by passing "--host=raspberrypi".
-
 Enjoy!
--- a/configure
+++ b/configure
@@ -2651,10 +2651,6 @@ if test -n "$_host"; then
 			append_var LDFLAGS "-L$RPI_ROOT/opt/vc/lib"
 			# This is so optional OpenGL ES includes are found.
 			append_var CXXFLAGS "-I$RPI_ROOT/opt/vc/include"
-			_savegame_timestamp=no
-			_eventrec=no
-			_build_scalers=no
-			_build_hq_scalers=no
 			# We prefer SDL2 on the Raspberry Pi: acceleration now depends on it
 			# since SDL2 manages dispmanx/GLES2 very well internally.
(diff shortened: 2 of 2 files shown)


lordhoto:
Please follow our http://wiki.scummvm.org/index.php/Commit_Guidelines. Thank you.

joolswills:
sorry - should have read that first.

joolswills:
Does that look better?

joolswills:
I removed the corresponding text from the readme also - let me know if you need me to go into more details as to why I think they should be removed.

sev-:
@vanfanel Any comments?

vanfanel:
@sev- yes, scalers are not needed at all on the Pi, as it scales very well using hardware with SDL2, and using software scalers (like those horrible hqx scalers) is too much for the weak ARM11 CPU on a Pi Zero or Pi1, so people will rightfully start complaining about scummvm being slow.
Also, few Pi's around me are internet-connected and the Pi does not have an onboard RTC, so timestamps are better disabled.
I  am totally against enabling those on the Pi, sorry.

joolswills:
I don't think the options should be dictated based on your use case alone. The options can be switched off using compile flags and it can be documented to do so on the wiki if people want. A few points:

I am one of the developers for RetroPie. The userbase is mostly pi2 users - Looking at download for the latest image - only ~7000 downloads for rpi1. rpi2 image has ~172000 - the scalers perform well on a RPI2. When built with the scalers they are still optional within the hui - I don't think something should be disabled just because someone doesn't like them. I think some of the hq scalers improve the look of some games - but it is down to personal choice.

The majority of users of RetroPie have their RPI connected to the internet - and so have the correct time - the rpi3 has wireless built in also. Since timestamps can be disabled from the configure script, surely it is enough to document this on the wiki rather than forcing it. Again, let's leave it up to the end user.

sev-:
@joolswills Is there a way to detect pi2 in configure as a distinct platform?

joolswills:
Sorry missed the reply

If building on the RPI itself - but won't help for cross compiling or chroot builds. eg in RetroPie we do

```
        case $(sed -n '/^Hardware/s/^.*: \(.*\)/\1/p' < /proc/cpuinfo) in
            BCM2708)
                __platform="rpi1"
                ;;
            BCM2709)
                local revision=$(sed -n '/^Revision/s/^.*: \(.*\)/\1/p' < /proc/cpuinfo)
                if [[ "$revision" == "a02082" || "$revision" == "a22082" ]]; then
                    if [[ "$architecture" == "aarch64" ]]; then
                        __platform="rpi3-64"
                    else
                        __platform="rpi3"
                    fi
                else
                    __platform="rpi2"
                fi
                ;;
        esac
```

I don't see why anything would be needed - having this stuff enabled by default is no issue on any RPI, as the scalers are optional and are not forced on the user (but they are there should they want to use them).

PR #699, opened 2016-03-09
SWORD25: Add engine option to use English speech instead of German

criezy:
When selecting a non-English language, the scripts default to using
German voices. This commit adds an option to allow using English
voices instead. This is done by editing the file path on the fly. This
implements [bug #6804 - SWORD25: add option to choose language speech](https://sourceforge.net/p/scummvm/bugs/6804/).

If the English voice pack is not present it falls back to using German voices.

The way this is implemented here however does not allow to use German
voices with English text or English voices with German text. This could be
achieved with a slightly different implementation of the same idea.

This is proposed as a pull request to get some feedback:
Is this a good idea?
Is there a better way to implement this?
Should we allow using German voices with English text or English voices with German text?

### DIFF
--- a/engines/sword25/POTFILES
+++ b/engines/sword25/POTFILES
@@ -1,0 +1,1 @@
+engines/sword25/detection.cpp
--- a/engines/sword25/detection.cpp
+++ b/engines/sword25/detection.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "base/plugins.h"
+#include "common/translation.h"
 #include "engines/advancedDetector.h"
 
 #include "sword25/sword25.h"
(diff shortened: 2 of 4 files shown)


eriktorbjorn:
I think it would be nice if the subtitle language could be selected independent of the speech language, rather than hardcoding a "Use English speech" option. Though I guess that would require either changes to the game's Lua scripts (to add that to its options dialog), or the ability to add a dropdown widget to the engine settings tab.

Since I don't know how the engine handles speech and subtitles, that may be just wishful thinking though.

criezy:
@eriktorbjorn: that is actually how I implemented it initially. I stored a speech_language in the ini file instead of a bool and used that in the code instead of hardcoding 'en'. The issue with this however is that it indeed required to edit the ini file manually (which was fine for me, but not great for casual users).

sev-:
Yes, super-proper solution would be to implement speech language selection withing the game options menu, but this could be not trivial. Maybe one day I take a look into it.

So far I think the current approach is still viable, taking into account that there are no projects to add 3rd language for voice-overs.

PR #700, opened 2016-03-09
ADL: New engine

waltervn:
This PR is for an engine that runs Mystery House (Apple II version, 1987 PD release). The name ADL stands for "Adventure Development Language", which is the original name given to it by Ken Williams, according to sources on the internet.

I suspect that the 1987 PD release uses a later version of ADL. Unfortunately, I have been unable to find the original version of Mystery House to make a comparison.

I assume that the other titles in the Hi-Res series will be quite similar to Mystery House internally, and that support for them can be added to this ADL engine in the future.

Before anyone looks at the code, we should probably discuss the following:

- Is Mystery House suitable for inclusion into ScummVM?

and if it is:

- Should ADL be a separate engine, or should it be converted into a Pre-AGI engine, like TrollVM.

While Mystery House shares nothing with AGI or Pre-AGI, later Hi-Res titles might use AGI-like PICs. Personally I prefer having it as a separate engine. We're probably talking about 10 or so games here across multiple platforms. While they might share some graphics formats with AGI, I don't think that is enough to put them under the AGI engine.

Thoughts?


sev-:
Great work. The engine is very clean

engines/adl/adl.cpp
```diff
@@ -0,0 +1,1038 @@
...
+
+	if (!outFile) {
+		warning("Failed to open file '%s'", fileName.c_str());
+		return Common::kUnknownError;
+	}
+
+	outFile->writeUint32BE(MKTAG('A', 'D', 'L', ':'));
+	outFile->writeByte(SAVEGAME_VERSION);
+
+	char name[SAVEGAME_NAME_LEN] = { };
```
lordhoto: minor: I think it would make sense to remove the default initialization and simply call `Common::strlcpy` instead of `strncpy` below.

engines/adl/adl.cpp
```diff
@@ -0,0 +1,1038 @@
...
+	bits += quadrant;
+
+	if (bits & 1)
+		p.x += (bits & 2 ? -1 : 1);
+	else
+		p.y += (bits & 2 ? 1 : -1);
+}
+
+void AdlEngine::drawLineArt(const Common::Array<byte> &lineArt, const Common::Point &pos, byte rotation, byte scaling, byte color) const {
+	const byte stepping[] = {
```
lordhoto: minor: I guess this could be made static.

engines/adl/adl.h
```diff
@@ -0,0 +1,247 @@
...
+	bool doOneCommand(const Commands &commands, byte verb, byte noun);
+	void doAllCommands(const Commands &commands, byte verb, byte noun);
+
+	const AdlGameDescription *_gameDescription;
+	bool _isRestarting, _isRestoring;
+	byte _saveVerb, _saveNoun, _restoreVerb, _restoreNoun;
+	bool _canSaveNow, _canRestoreNow;
+};
+
+AdlEngine *HiRes1Engine__create(OSystem *syst, const AdlGameDescription *gd);
```
lordhoto: This looks undefined only forward declared?

engines/adl/adl.cpp
```diff
@@ -0,0 +1,1038 @@
...
+			if (event.kbd.flags & Common::KBD_CTRL) {
+				if (event.kbd.keycode == Common::KEYCODE_q)
+					g_engine->quitGame();
+				continue;
+			}
+
+			switch (event.kbd.keycode) {
+			case Common::KEYCODE_BACKSPACE:
+			case Common::KEYCODE_RETURN:
+				key = convertKey(event.kbd.keycode);
```
lordhoto: I general I *strongly* suggest never ever to mix key codes and ascii values. `convertKey` looks like it works on ascii values not key codes. There is only some overlap. Simply converting the backspace and return keys directly here would be best.

engines/adl/detection.cpp
```diff
@@ -0,0 +1,247 @@
...
+	AD_EXTRA_GUI_OPTIONS_TERMINATOR
+};
+
+static const PlainGameDescriptor adlGames[] = {
+	{"hires1", "Hi-Res Adventure #1: Mystery House"},
+	{0, 0}
+};
+
+static const AdlGameDescription gameDescriptions[] = {
+	{ // Hi-Res Adventure #1: Mystery House - Apple II - 1987 PD release
```
lordhoto: We should flag the detection entry as unstable.

lordhoto:
The code itself seems mostly reasonably.

The character encoding and input handling is confusing to me, but that might be related to the time of the day. And I guess a displayChar might make sense too, there seem to be some places which convert a character into `Common::String` to display it.

About the game itself: I didn't really look into that, I'll try to do that tomorrow.

Strangerke:
@lordhoto Did you have a look at the games to confirm the games are in the scope?

lordhoto:
See, from the looks it's pre-point&click games. But honestly since some of these games (i.e. the pre-AGI ones) are already supported, it would seem odd to reject this...

Overall I'm still waiting for @wjp's input here though.

wjp:
No objections from me.

sev-:
Fine, merging then

PR #701, opened 2016-03-09
fix: error: format not a string literal and no format arguments [-Wer…

ckorn:
…ror=format-security]


lordhoto:
Please fix the commit message to follow our http://wiki.scummvm.org/index.php/Commit_Guidelines. Thank you.

PR #702, opened 2016-03-09
SHERLOCK: Add format to fix compile error

ckorn:
This patch fixes the compile error:
format not a string literal and no
format arguments [-Werror=format-security]

Just suply "%s" as the standard format.
This way the string is not interpreted as format
which may lead to security issues.


lordhoto:
Thanks.

PR #704, opened 2016-03-13
Implement Custom Audio CD Support for Win32, Linux, and Mac OS X.

lordhoto:
Allmost all of the work as been done by @clone2727 who also helped out bringing this up to speed for this PR. Thanks and major props to him.

The title really says it all. This implements custom audio CD support and stops relying on SDL for these platforms. This assures that even for SDL2 we do have audio CD support on those platforms. 

Pros:
- SDL2 builds now finally have audio CD support on the mentioned platforms. (This has been a major blocker for @sev- for making SDL2 the default.)
- Linux audio CD support is finally usable. SDL 1.2 relied on the audio drive to be connected to the sound card via an analog cable. Now we digitally play back the sound data.
- Cine, SCI, TeenAgent, and Kyra have been fixed to actually work with playing from real CD.
- We auto detect the drive now when the game is actually played from CD.
- The "cdrom" configuration parameter can now be used to specify a device or path too.
- Clean up of the AudioCDManager API to expose a unified API for real CD playback and emulation mode.

Cons:
- Win32 code is WinNT+ only because it relies on WinNT ioctl calls AFAICT. We are already discussing dropping Win95/Win98/WinME support on -devel. If someone, who raises his voice to keep Win95/Win98/WinME support, wants to fix this, any patch is welcomed.


sev-:
Regarding the Cons. I think it was mentioned that Win95/Me should stay on SDL1 anyway, thus this is not relevant. However, for still keeping it clean, this needs to be made compile-time switchable. Is it possible to add some nice #ifdefs around?

lordhoto:
> Is it possible to add some nice #ifdefs around?

It is. I think some people on -devel mentioned they wanted to jump in to maintain Win95/98/ME support. This looks like a nice task to get more familiar with our Win32 specific code.

sev-:
Hm, I had an impression that SDL2 is plainly not compatible with Win95. Thus, it makes porting of this code to that OS practically pointless, as SDL1 CD-Audio capabilities are in place. Or am I missing something?

lordhoto:
SDL2 is only working on WinXP+, yeah.
The new code has the advantage that it automatically picks the CD drive you play the game from (if you play from CD), and it allows easy configuration via drive letter. Both are not supported with SDL 1.2 because SDL 1.2 doesn't have any API for this. Thus, it makes sense to use it on WinXP even with SDL 1.2.
So, ideally switching would be done via a configure option (which works in a way that `create_project` does not need to be adapted, or we also adapt `create_project`). Since even our Win32 porter thinks about dropping pre-WinXP support, I don't really see what would justify the effort. Thus, I suggest if someone is feeling strong about this, they can just provide us with a patch in the future.

backends/audiocd/macosx/macosx-audiocd.cpp
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	for (uint32 i = 0; i < children.size(); i++) {
+		if (!children[i].isDirectory()) {
+			Common::String fileName = children[i].getName();
+
+			if (fileName.hasSuffix(".aiff") || fileName.hasSuffix(".cdda")) {
+				uint trackID = 0, j = 0;
+
+				for (; j < fileName.size() && !Common::isDigit(fileName[j]); j++)
```
csnover: This looks wrong

backends/audiocd/sdl/sdl-audiocd.cpp
```diff
@@ -67,44 +73,69 @@ bool SdlAudioCDManager::openCD(int drive) {
...
-	if (!_cdrom)
-		return;
+	// If we're set to only emulate, or have no CD, return here
+	if (onlyEmulate || !_cdrom)
+		return false;
 
+	if (!numLoops && !startFrame)
+		return false;
+
+	// FIXME: Explain this.
```
csnover: Please do :)

backends/audiocd/win32/win32-audiocd.cpp
```diff
@@ -0,0 +1,388 @@
...
+	// Prefer emulation
+	if (DefaultAudioCDManager::play(track, numLoops, startFrame, duration, onlyEmulate))
+		return true;
+
+	// If we're set to only emulate, or have no CD drive, return here
+	if (onlyEmulate || _driveHandle == INVALID_HANDLE_VALUE)
+		return false;
+
+	// HACK: For now, just assume that track number is right
+	// That only works because ScummVM uses the wrong track number anyway
```
csnover: There seems to be a check on the track number below, is this comment still accurate?

backends/audiocd/win32/win32-audiocd.cpp
```diff
@@ -0,0 +1,388 @@
...
+	_lastTrack = tocData.LastTrack;
+#if 0
+	debug("First Track: %d", tocData.FirstTrack);
+	debug("Last Track: %d", tocData.LastTrack);
+#endif
+
+	for (uint32 i = 0; i < (bytesReturned - 4) / sizeof(TRACK_DATA); i++)
+		_tocEntries.push_back(tocData.TrackData[i]);
+
+#if 0
```
csnover: Is there still value in keeping this dead code around?

backends/audiocd/macosx/macosx-audiocd.cpp
lordhoto: This loop skips over non digits (note the `!`, which is not present in the one below), the one below only goes over digits.

backends/audiocd/sdl/sdl-audiocd.cpp
clone2727: Endy added the line years ago, but could not remember why when questioned. Perhaps you'd like to have a shot at explaining it.

backends/audiocd/win32/win32-audiocd.cpp
clone2727: Yes. ScummVM still uses broken track numbering.

backends/audiocd/win32/win32-audiocd.cpp
clone2727: For debugging?

backends/audiocd/macosx/macosx-audiocd.cpp
csnover: Aha, yes, I see! This may be one of those very rare cases where it makes sense to actually comment what the code is doing, so dumb people like me don’t gloss over the one-character difference. But it’s up to you, and I could change my mind tomorrow when I’m not drunk-tired and trying to be “helpful” reviewing unfamiliar stuff :)

lordhoto: The loop below should be replaced by something like `strtol` anyway...

Kirben:
Completely dropping Windows 9x/ME support is fine, as it requires very old hardware, which I really doubt people are still using. It isn't clear if Windows 9x/ME is even still working, based on recent response from Simon Sawatzki on ScummVM devel mail list.

The updated Audio CD support seems fine on Windows 7, when briefly testing with Future Wars.

lordhoto:
Great. I'll merge this then. If somebody really wants Win95 support back, we can still fix this till 1.9.0 anyway.

PR #706, opened 2016-03-14
BBVS: Move much of engine initialization to newGame()

eriktorbjorn:
This should fix bug #7057 ("BBVS: No clean-up when restarting"). However, I'm not familiar with the game so I haven't done much testing other than using the savegame from that bug report and checked that my inventory is empty when starting the game anew.

If this is the correct fix, it may also be a good candidate for the 1.8 branch.


sev-:
Yes, this is the proper thing to do. Thank you!

PR #707, opened 2016-03-15
BUILD: prefer SDL2 by default

onlyjob:
Search for "sdl2-config" before looking for "sdl-config".


lordhoto:
SDL2 support stills is considered experimental. We will enable it when we think it's ready. Thank you for your effort though.

onlyjob:
OK thanks. FYI there is a report to suggest that SDL2 already works better on some architectures:

  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=807710

I briefly tested ScummVM-1.8.0 built with SDL2 on Debian/x86_64 and found it working well.

See also https://wiki.libsdl.org/MigrationGuide

onlyjob:
Is there a bug to track progress of SDL2 adaptation?

lordhoto:
We already spent a considerable amount of time to support SDL2. We will enable it in the near future. There is no place where any progress is tracked. It's not really a bug report anyway, it would be more of a feature request.

onlyjob:
Well, I meant _ticket_ not "bug" OK? ;)

a-detiste:
Sorry for the confusion, I'm responsible for this Debian bug :sweat: 

So here's a new Debian-only patch:

Maybe always pulling SDL1 + SDL2 & then doing this would be OK,
I hope 1.9.0 will be SDL2 all the way. (Wayland !)

-_sdlconfig=sdl-config
+_sdlconfig=debian/scripts/my-sdl-config

```
#!/bin/sh
if dpkg-vendor --is RaspBian then
    exec sdl2-config $@
else
    exec sdl-config $@
fi
```

PR #708, opened 2016-03-15
COMMON: [GSoC] Fix wrong warning message

vandalo:
Changed the err control of the game path refers

We were calling "err = Common::kPathNotDirectory;"
Changed in to:
	// Verify that the game path refers to an actual directory
    if (!dir.isDirectory())
        err = Common::kPathNotDirectory;
    else if(!dir.exists())
        err = Common::kPathDoesNotExist;

https://sourceforge.net/p/scummvm/bugs/6765/

### DIFF
--- a/engines/sci/graphics/celobj32.cpp
+++ b/engines/sci/graphics/celobj32.cpp
@@ -109,25 +109,42 @@ void CelObj::deinit() {
 
 template<bool FLIP, typename READER>
 struct SCALER_NoScale {
+#ifndef NDEBUG
+	const byte *_rowEdge;
+#endif
 	const byte *_row;
 	READER _reader;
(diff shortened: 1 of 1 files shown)


criezy:
Is there a reason why you are checking isDirectory() before exists()?

I wonder what isDirectory() returns for path that do not exists. The documentation is not clear on this. So to me it seems safer and more logical to first test that the path exist. What do you think?

criezy:
Sounds good now, except that only the first commit message follows our guideline.

I would recommend that you squash all the commits into one and make sure at the same time that you follow our guideline for the commit message.

vandalo:
I closed it, I think I made squash to much commits :(

PR #709, opened 2016-03-15
I18N: Chinese Pinyin GUI Translation added, support ISO-8859-1

lichenbo:
I planned to support Chinese Translation of GUI while the current system does not support 2-bytes characters. So instead I translate GUI into the Romanized Chinese with basic Latin alphabet. I've tested it with my compilation of Scummvm on OSX. It would be a help to further real Chinese characters translation when 2-bytes support is done, which is what I really want to work for as GSoC. Hope it would help.


criezy:
Thank you, that looks good.
Before I merge this, I would like to confirm with you that you will stay around to update the translation when we are preparing new releases (or more often if you want to). Providing a translation is not a one off contribution as some small updates need to be done from time to time.

Also in general I would prefer if you don't include the translations.dat into the same commit as the translation update. This make it a bit easier to merge or cherry-pick the commit into a release branch (as the binary translations.dat file might be different and need to be regenerated rather than merged).

lichenbo:
Yeah, I'd like to update the translation constantly. Since it's my first language, I could do it quickly. And also do I need to resubmit a new request for a pofile-only version of commit? Because I didn't find anyway to remove the file to this pull request. Thank you.

criezy:
If the issue is about removing the file from the commit, there are several ways to do that (for example I thing "git reset HEAD^ -- gui/themes/translations.dat" followed by "git commit --amend" should work if you are using command line git).

If the issue is about updating the pull request once you have done the change locally, you just have to force push to your fork on GitHub.

Note: I see that you made the changes in the master branch in your fork. I would recommend that you use a separate branch for each pull request that you do instead of using master (e.g. create a branch, do modifications in that branch, push to your fork, open pull request). That way you can keep your master branch clean and easily pull changes from our repository.

lichenbo:
Thank you, I will close this thread and make a new request. New stuff learned! :-)

PR #710, opened 2016-03-15
COMMON: [GSoC] Fix wrong warning message

vandalo:
The if was wrong, we split in two the validation of the path, first we check if it's a existing path and then if the path is a directory

Bug:
https://sourceforge.net/p/scummvm/bugs/6765/

### DIFF
--- a/base/main.cpp
+++ b/base/main.cpp
@@ -148,8 +148,11 @@ static Common::Error runGame(const EnginePlugin *plugin, OSystem &system, const 
 #endif
 
 	// Verify that the game path refers to an actual directory
-	if (!(dir.exists() && dir.isDirectory()))
+        if (!dir.exists()) {
+		err = Common::kPathDoesNotExist;
+        } else if (!dir.isDirectory()) {
 		err = Common::kPathNotDirectory;
+        }
 
 	// Create the game engine
 	if (err.getCode() == Common::kNoError) {


lordhoto:
Please apply manual line breaks in the commit message as suggested by our commit guidelines.
Also, please remove the GSoC tag from the commit message. This is only relevant for the title of the pull request.

base/main.cpp
```diff
@@ -148,7 +148,9 @@ static Common::Error runGame(const EnginePlugin *plugin, OSystem &system, const
 #endif
 
 	// Verify that the game path refers to an actual directory
-	if (!(dir.exists() && dir.isDirectory()))
+	if(!dir.exists())
```
lordhoto: There's a space missing between 'if' and '('.

criezy:
The commit content looks good. But the commit message could be improved.

It would be good to have a slightly more precise and coherent commit message. Is it a 'warning message' (as indicated on the first line) or an 'error message' (as indicated on the next line)? What error/warning is this? In what context?

Also we are going to move to a different bug tracker so maybe having a link to the bug tracker in the commit message is not that useful, Usually we just specify the bug id and title (which then allow us to quickly find the bug in the bug tracker - wherever that bug tracker is). For example:
This fixes bug #6765 - Wrong error code if directory missing

vandalo:
message commit changed

lordhoto:
The commit message still talks about a warning message even though the code suggests it is an error message?

base/main.cpp
```diff
@@ -148,7 +148,9 @@ static Common::Error runGame(const EnginePlugin *plugin, OSystem &system, const
 #endif
 
 	// Verify that the game path refers to an actual directory
-	if (!(dir.exists() && dir.isDirectory()))
+	if (!dir.exists())
```
lordhoto: Please add {} even for one line if statements.

lordhoto:
The commit message still reads "warning" instead of "error". Can you fix this and implement the changes I suggested? Then this should be good to go.

vandalo:
sorry for late fix, I have had exams and projects at university. I hope now it's all done!

vandalo:
I dont' know why when I made the commit, I have lost my indentation, I'll fix it today

vandalo:
Solved identation problems too, I hope now is all right

sev-:
Thanks. Merging.

PR #712, opened 2016-03-17
I18N: Chinese Pinyin GUI Translation added, support ISO-8859-1

lichenbo:
I planned to support Chinese Translation of GUI while the current system does not support 2-bytes characters. So instead I translate GUI into the Romanized Chinese with basic Latin alphabet. I've tested it with my compilation of Scummvm on OSX. It would be a help to further real Chinese characters translation when 2-bytes support is done, which is what I really want to work for as GSoC. Hope it would help.

### DIFF
--- a/po/zh_pinyin.po
+++ b/po/zh_pinyin.po
@@ -1,0 +1,3730 @@
+# LANGUAGE translation for ScummVM.
+# Copyright (C) YEAR ScummVM Team
+# This file is distributed under the same license as the ScummVM package.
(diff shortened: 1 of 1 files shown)


lichenbo:
This request is basically the same with #709 , except the binary .dat file is excluded and would be merged with the branch other than master. Thanks to @criezy for guidance.

criezy:
Thank you. I have added this to our repository and it will be present in the next release.

Note: I renamed the po file to zh-Latn_CN.po. Apparently there is no standard ISO language and country codes for Pinyin but common usage seems to be that one.

PR #713, opened 2016-03-19
BASE: [GSoC] Add --auto-detect command

iskrich:
Allows to display games in the current directory that are compatible with ScummVM.

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -757,12 +757,16 @@ static void listAudioDevices() {
 }
 
 /** Display all games in current directory */
-static void autoDetect(){
+static void autoDetect() {
+
 	//Current directory
 	Common::FSNode dir(".");
+
 	Common::FSList files;
+
 	//Collect all files from directory
 	dir.getChildren(files, Common::FSNode::kListAll);
+
 	GameList candidates(EngineMan.detectGames(files));
 	if (candidates.empty()) {
 		printf("%s\n","ScummVM could not find any game in the current directory" );
@@ -1000,7 +1004,7 @@ bool processSettings(Common::String &command, Common::StringMap &settings, Commo
 	} else if (command == "help") {
 		printf(HELP_STRING, s_appName);
 		return true;
-	} else if (command == "auto-detect"){
+	} else if (command == "auto-detect") {
 		autoDetect();
 		return true;
 	}
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -68,6 +68,7 @@ static const char HELP_STRING[] =
 	"  -z, --list-games         Display list of supported games and exit\n"
 	"  -t, --list-targets       Display list of configured targets and exit\n"
 	"  --list-saves=TARGET      Display a list of saved games for the game (TARGET) specified\n"
+	"  --auto-detect            Display a list of games from current directory\n"
 #if defined(WIN32) && !defined(_WIN32_WCE) && !defined(__SYMBIAN32__)
 	"  --console                Enable the console window (default:enabled)\n"
 #endif
@@ -392,6 +393,9 @@ Common::String parseCommandLine(Common::StringMap &settings, int argc, const cha
 			END_COMMAND
 
 			DO_COMMAND('z', "list-games")
+			END_COMMAND
+
+			DO_LONG_COMMAND("auto-detect")
 			END_COMMAND
 
 #ifdef DETECTOR_TESTING_HACK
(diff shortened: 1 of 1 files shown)


base/commandLine.cpp
```diff
@@ -752,6 +756,24 @@ static void listAudioDevices() {
 	}
 }
 
+/** Display all games in current directory */
+static void autoDetect(){
```
bluegr: Formatting (space between ) and { )

base/commandLine.cpp
```diff
@@ -752,6 +756,24 @@ static void listAudioDevices() {
...
+	//Current directory
```
bluegr: Formatting

base/commandLine.cpp
```diff
@@ -752,6 +756,24 @@ static void listAudioDevices() {
...
+	Common::FSNode dir(".");
+	Common::FSList files;
+	//Collect all files from directory
```
bluegr: Formatting

base/commandLine.cpp
```diff
@@ -978,6 +1000,9 @@ bool processSettings(Common::String &command, Common::StringMap &settings, Commo
 	} else if (command == "help") {
 		printf(HELP_STRING, s_appName);
 		return true;
+	} else if (command == "auto-detect"){
```
bluegr: Formatting

base/commandLine.cpp
iskrich: Could you explain about comments, code to which the comment must allocated by blank lines?

csnover: @iskrich The code formatting guidelines for ScummVM are at http://wiki.scummvm.org/index.php/Code_Formatting_Conventions.

lordhoto:
What exactly is the use case for detecting games inside the current directory? Why would you want to do that, what do you gain from that?

a-detiste:
I think this project (Steam-less Steam-like launcher) needed this: https://github.com/ssokolow/game_launcher

@ssokolow

a-detiste:
This seems pretty lightweight, so why not ?

iskrich:
A finite aim was to implement a launch games without GUI.
Using simple call scummvm, but then I realized that this is not a trivial task....
Specifically, this part will help the user to understand whether the game will be able to start, without causing GUI...

lordhoto:
Looks good then.

Can we make this work in conjunction with the --path/-p flag to look in some other directory?

ssokolow:
@a-detiste 
> I think this project (Steam-less Steam-like launcher) needed this: https://github.com/ssokolow/game_launcher

Yeah. Once I no longer have my degree project sapping all of my time, I want to get as close as possible to "Tell the launcher where you keep your games and it does the rest, right down to offering a list of saved games in the context menu if the backend provides an equivalent to `scummvm --list-saves=tentacle` and `scummvm --save-slot=1 tentacle`

iskrich:
In TODO  http://wiki.scummvm.org/index.php/TODO#Common_code.2C_infrastructure
provides the use specified path too.
I can look at the implementation of this flag

lordhoto:
> I can look at the implementation of this flag

Any news on this?

iskrich:
I looked at the other command-line commands. Adding path looks like a value assignment, for example: `--extrapath = PATH`,  `--soundfont = FILE.`
Add a secondary flag for autodetect will look like a violation of the style of these commands.
Alternatively, I can add `--autodetect=PATH`. But then there will be problems with auto-detection of current directory.

iskrich:
I looked at the other command-line commands. Adding path looks like a value assignment, for example:` --extrapath = PATH, --soundfont = FILE.`
Add a secondary flag will look like a violation of the style of these commands.
Alternatively, I can add it. But then there will be problems with auto-detection of the current directory

lordhoto:
There is a bit of a choice point here. The other optons you cite affect the configuration variables directly. Like `-p/--path` does. Which is what you need to pass when you run a game without it being in the config.

iskrich:
I still do not fully understand. Is it true that now is no command in console with an additional flag?
Commands that work with assignment, really work with the object settings. Which then sends the results of parsing to conf. All these are defined as options. 
If I want to use the command (as I understand commands do not affect to conf)  `--autodetect=PATH` I should be determined new define like 
`DO_LONG_COMMAND_OPT` and `DO_COMMAND_OPT`,  is it a good decision?

lordhoto:
> Is it true that now is no command in console with an additional flag?

There are some with the assignment style syntax. There are two things to consider here, if you run a game not added to the config (which is probably why you want to use --auto-detect in the first place), then you might have to specify a path via -p/--path (except you go for cwd) later on. So going for the same style here is something which can be considered consistency. Of course, then this is inconsistent with other options which take arguments they work on directly.

sev-:
@iskrich When do you plan to add -p option support to this patch?

criezy:
Thank you for starting the work. I have added the missing bit to use the path specified with -p/--path and merged everything in master. Further work would be to be able to start the detected game rather than just print detected games. But for now this will do.

PR #714, opened 2016-03-21
WINTERMUTE: Don't mix different path separators.

RyperHUN:
- Buf fix #7068 : https://sourceforge.net/p/scummvm/bugs/7068/
- Replacing '\' with '/' in filename - getNodeRelativePath()

The assertion happened when the getNodeRelativePath(); function was called, with a string containing '\' char and '/' at the same time. ( /COM0024/secnes\allotment\jonas_dia_04b\jonas_dia_04b\jonas_dia_4b.scene )
As you can see I just added a conversion which replaces the car '/' with '\' in the right place.

### DIFF
--- a/engines/wintermute/base/file/base_disk_file.cpp
+++ b/engines/wintermute/base/file/base_disk_file.cpp
@@ -150,7 +150,14 @@ Common::SeekableReadStream *openDiskFile(const Common::String &filename) {
 	}
 	// File wasn't found in SearchMan, try to parse the path as a relative path.
 	if (!file) {
-		Common::FSNode searchNode = getNodeForRelativePath(filename);
+		Common::String filenameBackSlash = filename;
+		for (size_t i = 0; i < filenameBackSlash.size(); i++) {
+			if (filenameBackSlash[i] == '/') {
+				filenameBackSlash.setChar('\\', i);
+			}
+		}
+
+		Common::FSNode searchNode = getNodeForRelativePath(filenameBackSlash);
 		if (searchNode.exists() && !searchNode.isDirectory() && searchNode.isReadable()) {
 			file = searchNode.createReadStream();
 		}


engines/wintermute/base/file/base_disk_file.cpp
```diff
@@ -150,7 +150,14 @@ Common::SeekableReadStream *openDiskFile(const Common::String &filename) {
 	}
 	// File wasn't found in SearchMan, try to parse the path as a relative path.
 	if (!file) {
-		Common::FSNode searchNode = getNodeForRelativePath(filename);
+		Common::String fileNameRIghtSlashes = filename;
+		for (size_t i = 0; i < fileNameRIghtSlashes.size(); i++) {
```
somaen: would you mind correcting that broken camel-casing in "RIght"?

somaen:
The commit message on this one does not follow our commit guidelines:

I don't have a quick reference around, but I think the notation commonly used for bug references is "Fixes #bugid" at the end of the Commit-message, using the space before that to say what was DONE.

engines/wintermute/base/file/base_disk_file.cpp
```diff
@@ -150,7 +150,14 @@ Common::SeekableReadStream *openDiskFile(const Common::String &filename) {
 	}
 	// File wasn't found in SearchMan, try to parse the path as a relative path.
 	if (!file) {
-		Common::FSNode searchNode = getNodeForRelativePath(filename);
+		Common::String fileNameRIghtSlashes = filename;
```
somaen: This is probably better named filenameBackSlash, as the "right" is ambiguous.

RyperHUN: Yeah your right, I corrected it.

somaen:
Both of these commits still don't follow the commit message guidelines I mentioned above.

Also, feel free to squash the two commits into a single one.

RyperHUN:
Fixed the naming issue, and I squashed the two commits together.

tobiatesan:
Um, without requiring posterity to git blame/git log to understand what's going on, would a comment like `"// HACK: this is because some scripts have spurious '/' chars"`, in the spirit of the one at line 116, help?

criezy:
Without knowing the details of the code and only having had a quick look at the file, rather than replace '/' with '\\' would it make sense to use fileFixedName instead of filename in the call to getNodeForRelativePath (and use '/' instead of '\\' in the tokenizer in getNodeForRelativePath)? We already fixed slashes ones in the function (albeit to use '/' everywhere and not '\\' everywhere) and it seems like it might be good to benefit from it.

A potential issue is that getNodeForRelativePath is also used in diskFileExists(), but the implementation there looks wrong anyway (it uses the same file name for SearchMan and for getNodeForRelativePath while it looks like the former expect the use of '/' and the latter currently expect '\\' - or at least that is what openDiskFile assumes).

tobiatesan:
Would this fit in PathUtil?

RyperHUN:
Tobiatesan: Should I put it there? I can create a new function for this in path util, but i think it is not neccesary. 

Criezy: fixedFilename is a path with forward slashes:
image
I don't know exactly what is happening in all of the functions which references to this method, but my method solves the mixing separator problem.

sev-:

bluegr:
This looks to be quite a simple and straightforward change, and it's been bitrotting for months now... is there anything blocking its addition to master?

tobiatesan:
@bluegr: *probably* I *guess* it would be a wise-ish idea to extract the relevant piece of logic, call it "normalizeSlashes" or something and then maybe integrate it with PathUtils? I'll get there myself in a minute.

bluegr:
@tobiatesan: sounds fine. You can cherry pick HyperHUN's initial commit from this PR into your branch, and then perform the changes you want on it. That way, all further work on this can continue in PR #733

tobiatesan:
@bluegr ah-hum, apparently my branch already has a normalizeFilename() for this very purpose I forgot I had in there.

I'll see if the bug this branch set out to fix is already solved there.

tobiatesan:
@bluegr I opened a separate PR #874 which includes this commit + my branch from #733.

sev-:
This is now obsoleted by #874

PR #715, opened 2016-03-22
GUI: Remove dummy FluidSynth Settings button from Edit Game

dudola:
Remove dummy FluidSynth Settings button from Edit Game.
The FluidSynth Settings button actually works only from the Options dialog.
https://sourceforge.net/p/scummvm/bugs/6821/

### DIFF
--- a/gui/options.cpp
+++ b/gui/options.cpp
@@ -868,10 +868,6 @@ void OptionsDialog::addMIDIControls(GuiObject *boss, const Common::String &prefi
 	_midiGainSlider->setMaxValue(1000);
 	_midiGainLabel = new StaticTextWidget(boss, prefix + "mcMidiGainLabel", "1.00");
 
-#ifdef USE_FLUIDSYNTH
-	new ButtonWidget(boss, prefix + "mcFluidSynthSettings", _("FluidSynth Settings"), 0, kFluidSynthSettingsCmd);
-#endif
-
 	_enableMIDISettings = true;
 }
 
@@ -1108,6 +1104,10 @@ GlobalOptionsDialog::GlobalOptionsDialog()
 	//
 	_midiTabId = tab->addTab(_("MIDI"));
 	addMIDIControls(tab, "GlobalOptions_MIDI.");
+
+#ifdef USE_FLUIDSYNTH
+	new ButtonWidget(tab, "GlobalOptions_MIDI.mcFluidSynthSettings", _("FluidSynth Settings"), 0, kFluidSynthSettingsCmd);
+#endif
 
 	//
 	// 4) The MT-32 tab


lordhoto:
Is there any special reason why it doesn't work from the Options dialog opened via Edit Game? Would it be possible to fix it, if that's sensible?

gui/options.cpp
```diff
@@ -869,6 +869,7 @@ void OptionsDialog::addMIDIControls(GuiObject *boss, const Common::String &prefi
 	_midiGainLabel = new StaticTextWidget(boss, prefix + "mcMidiGainLabel", "1.00");
 
 #ifdef USE_FLUIDSYNTH
+	if(prefix=="GlobalOptions_MIDI.")
```
sev-: This has incorrect code formatting and indentation on the following line.

dudola:
From https://sourceforge.net/p/scummvm/bugs/6821/ it is unclear if this was ever intended as a per-game setting or not . So should i fix it so that it works from the Edit Game also ?

lordhoto:
Well, there are two options you have:
1) remove it from the edit game dialog
2) make it work in the edit game dialog

If you want to go with 1) please clean up the design a bit. `OptionsDialog::addMIDIControls` should really only add widgets which are used for both dialogs, the global one and the game specific one. Thus, if only the global one uses it, there should be no code to add the widget here.

gui/options.cpp
```diff
@@ -1108,6 +1104,10 @@ GlobalOptionsDialog::GlobalOptionsDialog()
 	//
 	_midiTabId = tab->addTab(_("MIDI"));
 	addMIDIControls(tab, "GlobalOptions_MIDI.");
+
```
lordhoto: Please, remove trailing whitespaces (this includes tabs) here.

gui/options.cpp
```diff
@@ -1108,6 +1104,10 @@ GlobalOptionsDialog::GlobalOptionsDialog()
 	//
 	_midiTabId = tab->addTab(_("MIDI"));
 	addMIDIControls(tab, "GlobalOptions_MIDI.");
+	
+	#ifdef USE_FLUIDSYNTH
```
lordhoto: We usually don't indent ifdef checks in situations like these. Please adapt it.

lordhoto:
Looks good for the removal of the button now. Can you please fix the minor issues I pointed out and squash all together into one commit?

dudola:
Yeah sure , I will do this ASAP .

lordhoto:
There are still a lot of commits in this PR and one is even some odd merge commit. Please adapt the PR that only a single commit with your final changes is present.

PR #716, opened 2016-03-23
GUI: Fix timer's conflict

ottogin:
This patch fixes tooltip disapearing.
Try pressing the shift key and hovering over any button. The tooltip will appear and then disappear, it happens because pressing any button, closing the tooltip. As we do not distinguish any difference between the event of the button constantly being pressed and of it being pressed just now, during the next iteration processEvent tooltip closes. That's what I corrected by making a timer that monitors whether the tooltip has been created during the pressed down button (if so is not necessary to close), or whether the button was pressed after (then close)

### DIFF
--- a/gui/Tooltip.cpp
+++ b/gui/Tooltip.cpp
@@ -31,8 +31,8 @@
 namespace GUI {
 
 
-Tooltip::Tooltip() :
-	Dialog(-1, -1, -1, -1), _maxWidth(-1) {
+Tooltip::Tooltip(uint32 bornTime) :
+	Dialog(-1, -1, -1, -1), _maxWidth(-1), _bornTime(bornTime), _firstKeyDownTimeSetted(false){
 
 	_backgroundType = GUI::ThemeEngine::kDialogBackgroundTooltip;
 }
--- a/gui/Tooltip.h
+++ b/gui/Tooltip.h
@@ -27,6 +27,8 @@
 #include "common/str-array.h"
 #include "gui/dialog.h"
 
+#include "common/system.h"
+
 namespace GUI {
 
 class Widget;
@@ -34,9 +36,15 @@ class Widget;
 class Tooltip : public Dialog {
 private:
 	Dialog *_parent;
+	uint32 _bornTime;
+	uint32 _firstKeyDownTime;
+	bool _firstKeyDownTimeSetted;
+	enum{
+		kListenerDelay = 200
+	};
 
 public:
-	Tooltip();
+	Tooltip(uint32 bornTime);
 
 	void setup(Dialog *parent, Widget *widget, int x, int y);
 
@@ -55,7 +63,14 @@ protected:
 		_parent->handleMouseWheel(x + (getAbsX() - _parent->getAbsX()), y + (getAbsX() - _parent->getAbsX()), direction);
 	}
 	virtual void handleKeyDown(Common::KeyState state) {
-		close();
+		// Check if current Tooltip was born after first key down
+		// If it isn't - close Tooltip
+		if (!_firstKeyDownTimeSetted) {
+			_firstKeyDownTimeSetted = true;
+			_firstKeyDownTime = g_system->getMillis(true);
+		}
+		if (_firstKeyDownTime - _bornTime > kListenerDelay)
+			close();
 		_parent->handleKeyDown(state);
 	}
 	virtual void handleKeyUp(Common::KeyState state) {
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -351,7 +351,7 @@ void GuiManager::runLoop() {
 		if (tooltipCheck && _lastMousePosition.time + kTooltipDelay < _system->getMillis(true)) {
 			Widget *wdg = activeDialog->findWidget(_lastMousePosition.x, _lastMousePosition.y);
 			if (wdg && wdg->hasTooltip() && !(wdg->getFlags() & WIDGET_PRESSED)) {
-				Tooltip *tooltip = new Tooltip();
+				Tooltip *tooltip = new Tooltip(_system->getMillis(true));
 				tooltip->setup(activeDialog, wdg, _lastMousePosition.x, _lastMousePosition.y);
 				tooltip->runModal();
 				delete tooltip;


lordhoto:
Are you sure this is actually fixing the cause of bug #6841?

AFAICT the issue is that if you replicate the steps described in the bug report the "key up" event for the shift key is eat up, and thus the LauncherDialog is never notified properly.

How does your attempted fix relate to this?

ottogin:
lordhoto, Currently the description of the bug does not match with reality. If you follow all the steps in the bug list, you will not see any result. I.e. someone had tried to fix it, but in doing so has created another bug: try pressing the shift key and hovering over any button. The tooltip will appear and then disappear, it happens because the # 6891 was being corrected by pressing any button, closing the tooltip. As we do not distinguish any difference between the event of the button constantly being pressed and of it being pressed just now, during the next iteration processEvent tooltip closes. That's what I corrected by making a timer that monitors whether the tooltip has been created during the pressed down button (if so is not necessary to close), or whether the button was pressed after (then close)

criezy:
The code looked confusing at first but your last comment explains I think (maybe adding a comment in the code would help). So you are approximating the 'key was pressed before showing the tooltip' by we receive a 'key press event with 200ms of the tooltip creation'? This works because we get continuous key press events while the key remains pressed and the next processEvent iteration occurs within 200ms of the tooltip being created. That seems a bit unreliable (some platforms may be slower than others).

Also why is firstKeyDownTime a class variable and why is it tested in all calls to handleKeyDown? It looks to me as if we only need to test it once (when _firstKeyDownTimeSetted is set from false to true).

lordhoto:
> it happens because the # 6891 was being corrected by pressing any button, closing the tooltip

#6891 seems to be a bug which is not related to the GUI at all (https://sourceforge.net/p/scummvm/bugs/6891/ shows me some bug in the Sherlock engine). It would be nice if you could give the correct bug number. Thank you.

> If you follow all the steps in the bug report, you will not see any result. I.e. someone had tried to fix it, but in doing so has created another bug

I can reproduce the bug with the given steps. It is a bit tricky to do because it requires very careful timing. If you use something like valgrind which slows down execution time, it is easier to reproduce.

> try pressing the shift key and hovering over any button. The tooltip will appear and then disappear

That seems fine behavior, because we want to close on key down. And we do have a key repeat, which makes the shift button get repeated after some point.

> That's what I corrected by making a timer that monitors whether the tooltip has been created during the pressed down button (if so is not necessary to close), or whether the button was pressed after (then close)

In fact, your patch will not close at all on any button down afterwards anymore because you only use the time on the first key down you receive. Any subsequent key down events will not use the time when they were triggered.
Steps to reproduce:
1) Press Shift
2) Hover Add Game Button
3) Wait for tooltip to appear (it will now stay on screen)
4) Press backspace
As a fun side effect this opens up the remove game dialog. If you now click "No", the dialog will close, but the tooltip will still be visible till you move the mouse again.

As I see the this particular bug, it is related to the flushing of events after event closing in https://github.com/scummvm/scummvm/blob/master/gui/gui-manager.cpp#L328-L335
What makes this bug appareant is if this loop catches both the repeated key down event and the key up event. In this case they key down will close the tooltip, because it got closed the key up event will be ignored. As a result the LauncherDialog will never get notified the SHIFT state might have changed and not adapt the button description.

With your patch you simply avoid closing the tooltip on key down (if it happens too fast) and thus effectively not making the event loop eat up the key up event. However, the event loop still has this fragile behavior about possible ignoring important events. It would seem much more natural to fix the event loop instead.

ottogin:
criezy, _firstKeyDownTime is a class variable, because we want to know when Tooltip does "catch" first key down event. It will be different for any Tooltips.
lordhoto, I made a mistake: bug #6841. 
Thanks for the explanation, But are you sure that this is fine behavior(tooltip disappearing)? Looks like it is not =) 
I would like to fix it, but I can't still reproduce bug, even using Valgrind

lordhoto:
> I would like to fix it, but I can't still reproduce bug, even using Valgrind

You still need to let go shift really fast after you start seeing the tooltip. It might require some tries. As I said, it's timing sensitive.

criezy:
@ottogin I didn't mean to make it a global variable shared by all tooltips, but to make it a local variable in the handleKeyDown function because you only need to use it once. You only need to know when the tooltip does "catch" the first key down event when that first key down events is actually caught. When further key down events are caught, if you know they are not the first one (and you know it because of the firstKeyDownTimeSetted variable) you don't need to again compare the _bornTime with the firstKeyDownTime  because neither of those have changed.

But that might be irrelevant. I was only looking at your code but I have not checked what the bug actually is or tried to reproduce it in ScummVM. So you should first try to resolve issues raised by lordhoto.

ottogin:
criezy, I am working on it currently. Thank you.

criezy:
I now had a quick look at the bug report and then successfully reproduced the issue described there (without using valgrind). To me it indeed looks like you actually fixed a different bug than the one described in bug #6841(the issue you described - pressing shift, then trying to display a tooltip, and the tooltip disappears immediately), which is fine. But the way you fix it indeed has issues (the one described by lordhoto in his previous message) as it doesn't correctly discriminate between repeated key down events and key down events for a new key press.

I am wondering if there would be a better way to discriminate between those than using a time based comparaison. It looks like Event::synthetic might have provided such a way but unfortunately somebody decided that the GuiManager only needed to pass the KeyState to widgets and not the full event on a key press or key down event...

fuzzie:
So would it make sense to refactor the event handlers so that they get the full event?

criezy:
I think so. That was indeed what I was coming at.

criezy:
Closing this as this is not a correct way to fix the issue.
PR #730 provides a cleaner fix (that actually works) but I decided to close it after some discussions as we think it would be better to make some more fundamental changes to the way tooltips are implemented, and in the process fix this and other tooltip bugs.

You are of course welcome to work on this (and I think you already discussed with sev about this) and open a new PR.

I am not closing the door to merging a temporary fix, that actually works and is quite self contained.But that excludes both the approach taken here and in PR #730.

Also note that there is now a ticker in our bug tracker referencing this issue:
#7108 GUI: When shift is pressed, tooltips disappear quickly without moving the mouse

PR #717, opened 2016-03-23
GUI: [GSoC] EditableWidget: Add maxLength and redo setEditString()

vampcat:
In EditableWidget,
- Add maxLength
- Add function isCharValid() to filter input
- Modified tryInsertChar() to act as length check
- Modified DomainEditTextWidget() in launcher.cpp
  to use isCharValid() as input filter.

### DIFF
--- a/gui/widgets/editable.cpp
+++ b/gui/widgets/editable.cpp
@@ -28,13 +28,13 @@
 
 namespace GUI {
 
-EditableWidget::EditableWidget(GuiObject *boss, int x, int y, int w, int h, const char *tooltip, uint32 cmd)
-	: Widget(boss, x, y, w, h, tooltip), CommandSender(boss), _cmd(cmd) {
+EditableWidget::EditableWidget(GuiObject *boss, int x, int y, int w, int h, const char *tooltip, uint32 cmd, uint maxLength)
+	: Widget(boss, x, y, w, h, tooltip), CommandSender(boss), _cmd(cmd), _maxLength(maxLength) {
 	init();
 }
 
-EditableWidget::EditableWidget(GuiObject *boss, const String &name, const char *tooltip, uint32 cmd)
-	: Widget(boss, name, tooltip), CommandSender(boss), _cmd(cmd) {
+EditableWidget::EditableWidget(GuiObject *boss, const String &name, const char *tooltip, uint32 cmd, uint maxLength)
+	: Widget(boss, name, tooltip), CommandSender(boss), _cmd(cmd), _maxLength(maxLength) {
 	init();
 }
 
@@ -63,18 +63,31 @@ void EditableWidget::reflowLayout() {
 }
 
 void EditableWidget::setEditString(const String &str) {
-	// TODO: We probably should filter the input string here,
-	// e.g. using tryInsertChar.
-	_editString = str;
+	_editString.clear();
+
+	// Copy string char by char, ensuring each char is valid
+	uint len = 0;
+	for (uint i = 0; i < str.size() && len < _maxLength; i++) {
+		if (isCharValid((byte)str[i])) {
+			_editString += str[i];
+			len++;
+		}
+	}
+
 	_caretPos = 0;
 }
 
 bool EditableWidget::tryInsertChar(byte c, int pos) {
-	if ((c >= 32 && c <= 127) || c >= 160) {
+	if (_maxLength == 0 || _editString.size() < _maxLength) {
 		_editString.insertChar(c, pos);
 		return true;
 	}
+
 	return false;
+}
+
+bool EditableWidget::isCharValid(byte c) const {
+	return (c >= 32 && c <= 127) || c >= 160;
 }
 
 void EditableWidget::handleTickle() {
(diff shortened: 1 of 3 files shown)


gui/widgets/editable.cpp
```diff
@@ -63,15 +63,21 @@ void EditableWidget::reflowLayout() {
 }
 
 void EditableWidget::setEditString(const String &str) {
-	// TODO: We probably should filter the input string here,
-	// e.g. using tryInsertChar.
-	_editString = str;
+	_editString = "";
+	int pos = 0;
+	for (unsigned int i = 0; i < str.size(); i++)
+		if (tryInsertChar((byte)str[i], pos))
```
sev-: inserting string char by char like this is extremely ineffective. Please make it in a single call.

gui/widgets/editable.cpp
```diff
@@ -63,15 +63,21 @@ void EditableWidget::reflowLayout() {
 }
 
 void EditableWidget::setEditString(const String &str) {
-	// TODO: We probably should filter the input string here,
-	// e.g. using tryInsertChar.
-	_editString = str;
+	_editString = "";
+	int pos = 0;
+	for (unsigned int i = 0; i < str.size(); i++)
```
sev-: we have special type for unsigned int. Please use that.

gui/widgets/editable.cpp
```diff
@@ -63,15 +63,21 @@ void EditableWidget::reflowLayout() {
...
 	_caretPos = 0;
 }
 
 bool EditableWidget::tryInsertChar(byte c, int pos) {
 	if ((c >= 32 && c <= 127) || c >= 160) {
-		_editString.insertChar(c, pos);
+		if (_maxLength < 0 || _editString.size() < (unsigned int)_maxLength)
+			_editString.insertChar(c, pos);
+		else
+			_caretPos--;
```
sev-: Why are you moving cursor back here?

sev-:
Your commit log message dos not conform to our guidelines. Please fix.

gui/widgets/editable.cpp
vampcat: The calling code increments the position of _caretPos on it's own, so this is to counter-act that. If I don't decrement it here, it will cause the caret to be incremented even when no character is typed.

gui/widgets/editable.cpp
```diff
@@ -28,13 +28,13 @@
...
-EditableWidget::EditableWidget(GuiObject *boss, int x, int y, int w, int h, const char *tooltip, uint32 cmd, int maxLength)
+EditableWidget::EditableWidget(GuiObject *boss, int x, int y, int w, int h, const char *tooltip, uint32 cmd, uint maxLength)
 	: Widget(boss, x, y, w, h, tooltip), CommandSender(boss), _cmd(cmd), _maxLength(maxLength) {
 	init();
 }
 
-EditableWidget::EditableWidget(GuiObject *boss, const String &name, const char *tooltip, uint32 cmd, int maxLength)
-	: Widget(boss, name, tooltip), CommandSender(boss), _cmd(cmd), _maxLength(maxLength) {
+EditableWidget::EditableWidget(GuiObject *boss, const String &name, const char *tooltip, uint32 cmd, uint maxLength)
+	: Widget(boss, name, tooltip), CommandSender(boss), _cmd(cmd), _maxLength(5) {
```
criezy: You are ignoring the maxLength passed to the constructor. That looks wrong.

gui/widgets/editable.cpp
criezy: The calling code increments the position of _caretPos only if tryInsertChar returns true. And I would not expect tryInsertChar to return true if the character was not inserted.

gui/widgets/editable.cpp
vampcat: Ahh, that was for debugging. That's for pointing it out.

lordhoto:
It looks like you now have commits, where the last three have an identical commit message. It think you should simply squash them all into one commit.

gui/widgets/editable.cpp
```diff
@@ -63,20 +63,32 @@ void EditableWidget::reflowLayout() {
...
+			_editString += str[i];
+			len++;
+		}
+
 	_caretPos = 0;
 }
 
 bool EditableWidget::tryInsertChar(byte c, int pos) {
-	if ((c >= 32 && c <= 127) || c >= 160) {
+	if (_maxLength == 0 || _editString.size() < _maxLength) {
```
lordhoto: We have at least one subclass of EditableWidget which overrides tryInsertChar to do custom filtering of input, it can be found over [here](https://github.com/scummvm/scummvm/blob/master/gui/launcher.cpp#L99-L104).

They way I see it this overriding will now break the length checking you added.

It is also not clear to me why you don't call `isCharValid` inside `tryInsertChar`. Which seems the obvious place to do it. Then you could also simply use `tryInsertChar` in `setEditString`.

gui/widgets/editable.cpp
```diff
@@ -63,20 +63,32 @@ void EditableWidget::reflowLayout() {
...
 		_editString.insertChar(c, pos);
 		return true;
 	}
+
 	return false;
 }
 
+bool EditableWidget::isCharValid(byte c) {
```
lordhoto: This is probably a good candiate for a method to be made constant because it should not modify internal state of the widget.

gui/widgets/editable.cpp
```diff
@@ -63,20 +63,32 @@ void EditableWidget::reflowLayout() {
 }
 
 void EditableWidget::setEditString(const String &str) {
-	// TODO: We probably should filter the input string here,
-	// e.g. using tryInsertChar.
-	_editString = str;
+	_editString = "";
```
lordhoto: There is a `clear()` function in `String` which should be used for these cases.

lordhoto:
Looking at this again, I think it makes sense to do two precisely separated commits: one for the setEditString rework and one for the maxLength add.

sev-:
@vampcat Could you please split it? Then we merge.

sev-:
OK, closing due to the absence of reaction from the original author.

PR #718, opened 2016-03-23
GUI: Improve handling of button presses while moving the cursor

salty-horse:
Submitting this patch for review. It's pretty simple, but I want more eyes on it, as it may affect things I'm unaware of. I've tested it on buttons, radio buttons, checkboxes, load-game dialog (picture buttons), and the AGI predictive text dialog.

The patch keeps track of where a mouse press started when deciding how mouse-over
and mouse-up should behave.

This handles the following situations:

1) If a mouse press starts outside a button (e.g. the UI background),
mouse-up inside the button has no effect. Previously, it triggered a
button click.
2) If a mouse press starts inside a button, the cursor moves outside of
its region, then back inside, the button will show as pressed.
Previously, it showed as highlighted instead of pressed.


lordhoto:
This looks reasonable from a glance. I did not test it though.

sev-:
Seems OK to me too.

salty-horse:
Merged as 031e062.

PR #721, opened 2016-03-25
SCI: Cleanup of Save/Load Code

lordhoto:
This removes an old hack in savegame.cpp which relied on template specialization to synchronize types. It uses much more natural function overloading instead, which is also nicer to distribute over multiple files, if desired.

Additionally, I cleaned up serialization of `SegmentObjTable::Entry` objects. The former approach required each individual `SegmentObjTale` subclass to specialize the serialization method for `Entry` and make sure `next_free` is serialized correctly. Now, only the contained type in `SegmentObjTable` needs to be serializable with `syncWithSerializer` and a generic serialization for `Entry` exists which takes care of `next_free`.
SegmentObjTable can now be easily used with fundamental types and pointers too.

I tested this manually by loading older KQ6CD saved games. I loaded various (~10) and then walked around a bit to see for obvious brokenness. I could not find any.

Additionally, I did some more involved tests using `cmp` to check for binary differences in written save files. Here, I altered `gamestate_delayedrestore` to immediately save again (with zeroed out meta data like time to avoid differences there). I did this to both my branch and master. With this change in place I loaded all my saved games (4 KQ5CD, 7 KQ6 German, 49 KQ6 CD) from command line multiple times each.
Then I scanned for differences in the generated save files with `cmp`. It turns out saved games created using this method have a slight chance of being different in all cases. This means there can be slight differences between master and my branch. But also between different runs from master (or different runs from my branch). I did not really figure out where this might come from, but since it happens in master too, I don't think this is a regression from my changes. Since it does not consistently happen it could also be just normal behavior due to timing differences.

### DIFF
--- a/engines/sci/console.cpp
+++ b/engines/sci/console.cpp
@@ -1861,7 +1861,7 @@ bool Console::cmdSavedBits(int argc, const char **argv) {
 
 	for (uint i = 0; i < entries.size(); ++i) {
 		uint16 offset = entries[i].getOffset();
-		const Hunk& h = hunks->_table[offset];
+		const Hunk& h = hunks->at(offset);
 		if (strcmp(h.type, "SaveBits()") == 0) {
 			byte* memoryPtr = (byte *)h.mem;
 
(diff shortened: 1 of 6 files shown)


wjp:
While I haven't tested this, I did look it over in detail, and the changes seem very nice.

bluegr:
I haven't tested this either, but the changes look great. Well done :) :+1:

tramboi:
Great change, overload resolution is definitely the right way to do it.

PR #722, opened 2016-03-25
GUI: [GSoC] Widget: redraw when setAlign() called

Abde-:
redraw when StaticTextWidget is aligned (same problem as setLabel, text drawn on top of the old one)


lordhoto:
The change looks fine. How did you test this?

sev-:
Good change. Merging.

PR #727, opened 2016-03-25
Added TransformTool Test Unit

CodinGree:
Simple Test Unit pull request for GSoC 2016.

### DIFF
--- a/test/graphics/transform_tools.h
+++ b/test/graphics/transform_tools.h
@@ -1,0 +1,33 @@
+#include <cxxtest/TestSuite.h>
+
+#include <math.h>
+#include "graphics/transform_tools.h"
+
+class TransformToolsSuite : public CxxTest::TestSuite {
+public:
+
+	void test_transformPoint(){
+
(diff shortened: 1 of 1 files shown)


test/graphics/transform_tools.h
```diff
@@ -0,0 +1,33 @@
...
+			x = (x * zoom.x) / kDefaultZoomX;
+			y = (y * zoom.y) / kDefaultZoomY;
+			x = x * cos(rotateRad) - y * sin(rotateRad);
+			y = x * sin(rotateRad) + y * cos(rotateRad);
+	
+			
+			y *= -1;
+
+			// X
+			TS_ASSERT_EQUALS(test.x,x);
```
lordhoto: Comparing floating point values with ``==`` is not necessarily a good idea. During calculation you can hit rounding etc. issues which can lead to value differences.

sev-:
Due to lack of feedback reaction from the submitter I'm closing this.

PR #728, opened 2016-03-27
WINTERMUTE: Correctly find .ogg version of .wav files

tobiatesan:
As it was, it didn't work because it turned
"some\\windows\\path.wav" into "some/system/pathogg" (not ".ogg"), so any file called by the scripts with its ".wav" extension but actually present in compressed .ogg form on the .dcp archive could not be found.

base_disk_file.cpp (apparently) takes care of preprocessing the path to make it system-agnostic already for those files which reside on the disk.

This one should fix https://sourceforge.net/p/scummvm/bugs/7088/

The corresponding bit in the original engine is:

```
	// try to switch WAV to OGG file (if available)
	_splitpath(Filename, drive, dir, fname, ext);
	if(_stricmp(ext, ".WAV")==0){
		_makepath(NewFile, drive, dir, fname, ".OGG");
		CBFile* file = Game->m_FileManager->OpenFile(NewFile);
		if(file){
			Filename = NewFile;
			Game->m_FileManager->CloseFile(file);
		}
	}

```

This way it should work reliably on all platforms like this, but just to be sure I would love it if somebody who runs Windows would check it before merging (I won't probably be able to do it before tomorrow).

### DIFF
--- a/engines/wintermute/base/sound/base_sound_manager.cpp
+++ b/engines/wintermute/base/sound/base_sound_manager.cpp
@@ -100,15 +100,14 @@ BaseSoundBuffer *BaseSoundMgr::addSound(const Common::String &filename, Audio::M
 	BaseSoundBuffer *sound;
 
 	Common::String useFilename = filename;
+	useFilename.toLowercase();
 	// try to switch WAV to OGG file (if available)
-	AnsiString ext = PathUtil::getExtension(filename);
-	if (StringUtil::compareNoCase(ext, "wav")) {
-		AnsiString path = PathUtil::getDirectoryName(filename);
-		AnsiString name = PathUtil::getFileNameWithoutExtension(filename);
-
-		AnsiString newFile = PathUtil::combine(path, name + "ogg");
-		if (BaseFileManager::getEngineInstance()->hasFile(newFile)) {
-			useFilename = newFile;
+	if (useFilename.hasSuffix(".wav")) {
+		Common::String oggFilename = useFilename;
+		oggFilename.erase(oggFilename.size() - 4);
+		oggFilename = oggFilename + ".ogg";
+		if (BaseFileManager::getEngineInstance()->hasFile(oggFilename)) {
+			useFilename = oggFilename;
 		}
 	}
 


fuzzie:
So the changes here are (a) adding the dot before the extension, and then (b) removing the call to combine? I'm curious about (b) since your commit message says "probably don't want".. why probably? Also, you say "should fix #7088" - you don't have the game to test it?

salty-horse:
I tested the fix with the game (which is freely available online), and it works.

Two comments:
1) `PathUtil::combine` seems to do the "wrong" thing since the "wav" string looks like:
  `interface\speech\sounds\knight_sing.wav`
and the `combine`d string looks like:
  `interface/speech/sounds/knight_singogg`
I'd rather keep the current path separator. Maybe the code could be simplified to look at the last 3 chars and replace them (in a new copy of the string) without reconstructing the path + name parts.

2) The game crashes in libvorbis while playing `actors\dragon\dragon_roar_distant.ogg` - probably a separate issue.

tobiatesan:
@fuzzie:
> (a) adding the dot before the extension, and then

True

>  (b) removing the call to combine

True

>  you say "should fix #7088" - you don't have the game to test it?

I have it and it works for me. I have rephrased the commit, thanks.
I have a tendency to err on the side of conditional tense which probabl^N^N^N actually doesn't belong to commit messages.

> why probably?

Because I wouldn't want you to take my word for it until either @somaen (who originally wrote this bit) or @wjp have seen it.

> (BTW, the game crashes during the intro while playing actors\dragon\dragon_roar_distant.ogg if you're using an old version of libvorbis - it's unrelated to this bug)

Yes, it's #7089

tobiatesan:
@salty-horse:

> Maybe the code could be simplified to look at the last 3 chars and replace them (in a new copy of the string) without reconstructing the path + name parts.

I like this better as well. See updated commit.

engines/wintermute/base/sound/base_sound_manager.cpp
```diff
@@ -100,15 +100,15 @@ BaseSoundBuffer *BaseSoundMgr::addSound(const Common::String &filename, Audio::M
...
-		AnsiString name = PathUtil::getFileNameWithoutExtension(filename);
-
-		AnsiString newFile = PathUtil::combine(path, name + "ogg");
-		if (BaseFileManager::getEngineInstance()->hasFile(newFile)) {
-			useFilename = newFile;
+	if (useFilename.hasSuffix("wav")) {
+		Common::String oggFilename = useFilename;
+		oggFilename.erase(oggFilename.size() - 3);
+		assert(oggFilename.lastChar() == '.');
+		oggFilename = oggFilename + "ogg";
```
salty-horse: Isn't there an append function? I think `s = s + suffix` creates a new string.

Instead of erasing, maybe just set the last 3 chars manually:
```
size = s.size()
s[size - 3] = 'o';
s[size - 2] = 'g';
s[size - 1] = 'g';
```

wjp: I really wouldn't worry about that compared to the cost of opening a file. But see also #688.

engines/wintermute/base/sound/base_sound_manager.cpp
```diff
@@ -100,15 +100,14 @@ BaseSoundBuffer *BaseSoundMgr::addSound(const Common::String &filename, Audio::M
...
 	// try to switch WAV to OGG file (if available)
-	AnsiString ext = PathUtil::getExtension(filename);
-	if (StringUtil::compareNoCase(ext, "wav")) {
-		AnsiString path = PathUtil::getDirectoryName(filename);
-		AnsiString name = PathUtil::getFileNameWithoutExtension(filename);
-
-		AnsiString newFile = PathUtil::combine(path, name + "ogg");
-		if (BaseFileManager::getEngineInstance()->hasFile(newFile)) {
-			useFilename = newFile;
+	if (useFilename.hasSuffix(".wav")) {
```
lordhoto: Yes, but it does do a ``toLowercase`` on ``useFilename`` above, so it shouldn't be an issue (for this check).

tobiatesan: Do you prefer `getExtension`? I'm rather agnostic on the issue (however you polish I think... well, it's still a hack) I was mostly taking @salty-horse's advice.

engines/wintermute/base/sound/base_sound_manager.cpp
```diff
@@ -100,15 +100,14 @@ BaseSoundBuffer *BaseSoundMgr::addSound(const Common::String &filename, Audio::M
...
+		Common::String oggFilename = useFilename;
+		oggFilename.erase(oggFilename.size() - 4);
+		oggFilename = oggFilename + ".ogg";
```
salty-horse: You could erase just 3 chars, leaving the dot alone.

tobiatesan: If you look at the first commit that's how I phrased it originally.
But I feel this is... less ambiguous (a missing dot _was_ the original problem after all) and I think writing a single extra char everytime you visit a new room is hardly relevant.

tobiatesan: (Okay, `n` extra chars where `n` is at most the number of sounds in that scene).

PR #730, opened 2016-03-28
GUI: Add repeatEvent flag to widgets and dialogs handleKeyDown calls

criezy:
This allows widgets and dialogs to know if the call correspond to an initial key down event or a repeated key down event. This is used in the Tooltip to avoid closing the tooltip on repeated key down events. Otherwise this makes it for example impossible to see the tooltip on the Mass Add button (since we get Shift key down repeated events).

This is an alternate (and in my opinion cleaner solution) than what otto gin tried to implement in PR #716. In that PR I suggested we might want to pass the full event to the widgets and dialogs event handlers. But since I am lazy I decided to only modify handleKeyDown by adding a flag.

Note: I could have changed only Dialog (and classes that derive from it) but decided to also change the Widget class for consistency. No Widget currently use that flag.

I am submitting this as a pull request to get opinions on whether this is the best way to fix this tooltip issue.

### DIFF
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -517,6 +517,8 @@ void GuiManager::processEvent(const Common::Event &event, Dialog *const activeDi
 	Common::Point mouse(event.mouse.x - activeDialog->_x, event.mouse.y - activeDialog->_y);
 	switch (event.type) {
 	case Common::EVENT_KEYDOWN:
+		// TODO: Using event.synthetic here is not correct. Repeat key down events are synthetic
+		// but there are other synthetic events (for examples events replayed by the event recorder).
 		activeDialog->handleKeyDown(event.kbd, event.synthetic);
 		break;
 	case Common::EVENT_KEYUP:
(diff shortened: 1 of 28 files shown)


gui/gui-manager.cpp
```diff
@@ -517,7 +517,7 @@ void GuiManager::processEvent(const Common::Event &event, Dialog *const activeDi
 	Common::Point mouse(event.mouse.x - activeDialog->_x, event.mouse.y - activeDialog->_y);
 	switch (event.type) {
 	case Common::EVENT_KEYDOWN:
-		activeDialog->handleKeyDown(event.kbd);
+		activeDialog->handleKeyDown(event.kbd, event.synthetic);
```
lordhoto: I think it's best adding a TODO here which marks that using "synthetic" is not correct here since more than just keyboard repeat events are synthetic (it's "keyboard repeat => synthetic == true" not the other way round).

criezy: Yeah I was wondering about that. Other than the event recorder, are there any other way to generate synthetic events that are not repeat key down events? And do you know of a better way to detect repeat key down events? Your suggestion to just add a TODO suggests you do not, but I though I might ask anyway ;)

criezy:
After discussing with sev I am closing this pull request as he is thinking of a different approach in how tooltips should work.

gui/gui-manager.cpp
lordhoto: Well, extend the event structure to also carry this information if this is the right thing to do :-).

PR #732, opened 2016-03-30
WINTERMUTE: Log warning if addSound is called with empty filename

tobiatesan:
Bickadoodle does this.

<s>Fixes #6594</s> It doesn't, that is best handled in #733

### DIFF
--- a/engines/wintermute/base/sound/base_sound_manager.cpp
+++ b/engines/wintermute/base/sound/base_sound_manager.cpp
@@ -97,6 +97,11 @@ BaseSoundBuffer *BaseSoundMgr::addSound(const Common::String &filename, Audio::M
 		return nullptr;
 	}
 
+	if (filename.empty()) {
+		// At least one game, Bickadoodle, calls playSound with an empty filename, see #6594
+		BaseEngine::LOG(0, "addSound called with empty filename");
+	}
+
 	BaseSoundBuffer *sound;
 
 	Common::String useFilename = filename;


tobiatesan:
Sorry, don't merge yet.

tobiatesan:
PathUtil::getFileName which is where a failing assert is hit (char with index -1 is accessed) as a consequence probably deserves a check to be added as well, _but_ I would leave this here for clarity as well.

tobiatesan:
See #733; #733 alone _is sufficient_ to fix #6594.

engines/wintermute/base/sound/base_sound_manager.cpp
```diff
@@ -97,6 +97,12 @@ BaseSoundBuffer *BaseSoundMgr::addSound(const Common::String &filename, Audio::M
 		return nullptr;
 	}
 
+	if (filename == "") {
```
lordhoto: `empty()` is generally preferred to test if something is empty or not

tobiatesan: Thanks

engines/wintermute/base/sound/base_sound_manager.cpp
```diff
@@ -97,6 +97,12 @@ BaseSoundBuffer *BaseSoundMgr::addSound(const Common::String &filename, Audio::M
 		return nullptr;
 	}
 
+	if (filename.empty()) {
+		// At least one game, Bickadoodle, calls playSound with an empty filename, see #6594
+		BaseEngine::LOG(0, "addSound called with empty filename");
```
somaen: Are you sure that it actually does that in the scripts (and i.e. we don't screw something up along the way?)

Does WME proper allow for calling with invalid/blank filenames?

tobiatesan: @somaen: If I add a breakpoint in CBSoundBuffer::LoadFromFile I can reproduce in the original engine, so it's either a script bug or... a bugfeature of the original script interpreter.

In the original, of course, it just returns an E_FAIL when it can't find the file, instead of segfaulting.

I removed the `return nullptr` bit because it seems to me the segfaulting problem is best fixed in #733, I was originally wary of being that heavy handed, but after spending enough time with it and testing for a few edge cases I think it's solid enough.

I guess logging the anomaly would not hurt in case some similar bug pops up in the future, though?

bluegr:
This has ended up as a warning for a specific behavior, so there's no change in functionality with the changes in this pull request.

Will this be merged, or closed?

sev-:
Merging it as it is basic

PR #733, opened 2016-03-30
WINTERMUTE: Fix PathUtils

tobiatesan:
This avoids hitting a failing assert when passing strings == "".
E.g. Bickadoodle has a script bug that tries to load an empty filename.

This fixes #6594

### DIFF
--- a/engines/wintermute/utils/path_util.cpp
+++ b/engines/wintermute/utils/path_util.cpp
@@ -71,7 +71,11 @@ bool PathUtil::hasTrailingSlash(const Common::String &path) {
 Common::String PathUtil::getDirectoryName(const Common::String &path) {
 	Common::String newPath = unifySeparators(path);
 	Common::String filename = getFileName(path);
-	return Common::String(path.c_str(), path.size() - filename.size());
+	if (hasTrailingSlash(newPath)) {
+		return path;
+	} else {
+		return Common::String(path.c_str(), path.size() - filename.size());
+	}
 }
 
 //////////////////////////////////////////////////////////////////////////
--- a/engines/wintermute/utils/path_util.cpp
+++ b/engines/wintermute/utils/path_util.cpp
@@ -63,6 +63,10 @@ Common::String PathUtil::combine(const Common::String &path1, const Common::Stri
 	return newPath1 + newPath2;
 }
 
+bool PathUtil::hasTrailingSlash(const Common::String &path) {
+	return (path.size() > 0 && path[path.size() - 1 ] == '/');
+}
+
 //////////////////////////////////////////////////////////////////////////
 Common::String PathUtil::getDirectoryName(const Common::String &path) {
 	Common::String newPath = unifySeparators(path);
@@ -74,10 +78,10 @@ Common::String PathUtil::getDirectoryName(const Common::String &path) {
 Common::String PathUtil::getFileName(const Common::String &path) {
 	Common::String newPath = unifySeparators(path);
 	Common::String lastPart = Common::lastPathComponent(newPath, '/');
-	if (lastPart[lastPart.size() - 1 ] != '/') {
-		return lastPart;
+	if (hasTrailingSlash(newPath)) {
+		return Common::String("");
 	} else {
-		return path;
+		return lastPart;
 	}
 }
 
--- a/engines/wintermute/utils/path_util.cpp
+++ b/engines/wintermute/utils/path_util.cpp
@@ -32,8 +32,8 @@
 namespace Wintermute {
 
 //////////////////////////////////////////////////////////////////////////
-AnsiString PathUtil::unifySeparators(const AnsiString &path) {
-	AnsiString newPath = path;
+Common::String PathUtil::unifySeparators(const Common::String &path) {
+	Common::String newPath = path;
 
 	for (uint32 i = 0; i < newPath.size(); i++) {
 		if (newPath[i] == '\\') {
@@ -45,16 +45,16 @@ AnsiString PathUtil::unifySeparators(const AnsiString &path) {
 }
 
 //////////////////////////////////////////////////////////////////////////
-AnsiString PathUtil::normalizeFileName(const AnsiString &path) {
-	AnsiString newPath = unifySeparators(path);
+Common::String PathUtil::normalizeFileName(const Common::String &path) {
+	Common::String newPath = unifySeparators(path);
 	newPath.toLowercase();
 	return newPath;
 }
 
 //////////////////////////////////////////////////////////////////////////
-AnsiString PathUtil::combine(const AnsiString &path1, const AnsiString &path2) {
-	AnsiString newPath1 = unifySeparators(path1);
-	AnsiString newPath2 = unifySeparators(path2);
+Common::String PathUtil::combine(const Common::String &path1, const Common::String &path2) {
+	Common::String newPath1 = unifySeparators(path1);
+	Common::String newPath2 = unifySeparators(path2);
 
 	if (!newPath1.hasSuffix("/") && !newPath2.hasPrefix("/")) {
 		newPath1 += "/";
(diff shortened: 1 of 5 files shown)


engines/wintermute/utils/path_util.cpp
```diff
@@ -74,6 +74,11 @@ AnsiString PathUtil::getDirectoryName(const AnsiString &path) {
 AnsiString PathUtil::getFileName(const AnsiString &path) {
 	AnsiString newPath = unifySeparators(path);
 	Common::String lastPart = Common::lastPathComponent(newPath, '/');
+
+	if (lastPart.empty()) {
+		return path;
+	}
```
somaen: Technically, the following code should be else-if to this I guess.

On the other hand, does lastPathComponent actually return strings with the separator still in them? (Or could we simply do else { return lastPart; } ?

lordhoto: > On the other hand, does lastPathComponent actually return strings with the separator still in them? (Or could we simply do else { return lastPart; } ?

I'm sure it should not contain any path separator inside the result. So, returning the result of ``lastPathComponent`` should be fine AFAICT. I do not understand the reason for the `if` check below though.

tobiatesan: It is apparently not a WME thing:
https://bitbucket.org/MnemonicWME/wme1/src/62204728c14de49e49d5b6b70f93a00c310f2e0f/src/engine_core/wme_base/utils.cpp?at=default&fileviewer=file-view-default

Apparently, the whole getFilename method can be removed without damage, as in PR #737

Although: I didn't do a _lot_ of testing yet.

tobiatesan: Apparently it can wrap lastPathComponent straight away without much damage.

Beware that only I tested it _a little_.

tobiatesan: Given the context (https://github.com/scummvm/scummvm/commit/afee4aeacccf66f52fefacbb0904049f0ce45227#diff-cb293c2a8565f312766de8a207314d1eL90) I wouldn't be surprised if it was a leftover from... something.

tobiatesan:
Please don't merge just yet, I am in the process of adding a few test cases

tobiatesan:
@lordhoto @somaen I've added a bunch of tests and so far the only change in behaviour is that the segafaulting calls are no longer segfaulting.

Anything weird has yet to happen when actually running it.

I think this is safe.

tobiatesan:
I'm sorry - is there anything blocking this one? (Reminder - this actually causes a crash, it's not just cosmetic)
I guess once this is in along with #714 I could put a few test cases in place and refactor the whole thing in a less redundant fashion, if that's the issue.

test/engines/wintermute/path_utils.h
```diff
@@ -0,0 +1,229 @@
...
+				);
+		TS_ASSERT_EQUALS(
+				Wintermute::PathUtil::getDirectoryName(windowsCapPath),
+				Common::String("C:\\SOME\\")
+				);
+
+		/*
+		 * Failing assert in Common::Array
+		 */
+		// TS_ASSERT_EQUALS(
```
somaen: Why is this here then?

test/engines/wintermute/path_utils.h
```diff
@@ -0,0 +1,229 @@
...
+	const Common::String windowsRelativePath;
+	PathUtilTestSuite () :
+		unixPath("/some/file.ext"),
+		unixCapPath("/SOME/FILE.EXT"),
+		windowsPath("C:\\some\\file.ext"),
+		windowsCapPath("C:\\SOME\\FILE.EXT"),
+		emptyString(""),
+		dualExtPath("/some/file.tar.gz"),
+		manyExtPath("/some/file.tar.bz2.gz.zip"),
+		mixedSlashesPath1("C:\\this/IS_REALLY\\weird.exe"),
```
somaen: Why do we want to allow this?

test/engines/wintermute/path_utils.h
```diff
@@ -0,0 +1,229 @@
...
+	const Common::String emptyString;
+	const Common::String dualExtPath;
+	const Common::String manyExtPath;
+	const Common::String mixedSlashesPath1;
+	const Common::String mixedSlashesPath2;
+	const Common::String unixRelativePath;
+	const Common::String windowsRelativePath;
+	PathUtilTestSuite () :
+		unixPath("/some/file.ext"),
+		unixCapPath("/SOME/FILE.EXT"),
```
somaen: How about the case where you have a trailing slash?

test/engines/wintermute/path_utils.h
somaen: Additionally, absolute paths are bad(tm)

engines/wintermute/utils/path_util.cpp
```diff
@@ -73,12 +73,7 @@ Common::String PathUtil::getDirectoryName(const Common::String &path) {
 //////////////////////////////////////////////////////////////////////////
 Common::String PathUtil::getFileName(const Common::String &path) {
 	Common::String newPath = unifySeparators(path);
-	Common::String lastPart = Common::lastPathComponent(newPath, '/');
-	if (lastPart[lastPart.size() - 1 ] != '/') {
-		return lastPart;
-	} else {
-		return path;
-	}
+	return Common::lastPathComponent(newPath, '/');
```
somaen: Does this guarantee that i.e. /foo/bar/ returns ""?

test/engines/wintermute/path_utils.h
```diff
@@ -53,14 +53,10 @@ class PathUtilTestSuite : public CxxTest::TestSuite {
 				Wintermute::PathUtil::getDirectoryName(windowsCapPath),
 				Common::String("C:\\SOME\\")
 				);
-
```
somaen: Preferably remove these from the preceding commit, and just introduce them here.

test/engines/wintermute/path_utils.h
tobiatesan: Because while absolute paths are the spawn of the devil there are a few games that use them already, by mistake - the CR series seems to be a major offender - and I thought that we might like to test for a defined behavior in that case, just to be sure of what the function is doing in that case.

We might very well want to change that behaviour, of course, but I don't think it has P=0 of happening.

test/engines/wintermute/path_utils.h
tobiatesan: Ah, thanks, good point.

test/engines/wintermute/path_utils.h
tobiatesan: I un-comment it in 76b49c8, after I've fixed the failing behaviour.

I can remove it altogether, I see how it can raise eyebrows but in this particukar circumstances it seemed clearer this way to me.

test/engines/wintermute/path_utils.h
tobiatesan: Okay, thanks

sev-:

tobiatesan:
I am rebasing right now.

test/engines/wintermute/path_utils.h
tobiatesan: On a second thought, I'm not sure it's so good a point.
I don't think you can put a "directory" in a .DCP, so my guess is that it's probably best if anything with a trailing / is *not* found.

I have added a commit (ac0d476) that takes care of trailing "/", but if you want my opinion I'd guess it would be best *not* to merge that one.

test/engines/wintermute/path_utils.h
tobiatesan: Done

test/engines/wintermute/path_utils.h
tobiatesan: I am an idiot, my commit just reinstates the previous behaviour. Sorry, it's been a few months.

engines/wintermute/utils/path_util.cpp
tobiatesan: Now it does. By the way, is this intended behaviour?
I don't think accessing stuff that ends with a / inside .DCPs is defined behaviour, or is it?

test/engines/wintermute/path_utils.h
tobiatesan: Fixed

tobiatesan:
How do you like this?

test/engines/wintermute/path_utils.h
tobiatesan: Notice the big fat disclaimer at the beginning of the file. Please feel free to reword or dictate a different approach. 

However, those *do* pop up from time to time and I am afraid we must handle them on way or the other :|

tobiatesan:
@bluegr I opened a separate PR #874 which includes this commit

sev-:
This is now obsoleted by #874

PR #734, opened 2016-03-30
WINTERMUTE: Enable bilinear filtering in Wintermute

tobiatesan:
1. Templat-ify scale() and rotoscale()
2. Add a menu item for bilinear filtering
3. Read the value of the above in RenderTicket and pick the correct specialization accordingly

This is useful in Bickadoodle, which makes a very liberal use of rotozooming; without bilinear on you can see some conspicuous pixels in a few places.

### DIFF
--- a/graphics/transparent_surface.cpp
+++ b/graphics/transparent_surface.cpp
@@ -37,8 +37,6 @@
 #include "graphics/transparent_surface.h"
 #include "graphics/transform_tools.h"
 
-//#define ENABLE_BILINEAR
-
 namespace Graphics {
 
 static const int kBModShift = 0;//img->format.bShift;
@@ -677,6 +675,7 @@ systems.
 
 
 
+template <TFilteringMode filteringMode>
 TransparentSurface *TransparentSurface::rotoscale(const TransformStruct &transform) const {
 
 	assert(transform._angle != 0); // This would not be ideal; rotoscale() should never be called in conditional branches where angle = 0 anyway.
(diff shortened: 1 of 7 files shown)


graphics/transparent_surface.cpp
```diff
@@ -606,49 +606,56 @@ TransparentSurface *TransparentSurface::rotoscale(const TransformStruct &transfo
...
-				pc->a = (((t2 - t1) * ey) >> 16) + t1;
-			}
-#else
+#ifdef DISABLE_BILINEAR
 			if ((dx >= 0) && (dy >= 0) && (dx < srcW) && (dy < srcH)) {
 				const tColorRGBA *sp = (const tColorRGBA *)getBasePtr(dx, dy);
 				*pc = *sp;
 			}
+#else
+			if (transform._filteringMode == FILTER_BILINEAR) {
```
somaen: Performing this check inside the inner loop is slooooow, consider upping this to be a decision based on a template parameter instead, so that the compiler will create two versions of this function, thereby moving the branch outside the entire loop without having to duplicate the code. (I think we're already doing that in the various draw functions of TransparentSurface, otherwise you'll find examples in subr3v's TinyGL code from 2014).

Another alternative is to put the if outside the for-loop and have a new set of for-loops after, but that leads to code duplication.

graphics/transform_struct.cpp
```diff
@@ -25,10 +25,11 @@
 
 namespace Graphics {
 
-void TransformStruct::init(Common::Point zoom, uint32 angle, Common::Point hotspot, bool alphaDisable, TSpriteBlendMode blendMode, uint32 rgbaMod, bool mirrorX, bool mirrorY, Common::Point offset) {
```
somaen: This commit should be squashed into the preceeding commit, as adding a variable to the constructor is something you should do when you introduce it.

graphics/transparent_surface.cpp
tobiatesan: Whooops, that's quite right, thanks.

graphics/transparent_surface.cpp
```diff
@@ -606,50 +605,50 @@ TransparentSurface *TransparentSurface::rotoscale(const TransformStruct &transfo
...
-				sp += 1;
-				c01 = *sp;
-				sp += (this->pitch / 4);
-				c11 = *sp;
-				sp -= 1;
-				c10 = *sp;
-				if (flipx) {
-					cswap = c00; c00=c01; c01=cswap;
-					cswap = c10; c10=c11; c11=cswap;
+			if(filteringMode == FILTER_BILINEAR) {
```
bluegr: Space between the if and the left parenthesis

sev-:
This does not merge cleanly anymore due to conflicts in graphics/transparent_surface.cpp

graphics/transparent_surface.cpp
tobiatesan: Fixed

tobiatesan:
@sev- Now it should merge cleanly again, apparently it was just a stray }

sev-:
Cool!

PR #735, opened 2016-03-30
UPDATES: Add GUI. Opt-in dialog and Options

sev-:
Recently I finally managed to get Sparkle working. Sparkle is a software update framework, initially for Mac, and now for Windows too (as WinSparkle project). It came to ScummVM in 2011 by means of a Pull Request: https://github.com/scummvm/scummvm/pull/52 .

This PR implements an opt-in dialog, which is shown on the very first run of ScummVM with updates support, which offers the user to (a) agree to have updates (b) to select frequency. By default frequency is selected to weekly checks:

Step a:
Step 1 of Wizard

Step b:
Step 2 of Wizard

At any time the user may disable/enable the check in our GUI:

Options Dialog

I invite you to comment on the wording, design and overall workflow.

And yes, I am aware of duplicated list of the popup values, I am thinking about enhancing the updates API with the following calls:

  Common::Array<int> getUpdateIntervals();
  Common::String updateIntervalToString(int interval);

And then use it everywhere. Comments on this are welcome.


RichieSams:
When it detects a new version, what happens? Does it just inform the user, or does it auto-update the binaries as well?

gui/updates-dialog.cpp
```diff
@@ -0,0 +1,151 @@
...
+		y += kLineHeight;
+	}
+
+	y += kLineHeight;
+	_updatesPopUpDesc = new StaticTextWidget(this, 10, y, _w, kLineHeight,  _("How often you would like to check for updates?"), Graphics::kTextAlignCenter);
+	y += kLineHeight + 3;
+
+	_updatesPopUp = new PopUpWidget(this, 10, y, _w - 20, g_gui.xmlEval()->getVar("Globals.PopUp.Height", kLineHeight));
+
+	_updatesPopUp->appendEntry(_("Never"), Common::UpdateManager::kUpdateIntervalNotSupported);
```
criezy: I find it a bit confusing to include the "Never" option at this step. If the user sees this that means he just clicked Yes when being asked if he wants to enable Automatic Checks. And he also has the option to use "Cancel" if he changes his mind. But I might just be nitpicking.

criezy:
It looks good to me.

However you should also remove from the MacOSXUpdateManager the following two lines (there is actually a TODO comment about it):
	setAutomaticallyChecksForUpdates(kUpdateStateEnabled);
	setUpdateCheckInterval(kUpdateIntervalOneDay);

And they should be replace by code that reads the config file to set those.

@RichieSams: It informs the user and auto-update if the user ask for it.
This is all handled by Sparkle. In details, it displays a dialog with the release notes and ask the user if he wants to install the new version or skip this update. If the user click yes it then downloads the new version in background and once the download is finished the user gets another dialog telling him he needs to restart the application for the new version to be installed, and there is a button to do so. When the user click the button the application quits, is moved to the trash, replaced by the new version and is reopened (all this automatically).

gui/updates-dialog.cpp
```diff
@@ -0,0 +1,151 @@
...
+UpdatesDialog::UpdatesDialog() : Dialog(30, 20, 260, 124) {
+
+	const int screenW = g_system->getOverlayWidth();
+	const int screenH = g_system->getOverlayHeight();
+
+	int buttonWidth = g_gui.xmlEval()->getVar("Globals.Button.Width", 0);
+	int buttonHeight = g_gui.xmlEval()->getVar("Globals.Button.Height", 0);
+
+	const char *message = _(
+		"ScummVM now supports automatic check for updates\n"
```
somaen: This doesn't ring entirely right in my ear, either it should be "updates" or "checking", the former does sound slightly better, but someone with english as a first language might want to chime in on that.

Also, perhaps skip "now", as this might not be an upgrade for the user, but their first ever install.

criezy: Personally I think I would leave "now". Otherwise if a user is upgrading ScummVM he might wonder why he is getting this dialog now (unless he read the release notes, but who does?). While a user that installs ScummVM for the first time will still understand that the 'now' means previous versions did not have this feature.

lordhoto:
I only have some visual feedback on the dialog design for now. I will take a look at the code in the new days.

I think we should consider merging steps (a) and (b). Just having a step (b) like dialog seems much leaner and generally less confusing to me (see also the point @criezy brought up on "Never" being a possible selection in step (b)).

criezy:
I wrote to sev on IRC initially "After just a quick look I am not sure why step a and step b are two separate steps". After looking at the code and starting to comment here I decided not to raise that point as in my opinion it has the advantage of making it very clear what the user is doing, thus removing any risk of confusion.

But indeed we could probably have this as a single step and still keep it clear. For example we could maybe have something like this:
```
ScummVM now supports automatic check for updates
which requires access to the Internet.
Would you like to enable this feature?
You can always enable it later in the options dialog
on the Misc tab.

[ ] Check for updates automatically [ Daily     v]

                                           [ Close ]
```

With the frequency selector being enabled only when the check box is toggled on.

sev-:
Yep, I like the idea with a checkbox, will implement it today.

sev-:
I adjusted the dialog:
Single step GUI

I decided to leave 'Cancel' button on, so the users could close it right away at any time.

As of removing 'Never' option, I wanted to have the PopUp to show exactly same number of options as in the Options dialog, but there is no problem with removing it.

sev-:
..and now I got rid of the hardcoded list and implemented reading config values in the MacOS X manager.

The feature is complete now.

raziel-:
I've got a question:

Is this meant to work platform independant or only for Mac/Windows backends?

If not, could it be enhanced to check on which platform/backend it is running and check for them updates aswell?
Given that the (main) releases are (always) gathered in one place this "should" be possible, shouldn't it?

backends/updates/macosx/macosx-updates.mm
```diff
@@ -74,11 +75,13 @@
...
-	// checkForUpdates() here to check for updates on every startup)
-	// TODO: Should be removed when an update settings gui is implemented
-	setAutomaticallyChecksForUpdates(kUpdateStateEnabled);
-	setUpdateCheckInterval(kUpdateIntervalOneDay);
+	if (!ConfMan.hasKey("updates_check")
+			|| ConfMan.getInt("updates_check") == Common::UpdateManager::kUpdateIntervalNotSupported) {
+		setAutomaticallyChecksForUpdates(kUpdateStateDisabled);
+	} else {
+		setAutomaticallyChecksForUpdates(kUpdateStateEnabled);
+		setUpdateCheckInterval(ConfMan.getInt("updates_check"));
```
criezy: I wonder if we should check that the getInt gives us something that is one of the value in the UpdateInterval  enum (or one of the value in the updateIntervals array, which is the same). If a user edits the file he can change it to anything, and I am not sure what Sparkle does when we call setUpdateCheckInterval with a negative or null NSTimeInterval. Also checking for updates every seconds or 10 seconds would not be a very good idea. Is there any case where it would be useful for the user to specify an interval in the config file that is not one of the predefined ones?
Note: if we add it the check should maybe be in setUpdateCheckInterval and not here.

criezy:
@sev:
> I decided to leave 'Cancel' button on, so the users could close it right away at any time.
With a unique button named 'Close' the user would also have the possibility to close the dialog right away at any time.
My worry with a button Cancel is that the user may think the dialog will close and nothing will be done, while actually it will set the options to never to automatic check (so clicking Cancel will result in a change in the config, which seems a strange behaviour to me). With Close, it seems clearer to me that the config will be change (to either enable or disable automatic updates check depending on the state of the checkbox).

backends/updates/macosx/macosx-updates.mm
sev-: Good point. I just implemented normalizing while setting via API and in the GUI. When incorrect value is provided I fall back to one day.

Currently Sparkle itself restricts internally updates to 3600 seconds, e.g. once per hour.

criezy:
@raziel- 
The Updates API is platform independent, and the GUI code in this pull request is also platform independent but will only be used on platforms where the Updates API is implemented. And currently there is only one (OS X - is uses Sparkle to implement the API).
As indicated by sev the Updates API could also quite easily be implemented on Windows using WinSparkle, but our Windows maintainer doesn't want automatic updates check to be included in the Windows port.
If you know of a way to implement such a feature on Amiga then the API could also be implemented there. And the same goes for Linux or other platforms.

sev-:
@raziel- The GUI is platform-independent. Updates backends is currently implemented for Mac, and I have plans to implement it for Windows.

Other systems may consider implementing it as well where applicable, e.g. there is no sense to do that, say, on Android/iOS, since they have their own update systems, or some platform like DC may have no network connection normally.

raziel-:
Thank you for all mthe answers, guys.

I think i'll go with a native updater

sev-:
I forgot to mention that I added immediate check button to the Options.

Here is the screenshot:

Check now button

common/updates.h
```diff
@@ -85,18 +85,21 @@ class UpdateManager {
...
 
 	/**
 	 * Gets the update check interval.
 	 *
 	 * @return  the update check interval.
 	 */
-	virtual UpdateInterval getUpdateCheckInterval() { return kUpdateIntervalNotSupported; }
+	virtual int getUpdateCheckInterval() { return kUpdateIntervalNotSupported; }
+
+	static const int *getUpdateIntervals();
```
lordhoto: We need to document this.

common/updates.h
```diff
@@ -85,18 +85,21 @@ class UpdateManager {
...
+	static const char *updateIntervalToString(int interval);
```
lordhoto: We also need some documentation here.

common/updates.h
```diff
@@ -85,18 +85,21 @@ class UpdateManager {
 	 *
 	 * @param  interval    The interval.
 	 */
-	virtual void setUpdateCheckInterval(UpdateInterval interval) {}
+	virtual void setUpdateCheckInterval(int interval) {}
```
lordhoto: So what is the reasoning we do allow arbitrary values now?

gui/options.cpp
```diff
@@ -1229,6 +1231,25 @@ GlobalOptionsDialog::GlobalOptionsDialog()
...
+	_updatesPopUp = new PopUpWidget(tab, "GlobalOptions_Misc.UpdatesPopup");
+
+	const int *vals = Common::UpdateManager::getUpdateIntervals();
+
+	while (*vals != -1) {
+		_updatesPopUp->appendEntry(Common::UpdateManager::updateIntervalToString(*vals), *vals);
+		vals++;
+	}
+
+	if (ConfMan.hasKey("updates_check"))
```
lordhoto: In case this does not exist, the dialog will open at the start of ScummVM. This makes it seem to me that this case is always hit?

gui/options.cpp
```diff
@@ -1229,6 +1231,25 @@ GlobalOptionsDialog::GlobalOptionsDialog()
...
+		_updatesPopUp->setSelectedTag(ConfMan.getInt("updates_check"));
```
lordhoto: ``setSelectedTag`` will select nothing in case the value from the config file is not one of our predefined values. It looks like this case needs to be handled specially. I can't really tell from our sources what the implications here are, but it might be that ``getSelectedTag`` will return ``0xFFFFFFFF`` in this case. And I have no idea how this affects the GUI logic from a quick glance, but it looks like no precautions are made here?

lordhoto: ``setSelectedTag`` will select nothing in case the value from the config file is not one of our predefined values. It looks like this case needs to be handled specially. I can't really tell from our sources what the implications here are, but it might be that ``getSelectedTag`` will return ``0xFFFFFFFF`` in this case. And I have no idea how this affects the GUI logic from a quick glance, but it looks like no precautions are made here?

lordhoto:
Since we plan to add update support to more platforms in the future, wouldn't it make sense to have helper functionality for configuration file handling (i.e. setting and retrieving intervals) in common/updates.[h/cpp]? Currently it looks like the GUI and the MacOS X implementation seem to duplicate logic here. There is definitely some room for improvment and clean up here.

Another place where code duplication occurs is between the global options and the new UpdatesDialog shown at startup. Not sure how viable refactoring here is to reduce code duplication but I think we should give it a thought.

Is the checkbox checked by default? Or don't we want to suggest that we think it makes sense to enable it by default?

gui/options.cpp
sev-: I implemented and use the relevant method for normalizing the values.

gui/options.cpp
sev-: RIght, removed the check.

sev-:
Yes, I plan to move the config file manipulation after the merge. Although those will be very small methods.

The checkbox is not checked by default in order to make this feature opt-in, not opt-out. In the first two step implementation the user had to push 'Yes, I want the updates' button first. Same logic is here. If you, however, feel that we may check the box by default, it could be easily done.

sev-:
Guys, OK to merge?

criezy:
OK for me.

PR #737, opened 2016-03-31
WINTERMUTE: Remove PathUtil::getFilename

tobiatesan:
Use Common::lastPathComponent instead.
Fixes #6594.

More heavy-handed alternative to PR #733


tobiatesan:
Disregard for the moment, sorry.

PR #738, opened 2016-03-31
MATH: Move round() and roundUp() to common/math.h

tobiatesan:
This borders on useless, but I think those probably fit in common/ rather than inside the engine.

Unless I'm mistaken these are ScummVM originals, there seems to be no MathUtil.cpp in wme1/


sev-:
I think it is a good idea.

But being in common/, could you please add documentation to the methods?

wjp:
Maybe, but roundUp() is presumably just ceil(), and round() can be done much shorter with floor(val + 0.5f).

tobiatesan:
@wjp I have no idea how it didn't occur to me that roundUp is actually ceil. 
Thanks.

lordhoto:
I don't think it's a good idea to have a function with the same name as a standard library function but different behavior (i.e. standard C99's ``round`` rounds away from zero, while our version in ``Common`` does not do this.)

tobiatesan:
Very fair point. "`rint`"?

tobiatesan:
Very fair point. "`nearestIntf`"?

sev-:
@lordhoto, what do you think now?

lordhoto:
I think what makes more sense is to re-evaluate whether Wintermute really wants to use this specific rounding beahvior (instead of, for example, simply using `round`). Seeing that it is currently only used in a function, which returns before the actual code using it, it seems questionable we need anything like `nearestIntf`. @tobiatesan, @somaen: Any ideas here?

Apart: The name can still be confusing I think. It's basically the same name as the standard description of `round`. There need to be more clear docs and unit tests which check for this specific rounding behavior if we keep this.

lordhoto:
That said: Thinking about it some more: if wintermute really uses such a specific rounding behavior, it's probably best to keep it local to Wintermute. Otherwise people can easily get confused why this exists or they accidently use it without realizing the implications.

tobiatesan:
> whether Wintermute really wants to use this specific rounding beahvior

That's probably a question for @wjp / @somaen

I peeked at the log for the file in question, but it seems to stem from a single giant commit.

sev-:
Well, in the light of the above discussion, let's not merge it. Closing.

PR #739, opened 2016-03-31
SCUMM: Add checkbox for object_labels setting in "Edit Game" dialog

salty-horse:
This has been split off from PR #736.

Adds a custom engine setting for COMI's object_labels.
At the moment this will only appear in the launcher, but the other PR makes it also appear in the GMM Options dialog.


lordhoto:
Looks good to me. If we ever add more inside SCUMM, we should think of a better way to organize the settings and querying them though.

PR #740, opened 2016-04-01
DEBUGGER: Drop events when using readline debugger

fuzzie:
Some Windows users find that Windows marks ScummVM as unresponsive when using the readline debugger, because it doesn't consume events. This patch fixes this by dropping all incoming events while readline is waiting for input (thanks @waltervn for verifying).

I have not tested this with non-GNU readline (i.e. the OS X one), which might not have rl_event_hook, so it would be nice if someone could test it there (and we can add an extra check to configure if necessary).

### DIFF
--- a/gui/debugger.cpp
+++ b/gui/debugger.cpp
@@ -42,6 +42,7 @@
 #elif defined(USE_READLINE)
 	#include <readline/readline.h>
 	#include <readline/history.h>
+	#include "common/events.h"
 #endif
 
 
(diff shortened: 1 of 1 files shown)


lordhoto:
Looks fine. Our GUI based debugger will effectively eat up all events too, thus this should be fine behavior. We might consider adding some way to simply update the event queue instead of eating up the events in the long run. This way we can get responsiveness while not "losing" events.

I think we can safely ignore issues like the while loop reading *all* queued up events and blocking too long here.

sev-:
Great thing, thanks a lot!

PR #741, opened 2016-04-01
SCI: Clean up some aspects of call handling

wjp:
This has a few risky commits:

"Improve kernel subfunction logging" adds kernel subfunction information to ExecStack, and removes the re-use of debugSelector for kernel call information. I think I untangled all uses of this, but this could use review and testing. (Especially in workaround processing.)

"Remove unclear &rest handling" is potentially dangerous. It removes a change to a value _above_ the current stack pointer, which seems to have no clear effect. My guess is that this is an unintended part of these two commits from FreeSCI[*], attempting to adjust argc on the stack. Incorrectly setting argc would not necessarily have been noticed, as `add_exec_stack_entry` would store it on the stack properly afterwards. This also needs review and testing.

[*] wjp/freesci-archive@349958902d0e7a3f66e949c46cdac4633462cd8c and wjp/freesci-archive@b4daf42d0c75c042b2c0ce076e78167431879291 

"Remove unexpected side effect from ExecStack constructor" moves setting `argp[0]` to `argc` from its very well-hidden spot in a constructor to the calling code, to make stack handling more explicit. This needs review too.

### DIFF
--- a/engines/sci/engine/kernel.cpp
+++ b/engines/sci/engine/kernel.cpp
@@ -84,23 +84,15 @@ uint Kernel::getKernelNamesSize() const {
 }
 
 const Common::String &Kernel::getKernelName(uint number) const {
-	// FIXME: The following check is a temporary workaround for an issue
-	// leading to crashes when using the debugger's backtrace command.
-	if (number >= _kernelNames.size())
-		return _invalid;
+	assert(number < _kernelFuncs.size());
 	return _kernelNames[number];
 }
 
 Common::String Kernel::getKernelName(uint number, uint subFunction) const {
-	// FIXME: The following check is a temporary workaround for an issue
-	// leading to crashes when using the debugger's backtrace command.
-	if (number >= _kernelFuncs.size())
-		return _invalid;
+	assert(number < _kernelFuncs.size());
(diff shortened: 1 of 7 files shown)


sev-:
Looks like nobody has time to review or comment on this. Feature branch development requires some additional dedication.

engines/sci/engine/kernel.cpp
```diff
@@ -91,6 +91,20 @@ const Common::String &Kernel::getKernelName(uint number) const {
 	return _kernelNames[number];
 }
 
+Common::String Kernel::getKernelName(uint number, uint subFunction) const {
+	// FIXME: The following check is a temporary workaround for an issue
+	// leading to crashes when using the debugger's backtrace command.
```
csnover: Do you know what the issue is and can maybe explain it in more detail in the comment?

wjp: This comment is copied from the same check in the other `getKernelName` function. It seems to be a very old check from the time we stored kernel functions differently on the stack. I could find no caller of `getKernelName` that should now be able to trigger this, so I've replaced them with asserts in 4b72a42da93da5560ecde38010d328196ff727fd.

PR #742, opened 2016-04-02
GUI: Fix the cursor lags in the launcher

rootfather:
_Disclaimer: I only have basic understanding of C++. The commit I did fixed the problem for my, but I can absolutely not guarantee it has some unwanted side effects. Extensive testing on all backends is highly recommended before merging. It's also very possible I fix this the wrong way._

_The main purpose of this pull request is discussing this issue._

On my current Windows 10 laptop, the mouse cursor in the launcher has a very high lag. It doesn't move smoothly, but stutters. I had the same problem on my Intel Quad-Core with Windows 7 and on Ubuntu. @raziel- reported the same issues on the AmigaOS builds. I don't know if more backends are affected.

The two lines I re-enabled were introduced back in 2008 with commit 85c36885f5bbf2d47276c7702f1b8ccbf22ecc34.

After removing the comment marks, the cursor behaves very smooth on my weak Windows 10 laptop. I couldn't find any side effects, but again, this is mainly for discussion and needs some testing. I can't imagine the original author commented these lines out just for fun. Unfortunately, I can't test with Linux, because I never get a good performance inside a VM and I don't have any spare PC to test.

### DIFF
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -304,9 +304,9 @@ void GuiManager::runLoop() {
 //		_system->updateScreen();
 
 		if (lastRedraw + waitTime < _system->getMillis(true)) {
+			lastRedraw = _system->getMillis(true);
 			_theme->updateScreen();
 			_system->updateScreen();
-			lastRedraw = _system->getMillis(true);
 		}
 
 		Common::Event event;
@@ -342,9 +342,9 @@ void GuiManager::runLoop() {
 
 
 			if (lastRedraw + waitTime < _system->getMillis(true)) {
+				lastRedraw = _system->getMillis(true);
 				_theme->updateScreen();
 				_system->updateScreen();
-				lastRedraw = _system->getMillis(true);
 			}
 		}
 
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -286,7 +286,7 @@ void GuiManager::runLoop() {
 
 	Common::EventManager *eventMan = _system->getEventManager();
 	uint32 lastRedraw = 0;
-	const uint32 waitTime = 1000 / 45;
+	const uint32 waitTime = 1000 / 60;
 
 	bool tooltipCheck = false;
 


raziel-:
It does "feel" a little more responsive, then again the lag isn't as bad as i remember.
Most of the lagging i get when moving the mouse around fast or clicking on buttons/tabs which open new pages/windows...this really takes a long time

rootfather:
Indeed, opening a new window or switching between tabs is still very unresponsive. This only "attacks" the cursor lag itself, unfortunately, overall launcher performance is still quite bad.

criezy:
What commit 85c3688 did was throttle the screen updates to 45Hz by adding the if condition below:

```
if (lastRedraw + waitTime < _system->getMillis()) {
 	_theme->updateScreen();
 	_system->updateScreen();
 	lastRedraw = _system->getMillis();
 }
```
with waitTime set to
`const uint32 waitTime = 1000 / 45;`

(and it also added a similarly throttled screen update inside the pollEvent while loop, while before there was no update at all there)

So I assume that change was made to avoid high CPU usage with unnecessary screen updates on fast computers. And I am not sure why it would cause a lag on some computers.

wjp:
@criezy: that logic only limits redraws to 45Hz if screen updates are instant, since it updates `lastRedraw` only after the update. If updates are slower, the refresh rate will be limited further.

criezy:
@wjp: That's right and I wondered why that lastRedraw update was done after the two calls to screenUpdate. Would it make sense to move it up?
But I would not expect this to make much of a difference. For platforms where the screen update takes more than 22ms (and assuming the rest of the loop is fast), instead of doing another one on the next iteration it will be delayed by a further 22ms. Would that be sufficient to make it look choppy?

rootfather:
> Would that be sufficient to make it look choppy?

Without much understanding of the code as mentioned before - it works for me.

wjp:
Sorry, what does exactly?

rootfather:
Sorry, I should be more precise.

The way I removed the comments of the two lines in my commit fixes the choppy look and feel of the cursor for me. I haven't tried anything else yet, and I can't explain why these two lines make any difference.

fuzzie:
Can you try moving the '_lastRedraw =' line to before the updateScreen calls inside the 'if', instead?

(Which scaler are you using? Does it help if you use the OpenGL backend?)

rootfather:
In the launcher, I use the default scaling settings. When I use the OpenGL backend, performance in the launcher doesn't change much, it's maybe a little bit more laggy than the default backend, but this might be just my imagination.

When I move the _lastRedraw lines as you suggested, performance in OpenGL increases a little bit, but not as smooth as my original approach.

To make it clear, in-game performance is always great with SDL and OpenGL rendering.

rootfather:
Is there any specific reason waitTime is set to 1000/45?

I lowered it to 500/45 out of curiosity, and in combination with moving the _lstReadraw= line before the updateScreen calls I get a very smooth cursor.

raziel-:
@rootfather 
See @criezy 's first comment

rootfather:
For me, there's no huge increase in CPU usage when decreasing the waitTime from ~20ms to ~10ms.

rootfather:
I reverted my previous changes and did what fuzzie recommended.

Updating the _lastRedraw variable before calling updateScreen has a huge impact in the cursor performance, it's almost smooth, especially when using the openGL renderer. There are only minor lags which are cause by the 22ms delay I guess. To prevent that, I increased the update screen rate to 60Hz -- now the cursor is quite smooth on my system without any impact on CPU load.

So -- what do you think -- has this any chance of getting merged some time or not? Could you please test on your platform if these changes have any (positive or negative) impact?

raziel-:
I can confirm that on AmigaOS4 the adapted changes show an exceptional improvement.

Not only is cursor movement lightning fast now (as i know it to be from the classic theme), but also moving it over buttons doesn't feel like driving over speedbumps anymore.

CPU usage visibly do go up for me, especially using OpenGL and insanely moving the cursor over buttons, but given the fact this is only used in the launcher and in-game menues it shouldn't impact on game speed et al, should it?

fuzzie:
This looks fine to me, thanks; changing it to 60hz makes sense. If there are performance problems on some lower-end platforms then we can always add an option to make them 30hz or so instead..

criezy:
This looks fine to me as well. I just gave it a quick test on OS X and on my Windows 10 and Linux VMs and I can't see any difference in term of CPU usage (or in term of behaviour - I didn't have any lag before myself) compared to the code without that change.

I pushed this. So you can blame me if this breaks something :P

PR #743, opened 2016-04-02
COMMON: Add Apple II platform

fuzzie:
Add Apple II to our possible game platforms list (which is not the same as the Apple IIgs which is already there).


sev-:
Thanks.

PR #744, opened 2016-04-06
GUI: When dialogs gain focus, inform them of the current mouse position

salty-horse:
Previously, they only reacted to the mouse position once it was moved.
This meant that if the cursor was on a button that just gained focus, it did not highlight.

Fixes [#7101](https://sourceforge.net/p/scummvm/bugs/7101/).

A few notes:

1. If you want to test the case where a new dialog opens, and the mouse cursor is above a button, you could use a hotkey in the a launcher to open "Add Game" when reading the mouse to be over the "Go Up" button which has a tooltip. However, any hotkeys get caught by the game list. I managed to test it by having `ListWidget::handleKeyDown()` return `false` immediately.

2. I removed `tooltipCheck` from the GUI loop. I wasn't sure how to keep it around, since I wanted tooltips to work on the newly-focused dialog without having the mouse move. Without it, there's a new "bug": if a tooltip shows up, and you make it go away by a method other than moving the mouse (like pressing "alt") it shows up again after a few seconds. Keeping that behavior seemed complicated, and I found some other issues with tooltips (e.g. they shouldn't re-appear as long as the mouse is still hovering over a widget, regardless of mousemove), so I figured it could be fixed later.

### DIFF
--- a/gui/Tooltip.h
+++ b/gui/Tooltip.h
@@ -41,6 +41,8 @@ public:
 	void setup(Dialog *parent, Widget *widget, int x, int y);
 
 	void drawDialog();
+
+	virtual void receivedFocus(int x = -1, int y = -1) {}
 protected:
 	virtual void handleMouseDown(int x, int y, int button, int clickCount) {
 		close();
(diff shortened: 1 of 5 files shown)


salty-horse:
As reported in the bug, there are other cases when buttons don't get highlight unless the mouse is moved. I think it's out-of-scope for this PR.

>Found another case where elements aren't aware of the current mouse position:
>
>Hover over a button widget, and press the mouse button. Hover over a second button widget and release the mouse press.
>
>The second button widget should get a highlight. It doesn't, until the mouse is moved again.

sev-:
Cool, finally!

PR #747, opened 2016-04-14
GOB: Reduce audio header dependencies

salty-horse:
This is an attempt to decouple the audio parts of GOB from the rest of the engine.

I moved the background atmosphere audio enum from `bgatmosphere.h` to `sound.h`, so files that want to use `sound.h` won't have to include `bgatmosphere.h` which includes the `audio/` headers.

The result:
If you change a file in `audio/`, only 8 gob files are recompiled instead of 39.


sev-:
Thanks!

PR #748, opened 2016-04-14
FULLPIPE: Reduce audio header dependencies

salty-horse:
This is an attempt to decouple the audio parts of FULLPIPE from the rest of the engine.

I changed all the `SoundHandle`s to pointers, so `fullpipe.h` and `sound.h` don't have to include `audio/mixer.h`.

The result:
If you change `audio/mixer.h`, only 3 fullpipe files are recompiled instead of 64.

### DIFF
--- a/engines/fullpipe/fullpipe.cpp
+++ b/engines/fullpipe/fullpipe.cpp
@@ -24,6 +24,7 @@
 
 #include "common/archive.h"
 #include "common/config-manager.h"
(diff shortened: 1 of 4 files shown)


salty-horse:
I tested this on a demo and it seems to work, but maybe I didn't trigger all the relevant code paths.

sev-:
Great! Thanks a lot!

PR #749, opened 2016-04-14
GUI: Remove 'sticky button' feature

salty-horse:
This feature made pressed buttons wait a few moments before returning to an unpressed state. It was half-implemented, and caused several visual bugs.

Fixes [#7083](https://sourceforge.net/p/scummvm/bugs/7083/), which details the bugs involved.

The predictive dialog is buggy. It was buggy before this change, and I only made enough changes so it would compile. If you try to press two num keys at once on the keyboard, you'll notice that the first button pressed will maintain its "pressed" animation.

Once this is merged, I will open a new bug about the predictive dialog, and we can discuss how it should handle multiple key presses.


sev-:
Thanks, please proceed with the further PR

salty-horse:
Opened a separate bug for predictive dialog: https://sourceforge.net/p/scummvm/bugs/7125/

PR #752, opened 2016-04-15
SCI: Add Russian translate of Torin's Passage (Windows version)

CrazyMax:
Official russian translate from SoftClub


sev-:
Thanks!

PR #756, opened 2016-04-30
SCUMM: Add detection for Monkey Island 1+2 Unofficial SE Talkies

rootfather:
This implements proper detection for the unofficial talkie versions of Monkey Island 1+2. 

Please see my feature request for further information and reference.
https://sourceforge.net/p/scummvm/feature-requests/708/


tsoliman:
Has anything changed since #90?

rootfather:
No, nothing changed there as far as I am aware.

All this patch does is adding the MD5 values to the detection table and create a new game variant with the appropriate settings, so it's not possible to select the SegaCD or FM-TOWNS version by accident.

I don't see a problem in adding the MD5 hashes to the detection table - this patch is sort of "semi-official" because it uses the original speech files from the original Re-Release.

tsoliman:
So did #90. Read the comments on that PR for the problem the SCUMM team had with adding the md5 hashes.

What I meant by "has anything changed" is: has the SCUMM engine team position/opinion changed"

rootfather:
Well, a few days ago I asked @sev- if such a patch would be accepted, and he said yes. Not sure how the rest of the team thinks about it.

sev-:
The #90 was talking about adding extra code, thus, it was remade in #91. Adding md5s is OK. Moreover, classic gfx and speech is part of the new gameplay of SE, and while it is not trivial to support the enhanced gfx, having this hybrid properly detected is fine in my opinion.

rootfather:
Thank you for the positive feedback! 

So, is it OK to merge this?

rootfather:
Would it be safe to remove the fallback detection of the talkie versions, know that there is a detection based on the MD5 hashes? Or should I leave it in there just in case there are users that are still using an old version of the patch (which I doubt - the older versions have quite a lot of bugs...)?
```
// Detect if there are speech files in this unknown game
if (detectSpeech(fslist, g)) {
	if (strchr(dr.game.guioptions, GUIO_NOSPEECH[0]) != NULL) {
		if (g->id == GID_MONKEY || g->id == GID_MONKEY2)
			// TODO: This may need to be updated if something important gets added in the top detection table for these game ids
			dr.game.guioptions = GUIO0();
		else
			warning("FIXME: fix NOSPEECH fallback");
	}
}
```

tsoliman:
@rootfather 
I am all for adding game entries since it makes it unambiguous. 👍 

@sev- 
The response I got from @Kirben in #90 was clearly against adding md5 game entries for the MI1&2 UTEs (emphasis mine)

> It would be good if we could avoid adding the additional game detection code **and extra game entries**, for these fan patches. Since we don't accept md5 entries for fan translations of other SCUMM games.

So we all came up with clever ways to avoid adding game entries in subsequent PRs, ending up with the code in #99.

I am happy to hear that adding md5s of fan-patches of SCUMM games is now ok.

rootfather:
Are there any more objections against this PR? Otherwise, I would happily merge it to the main branch.

PR #759, opened 2016-05-21
NEVERHOOD: Reduce audio header dependencies

salty-horse:

When modifying `audio/mixer.h`, only 17 files require recompilation instead of 59.


sev-:
Great work!

PR #760, opened 2016-05-21
SCUMM: Reduce audio header dependencies

salty-horse:

When modifying `audio/mixer.h`, only 31 files require recompilation instead of 58.


sev-:
Nice! Thanks a lot for your work!

PR #763, opened 2016-05-22
Add support for WinSparkle

bluegr:
This is a pull request to add support for WinSparkle - an update framework. WinSparkle is a port of Sparkle for MacOS, which is already supported, so the update logic was already there.

Sparkle is using special XML files for update information, called appcasts. The same appcast can be used for both Windows and MacOS updates, and is currently located here:

https://www.scummvm.org/appcasts/macosx/release.xml

Here is some more information, from Sparkle and WinSparkle:

    Sparkle is a software update framework for Mac OS X which uses appcasts for
     release information. Appcasts are RSS-like XML feeds which contain information
     about the most current version at the time. If a new version is available, the
     user is presented the release-notes/changes/fixes and is asked if he wants to
     update, and if yes the Sparkle framework downloads a signed update package
     from the server and automatically installs and restarts the software.
     More detailed information is available at the following address:
     http://sparkle.andymatuschak.org/
    
     WinSparkle is a heavily (to the point of being its almost-port) inspired by the
     Sparkle framework originally by Andy Matuschak that became the de facto standard
     for software updates on OS X.
     More detailed information is available at the following address:
     https://winsparkle.org/

### DIFF
--- a/backends/updates/win32/win32-updates.cpp
+++ b/backends/updates/win32/win32-updates.cpp
@@ -1,0 +1,132 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+// Disable symbol overrides so that we can use system headers.
+#define FORBIDDEN_SYMBOL_ALLOW_ALL
+
+#include "common/system.h"
+#include "backends/updates/win32/win32-updates.h"
+
+#ifdef USE_SPARKLE
+#include "common/translation.h"
+#include "common/config-manager.h"
+
+#include <time.h>
+#include <winsparkle.h>
+
+/**
+ * Sparkle is a software update framework for Mac OS X which uses appcasts for
+ * release information. Appcasts are RSS-like XML feeds which contain information
+ * about the most current version at the time. If a new version is available, the
+ * user is presented the release-notes/changes/fixes and is asked if he wants to
+ * update, and if yes the Sparkle framework downloads a signed update package
+ * from the server and automatically installs and restarts the software.
+ * More detailed information is available at the following address:
+ * http://sparkle.andymatuschak.org/
+ *
+ * WinSparkle is a heavily (to the point of being its almost-port) inspired by the
+ * Sparkle framework originally by Andy Matuschak that became the de facto standard
+ * for software updates on OS X.
+ * More detailed information is available at the following address:
+ * https://winsparkle.org/
+ *
+ */
+Win32UpdateManager::Win32UpdateManager() {
+	const char *appcastUrl = "https://www.scummvm.org/appcasts/macosx/release.xml";
+
+	win_sparkle_set_appcast_url(appcastUrl);
+    win_sparkle_init();
+	
+	if (!ConfMan.hasKey("updates_check")
+			|| ConfMan.getInt("updates_check") == Common::UpdateManager::kUpdateIntervalNotSupported) {
+		setAutomaticallyChecksForUpdates(kUpdateStateDisabled);
+	} else {
+		setAutomaticallyChecksForUpdates(kUpdateStateEnabled);
+		setUpdateCheckInterval(normalizeInterval(ConfMan.getInt("updates_check")));
+	}
+}
+
+Win32UpdateManager::~Win32UpdateManager() {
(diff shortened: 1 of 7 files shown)


backends/updates/win32/win32-updates.cpp
```diff
@@ -0,0 +1,132 @@
...
+ * for software updates on OS X.
+ * More detailed information is available at the following address:
+ * https://winsparkle.org/
+ *
+ */
+Win32UpdateManager::Win32UpdateManager() {
+	const char *appcastUrl = "https://www.scummvm.org/appcasts/macosx/release.xml";
+
+	win_sparkle_set_appcast_url(appcastUrl);
+    win_sparkle_init();
```
sev-: Please fix indentation here. Spaces instead of tabs.

sev-:
Really cool, thank you! Looking for more comments.

bluegr:
Thanks :) Indentation fixed

uruk:
I can only repeat Eugene, thanks for your work! :)

uruk:
Any objections against merging this? I guess since it's not an engine, it doesn't need the consent of the whole core team.

bluegr:
Please note that the Makefile will need to be updated accordingly... Theoretically, it should work by specifying --enable-updates and --enable-sparkle, but I have only tested this with MSVC

sev-:
Thanks, merging.

criezy:
Regarding the last note from bluer, I don't think specifying --enable-updates and --enable-sparkle will work when using configure+make. Currently in configure sparkle is always set to 'no' when not compiling for darwin (thus the --enable-sparkle will have no impact when compiling for Windows) and the USE_SPARKLE define will only be added when compiling for darwin.

rootfather:
@criezy, I used `./configure --enable-updates --enable-sparkle` and compiled with mingw-w64 on Windows 10. The configure skript confirms that updates are enabled, and I also get the GUI dialog about enabling the automatic updates on first startup and in the Options dialog.

I have to test with MSVC later.

criezy:
@rootfather: There are two defines added by configure to use sparkle updates, 'USE_UPDATES' and 'USE_SPARKLE'. I suspect that currently when using `./configure --enable-updates --enable-sparkle` on Windows will only add the `USE_UPDATES` define, but the `USE_SPARKLE` will be missing. This probably results in the GUI dialog being added, but automatic updates will not work since the winsparkle backend will be missing.

rootfather:
criezy, thanks for your reply. Is there any way to test if the winsparkle backend is integrated?

Btw, what should happen if I press the "Check now" button in the update options dialog? On my build, it simply does nothing. Is there any pop-up window that should appear?

criezy:
When checking for updates, a popup window will only appear if a new update is found I think. So it appearing to do nothing for you is not surprising and doesn't really help to know if the win sparkle backend is integrated or not.

Running configure generates a config.h file. You can check the content of this file, and in particular if it contains a line `#define USE_SPARKLE`. If that line is not present then you will be missing the sparkle backend in your build.

You can also look at what the configure script tells you. If it all works you should have:
`Checking for Sparkle... yes`
and a bit below:
`Building updates support... Sparkle`

I suspect that on Windows it currently skips the Sparkle check and you only get:
`Building updates support... yes`

rootfather:
Yep, you're right.

There's no `#define USE_SPARKLE` in config.h and I only get `Building updats support ...yes` without any hint about Sparkle.

criezy:
@rootfather: Can you try with the following changes: https://github.com/criezy/scummvm/commit/c13b2f94ac201314f7db97b2e57359804f230cb7

I added auto-detection, so try with `./configure --enable-updates` first to see if auto detection works.
Then you can also try with `./configure --enable-updates --enable-sparkle`

I also added support for --with-sparkle-prefix, and you may need to use it to provide the path to where you have your Sparkle library. It assumes the layout of the download:
```
WinSparkle-0.5.1
  |_ include
     |_ winsparkle.h
  |_ Release
     |_ WinSparkle.dll
     |_ WinSparkle.lib
  |_ x64
     |_ Release
        |_ WinSparkle.dll
        |_ WinSparkle.lib
```
And you need to pass the path to the root directory:
`./configure --enable-updates --with-sparkle-prefix=C:\path\to\WinSparkle-0.5.1`

rootfather:
@criezy, thanks for the update!

`./configure --enable-updates --with-sparkle-prefix=C:\path\to\WinSparkle-0.5.1`
results in `#undef USE_SPARKLE` in the config.h file.

`./configure --enable-updates --enable-sparkle --with-sparkle-prefix=...`
gives me the proper `Building updates support... Sparkle` message and a nice `#define USE_SPARKLE`.

criezy:
@rootfather, thank you for the tests. Could you also provide the generated config.mk? This might help understand why the compilation doesn't work.

rootfather:
@criezy, here you go:
https://gist.github.com/rootfather/04051b8b296c4d2aa183914f6bbfc12f#file-config-mk

rootfather:
@criezy, I am now a step further.

Using `/c/scummvm/scummvm_libs_2015/winsparkle` as sparkle prefix seems to do the trick - auto detection of Sparkle works and it compiles nicely, but after linking, I get the following error message:
```
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/5.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: skipping incompatible C:/ScummVM/scummvm_libs_2015/winsparkle/Release/WinSparkle.lib when searching for -lWinSparkle
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/5.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: skipping incompatible C:/ScummVM/scummvm_libs_2015/winsparkle/Release/WinSparkle.dll when searching for -lWinSparkle
```

criezy:
Thank you. I committed my changes to configure.
And yes, the messages are normal. In configure it adds the path to both the 32 bits and 64 bits DLLs, and thus the linker ignores one of the two and you get these messages.

PR #765, opened 2016-05-22
ANDROID: Add support for Leanback Launcher on Android TV

mjg59:
Android TV needs some modifications to the manifest and a new icon in order
for apps to appear in the launcher.


fuzzie:
This looks like it should be pretty harmless, at least.

sev-:
Could you please make the icon background orange? Just like on our other similar icons: https://github.com/scummvm/scummvm/blob/master/dists/androidsdl/scummvm/banner.png

mjg59:
Updated the icon and added an extra commit that makes it possible to get at the in-game menu

sev-:
Thanks a lot! Merging.

fuzzie:
This broke buildbot because some of the new manifest elements (isGame, banner) aren't supported by the old SDK it's using. We should upgrade it.

magic-ian:
The Google Play Store does not surface this as Android TV compatible, so there must still be something missing perhaps in the manifest or elsewhere that prevent installation from this source. Enjoying it otherwise from the direct download on Android TV. Added a comment to the [following existing ticket](https://bugs.scummvm.org/ticket/11932#comment:4).

PR #767, opened 2016-05-29
GUI: Make (C) sign consistent in all engines

raziel-:
The About dialog shows lower case copyright signs (c) on 9 engines, while the majority of engines displays it upper case (C) - so i guess upper case is the way to go.
Changed lower case to upper case on affected engines.

Question: Is there a reason the ASCII char © isn't used?
Is it because some backends can't display those or is ScummVM's font limited or maybe the compiler chokes?

### DIFF
--- a/engines/access/detection.cpp
+++ b/engines/access/detection.cpp
@@ -94,7 +94,7 @@ public:
 	}
 
 	virtual const char *getOriginalCopyright() const {
-		return "Access Engine (c) 1989-1994 Access Software";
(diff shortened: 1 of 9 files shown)


sev-:
Nice, catch, thanks!

somaen:
Re: your question:

© isn't ASCII, it is however available in ISO-8859-1, but that means adding non-ASCII characters to our source code, which I'm not sure if we do or not.

raziel-:
@somaen 

So no real gain in adding those.

Thank you

PR #769, opened 2016-06-15
SCI32: New audio engine (kDoAudio, kDoSound digital effects)

csnover:
This PR provides a new audio engine for SCI32 that implements the software mixer and features required for most SCI2.1 games.

In order to support WAVE audio without duplicating a bunch of code, this PR also includes minor changes to the common audio subsystem, allowing WAV files containing uncompressed audio to be seekable.

It may also be possible for this audio engine to totally replace the existing audio engine used for SCI16 games, since the major differences seem to be that (1) SCI1.1 only supported one digital audio playback channel (this is configurable already), and (2) it had extra commands for CD audio playback and queued sample playback, which don’t exist in SCI32. But, for now at least, the two are kept separate, to avoid regressing any functionality that I can’t test.

The code to support playback of digital samples via kDoSound is a little hacky, but actually less hacky than what was in the original engine. I heard that the MIDI code was going to be overhauled at some point in the future anyway, so these additions may look a little bit scrappy but seem to work OK, and will hopefully get redone when the MIDI subsystem is redone.

There is one known problem with KQ7, the intro animation plays duplicate samples because of the way the script works (on each loop though the VM, if the displayed cel matches a number, then it calls to play the sample; the cels are shown for more than one frame, so the audio gets restarted); I think this might have only worked in the original engine because the audio engine ran mostly in the main thread. Anyway, this is a minor problem that can hopefully be fixed when focusing more on games other than SQ6 in the future.


engines/sci/sound/audio32.cpp
```diff
@@ -0,0 +1,920 @@
...
+#include "sci/sound/decoders/sol.h" // for makeSOLStream
+
+namespace Sci {
+
+bool detectSolAudio(Common::SeekableReadStream &stream) {
+	const size_t initialPosition = stream.pos();
+
+// TODO: Resource manager for audio resources reads past the
+// header so even though this is the detection algorithm
+// in SSCI, ScummVM can't use it
```
bluegr: The SCI resource manager checks the type of each resource that is loaded. Thus, the resource type check does happen in ScummVM, but it occurs inside the resource manager, not in the SOL audio detection function

engines/sci/sound/audio32.cpp
```diff
@@ -0,0 +1,920 @@
...
+			continue;
+		}
+
+		if (channel.robot) {
+			// TODO: Robot audio into output buffer
+			continue;
+		}
+
+		if (channel.vmd) {
+			// TODO: VMD audio into output buffer
```
bluegr: VMD videos are actually not played within the engine - unlike robot videos, they're full screen videos, and can be placed in specific screen areas. Thus, in ScummVM, VMD videos run within the video decoder interface, so their audio handling occurs outside the engine. Thus, this VMD audio handling code is probably not needed with our implementation

engines/sci/sound/audio32.cpp
```diff
@@ -0,0 +1,920 @@
...
+			// engine was supposed to allow counter loops at one
+			// point, but ended up only using loop as a boolean.
+			loop = (bool)argv[5].toSint16();
+		}
+
+		if (argc < 7 || argv[6].toSint16() < 0 || argv[6].toSint16() > Audio32::kMaxVolume) {
+			volume = Audio32::kMaxVolume;
+
+			if (argc >= 7) {
+				monitor = true;
```
bluegr: I'm confused here. Is there a specific call that marks a channel to be monitored for signals? It seems like the 7th argument is the volume, and you're checking for all values that are out of the ordinary? Isn't there a specific mark, like -1?

Also, you are checking for all calls with 7 or more arguments... Shouldn't this be exactly 7 arguments?

engines/sci/sound/audio32.cpp
```diff
@@ -0,0 +1,920 @@
...
+			if (candidate.robot == false && candidate.id == searchId) {
+				return i;
+			}
+		}
+	}
+
+	return kNoExistingChannel;
+}
+
+int16 Audio32::findChannelById(const ResourceId resourceId, const reg_t soundNode) const {
```
bluegr: So, findChannelByArgs() can find a channel for both Audio and Audio36 resources, but findChannelById() can only find a channel for Audio resources. Please, indicate that in the function comments in the header file.

engines/sci/sound/audio32.cpp
```diff
@@ -0,0 +1,920 @@
...
+	if (_numActiveChannels == _channels.size()) {
+		warning("Audio mixer is full when trying to play %s", resourceId.toString().c_str());
+		return 0;
+	}
+
+	// NOTE: SCI engine itself normally searches in this order:
+	//
+	// For Audio36:
+	//
+	// 1. First, request a FD using Audio36 name and use it as the
```
bluegr: Please, indicate that FD is "File Descriptor"

engines/sci/sound/audio32.cpp
```diff
@@ -0,0 +1,920 @@
...
+	// descriptor + offset that it can use to read streamed audio, or it
+	// has a memory ID that it can use to read cached audio.
+	//
+	// Here in ScummVM we just ask the resource manager to give us the
+	// resource and we get a seekable stream.
+
+	// TODO: This should be fixed to use streaming, which means
+	// fixing the resource manager to allow streaming, which means
+	// probably rewriting a bunch of the resource manager.
+	Resource *resource = _resMan->findResource(resourceId, true);
```
bluegr: Why is streaming needed here?

engines/sci/sound/audio32.cpp
```diff
@@ -0,0 +1,920 @@
...
+				_monitoredChannelIndex = i;
+			}
+		}
+	}
+
+	// NOTE: SSCI stops the DSP interrupt and frees the
+	// global decompression buffer here if there are no
+	// more active channels
+	if (_numActiveChannels == 0) {
+		_mixer->pauseHandle(_handle, true);
```
bluegr: This indicates that the audio should be stopped.... so why pause the handle, instead of stopping it?

engines/sci/sound/decoders/sol.cpp
```diff
@@ -0,0 +1,269 @@
...
+	if (flags & kCompressed) {
+		if (flags & kStereo && flags & k16Bit) {
+			return new SOLStream<true, true>(new Common::SeekableSubReadStream(stream, initialPosition, initialPosition + dataSize, disposeAfterUse), disposeAfterUse, headerSize, sampleRate, dataSize);
+		} else if (flags & kStereo) {
+			return new SOLStream<true, false>(new Common::SeekableSubReadStream(stream, initialPosition, initialPosition + dataSize, disposeAfterUse), disposeAfterUse, headerSize, sampleRate, dataSize);
+		} else if (flags & k16Bit) {
+			return new SOLStream<false, true>(new Common::SeekableSubReadStream(stream, initialPosition, initialPosition + dataSize, disposeAfterUse), disposeAfterUse, headerSize, sampleRate, dataSize);
+		} else {
+			return new SOLStream<false, false>(new Common::SeekableSubReadStream(stream, initialPosition, initialPosition + dataSize, disposeAfterUse), disposeAfterUse, headerSize, sampleRate, dataSize);
+		}
```
bluegr: Why don't you just pass the stereo and 16-bit flags as parameters to the constructor?

bluegr:
Nice work overall!

I haven't checked the differences between SCI16 and SCI32 audio, but if this code suites us for audio in SCI16, we should consider merging the two in the future, as you mentioned.

engines/sci/sound/audio32.cpp
csnover: There is no separate call for monitoring. The channel is set when `play` or `waitForPlay` is called by the game engine, according to the number of arguments, just like this code does it. While the signature checking in ScummVM means there will never be more than seven arguments to this call, it is identical to SSCI logic and there is no harm in using the same operator (maybe SCI3 has more arguments? I don’t know).

engines/sci/sound/audio32.cpp
csnover: For memory efficiency, especially considering low-memory targets. Audio data can be fairly large and there isn’t a good reason to load it all into memory instead of streaming, since the audio engine already treats all audio as streamed. Some games, like Phantasmagoria, have audio samples that are 12MiB—and there can be 5 or 8 streams at once, which means an even larger audio footprint if we’re mixing several sounds together (background music + ambient audio + speech + sfx, for instance).

engines/sci/sound/audio32.cpp
csnover: Stopping the handle releases memory that will then just get allocated again when the handle is restarted later. This happens frequently and has the potential to cause memory fragmentation. It also wastes CPU reinitialising a channel over and over, whereas pausing just sets a flag.

engines/sci/sound/decoders/sol.cpp
csnover: Because using a template allows elimination of conditionals at runtime. It’s not hot enough code to likely matter, but there’s no real ergonomic benefit to not using a template here.

engines/sci/sound/audio32.cpp
csnover: Thanks for noticing this; while `findChannelById` is used only by sound effects calls from `kDoSound` right now, it should be usable regardless of the audio resource type, especially since later engine versions change kernel calling conventions which means `findChannelById` is probably going to be used more frequently later.

engines/sci/sound/audio32.cpp
csnover: Since I have not looked much at all about how games use VMD I can’t be sure whether this is required or not, just that it exists in the engine and this is where it would go. VMD playback appears to allow yielding back to the VM periodically, which potentially means that other audio may be played and mixing of the VMD into the SCI audio mixer is critical for correct audio playback. Until VMD implementation is done and we know the audio can always be run independently of the SCI audio mixer, this is just a reminder that the VMD audio playback would go here if it needs to go through the SCI mixer.

csnover:
This is landed!

PR #771, opened 2016-06-25
SCI32: Store bitmap resources in dynamic instead of hunk memory

bluegr:
Hunk memory is volatile, and is not preserved in saved games. The
original stores bitmap resources in a separate kind of memory, which
is preserved in saved games. We preserve the same behavior using
dynamic memory. This allows dynamic text objects to be preserved in
saved games, in scenes where text is dynamically created and shown
together with other scene objects. An example is the movie
title in the cinema marquee in the first scene of SQ6

### DIFF
--- a/engines/sci/engine/kgraphics32.cpp
+++ b/engines/sci/engine/kgraphics32.cpp
@@ -539,7 +539,7 @@ reg_t kBitmapCreate(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kBitmapDestroy(EngineState *s, int argc, reg_t *argv) {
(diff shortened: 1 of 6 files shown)


csnover:
I’m glad this resolves some save game problems and is a pretty easy fix, though wasn’t the plan to add a new bitmap segment to the VM?

bluegr:
Yeah, that would be another approach. I can do it in a separate pull request, and we can then decide which one to go with

bluegr:
The requested functionality has now been implemented with a new segment type in pull request #791. Closing this one

PR #772, opened 2016-06-29
Bugfixes

BenCastricum:
Please have a look at these suggested changes. It fixes 2 bugs and one janitorial commit (savegame -> saved game). Feel free to cherry-pick only the ones you like.


engines/scumm/saveload.cpp
```diff
@@ -1137,6 +1137,11 @@ void ScummEngine::saveOrLoad(Serializer *s) {
 	if (s->isLoading() && s->getVersion() < VER(14))
 		upgradeGfxUsageBits();
 
+	// When loading, reset the setShakePos. Fixes bug #7141 Partially
```
bluegr: Why partially?

wjp: And as a meta-comment, references to bugs generally fit better in commit messages than in comments. (Since the code comment should describe the code as it is, not compared to some older broken state.)

BenCastricum: @bluegr Bug #7141 can be split up into 2 issues. The shakePos is not reset on loading a saved game, and when shakePos is not 0, mousepointer is not handled correctly. The second part is also present during normal gameplay, but less visible.

@wjp I changed the commit messages/comments a bit.

sev-:
Now it is in a good shape. Thanks, merging.

BenCastricum:
Thanks, bugs #6886 and #7141 can also be closed now.

PR #773, opened 2016-06-30
SCI32: Graphics rendering code updates

csnover:
The goal with this pass was primarily to get documentation in place, and try to clean up by giving variables more descriptive names, but while I was working on doing this I discovered some rendering errors in Torin when scrolling the transparent transcript window (opened by the book icon at the bottom-left). This sent me on a wild goose chase for a while that had me first try implementing `kPlaneTypeTransparentPicture`, before discovering that `_moved` and `_updated` were reversed in a couple of places and that’s why things were screwy. (As it turns out that swap was messing up rendering in quite a few games—just not the ones we’ve been mostly looking at!)

I also discovered that some 320x200 constant values were still existing in games with high-resolution scripts, so it seems that the ScreenItem rect calculation code doesn’t care whether the cel and game script resolutions are mismatched so much; rather, it seems to care whether or not the cels are low-resolution.

I also found looks like some late SCI games seem to use the “hi res pics” rendering path, so left that in for now, although it’s still always disabled (it still acts weird when it’s turned on in SQ6 and GK1—not sure if that is because it’s buggy/wrong or because it just can’t handle the way those games build their rendering trees).

### DIFF
--- a/engines/sci/graphics/screen_item32.h
+++ b/engines/sci/graphics/screen_item32.h
@@ -236,6 +236,24 @@ public:
 		return false;
 	}
 
+	inline bool operator>(const ScreenItem &other) const {
+		if (_priority > other._priority) {
+			return true;
+		}
+
+		if (_priority == other._priority) {
+			if (_position.y + _z > other._position.y + other._z) {
+				return true;
+			}
+
+			if (_position.y + _z == other._position.y + other._z) {
+				return _object > other._object;
+			}
+		}
+
+		return false;
+	}
+
 	/**
 	 * Calculates the dimensions and scaling parameters for
 	 * the screen item, using the given plane as the parent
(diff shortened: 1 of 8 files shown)


engines/sci/graphics/screen_item32.h
```diff
@@ -236,6 +236,24 @@ class ScreenItem {
...
+			}
+
+			if (_position.y + _z == other._position.y + other._z) {
+				return _object > other._object;
+			}
+		}
+
+		return false;
+	}
+
```
bluegr: Same behavior as in SCI16, check sortHelper() in animate.cpp:181

wjp:
Nice catch on the `_moved`/`_updated` mixup. Looks good overall.

PR #774, opened 2016-07-02
Add ScrollContainerWidget

Tkachov:
The following adds a special ScrollContainerWidget, which is used to create scrollable areas with widgets in it. Similarly to TabWidget, it's added as `<widget>` in .stx files, and its contents are described in a separate `<dialog>`. In the code one must create a ScrollContainerWidget and pass it as boss to all its child widgets.

If inner widgets require more space than container get provide, it shows a scrollbar, and inner widgets are clipped. These commits add clipping versions of ThemeEngine's and VectorRenderer's methods. Original non-clipping methods might be faster, so they are automatically used when no clipping is required.

Clipping is on for all widgets, not only for those kept inside Container. Widgets use their boss's area (absolute coordinates and width, height) as clipping rectangle. The difference is that Container provides a scrollbar, so all widgets are available to user. Without this container widgets which get out of boss's area get clipped and thus might become invisible.

Example of Container clipping


sev-:
I reviewed the code, and it is of a pretty high quality. Merging.

wjp:
@csnover points out on IRC that this breaks some dialog layouts. I took some screenshots: http://www.usecode.org/scummvm/options1.png http://www.usecode.org/scummvm/options2.png

Tkachov:
That's very strange. I've seen such bug in lowres (not sure how to fix it yet), but I believe it was there before ScrollContainer, because there was a "padding-bottom = -8" hack applied, which I removed in this PR.

No such bug - or fix - was found in default 640x400.

Took a few screenshots just now:

640x400
320x200

sev-:
@csnover May it happen that you have an old theme .zip somewhere?

wjp:
I probably did, but @csnover probably didn't. Snover's screenshots on IRC show a more subtle overlap of shadow with the buttons.

sev-:
@wjp, do I understand correctly, that you see the problem too, or just @csnover?

Tkachov:
Looks like I've discovered the source of the problem I see. The thing is I've updated `ThemeEval` to create `ThemeLayoutTabWidget` instead of simple `ThemeLayoutWidget`. That's needed because `TabWidget` takes some space for its tabs, so children widgets should get less space, but layouts were not informed about that. `ThemeLayoutTabWidget` is: it decreases the layout height of `TabWidget` correspondingly.

But I've made `ThemeEval` create it only if widget's type in stx is set to "TabWidget". Obviously, I wasn't careful enough and didn't specify the type everywhere. It's set in `scummmodern_layout.stx` for "GlobalOptions" and in `scummmodern_layout_lowres.stx` for "GameOptions". So, it should be set for all three dialogs - "GlobalOptions", "GameOptions" and "FluidSynthSettings" - in all stx files, including the classic theme this time.

Moreover, Eugene noticed another height-related bug: `TabWidget`'s contents are clipped "too early". To fix that, we should override `TabWidget`'s `getHeight()` to `return _h + _tabHeight;`.

Plus, as I didn't touch the classic theme in this PR, I didn't fix the "-8 bottom padding" hack in lowres classic theme, which seems to be there.

PR #775, opened 2016-07-02
SCI32: Improve FileIO read-write support

wjp:
It adds a ReadStream/WriteStream to support this, replacing the previous VirtualIndexFile which required reimplementation of most read/write functions.

This new implementation fixes corrupting phantsg.dir when creating new games in Phantasmagoria, and updating swat.dat on PQ:SWAT startup (and elsewhere).

### DIFF
--- a/engines/sci/engine/file.cpp
+++ b/engines/sci/engine/file.cpp
@@ -22,6 +22,7 @@
 
 #include "common/savefile.h"
 #include "common/stream.h"
+#include "common/memstream.h"
 
 #include "sci/sci.h"
 #include "sci/engine/file.h"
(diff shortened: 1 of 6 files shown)


sev-:
@wjp: What do you think about the similar functionality implemented here:

https://github.com/Tkachov/scummvm/commit/e83a3e27f636e6c17c2e895950dd1fc02496f787#diff-0651c045bfb7e1808db63a2269683be0R212

Maybe a better name should be used for this upcoming one ring buffer-based? E.g. as I understand, you need to keep the written file, while the one by Tkachov discards the data as soon as it was read.

wjp:
The main use of `MemoryDynamicRWStream` in my branch here is modifying an existing (small) file and then writing the modified file back to disk, so it should indeed keep all contents as you say.

The FIFO that Tkachov implements seems to have quite different semantics, including different `read` behaviour and separate non-seekable positions.

I think the `MemoryReadWriteStream` name is indeed a bit misleading, since none of our other `Read/WriteStreams` discard data when it is read (OTOH, `AudioStreams` do.), so I think I'd indeed be in favour of a different name there. Maybe something with FIFO, queue, ...?

wjp:
I kind of like `DataStream` or `ByteStream` too, I think.

wjp:
`VirtualIndexFile` was indeed corrupting `phantsg.dir`; extra nul bytes were appearing in there, which broke creating new games. I didn't bother tracking down the exact cause, as it was duplicating existing functionality from our common stream classes and I preferred just removing the duplication.

As the TODO also states, it might be worth it to keep that file open, and that can be added easily enough if necessary. However, when looking at how it's used currently, I didn't see all that much accesses to it at run-time, so I left this open to re-evaluate. It seemed plausible that the excessive accesses are (or were) in fact caused by other bugs that may now be fixed.

Both `MemoryDynamicRWStream` and `SaveFileRewriteStream` are potentially generally enough to put in common, true. Are you aware of other engines that could use either of them?

engines/sci/engine/file.cpp
```diff
@@ -32,6 +33,110 @@
...
+	} else {
+		_changed = true;
+	}
+}
+
+SaveFileRewriteStream::~SaveFileRewriteStream() {
+	writeBack();
+}
+
+void SaveFileRewriteStream::writeBack() {
```
csnover: `commit` or `commitToDisk`, for clarity?

engines/sci/engine/file.cpp
```diff
@@ -91,6 +196,17 @@ reg_t file_open(EngineState *s, const Common::String &filename, int mode, bool u
 		break;
 	}
 
+#ifdef ENABLE_SCI32
+	if (mode != _K_FILE_MODE_OPEN_OR_FAIL && (
+	    (g_sci->getGameId() == GID_PHANTASMAGORIA && filename == PHANTASMAGORIA_SAVEGAME_INDEX) ||
```
csnover: Is there a reason to keep this `#define PHANTASMAGORIA_SAVEGAME_INDEX` around?

engines/sci/engine/file.cpp
```diff
@@ -126,15 +242,7 @@ reg_t file_open(EngineState *s, const Common::String &filename, int mode, bool u
...
-	// Find a free file handle
-	uint handle = 1; // Ignore _fileHandles[0]
-	while ((handle < s->_fileHandles.size()) && s->_fileHandles[handle].isOpen())
-		handle++;
-
-	if (handle == s->_fileHandles.size()) {
-		// Hit size limit => Allocate more space
-		s->_fileHandles.resize(s->_fileHandles.size() + 1);
-	}
+	uint handle = findFreeFileHandle(s);
```
csnover: There is some `VIRTUALFILE_HANDLE_START` and `VIRTUALFILE_HANDLE_END` a bit below here at `getFileFromHandle` (line 255) but I don’t actually see anything that would give a handle in that range. Is this a game script thing, or just dead code?

wjp: There's still `VIRTUALFILE_HANDLE_SCIAUDIO`.

wjp:
New commits take comments into account. I also realized it should fallback to try to read from regular non-save datafiles.

bluegr:
Regarding the move to common: I am not aware of other games that would use this functionality, but this class is generic enough to warrant a move to common code. There could be games that use this functionality.

PR #776, opened 2016-07-03
SCI32: kPlayVMD

csnover:
This *should* be the complete set of VMD functionality, modulo hot rectangles and possibly a couple extra subops used by Phant1 which are currently dummied out (to see if they’re ever called). Video black lines is added as an optional game engine feature that is turned off by default.


engines/sci/detection_tables.h
```diff
@@ -4330,31 +4331,31 @@ static const struct ADGameDescription SciGameDescriptions[] = {
 		{"resmap.000", 0, "e55c3097329b3c53752301e01c6af2fb", 9787},
 		{"ressci.000", 0, "118f9bec04bfe17c4f87bbb5ddb43c18", 56127540},
 		AD_LISTEND},
-		Common::DE_DEU, Common::kPlatformWindows, ADGF_UNSTABLE | ADGF_CD, GUIO4(GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
+		Common::DE_DEU, Common::kPlatformWindows, ADGF_UNSTABLE | ADGF_CD, GUIO5(GAMEOPTION_ENABLE_BLACK_LINED_VIDEO, GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
```
sev-: I recommend to put these into #defines for the compactness sake.

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+		g_sci->getEngineState()->speedThrottler(_decoder->getTimeToNextFrame());
+		g_sci->getEngineState()->_throttleTrigger = true;
+		if (_decoder->needsUpdate()) {
+			renderFrame();
+		}
+
+		const int currentFrameNo = _decoder->getCurFrame();
+
+		if (
```
sev-: This doesn't conform to our code formatting guidelines.

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+	_startColor(0),
+	_endColor(255),
+	_blackPalette(false),
+
+	_boostPercent(100),
+	_boostStartColor(0),
+	_boostEndColor(255),
+
+	 _showCursor(false) {}
```
sev-: An extra indentation?

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	if (!_isInitialized) {
+		_isInitialized = true;
+
+		if (!_showCursor) {
+			g_sci->_gfxCursor->kernelHide();
+		}
+
+		Common::Rect vmdRect(
```
sev-: Please fix code formatting.

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,393 @@
...
+		g_sci->getSciDebugger()->onFrame();
+		g_sci->_gfxFrameout->frameOut(true);
+		g_sci->_gfxFrameout->throttle();
+	}
+}
+
+void VMDPlayer::fillPalette(Palette &palette) const {
+	const byte *vmdPalette = _decoder->getPalette() + _startColor * 3;
+	for (uint16 i = _startColor; i <= _endColor; ++i) {
+		int16 r = *vmdPalette++;
```
sev-: Shouldn't it be rather uint16?

engines/sci/graphics/video32.h
```diff
@@ -0,0 +1,312 @@
...
+	bool _isOpen;
+
+	/**
+	 * Whether or not a VMD player has been initialised
+	 * with `init`.
+	 */
+	bool _isInitialized;
+
+	/**
+	 * For VMDs played with the `kEventFlagYieldToVM` flag,
```
sev-: Kudos for great documentation.

engines/sci/detection_tables.h
csnover: Excellent idea. I will do this in a separate commit once this is landed, for SCI32. It should also be done for SCI16 games too probably, but it’s really tedious, so I don’t want to do any more of it right now… :)

engines/sci/graphics/video32.cpp
csnover: Good catch. Xcode invisibles are [broken](http://stackoverflow.com/questions/9244249/how-can-i-change-the-color-of-invisibles-in-xcode4) and so sometimes things like this sneak in despite my best efforts. :\

engines/sci/graphics/video32.cpp
csnover: I don’t know what you are referring to here, this does follow [the written conventions](http://wiki.scummvm.org/index.php/Code_Formatting_Conventions).

engines/sci/graphics/video32.cpp
csnover: Again, I’m unsure of what guideline you are referring to here. Could you please point me to it? The only one I can figure is the common sense rule, which this seems to follow; the condition is complex, so it is split to multiple lines for clarity, and indented with the closing bracket on its own line so the end of the condition is clear.

engines/sci/graphics/video32.cpp
sev-: I am talking about this (github eats leading spaces)


>   foo(
>      blah,
>      bleh,
>      baz
>   );

If you want to align the params vertically, then it should be:

> foo(blah,
>      bleh,
>      baz);

engines/sci/graphics/video32.cpp
csnover: Hmm. I don’t *think* it should matter, the largest calculated value should always be well below the maximum allowed value in either case.

engines/sci/graphics/video32.cpp
csnover: > github eats leading spaces

See [creating and highlighting code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/) :)

> If you want to align the params vertically, then it should be: […]

Hm… I don’t see it in the guidelines to do it that way, and there are around 420 instances of this ‘wrong’ style in ScummVM (grepping for `\($`).

I don’t really want to bikeshed on this, so if this is supposed to be some hard-and-fast rule about how to author complex conditions or function calls that need to spill into multiple lines for clarity, it should definitely be added into the guidelines…

sev-: My primary concern is about the hanging ');'.

grep -r '^\s*[)];$' engines/*|grep -v matches|wc

gives me 47 instances, 15 of which are in SCI. That apparently was overlooked and slipped in.

If you want to raise this new extension to our coding guidelines, please use -devel, in the meantime, this needs to be corrected.

sev-: And speaking of the bikesheding (pardon?), we had a lot of discussions in the past, and it is very important to have common code formatting in a 1.8M+ SLOC codebase. Otherwise it quickly will turn into an unmaintainable mess.

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,364 @@
...
+			return kEventFlagEnd;
+		}
+	}
+
+	_decoder->pauseVideo(false);
+
+	if (flags & kEventFlagReverse) {
+		// NOTE: This flag may not work properly since SCI does not care
+		// if a video has audio, but the VMD decoder does, for some reason.
+		assert(_decoder->setReverse(true));
```
dreammaster: I can attest from personal experience that it's bad form to put a call that has to happen inside an assert statement, since when the codebase is built in release mode, the entire line will be removed. You should instead get the result from the function, and assert the value on a separate line.

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,390 @@
...
+		if (_yieldInterval > 0 &&
+			currentFrameNo != _lastYieldedFrameNo &&
+			(currentFrameNo % _yieldInterval) == 0
+		) {
+			_lastYieldedFrameNo = currentFrameNo;
+			stopFlag = kEventFlagYieldToVM;
+			break;
+		}
+
+		const SciEvent event = _eventMan->getSciEvent(SCI_EVENT_MOUSE_PRESS | SCI_EVENT_KEYBOARD);
```
wjp: While in this case I think there aren't many events you could be missing, this way of calling `getSciEvent` lets unhandled events queue up instead of being discarded.

wjp: Oh wait, there's `MOUSE_RELEASE`.

csnover: That’s a point, though it was the same way in the original engine.

…actually, re-checking this, it doesn’t look like SSCI *ever* consumed the events in VMD playback, just peeked. So this does need to be updated… to queue *more* events.

wjp:
Other than these few minor points, looks good. Nice work!

engines/sci/detection_tables.h
```diff
@@ -1727,55 +1728,55 @@ static const struct ADGameDescription SciGameDescriptions[] = {
...
+		Common::EN_ANY, Common::kPlatformWindows, ADGF_UNSTABLE | ADGF_CD, GUIO5(GAMEOPTION_ENABLE_BLACK_LINED_VIDEO, GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
 
 	// King's Quest 7 - English DOS (from FRG)
 	// SCI interpreter version 2.100.002, VERSION file reports "2.00b"
 	{"kq7", "", {
 		{"resource.map", 0, "8676b0fbbd7362989a029fe72fea14c6", 18709},
 		{"resource.000", 0, "51c1ead1163e19a2de8f121c39df7a76", 200764100},
 		AD_LISTEND},
-		Common::EN_ANY, Common::kPlatformDOS, ADGF_UNSTABLE | ADGF_CD, GUIO4(GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
+		Common::EN_ANY, Common::kPlatformDOS, ADGF_UNSTABLE | ADGF_CD, GUIO5(GAMEOPTION_ENABLE_BLACK_LINED_VIDEO, GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
```
bluegr: The DOS version of KQ7 does not use VMD videos, AFAIK. Only the Windows version has VMD videos. Will double-check this later on today

engines/sci/detection_tables.h
```diff
@@ -4230,39 +4231,39 @@ static const struct ADGameDescription SciGameDescriptions[] = {
...
+		Common::EN_ANY, Common::kPlatformDOS, ADGF_CD | ADGF_UNSTABLE, GUIO5(GAMEOPTION_ENABLE_BLACK_LINED_VIDEO, GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
 
 	// Space Quest 6 - English DOS/Win3.11 CD ver 1.11 (from FRG)
 	// SCI interpreter version 2.100.002 (just a guess)
 	{"sq6", "", {
 		{"resource.map", 0, "e0615d6e4e10e37ae42e6a2a95aaf145", 10528},
 		{"resource.000", 0, "c4259ab7355aead07773397b1052827d", 41150806},
 		AD_LISTEND},
-		Common::EN_ANY, Common::kPlatformDOS, ADGF_CD | ADGF_UNSTABLE, GUIO4(GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
+		Common::EN_ANY, Common::kPlatformDOS, ADGF_CD | ADGF_UNSTABLE, GUIO5(GAMEOPTION_ENABLE_BLACK_LINED_VIDEO, GUIO_NOASPECT, GAMEOPTION_PREFER_DIGITAL_SFX, GAMEOPTION_ORIGINAL_SAVELOAD, GAMEOPTION_FB01_MIDI)	},
```
bluegr: SQ6 does not use VMD videos IIRC. Will verify this later on today, as well...

engines/sci/graphics/video32.cpp
bluegr: Agree with dreammaster... MSVC in particular removes all assert() calls in release binaries. Please, only check for variables inside asserts, do not call other functions in them.

wjp: It's perfectly fine to call functions, just not functions with (relevant) side effects.

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,390 @@
...
+	EventFlags stopFlag = kEventFlagNone;
+	while (!g_engine->shouldQuit()) {
+		if (_decoder->endOfVideo()) {
+			stopFlag = kEventFlagEnd;
+			break;
+		}
+
+		g_sci->getEngineState()->speedThrottler(_decoder->getTimeToNextFrame());
+		g_sci->getEngineState()->_throttleTrigger = true;
+		if (_decoder->needsUpdate()) {
```
bluegr: Hm... why do you need to delay each frame using the speed throttler? VMD rendering occurs outside of the engine's main loop, so we can just delay by getTimeToNextFrame() here

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+	EventFlags stopFlag = kEventFlagNone;
+	while (!g_engine->shouldQuit()) {
+		if (_decoder->endOfVideo()) {
+			stopFlag = kEventFlagEnd;
+			break;
+		}
+
+		g_sci->getEngineState()->speedThrottler(_decoder->getTimeToNextFrame());
+		g_sci->getEngineState()->_throttleTrigger = true;
```
bluegr: Hm... why do you need to delay each frame using the speed throttler? VMD rendering occurs outside of the engine's main loop, so we can just delay by getTimeToNextFrame() here

engines/sci/graphics/video32.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+		if ((flags & kEventFlagEscapeKey) && event.type == SCI_EVENT_KEYBOARD && event.character == SCI_KEY_ESC) {
+			stopFlag = kEventFlagEscapeKey;
+			break;
+		}
+
+		// TODO: Hot rectangles
+		if ((flags & kEventFlagHotRectangle) /* && event.type == SCI_EVENT_HOT_RECTANGLE */) {
+			warning("Hot rectangles not implemented in VMD player");
+			stopFlag = kEventFlagHotRectangle;
```
bluegr: AFAIK, only Phantasmagoria 1 actually uses the Hot rectangles functionality, and that does not use VMD videos. So, in theory, this functionality is not actually used anywhere, so we could safely error out here

engines/sci/detection_tables.h
```diff
@@ -30,6 +30,7 @@ namespace Sci {
 #define GAMEOPTION_SQ4_SILVER_CURSORS       GUIO_GAMEOPTIONS6
 #define GAMEOPTION_EGA_UNDITHER             GUIO_GAMEOPTIONS7
 #define GAMEOPTION_HIGH_RESOLUTION_GRAPHICS GUIO_GAMEOPTIONS8
+#define GAMEOPTION_ENABLE_BLACK_LINED_VIDEO GUIO_GAMEOPTIONS9
```
bluegr: Not all of the games that have been changed have VMD videos, so this option should only be available to the ones that actually havge and use them.

In short, VMD videos are only used in the following games:
- GK2 (full version only)
- LSL7 (full version only)
- Lighthouse (full version only)
- Phantasmagoria 1 (demo and full version)
- RAMA (demo and full version)
- Shivers (full version only)
- SQ6 (full version only)
- Torin's Passage (full version only)

The following DO NOT use VMD videos, so the changes to their detection entries should be reverted:
- GK2 demo: Does not use VMD videos - it uses AVI videos
- KQ7 demo: does not have any video
- KQ7: Does not use VMD videos - it uses AVI and Robot videos
- LSL7 demo: does not have any video
- Lighthouse demo: uses Robot videos
- Mixed Up Mother Goose Deluxe: does not have any video
- Phantasmagoria 2: Does not use VMD videos - it uses AVI DUCK videos
- Police Quest: SWAT (demo and full version): Does not use VMD videos - it uses Robot videos
- Shivers demo: Does not have any video
- SQ6 demo: Does not have any video
- Torin's Passage demo: Does not have any video

PR #777, opened 2016-07-03
SDL: Fix keyboard on macOS, when using SDL2, fix directional keypad

m-kiewitz:
This will hopefully fully fix keyboard issues for macOS / SDL2,
for example when pressing "alt-x".

Also reset .ascii to 0, when Num-Lock is NOT enabled and keypad
directional keys are pressed. This was fixed inside AGI+SCI
previously.

The latter shouldn't cause issues, but in case it does, the
affected engine should get fixed and use keycodes instead.

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -110,13 +110,51 @@ SdlEventSource::~SdlEventSource() {
 int SdlEventSource::mapKey(SDLKey sdlKey, SDLMod mod, Uint16 unicode) {
 	Common::KeyCode key = SDLToOSystemKeycode(sdlKey);
 
+	// Keep unicode in case it's regular ASCII text or in case we didn't get a valid keycode
+	//
+	// We need to use unicode in those cases, simply because SDL1.x passes us non-layout-adjusted keycodes.
+	// So unicode is the only way to get layout-adjusted keys.
+	if (unicode < 0x20) {
+		// don't use unicode, in case it's control characters
+		unicode = 0;
+	} else {
+		// Use unicode, in case keycode is invalid.
+		// Umlauts and others will set KEYCODE_INVALID on SDL2, so in such a case always keep unicode.
+		if (key != Common::KEYCODE_INVALID) {
+			// keycode is valid, check further also depending on modifiers
+			if (mod & (KMOD_CTRL | KMOD_ALT)) {
+				// Ctrl and/or Alt is active
+				//
+				// We need to restrict unicode to only up to 0x7E, because on macOS the option/alt key will switch to
+				// an alternate keyboard, which will cause us to receive Unicode characters for some keys, which are outside
+				// of the ASCII range (e.g. alt-x will get us U+2248). We need to return 'x' for alt-x, so using unicode
+				// in that case would break alt-shortcuts.
+				if (unicode > 0x7E)
+					unicode = 0; // do not allow any characters above 0x7E
+			} else {
+				// We must not restrict as much as when Ctrl/Alt-modifiers are active, otherwise
+				// we wouldn't let umlauts through for SDL1. For SDL1 umlauts may set for example KEYCODE_QUOTE, KEYCODE_MINUS, etc.
+				if (unicode > 0xFF)
+					unicode = 0; // do not allow any characters above 0xFF
+			}
+		}
+	}
+
+	// Attention:
+	// When using SDL1.x, we will get scancodes via sdlKey, that are raw scancodes, so NOT adjusted to keyboard layout/
+	// mapping. So for example for certain locales, we will get KEYCODE_y, when 'z' is pressed and so on.
+	// When using SDL2.x however, we will get scancodes based on the keyboard layout.
+
 	if (key >= Common::KEYCODE_F1 && key <= Common::KEYCODE_F9) {
 		return key - Common::KEYCODE_F1 + Common::ASCII_F1;
 	} else if (key >= Common::KEYCODE_KP0 && key <= Common::KEYCODE_KP9) {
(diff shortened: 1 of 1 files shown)


sev-:
That's the lucky number!

m-kiewitz:
Thank god it's not 666.

bgK:
Using the keycode instead of the unicode value changes existing behavior. For example:
- With my AZERTY keyboard, SDL1, Windows, I am currently able to press AltGr-$ to get the "¤" character (code 164). With these changes, it returns "]".
- Some eastern europe keyboard layouts give unmodded access to characters with unicode values > 255, such as "č". With SDL1, pressing those in text input fields will return the US layout values instead of doing nothing as currently.

bluegr:
Looks good!

Out of curiosity, why is MacOS only affected by this?

m-kiewitz:
@bluegr
Because SDL is inconsistent between platforms (and parts are inconsistent between SDL1+SDL2). I guess we could also change the code to platform specific code, but then we could also simply dump SDL for that and process keyboard input directly. I guess that would be the easier approach.

@bgK 
Hmmm, why is Alt-Gr seen as KMOD_ALT? And are there actually games that use AltGr-anything? The point of those changes is to make Alt/Ctrl-shortcuts work consistently. Those changes could of course also be limited to MacOS, but we would have to make all of that consistent between different platforms (SDL should actually do that already). I'm pretty sure that AltGr-keys are surely also not working consistently atm.

Do Alt/Ctrl-shortcuts still work?

bgK:
The reason this happens only on OS X is not because of inconsistencies related to SDL, but because OS X has characters mapped to key combinations with alt pressed:
http://www.macfreek.nl/memory/File:MacLaptop-US-option.png

Alt-Gr and the right Alt key on US keyboards are not distinguished by SDL. This may or may not be due to the fact that on Windows Alt-Gr is just a shortcut for pressing both Alt and Ctrl at the same time. (This is not the case on Linux)

My point of view on this subject is that the design of the keyboard events is good enough, and it's best to follow it:
- Use the ascii value of keyboard events for text input
- Use the keycode value for shortcuts

I don't think it's a good idea to work around issues caused by engines not following these rules in the backend code.

m-kiewitz:
Eh moment, because of SDL1 issues, possibly every engine that wanted for example Alt-Y and Alt-Z to work properly, would use the .ascii value for those. Because the keycode never worked for those on certain mappings on Windows. That's also why I implemented it that way in AGI/SCI.

This should have been fixed in the backend ages ago, but never was. And changing it now especially setting .ascii to 0 for Alt- / Ctrl- shortcuts will result in engines breaking, which means they will need to be modified as well.

The original issue was caused by SDL not applying mappings properly on at lesat Windows. And I don't want to know what else doesn't work properly on some platforms.

And then there are also issues like NumLock off and keypad usage. That also doesn't return the correct values everywhere. I mean there has to be some way to notice that keycode should be used and not ascii. The easiest approach is probably returning 0 in .ascii. Or INVALID_KEYCODE in keycode for regular ascii. But that also doesn't work consistently.

I personally don't like it at all to fix all sorts of broken things in SDL in the backend code. I think a proper approach would be to handle keyboard stuff ourselves. Sure, that's a bit more work, on the other hand all sorts of engines will need to get changed anyway (see above).

m-kiewitz:
And also: ScummVM engines should get the same keycodes/ASCII for certain key shortcuts.
And that's why SDL should behave somewhat the same on every platform or well we would have to make it platform-independent ourselves. When you press Alt-X in a DOS game, that runs in ScummVM, then the engine + the game should get the exact same data and not one data here for MacOS and another data for Windows and another one for Linux.

SDL should deliver a way, that makes it possible to write software for a bunch of platforms and w/o caring about platforms. But when it returns different keys depending on platform then it doesn't deliver on what it should. It gets complicated of course in case special keys are pressed, but not even regular Alt- / Ctrl- shortcuts return the same values. And when it doesn't work on some platform, we are blamed not SDL.

I think I also fix incorrect/inaccurate/different data for Ctrl-shortcuts in SCI, because I don't get the same data in Linux + Windows.

I mean it would possibly make more sense to fix SDL1 and then bundle that fixed library with ScummVM. But then we could simply do it ourselves and get rid of SDL keyboard processing. Fixing all sorts of flaws in SDL1 (and surely SDL2) is basically the same as changing engines instead of the actual backend code. The fault is in SDL, not in the backend code (well unless one of those fixes are responsible), but I hope you get my point.

sev-:
@m-kiewitz as I understand, we need to change some of the engines as well. Then we probably need to update our Engines HOWTO with the clear instructions, and then start looking into the individual engines. I can cover most of them by myself, but for AGI/SCI I would prefer to rely on those engine developers.

m-kiewitz:
@sev- 
We can't do that atm. For example Windows CE wasn't checked and I doubt that the mapping was done for Windows CE back with SDL1. If we change engines now, we will definitely break some platforms.

sev-:
I plan to remove Windows CE backend as soon as I find time for bringing the Scaler Plugin PR up-to-date and merging it. The backend is dead, so is the platform (unfortunately).

m-kiewitz:
@sev- all sorts of other SDL platforms have to be checked, not just Windows CE.

csnover:
Can this get landed soonish so that there is the maximum amount of time for any necessary engine adjustments between releases?

csnover:
Bueller?

sev-:
OK, it is time to break the buildbot...

bgK:
I still would very much prefer if the hack done to the ascii value when CTRL or ALT are pressed was done only in the engines where it is required instead of in the shared input code.
It introduces non-standard behavior and breaks input in the GUI in some cases as stated in my first comment.

PR #782, opened 2016-07-06
SDL: Fix Windows SDL1 keycodes to use the active keyboard layout

bgK:
On Windows, SDL 1.2 ignores international keyboard layouts and returns keycodes valid for US keyboards. For example, pressing 'A' on an AZERTY keyboards returns a keycode with the value 'q'. This prevents engines from being able to reliably use the keycodes to identify keys.
SDL 2 does not have this problem, nor SDL 1.2 on macOS and Linux/X11.

The proposed change consists in making Windows API calls to obtain the keycode for the scancode sent by SDL using the currently active keyboard layout.

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -517,15 +517,17 @@ bool SdlEventSource::handleKeyDown(SDL_Event &ev, Common::Event &event) {
 
 	SDLModToOSystemKeyFlags(SDL_GetModState(), event);
 
+	SDLKey sdlKeycode = obtainKeycode(ev.key.keysym);
+
 	// Handle scroll lock as a key modifier
-	if (ev.key.keysym.sym == SDLK_SCROLLOCK)
+	if (sdlKeycode == SDLK_SCROLLOCK)
 		_scrollLock = !_scrollLock;
 
 	if (_scrollLock)
 		event.kbd.flags |= Common::KBD_SCRL;
 
 	// Ctrl-m toggles mouse capture
-	if (event.kbd.hasFlags(Common::KBD_CTRL) && ev.key.keysym.sym == 'm') {
+	if (event.kbd.hasFlags(Common::KBD_CTRL) && sdlKeycode == 'm') {
 		if (_graphicsManager) {
 			_graphicsManager->getWindow()->toggleMouseGrab();
 		}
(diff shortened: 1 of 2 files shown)


bluegr:
Looks good at a first glance, but I'm unable to test this, as I don't have a non-QWERTY keyboard

rootfather:
I have a German QWERTZ keyboard. How can I test this?

bgK:
@rootfather According to this piece of code https://github.com/scummvm/scummvm/blob/master/backends/events/sdl/sdl-events.cpp#L548, ctrl-z should quit. With your QWERTZ keyboard on Windows before the fix ctrl-y should quit instead.

rootfather:
Confirmed.

Without the patch: CTRL-Z does nothing, CTRL-Y quits.
With the patch: CTRL-Z quits, CTRL-Y does nothing, as expected.

backends/events/sdl/sdl-events.cpp
```diff
@@ -868,6 +871,42 @@ bool SdlEventSource::handleResizeEvent(Common::Event &event, int w, int h) {
...
+		// Querying the keyboard layout for those would return the base key values
+		// for AZERTY keyboards, which are not numbers. For example, SDLK_1 would
+		// map to SDLK_AMPERSAND. This is theoratically correct but practically unhelpful,
+		// because it makes it impossible to handle key combinations such as "ctrl-1".
+		return keySym.sym;
+	}
+
+	int vk = MapVirtualKey(keySym.scancode, MAPVK_VSC_TO_VK);
+	if (vk) {
+		int ch = (MapVirtualKey(vk, MAPVK_VK_TO_CHAR) & 0x7FFF);
```
bluegr: Why are you discarding the leftmost bit here?

bgK: According to MSDN:
> MAPVK_VK_TO_CHAR : Code is a virtual-key code and is translated into an unshifted character value in the low-order word of the return value. Dead keys (diacritics) are indicated by setting the top bit of the return value. If there is no translation, the function returns 0.

Which means dead keys are not discarded, and their value is that of the diacritic alone. This is what SDL2 does as well.

bluegr: I think it would be best to add what you said in a comment here. It would make things clearer for someone who's reading the code

bgK: There you go.

bluegr:
Great work, thanks!

The usage of MapVirtualKey is pretty self-explanatory:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms646306(v=vs.85).aspx

Merging

PR #783, opened 2016-07-07
DEVTOOLS: Add a CMake project generator to create_project

bgK:
This adds a CMake project generator to create_project. You might wonder why it would be a good idea to make a build project generator output a configuration file for another build project generator. I use it to work on ScummVM with CLion as an IDE. CLion uses CMake build files as its native project model.
The CMake build file generated by create_project could probably be used to generate projects for other tools, but I have not tried it.

The generated CMake files have been tested to produce working builds of ScummVM on the following platforms:
- Linux, CMake 3.5, native GCC 6.1
- Windows, CMake 3.4, MSYS2/MinGW64 GCC 5.4

Other platforms may require a bit of tweaking.

The generator has the following known shortcomings which I am not planning to address:
- It is impossible to generate CMake files for the tests and devtools
- create_project must be built on the same platform as the target of the CMake build (no cross compilation support)
- CMake's configuration features are only used for part of the libraries. Other libraries are hardcoded, and are expected to be found at the same location as configured ones.
- ASM scalers and updates are not supported.

This code is based on original work by @somaen and @klusark.

### DIFF
--- a/devtools/create_project/cmake.cpp
+++ b/devtools/create_project/cmake.cpp
@@ -1,0 +1,323 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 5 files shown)


sev-:
Very well, thank you, merging.

PR #785, opened 2016-07-17
SCUMM: Maniac V0: Walkbox / Timing improvements

segrax:
Hey All,

There is currently a number of small issues with MM V0. This patch addresses them, 

1. In the 'V0 Demo' room 5 during the 'hands off demo mode' Sandy runs down, then left to escape Purple Tentacle, instead of going around the stair case (this is due to the original engine finding the first path, not necessarily the fastest. This behavior is now implemented)

1. During the intro, the timing is off between the audio/video. The music builds up, but the "A LucasFilm Production" text has already been displayed.

1. Direction of actors after loading a save game is always putting right

1. Top text area is always cleared prior to screen fade in V0

Notes:

1. Original walkbox algorithm is implemented, we now calculate the walk boxes before walking, then remove them from the array as we walk to them.

2. Counters have been added, and slight delays are added for each object/strip/actor/limb being refreshed, as well as screen scrolls.
The reason this is done, is because the original engine slows down to < 60Hz when certain routines are called (especially screen refresh related). some scripts make use of the 'breakHere' opcode instead of 'Delay', as the script author had the expectation of a previously given instruction taking more than 1 tick worth of time to complete (ie. Maniac Mansion V1-V2 both use Delay(30) to pause between the Mansion lights turning on during the introduction, whereas V0 uses breakHere -- causing the lights to turn on almost instantly in ScummVM, a C64 on the other hand has an expected delay, as the screen is redrawn based on the object change of state  )
The values used are based on visual observation, and are not 100% correct. It is close to the original though (and this removes the workaround for the Dungeon Door).
Comparison of ScummVM vs WinVice here [ScummVM vs WinVice](https://youtu.be/XqHx16AzmKU)
As you play in ScummVM now, it now gives you the sense of playing on a C64 :)

3. Loading/Saving of V0 has been fixed properly, and now supports saving the walkbox queue

4. Fixed

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -272,23 +272,23 @@ void Actor_v0::walkBoxQueueReset() {
 	_walkboxHistory.clear();
 	_walkboxQueueIndex = 0;
 
-	for (uint i = 0; i < ARRAYSIZE( _walkboxQueue ); ++i) {
+	for (uint i = 0; i < ARRAYSIZE(_walkboxQueue); ++i) {
 		_walkboxQueue[i] = kInvalidBox;
 	}
 }
 
 bool Actor_v0::walkBoxQueueAdd(int box) {
 
-	if (_walkboxQueueIndex == ARRAYSIZE( _walkboxQueue ))
+	if (_walkboxQueueIndex == ARRAYSIZE(_walkboxQueue))
 		return false;
 
-	_walkboxQueue[_walkboxQueueIndex++] = box ;
-	_walkboxHistory.push_back( box );
+	_walkboxQueue[_walkboxQueueIndex++] = box;
+	_walkboxHistory.push_back(box);
 	return true;
 }
 
 void Actor_v0::walkboxQueueReverse() {
-	int j = ARRAYSIZE( _walkboxQueue ) - 1;
+	int j = ARRAYSIZE(_walkboxQueue) - 1;
 
 	while (_walkboxQueue[j] == kInvalidBox && j >= 1)
 		--j;
@@ -328,21 +328,21 @@ bool Actor_v0::walkBoxQueuePrepare() {
 	// Build a series of walkboxes from our current position, to the target
 	do {
 		// Add the current box to the queue
-		if (!walkBoxQueueAdd( BoxFound ))
+		if (!walkBoxQueueAdd(BoxFound))
 			return false;
 
 		// Loop until we find a walkbox which hasn't been tested
 		while (_walkboxQueueIndex > 0) {
 
 			// Check if the dest box is a direct neighbour
-			if ((BoxFound = _vm->getNextBox( BoxFound, _walkdata.destbox )) == kInvalidBox) {
+			if ((BoxFound = _vm->getNextBox(BoxFound, _walkdata.destbox)) == kInvalidBox) {
 
 				// Its not, start hunting through this boxes immediate connections
-				byte* boxm = _vm->getBoxConnectionBase( _walkboxQueue[_walkboxQueueIndex - 1] );
+				byte* boxm = _vm->getBoxConnectionBase(_walkboxQueue[_walkboxQueueIndex - 1]);
 
 				// Attempt to find one, which we havn't already used
 				for (; *boxm != kInvalidBox; ++boxm) {
-					if (walkBoxQueueFind( *boxm ) != true)
+					if (walkBoxQueueFind(*boxm) != true)
 						break;
 				}
 
@@ -357,7 +357,7 @@ bool Actor_v0::walkBoxQueuePrepare() {
 
 					_newWalkBoxEntered = true;
 
-					walkBoxQueueAdd( BoxFound );
+					walkBoxQueueAdd(BoxFound);
 
 					walkboxQueueReverse();
 					return true;
@@ -750,7 +750,7 @@ void Actor::startWalkActor(int destX, int destY, int dir) {
 	_walkdata.curbox = _walkbox;
 
 	if (_vm->_game.version == 0) {
-		((Actor_v0*)this)->walkBoxQueuePrepare();
+		((Actor_v0 *)this)->walkBoxQueuePrepare();
 
 	} else if (_vm->_game.version <= 2) {
 		_moving = (_moving & ~(MF_LAST_LEG | MF_IN_LEG)) | MF_NEW_LEG;
@@ -1507,9 +1507,9 @@ void Actor::putActor(int dstX, int dstY, int newRoom) {
 
 	if (_vm->_game.version == 0) {
 
-		((Actor_v0*)this)->_newWalkBoxEntered = false;
-		((Actor_v0*)this)->_CurrentWalkTo = _pos;
-		((Actor_v0*)this)->_NewWalkTo = _pos;
+		((Actor_v0 *)this)->_newWalkBoxEntered = false;
+		((Actor_v0 *)this)->_CurrentWalkTo = _pos;
+		((Actor_v0 *)this)->_NewWalkTo = _pos;
 
 		// V0 always sets the actor to face the camera upon entering a room
 		setDirection(oldDirToNewDir(2));
@@ -2512,7 +2512,7 @@ void Actor_v0::limbFrameCheck(int limb) {
 void Actor_v0::animateCostume() {
 	speakCheck();
 
-	byte count = _vm->_costumeLoader->increaseAnims( this );
+	byte count = _vm->_costumeLoader->increaseAnims(this);
 
 	if (count) {
 		_vm->_V0Delay._actorLimbRedrawDrawCount += count;
@@ -3430,7 +3430,7 @@ void Actor_v0::actorSetWalkTo() {
 
 	_newWalkBoxEntered = false;
 
-	int nextBox = ((ScummEngine_v0*)_vm)->walkboxFindTarget(this, _walkdata.destbox, _walkdata.dest);
+	int nextBox = ((ScummEngine_v0 *)_vm)->walkboxFindTarget(this, _walkdata.destbox, _walkdata.dest);
 	if (nextBox != kInvalidBox) {
 		_walkdata.curbox = nextBox;
 	}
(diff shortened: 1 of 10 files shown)


engines/scumm/script_v5.cpp
```diff
@@ -2490,7 +2490,7 @@ void ScummEngine_v5::walkActorToActor(int actor, int toActor, int dist) {
 	if (_game.version <= 2) {
 		x /= V12_X_MULTIPLIER;
 		y /= V12_Y_MULTIPLIER;
-	}
+	}
```
sev-: This whole file needs to be reverted, as it contains only the whitespace changes.

engines/scumm/scumm.cpp
```diff
@@ -2079,13 +2108,24 @@ Common::Error ScummEngine::go() {
...
-		// WORKAROUND: walking speed in the original v0/v1 interpreter
+		// WORKAROUND: Unfortunately the MOS 6502 wasn't always fast enough for MM
+		//  a number of situations can lead to the engine running at less than 60 ticks per second, without this drop
+		//	- A single kid is able to escape via the Dungeon Door (after pushing the brick)
+		//	- During the intro, calls to 'SetState08' are made for the lights on the mansion, with a 'breakHere'
+		//	  in between each, the reduction in ticks then occurs while affected stripes are being redrawn.
+		//	  The music buildup is then out of sync with the text "A Lucasfilm Games Production".
+		//	  Each call to 'breakHere' has been replaced with calls to 'Delay' in the V1/V2 versions of the game
+		if (_game.version == 0) {
+			delta += ((ScummEngine_v0*) this)->DelayCalculateDelta();
```
sev-: it must be '(ScummEngine_v0 *)this' according to our code formatting conventions.

engines/scumm/scumm.cpp
```diff
@@ -2079,13 +2108,24 @@ Common::Error ScummEngine::go() {
...
+		}
+		
+		// WORKAROUND: walking speed in the original v1 interpreter
 		// is sometimes slower (e.g. during scrolling) than in ScummVM.
 		// This is important for the door-closing action in the dungeon,
 		// otherwise (delta < 6) a single kid is able to escape.
-		if ((_game.version == 0 && isScriptRunning(132)) ||
-			(_game.version == 1 && isScriptRunning(137)))
-			delta = 6;
+		if (_game.version == 1 && isScriptRunning( 137 )) {
```
sev-: Please remove spaces around '137'

engines/scumm/scumm.h
```diff
@@ -1131,6 +1140,8 @@ class ScummEngine : public Engine {
 
 	byte getNumBoxes();
 	byte *getBoxMatrixBaseAddr();
+	byte *getBoxConnectionBase( int box );
```
sev-: Same here, please remove spaces around the parameter.

engines/scumm/scumm.cpp
sev-: Also, why script 132 does not trigger the adjustment anymore. Was it fixed by the new code?

engines/scumm/actor.cpp
```diff
@@ -3350,10 +3469,33 @@ void Actor_v0::saveLoadWithSerializer(Serializer *ser) {
...
+
+		MKARRAY(Actor_v0, _walkboxQueue[0], sleByte, 16, VER(98)),
+		MKLINE(Actor_v0, _walkboxQueueIndex, sleByte, VER(98)),
 		MKEND()
 	};
 
 	ser->saveLoadEntries(this, actorEntries);
+
+	if (ser->isLoading()) {
+		if (_costCommand != 0xFF ) {
```
sev-: This block needs to be reformatted.

engines/scumm/actor.cpp
```diff
@@ -2404,8 +2518,13 @@ void Actor_v0::limbFrameCheck(int limb) {
 void Actor_v0::animateCostume() {
 	speakCheck();
 
-	if (_vm->_costumeLoader->increaseAnims(this))
+	byte count = _vm->_costumeLoader->increaseAnims( this );
```
sev-: '( this )' -> '(this)'

engines/scumm/actor.cpp
```diff
@@ -1398,13 +1511,13 @@ void Actor::putActor(int dstX, int dstY, int newRoom) {
...
 	}
 
-	// V0 always sets the actor to face the camera upon entering a room
 	if (_vm->_game.version == 0) {
-		_walkdata.dest = _pos;
 
-		((Actor_v0*)this)->_newWalkBoxEntered = true;
+		((Actor_v0*)this)->_newWalkBoxEntered = false;
 		((Actor_v0*)this)->_CurrentWalkTo = _pos;
+		((Actor_v0*)this)->_NewWalkTo = _pos;
```
sev-: '(Actor_v0 *)' -- apparently it slipped it unnoticed in your previous patch as well.

sev-:
Very good pull request. Thank you for continuing with fixing these.

However, the code formatting needs to be adjusted slightly to conform our Code Formatting Conventions. Once it is done, the PR could be merged.

engines/scumm/scumm.cpp
segrax: Yeah, DelayCalculateDelta handles the adjustments

segrax:
Thanks Sev,

Will take care of them tonight when I get home

engines/scumm/boxes.cpp
```diff
@@ -1195,7 +1195,7 @@ byte ScummEngine_v0::walkboxFindTarget(Actor *a, int destbox, Common::Point walk
 	// Next box reached
 	if (nextBox != Actor::kInvalidBox && nextBox != a->_walkbox) {
 
-		getClosestPtOnBox(getBoxCoordinates(nextBox), a->getPos().x, a->getPos().y, Actor->_NewWalkTo.x, Actor->_NewWalkTo.y);
+		getClosestPtOnBox(getBoxCoordinates(nextBox), a->getRealPos().x, a->getRealPos().y, Actor->_NewWalkTo.x, Actor->_NewWalkTo.y);
```
segrax: I've also made this change, as the values being returned was not on the same scale as the Box Coordinates (its surprising this hasn't caused an issue before)

segrax:
@sev- All formatting issues should be now fixed, and one last bug fix was applied.

engines/scumm/actor.cpp
```diff
@@ -259,13 +259,125 @@ void Actor_v0::initActor(int mode) {
...
 	} else {
 		_sound[0] = v0ActorTalk[_number];
 	}
 }
 
+void Actor_v0::walkBoxQueueReset() {
+	_walkboxHistory.clear();
+	_walkboxQueueIndex = 0;
+
+	for (uint i = 0; i < ARRAYSIZE( _walkboxQueue ); ++i) {
```
sev-: Nope, we are not putting spaces after and before the parens. Should be '(_walkboxQueue)', and further, and with '( box )' laterm etc.

engines/scumm/actor.cpp
```diff
@@ -259,13 +259,125 @@ void Actor_v0::initActor(int mode) {
...
+	do {
+		// Add the current box to the queue
+		if (!walkBoxQueueAdd( BoxFound ))
+			return false;
+
+		// Loop until we find a walkbox which hasn't been tested
+		while (_walkboxQueueIndex > 0) {
+
+			// Check if the dest box is a direct neighbour
+			if ((BoxFound = _vm->getNextBox( BoxFound, _walkdata.destbox )) == kInvalidBox) {
```
sev-: same here and below.

engines/scumm/actor.cpp
```diff
@@ -3350,10 +3463,38 @@ void Actor_v0::saveLoadWithSerializer(Serializer *ser) {
...
+				setDirection( _facing );
+				speakCheck();
+
+			} else {
+				// Force limb direction update
+				_facing = 0;
+				directionUpdate();
+
+				// Begin walking
+				animateActor( newDirToOldDir( _facing ) );
```
sev-: and here

engines/scumm/actor.cpp
```diff
@@ -3496,6 +3637,8 @@ void Actor::saveLoadWithSerializer(Serializer *ser) {
 			_walkdata.point3.x >>= V12_X_SHIFT;
 			_walkdata.point3.y >>= V12_Y_SHIFT;
 		}
+
+		setDirection( _facing );
```
sev-: Here also

engines/scumm/actor.h
```diff
@@ -380,6 +385,12 @@ class Actor_v0 : public Actor_v2 {
 
 	bool _limb_flipped[8];
 
+private:
+
+	bool walkBoxQueueAdd( int box );
```
sev-: Ditto

engines/scumm/boxes.cpp
```diff
@@ -692,6 +692,22 @@ byte *ScummEngine::getBoxMatrixBaseAddr() {
...
 }
 
+byte *ScummEngine::getBoxConnectionBase( int box ) {
+	byte *boxm = getBoxMatrixBaseAddr();
+
+	int boxIndex = 0;
+
+	for (; boxIndex != box; ++boxIndex) {
+
+		while ( *boxm != 0xFF) {
```
sev-: Once more, and above, with the function declaration.

engines/scumm/gfx.cpp
```diff
@@ -3795,6 +3795,11 @@ void ScummEngine::fadeOut(int effect) {
 		_textSurface.fillRect(Common::Rect(0, vs->topline * _textSurfaceMultiplier, _textSurface.pitch, (vs->topline + vs->h) * _textSurfaceMultiplier), 0);
 #endif
 
+	// V0 wipes the text area before fading out
+	if (_game.version == 0) {
+		updateDirtyScreen( kTextVirtScreen );
```
sev-: Here too.

segrax:
@sev- :( so it turns out my ide was auto formatting, fixed, and double checked :) if i've missed or caused anymore, its going to be really embarrassing

sev-:
Great! Merging

rootfather:
@segrax:

I'm currently updating our NEWS file. Do you think the following entries are appropriate to your PR?
* Improve pathfinding in Maniac Mansion V0
* Game speed is now closer to the original in certain situations

segrax:
@rootfather yeah, sounds good

rootfather:
@segrax Thanks for your reply.

One more question: What is the "V0" referring to? I only find references to a Version 1 and Version 2 on various websites.

segrax:
@rootfather V0 is the Commodore 64 and Apple 2 version, the graphics/sound are very similar between V0 and V1 versions, but the engine itself was completely rewritten in C for V1 (whereas V0 was written in 6502 assembly)

PR #786, opened 2016-07-19
TITANIC: Starship Titanic engine

dreammaster:
This is a pull request for my in-progress work on the Starship Titanic engine/game. The game isn't anywhere near feature complete yet, but it's certainly getting there. There's still parts of the NPC script system to figure out, and lots of classes needed by the Star Control display (which uses a lot of messy 3D projection code to draw points in a starmap). So you'll notice that there's a lot of classes in titanic/star_control/ that have names like "CStarControlSub6" with methods like "proc3" or "fn2". These will gradually get renamed as the star control class hierarchy is more fully implemented and understood. Also missing is the hard-coded game logic in all the various game classes, although the core base class hierarchy is pretty much feature complete.

Another thing to mention is that the engine makes the following changes to the core AVIDecoder:
* Titanic is added to the comment listing the engines that use it
* The root RIFF chunk's filesize is now respected when decoding the chunks of the AVI file. Previously, the decoder was simply processing chunks until false was returned, but in Titanic, some of the videos have a few bytes of padding garbage at the end of the file. Hopefully the use of the filesize wasn't disabled deliberately for some other engine having bad AVI files?
* STRN chunks are now processed and store the stream name for video tracks in their stream header structures. This is needed by the engine, since some videos use special stream names to indicate what the transparency pixel is
* Finally, I added a new optional callback function parameter to the AVIDecoder constructor. When provided (for Titanic), it will be passed the details of any video & audio track about to be added, and returns whether it should be added. This is needed because some Titanic videos have a second video track, and this allows Titanic to create two AVIDecoder instances, one with each video track.

Finally, the engine also needs it's own .dat file. I've created a 'create_titanic' tool which creates the data file based on the current GOG release (maybe others as well, but I haven't tested it on the original release executable).  I previously had multiple commits of the generated datafile in dists/engine-data, but it's changed so frequently as I added new data, I removed all previous revisions of the file from the branch's history. I'll commit a final version in at some point in the future once I'm sure no further data needs to be added.


dreammaster:
Hmmm.. there seems to be some weird conflict issues on a lot of stuff that I never directly modified. I fear my prior interactive rebases on the history, such as to remove old copies of the DAT file, have screwed something up. Hopefully I can get some assistance from someone prior to the merge to more properly fix the branch so it can be cleanly merged.

wjp:
You rebased about 500 "other" commits from master into this branch, which is causing these conflicts. 

I doubt an interactive rebase would've done this; more likely you rebased master on top of titanic instead of merging it, since the extra commits are exactly the commits since the previous normal merge of master with titanic at that time.

I can clean this up later.

dreammaster:
Ah, that's the cause. Thanks.

PaulGilbert:
A while back Strangerke had problems getting anything to display; it turned out that his installation didn't have the JPEG library included, and the engine requires it. I can't remember.. is there a standardized way where we specify an engine needs a particular library to run? And/or detect if it's not present at startup without going through trying to actually decode one of the game's JPEGs?

wjp:
In theory it shouldn't even build the engine since you specify `jpeg` as a dependency in `configure.engine`.

PaulGilbert:
Ah; seems like Strangerke added that in himself after he first had the problem. That takes care of it, then.

sev-:
I will hold off my review until the history is fixed. Currently it lists way too many non-related changes.

bluegr:
Nice work, but as @sev said, it's not easy to review this until its history is fixed :(

wjp:
I suppose I should go do that then.

wjp:
I've force-pushed the fixed up history. (I took the liberty of also replacing a few commit message prefixes like TTIANIC, TTITANIC by TITANIC.)

Github still seems to be a bit confused about the exact state of the pull request, but hopefully that clears itself up.

wjp:
Github seems to be un-confused again, but this branch is still causing merge conflicts because it touches a file in the voyeur engine.

PaulGilbert:
Many thanks. I double checked, and that change to Voyeur was an accidental copy of something I already did in master, rearranging fields to stop Visual Studio giving warnings about structure field alignment. Do you feel up to doing one more rebase to remove the file from the commit? It would be cleaner to do it now pre-merge, and at this point I feel more confident about your rebasing skills than mine :)

wjp:
Was just doing that :-)

Done now, and force-pushed again.

dreammaster:
Bonza :)

sev-:
As usual, very high-quality code.

Couple notes so far:

1. I am a bit worried by copying substantial chunks of the common/ code to the create_titanic utility. Maybe it makes sense to use STL instead? It is allowed and is in fact used by some of the devtools as we do not compile them on all platforms

2. I would recommend having the macros ON_MESSAGE(), DECLARE_MESSAGE_MAP, CLASSDEF, DEFFN() etc requiring a semicolon at the end, just as we do with the coroutines stuff. It will make the code look more c-like :)

3. Why do you use 'C' prefix for most of the class names? Just being curious

4. Some guard defines are missing the correct directory structure. E.g. TITANIC_PLUG_IN_H for titanic/carry/plug_in.h file, TITANIC_FILE_ITEM_H for titanic/core/file_item.h

5. Great list of includes in engines/titanic/core/saveable_object.cpp

6. The engine is 16bpp, however, you do not enlist it as a requirement in the configure.engine

dreammaster:
Thanks Sev.

1. You're right; I did have to copy some code from core, since including the original added unwanted dependencies to g_system. However, I don't think it's an egregious amount. The biggest amount comes from the EXE resource reading code, and I'm not aware of STL having convenient classes for reading from a Windows EXE. Since everything is already working as is, and it's not part of the main codebase, I really don't see any advantage to any significant refactoring of it.

2. Okay, I added semicolons to some of them. It's not possible for all of them, though.. in particular, the BEGIN_MESSAGE_MAP and ON_MESSAGE macros can have semicolons after them, because their internal definition has BEGIN ending with the starting definition for an array, and each ON_MESSAGE adding a row to the messages array. So I can't have semicolons present without making the array definition invalid.

3. Mostly  because that's the way the original class hierarchy named the classes, and I decided to follow their example. It used 'C' prefixes for most standard classes, as well as "OS" as a prefix for operating specific classes like OSVideoSurface, and "TT" as a prefix for True Talk (conversation) related classes.

4. Actually, off the top of my head, I don't think any of the header defines actually include the folder. So for example, the root tree item class in core/tree_item.h uses TITANIC_TREE_ITEM_H rather than TITANIC_CORE_TREE_ITEM_H. Given that there aren't any duplicate filenames, I'd prefer not to have to rename the defines across hundreds of header files :)

5. Yes indeed. Unfortunately, I'm not aware of any way around it. The file needs to have access to each class's ClassDef definition so that when loading a savegame, an instance of any type of valid object can be created. I experimented with ways not to have to include the includes, but even using partial class definitions automatically defined by the macros, the compiler errors out that the classes aren't fully defined.

6. Oh right, Added in.

sev-:
As you may see from [the Travis CI output](https://travis-ci.org/scummvm/scummvm/jobs/148200931), the create_titanic tool is not compilable.

sev-:
Time to merge

PR #787, opened 2016-07-21
DIRECTOR: New engine

sev-:
This is the WIP of the Director engine, which was part of GSoC.

As you know, iskrich quit the GSoC, so this is now currently on me and (probably) fuzzie.

I was working on the Lingo implementation, and still continue working on it. After I have the language completed, I plan to turn back to the Director itself.

Current status:
Director: about 60%
Lingo: about 60% of the language features, about 20% of the built-in functions

Things missing in Director:
- Autostart (currently there is one movie hardcoded)
- Input
- Hooks back to Lingo (about 40% done)
- Sound control

Things missing in Lingo language features:
- Lists
- Case statement
- Factories
- (probably) Some combinations of if() statements, need to see them in the wild

There is also a 'directortest' pseudo-target, which runs all lingo tests, which you can find in lingo/tests directory.

Lingo is implemented as a p-code compiler, with barebones taken from hoc. I base it on flex/bison, but neither is required for compilation, since I committed their output as well. For the development they are needed. Also I base on bison2, not bison3.


engines/director/lingo/lingo-builtins.cpp
```diff
@@ -0,0 +1,219 @@
...
+	Datum d = g_lingo->pop();
+	d.toFloat();
+	d.u.f = log(d.u.f);
+	g_lingo->push(d);
+}
+
+void Lingo::b_pi() {
+	Datum d;
+	d.toFloat();
+	d.u.f = 3.14159265358979;
```
bluegr: There's M_PI defined in math.h for this

engines/director/lingo/lingo-code.cpp
```diff
@@ -0,0 +1,743 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+// Heavily inspired by hoc
```
bluegr: hoc? You mean the programming language? How has this affected Lingo?

bluegr:
Overall, nice and well written!

The vast majority of this is Lingo...

engines/director/lingo/lingo-builtins.cpp
sev-: Thanks, fixed

engines/director/lingo/lingo-code.cpp
sev-: I am using same stack machine as hoc. I started from the barebones of hoc, and then modified and extended it for Lingo.

Lingo itself is Forth-based.

sev-:
> The vast majority of this is Lingo...

Not really.

Lingo is 2570 lines of code
Director is 3117 lines of code

However, since I commit both flex and bison output, it looks like there is more Lingo than Director.

Also iskrich wrote about 300 lines of code in Lingo too.

I forgot to mention about flex/bison, so I'll modify the PR description now.

bluegr:
Just for completeness sake... the current TODO is still valid, right?

sev-:
No, the numbers could be bumped. Moreover, I found out that lists belong to director 5+, thus are out of current scope. Same goes for the switch statement.

Also iskrich made couple of additions to 'the' entities in the meantime.

PR #788, opened 2016-07-21
ALL: Add Cloud storage support

Tkachov:
This adds the Cloud storages support feature. Four providers supported: Dropbox, OneDrive, Google Drive and Box. These are controlled by CloudManager, which could be accessed as CloudMan singleton. CloudMan saves the information in scummvm.ini within "cloud" section.

Storages use OAuth2 and interact with providers' API by giving the access_token. OneDrive, Google Drive and Box has to refresh token from time to time. To send the request, libcurl is used. There is a small system of Requests built on top of it. Requests are managed by ConnectionManager, which works in a separate thread (implemented with Timer) and keeps them running. If there are no more Requests, ConnectionManager stops the timer. While Storage is working (i.e. Requests are running), a pulsating Cloud icon is shown in the corner of the screen.

Cloud icon

Cloud-related information is available at the new Cloud tab of the Options dialog. One can switch connected storages there, refresh the information about these or connect a new one. The main features of the Cloud support are automatic saves sync (starts on each manual save, autosave, startup and on SaveLoadDialog opening) and games download (could be used from the Cloud tab). Before downloading ScummVM checks whether connection is "limited" and notifies user about that. Networking::Connection::isLimited() has only Android backend yet, where it checks whether Wi-Fi or a mobile network is used.

Options dialog Cloud tab

The Wi-Fi sharing feature is also added. ScummVM now can start a local webserver (using SDL_Net), which provides the Files Manager: users can navigate through directories, download files, create new directories or upload files using the browser on another device. Information about the server is available on the same Cloud tab. Users could change the webserver's port there.

AJAX Files Manager
LocalWebserver's port in Cloud tab

To connect a Storage, users must navigate to the link we show on the screen, allow our app access to their account, and receive the code which would be used by ScummVM to get the access_token. Some features were added in order to simplify that. First of all, we show a special short scummvm.org link. Secondly, there is a special "Open URL" button, which opens that link in a default browser, if the corresponding Networking::Browser::openUrl() backend is implemented.

Storage Connection Dialog: "Open URL" button, 8 fields

The code is quite long, so scummvm.org page transforms it into a few short groups with hashsum added. ScummVM dialog shows 8 fields, where user should type these groups, and checks that hashsum is OK. Then I also added SDL2 clipboard support in EditableWidget, so now users could paste these short groups into these fields.

scummvm.org code page

Finally, if there is LocalWebserver available, a bit different link is shown and cloud providers redirect user to the localhost:12345 page instead of scummvm.org. This way ScummVM can get the code without user typing it in the fields, which makes connecting a Storage extremely simple.

Storage Connection Dialog: no fields

----

It lacks iOS implementations of Networking::Browser::openUrl() and Networking::Connection::isLimited().

There are some commits from @sev- 's [gui-improvement](https://github.com/sev-/scummvm/commits/gui-improvement). This code was in my container-box branch while I was working on it, so there are clipping versions for transparent surfaces from gui-improvement. Yet, I had to remove it while preparing PR with Container box, so when I merged that back into my cloud branch, I didn't add those clipping versions back. In short, there are commits for transparent pictures, but those don't get clipped.

There is no code overriding loading Storage's KEY and SECRET from scummvm.ini.

All the short links are redirecting to giving access to my test applications. Actual ScummVM applications must be created with their real keys and secrets, and links must be edited to redirect to these.

Some TODOs are still in the code, I'm working on these.


engines/sci/detection.cpp
```diff
@@ -784,6 +785,8 @@ SaveStateList SciMetaEngine::listSaves(const char *target) const {
 	return saveList;
 }
 
+bool SciMetaEngine::simpleSaveNames() const { return true; }
```
wjp: SCI _usually_ uses standard savegame names, but not for all games, and some games additionally store other files using the savegame system (containing settings and such).

engines/metaengine.h
```diff
@@ -116,6 +116,20 @@ class MetaEngine : public PluginObject {
...
+	* number.
+	*
+	* If that's not true or engine is using some unusual way
+	* of detecting saves and slot numbers, this should return
+	* false. In that case Save/Load dialog would be unavailable
+	* during cloud saves sync.
+	*
+	* @return	true, if "<target>.###" is OK for this engine
+	*/
+	virtual bool simpleSaveNames() const { return false; }
```
wjp: Wouldn't the `simpleSaveNames` thing fit better as an engine feature? (See `MetaEngineFeature`)

Tkachov: Well, it looks like it really could be implemented as an engine feature. Didn't see that back then. Noted.

backends/networking/sdl_net/localwebserver.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+	// Create a listening TCP socket
+	IPaddress ip;
+	if (SDLNet_ResolveHost(&ip, NULL, _serverPort) == -1) {
+		error("SDLNet_ResolveHost: %s\n", SDLNet_GetError());
+	}
+
+	resolveAddress(&ip);
+
+	_serverSocket = SDLNet_TCP_Open(&ip);
```
wjp: I'm not too familiar with SDLNet specifically, but why not only listen on 127.0.0.1?

engines/sci/detection.cpp
wjp: To elaborate, the SCI engine could give a list of all savegames (and other saved files) for the current game, but it won't be limited to the usual `target.###`.

backends/networking/sdl_net/localwebserver.cpp
Tkachov: Well, passing NULL up there means we want to use 0.0.0.0 ("listen on every available network interface").

Address resolving is needed because local webserver (in case of Wi-Fi sharing feature) is meant to be accessed from another device on the same network, thus actual local IP is required instead of plain 127.0.0.1.

wjp: What is a wifi sharing feature, why do we need it, and where I can find information about your use of it?

wjp:
Given the big impact of this, is there any technical design document? (In particular about the various security considerations.)

backends/networking/sdl_net/localwebserver.cpp
sev-: This is a local webserver running on top of ScummVM, so a user can open the provided URL in their local network and have direct access to the device filesystem. Then they can upload the game files or download saves from their device running ScummVM.

This is an alternative to the external cloud usage.

Tkachov: Some information about it is in the [PR description](https://github.com/scummvm/scummvm/pull/788) and in [my blog](http://tkachov.ru/gsoc/post_07_10_2016_game_download_and_wifi_sharing).

We need it because it's cool and some other apps - for example, VLC - have such feature.

Almost all "handlers" in the `backends/networking/sdl_net/` are implementing this feature.

In short, it allows one to access files via network - upload and download files from the device ScummVM is working on. For example, one can upload a game from a PC onto a smartphone without USB cable.

wjp: Hm, I see how that could be nice. Is the same local webserver used for this feature and the OAuth stuff?

sev-:
Yes, it was highlighted in the Alex's blog. For instance, see here: http://tkachov.ru/gsoc/post_05_28_2016_file_downloading_week

backends/networking/sdl_net/localwebserver.cpp
Tkachov: Yes, it is.

backends/base-backend.cpp
```diff
@@ -39,6 +39,20 @@ void BaseBackend::displayMessageOnOSD(const char *msg) {
 	dialog.runModal();
 }
 
+void BaseBackend::copyRectToOSD(const void *buf, int pitch, int x, int y, int w, int h) {
+	warning("BaseBackend::copyRectToOSD not implemented"); //TODO
+}
+
+void BaseBackend::clearOSD() {
+	warning("BaseBackend::clearOSD not implemented"); //TODO
+	//what should I do? Remove all TimedMessageDialogs?
```
sev-: Yes, skip those. Additionally, we could add another backend feature: kFeatureOSD.

backends/cloud/box/boxlistdirectorybyidrequest.cpp
```diff
@@ -0,0 +1,169 @@
...
+		    warning("Box returned error: %s", responseObject.getVal("error_summary")->asString().c_str());
+		    error.failed = true;
+		    error.response = json->stringify();
+		    finishError(error);
+		    delete json;
+		    return;
+		}
+		*/
+
+		//TODO: check that ALL keys exist AND HAVE RIGHT TYPE to avoid segfaults
```
sev-: Right, it is a must for implementing. Could also help with testing if the providers change their API.

backends/cloud/box/boxlistdirectorybyidrequest.cpp
```diff
@@ -0,0 +1,169 @@
...
+				Common::String name = item.getVal("name")->asString();
+				bool isDirectory = (item.getVal("type")->asString() == "folder");
+				uint32 size = 0, timestamp = 0;
+				if (item.contains("size")) {
+					if (item.getVal("size")->isString())
+						size = item.getVal("size")->asString().asUint64();
+					else if (item.getVal("size")->isIntegerNumber())
+						size = item.getVal("size")->asIntegerNumber();
+					else
+						warning("strange type for field 'size'");
```
sev-: It is always helpful to print out the unexpected value when possible.

backends/cloud/box/boxlistdirectorybyidrequest.cpp
```diff
@@ -0,0 +1,169 @@
...
+		if (responseObject.contains("offset") && responseObject.getVal("offset")->isIntegerNumber())
+			received = responseObject.getVal("offset")->asIntegerNumber();
+		if (responseObject.contains("limit") && responseObject.getVal("limit")->isIntegerNumber())
+			received += responseObject.getVal("limit")->asIntegerNumber();
+		bool hasMore = (received < totalCount);
+
+		if (hasMore) makeRequest(received);
+		else finishListing(_files);
+	} else {
+		warning("null, not json");
```
sev-: Please put prefix, like: "BoxListDirectoryByIdRequest::responseCallback(): null, not json". Will help greatly with the troubleshooting.

Tkachov: Sure.

bluegr:
From a quick glance: quite impressive work, overall! Well done! :)

I'll check this out later on

backends/cloud/box/boxlistdirectorybyidrequest.cpp
Tkachov: Yep, that's what I'm working on.

wjp:
I mean a design document for the whole general cloud functionality. This feature will require long-term maintainability more than most of our current features, and also potentially urgent security fixes 
sometimes. I don't think blog posts or a pull request description (while nice) are really sufficient for this.

backends/networking/sdl_net/localwebserver.cpp
wjp: That has the unfortunate side effect of opening a large attack surface unnecessarily when not using this feature.

backends/cloud/box/boxstorage.cpp
```diff
@@ -0,0 +1,299 @@
...
+		request->addPostField("grant_type=authorization_code");
+		request->addPostField("code=" + code);
+	} else {
+		request->addPostField("grant_type=refresh_token");
+		request->addPostField("refresh_token=" + _refreshToken);
+	}
+	request->addPostField("client_id=" + Common::String(KEY));
+	request->addPostField("client_secret=" + Common::String(SECRET));
+	/*
+	if (Cloud::CloudManager::couldUseLocalServer()) {
```
sev-: Is this working?

backends/cloud/box/boxstorage.cpp
```diff
@@ -0,0 +1,299 @@
...
+	Common::JSONValue *json = response.value;
+	if (!json) {
+		warning("BoxStorage: got NULL instead of JSON");
+		if (callback) (*callback)(BoolResponse(nullptr, false));
+		return;
+	}
+
+	Common::JSONObject result = json->asObject();
+	if (!result.contains("access_token") || !result.contains("refresh_token")) {
+		warning("Bad response, no token passed");
```
sev-: Same thing, this need to have a prefix, so you could find it in the source code with ease.

backends/cloud/box/boxstorage.cpp
```diff
@@ -0,0 +1,299 @@
...
+		debug("%s", json->stringify().c_str());
```
sev-: Please hide this to some deep level, like 7 or deeper.

backends/networking/sdl_net/localwebserver.cpp
Tkachov: Webserver has to be turned on manually with a special button in the Cloud tab.

It's also turned on automatically while connecting a Storage, but it's turned off when Storage is connected.

As ScummVM isn't working on routers, I doubt someone could access files with that feature through the Internet. If the user is smart enough to configure port forwarding, I guess he/she is smart enough to use that with caution.

If the malefactor is on the same local network... well, user shouldn't have run the server while using a non-trusted network, then.

backends/cloud/box/boxstorage.cpp
```diff
@@ -0,0 +1,299 @@
...
+		delete outerCallback;
+	}
+
+	delete json;
+}
+
+Networking::Request *BoxStorage::createDirectoryWithParentId(Common::String parentId, Common::String name, BoolCallback callback, Networking::ErrorCallback errorCallback) {
+	if (!errorCallback) errorCallback = getErrorPrintingCallback();
+
+	Common::String url = "https://api.box.com/2.0/folders";
```
sev-: I recommend to move these to the very top of the file as constants with #define.

backends/networking/sdl_net/localwebserver.cpp
wjp: Is that really your attitude towards security?

sev-: We need to ensure that there is no way to go beyond a sandboxed directory where all game-related files could be stored. Also, ScummVM is not supposed to execute anything directly, all code in the engines so far is using VMs.

My recommendation is to enforce the server close once the dialog with upload files is closed. In this case this will be yet another safety feature.

However, we are using the local server for accepting keys from the local browser (e.g. we provide 127.0.0.1:12345 as a callback URL). In this case I would enforce additional checks on the incoming URL formats. We are using it only for extracting the keys provided in the GET URLs

Tkachov:
Sure, I'd compile the detailed information on a wiki page.

backends/networking/sdl_net/localwebserver.cpp
Tkachov: As I mentioned, server is turned off automatically after it's used in the Storage connection. As Eugene said, we can additionally turn off the server when user closes the Options dialog, so it would be available only while user actually works with it.

backends/cloud/box/boxuploadrequest.cpp
```diff
@@ -0,0 +1,212 @@
...
+			long code = stream->httpResponseCode();
+			error.httpResponseCode = code;
+		}
+	}
+
+	if (error.httpResponseCode != 200 && error.httpResponseCode != 201)
+		warning("looks like an error");
+
+	Common::JSONValue *json = response.value;
+	if (json) {
```
sev-: For simplicity I would negate it. E.g.:
```
if (!json) {
   warning("BoxUploadRequest::uploadedCallback(): null, not json");
   finishError(error);
   return;
}

if (!json->isObject()) {
   warning...
   return;
}

Common::JSONObject object = json->asObject();
...
```

sev-:
I just rebased everything to the tip of the master

Tkachov:
I've added a detailed description of Cloud and Local Webserver / Wi-Fi Sharing features to the wiki:
[Cloud Storage Support](http://wiki.scummvm.org/index.php/Cloud_Storage_Support)
[Local Webserver](http://wiki.scummvm.org/index.php/Local_Webserver)

uruk:
I also added the two documentation pages to our Developer Central: http://wiki.scummvm.org/index.php/Developer_Central#Networking

backends/networking/sdl_net/uploadfileclienthandler.cpp
```diff
@@ -0,0 +1,194 @@
...
+				_headersStream = new Common::MemoryReadWriteStream(DisposeAfterUse::YES);
+
+			if (client->readBlockHeaders(_headersStream)) {
+				handleBlockHeaders(client);
+				continue;
+			}
+			break;
+
+		case UFH_READING_BLOCK_CONTENT:
+			// _contentStream is created by handleBlockHeaders() if needed
```
wjp: I don't see how you're handling the error case where it isn't created.

criezy:
I have not looked at the code yet, but from the description in this PR I have one suggestion and one question.

The suggestion would be to have a "Paste from Clipboard" button in the tab where we need to enter the access token. From what I understand, the way pasting is implemented currently we would have to do it eight time for each pieces fo the access token? The possibility to paste the full token in one go would be great. Also I am assuming currently pasting is done using a keyboard shortcut? In that case the button would help on devices that do not have a keyboard (it suspect it would at least be easier to implement than the long press usually used to popup the copy/paste menu, although that would be great too :P).

The question is about the local server and how tied it is to SDL Net? Is it implemented in such a way (e.g. with abstraction layers) that it would be easy for porters for platforms that do not use SDL to implement it in a different way?

criezy:
There are various warnings and compilation errors.

Some examples of warnings:
```
engines/testbed/cloud.cpp:121:82: warning: format specifies type 'unsigned int' but the argument has type 'const char *' [-Wformat]
                Testsuite::logPrintf("Info! First directory is '%u' and first file is '%s'\n", directory.c_str(), file.c_str());
                                                                ~~                             ^~~~~~~~~~~~~~~~~
                                                                %s
engines/testbed/cloud.cpp:137:95: warning: format specifies type 'unsigned long' but the argument has type 'uint32' (aka 'unsigned int')
      [-Wformat]
        Testsuite::logPrintf("Info! It's id = '%s' and size = '%lu'\n", response.value.id().c_str(), response.value.size());
                                                               ~~~                                   ^~~~~~~~~~~~~~~~~~~~~
                                                               %u
engines/testbed/cloud.cpp:154:95: warning: format specifies type 'unsigned long' but the argument has type 'size_type' (aka 'unsigned int')
      [-Wformat]
                Testsuite::logPrintf("Warning! %lu files were not downloaded during folder downloading!\n", response.value.size());
                                               ~~~                                                          ^~~~~~~~~~~~~~~~~~~~~
                                               %u
./gui/storagewizarddialog.h:49:7: warning: private field '_stopServerOnClose' is not used [-Wunused-private-field]
        bool _stopServerOnClose;
             ^
```

And some example of errors (disclaimer: I have disabled the use of SDL Net):
```
gui/options.cpp:1322:28: error: no member named 'LocalWebserver' in namespace 'Networking'
        uint32 port = Networking::LocalWebserver::getPort();
                      ~~~~~~~~~~~~^
gui/storagewizarddialog.cpp:198:7: error: use of undeclared identifier 'kStorageCodePassedCmd'
        case kStorageCodePassedCmd:
             ^
gui/storagewizarddialog.cpp:199:39: error: use of undeclared identifier 'LocalServer'
                CloudMan.connectStorage(_storageId, LocalServer.indexPageHandler().code());
                                                    ^
```

I am aware this is all code in progress, so I will stop there in error reporting. But they should all be fixed before this PR is merged.

Tkachov:
Good suggestion about the "Paste from clipboard" button! Right now it works when built with SDL2, via Ctrl+V shortcut.

`LocalWebserver` class itself uses SDL_Net sockets (and socket set). `Client` does too. Other classes interact with these, so these would use sockets and return some result. SDL_Net sockets are not really complex, and I believe could be easily replaced with usual sockets (IIRC there were socket sets of some kind in POSIX). Yet, in order to use other sockets instead of SDL_Net's, we'd have to introduce a few own classes, as I didn't.

I fixed compilation without SDL_Net or libcurl a few times, but I guess I've broken that again. One of the warnings you've posted is really useful (`%u` for string somehow). I've used a lot of uint64, so I guess I used `%lu` automatically in the other two. We could avoid the last one, if we'd hide the field with `#ifdef`, but after that we might end up having an `undefined identifier` error instead. I'll try to fix these.

criezy:
I had some time to look a bit more at the code. Great work!
However I have one more warning and a couple of remarks/suggestions you might want to discuss with your mentor.

First the warning is:
```
gui/storagewizarddialog.cpp:50:43: warning: commas at the end of enumerator lists are a C++11 extension [-Wc++11-extensions]
        kStorageWizardContainerReflowCmd = 'SWCr',
                                                 ^
```

The first remark is about the copy/paste feature. I am thinking that rather than use directly SDL2 in the gui code it might be better to extend OSystem and let backends implement the functionality (with for example SDL2, but not necessarily). For example you might want in OSystem to add:

- kFeatureClipboard
- bool hasTextInClipboard
- String getTextFromClipboard

(and if we need copy at some point we can add a copyTextToClipboard)

The second remark is about the use of CURLOPT_XFERINFOFUNCTION, which was added in curl 7.32 (which is "only" 3 years old :P ). This causes compilation to fails on my system. I suppose I could compile a more recent curl manually and use that, but I am wondering about the possibility to use CURLOPT_PROGRESSFUNCTION rather than CURLOPT_XFERINFOFUNCTION to support a wider range of curl versions and ease the pain for port maintainers.

Tkachov:
Got it about the clipboard, would move it to OSystem.

While reading the docs, I thought `CURLOPT_PROGRESSFUNCTION` is deprecated. It actually is "obsolete". In their code they use both by adding `#if LIBCURL_VERSION_NUM >= 0x072000`. I'll do the same then.

backends/networking/sdl_net/uploadfileclienthandler.cpp
Tkachov: Woah, I thought the comment is about `new` up there and handling a situation when object is not allocated. Well, I'm not handling such situations mostly, because it's difficult to do something when you can't allocate more memory.

As for `_contentStream`, if it isn't created, it's `nullptr`. It's passed to `Client::readBlockContent()`. As mentioned [in the wiki](http://wiki.scummvm.org/index.php/Local_Webserver#How_LocalWebserver_works) (and also in `Client`'s description comment), `Client` uses `Reader` internally.

And when you pass `nullptr` as a stream to `Reader`, it reads contents into the void (reads, sees there is no stream, does nothing). So, passing `nullptr` here is similar to `> /dev/null` in a terminal.

backends/cloud/box/boxstorage.cpp
Tkachov: Those short links work fine. It looks like Box doesn't check the `redirect_uri` parameter here.

But I just checked Box docs/app editing page... Box allows only one redirect_uri. And localhost option is available for apps in development only. I guess we'd have to use https://www.scummvm.org/c/code option for Box all the time, even if `LocalWebserver` is available.

backends/platform/sdl/sdl.cpp
```diff
@@ -64,6 +64,11 @@
 #include <SDL/SDL_net.h>
 #endif
 
+#if SDL_VERSION_ATLEAST(2, 0, 0)
```
mgerhardy: This looks weird - the SDL_VERSION_ATLEAST macro should only be known if you already included SDL.h.

also (not sure about the scummvm guidelines) it looks wrong to include SDL2/SDL.h - just include SDL.h and let the include paths be configured correctly (e.g. pkg-config --cflags sdl2 will print /usr/include/SDL2 on my system - and this is also the default that SDL2 configures in their pc files). If you already include SDL.h for version >= 2.0.0, SDL_clipboard.h is also automatically included.

mgerhardy:
I've read in your blog post that ios openurl is still missing, check out this: https://github.com/mgerhardy/caveexpress/blob/master/src/modules/common/ports/IOSObjc.mm

Tkachov:
@mgerhardy, yeah, I guess that should do. I can add it then, but I have no means to check that it compiles and works fine.

mgerhardy:
@Tkachov You know that Travis CI also supports macosx and thus ios builds, no? See  e.g here: https://travis-ci.org/mgerhardy/caveexpress - quite useful, as I also don't fire up my mac for each commit and don't even have windows (see appveyor ci - e.g. https://ci.appveyor.com/project/mgerhardy/caveexpress/history).

Tkachov:
@mgerhardy Well, I like checking that something compiles and works before I commit it, not after. Plus, I'm not really sure how these .mm work, or which headers are required (in that example you've posted `<SDL.h>` is included though I don't think it's needed), or whether I'd need to define `FORBIDDEN_SYMBOL_ALLOW_ALL`, etc.

I mean, I can try commiting something I think should compile and work, then look at the result, try fixing something and then commiting again until it works... But I'd rather leave iOS code to someone who knows how to handle it.

uruk:
Sorry for not stating my opinion on this PR sooner! Somehow I assumed since I am the mentor, everybody automatically thought that I reviewed and approved all of the code before advising Alexander to start the PR... :) Obviously it was not the case, sorry for this false assumption.
So in my opinion the code is in a very good shape and ready for merge. I think the documentation both on our wiki and on Alexander's blog is sufficient and thorough and it'll help future developers to start hacking on this part of ScummVM. I also don't have any concerns regarding the security of the local webserver code that haven't been dealt with so far, thanks to the valuable comments above.
If nobody else has any objections (mainly @sev- and @wjp), I think it's time to merge this PR, before it starts to rot.

sev-:
Thus, after discussing with @wjp, we're merging this PR.

I do not think it will go into 1.9.0, which will be started soon, but we need to work on it so it ends up in 1.10.0

Tkachov, congratulations!

Tkachov:
Hooray! =)

uruk:
Congratulations Alexander, great job! :)

PR #790, opened 2016-07-23
ALL: Don't use 'defined' in macro definitions

salty-horse:
This is undefined behavior and clang warns about it. Specifically, it's not supported by Visual Studio.
See <http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20160118/147239.html>.

I haven't tested these changes carefully. Would be helpful to have more eyeballs :)

### DIFF
--- a/backends/graphics/opengl/opengl-sys.h
+++ b/backends/graphics/opengl/opengl-sys.h
@@ -48,9 +48,15 @@
 //  0 - Force OpenGL context
 //  1 - Force OpenGL ES context
 //  2 - Force OpenGL ES 2.0 context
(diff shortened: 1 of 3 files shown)


salty-horse:
Please hold, turns out there are other places where this is a problem that I missed due to a flood of other warnings :)

salty-horse:
I forgot to mention that it's clang 4.0 (dev build) that warns about it.

wjp:
Looks good to me.

wjp:
All buildbot toolchains seem happy with it too. Thanks!

PR #791, opened 2016-07-27
SCI32: Transitions (kSetShowStyle, kSetScroll)

csnover:
This PR cleans up the existing transitions code, and implements all other transitions that I could find in SCI32 games:

* Iris in/out (SCI2/2.1early, used regularly by PQ4)
* Pixel dissolve (SCI2/2.1early, ‘used’ by GK1 though not visible; SCI2.1mid+, used by a single scene in SQ6)
* Plane picture scroll (LSL6 hires intro)
* Palette morph (seems to be called incidentally by at least KQ7 1.51)

I feel like there are probably more used transitions than these, but since kSetShowStyle is called indirectly through the Styler game script, it’s a challenge to find all places where it is used. So, for the moment, just like with the magnifier view, any time a game calls to use one of the unknown transition styles, the engine will emit an error message asking the user to tell us what they were doing when that happened, and quit.

This PR also includes a commit that should allow hunk allocations to opt-out of VM garbage collection. This is necessary to avoid GC of the pixel dissolve bitmap, which is never retained by a game script. (The same problem occurs with Robot cels, so this change is needed for Robot, too, and was also an issue for ScrollWindow, which had previously been hooked directly into the GC.)


engines/sci/engine/seg_manager.h
```diff
@@ -225,9 +225,11 @@ class SegManager : public Common::Serializable {
 	 * @param[in] size		Number of bytes to allocate for the hunk entry
 	 * @param[in] hunk_type	A descriptive string for the hunk entry, for
 	 *						debugging purposes
+	 * @param[in] gc        Whether to make the hunk eligible for garbage
+	 *                      collection
```
bluegr: IMHO, it would be more preferable to store persistent hunk entries in a separate segment type (e.g. PersistentHunk), instead of changing the behavior of the Garbage Collector. Such entries can then be persisted in saved games, as well (like dynamic text objects).

bluegr:
Great work overal, well donel!
The persistent memory issue should be resolved, before this is merged

csnover:
OK, added the bitmap segment. Any thing else?

wjp:
Looks good!

bluegr:
Thanks for the changes :)

Could you please implement the bitmap sync code for saved games, so that this whole feature is fully implemented before merging it into master? Other than that missing functionality, it looks great :)

bluegr:
Excellent, thanks! :)

PR #793, opened 2016-07-28
WAGE: Add save/load code

Tkachov:
This adds some save/load code (based on asvitkine/wage-engine). It works, but there is no GUI dialogs to save or load yet.

### DIFF
--- a/engines/wage/entities.cpp
+++ b/engines/wage/entities.cpp
@@ -80,6 +80,8 @@ Context::Context() {
 }
 
 Scene::Scene() {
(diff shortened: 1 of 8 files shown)


bluegr:
Travis-CI reports that this currently breaks compilation: in saveload.cpp, Wage::Chr::ChrArmorType is not defined

Tkachov:
Yeah, I'll look into it soon. It was fine in VS, so I missed that.

sev-:
Cool, merging.

PR #794, opened 2016-07-28
Fix tests when building with the event recorder

bgK:
Remove the EventRecorder dependency from OSystem, pushing it down to SDL. The event recorder only works with SDL backends, so it should be fine.


sev-:
Fine, thank you.

PR #795, opened 2016-07-29
POSIX: Add $SNAP to search path if available.

fuzzie:
This allows ScummVM to find data files while running in a snap (e.g.
from the new Ubuntu store).

### DIFF
--- a/backends/platform/sdl/posix/posix.cpp
+++ b/backends/platform/sdl/posix/posix.cpp
@@ -141,6 +141,22 @@ Common::String OSystem_POSIX::getDefaultConfigFileName() {
 	return configFile;
 }
 
+void OSystem_POSIX::addSysArchivesToSearchSet(Common::SearchSet &s, int priority) {
+	const char *snap = getenv("SNAP");
+	if (snap) {
+		Common::String dataPath = Common::String(snap) + DATA_PATH;
+		Common::FSNode dataNode(dataPath);
+		if (dataNode.exists() && dataNode.isDirectory()) {
+			// This is the same priority which is used for the data path (below),
+			// but we insert this one first, so it will be searched first.
+			s.add(dataPath, new Common::FSDirectory(dataNode, 4), priority);
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


fuzzie:
This should respect the prefix, and the snap path should be higher priority than the data path added by the superclass. I'll fix these shortly.

I think it makes sense to keep the data path in the search path, to avoid breaking things if someone has $SNAP set to something random. That's easy enough to change too, though.

backends/platform/sdl/posix/posix.cpp
```diff
@@ -141,6 +141,22 @@ Common::String OSystem_POSIX::getDefaultConfigFileName() {
 	return configFile;
 }
 
+void OSystem_POSIX::addSysArchivesToSearchSet(Common::SearchSet &s, int priority) {
+	const char *snap = getenv("SNAP");
+	if (snap) {
```
sev-: Why not reverse the condition? E.g.

```
    if (!snap)
        return;

    Common::String dataPath = Common::String(snap) + DATA_PATH;
    ...
```

wjp: It should still call `OSystem_SDL::addSysArchivesToSearchSet` even when `SNAP` is not set.

sev-:
Nice, merging.

rootfather:
@fuzzie, @spads-spads: Do you consider the snapcraft support to be stable enough to be mentioned in the Linux port section in our NEWS file?

spads-spads:
@rootfather It's stable enough that I've been able to build arbitrary revisions from git and play Beneath A Steel Sky on them.  There's work to be done to tighten the access required, so we can snap up self-contained games with scummvm inside so they don't need external filesystem access.
But from a simple packaging formats standpoint I think this is stable enough to announce!

rootfather:
@spads-spads Thanks for your reply. If you don't have any objections, I'll mark the snapcraft support as "preliminary" or "basic" in the appropriate NEWS item until full functionality is achieved.

spads-spads:
Sounds good!

PR #797, opened 2016-07-29
Added an appdata.xml file

Mailaender:
This adds metadata for Linux software stores. See https://people.freedesktop.org/~hughsient/appdata/ for reference. My main motivation is to get you listed in https://software.opensuse.org/packages/Games

### DIFF
--- a/ports.mk
+++ b/ports.mk
@@ -20,6 +20,8 @@ install:
 	$(INSTALL) -c -m 644 $(DIST_FILES_THEMES) $(DIST_FILES_ENGINEDATA) "$(DESTDIR)$(datadir)/"
 	$(INSTALL) -d "$(DESTDIR)$(datarootdir)/applications"
 	$(INSTALL) -c -m 644 "$(srcdir)/dists/scummvm.desktop" "$(DESTDIR)$(datarootdir)/applications/scummvm.desktop"
+	$(INSTALL) -d "$(DESTDIR)$(datarootdir)/appdata"
+	$(INSTALL) -c -m 644 "$(srcdir)/dists/scummvm.appdata.xml" "$(DESTDIR)$(datarootdir)/appdata/scummvm.appdata.xml"
 ifdef DYNAMIC_MODULES
 	$(INSTALL) -d "$(DESTDIR)$(libdir)/scummvm/"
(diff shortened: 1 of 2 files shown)


sev-:
My understanding is that this file then must be installed in every system to /usr/share/appdata/scummvm.appdata.xml. If so, could you then help with modification or at least testing of the Makefile change? The file to modify is ports.mk, line 20 or so.

sev-:
Excellent, thank you! Merging.

Mailaender:
WAIT!

sev-:
ah, wait. Sorry for the late pointing out, but I completely missed that your commit log messages do not conform with our conventions.

Please add 'DISTS:" and "BUILD:" prefixes to the commits, and then I merge.

ports.mk
```diff
@@ -18,6 +18,8 @@ install:
 	$(INSTALL) -c -m 644 $(DIST_FILES_DOCS) "$(DESTDIR)$(docdir)"
 	$(INSTALL) -d "$(DESTDIR)$(datadir)"
 	$(INSTALL) -c -m 644 $(DIST_FILES_THEMES) $(DIST_FILES_ENGINEDATA) "$(DESTDIR)$(datadir)/"
+	$(INSTALL) -d "$(DESTDIR)$(datadir)/applications"
```
Mailaender: You have weird definition of `datadir`. It expands to `/usr/share/scummvm/` instead of `usr/share/`.

ports.mk
```diff
@@ -18,6 +18,10 @@ install:
 	$(INSTALL) -c -m 644 $(DIST_FILES_DOCS) "$(DESTDIR)$(docdir)"
 	$(INSTALL) -d "$(DESTDIR)$(datadir)"
 	$(INSTALL) -c -m 644 $(DIST_FILES_THEMES) $(DIST_FILES_ENGINEDATA) "$(DESTDIR)$(datadir)/"
+	$(INSTALL) -d "$(DESTDIR)$(datarootdir)/applications"
```
sev-: You also need to put it to the install-strip, and to uninstall targets

Mailaender:
Updated and successfully tested the install rules https://build.opensuse.org/request/show/415957

sev-:
Thanks!

PR #798, opened 2016-07-29
SCI32: Support Torin's autosave system

wjp:
Torin has two sets of saves: autosave.### and torinsg.###, both with
their own slots and .cat file. The autosave system uses autosave.000
and autosave.001. It also checks the presence of autosave.cat to
determine if it should show the chapter selection menu on startup.

### DIFF
--- a/engines/sci/engine/kfile.cpp
+++ b/engines/sci/engine/kfile.cpp
@@ -688,7 +688,7 @@ reg_t kSave(EngineState *s, int argc, reg_t *argv) {
 #endif
 
 reg_t kSaveGame(EngineState *s, int argc, reg_t *argv) {
-	Common::String game_id;
+	Common::String game_id = !argv[0].isNull() ? s->_segMan->getString(argv[0]) : "";
 	int16 virtualId = argv[1].toSint16();
 	int16 savegameId = -1;
 	Common::String game_description;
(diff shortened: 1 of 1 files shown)


csnover:
I’ve been running this patch the last few days and it seems to not break anything and allows the main menu in TP to show up. However, I’m curious that there doesn’t seem to be any autosave-specific code in SSCI that I can see (by searching for "autosave"). Are we layering hacks on hacks here, or is there some specific thing I don’t know about how ScummVM is handling saves?

bluegr:
That's not a hack. Games in the SCI engine create saved games based on the game ID, plus 3 digits for the save slot number, e.g. torin.010. In this case, Torin creates a different set of automatic saves, which are then used in the chapter selection screen. We can't use the name "autosave.xxx", as that would be too generic, and all games create saved games in the same folder, so there could be a name clash. Thus, this pull request effectively makes the engine create saves like gameid-autosave.xxx, e.g. torin-autosave.001.

Having checked this, I can say that it's great, and I'm OK to merge it as-is. Kudos, nice work! :)

wjp:
To elaborate on @bluegr's answer slightly: we ignored the `gameID` argument to the saving kernel calls. (It is a kind of namespace for save games.) This PR adds just enough support for it to support Torin's autosaves.

wjp:
I added a commit that bypasses `autosave.cat`.

bluegr:
Excellent, thanks! :)

For completeness sake: is kAutoSave also used for auto-saving in Torin?

wjp:
Haven't seen kAutoSave in there.

bluegr:
Thanks. Overall, great work :) Merging

PR #799, opened 2016-07-30
SCI32: kShowMovie

csnover:
This is the implementation of SCI32 version of kShowMovie for DOS and Windows. I tested all the SCI32 games I’m aware of that use AVI/SEQ video and everything seems to work as it does in the actual engine, with these intentional exceptions:

1. GK1 SEQ sequences are scaled in hi-res mode (in the original, it drew a 200x160 square in a 640x480 window)
1. GK1 AVI credits sequence is unscaled since it is a 640x480 AVI (in the original, it gets weirdly downscaled in a way that is obviously wrong)

And this one defect, which I don’t think is likely to be fixed terribly easily/without some hacks:

1. KQ7 1.51 movie playback backgrounds are white instead of black due to palette conflicts. The video uses palette entry 0 for white, but the background pic expects entry 0 to be black. I *think* this works in SSCI because the video palette is held separately by VfW, but I am not sure, since it is currently not possible to disassemble the Phar Lap binary at all. This is not a problem for KQ7 2.00b.

### DIFF
--- a/engines/sci/graphics/video32.cpp
+++ b/engines/sci/graphics/video32.cpp
@@ -161,6 +161,9 @@ AVIPlayer::IOStatus AVIPlayer::open(const Common::String &fileName) {
 }
 
 AVIPlayer::IOStatus AVIPlayer::init1x(const int16 x, const int16 y, int16 width, int16 height) {
+	if (_status == kAVINotOpen) {
+		return kIOFileNotFound;
+	}
 
 	_pixelDouble = false;
 
@@ -189,7 +192,7 @@ AVIPlayer::IOStatus AVIPlayer::init1x(const int16 x, const int16 y, int16 width,
 	// SCI2.1mid uses init2x to draw a pixel-doubled AVI, but SCI2 has only the
 	// one play routine which automatically pixel-doubles in hi-res mode
 	if (getSciVersion() == SCI_VERSION_2) {
-		// NOTE: This is somewhat of a hack; credits.avi from GK1 is not
+		// This is somewhat of a hack; credits.avi from GK1 is not
 		// rendered correctly in SSCI because it is a 640x480 video, but the
 		// game script gives the wrong dimensions. Since this is the only
 		// high-resolution AVI ever used, just set the draw rectangle to draw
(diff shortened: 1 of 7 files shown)


engines/sci/graphics/video32.cpp
```diff
@@ -29,11 +29,401 @@
...
+	// the animation does not need the extra offsets from the game in order to
+	// be correctly positioned in the middle of the window, so we ignore them.
+	_screenItem = new ScreenItem(_plane->_object, celInfo, Common::Point(0, 0), ScaleInfo());
+	g_sci->_gfxFrameout->addScreenItem(*_screenItem);
+	g_sci->_gfxFrameout->frameOut(true);
+	_decoder->start();
+
+	while (!g_engine->shouldQuit() && !_decoder->endOfVideo()) {
+		if (_decoder->endOfVideo()) {
+			break;
```
wjp: This looks like a duplicate check.

wjp:
One small issue which I accidentally ran into: it doesn't behave very nicely if an .avi file is missing. I'm unsure if this is because of a wrong error code, scripts not handling it, or something else. (I get a message `Could not initialize color format.` because it tries to set an invalid pixel format from an un-opened AVIPlayer.)

wjp:
Other than that, and the palette issue you already mention, looks good!

engines/sci/graphics/video32.cpp
csnover: Good catch. That’s what I get for copying stuff from the VMD code :)

csnover:
All this has landed at c8affb54cca259f37522216bad739be085bf9caa.

Added some error handling (which was just plain missing in SSCI), so now if the game can’t find the videos they will simply not be played and everything will continue on.

Also, in a moment of clarity, I was able to resolve easily the problem with the KQ7 1.51 videos by rewriting pixels on the fly. So that’s fixed too.

csnover:
(P.S. Thanks for the review!)

csnover:
Backed out due to buildbot failures that I can’t fix tonight.

csnover:
Relanded successfully today at 956cbf047a580ae5907f437b89fc2ae04efe2860.

PR #800, opened 2016-08-02
Add snapcraft.yaml so you can `snapcraft build`

spads-spads:
Now that https://github.com/scummvm/scummvm/pull/795 is merged, enabling snaps, we can merge the snapcraft.yaml.

This file is essentially the "Makefile" for the `snapcraft` binary, which will build a squashfs that can be mounted and run in a heavily AppArmor-constrained process space.

Right now this snapcraft.yaml allows access to the user's home directory through the `home` plug on line 15, but in future it would be possible to bundle games into their own snaps along with specific revisions of scummvm, and those could operate entirely without filesystem access outside the snap jail.  This can work no matter the base OS, as the environment overlays atop a ~60MB `ubuntu-core` image.

I've demonstrated this work in person to fuzzie, and have built 1.9.0git snaps using `snapcraft cleanbuild` which even does the build step in an lxd of the `ubuntu-core` environment.

I'm not sure how best to document this build information, so any guidance there would be welcomed.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,63 @@
+name: scummvm
+version: "1.9.0git"
+summary: ScummVM
+description: |
+    ScummVM is a program which allows you to run certain classic graphical
+    point-and-click adventure games, provided you already have their data
+    files. The clever part about this: ScummVM just replaces the executables
+    shipped with the game, allowing you to play them on systems for which
+    they were never designed!
+confinement: strict
+
+apps:
+  scummvm:
+    command: scummvm
+    plugs: [x11, home, pulseaudio, unity7, opengl]
+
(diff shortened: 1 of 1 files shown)


spads-spads:
It's also possible to have travis build these and upload to the "edge" channel, as in: https://github.com/evandandrea/cassandra-snap/blob/master/.travis.yml

sev-:
It would be cool if you add the description to our Wiki. Also if you have time for building these for the releases, and probably help with the packaging subsystem, which describes the game and has several output interfaces, including the snapcraft format, that would be awesome.

wjp:
Can we move this to a subdirectory? `dists/snapcraft` maybe?

PR #801, opened 2016-08-02
MOHAWK: Load games from outside the options dialog loop

criezy:
Loading games from inside the dialog loop may cause cursor glitches. This commits applies the same fix to the MohawkOptionsDialog as the one that already existed for the global main menu. This finished to fix bug #7164.

Since I am not familiar with the mohawk engine I prefer to propose this as a pull request rather than push directly to master.

### DIFF
--- a/engines/titanic/core/game_object.cpp
+++ b/engines/titanic/core/game_object.cpp
@@ -1371,11 +1371,6 @@ void CGameObject::petDisplayMessage(const CString &msg) {
 		pet->displayMessage(msg);
 }
 
-int CGameObject::petGetRooms1D0() const {
-	CPetControl *petControl = getPetControl();
-	return petControl ? petControl->getRooms1D0() : 0;
-}
-
 void CGameObject::petInvChange() {
 	CPetControl *pet = getPetControl();
 	if (pet)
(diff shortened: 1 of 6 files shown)


bgK:
Good catch !
However I would prefer if you did not reuse the same variable as the GMM for the slot to be loaded. It introduces a dependency between MohawkOptionsDialog and the GMM's loading facilities which were previously unrelated. Can you please add a variable in MohawkOptionsDialog and use it instead?

criezy:
Sure I will do the change this evening.

Re-using the variable from the base Engine class was one aspect I wasn't sure about. It looked more related to the Engine loadGameState facility to me than specifically to the GMM, which is why I decided to use it in the end. But I understand you might prefer to instead use a dedicated variable in MohawkOptionsDialog.

criezy:
For some reason my force push to my branch caused the pull request to close itself and indicate that there is no commit in the PR (despite the force push replacing one commit with another). I am a bit confused by that...

Anyway I will commit to master directly rather than open another PR since the change is quite simple and now follows your suggestion to add a variable in the MohawkOptionsDialog.

bgK:
Thanks

PR #802, opened 2016-08-07
TITANIC: Add detection entry for German version

rootfather:
This adds an entry to the detection table for the German version of Starship Titanic, GoG release.

The detection itself works, but there seems to be something going wrong. The create_titanic tool is unable to process the German ST.exe. The application just hangs and creates a broken file. Using the titanic.dat file from the English GoG release causes the game to exit immediately with a black screen.

Maybe this detection entry is useful at the moment. If not, feel free to simply close this PR.

### DIFF
--- a/engines/titanic/detection_tables.h
+++ b/engines/titanic/detection_tables.h
@@ -34,6 +34,18 @@ static const TitanicGameDescription gameDescriptions[] = {
 			GUIO1(GUIO_NONE)
 		},
 	},
+	
+	{
+		{
+			"titanic",
+			0,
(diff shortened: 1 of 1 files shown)


dreammaster:
It certainly won't hurt to have it in the detection list as disabled for now, if only as a reminder that the German version exists.

Interesting that the English dat file can't be used with the German data, nor can the tool generate a valid DAT file from it. That means that the German version definitely contains differences from the English version. I was kind of hoping, given the game's support for synonym lists, that the German version was pretty much identical to the English version, just with different text within the dialogue files, dialogs, and speech. This seems to indicate otherwise.

PR #804, opened 2016-08-12
SCI32: Remove GfxScreen from SCI32

csnover:
This PR removes the last remaining bits of GfxScreen from SCI32, so graphics for SCI16 games can be handled exclusively by GfxScreen and graphics for SCI32 games are handled exclusively by GfxFrameout.

It also includes updated mouse implementation that fixes a few bugs:

1. Mouse cursor is now warped to the correct place;
2. Games that mix low-resolution and high-resolution cursor views will now only upscale the low-resolution cursors (based on a heuristic of the size of the view);
3. Mouse cursor is now correctly bound to the restricted rectangle, so views don’t bounce when the edges of the restricted rectangle are reached. This also fixes the binoculars in GK1.

### DIFF
--- a/engines/sci/engine/savegame.cpp
+++ b/engines/sci/engine/savegame.cpp
@@ -48,6 +48,7 @@
 #include "sci/sound/music.h"
 
 #ifdef ENABLE_SCI32
+#include "sci/graphics/cursor32.h"
 #include "sci/graphics/frameout.h"
 #include "sci/graphics/palette32.h"
 #include "sci/graphics/remap32.h"
@@ -887,6 +888,29 @@ void GfxRemap32::saveLoadWithSerializer(Common::Serializer &s) {
 
 	if (s.isLoading()) {
 		_needsUpdate = true;
+	}
+}
+
+void GfxCursor32::saveLoadWithSerializer(Common::Serializer &s) {
+	if (s.getVersion() < 37) {
+		return;
+	}
+
+	s.syncAsSint32LE(_hideCount);
+	s.syncAsSint16LE(_restrictedArea.left);
+	s.syncAsSint16LE(_restrictedArea.top);
+	s.syncAsSint16LE(_restrictedArea.right);
+	s.syncAsSint16LE(_restrictedArea.bottom);
+	s.syncAsUint16LE(_cursorInfo.resourceId);
+	s.syncAsUint16LE(_cursorInfo.loopNo);
+	s.syncAsUint16LE(_cursorInfo.celNo);
+
+	if (s.isLoading()) {
+		hide();
+		setView(_cursorInfo.resourceId, _cursorInfo.loopNo, _cursorInfo.celNo);
+		if (!_hideCount) {
+			show();
+		}
 	}
 }
 #endif
--- a/engines/sci/graphics/celobj32.cpp
+++ b/engines/sci/graphics/celobj32.cpp
@@ -45,7 +45,7 @@ void CelScaler::activateScaleTables(const Ratio &scaleX, const Ratio &scaleY) {
 		}
 	}
 
-	int i = 1 - _activeIndex;
+	const int i = 1 - _activeIndex;
 	_activeIndex = i;
 	CelScalerTable &table = _scaleTables[i];
 
@@ -65,7 +65,7 @@ void CelScaler::activateScaleTables(const Ratio &scaleX, const Ratio &scaleY) {
 void CelScaler::buildLookupTable(int *table, const Ratio &ratio, const int size) {
 	int value = 0;
 	int remainder = 0;
-	int num = ratio.getNumerator();
+	const int num = ratio.getNumerator();
 	for (int i = 0; i < size; ++i) {
 		*table++ = value;
 		remainder += ratio.getDenominator();
@@ -204,7 +204,7 @@ struct SCALER_Scale {
 		if (g_sci->_gfxFrameout->getCurrentBuffer().scriptWidth == kLowResX) {
 			const int16 unscaledX = (scaledPosition.x / scaleX).toInt();
 			if (FLIP) {
-				int lastIndex = celObj._width - 1;
+				const int lastIndex = celObj._width - 1;
 				for (int16 x = targetRect.left; x < targetRect.right; ++x) {
 					_valuesX[x] = lastIndex - (table->valuesX[x] - unscaledX);
 				}
@@ -220,7 +220,7 @@ struct SCALER_Scale {
 			}
 		} else {
 			if (FLIP) {
-				int lastIndex = celObj._width - 1;
+				const int lastIndex = celObj._width - 1;
 				for (int16 x = 0; x < targetRect.width(); ++x) {
 					_valuesX[targetRect.left + x] = lastIndex - table->valuesX[x];
 				}
@@ -261,7 +261,7 @@ private:
 #ifndef NDEBUG
 	const int16 _sourceHeight;
 #endif
-	byte *_pixels;
+	const byte *_pixels;
 	const int16 _sourceWidth;
 
 public:
@@ -270,7 +270,7 @@ public:
 	_sourceHeight(celObj._height),
 #endif
 	_sourceWidth(celObj._width) {
-		byte *resource = celObj.getResPointer();
+		const byte *resource = celObj.getResPointer();
 		_pixels = resource + READ_SCI11ENDIAN_UINT32(resource + celObj._celHeaderOffset + 24);
 	}
 
@@ -282,7 +282,7 @@ public:
 
 struct READER_Compressed {
 private:
-	byte *_resource;
+	const byte *const _resource;
 	byte _buffer[1024];
 	uint32 _controlOffset;
 	uint32 _dataOffset;
@@ -301,7 +301,7 @@ public:
 	_maxWidth(maxWidth) {
 		assert(maxWidth <= celObj._width);
 
-		byte *celHeader = _resource + celObj._celHeaderOffset;
+		const byte *const celHeader = _resource + celObj._celHeaderOffset;
 		_dataOffset = READ_SCI11ENDIAN_UINT32(celHeader + 24);
 		_uncompressedDataOffset = READ_SCI11ENDIAN_UINT32(celHeader + 28);
 		_controlOffset = READ_SCI11ENDIAN_UINT32(celHeader + 32);
@@ -311,14 +311,14 @@ public:
 		assert(y >= 0 && y < _sourceHeight);
 		if (y != _y) {
 			// compressed data segment for row
-			byte *row = _resource + _dataOffset + READ_SCI11ENDIAN_UINT32(_resource + _controlOffset + y * 4);
+			const byte *row = _resource + _dataOffset + READ_SCI11ENDIAN_UINT32(_resource + _controlOffset + y * 4);
 
 			// uncompressed data segment for row
-			byte *literal = _resource + _uncompressedDataOffset + READ_SCI11ENDIAN_UINT32(_resource + _controlOffset + _sourceHeight * 4 + y * 4);
+			const byte *literal = _resource + _uncompressedDataOffset + READ_SCI11ENDIAN_UINT32(_resource + _controlOffset + _sourceHeight * 4 + y * 4);
 
 			uint8 length;
 			for (int16 i = 0; i < _maxWidth; i += length) {
-				byte controlByte = *row++;
+				const byte controlByte = *row++;
 				length = controlByte;
 
 				// Run-length encoded
@@ -581,7 +581,7 @@ void CelObj::submitPalette() const {
 int CelObj::_nextCacheId = 1;
 CelCache *CelObj::_cache = nullptr;
 
-int CelObj::searchCache(const CelInfo32 &celInfo, int *nextInsertIndex) const {
+int CelObj::searchCache(const CelInfo32 &celInfo, int *const nextInsertIndex) const {
 	*nextInsertIndex = -1;
 	int oldestId = _nextCacheId + 1;
 	int oldestIndex = 0;
@@ -792,26 +792,27 @@ void CelObj::scaleDrawUncompNoMD(Buffer &target, const Ratio &scaleX, const Rati
 #pragma mark CelObjView
 
 int16 CelObjView::getNumLoops(const GuiResourceId viewId) {
-	Resource *resource = g_sci->getResMan()->findResource(ResourceId(kResourceTypeView, viewId), false);
+	const Resource *const resource = g_sci->getResMan()->findResource(ResourceId(kResourceTypeView, viewId), false);
 
 	if (!resource) {
 		return 0;
 	}
 
+	assert(resource->size >= 3);
 	return resource->data[2];
 }
 
 int16 CelObjView::getNumCels(const GuiResourceId viewId, const int16 loopNo) {
-	Resource *resource = g_sci->getResMan()->findResource(ResourceId(kResourceTypeView, viewId), false);
+	const Resource *const resource = g_sci->getResMan()->findResource(ResourceId(kResourceTypeView, viewId), false);
 
 	if (!resource) {
 		return 0;
 	}
 
-	byte *data = resource->data;
+	const byte *const data = resource->data;
 
-	uint16 loopCount = data[2];
-	if (loopNo >= loopCount) {
+	const uint16 loopCount = data[2];
+	if (loopNo >= loopCount || loopNo < 0) {
 		return 0;
 	}
 
@@ -819,10 +820,15 @@ int16 CelObjView::getNumCels(const GuiResourceId viewId, const int16 loopNo) {
 	const uint8 loopHeaderSize = data[12];
 	const uint8 viewHeaderFieldSize = 2;
 
-	byte *loopHeader = data + viewHeaderFieldSize + viewHeaderSize + (loopHeaderSize * loopNo);
+#ifndef NDEBUG
+	const byte *const dataMax = data + resource->size;
+#endif
+	const byte *loopHeader = data + viewHeaderFieldSize + viewHeaderSize + (loopHeaderSize * loopNo);
+	assert(loopHeader + 3 <= dataMax);
 
 	if ((int8)loopHeader[0] != -1) {
 		loopHeader = data + viewHeaderFieldSize + viewHeaderSize + (loopHeaderSize * (int8)loopHeader[0]);
+		assert(loopHeader >= data && loopHeader + 3 <= dataMax);
 	}
 
 	return loopHeader[2];
@@ -838,7 +844,7 @@ CelObjView::CelObjView(const GuiResourceId viewId, const int16 loopNo, const int
 	_transparent = true;
 
 	int cacheInsertIndex;
-	int cacheIndex = searchCache(_info, &cacheInsertIndex);
+	const int cacheIndex = searchCache(_info, &cacheInsertIndex);
 	if (cacheIndex != -1) {
 		CelCacheEntry &entry = (*_cache)[cacheIndex];
 		const CelObjView *const cachedCelObj = dynamic_cast<CelObjView *>(entry.celObj);
@@ -854,15 +860,14 @@ CelObjView::CelObjView(const GuiResourceId viewId, const int16 loopNo, const int
 	// generates view resource metadata for both SCI16 and SCI32
 	// implementations
 
-	Resource *resource = g_sci->getResMan()->findResource(ResourceId(kResourceTypeView, viewId), false);
+	const Resource *const resource = g_sci->getResMan()->findResource(ResourceId(kResourceTypeView, viewId), false);
 
 	// NOTE: SCI2.1/SQ6 just silently returns here.
 	if (!resource) {
-		warning("View resource %d not loaded", viewId);
-		return;
+		error("View resource %d not found", viewId);
 	}
 
-	byte *data = resource->data;
+	const byte *const data = resource->data;
 
 	_scaledWidth = READ_SCI11ENDIAN_UINT16(data + 14);
 	_scaledHeight = READ_SCI11ENDIAN_UINT16(data + 16);
@@ -881,7 +886,7 @@ CelObjView::CelObjView(const GuiResourceId viewId, const int16 loopNo, const int
 		}
 	}
 
-	uint16 loopCount = data[2];
+	const uint16 loopCount = data[2];
 	if (_info.loopNo >= loopCount) {
 		_info.loopNo = loopCount - 1;
 	}
@@ -896,7 +901,7 @@ CelObjView::CelObjView(const GuiResourceId viewId, const int16 loopNo, const int
 	const uint8 loopHeaderSize = data[12];
 	const uint8 viewHeaderFieldSize = 2;
 
-	byte *loopHeader = data + viewHeaderFieldSize + viewHeaderSize + (loopHeaderSize * _info.loopNo);
+	const byte *loopHeader = data + viewHeaderFieldSize + viewHeaderSize + (loopHeaderSize * _info.loopNo);
 
 	if ((int8)loopHeader[0] != -1) {
 		if (loopHeader[1] == 1) {
@@ -911,10 +916,14 @@ CelObjView::CelObjView(const GuiResourceId viewId, const int16 loopNo, const int
 		_info.celNo = celCount - 1;
 	}
 
+	if (_info.celNo < 0) {
+		error("Cel is less than 0!");
+	}
+
 	_hunkPaletteOffset = READ_SCI11ENDIAN_UINT32(data + 8);
 	_celHeaderOffset = READ_SCI11ENDIAN_UINT32(loopHeader + 12) + (data[13] * _info.celNo);
 
-	byte *celHeader = data + _celHeaderOffset;
+	const byte *const celHeader = data + _celHeaderOffset;
 
 	_width = READ_SCI11ENDIAN_UINT16(celHeader);
 	_height = READ_SCI11ENDIAN_UINT16(celHeader + 2);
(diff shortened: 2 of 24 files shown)


engines/sci/graphics/cursor32.cpp
```diff
@@ -0,0 +1,394 @@
...
+	const int16 screenWidth = g_sci->_gfxFrameout->getCurrentBuffer().screenWidth;
+	const int16 screenHeight = g_sci->_gfxFrameout->getCurrentBuffer().screenHeight;
+
+	_position.x = (position.x * Ratio(screenWidth, scriptWidth)).toInt();
+	_position.y = (position.y * Ratio(screenHeight, scriptHeight)).toInt();
+
+	g_system->warpMouse(_position.x, _position.y);
+}
+
+void GfxCursor32::gonnaPaint(Common::Rect paintRect) {
```
bluegr: Slang: "paint" should be pretty self-explanatory

bluegr:
Excellent work, well done! :)

Nice and clean separation of the graphics code, at last.

+1

engines/sci/graphics/cursor32.cpp
wjp: Except that that's not what it does...

engines/sci/engine/savegame.cpp
```diff
@@ -889,6 +890,29 @@ void GfxRemap32::saveLoadWithSerializer(Common::Serializer &s) {
 		_needsUpdate = true;
 	}
 }
+
+void GfxCursor32::saveLoadWithSerializer(Common::Serializer &s) {
+	if (s.getVersion() < 36) {
```
wjp: This should now be 37 I think.

engines/sci/graphics/celobj32.cpp
```diff
@@ -791,6 +791,43 @@ void CelObj::scaleDrawUncompNoMD(Buffer &target, const Ratio &scaleX, const Rati
...
+	const uint8 viewHeaderFieldSize = 2;
+
+	byte *loopHeader = data + viewHeaderFieldSize + viewHeaderSize + (loopHeaderSize * loopNo);
+
+	if ((int8)loopHeader[0] != -1) {
+		loopHeader = data + viewHeaderFieldSize + viewHeaderSize + (loopHeaderSize * (int8)loopHeader[0]);
+	}
+
+	return loopHeader[2];
+}
```
m-kiewitz: No resource size checks are done, which means corrupted resource could cause invalid memory access.

engines/sci/graphics/frameout.cpp
```diff
@@ -55,31 +57,37 @@
...
 		_isHiRes = false;
 	}
 
+	if (g_sci->getGameId() == GID_PHANTASMAGORIA) {
+		_currentBuffer = Buffer(630, 450, nullptr);
+	} else if (_isHiRes) {
+		_currentBuffer = Buffer(640, 480, nullptr);
+	} else {
+		_currentBuffer = Buffer(320, 200, nullptr);
+	}
```
m-kiewitz: It seems that according to the original code GK1 on Macintosh only supports hires mode and atm that is not implemented in the same manner.

I think clone2727 owned that version of the game. I can't check it.

engines/sci/graphics/frameout.cpp
```diff
@@ -1334,14 +1366,15 @@ void GfxFrameout::kernelSetNowSeen(const reg_t screenItemObject) const {
 
 	ScreenItem *screenItem = plane->_screenItemList.findByObject(screenItemObject);
 	if (screenItem == nullptr) {
-		error("kSetNowSeen: Screen item %04x:%04x not found in plane %04x:%04x", PRINT_REG(screenItemObject), PRINT_REG(planeObject));
+		return false;
```
m-kiewitz: why is a non existing object now ignored inside kSetNowSeen?
in sci32 that (also) results in a fatal error, so the original code was accurate.

engines/sci/graphics/screen.cpp
```diff
@@ -30,6 +30,9 @@
 #include "sci/engine/state.h"
 #include "sci/graphics/screen.h"
 #include "sci/graphics/view.h"
+#ifdef ENABLE_SCI32
+#include "sci/graphics/frameout.h"
+#endif
```
m-kiewitz: Makes no sense to me to add the header here.

engines/sci/graphics/screen.cpp
```diff
@@ -212,7 +192,7 @@ GfxScreen::GfxScreen(ResourceManager *resMan) : _resMan(resMan) {
 			initGraphics(_displayWidth, _displayHeight + 28 + 2, _displayWidth > 320);
 		else
 			error("Unknown SCI1.1 Mac game");
-	} else
+	} else if (getSciVersion() < SCI_VERSION_2)
 		initGraphics(_displayWidth, _displayHeight, _displayWidth > 320);
```
m-kiewitz: this is not needed and if something goes wrong, no initGraphics() call would be done.
So better keep it like it was or add an error() in case SciVersion is => SCI_VERSION_2.

engines/sci/graphics/screen.cpp
```diff
@@ -632,13 +612,13 @@ void GfxScreen::setVerticalShakePos(uint16 shakePos) {
...
-		if (directions & SCI_SHAKE_DIRECTION_VERTICAL)
+		if (directions & kShakeVertical)
 			setVerticalShakePos(10);
 		// TODO: horizontal shakes
 		g_system->updateScreen();
 		g_sci->getEngineState()->wait(3);
 
-		if (directions & SCI_SHAKE_DIRECTION_VERTICAL)
+		if (directions & kShakeHorizontal)
 			setVerticalShakePos(0);
```
m-kiewitz: Unneeded change and this is checking ShakeHorizontal instead of Vertical, which means it introduces a regression.

engines/sci/graphics/cursor32.cpp
```diff
@@ -0,0 +1,394 @@
...
+	if (_cursorBack.rect.isEmpty()) {
+		return;
+	}
+
+	readVideo(_cursorBack);
+	_drawBuff1.rect = _cursor.rect;
+	copy(_drawBuff1, _cursorBack);
+	paint(_drawBuff1, _cursor);
+	drawToHardware(_drawBuff1);
+}
```
m-kiewitz: Why can't we use SCUMMVM system functions for the cursor just like in the other cursor.cpp?

I'm not sure if that's the case but platforms could implement cursors in a way that performs better than doing all of that by ourselves.

engines/sci/graphics/cursor32.cpp
```diff
@@ -0,0 +1,394 @@
...
+	const int16 scriptWidth = g_sci->_gfxFrameout->getCurrentBuffer().scriptWidth;
+	const int16 scriptHeight = g_sci->_gfxFrameout->getCurrentBuffer().scriptHeight;
+	const int16 screenWidth = g_sci->_gfxFrameout->getCurrentBuffer().screenWidth;
+	const int16 screenHeight = g_sci->_gfxFrameout->getCurrentBuffer().screenHeight;
+
+	_position.x = (position.x * Ratio(screenWidth, scriptWidth)).toInt();
+	_position.y = (position.y * Ratio(screenHeight, scriptHeight)).toInt();
+
+	g_system->warpMouse(_position.x, _position.y);
+}
```
m-kiewitz: One should check if there are SCI32 games, that set the cursor to some location and then expect it to be there right after that. This happened in a few SCI16 games and caused issues with touchscreens and Wiimote on the Wii.

I added workaround code for such issues in cursor.cpp. Just to note.

engines/sci/graphics/frameout.cpp
m-kiewitz: And maybe we should also change to hires mode in case platform == windows.

Did all SCI32 Windows versions do hires mode? In that case we should implement it the same way. The hires flag should get used as well, especially because of games like Gabriel Knight 1, that supported hires in DOS and people may also select DOS instead of Windows without knowing that Windows would offer hires, which means we should also check that hires GUI option as well and default to hires in any case.

Although maybe the hires GUI option should not be available for games with platform == windows. In theory game scripts could expect hires mode as soon as platform == Windows.

engines/sci/graphics/cursor32.cpp
```diff
@@ -0,0 +1,394 @@
...
+			(g_sci->getGameId() == GID_GK1 ||
+			(g_sci->getGameId() == GID_PQ4 && _width <= 22 && _height <= 22))) {
+
+			_width *= 2;
+			_height *= 2;
+			_hotSpot.x *= 2;
+			_hotSpot.y *= 2;
+			pixelDouble = true;
+		}
+
```
m-kiewitz: Maybe we should go through all cursors of SCI32 games and then maybe create some sort of heuristic for it?

Harcoded GameIDs should be avoided whenever possible.

engines/sci/graphics/cursor32.cpp
```diff
@@ -0,0 +1,394 @@
...
+void GfxCursor32::setView(const GuiResourceId viewId, const int16 loopNo, const int16 celNo) {
+	hide();
+
+	_cursorInfo.resourceId = viewId;
+	_cursorInfo.loopNo = loopNo;
+	_cursorInfo.celNo = celNo;
+
+	if (viewId != -1) {
+		CelObjView view(viewId, loopNo, celNo);
+
```
m-kiewitz: On non existing views this will crash?!

engines/sci/graphics/frameout.cpp
lskovlun: Clone2727's Mac-specific code has been removed on a number of occasions previously. I'm not saying it was right to do so, only that this follows a general trend.

m-kiewitz: well in this case the deleted code right in this pull request did force hires mode for GK1 Mac.
And it seems GK1 Mac really is hires only.

It's a shame that probably no one of the team currently owns GK1 Mac.

csnover: Much like SQ6 was the primary focus of SCI32 development for many months in order to maintain focus, currently only DOS/Win are the platforms being target (and are the only ones that I am able to target, as I have no Mac version of anything). So, please don’t concern yourself at this point about that platform, it’s not time yet to try to do that work. :)

m-kiewitz: This is not about trying to get something to work. This is about keeping existing code behavior and not just removing it without any reason to do so.

We already know that GK1 on Mac seems to be hires only, so why not just keep that information inside the code instead of removing it now and having to add it again in some months. When I can not test something and verify that certain code is incorrect, I won't remove that code. Macintosh versions are not that easy to come by, which means that any information about those is highly valuable and shouldn't be discarded easily.

engines/sci/graphics/frameout.cpp
csnover: Per IRC discussion, all SCI2/2.1early games + *some* early SCI2.1mid games (SQ6, maybe MGDX, probably not Shivers) will quit with an error if a ScreenItem is not found, but all other SCI2.1mid+ games do what this code does. We can error out in the kernel function if it seems to be necessary for those older games to run properly. So far it doesn’t seem to be necessary.

engines/sci/graphics/cursor32.cpp
csnover: There is no point to such an endeavour, these are the only two games that exhibit this behaviour as they are the only two games ever written that let you toggle choose between lo-res and hi-res modes.

engines/sci/graphics/frameout.cpp
m-kiewitz: We should at the very least output a warning, so that errors are recognizable as such.
For SCI2/early games maybe we should even keep the error. It's not about running games properly, but being able to detect errors and not just skip over them, which could then cause all sorts of weird issues and could take ages to actually track down.

With the new code kSetNowSeen calls with invalid screen items could cause all sorts of issues, which would then be invisible to us and you don't know if the current code is perfect - and that even including the games, where such an issue in the original interpreter would literally "crash" the game (and where it would definitely be either a script error or our own fault). I don't see any issues with adding a warning. When someone plays through the games, that use the new code, we would at the very least see if such situations actually occur, could then write them down and maybe make sure that it's not our fault but how the games actually work.

And well, we could also just ignore the warnings anyway. But at least they would be there. If such a warning would happen all the time in those other games, then that's a different matter. But that wouldn't change SCI2/early behavior.

engines/sci/graphics/cursor32.cpp
m-kiewitz: Then I guess the comment should be fixed accordingly.

"Early SSCI" games is also irritating, because "Early SSCI games" normally means SCI0 games. SSCI even means Sierra-SCI, which is normally used to specify that something worked or didn't work when using the original interpreter and - when being pedantic - SSCI even means Sierra's Sierra's Creative Interpreter. Maybe "Early SCI32" games instead? That would be obvious.

Also maybe: "To address this, we make some slight adjustments to cursor display in the only two SCI32 games, that support both low and hires mode (and have those issues)."

engines/sci/graphics/frameout.cpp
csnover: > This is about keeping existing code behavior and not just removing it without any reason to do so.

All SCI32 graphics code was guesswork and hacks before I started working on it. I am not sure why any of the SCI32 Mac code would be exempt from this fact. In fact… (read next response)

> We already know that GK1 on Mac seems to be hires only, so why not just keep that information inside the code instead of removing it now and having to add it again in some months.

The default value for `enable_high_resolution_graphics` is true. So GK1 Mac will already be hires without adding a single line of special code to support it.

> Otherwise we could also just remove any Macintosh code altogether including all SCI32 Mac detection entries […]

This is a false equivalence. Detection table hashes are data, not code, and weren’t based on guesswork.

> And well, I could also remove Hoyle1+2+SQ3 Mac detection entries too, because those are also not working correctly atm and I had not enough nerve to try to figure out Sierra's upscaling algo used by those games.

If the games are totally broken they should at least have the `ADGF_UNSTABLE` flag.

engines/sci/graphics/cursor32.cpp
csnover: I don’t understand what the purpose of this comment is.

m-kiewitz: I mean it seems on non-existing views, CelObjView won't be fully initialized and the follow up code doesn't check for that, which means it will then work with uninitialized variables, correct?

engines/sci/graphics/frameout.cpp
m-kiewitz: GK1 Mac actually does not have the enable_high_resolution_graphics flag, which also makes sense when it's hires only. It wouldn't make sense to offer that option. It even seems that GK1 floppy also didn't offer hires graphics (I can't check that, I only own multiple CD versions) and only the DOS CD version did. So GK1 DOS floppy defaults to low-res and GK Mac floppy shouldn't. The original code made sense in that regard.

It's also a difference between the obvious guesswork code and code that for example sets resolutions and enforces such modes etc. At least clone2727 owned that version of the game and he wouldn't add such code, when the game is low-res only. I know that he made that version "work" back then even with the other guess-work code, so enforcing hires simply can't be bad.

The commit even says 24e983610dfb118c943d6a1867bf3ea5d81e18e4
"SCI: Fix GK1 Mac views, the game now starts"

At the very least you could add a comment saying "originally we enforced hires for GK1 mac here, but no one owns it atm, so we can't verify to make absolutely sure", but atm the code is simply removed and would be forgotten about and that's like throwing away a wiki page just for fun.

Anyway are you sure that the hashes were properly taken? I mean that's the same logic. Maybe those are also bad, who knows. No one can currently check GK1 Mac. Maybe they are incomplete? How do you know that they weren't based on guesswork and are perfectly fine, when you can not verify them?

I personally have to assume that both were correct. And that wasn't complicated code. It simply enforces hires for GK1 Mac. Hell, it could have been a simple option as well inside the detection tables instead, so we are basically discussing about a simple data flag and nothing more. I mean atm GK1 Mac got the options like GK1 DOS floppy, which means no hires option. How do you know that this data is accurate? You can't.

engines/sci/graphics/celobj32.cpp
csnover: Yes, that is true, even in SCI16’s GfxView code, and in SSCI itself. What is special about it here?

engines/sci/graphics/cursor32.cpp
csnover: The same thing would happen in SSCI. In ScummVM, the CelObjView constructor will emit a warning if a non-existing resource ID is given, if this ever actually happens (hasn’t so far!).

engines/sci/graphics/celobj32.cpp
m-kiewitz: This is new code, so one could do it properly. It's not like there would be 100 checks anyway and it would be tons of work.

I for example added checks in the new AGI view code, when I rewrote it. I also added tons of checks in the sherlock holmes code that I wrote.

Even in the current SCI view code, there are quite a lot of checks. Quite a few are also missing and that's also bad, but that's not the point. That should also get fixed. Same with the saved game code. My new save game code also checks for all sorts of issues. The current one is prone to crash in all sorts of ways, which is always bad.

engines/sci/graphics/cursor32.cpp
csnover: I don’t think I can do that before this lands. Playtesting should find any such issues later, if they exist, and then it can be addressed.

engines/sci/graphics/celobj32.cpp
m-kiewitz: I mean otherwise we can also simply stop using coverity, because who cares, right? I even wonder why parts of the current SCI view code wasn't complained about by coverity. It complained quite a bit about sherlock holmes code and I fixed all of those issues (at least hopefully).

engines/sci/graphics/cursor32.cpp
m-kiewitz: Ehm, SSCI checked basically nothing at all. That's not acceptable and it's very bad.
Is it a problem to error() out in such case? Or at the very least skip over the code?

SSCI also allowed the scripts to use uninitialized variables. And it didn't check kernel call parameters. SSCI also liked to crash a lot because of all those things. We could have been fine with that, but instead we did it properly and detect such issues. And that's why SCI games, at least the SCI16 ones normally run better in ScummVM than when using the original interpreter. ScummVM crashing is not really acceptable, especially when it's avoidable with a simple error().

engines/sci/graphics/cursor32.cpp
m-kiewitz: I wanted to make a note about that, so that these situations are known and that it's also known that there already is existing code to circumvent them.

engines/sci/graphics/celobj32.cpp
csnover: > This is new code, so one could do it properly. It's not like there would be 100 checks anyway and it would be tons of work.

You want only bounds checking in these two functions? OK, but it won’t stop OOB reads in the many other places resource data is consumed, and the only thing this will guard against is invalid large offsets in the view headers.

> otherwise we can also simply stop using coverity, because who cares, right?

Hyperbolic comments like this are not helpful. You know Coverity does a lot more than notifying about potentially unsafe uses of external data.

engines/sci/graphics/frameout.cpp
csnover: > GK1 Mac actually does not have the enable_high_resolution_graphics flag, which also makes sense when it's hires only. […]

I feel like there is a misunderstanding here.

When games are not given the `enable_high_resolution_graphics` (a.k.a. `GAMEOPTION_HIGH_RESOLUTION_GRAPHICS`) GUIO, that configuration option defaults to `true`. GK1 Mac is not given the GUIO option, so it always runs with high-resolution graphics. There’s no need to add extra GK1 Mac specific code, because it will get hires mode already, because of the way the system is designed. I hope this explanation clears things up.

> So GK1 DOS floppy defaults to low-res

It doesn’t in ScummVM at the moment; the GUIOs for all SCI32 games aren’t correct right now, which is something that needs to be addressed separately at a later time (like around the time we’re almost ready to start playtesting). I put it on my TODO list.

> Anyway are you sure that the hashes were properly taken? I mean that's the same logic. Maybe those are also bad, who knows. No one can currently check GK1 Mac. Maybe they are incomplete? How do you know that they weren't based on guesswork and are perfectly fine, when you can not verify them?

I have a somewhat higher confidence about the hashes because they’re automatically generated by code that is known to generate correct hashes, whereas I’ve been working on the SCI32 code for a while now and have a pretty low confidence that any of the old SCI32-specific code is actually correct.

m-kiewitz: Then that's a horrible ConfigManager design oversight. I thought it was already fixed, I think AGI had a similar issue. This definitely should not be the case. When the option is available, it should be true on default. But when it's not, it should return false. Atm it returns true for all sorts of SCI0->SCI1.1 games, which is of course incorrect.

>I have a somewhat higher confidence about the hashes because they’re automatically generated by code that is 
>known to generate correct hashes

The GUI-Options aren't automatically generated and several other information is also not generated, but added manually by hand.

Anyway, the point is still: this "code" does nothing else than a simple bit-switch. It forces hi-res, nothing else.
So in case hi-res mode was part of the detection-data, I have to assume that you would blindly trust it - on the other hand you don't trust any code, even when it's doing nothing else but a simple bit-switch.

Which to me makes no sense.

You are basically saying "who knows if Gabriel Knight 1 on Mac is really hi-res", although every information available points to that being true. You can't disprove it, because you don't own that version, but whatever. On the other hand you say "but those tables here, those are totally fine, although parts of them were added manually as well, but that data isn't code, so it's surely absolutely correct."

So I guess if clone2727 would have added a GUI-Option for enforcing hi-res, then that would be fine to you and you would trust it. But he added a tiny bit of code, so that's different.

m-kiewitz: Wait, actually there was already code in place to only read the GUI-option for King's Quest 6, Gabriel Knight 1 and Police Quest 4, when it's a CD version and DOS/windows.

But the current code in GfxFrameout reads it directly, so it's incorrect for all sorts of games.
The code should have used g_sci->forceHiresGraphics() instead. I also see the same direct read inside kPlatform. Will fix the latter right now.

In any case, I can also change that code to not use gameIDs anymore, but check the actual detection entry.

m-kiewitz: I just see that you changed some of that kPlatform code, when you implemented ef2c44bf1f2cf128a20096b3bcca2578634814b1. And your code basically breaks how King's Quest 6 worked.

King's Quest 6 was supposed to be hi-res when "force hi-res" is enabled OR platform is set to Windows. And now the flag is read directly, which is not only incorrect, but platform being Windows is fully ignored now.

Possibly other games may be affected too by those changes, because now all sorts of games get bad answers on those kPlatform calls.

So I guess I will not touch it (otherwise I break something in SCI32), but will instead ask you to please fix it.

engines/sci/graphics/cursor32.cpp
bluegr: "goingToPaint" then? "gonnaPaint" is slang, which we don't use

engines/sci/graphics/frameout.cpp
bluegr: I have GK1 Mac, and it is hires only. Here's a video, if it helps:
https://youtu.be/W49FgsXhE7Y

I am in vacations right now, and I can test in about a week or so

engines/sci/graphics/frameout.cpp
bluegr: It would make sense to keep a warning for this behavior, at least during initial game playthroughs, with a TODO to remove it. That way, we can at least check that this won't cause unexpected behavior and errors in other parts of the engine.

engines/sci/graphics/frameout.cpp
csnover: > I just see that you changed some of that kPlatform code, when you implemented ef2c44b. And your code basically breaks how King's Quest 6 worked.

Sorry about that. I looked at some other SCI1.1 games and they only had one operation (return platform ID), so I made an incorrect assumption that the other code was also some SCI32 guesswork (since SCI32 do have a few extra platform options). KQ6 is quite the unique snowflake, isn’t it… anyway, I am happy to fix this. Thanks for pointing it out!

engines/sci/graphics/frameout.cpp
csnover: > We should at the very least output a warning, so that errors are recognizable as such.

OK, I will add a warning here.

engines/sci/graphics/frameout.cpp
m-kiewitz: Actually the SCI1.1 DOS interpreter basically do almost nothing at all.
The windows 3.11 interpreters are different. KQ6 is a special snowflake because of kPortrait, but kPlatform is exactly the same as in other SCI1.1 interpreters for the DOS version.

engines/sci/graphics/cursor32.cpp
csnover: > Is it a problem to error() out in such case? Or at the very least skip over the code?

The view constructor has a warning already that could be turned into an error; it was a silent early return in SSCI.

engines/sci/graphics/frameout.cpp
csnover: > The windows 3.11 interpreters are different.

Both EcoQuest CD and SQ4 CD Windows kPlatform is identical to DOS kPlatform except it sets the accumulator to 2 instead of 1.

It’s trivia, I know, but I am curious which games other than KQ6 had a Windows interpreter with a different kPlatform function? The LSL6 lores CD version I have didn’t even have a Windows interpreter.

engines/sci/graphics/frameout.cpp
csnover: A warning has been added to kSetNowSeen32.

engines/sci/graphics/cursor32.cpp
csnover: Changed “early SSCI” to “early SCI32”.

engines/sci/graphics/cursor32.cpp
csnover: Warnings in CelObj constructors that rely on external data have been converted to errors.

engines/sci/graphics/cursor32.cpp
csnover: IIRC, it causes bouncing along the edges of the cursor restricted area because CursorManager uses unfiltered mouse events directly from the backend. Since this feature is not even 350 LOC, I’m not concerned if it turns out sometime in the future that this really needs to be changed to use CursorManager for some reason. (As far as performance goes, I suspect this code will always be faster than CursorManager anyway, since it is less generalised.)

csnover:
Landed 9bfeb3c297a7868c53d4c37775148067d72f44c0 through 7da359755d31b6a5a98fc791f795139441c82b56

PR #805, opened 2016-08-12
SCI32: Implement kRobot

csnover:
Not much else to say about this one. Makes a bunch of stuff work. Many many thanks to @wjp for helping figure out the craziest details of the audio format.

### DIFF
--- a/engines/sci/graphics/screen_item32.h
+++ b/engines/sci/graphics/screen_item32.h
@@ -31,6 +31,7 @@ namespace Sci {
 
 enum ScaleSignals32 {
 	kScaleSignalNone                    = 0,
+	// TODO: rename to 'manual'
 	kScaleSignalDoScaling32				= 1, // enables scaling when drawing that cel (involves scaleX and scaleY)
 	kScaleSignalUseVanishingPoint       = 2,
 	// TODO: Is this actually a thing? I have not seen it and
--- a/engines/sci/console.cpp
+++ b/engines/sci/console.cpp
@@ -54,7 +54,6 @@
 #include "sci/graphics/frameout.h"
 #include "sci/graphics/paint32.h"
 #include "video/coktel_decoder.h"
-#include "sci/video/robot_decoder.h"
 #endif
 
 #include "common/file.h"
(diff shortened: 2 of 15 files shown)


sev-:
Wow. Just wow.

csnover:
@sev-, yeah, that is a very familiar emotion, having spent like a month on this one feature. :)

engines/sci/graphics/screen_item32.h
```diff
@@ -31,6 +31,7 @@ namespace Sci {
 
 enum ScaleSignals32 {
 	kScaleSignalNone                    = 0,
+	// TODO: rename to 'manual'
```
bluegr: Does that refer to kScaleSignalNone?

bluegr:
Excellent work, for a very weirdly designed functionality :)

I had done progress on this myself, but then I read on IRC that you were working on this too, so, since my free time is very limited these days, I did not progress afterwards.

Overall, this looks good, however it's merged into one gigantic commit, which makes it quite hard to follow. Could you please split this up into smaller commits?

wjp:
Usually I'm all in favour of smaller commits, but since practically all of this commit is adding new, related functionality and I don't quite see which parts could be split up, one single commit makes sense to me here.

wjp:
Amazing work on this. It's gone through several debugging passes already, and I think this can be be merged as-is.

bluegr:
What is the status of V4 Robot videos? IIRC, these were only used in the SWAT demo... which used to work fine with the previous code. Does that still work? There are a lot of V4-related TODOs.

Furthermore, for completeness sake, can a comment be added mentioning which games use V4, V5 and V6 robot files?

I believe that it goes like this:
  - V4 robots: SWAT demo
  - V5 robots: KQ7, Phantasmagoria, SWAT
  - V6 robots: RAMA

IIRC, Lighthouse does not use Robot videos, right? Can the list above be double checked, and added as a comment?

Finally, I'd like a reply to my comment above (about kScaleSignalNone).

bluegr:
Sorry, Lighthouse does use robot videos. Thus, if I'm not mistaken, the list above is:

  - v4: SWAT demo
  - v5: KQ7, Phantasmagoria, SWAT
  - v6: RAMA, Lighthouse

csnover:
Landed at 0f2748b.

bluegr:
And what about my questions and request above?

PR #806, opened 2016-08-15
BUILD: Switch SDL backend to SDL2 by default. SDL1 is still a fallback.

sev-:
This is the switch of SDL backend to SDL2. All the work has been already prepared, and the use of this library version tested many times. Thus, this patch boils down to adding new references in our build system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 language:
   - cpp
 
-sudo: false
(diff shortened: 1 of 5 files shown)


tsoliman:
No compile errors on Maemo 
👍 

(no SDL2 installed nor will it likely ever be)

PR #808, opened 2016-08-16
WINTERMUTE: Add flag for games using low-spec resources

JenniBee:
I created three game for Wintermute that use 320x240 resolution, with 8-bit BMP art assets and WAV music, so they don't require the PNG, JPG, or VORBIS libraries.

The games are available here:
https://jennibee.itch.io/dreamcat
https://jennibee.itch.io/openquest-wintermute
https://jennibee.itch.io/dfaf-adventure

### DIFF
--- a/engines/wintermute/configure.engine
+++ b/engines/wintermute/configure.engine
@@ -1,3 +1,3 @@
 # This file is included from the main "configure" script
 # add_engine [name] [desc] [build-by-default] [subengines] [base games] [deps]
-add_engine wintermute "Wintermute" yes "" "" "jpeg png zlib vorbis 16bit highres"
+add_engine wintermute "Wintermute" yes "" "" "zlib 16bit highres"
--- a/engines/wintermute/base/sound/base_sound_buffer.cpp
+++ b/engines/wintermute/base/sound/base_sound_buffer.cpp
@@ -33,7 +33,9 @@
 #include "engines/wintermute/wintermute.h"
 #include "audio/audiostream.h"
 #include "audio/mixer.h"
+#ifdef USE_VORBIS
 #include "audio/decoders/vorbis.h"
+#endif
 #include "audio/decoders/wave.h"
 #include "audio/decoders/raw.h"
 #include "common/system.h"
@@ -102,9 +104,12 @@ bool BaseSoundBuffer::loadFromFile(const Common::String &filename, bool forceRel
 	}
 	Common::String strFilename(filename);
 	strFilename.toLowercase();
-	if (strFilename.hasSuffix(".ogg")) {
-		_stream = Audio::makeVorbisStream(_file, DisposeAfterUse::YES);
-	} else if (strFilename.hasSuffix(".wav")) {
+	#ifdef USE_VORBIS
+		if (strFilename.hasSuffix(".ogg")) {
+			_stream = Audio::makeVorbisStream(_file, DisposeAfterUse::YES);
+		}
+	#endif
+	if (strFilename.hasSuffix(".wav")) {
 		int waveSize, waveRate;
 		byte waveFlags;
 		uint16 waveType;
--- a/engines/wintermute/detection_tables.h
+++ b/engines/wintermute/detection_tables.h
@@ -43,7 +43,7 @@ static const PlainGameDescriptor wintermuteGames[] = {
 	{"conspiracao",     "Conspiracao Dumont"},
 	{"corrosion",       "Corrosion: Cold Winter Waiting"},
 	{"deadcity",        "Dead City"},
-	{"dfafadventure",        "DFAF Adventure"},
+	{"dfafadventure",   "DFAF Adventure"},
 	{"dreamcat",        "Dreamcat"},
 	{"dreaming",        "Des Reves Elastiques Avec Mille Insectes Nommes Georges"},
 	{"dirtysplit",      "Dirty Split"},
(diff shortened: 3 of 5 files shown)


engines/wintermute/configure.engine
```diff
@@ -1,3 +1,3 @@
 # This file is included from the main "configure" script
 # add_engine [name] [desc] [build-by-default] [subengines] [base games] [deps]
-add_engine wintermute "Wintermute" yes "" "" "jpeg png zlib vorbis 16bit highres"
```
somaen: This has some interesting side-effects that I'd have to handle before merging this. Namely that as far as I know, Dreamcat would be the _only_ game that works with ScummVM, if built without any of the above requirements.

somaen:
The 8-bit-ness doesn't really affect much beyond initial loading, the ScummVM implementation of the Wintermute engine is hard-coded to work in 32-bpp right now anyhow (and hence will spend the CPU resources/memory required to do so). This all boils down to how most of the games that were implemented for it were 24/32 bpp games anyhow.

If there actually is a need/want for 16/8-bpp modes, I _might_ look into it adding some optimized paths for it at some point, but in practice, you should get the same speed for 16/24/32 bpp rendering....

What _really_ slows things down, is using alpha beyond simple binary transparency (as allowed for by colour-keying).

JenniBee:
I know the Dreamcast has 32-bit color support (but I'm not sure if that's just when output through the VGA cable).  I do believe that it has 16-bit color support otherwise (at least I've seen projects that use this and don't mention they require a VGA cable),  But, I do believe that there are platforms that ScummVM supports that only run 320x240 resolutions at 8-bpp, so it would be helpful to have these lower resolutions supported (Wintermute is quite versatile when it comes to supporting resolutions, and I have seen quite a few posts on their forums that mentioned people wanting 320x240 8-bit support (but switching to AGS as there weren't any projects that used it at the time).  I personally plan on making at least a few more 320x240 bpp games in Wintermute at least.  

For now though, I suppose Dreamcast can still be enabled since it should still support Dreamcat in ScummVM's Wintermute engine when run through 32-bpp, at least on VGA.  Some of the other dependencies would still need to be checked off though, as I don't believe the Dreamcast port supports them (such as OGG).  

Perhaps for the dependencies issue, a "wintermute lowres" flag or something of the sort would be the route to go, similar to the "scumm_7_8 he" flag in SCUMM?

sev-:
Really interesting experiment!

If you think about removing PNG requirement, in this case, a code which shows a warning and does not try to load most of the games should be run (could be cleanly achieved via Game Features). Same goes for Jpeg/Vorbis/Zlib and Highres.

I am curious, does it even compile and run without these libraries?

Instead of marking all games as requiring these, I would add a feature GF_LOWRES_ASSETS or something similar.

If you want to have the detection entry added earlier, without modifying the engine settings, this could be done anytime.

sev-:
Why did you close it? Were you disappointed by something?

JenniBee:
I rebased, and it auto-closed (since there was only one commit).  I finally was able to run this.  It compiled cleanly, but when it was run there were problems without highres or 16bit (as somaen said - it needs 32-bit currently - and it gives you a warning and exits when you try to run it at a lower resolution) and zlib (as the package files made by the wintermute engine are compressed).  It doesn't need png. jpg, or vorbis though (which prevents it from being run on Dreamcast).  I'm going to rebase this commit to only disable those dependencies and reopen this commit.

JenniBee:
I added a GF_LOWRES_ASSETS flag, and set it up to trigger a GUI message if you try to run a  game that needs PNG, JPEG, or Vorbis if those aren't available.  I also set it to initialize at 320x240 if the GF_LOWRES_ASSETS flag was set, so that it doesn't try to initialize at a resolution higher than is available to a lower spec system.

JenniBee:
OK. I rebased with the formatting fixes, and I tested it with a build with jpeg disabled, and one with png, jpeg, and vorbis enabled, and everything appears to work as it should.

JenniBee:
I rebased this commit to reflect the additions to the main codebase by raziel in #879 

Thanks for that, by the way.  I was planning on making a separate commit for the games themselves, as sev suggested above, but now I don't have to. :+1:

JenniBee:
Since it's been a while since this pull request was opened, I'm wondering. Does anything have to be done on my end before this can be merged?

engines/wintermute/configure.engine
tobiatesan: I don't understand the point of this change, what am I missing? Thank you

csnover:
From reviewing the history it appears that there was still an unresolved concern regarding the changes to `configure.engine`, and also some unresolved discussion regarding the bit depth that the engine processes in? As the perpetual wet blanket, I would also raise an extra question regarding the time needed to properly address these outstanding concerns and maintaining this extra code to support one game (plus maybe a couple more in the future?) on a very old platform. But as I do not work on Wintermute that last question is best answered by the people that will be supporting and maintaining it.

JenniBee:
I had to rebase for conflicts, so I used that to address the lingering concerns. This is now not focused on the Dreamcast port, but on the games itself.

I changed the configure.engine to simply not require PNG, JPG, or VORBIS (since those aren't required for Wintermute games).  It no longer touches the highres or 16bit flags, since the engine is coded for 32bpp.

I kept the GF_LOWRES flag, since the three games that use it (OpenQuest, DFAF Adventure, and Dreamcat) are 320x240, so this makes sure it initiates the engine at the correct resolution.

That should clear everything up so this can hopefully finally be merged.

bonki:
This has been lingering for a while and I wonder what exactly needs to be done on our end so we can give @JenniBee some closure?

JenniBee:
I rebased and changed the flag from GF_LOWRES_ASSETS to GF_LOWSPEC_ASSETS, to better reflect the pull request as it is now, in that it doesn't require PNG, JPG, and VORBIS for games that don't use them.

This should hopefully allow it to finally be merged to master. If anything still needs to be changed on my end, please let me know and I'll be happy to oblige.

bonki:
Thanks for not giving up on this :)

To me it looks OK now, does anybody still object to merging?

bonki:
@JenniBee Right now it won't compile, can you fix?

JenniBee:
I think I see what might be causing it (the lack of end line breaks). I've rebased again. Hopefully, it'll be good to go once again.

JenniBee:
D'oh. I see where the problem is. When I rebased, I forgot to initialize GF_LOWSPEC_ASSETS

Now it should pass the test.

JenniBee:
I tested this one locally and it built fine, so it should successfully build now (fingers crossed)

PR #810, opened 2016-08-21
SDL: Fix build with MacPorts SDL2

tsoliman:
SDL2 (2.0.4) that comes with MacPorts uses MacPorts' libiconv which ends up depending on `time.h` forbidden symbols

Here is the include-chain to time.h (pre-patch)
```
../backends/platform/sdl/sdl-sys.h:100
/opt/local/include/SDL2/SDL.h:32
/opt/local/include/SDL2/SDL_main.h:25
/opt/local/include/SDL2/SDL_stdinc.h:87
/opt/local/include/iconv.h:111
/usr/include/wchar.h:91
/usr/include/time.h
```

Homebrew doesn't have this problem because it uses the native (XCode) libiconv

### DIFF
--- a/backends/platform/sdl/sdl-sys.h
+++ b/backends/platform/sdl/sdl-sys.h
@@ -67,6 +67,53 @@ typedef struct { int FAKE; } FAKE_FILE;
 #define system FAKE_system
 #endif
 
+// Fix compilation with MacPorts SDL 2
+// It needs various (usually forbidden) symbols from time.h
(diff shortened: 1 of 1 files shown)


criezy:
I don't use MacPorts and thus do not have this issue, but the changes proposed look fine to me.

I will let a few more days for others to complain and if I don't hear back by then I will merge it this weekend.

sev-:
No objections, looks fine to me.

PR #811, opened 2016-08-21
BACKENDS: timidity: Get rid of gethostbyname, use getaddrinfo.

susnux:
As the title says this will port the timidity code to use `getaddrinfo` instead of `gethostbyname`.
Not even that `gethostbyname` is deprecated, but also it will limit the address resolution to one possible ip (but maybe that one is not working?) and will limit the name resolution to IPv4.
That is fixed by this commit, moreover on at least openSUSE the buildsystem warns if code is using `gethostbyname`.
Also some other minor cleanup (fixed a FIXME etc).

### DIFF
--- a/backends/midi/timidity.cpp
+++ b/backends/midi/timidity.cpp
@@ -52,23 +52,17 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/param.h>
-#include <netdb.h>		/* for gethostbyname */
+#include <netdb.h>  /* for getaddrinfo */
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <stdarg.h>
 #include <stdlib.h>
 #include <errno.h>
 
-// WORKAROUND bug #1870304: Solaris does not provide INADDR_NONE.
-#ifndef INADDR_NONE
-#define INADDR_NONE 0xffffffff
-#endif
-
 // BeOS BONE uses snooze (x/1000) in place of usleep(x)
 #ifdef __BEOS__
 #define usleep(v) snooze(v/1000)
 #endif
-
 
 #define SEQ_MIDIPUTC 5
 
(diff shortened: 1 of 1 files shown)


backends/midi/timidity.cpp
```diff
@@ -215,9 +205,15 @@ int MidiDriver_TIMIDITY::open() {
 	/*
 	 * open data connection
 	 */
-	data_port = atoi(res + 4);
+	num = atoi(res+4);
```
bluegr: Formatting (spaces around "+")

bluegr:
Looks good overall, thanks! :)

susnux:
Ok fixed that

backends/midi/timidity.cpp
```diff
@@ -277,49 +273,36 @@ void MidiDriver_TIMIDITY::teardown() {
...
+	for (rp = result; rp != NULL; rp = rp->ai_next) {
+		if ((fd = socket(rp->ai_family, rp->ai_socktype, rp->ai_protocol)) == -1)
+			continue;
+		if (connect(fd, rp->ai_addr, rp->ai_addrlen) != -1)
+			break;
 		::close(fd);
+	}
+
+	if (rp == NULL) {
+		warning("TiMidity: Could not connect\n");
```
mgerhardy: Isn't there a small memleak here - freeaddrinfo(result) call is missing.

susnux:
@mgerhardy Yes in some situation that could maybe leak memory.
Fixed it by moving the free above the check.

fuzzie:
backends/midi/timidity.cpp:225:90: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘char*’ [-Wformat=]

Another format string needs updating?

susnux:
@fuzzie  You were right, fixed it.

sev-:
Thanks, merging.

PR #812, opened 2016-08-23
SCI32: Torin's Passage maze functionality

bluegr:
This pull request is a WIP implementation of kBitmapCreateFromView and kBitmapGetColor

These are used in Torin room 40300 (lava maze). They are used to
create a path mask, used to indicate walkable areas. The current
implementation is still not right, as the walkable areas returned
are wrong, and Torin can't move.

To test, change the current room to 40300 and try to move
around. The end result should be something like this:

https://www.youtube.com/watch?v=BIDT0A_6cdQ#t=1h39m57s

**Edit:** Forgot to mention: in this screen, game scripts check kBitmapGetColor for walkable areas, providing 3 parameters: a reference to the bitmap created in kBitmapCreateFromView, and the x/y coordinates to check. If the function returns 0, the requested point is walkable, otherwise it is barred.


csnover:
Addressed by PR #835.

PR #813, opened 2016-08-24
SCI32: Fix dynamic line creation

bluegr:
This makes the crystal puzzle in Torin (room 43000) work correctly.
The results have been verified with LSL6 as well

- Fix the screen item rect to be the one returned by makeLineBitmap.
  This fixes line placement
- Improve sanity check in the line plotter to work with coordinates
  instead of bitmap buffer size
- Fix off-by-one error in the line screenitem rectangle
- Don't create a draw rectangle from the line coordinates, as these
  may not form a valid rectangle

### DIFF
--- a/engines/sci/graphics/paint32.cpp
+++ b/engines/sci/graphics/paint32.cpp
@@ -38,7 +38,6 @@ reg_t GfxPaint32::kernelAddLine(const reg_t planeObject, const Common::Point &st
 
 	Common::Rect gameRect;
 	reg_t bitmapId = makeLineBitmap(startPoint, endPoint, priority, color, style, pattern, thickness, gameRect);
-	SciBitmap &bitmap = *_segMan->lookupBitmap(bitmapId);
 
 	CelInfo32 celInfo;
 	celInfo.type = kCelTypeMem;
@@ -48,7 +47,7 @@ reg_t GfxPaint32::kernelAddLine(const reg_t planeObject, const Common::Point &st
 	// `kUpdateLine` can get the originally used color
 	celInfo.color = color;
 
-	ScreenItem *screenItem = new ScreenItem(planeObject, celInfo, Common::Rect(startPoint.x, startPoint.y, startPoint.x + bitmap.getWidth(), startPoint.y + bitmap.getHeight()));
+	ScreenItem *screenItem = new ScreenItem(planeObject, celInfo, gameRect);
 	screenItem->_priority = priority;
 	screenItem->_fixedPriority = true;
 
@@ -89,9 +88,13 @@ void GfxPaint32::plotter(int x, int y, int color, void *data) {
 	LineProperties &properties = *static_cast<LineProperties *>(data);
 	byte *pixels = properties.bitmap->getPixels();
 
-	const uint32 index = properties.bitmap->getWidth() * y + x;
-
-	if (index < properties.bitmap->getDataSize()) {
+	const uint16 bitmapWidth  = properties.bitmap->getWidth();
+	const uint16 bitmapHeight = properties.bitmap->getHeight();
+	const uint32 index = bitmapWidth * y + x;
+	
+	// Only draw the points in the bitmap, and ignore the rest. SSCI scripts
+	// can draw lines ending outside the visible area (e.g. negative coordinates)
+	if (x >= 0 && x < bitmapWidth && y >= 0 && y < bitmapHeight) {
 		if (properties.solid) {
 			pixels[index] = (uint8)color;
 			return;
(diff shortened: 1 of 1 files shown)


csnover:
The game crashes with this patch applied with this crystal configuration:

screen shot 2016-08-24 at 09 57 03

The stack trace shows an attempted write to y=-1.

engines/sci/graphics/paint32.cpp
```diff
@@ -113,24 +112,26 @@ void GfxPaint32::plotter(int x, int y, int color, void *data) {
...
-	outRect.bottom = (startPoint.y > endPoint.y ? startPoint.y : endPoint.y) + halfThickness + 1;
+	outRect.left   = MIN<int16>(startPoint.x, endPoint.x) - halfThickness;
+	outRect.top    = MIN<int16>(startPoint.y, endPoint.y) - halfThickness;
+	outRect.right  = MAX<int16>(startPoint.x, endPoint.x) + (halfThickness + 1) + 1;	// round up halfThickness
+	outRect.bottom = MAX<int16>(startPoint.y, endPoint.y) + (halfThickness + 1) + 1;	// round up halfThickness
 
 	reg_t bitmapId;
-	SciBitmap &bitmap = *_segMan->allocateBitmap(&bitmapId, outRect.width(), outRect.height(), skipColor, 0, 0, g_sci->_gfxFrameout->getCurrentBuffer().scriptWidth, g_sci->_gfxFrameout->getCurrentBuffer().scriptHeight, 0, false, true);
+	uint16 scaledWidth = g_sci->_gfxFrameout->getCurrentBuffer().scriptWidth;
+	uint16 scaledHeight = g_sci->_gfxFrameout->getCurrentBuffer().scriptHeight;
```
csnover: Mismatched variable names.

bluegr: Are you referring to spacing?

bluegr:
Interesting, thanks for finding that! Will test later on today

engines/sci/graphics/paint32.cpp
csnover: No, I am referring to the mismatched variable names “scaled” and “script”.

engines/sci/graphics/paint32.cpp
```diff
@@ -152,12 +153,11 @@ reg_t GfxPaint32::makeLineBitmap(const Common::Point &startPoint, const Common::
...
-		startPoint.x - outRect.left,
-		startPoint.y - outRect.top,
-		endPoint.x - outRect.left,
-		endPoint.y - outRect.top
-	);
+	// Change coordinates to be relative to the bitmap
+	int16 x1 = startPoint.x - outRect.left;
+	int16 y1 = startPoint.y - outRect.top;
+	int16 x2 = endPoint.x - outRect.left;
+	int16 y2 = endPoint.y - outRect.top;
```
csnover: Const correctness throughout would be appreciated.

engines/sci/graphics/paint32.cpp
```diff
@@ -113,24 +112,26 @@ void GfxPaint32::plotter(int x, int y, int color, void *data) {
 			properties.patternIndex = 0;
 		}
 	} else {
-		warning("GfxPaint32::plotter: Attempted to write out of bounds (%u >= %u)", index, properties.bitmap->getDataSize());
+		warning("GfxPaint32::plotter: Attempted to write out of bounds (%u, %u) - bounds are (%u, %u)", x, y, properties.bitmap->getWidth(), properties.bitmap->getHeight());
```
csnover: This warns a lot; I wonder if there is another defect somewhere else causing overdraw.

bluegr:
All the issues mentioned have been addressed now. I've checked the behavior against the original, and the results are the same. Check a reference video here:

https://www.youtube.com/watch?v=BIDT0A_6cdQ#t=1h46m17s

csnover:
The wide line is not drawing at the correct position when slope = 0 or ∞.

untitled-2

bluegr:
Thanks for the feedback, the issue was with the way our thick line implementation differs from Sierra's. The issue in the picture has been fixed now

bluegr:
Update: Verified with @sev that the difference in our thick line implementation was unintentional, so the fixes for thick line drawing in this pull request have been pushed to common code

csnover:
There are still some problems with the line drawing, most noticeably where slope = 1 (lower-right).

untitled-3

bluegr:
Looks like a similar bug to the one above, but this one will be harder to fix. Will check it out

bluegr:
The problem above seems to be an issue with thick line drawing, where an extra padding of 1 pixel is added in this case.

We are using the same thick line algorithm as SSCI, Bresenham by Foley & Van Dam, and have taken our thick line drawing algorithm from gd (function gdImageLine):
https://github.com/libgd/libgd/blob/master/src/gd.c

Line thickness in this case is 3 for thick lines, and 1 for normal lines.

This will need more investigation...

csnover:
Addressed by PR #835.

bluegr:
The commit that has been added in PR #835 is an earlier version of the final commit in this PR. I'll update master with the changes that I made.

PR #814, opened 2016-08-25
SCI32: Implement dynamic video rectangles (SetHotRectangles)

bluegr:
This is an very rare functionality of the SCI engine, used during the
final chase scene in Phantasmagoria. Essentially, hot rectangles are
given in a list by the scripts while a VMD video is playing, and are
used to stop the video when the mouse enters or leaves a rectangle.
SSCI generates specific hot rectangle events, which are then picked up
by the VMD player, which also include the index of the hot rectangle
that fired the event, however this was never used.

Since this is only checked in one place (the VMD player), in this
implementation we are just storing a flag in the cursor handler when
a hot rectangle event occurs, which is then checked by the VMD player,
instead of injecting custom events


csnover:
Addressed by PR #835.

PR #815, opened 2016-08-25
SCI: Make kFileIORename use wrapped filenames

wjp:
This fixes deleting savegames in Phantasmagoria 1.

I'm not sure if there's SCI11 usage of kFileIORename that this would break, although it seems unlikely. I'm not aware of any yet, though, so can't test.

### DIFF
--- a/engines/sci/engine/kfile.cpp
+++ b/engines/sci/engine/kfile.cpp
@@ -632,6 +632,15 @@ reg_t kFileIORename(EngineState *s, int argc, reg_t *argv) {
 	Common::String oldName = s->_segMan->getString(argv[0]);
 	Common::String newName = s->_segMan->getString(argv[1]);
 
+	// We don't fully implement all cases that could occur here, and
+	// assume the file to be renamed is a wrapped filename.
+	// Known usage: In Phant1 and KQ7 while deleting savegames.
+	// The scripts rewrite the dir file as a temporary file, and then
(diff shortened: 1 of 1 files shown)


bluegr:
For reference: when Phantasmagoria deletes a save file, it writes the rest of the saved games in a new temporary savegame index file, xyz.tmp. It then renames the temporary index file to phantsg.dir, overwriting the existing one (yikes!)

Without this change, the game will create the temporary index file (with a wrapped name), but then it will try to rename the unwrapped name, which will fail, thus it will destroy the savegame index.

We should not even support renaming actual game files, only generated files, so this change makes even more sense :)

Merging, as-is, thanks for your work, once again! :)

wjp:
No, that's not correct. It first deletes phantsg.dir, and then does the rename of xyz.tmp to phantsg.dir afterwards.

wjp:
Also, please don't merge pull requests that are clearly still open for discussion unilaterally like this.

bluegr:
Sorry, it was my understanding that this would be good to merge because:
- I've tested it and it works
- It was a very small, non-invasive change to one kernel function
- It ensures that only saved games are renamed, not game files
- It is something that is used almost exclusively in SCI32 games, which are WIP
- It prevents Phantasmagoria from destroying its save index

All in all, this seemed to be quite a positive, correct and small change.

PR #816, opened 2016-08-26
GRAPHICS: [GSoC] Fix leaks and bugs in MacGui

blorente:
Since the addition of MacGui to mainline, a number of leaks and minor bugs in the code have been found, via Valgrind. They were mostly due to `free()` not being called when appropriate, which in a particular case created a 10MB leak every time the GUI was drawn. Regarding the bugs, there was only one found, which creashed the game when closing a window, due to a leftover index.

This PR is part of the cleanup of my GSoC project.


sev-:
Nice, thanks a lot!

PR #818, opened 2016-08-27
XEEN: Might & Magic engine

dreammaster:
This is the pull request for my in-progress engine for the earlier Might & Magic games. At the moment, I'm focusing on World of Xeen, which is the combination of Might & Magic 4 & 5. It will eventually also be able to play the games individually, as well as hopefully Might and Magic 3 Isles of Terra, and the fan-produced Swords of Xeen. I call the engine Xeen in the absence of any better name.

At the moment, the in-game functionality in a reasonably good state. You can walk around, go to different areas, and even fight monsters (though combat does still need some work). Much of the character dialogs are done, and likewise the in-town functionality, though there are still some aspects still to do.

One major reason I'm creating a pull request for this now is not only because it seems fitting with the Dungeon Master engine shortly to have it's own pull request, but also because by the time it's accepted and hits master, I'll likely be finished writing all the game logic classes for Titanic.. and starting the daunting task of slowly debugging all the code I wrote.  Having something to regularly switch to will likely help stop me going bonkers, and it'll more convenient to have it in master.

Moving forward, I'll be initially focusing more on implementing the cutscenes (intro & ending) for both Clouds and Dark Side, and then the startup dialogs for creating characters. I'll then return to polishing up the in-game functionality later on.


engines/xeen/character.cpp
```diff
@@ -0,0 +1,1840 @@
...
+	_weapons.clear();
+	_armor.clear();
+	_accessories.clear();
+	_misc.clear();
+}
+
+void Character::synchronize(Common::Serializer &s) {
+	char name[16];
+	Common::fill(&name[0], &name[16], '\0');
+	strncpy(name, _name.c_str(), 16);
```
sev-: This may lead to a buffer overflow, as strncpy doesn't take into account 0. It is safer to use Common::strlcpy which does take it into account.

sev-:
I looked through the code, excellent work, as usual.

Nothing needs to be done before the merge.

sev-:
Ah, there is a name clash warning in xsurface.cpp. Might be worth fixing before the merge.

Also there are 2 writes past the array size, see the Travis clang log. Those will lead to crashes.

PR #819, opened 2016-08-28
SCI32: kNumCels

csnover:
Every version of SCI32 has a logic error in `kNumCels` (comparison `>` instead of `>=`) that causes games to read garbage data if they ask for a cel count for the one-past-the-end loop. Unfortunately, GK1 does exactly this, and it is plausible that other games do too since the bug in SSCI was never fixed.

This PR does two things:

1. Generalises the code used to get information on the current callee, so that information about buggy scripts can be more easily displayed in errors;
2. Raises an error whenever `kNumCels` is detected to be used in the buggy way, so game scripts relying on invalid reads can be fixed

(Please feel free to bikeshed on more appropriate names for the new EngineState method!)

### DIFF
--- a/engines/sci/graphics/celobj32.cpp
+++ b/engines/sci/graphics/celobj32.cpp
@@ -812,7 +812,21 @@ int16 CelObjView::getNumCels(const GuiResourceId viewId, const int16 loopNo) {
 	const byte *const data = resource->data;
 
 	const uint16 loopCount = data[2];
-	if (loopNo >= loopCount || loopNo < 0) {
+
+	// Every version of SCI32 has a logic error in this function that causes
+	// random memory to be read if a script requests the cel count for one
+	// past the maximum loop index. At least GK1 room 800 does this, and gets
+	// stuck in an infinite loop because the game script expects this method
+	// to return a non-zero value.
+	// The scope of this bug means it is likely to pop up in other games, so we
+	// explicitly trap the bad condition here and report it so that any other
+	// game scripts relying on this broken behavior can be fixed as well
+	if (loopNo == loopCount) {
+		const SciCallOrigin origin = g_sci->getEngineState()->getCurrentCallOrigin();
+		error("[CelObjView::getNumCels]: loop number is equal to loop count in method %s::%s (room %d, script %d, localCall %x)", origin.objectName.c_str(), origin.methodName.c_str(), origin.roomNr, origin.scriptNr, origin.localCallOffset);
+	}
+
+	if (loopNo > loopCount || loopNo < 0) {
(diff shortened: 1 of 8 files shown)


bluegr:
Looks good, +1

engines/sci/graphics/celobj32.cpp
```diff
@@ -812,7 +812,21 @@ int16 CelObjView::getNumCels(const GuiResourceId viewId, const int16 loopNo) {
...
+	// random memory to be read if a script requests the cel count for one
+	// past the maximum loop index. At least GK1 room 800 does this, and gets
+	// stuck in an infinite loop because the game script expects this method
+	// to return a non-zero value.
+	// The scope of this bug means it is likely to pop up in other games, so we
+	// explicitly trap the bad condition here and report it so that any other
+	// game scripts relying on this broken behavior can be fixed as well
+	if (loopNo == loopCount) {
+		const SciCallOrigin origin = g_sci->getEngineState()->getCurrentCallOrigin();
+		error("[CelObjView::getNumCels]: loop number is equal to loop count in method %s::%s (room %d, script %d, localCall %x)", origin.objectName.c_str(), origin.methodName.c_str(), origin.roomNr, origin.scriptNr, origin.localCallOffset);
```
bluegr: Question: The way this is written, it never allows for any workarounds to be set... is this intentional?

bluegr:
Instead of this solution, it would be much more preferable to also use a workaround table for this specific bug.  The way the code in this pull request is written, it makes the engine error out without attempting to fix the bug. For example, you can check vm_types.cpp:68, method lookForWorkaround(), or vm.cpp:120, method read_var() or vm.cpp:352, method callKernelFunc() or paint16.cpp:476, method kernelDisplay()

wjp:
Why would you already add a workaround table if we don't even know if there are workarounds to be added? That makes no sense.

bluegr:
The comment in the code states that "At least GK1 room 800 does this, and gets stuck in an infinite loop because the game script expects this method to return a non-zero value".

csnover:
Addressed by PR #835.

bluegr:
I'm sorry, but this has not been addressed at all, so I don't believe this should be closed. The PR that is linked merely contains the commits from this PR, without actually addressing the flaw in the logic: the current code errors out when this situation occurs, instead of providing a workaround, like all other places. I'm reopening this until this situation is resolved.

wjp:
No, there is workaround table now. 2035e21667b746c346e9f95658109904a86597d3.

bluegr:
Ah, right, I missed that one, my apologies!

PR #820, opened 2016-08-29
COMMON: Add Atari 8-bit platform

waltervn:
This adds a platform for the [Atari 8-bit family](https://en.wikipedia.org/wiki/Atari_8-bit_family). As 'atari' is already used for the Atari ST platform, I set all the codes to 'atari8'.


bluegr:
Perhaps it would also make sense to rename the current 'atari' platform to 'atarist' or 'atari-st', to avoid confusion?

sev-:
We already accept atari-st there. If we rename default option, that would require implementing autoupdating the scummvm.ini which is not trivial.

I'm merging this as is, thanks.

PR #822, opened 2016-09-02
CREATE_PROJECT: Fix networking backend file targets in Xcode

csnover:
I think this is right, but since I don’t own any of the networking code I just wanted to double check that this is the correct and expected build setting for iOS.


sev-:
Yep, thanks.

PR #824, opened 2016-09-05
OSYSTEM: Allow drawing a background activity icon on the OSD

bgK:
The recent cloud feature set added the following methods to the OSystem interface:
```c++
void copyRectToOSD(const void *buf, int pitch, int x, int y, int w, int h);
void clearOSD();
Graphics::PixelFormat getOSDFormat();
```

These methods allow to draw pixel data that is expected to overlay the regular
screen, with alpha transparency.

These methods are, as of now,  only used to display an icon indicating that ScummVM
is synchronizing in the background with some remote data storage.

There are IMO several issues:
* The backends are expected to implement the OSD as a full screen surface.
  It is not the case. For example the Android backend displays the OSD messages
  with native pop-up messages. The SDL backend did have a full screen OSD surface,
  but it was an implementation detail that I don't think should be exposed.
* It is not clear when the OSD surface should be drawn. The initial implementation
  drew it all the time resulting in a performance issue.
* The "Surface SDL" backend uses 16bits surfaces, which don't have a large enough
  alpha channel for proper alpha transparency.

This pull request removes the ability to directly draw to the OSD surface and
replaces it with the possibility to draw an icon indicating background activity:
```c++
void displayActivityIconOnOSD(const Graphics::Surface *icon)
```

The backend may display the icon as it sees fit.

~~I'm submitting this as a RFC to review the idea and proposed interface. Please don't merge, and please don't review in depth the implementation.~~

~~If the review is positive, I will polish the implementation, provide an implementation for the OpenGL backend, and do my best to fix breakage in the ports.~~

### DIFF
--- a/backends/graphics/opengl/opengl-graphics.cpp
+++ b/backends/graphics/opengl/opengl-graphics.cpp
@@ -57,7 +57,8 @@ OpenGLGraphicsManager::OpenGLGraphicsManager()
       _cursorKeyColor(0), _cursorVisible(false), _cursorDontScale(false), _cursorPaletteEnabled(false),
       _forceRedraw(false), _scissorOverride(3)
 #ifdef USE_OSD
-      , _osdAlpha(0), _osdFadeStartTime(0), _osd(nullptr)
+      , _osdMessageChangeRequest(false), _osdMessageAlpha(0), _osdMessageFadeStartTime(0), _osdMessageSurface(nullptr),
+      _osdIconChangeRequest(false), _osdIconSurface(nullptr)
 #endif
     {
 	memset(_gamePalette, 0, sizeof(_gamePalette));
@@ -69,7 +70,9 @@ OpenGLGraphicsManager::~OpenGLGraphicsManager() {
 	delete _overlay;
 	delete _cursor;
 #ifdef USE_OSD
-	delete _osd;
+	delete _osdMessageSurface;
+	delete _osdIconSurface;
+	_osdIconNextData.free();
 #endif
 #if !USE_FORCED_GLES
 	ShaderManager::destroy();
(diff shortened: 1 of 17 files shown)


criezy:
In principle this look like a good change to me.

Having the ability to directly draw to the OSD surface means we can display anything on the OSD, but I don't think we need that flexibility. And indeed that imposed a restriction on how the OSD is implemented in backends.

sev-:
Yes, we had some conceptual troubles with implementing it. I like your idea, it will have much less performance hit, still achieving the required results.

Initially we thought about adding some messages next to the icon, but in the end the idea was not implemented, which makes this very PR possible.

I'd say, go ahead, and implement it. Although if possible, put as much code as possible to the shared SDL codebase, so you do not end up copying it over for other ports.

Also I would suggest to mark it as another backend feature (see Feature enum in system.h), and currently also tie it to the USE_CLOUD define. And I like the way you implemented, that there is no performance hit when icon is not visible.

sev-:
@bgK Would it be possible to get this finished for 1.9.0?

bgK:
The OpenGL implementation turned out to be more complex than expected because the draw call happens from another thread. I have a draft implementation that needs finishing and clean up. It should be ready by the middle of next week.

bgK:
I've pushed a complete version of the changes. Please review.
This is ATM a huge commit because development was erratic on my part. I can try to split it if it too difficult to review.

backends/graphics/opengl/opengl-graphics.cpp
```diff
@@ -707,81 +736,126 @@ void OpenGLGraphicsManager::displayMessageOnOSD(const char *msg) {
...
+	_osdMessageChangeRequest = false;
 }
+#endif
 
-void OpenGLGraphicsManager::clearOSD() {
+void OpenGLGraphicsManager::displayActivityIconOnOSD(const Graphics::Surface *icon) {
 #ifdef USE_OSD
 	// HACK: Actually no client code should use graphics functions from
-	// another thread. But the MT-32 emulator still does, thus we need to
+	// another thread. But the network synchronization still does, thus we need to
```
sev-: actually it appears now that both MT-32 and network are applicable

sev-:
I started to look into it. One thing to mention is that it must be split on per-backend basis.

sev-:
But in general, I like the way it is done. Once you split it, it could be merged and we may start the 1.9.0 release cycle.

bgK:
Ok, I've split the commit.

sev-:
Yay, thanks! Merging.

PR #825, opened 2016-09-07
AUDIO: Fix CMS chips incorrect pitch.

jepael:
The CMS emulation assumes the chips run at 8 MHz clock,
but in PCs they run at 7.15909 MHz, so the emulated pitch
is too high. Adjusting the requested sampling rate higher
by matching amount the pitch is lowered down to normal.


sev-:
Thanks, merging.

PR #826, opened 2016-09-10
IMAGE: Indeo 4 & 5 decoders

dreammaster:
This implements an Indeo 4 decoder, which is needed for some of the videos in Starship Titanic. The code is taken from ffmpeg. Indeo 4 and 5 share a lot of code in common, and so I've kept that in my ScummVM version as well, in case we need to implement support for Indeo 5 as well in the future. Another thing to note is that there was too much code for me to be comfortable putting everything in a single file, so I created a new image/codecs/indeo/ folder which contains the extra classes that are needed, including the common Indeo decoding base class.


DrMcCoy:
> added GetBits class for reading

Wouldn't it be better to use Common::Bitstream for that? That's kinda what it was supposed to be, a replacement for ffmpeg's macro mess when I adapted their Bink code.

DrMcCoy:
Also, are their commonalities between Indeo3 and Indeo4/Indeo5 that could be leveraged here? Or are these codecs completely different from each other?

DrMcCoy:
Hmm, the code itself is also still very ffmpeg-like, with its av_malloc, snake_case method names and things like that.

On the one hand, properly helps keeping up with fixes from ffmpeg (if there will be any). On the other hand, it muddies the ScummVM style.

dreammaster:
Thanks for the prompt review. You're right regarding the differences in coding style.. I was a bit hesitant to go with large scale changes to the code, for partly that, and also to ease adding in Indeo5 later on if necessary. I wasn't aware of Common::BitStream.. wonder what else is tucked away in Common that I'm not aware of. I'll at least have a look at replacing my converted GetBits class with BitStream.

dreammaster:
Oh, and no, the differences to Indeo3 are signifcant, so the existing Indeo3 decoder won't benefit from any of this.

dreammaster:
Actually, on review, it'll be better to hold off on any simplification for GetBits until after I take a look at adding in Indeo5 support for Sev, for The Full Pipe.

dreammaster:
Thankfully, implementing the Indeo 5 decoder proved as straightforward as I'd hoped. The Full Pipe intro AVI verified to play fine.

dreammaster:
Okay, I've cleaned up everything I could find to match the coding conventions, and changed GetBits to derive from Common::BitStream and removed all redundant code from it. I think it's ready for merging.

image/codecs/indeo/get_bits.cpp
```diff
@@ -0,0 +1,60 @@
...
+namespace Indeo {
+
+int GetBits::getVLC2(int16 (*table)[2], int bits, int maxDepth) {
+	int code;
+    int n, nbBits;
+    unsigned int index;
+
+	index = peekBits(bits);
+    code  = table[index][0];
+    n     = table[index][1];
```
csnover: some spaces instead of tabs here

image/codecs/indeo/mem.cpp
```diff
@@ -0,0 +1,183 @@
...
+}
+
+void *avMallocZArray(size_t nmemb, size_t size) {
+	if (!size || nmemb >= MAX_INTEGER / size)
+		return NULL;
+
+	return avMallocZ(nmemb * size);
+}
+
+void avFree(void *ptr) {
```
sev-: I think this is redundant.

image/codecs/indeo/mem.cpp
```diff
@@ -0,0 +1,183 @@
...
+	// are both less than sqrt(SIZE_MAX).
+	if ((a | b) >= ((size_t)1 << (sizeof(size_t) * 4)) && a && t / a != b)
+		return -1;
+	*r = t;
+	return 0;
+}
+
+/*------------------------------------------------------------------------*/
+
+void *avMalloc(size_t size) {
```
sev-: This is also redundant

image/codecs/indeo/indeo.cpp
```diff
@@ -0,0 +1,1730 @@
...
+
+	_ctx._gb->align();
+
+	return result;
+}
+
+void IndeoDecoderBase::recomposeHaar(const IVIPlaneDesc *_plane,
+		uint8 *dst, const int dstPitch) {
+	int           x, y, indx, b0, b1, b2, b3, p0, p1, p2, p3;
+	const short * b0Ptr, *b1Ptr, *b2Ptr, *b3Ptr;
```
sev-: Formatting

image/codecs/indeo/indeo.cpp
```diff
@@ -0,0 +1,1730 @@
...
+}
+
+void IndeoDecoderBase::recompose53(const IVIPlaneDesc *_plane,
+		uint8 *dst, const int dstPitch) {
+	int           x, y, indx;
+	int32         p0, p1, p2, p3, tmp0, tmp1, tmp2;
+	int32         b0_1, b0_2, b1_1, b1_2, b1_3, b2_1, b2_2, b2_3, b2_4, b2_5, b2_6;
+	int32         b3_1, b3_2, b3_3, b3_4, b3_5, b3_6, b3_7, b3_8, b3_9;
+	int32         _pitch, back_pitch;
+	const short * b0Ptr, *b1Ptr, *b2Ptr, *b3Ptr;
```
sev-: Formatting

image/codecs/indeo/indeo.cpp
```diff
@@ -0,0 +1,1730 @@
...
+		b0Ptr += _pitch + 1;
+		b1Ptr += _pitch + 1;
+		b2Ptr += _pitch + 1;
+		b3Ptr += _pitch + 1;
+	}
+}
+
+void IndeoDecoderBase::outputPlane(IVIPlaneDesc *_plane, uint8 *dst, int dstPitch) {
+	int           x, y;
+	const int16 * src = _plane->_bands[0]._buf;
```
sev-: Looks like it is everywhere

sev-:
Now it is in a good shape. IMHO, could be merged

sev-:
Great, merging.

PR #832, opened 2016-09-18
SCI: Set default MT-32 reverb before each sound

rklaver:
While playing LSL3 with a MT-32 I noticed that after falling off a cliff there was suddenly a lot of reverb in the sound. This happens both with the emulator and with a real MT-32. After researching this, it turned out that the reverb setting was changed before playing the falling sound, but never reset. After comparing a MIDI capture from ScummVM and DOSBox I concluded that the behaviour of the original interpreter is to send the "default" reverb configuration, as present in the MT-32 patch data resource (or in earlier games MT32.DRV directly) before playing each song, if it got changed.

This fix introduces that behaviour for SCI0 games. I'm not 100% sure about the style of the fix, maybe a regular maintainer can weigh in on this.

I have also tried to verify this with PQ2 (as the reverb in this game was mentioned in an old bug report). The original interpreter sets the reverb data for the intro and if you then walk into the road it will reset the reverb data to a slightly different value (not actually audible to me) that is the default. ScummVM currently does not do this.

(Interestingly, there is a reset of the reverb to the default in the MIDI data of the falling sound resource in LSL3, but the sound is stopped before the player gets to it)

### DIFF
--- a/engines/sci/sound/drivers/midi.cpp
+++ b/engines/sci/sound/drivers/midi.cpp
@@ -65,9 +65,9 @@ public:
 	int getFirstChannel() const;
 	int getLastChannel() const;
 	void setVolume(byte volume);
+	void onNewSound();
 	int getVolume();
 	void setReverb(int8 reverb);
-	void setDefaultReverb();
 	void playSwitch(bool play);
 
 private:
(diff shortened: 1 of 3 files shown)


waltervn:
There are other SCI0 sound drivers that perform certain actions when a new sound is loaded. I suggest that we add a more generic function that can be used for all. I have some new sound drivers lined up, and these use an onNewSound function that also communicates which channels will be used (this is needed for some drivers): https://github.com/waltervn/scummvm/commit/1f397fbebf0849d5b38ebc02c2ae355f423ac2ec

rklaver:
Right, it does not need to be reverb specific, just as long as the MIDI player gets a call when starting a new sound. It that case however I would say the SCI0 check needs to be within the MIDI player implementation itself.

rklaver:
If it is not a problem that this method is only called for SCI0, sure it can. :)

csnover:
Superseded by gh-1023.

PR #833, opened 2016-09-18
SCI: Add lsl3 amiga detection checksums

h3xx:
Game reports v1.027, 1990-01-27.

screenshot-2016-0918-1637


bluegr:
Thanks, merging

PR #834, opened 2016-09-20
ANDROIDSDL: fixed switching to capital letter

lubomyr:


wjp:
It looks like you may have left some remnants of debugging code in there.

lubomyr:
Fixed, debugging code was deleted.

bluegr:
It would be good if you added a small comment, explaining why this implementation is different from the default one.

lubomyr:
In compare with original implementation of mapKey function were removed two lines, which with unknown reason broken switching to capital letter in text input.

removed lines:
	} else if (unicode) {
return unicode;

PR #835, opened 2016-09-28
SCI32: kArray, kString, kSave, fixes, clean-ups, kitchen sink

csnover:
This PR encompasses work to complete kArray/kString, which then necessitated work on kSave to use the new array structures. Also includes fixes for a variety of bugs that were discovered during testing.


wjp:
I had a review/comment session about this with @csnover on IRC, and I think this can be merged. Excellent work.

engines/sci/engine/vm.h
```diff
@@ -136,6 +136,21 @@ enum {
...
+	kDefaultPlane   = 3,  // SCI32
+	kQuit           = 4,
+	kPlanes         = 10, // SCI32
+	kCurrentRoomNo  = 11,
+	kPreviousRoomNo = 12,
+	kNewRoomNo      = 13,
+	kFastCast       = 84, // SCI16
+	kMessageType    = 90
+};
+
```
m-kiewitz: Shouldn't those be better called kGlobalVarSpeed and such?
I don't really like calling those "kQuit" etc. because you don't know immediately that those are GlobalVar enums.

I'm also not sure if it's wise to mix sci16 and sci32 together, because all sorts of things are not the same.
So maybe: "kGlobalVar16Speed" and so on. For the mixed ones, one could drop the 16/32.

wjp: See also 8a3c4bfa3761d6f75662e62df2d6dfbd983bb205.

m-kiewitz: I have seen that after writing, but it was changed to kGlobalQuit instead of kGlobalVarQuit. And I would also prefer it in case the SCI16/SCI32 exclusive ones were called for example kGlobalVar32Quit or kGlobalVarQuit32.

In other cases we have used the full enum name for the variables too. For example kDebugLevels - that one has kDebugLevelError, kDebugLevelNodes etc.

(in any case this is of course a minor thing and not a big problem - other than that everything looks fine to me (well the SCI16 stuff), excellent work)

PR #837, opened 2016-09-30
SCI32: Restore return to launcher functionality

bluegr:
RTL functionality has been broken by commit d0517f515e.
There is a check in EventManager::updateScreen() for shouldQuit(),
which is throttled to 60fps. This also restores the screen update
throttling to 60fps via the EventManager


bluegr:
Any comment on this? If there are no objections, this can be merged

csnover:
This is not the right way to fix RTL. Throttling of rendering to surface is already accomplished in frameout. The event manager should be listening for the RTL event to quit the VM the same way that it listens for the quit event.

PR #839, opened 2016-10-03
VIDEO: AVIDecoder changes for Titanic videos

dreammaster:
This pull request implements changes to the Titanic engine and AVIDecoder to more properly handle the transparency data video track that some AVI videos in Titanic have. Since it makes changes to AVIDecoder, I figure better to put it in a pull request for review.

The only change to the base VideoDecoder class is to make seekToFrame virtual, so that I can override it in AVIDecoder, since the default VideoDecoder implementation exits out if there are multiple video tracks. In AVIDecoder, I detect when there's a second video track, I figure out the offset for the first frame of the second track by scanning the _indexEntries array. Then in seekIntern, I also seek on the second video track as well by scanning the index for the offset for the second video track's frame


bluegr:
Nice work overall, well done! +1 :)

Note that we never had a sample of an AVI video with more than 1 video track until now, so the existing implementation was never tested.

sev-:
The changes are fine. Merging.

PR #842, opened 2016-10-07
I18N: Update Belarusian and Russian translations

greencis:
As always, I've updated Belarusian translation and have fixed some typos in Russian translation.


sev-:
Great, thank you!

PR #843, opened 2016-10-09
I18N: Update Spanish translation

Truido:
[es_ES.zip](https://github.com/scummvm/scummvm/files/518257/es_ES.zip)


sev-:
Thanks, I merged it in manually, since there were several problems with the change.

First, you are trying to merge in 41 commits from branch-1-9, and second, the tool which you used for translation, altered formatting even for the English entries, which created a huge but irrelevant diff file.

PR #844, opened 2016-10-09
I18N: Czech translation update

Tsbook:
I have updated the .po file with the latest changes and the ReadMe file with only minor edits.


sev-:
Thanks, merging

PR #847, opened 2016-10-13
Add graphics linear filtering feature

criezy:
In our Surface SDL code with SDL2 we are using a fake fullscreen mode that uses the desktop resolution and lets SDL scale the display to this resolution. The default SDL scaling is done using a nearest method, which results in  a pixelated image. This was reported by rootfather in https://bugs.scummvm.org/ticket/9573

As noted in the comments in this ticket, in our OpenGL code we currently have two graphics mode: "OpenGL" and "OpenGL (No filtering)" to toggle between nearest and linear filtering while scaling Duplicating all the Surface SDL scalers to add the same option does't look like a very good idea though. Instead wjp suggested to add a separate option that would affect both SDL2 and OpenGL.

This pull request attempts to implement this. Comments on whether this is a good idea, whether this is the correct way to implement it and on code details are all welcome. In particular I didn't really think of the impact on non-desktop backends that may be using SDL or OpenGL.


rootfather:
@criezy, thank you very much for this PR! I tested it on Windows 10, and it works like a charm. Very crisp and sharp graphics in VGA games, and nice and smooth graphics on the higher resolutions.

Works exactly like expected!

criezy:
@rootfather, thanks for testing.
Could you try something?
In backends\graphics\surfacesdl\surfacesdl-graphics.cpp on lines 2597 and 2631 you can replace "linear" with "best":
`SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, _videoMode.filtering ? "best" : "nearest");`

On Linux and OS X this does't change anything, but according to the documentation on Windows with Direct3D this should use an anisotropic filtering.

In the changes I made I tried to avoid using "linear" in the documentation and GUI so that backends have more freedom on the way to implement filtering. And I am in particular wondering whether to use the anisotropic filtering instead of linear filtering on Windows.

Currently I am using linear for two main reasons:
- For consistency of result across platforms and between SDL and OpenGL.
- Because I cannot test the Direct3D anisotropic filtering and I don't know how it performs in term of quality and in term of performances.

So it would be great if you could test it and report the result, both in term of performances and quality for the launcher and for low resolution games.

Also assuming the result is positive, I would welcome feedback on whether to use it or not.

rootfather:
@criezy, here are my test results. My test subjects are "The Secret of Monkey Island" for the low-res games, "Myst (ME)" for non-VGA resolutions and "The Curse of Monkey Island" for higher-res games.

Without filtering enabled, the graphics look pixelated as expected. Without any filtering options enabled, I noticed a few small graphic distortions. I already reported this behavior in [ticket 9592](https://bugs.scummvm.org/ticket/9592).

**Screenshots with linear filtering enabled**
The launcher looks great in my opinion. Monkey Island 1 is of course not as razor-sharp as previously, but still very good.
launcher-linear
launcher-linear2
monkey-linear
monkey-linear2

With linear filtering, Myst looks better than without filtering. However, I noticed that with linear filtering enabled, the last pixel columns on the sides of the image are replicated to the opposite side. Quite hard to explain, but visible on the fullscreen screenshots.

This issue doesn't happen when filtering is disabled.
Please notice the duplicated pixel line on the left side of the image:
myst-linear
Here, we have a duplicated pixel line on the right side of the image. Best visible on the bottom half.
myst-linear2

In Monkey Island 3, the image looks great with linear filtering. Again, if you look very closely, you see that the pixel columns on the very left and right are "wrapping around". In Monkey 3, it is best visible when you move the white mouse cursor to the very right/left and have a look on the opposite side of the screen.
monkey3-linear
monkey3-linear2

Now, **to the anisotropic/best filtering**
On my Windows 10 machine, the images look basically exactly the same, I couldn't spot any differences. I made two side-by-side comparisons. The read line marks the border between both rendering methods.
monkey3-sidebyside
monkey-sidebyside

The issue with the "wrapped" pixel columns at the edge of the image still persist.

I couldn't notice any differences in performance. Maybe there are setups out there that might profit from the anisotropic filtering, but on my machine and in my eyes, it makes absolutely no difference.

Feel free to ask if I should test more stuff. 👍

criezy:
Thank you for the detailed tests.
I am not familiar with SDL on Windows but I think you can actually choose if it is using Direct3D or OpenGL, perhaps at compile time (when compiling SDL, not ScummVM) or with an environment variable? That might explain why you didn't see a different when trying to use the "best" filtering.

For now I will just land this PR as it is now with linear filtering. We can always change it later if we want to.

rootfather:
I know that in SDL1, you could set an environment variable that enforced the directx/ddraw API. I have to check if this is still available for SDL2, or if SDL2 selects the "best" video driver itself.

Do you have any clue if the "pixel line" and graphics distortion in the launcher when filtering is disabled is caused by my video driver or an issue with ScummVM?

PR #848, opened 2016-10-13
RASPBERRYPI: Add specific README and launch script for using SDL 2.0.4 included in the release package. 

vanfanel:
This PR adds the specific README_raspberrypi file for notes and instructions specific to the Raspberry Pi port, and a small launch script so the scummvm port uses the included SDL 2.0.4 shared library included in libs with the binary package, so people doesn't have to build SDL 2.0.4 themselves (since the Pi foundation is taking very long to include SDL 2.0.4 with Raspbian by default).


sev-:
Great, thank you!

PR #849, opened 2016-10-16
CHEWY: Add detection entries for German versions

rootfather:
I own two original disks of "Chewy: ESC von F5", and here are the detection entries.

The game was originally published in German, the English version followed later. Both game entries will show the first scene and the "This is a test message".


Strangerke:
Thanks!

PR #261, opened 2012-07-28
COMPOSER: Support GMM saving/loading and load from launcher

upthorn:
Adds fairly robust savegame support to the composer engine module.

PR #851, opened 2016-10-18
(Updated) COMPOSER: Support GMM saving/loading and load from launcher

angstsmurf:
This is #261 with minimal changes to make it work with the latest ScummVM code. It has only been tested with Darby the Dragon on Mac OS.

### DIFF
--- a/engines/composer/composer.cpp
+++ b/engines/composer/composer.cpp
@@ -21,6 +21,7 @@
  */
 #include "common/scummsys.h"
 
+#include "common/config-manager.h"
 #include "common/events.h"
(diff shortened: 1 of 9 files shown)


sev-:
Thank you for updating the patch.

@fuzzie I'm merging it before it get rot again;

digitall:
This has caused a major regression in the graphics animation code for Darby. See bug 10861: https://bugs.scummvm.org/ticket/10861

Any help tracking this down and fixing would be appreciated...

bonki:
This was fixed in commit e45dd70.

PR #852, opened 2016-10-18
Fix some of the win32 dists file

hpvb:
Migrations.txt needs to list WinSparkle.dll, in addition several of the later LICENSE files were missing. The iss file needs to also deploy WinSparkle.dll.


Kirben:
WinSparkle would be better static linked, the Makefile and patch used by MSYS2 at http://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-winsparkle-git might be useful

hpvb:
Thanks, that is super-helpful and looks like it will work. I'll do this for 1.9.1

PR #853, opened 2016-10-20
IOS: Disable idle timer while an engine is running

criezy:
After my iPad went to sleep while I was looking at the DOTT intro cutscene I though that maybe it would be nice to disable the idle timer while an engine is running to prevent the iPad going to sleep.

The idle timer is only disabled while an engine is running. It is enabled again when returning to the launcher, so if the user doesn't do anything for a while while in the launcher, the iPad will go to sleep.

I am not sure if this is a good idea or not, so I am submitting this change as a PR.

### DIFF
--- a/backends/platform/ios7/ios7_osys_main.h
+++ b/backends/platform/ios7/ios7_osys_main.h
@@ -123,6 +123,9 @@ public:
 	static OSystem_iOS7 *sharedInstance();
 
 	virtual void initBackend();
+	
+	virtual void engineInit();
+	virtual void engineDone();
 
 	virtual bool hasFeature(Feature f);
 	virtual void setFeatureState(Feature f, bool enable);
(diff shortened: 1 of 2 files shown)


mgerhardy:
Maybe just disable this while watching cutscenes

criezy:
Yes, I though about only disabling the idle timer during cutscenes, but that would require the engines to notify the backends when playing a cutscene, which is as far as I know not the case currently.

criezy:
Note to feed the discussion: here is what the official documentation says about disabling the idle timer.

> Important
You should set this property only if necessary and should be sure to reset it to NO when the need no longer exists. Most apps should let the system turn off the screen when the idle timer elapses. This includes audio apps. With appropriate use of Audio Session Services, playback and recording proceed uninterrupted when the screen turns off. The only apps that should disable the idle timer are mapping apps, games, or programs where the app needs to continue displaying content when user interaction is minimal.

This is taken from https://developer.apple.com/reference/uikit/uiapplication/1623070-idletimerdisabled?language=objc

sev-:
I like the change.

PR #854, opened 2016-10-23
ANDROIDSDL: Changed logic for back button & getting sdcard storage location

lubomyr:
More info in commit titles.


fuzzie:
LGTM other than the question I asked above (and can always merge and fix that later).

sev-:
Cool, merging then

PR #855, opened 2016-10-23
OSD messages from any thread

criezy:
This is one attempt to fix bug #9609: SDL: Crash with displayMessageOnOSD called from separate thread in cloud code.

The OpenGL Graphics Manager was not having the issue because there was a hack to handle this case. But instead of also adding a hack to the SurfaceSDL Graphics Manager, or changing the cloud code to make sure all OSD messages are displayed from the graphics thread, I have instead added a OSD message queue class that takes messages from any thread which can then be polled by the graphics thread to be displayed.

I then modified the cloud code and MT32 code to use this new class and removed the hack from the OpenGL Graphics Manager.

I am opening this PR to get opinion on whether this is a good way to fix the issue, and also to get feedback on the actual implementation. For example I used a minimum 1s delay between messages so that each message gets displayed for at least 1s. Should we use a longer delay? Or drop messages to only display the last one when we get several messages in quick succession?

### DIFF
--- a/audio/softsynth/mt32.cpp
+++ b/audio/softsynth/mt32.cpp
@@ -42,6 +42,7 @@
 #include "common/archive.h"
 #include "common/textconsole.h"
 #include "common/translation.h"
+#include "common/osd_message_queue.h"
 
 #include "graphics/fontman.h"
 #include "graphics/surface.h"
@@ -79,7 +80,7 @@ protected:
 		error("MT32emu: Init Error - Missing PCM ROM image");
 	}
 	void showLCDMessage(const char *message) {
-		g_system->displayMessageOnOSD(message);
+		Common::OSDMessageQueue::instance().addMessage(message);
 	}
 };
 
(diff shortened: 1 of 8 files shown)


hpvb:
I think this looks good but perhaps we shouldn't let error messages expire automatically at all. But I suppose that's not really in scope for this change?

criezy:
Right, I assume you are referring to some of the cloud messages. It might indeed be better for example to use a MessageDialog, and also call logMessage() for those error messages. This might be related to this change if we decide that the best way to do it is to push the idea a bit further and convert the OSDMessageQueue from this PR into a more generic MessageQueue that would handle error and info messages in a different way.

hpvb:
Well, this is strictly a bugfix for the existing functionality right? 

Perhaps it would make more sense to merge this, as we definitely need to fix this, then do another PR to work out what to do with errors? I can also see using a sort of toaster-type notifications for non-fatal errors for instance?

fuzzie:
Merging this first for the normal OSD messages seems fine.

I'm not sure it's a good idea to deregister the event source in the destructor. In any case, it's a singleton, so it doesn't get deleted - and I guess it should be referenced early at startup, to make sure it's created at a sensible point (and not created on another thread)? Maybe it shouldn't be a singleton but just a class member? I guess that's more complicated than this is worth for now though. :-P

PR #856, opened 2016-10-26
I18N: Update Dutch translation

BenCastricum:


hpvb:
Zit er goed uit! (looks good)

criezy:
Thank you.

PR #857, opened 2016-10-29
SDL 1.2 builds for windows

JustinCB:
Please create an alternate build with SDL 1.2 for windows 9x.  Windows 9x can not run SDL 2, but should still be supported with the fallback SDL 1.2.  This will save a lot of Windows 9x users the time and energy it takes to set up MinGW to compile it.


JustinCB:
Am I requesting this wrong?

wjp:
A broken, random pull request is not the right place to have this discussion. I would suggest the forum, mailing list or bug tracker.

hpvb:
Also note that the 1.9.x series of official win32 releases will continue to support 1.2. For 1.10 this has not yet been decided.

PR #859, opened 2016-10-30
Gui apply changes

criezy:
This pull request contains two somewhat related but separate commits. Either one could be use without the other.

The first commit adds a Apply button in the global options dialog to apply the change without closing the dialog. This is mostly useful for options that result in a visual change in the dialog when we want to see the change before eventually making additional changes. I can see three options concerned by this:
- Change any graphic options.
- Change the Theme.
- Change the Language.

However currently changing the language requires to restart ScummVM. So applying the change without closing the dialog will not be very useful for this one yet. But maybe that will change.

Also currently selecting a new Theme results in the theme being changed straight away, unlike other options that are only changed when clicking on OK (or Apply with this PR). So again the Apply is not very useful for themes.

But that brings me to the second commit: it change the behaviour of changing theme to bring it in line with the other option changes and only apply the change when clicking on OK (or Apply).

Are those two changes a good idea? There was a FIXME comment in the code for the second one, so I guess I least one other developer though it would be a good idea to be consistent. Any comments on this, or on the implementation, are welcome.


criezy:
I have now added a bunch of commits that also allow to change the GUI language without restarting ScummVM. This is done by fully rebuilding the Launcher and Options dialogs, which I think are the only dialogs that can exist when we change the setting. I have been wondering though if the possibility to rebuild dialogs is something that should be extended to all dialogs.

sev-:
Great stuff! Merging.

PR #860, opened 2016-10-31
JANITORIAL: Typos found by lintian

a-detiste:
I've not fixed this one, maybe it's on purpose:

RELASE -> RELEASE

engines/mads/staticres.cpp:const char *const kGameReleaseTitleStr =
"GAME RELASE VERSION INFO";


sev-:
Nice, thank you!

dreammaster:
It was a misspelling not present in the original, thanks for pointing it out. I'll fix it now.

PR #862, opened 2016-11-26
I18N: Update/add POTFILES, enable some more strings for translation

rootfather:
This PR adds and updates some POTFILES, mostly for consistency purposes. Besides that, a few strings were added to translation. I also enabled translation for the "unknown game warning" in the SCUMM engine - we have it enabled in advancedDetector.cpp since ages.

Let me know if I did something wrong 👍

### DIFF
--- a/engines/gnap/POTFILES
+++ b/engines/gnap/POTFILES
@@ -1,0 +1,1 @@
+engines/gnap/menu.cpp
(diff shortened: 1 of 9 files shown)


sev-:
Looks great. Let's test Weblate now with it.

PR #863, opened 2016-11-26
MT32: Update Munt to 2.0.0

csnover:
This update switches to use the new Munt C++ interface, which
will allow ScummVM to link to an external Munt library instead
of requiring it to be built-in in the future. For the moment,
the emulator is still built-in, since it is not available from
most package repositories.

~A small patch is applied to this revision of Munt code since the
new interface does not expose two APIs that are used by ScummVM.
The patch is available at <https://github.com/munt/munt/pull/30>.~ Patching has been replaced with a call to writeSysex.

This changeset also removes other unnecessary direct modifications
to Munt code to ease future updates. To update Munt code in the
future:

1. Replace all source files in the `softsynth/mt32` directory with
   new files from the upstream `mt32emu/src` directory;
2. Update `config.h` with the correct version number for the new
   version of Munt;
3. Update `module.mk` to add any new source files that need to be
   built.

### DIFF
--- a/audio/softsynth/mt32/FileStream.h
+++ b/audio/softsynth/mt32/FileStream.h
@@ -15,48 +15,32 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef MT32EMU_TABLES_H
-#define MT32EMU_TABLES_H
+#ifndef MT32EMU_FILE_STREAM_H
+#define MT32EMU_FILE_STREAM_H
+
+#include <fstream>
 
 #include "globals.h"
 #include "Types.h"
+#include "File.h"
 
 namespace MT32Emu {
 
-class Tables {
-private:
-	Tables();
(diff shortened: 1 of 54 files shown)


audio/softsynth/mt32/FileStream.h
```diff
@@ -0,0 +1,46 @@
...
+ *  GNU Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MT32EMU_FILE_STREAM_H
+#define MT32EMU_FILE_STREAM_H
+
+#include <fstream>
```
sev-: As I understand, it is not used. Is it possible to get rid of it?

sev-:
OK, let's give it a second try. Nice work.

PR #865, opened 2016-12-03
CONFIGURE: Allow ppc as alias for powerpc

dwatteau:
OpenBSD, for example, uses ppc instead of powerpc.

This lets OpenBSD/macppc use ``_need_memalign=no``.

A similar fix, for amd64/x86_64, was commited in ed53292194682fb2ebc0d5a7ecd23e482899aefc.


criezy:
That looks good to me. Merging.

PR #866, opened 2016-12-06
ALL: Unify messages concerning engine data files

BenCastricum:
The helps translators a bit by removing filenames from messages so no duplicate messages have to be translated. Some engine didn't display a dialog box about the problem found. Some message were not yet translated.
Feedback appreciated.

### DIFF
--- a/README
+++ b/README
@@ -1742,6 +1742,28 @@ Broken Sword
     cz  - Czech
 
 
+5.6) Common issues:
+---- --------------
+"Engine data file missing/corrupt/wrong version"
+
+Some engines require some additional data (besides the game itself) to run
+a game. This data usually is included in the ScummVM installation package to
+ensure you always have the right version. But if for some reason you get an
+error about the engine data file, here are a few things you can try:
+
+- (Re)download and install the latest version of ScummVM package
+Because the package should contain the required data, this should fix it.
+
+- Download the engine data file manually from github
+On https://github.com/scummvm/scummvm/blob/master/dists/engine-data/ you can
+find the latest versions of all engine data files.
+
+- Check your path settings
+ScummVM will scan it's own directory and the Extra Path directory for the
+engine data files. Try moving the engine data file into those one of those
+directories.
+
+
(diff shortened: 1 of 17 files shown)


wjp:
Unifying this makes a lot of sense, thanks. But if we're going to pick one, this might not be the best phrase. I think the data files should always be in the installation package, tarball, or git repository, so it might be more likely that is a path problem than that a file needs to be downloaded.

BenCastricum:
I added a commit for alternate messages referring to the README instead of the website. Makes a bit more sens I think. Not quite sure if the README is the right place for it though.

rootfather:
I think a README bundled with every ScummVM release is a better place for such instructions than a link to the website.

README
```diff
@@ -1742,6 +1743,28 @@ Broken Sword
...
+- (Re)download and install the latest version of ScummVM package
+Because the package should contain the required data, this should fix it.
+
+- Download the engine data file manually from github
+On https://github.com/scummvm/scummvm/blob/master/dists/engine-data/ you can
+find the latest versions of all engine data files.
+
+- Check your path settings
+ScummVM will scan it's own directory and the Extra Path directory for the
+engine data files. Try moving the engine data file into those one of those
```
raziel-: One too many "those"

BenCastricum:
Fixed that one. But is the README the desired place? The FAQ is also a nice place for this info and already has a trouble shooting section where it could fit in.

BenCastricum:
Here's the proposal for not mentioning any instructions in the message. We don't do that for other messages either, so the user will follow the "normal" procedure for getting help. I would recommend a entry for this issue into the FAQ if it really turns out to be a FAQ.

Before merging (if this is the agreed way to go) I would like to squash the PR first.

criezy:
I think having the instructions in the README is fine.
Then if it turns out that this is a FAQ it can indeed be added there as well. The two are not exclusive.

But overall this is  nice pull request. Unifying those messages is indeed a good idea.

PR #868, opened 2016-12-09
CRUISE: font: Don't overwrite transparent pixels

danpla:
Fixes wrong rendering when outlines overlap. Current:
c2
With patch:
c3
Original (DOSBox):
c1


sev-:
Great! Thanks you for fixing it.

PR #869, opened 2016-12-10
CRUISE: Fix font space width

danpla:
The game uses 5px space instead of 4px.

Some screenshots for comparison. First DOSBox, then the current ScummVM.
English version:
e1
e2
Italian version. Symbols are narrower, but the space is still the same:
e3


sev-:
Please keep them coming!

PR #870, opened 2016-12-19
COMMON: Add strnlen for safer C string length reads

csnover:
This API is intended for use in cases where C strings come
from untrusted sources like game files, where malformed data
missing the null terminator would cause strlen to read out of
bounds.

### DIFF
--- a/common/str.cpp
+++ b/common/str.cpp
@@ -942,6 +942,12 @@ size_t strlcat(char *dst, const char *src, size_t size) {
 	return dstLength + (src - srcStart);
 }
 
+size_t strnlen(const char *src, size_t maxSize) {
+	size_t counter = 0;
+	while (counter != maxSize && *src++) ++counter;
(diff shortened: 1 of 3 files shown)


wjp:
Looks good to me.

common/str.cpp
```diff
@@ -942,6 +942,12 @@ size_t strlcat(char *dst, const char *src, size_t size) {
 	return dstLength + (src - srcStart);
 }
 
+size_t strnlen(const char *src, size_t maxSize) {
+	size_t counter = 0;
+	while (counter != maxSize && *src++) ++counter;
```
bluegr: Code formatting: please split this in two

bluegr:
Other than the minor formatting change, this looks OK to be merged. Kudos! :)

wjp:
Fixed the formatting and a comment, and committed manually. Thanks!

PR #871, opened 2016-12-19
COMMON: Add stddef.h to scummsys.h for ptrdiff_t

csnover:
As mentioned by @wjp, this header is included already by Director, SCUMM, Sky, and Sword25 engines so it should be fine to load in scummsys.h, but I always want to get a double-check before I put anything new into common.


wjp:
No objections here.

sev-:
Sure, no objections. Merging.

PR #872, opened 2016-12-19
SCI32: Small changes to get the demos of LSL7, Lighthouse and RAMA working

bluegr:
Here are some small changes to get the demos of LSL7 and Lighthouse working. With these changes, both demos work fine. They are both SCI2.1 late (unlike the full versions, which are SCI3).

The demo of RAMA is also SCI2.1 late, and needs a change to work, but that requires some further thought (it tries to reference a script as an array)

### DIFF
--- a/engines/sci/engine/kernel_tables.h
+++ b/engines/sci/engine/kernel_tables.h
@@ -868,7 +868,7 @@ static SciKernelMapEntry s_kernelMap[] = {
 	// our own memory manager and garbage collector, thus we simply call FlushResources, which in turn invokes
 	// our garbage collector (i.e. the SCI0-SCI1.1 semantics).
 	{ "Purge", kFlushResources,    SIG_EVERYWHERE,           "i",                     NULL,            NULL },
-	{ MAP_CALL(SetShowStyle),      SIG_EVERYWHERE,           "ioiiiii([ri])(i)",      NULL,            NULL },
+	{ MAP_CALL(SetShowStyle),      SIG_EVERYWHERE,           "ioiiiii([ri])([ri])",   NULL,            NULL },
 	{ MAP_CALL(String),            SIG_EVERYWHERE,           "(.*)",                  kString_subops,  NULL },
 	{ MAP_CALL(UpdatePlane),       SIG_EVERYWHERE,           "o",                     NULL,            NULL },
 	{ MAP_CALL(UpdateScreenItem),  SIG_EVERYWHERE,           "o",                     NULL,            NULL },
--- a/engines/sci/engine/kgraphics32.cpp
+++ b/engines/sci/engine/kgraphics32.cpp
@@ -390,7 +390,7 @@ reg_t kSetShowStyle(EngineState *s, int argc, reg_t *argv) {
 		divisions = argc > 7 ? argv[7].toSint16() : -1;
 	}
 	// SCI 2.1mid–2.1late
-	else if (getSciVersion() < SCI_VERSION_3) {
+	else if (getSciVersion() < SCI_VERSION_3 && g_sci->getGameId() != GID_LSL7) {
 		blackScreen = 0;
 		pFadeArray = argc > 7 ? argv[7] : NULL_REG;
 		divisions = argc > 8 ? argv[8].toSint16() : -1;
--- a/engines/sci/engine/klists.cpp
+++ b/engines/sci/engine/klists.cpp
@@ -813,6 +813,9 @@ reg_t kArrayGetSize(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kArrayGetElement(EngineState *s, int argc, reg_t *argv) {
+	if (s->_segMan->getSegmentType(argv[0].getSegment()) == SEG_TYPE_SCRIPT)
+		return kStrAt(s, 2, argv);
+
 	SciArray &array = *s->_segMan->lookupArray(argv[0]);
 	return array.getAsID(argv[1].toUint16());
 }
@@ -824,6 +827,13 @@ reg_t kArraySetElements(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kArrayFree(EngineState *s, int argc, reg_t *argv) {
+	if (s->_segMan->getSegmentType(argv[0].getSegment()) == SEG_TYPE_SCRIPT) {
+		reg_t source = make_reg(argv[0].getSegment(), argv[0].getOffset() + 1);
+		reg_t dest = argv[0];
+		s->_segMan->strcpy(dest, source);
+		return s->r_acc;
+	}
+
 	s->_segMan->freeArray(argv[0]);
 	return s->r_acc;
 }
--- a/engines/sci/engine/workarounds.cpp
+++ b/engines/sci/engine/workarounds.cpp
@@ -342,6 +342,8 @@ const SciWorkaroundEntry uninitializedReadWorkarounds[] = {
 	{ GID_LSL6HIRES,      -1,    85,  0,             "LL6Inv", "init",                            NULL,     0, { WORKAROUND_FAKE,   0 } }, // when creating a new game
 	{ GID_LSL6HIRES,      -1, 64950,  1,            "Feature", "handleEvent",                     NULL,     0, { WORKAROUND_FAKE,   0 } }, // at least when entering swimming pool area
 	{ GID_LSL6HIRES,      -1, 64964,  0,              "DPath", "init",                            NULL,     1, { WORKAROUND_FAKE,   0 } }, // during the game
+	{ GID_LSL7,           50, 64017,  0,             "oFlags", "clear",                           NULL,     0, { WORKAROUND_FAKE,   0 } }, // when starting the demo
+	{ GID_LSL7,         9000, 64892,  0,      "oEventHandler", "killAllEventHogs",                NULL,     1, { WORKAROUND_FAKE,   0 } }, // when clicking on help in the dice mini game
 	{ GID_MOTHERGOOSE256, -1,     0,  0,                 "MG", "doit",                            NULL,     5, { WORKAROUND_FAKE,   0 } }, // SCI1.1: When moving the cursor all the way to the left during the game - bug #5224
 	{ GID_MOTHERGOOSE256, -1,   992,  0,             "AIPath", "init",                            NULL,     0, { WORKAROUND_FAKE,   0 } }, // Happens in the demo and full version. In the demo, it happens when walking two screens from mother goose's house to the north. In the full version, it happens in rooms 7 and 23 - bug #5269
 	{ GID_MOTHERGOOSE256, 90,    90,  0,        "introScript", "changeState",                     NULL,    65, { WORKAROUND_FAKE,   0 } }, // SCI1(CD): At the very end, after the game is completed and restarted - bug #5626
--- a/engines/sci/graphics/font.cpp
+++ b/engines/sci/graphics/font.cpp
@@ -38,7 +38,9 @@ GfxFontFromResource::GfxFontFromResource(ResourceManager *resMan, GfxScreen *scr
 
 	_resource = resMan->findResource(ResourceId(kResourceTypeFont, resourceId), true);
 	if (!_resource) {
-		error("font resource %d not found", resourceId);
+		// The Lighthouse demo does not contain any fonts
+		warning("font resource %d not found", resourceId);
+		return;
 	}
 	_resourceData = _resource->data;
 
@@ -54,8 +56,10 @@ GfxFontFromResource::GfxFontFromResource(ResourceManager *resMan, GfxScreen *scr
 }
 
 GfxFontFromResource::~GfxFontFromResource() {
-	delete[] _chars;
-	_resMan->unlockResource(_resource);
+	if (_resource) {
+		delete[] _chars;
+		_resMan->unlockResource(_resource);
+	}
 }
 
 GuiResourceId GfxFontFromResource::getResourceId() {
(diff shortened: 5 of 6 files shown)


lskovlun:
> The demo of RAMA is also SCI2.1 late, and needs a change to work, but that requires some further thought (it tries to reference a script as an array)
This is a regression, I have a snapshot from August that gets past this (it crashes for some other reason that I haven't investigated). Wasn't the string/array code changed at some point?

bluegr:
@lskovlun: Yes, the array-related issues with the RAMA demo are a regression from 3f9172676526aa04983f148d1262af6ea9fb53ef. There was logic inside kArray, which handled calls to kString from within kArray (engine/klists.cpp:685), which has been removed with the rewrite by @csnover.

wjp:
What do you mean by 'tries to reference a script as an array'?

bluegr:
@wjp: in this case, a script calls kArrayGetElement on a string stored in another script's heap space

bluegr:
Update: I've also made some changes and fixes for the RAMA demo, which now works correctly as well

bluegr:
Any comments on this? If there are no objections, I'd like to have this merged by Sunday

engines/sci/engine/kernel_tables.h
```diff
@@ -868,7 +868,7 @@ static SciKernelMapEntry s_kernelMap[] = {
 	// our own memory manager and garbage collector, thus we simply call FlushResources, which in turn invokes
 	// our garbage collector (i.e. the SCI0-SCI1.1 semantics).
 	{ "Purge", kFlushResources,    SIG_EVERYWHERE,           "i",                     NULL,            NULL },
-	{ MAP_CALL(SetShowStyle),      SIG_EVERYWHERE,           "ioiiiii([ri])(i)",      NULL,            NULL },
+	{ MAP_CALL(SetShowStyle),      SIG_EVERYWHERE,           "ioiiiii([ri])([ri])",   NULL,            NULL },
```
bluegr: In that case, this part can be reverted in order to merge your changes in

engines/sci/engine/kgraphics32.cpp
```diff
@@ -390,7 +390,7 @@ reg_t kSetShowStyle(EngineState *s, int argc, reg_t *argv) {
 		divisions = argc > 7 ? argv[7].toSint16() : -1;
 	}
 	// SCI 2.1mid–2.1late
-	else if (getSciVersion() < SCI_VERSION_3) {
+	else if (getSciVersion() < SCI_VERSION_3 && g_sci->getGameId() != GID_LSL7) {
```
bluegr: Same as above, in that case we can use your changes and revert this

engines/sci/engine/klists.cpp
```diff
@@ -813,6 +813,9 @@ reg_t kArrayGetSize(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kArrayGetElement(EngineState *s, int argc, reg_t *argv) {
+	if (s->_segMan->getSegmentType(argv[0].getSegment()) == SEG_TYPE_SCRIPT)
+		return kStrAt(s, 2, argv);
```
bluegr: Fair enough, this change was the most controversial of all. This behavior is indeed strange, which is why I made a pull request out of it in the first place. Anyway, I'll leave that out

engines/sci/engine/klists.cpp
```diff
@@ -824,6 +827,13 @@ reg_t kArraySetElements(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kArrayFree(EngineState *s, int argc, reg_t *argv) {
+	if (s->_segMan->getSegmentType(argv[0].getSegment()) == SEG_TYPE_SCRIPT) {
```
bluegr: No, this function is not properly named. It frees a single ELEMENT of an array, not the whole array:

    void SegManager::freeArray(reg_t addr) {
        // (...)
        ArrayTable &arrayTable = *(ArrayTable *)_heap[addr.getSegment()];
        // (...)
	arrayTable.freeEntry(addr.getOffset());
    }

bluegr: In SSCI, this function is named "del", rather than "freeArray" - our naming very confusing, and wrong. This function should only delete a single element from an array, shifting all the rest, and that's what it does, but its naming is wrong. And SSCI does exactly the same for strings, which is what the code that I've included performs. So, it DOES make sense, since it's what the original is doing.... it's our naming of kArrayFree which is confusing.

engines/sci/engine/workarounds.cpp
```diff
@@ -342,6 +342,8 @@ const SciWorkaroundEntry uninitializedReadWorkarounds[] = {
 	{ GID_LSL6HIRES,      -1,    85,  0,             "LL6Inv", "init",                            NULL,     0, { WORKAROUND_FAKE,   0 } }, // when creating a new game
 	{ GID_LSL6HIRES,      -1, 64950,  1,            "Feature", "handleEvent",                     NULL,     0, { WORKAROUND_FAKE,   0 } }, // at least when entering swimming pool area
 	{ GID_LSL6HIRES,      -1, 64964,  0,              "DPath", "init",                            NULL,     1, { WORKAROUND_FAKE,   0 } }, // during the game
+	{ GID_LSL7,           50, 64017,  0,             "oFlags", "clear",                           NULL,     0, { WORKAROUND_FAKE,   0 } }, // when starting the demo
+	{ GID_LSL7,         9000, 64892,  0,      "oEventHandler", "killAllEventHogs",                NULL,     1, { WORKAROUND_FAKE,   0 } }, // when clicking on help in the dice mini game
```
bluegr: That is quite odd... it does happen for me. Perhaps it's another change you've made locally that addresses these?

engines/sci/graphics/font.cpp
```diff
@@ -38,7 +38,9 @@ GfxFontFromResource::GfxFontFromResource(ResourceManager *resMan, GfxScreen *scr
 
 	_resource = resMan->findResource(ResourceId(kResourceTypeFont, resourceId), true);
 	if (!_resource) {
-		error("font resource %d not found", resourceId);
+		// The Lighthouse demo does not contain any fonts
```
bluegr: That is odd. The demo files that I have are different from the ones provided by Sierrahelp. The ones from Sierrahelp do include fonts, the ones I got don't. Also, the ones I got seem to be using an older 16-bit executable, which is consistent with SCI2/SCI2.1, but the ones in Sierrahelp have a 32-bit executable.

The demos from Sierrahelp have been modified, as the page states, so this could be the reason

bluegr:
As stated, these demos are using SCI2.1 late, and are in-between versions (like QFG4 and GK1, where the demos used SCI1.1, and the full versions SCI2/SCI2.1). The main major change with SCI3 is its script interpreter, which is able to overcome the 64kb heap limit of previous versions. These demos do not use the new scripting system, rather than some small changes to kernel functions (which you've implemented already).

Regarding guesswork, as I said, I'm reading the actual code and script code before proposing changes, so I'm not trying to force any guesswork in the code, and it saddens me that you seem to be persistent about this in each of your replies. I am aware of what has been discussed, and please, don't start discussions about my time, energy or devotion to this project, this is not a contest, and please don't urge me to stop working because you believe that your work is somehow superior.

engines/sci/engine/klists.cpp
csnover: > No, this function is not properly named. It frees a single ELEMENT of an array, not the whole array: [code]

That code doesn’t do what you are saying at all. `ArrayTable` is the *segment table* that holds all the SCI32 arrays. Calling `SegManager::freeArray` deletes one *whole array*.

>In SSCI, this function is named "del"

There is definitely no such “del” function in SSCI for arrays. I can only assume you are very confused and think of a totally different engine.

> This function should only delete a single element from an array, shifting all the rest, and that's what it does

The disassembly for kArrayFree in SSCI is very straightforward, so all I can think is that you didn’t look, or were inadvertently looking at something else:

screen shot 2017-01-06 at 19 43 51

bluegr: I was referring to the SCI3 deletion function. I stand corrected regarding the arrays segment. you rewrote that code in 3f91726765 and now its behavior is what you mention.

engines/sci/graphics/font.cpp
bluegr: Update: Here's the Lighthouse demo I'm referring to:

http://www.scummvm.org/frs/demos/sci/lighthouse-win-ni-demo-en.zip

engines/sci/engine/workarounds.cpp
bluegr: Update: The demo version I've tried with is this one... perhaps yours is different?

http://www.scummvm.org/frs/demos/sci/lsl7-dos-demo-en.zip

bluegr:
So, to sum up:

- The changes in 96bf7068a08f166b84fa05cb9b6ad95f9e913f87 are done in 7b064a20b848eaf321530c835df51640eeb181b6
- The LSL7 workarounds in a09fe7c8b94ccb515e17823bff42f32e671ebe6b do not appear to occur for snover, which is quite odd (different version, perhaps?)
- The font workaround done in eba05d617b35922ee75e485d81f3cd2e27504c3c is used for one of the Lightouse demos (probably unmodified), not for the one that has been modified by SierraHelp, so this is needed, IMHO
- d95acdea70260f0a8358cbada206c3601a51cff9 has been merged
- 66904e0a201101581ff5c7cf0a0c80a0ce7268b5 is a bit controversial, but the related behavior in that demo is certainly odd anyway...

engines/sci/graphics/font.cpp
lskovlun: Some (all?) SCI3 interpreters (and it seems SCI2.1 as well) contain a fallback font embedded in the interpreter, which they will use if given the font number -1 (65535). A backtrace shows that the bad font lookup comes from the constructor to GfxText32 and not from the scripts... if we absolutely must have a font, would it not be better to implement this embedded font and use it?

csnover: @lskovlun The embedded system font has been there since at least SCI2, and is what I expected to see in the demo, but I don’t think it is code worth adding to support a non-interactive demo. Anyone who wants to run this demo can just open 2.VMD in VLC and get the same experience.

bluegr: That demo does not use any font at all, so the only needed change is to throw a warning instead of erroring out. So, it's not any major change to get it working

wjp: In any case, this commit doesn't do what the commit message claims, and I don't agree that this particular error->warning change is 'needed'.

If the error is caused by the GfxText32 constructor, that seems to mean we look up a font before it's actually necessary. Can we ignore a failing load only for that single call?

wjp: Note that most callers to getFont assume that the result is valid, so we can't simply remove this error.

lskovlun: Just tested the Lighthouse demo, and it works without a loaded font (commented out the getFont call).
Allowing a getFont(0) to fail definitely isn't right, but allowing getFont(65535) to fail might be, assuming Colin is right that nothing else requires it at the moment. That means changing this variable initialization
`int16 GfxText32::_defaultFontId = 0;`
to read 65535 (or -1) instead and special-case *that* (I'm a bit worried that this variable is signed; should I be?). If it turns out that something else does need this, of course, supporting an actual embedded font would be the answer.

csnover: This problem is addressed in 192987288a9b850cd29ac2f203341035b922b036 which will land along with a backlog of other fixes once the new bounds checking code lands.

engines/sci/engine/workarounds.cpp
bluegr: I can't reproduce the problem with "how to play" anymore. Also, the oFlags workaround needs to be applied to all rooms - it's reproducible whenever the pop up menu is clicked to jump to the mini-game. I'll adapt this accordingly and submit to master

bluegr:
Update:

- I've pushed one of the workarounds for LSL7 in 09d06855c550c5038a0a733fa610e6cf09dd71d8 - I couldn't reproduce the other one with the latest changes on master. After snover's related changes on master, the LSL7 demo works now
- With snover's font-related changes for default fonts, the Lighthouse demo should be working (have to test it)
- The RAMA demo is SCI21 late, but with the latest changes to kSetShowStyle, it errors out with an invalid signature:

    parameter 0: 0000:0000 (null, integer), should be null, integer
    parameter 1: 0017:1392 (object), should be object
    parameter 2: 0000:0001 (integer), should be null, integer
    parameter 3: 0000:ffff (integer), should be null, integer
    parameter 4: 0000:00c8 (integer), should be null, integer
    parameter 5: 0000:0000 (null, integer), should be null, integer
    parameter 6: 0000:0000 (null, integer), should be null, integer
    parameter 7: 000a:0003 (reference), should be null, integer
    parameter 8: not passed, should be reference (optional)
    parameter 9: not passed, should be null, integer (optional)
    [VM] kSetShowStyle[2e]: signature mismatch in method Styler::doit (room 0, script 64909, localCall ffffffff)!

- The changes needed for the RAMA demo will apparently need more work

engines/sci/graphics/font.cpp
bluegr: This has now been added to master in 4c942758c073ac007f8eefc819bcf7e2ac58557c, thanks csnover! :)

bluegr:
So, overall:
- SCI2.1 non-interactive demo of Lighthouse works now after the addition of the default font
- the demo of LSL7 is working now
- For completeness' sake, the SCI3 interactive demo of Lighthouse is also broken, but it is SCI3, which is still WIP
- The RAMA demo is SCI2.1 late, but it uses the earlier kSetShowStyle semantics. A workaround should be added for that
- The RAMA demo is still doing a weird array element dereference on startup - a script calls kArrayGetElement on a string stored in another script's heap space. This looks to be a script bug, so it would be best if this code was patched

I'm closing this PR since most of the changes have been merged into master now. The RAMA demo is the only one left from these demos.

Thanks again!

PR #873, opened 2016-12-23
SCI: Unconditionally save palvary state

wjp:
Additionally, add workaround to fix up old QfG3 saves with broken
_palVaryPaused state. Fixes bug #9674.

### DIFF
--- a/engines/sci/graphics/palette.cpp
+++ b/engines/sci/graphics/palette.cpp
@@ -852,6 +852,7 @@ void GfxPalette::palVaryCallback(void *refCon) {
 }
 
 void GfxPalette::palVaryIncreaseSignal() {
+	// FIXME: increments from another thread aren't guaranteed to be atomic
 	if (!_palVaryPaused)
 		_palVarySignal++;
 }
--- a/engines/sci/engine/savegame.cpp
+++ b/engines/sci/engine/savegame.cpp
@@ -764,17 +764,33 @@ void GfxPalette::saveLoadWithSerializer(Common::Serializer &s) {
 			palVaryRemoveTimer();
 
 		s.syncAsSint32LE(_palVaryResourceId);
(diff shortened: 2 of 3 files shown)


bluegr:
Looks good, but why do you only fix QFG3 saved games, and not for all SCI1.1 games?

wjp:
FYI: I checked the scripts (decompiled with SCI Companion) of ecoquest2, freddy pharkas, kq6, lb2, mgoose11, pq1, qg1, qg3, sq4, sq5, and none of them use palVaryPause.

bluegr:
I'm confused... You included QFG3 to the list... is that the only game using this feature? Or is our implementation wrong?

Since you made this research, please add a relevant comment in the source, to avoid confusion in the future

wjp:
Um, yes, except qg3 of course.

bluegr:
OK. Then please add a relevant comment in the code, explaining that this feature is only used in QFG3. Thanks! :)

engines/sci/graphics/palette.cpp
```diff
@@ -853,7 +853,7 @@ void GfxPalette::palVaryCallback(void *refCon) {
 
 void GfxPalette::palVaryIncreaseSignal() {
 	if (!_palVaryPaused)
-		_palVarySignal++;
```
bluegr: And since this isn't apparent, the comment where I'm asking for changes/commits to be addressed refers to this change

wjp:
Right, good point. I put back the FIXME for now then.

bluegr:
Thanks! Looks good to be merged now :)

bluegr:
Any comments on this? If there are no objections, I'd like to have this merged by Sunday

wjp:
I'll do the palVarySignal thread safety in a separate commit and/or PR later.

PR #874, opened 2016-12-26
WINTERMUTE: Fix PathUtils and add workaround for mixed separators

tobiatesan:
TL;DR: This is PR #733 + #714.
Fixes bugs 7068 and 6594

 ---- 
This is the same exact branch as PR #733;

In addition, I cherry-picked Zoltan's commit from #714; and added a patch on top of it so that it uses the existing normalizeFilename() method in PathUtils for the sake of having a bit less duplication.

After doing so, #7068 *appears* to not have come back, if I have understood the bug report correctly (I can get to the end of the conversation).

I also reworded Zoltan's commit so that it includes the line "fixed #7068".
it is a one-liner commit, but I find he deserves the credit for tracking down the bug.

If you like it enough, you can merge it directly, otherwise merging #733 and #714 separately and then maybe adding this on top is probably equivalent.


sev-:
Thanks, let's merge it

PR #876, opened 2016-12-27
TEST: Only build Wintermute tests when that engine is enabled

csnover:
This fixes `make test` when wintermute engine is disabled after commit e3fdd8d5fe5b739b28c67539bbdb0b596f9dacbe, but I am not 100% sure that this is the right solution.

### DIFF
--- a/test/module.mk
+++ b/test/module.mk
@@ -5,8 +5,13 @@
 #
 ######################################################################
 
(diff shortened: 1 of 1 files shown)


sev-:
Yes, that's a clean way. Thanks, merging.

PR #877, opened 2016-12-27
SCUMM: Fix bug #2285 (ZAK C64/V1/V2 Caponians dont disguise after using crystal)

segrax:
This is one possible fix for #2285, which is actually a scripting oversight, not a ScummVM bug.

I'm not really sure if bugs like this are generally fixed. But after spending the time investigating it, a fix was simple enough to put together. 

This bug affects C64, V1, and V2 versions of Zak (I do not have the FMTowns version, so can not check it)
I'll copy and paste the analysis of the problem from the bug report,

"
The cut scene (script-96) inside the telephone company triggers just after first using the crystal, setting Var[245] (Actor 7's Costume) to 0x1C (Disguise On) and locking object 344. This cut scene is not played again.

When you fly to the Bermuda Triangle and abducted, you can ring the door bell... this sets Var[245] back to 0x18 (Disguise Off). The previous variable state is not stored, and thus not restored upon leaving the craft.

When you use the blue crystal again, object 344 being locked prevents script 96 from being started, thus Var[245] is not changed. 

You can also visit the telephone company which will set Var[245] to 0x1C (Disguise on)
"

### DIFF
--- a/engines/scumm/script_v2.cpp
+++ b/engines/scumm/script_v2.cpp
@@ -1277,6 +1277,35 @@ void ScummEngine_v2::o2_putActorInRoom() {
 
 		a->putActor(0, 0, 0);
 	}
+
+	// WORKAROUND bug #2285: Caponians dont disguise after using blue crystal
+	// This is for a game scripting oversight.
+	// After first using the blue crystal, a cutscene of the two Caponians plays (script-96),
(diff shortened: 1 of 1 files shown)


sev-:
This is in manner as we did other fixes. Thanks, merging

PR #878, opened 2016-12-30
FULLPIPE: Fix leaks in surface and bitmap usage.

jammm:
Fixes part of bug #9654.


sev-:
Thanks, this is a step in the right direction

PR #879, opened 2016-12-31
WME: Add three fangames from jennibee

raziel-:
- DFAF Adventure
- Dreamcat
- Open Quest

Freely available from https://jennibee.itch.io/ those are three (very short) fangames.
All of those games come in three packages (Linux, OSX, Windows32) but all the packages differ only in the included platform's main .exe, the data.dcp stays the same, so the entry is limited to one for all three games.

The packages even come with ScummVM 1.8.1

All three games tested and working (didn't solve the Open Quest final riddle though)
One script runtime error encountered in DFAF Adventure on examining the "speaker", other than that all games are fine and none crash.

Feel free to give hints on what i should change.

Thanks a lot


raziel-:
Ugh, sorry...partly duplicate from #808 

...i knew i heard the name @JenniBee  before :-(

I guess it's best to close?

sev-:
Great, thanks!

PR #881, opened 2017-01-01
COMMON: Add Span to common library

csnover:
(This PR contains code that relies on #870.)

Span is roughly modelled on the GSL span<T> type, and is intended to replace direct access to raw pointers—especially pointers that are passed to functions along with a separate size parameter. It provides low-cost bounds-checked reads and writes, as well as convenience functions for reading common values (integers of varying endianness, strings, etc.). While similar to MemoryReadStream in purpose, Span is superior in cases where memory is writable, where memory is accessed randomly rather than sequentially, or where any invalid access should be treated as an unrecoverable error. It should also be more efficient than a MemoryReadStream because it is implemented using CRTP, so there is no runtime overhead from dynamic dispatch.

NamedSpan is an extension of Span which provides enhanced debugging information when out-of-bounds memory accesses occur. It allows programmers to name the memory span at construction time, and it also tracks the offsets of subspans so that the absolute byte offset of the original memory can be provided in the error message if an out-of-bounds access occurs (i.e. if you were to create a subspan to read a block of data from a resource, and there was an access violation, the error will include information on the position within the original resource, not just the position within the subspan).

SpanOwner is similar to ScopedPtr but has awareness of the design of Span objects, so allows the memory pointed to by the Span object inside the SpanOwner to be freed when the SpanOwner is freed without requiring holding a separate pointer to the start of memory. It also provides some copy semantics, so unlike a ScopedPtr, SpanOwners can be held by objects in movable containers like Common::Array -- but note that because there are no move semantics in C++98, this means that a new, complete memory copy of the pointed-to data will be created, rather than just a new Span pointing to the same block of memory, when a container holding a SpanOwner expands.

This PR also includes a bit of extra support code that hasn’t been PR’d separately: (1) type traits and (2) safe bool idiom. The safe bool code was lifted from boost so has a separate header license.

This code currently is used in my working branch for bounds checking resources in SCI engine. I will send a separate PR with those changes, which hopefully will provide insight into the design of this API.

One feature I didn’t feel very happy with (giving Spans a forward-iterator interface) ended up going into the SCI-specific span subclass, since it was convenient for simplifying some of the reads in SCI, but it interacts poorly with e.g. SpanOwner, so I’m not sure it should be encouraged for global use.

Other than that, the things I like the least about this code, and would be most interested for feedback on, are:

1. The existence of SpanIterators, versus adding one additional size_t-sized offset property to all Spans, which would allow them to be treated mostly like regular pointers (but also cause them to be 50% larger than necessary);
2. Performing memory allocations through Span (convenient, but it seems weird for a view-like construct to include functions that allow new memory allocations);
3. Overload some of the existing READ_FOO and WRITE_FOO helpers instead of using member functions (member functions seemed to me to result in more readable code, but I don’t know if the boilerplate is worth the trade-off).

Thanks,


sev-:
...and your tests are broken. See Travis logs

csnover:
> ...and your tests are broken. See Travis logs

Apple clang is either less strict than others, or there is a difference in the way builds in Travis-CI are configured. It is OK now.

csnover:
I’ve increased the amount of test coverage and reduced some apparently unnecessary boilerplate code. I also fixed some bugs with non-byte-sized span value types, and added compile-time errors for attempts to use the helper functions in ways that would involve system endianness.

wjp:
There's still a lot of shadowing warnings with gcc 4.9, but other than that, I think this is ready to be merged. Great work.

csnover:
The shadow warnings have been fixed. @sev- any other thoughts?

sev-:
Cool beans!

PR #882, opened 2017-01-01
SCI: Bounds checking

csnover:
This is the follow-up to #881, containing just the modifications to SCI to add bounds checking of resources. ~~This PR relies on code not included in this branch, so it will not build if you try :). [This ZIP](https://github.com/csnover/scummvm/archive/28054e84eac78856cace0d092dcc7efed3de034e.zip) (my [working branch](https://github.com/csnover/scummvm/tree/working) at the time of this PR) contains all the code necessary for a running build.~~

Here are the games I have tested with this code (versions listed are game versions):

Full playthrough:
* CHRISTMAS1988 (1.04)
* LSL2 (1.002.000)
* SHIVERS (1.02)
* SQ6 (1.0)

Partial testing (start game, interact through a few rooms):
* ECOQUEST (floppy 1.000, CD 1.1)
* GK1 (floppy 1.0, CD 1.0a)
* GK2 (1.0)
* HOYLE5 (demo)
* ICEMAN (1.023)
* ISLANDBRAIN (1.0)
* KQ1 (1.000.051)
* KQ4 (1.006.004)
* KQ5 (1.000.052)
* KQ6 (1.000.00G)
* LAURABOW (1.000.046)
* LAURABOW2 (floppy 1.000, CD 1.1)
* LSL1VGA (2.1)
* LSL3 (1.021)
* LSL5 (1.000)
* LSL6 (1.000.000)
* LSL6HIRES (1.000.000)
* MOTHERGOOSE256 (floppy 2.000)
* PHANTASMAGORIA (1.000.000)
* PQ1VGA (2.000)
* PQ2 (1.002.011)
* PQ3 (1.00)
* PQ4 (1.100.000)
* QFG1 (EGA 1.200, VGA 2.000)
* QFG2 (1.105)
* QFG3 (1.1)
* QFG4 (Ger floppy 1.1, CD 1.0)
* SQ1VGA (2.000)
* SQ3 (1.018)
* SQ4 (CD 1.0)
* SQ5 (1.04)
* TORIN (1.0)

### DIFF
--- a/engines/sci/engine/kvideo.cpp
+++ b/engines/sci/engine/kvideo.cpp
@@ -27,8 +27,10 @@
 #include "sci/graphics/cursor.h"
 #include "sci/graphics/palette.h"
 #include "sci/graphics/screen.h"
+#include "sci/util.h"
 #include "common/events.h"
 #include "common/keyboard.h"
+#include "common/span.h"
 #include "common/str.h"
 #include "common/system.h"
 #include "common/textconsole.h"
@@ -53,19 +55,21 @@ void playVideo(Video::VideoDecoder *videoDecoder, VideoState videoState) {
 
 	videoDecoder->start();
 
-	byte *scaleBuffer = 0;
+	Common::SpanOwner<SciSpan<byte> > scaleBuffer;
 	byte bytesPerPixel = videoDecoder->getPixelFormat().bytesPerPixel;
 	uint16 width = videoDecoder->getWidth();
 	uint16 height = videoDecoder->getHeight();
 	uint16 pitch = videoDecoder->getWidth() * bytesPerPixel;
 	uint16 screenWidth = g_sci->_gfxScreen->getDisplayWidth();
 	uint16 screenHeight = g_sci->_gfxScreen->getDisplayHeight();
+	uint32 numPixels;
 
 	if (screenWidth == 640 && width <= 320 && height <= 240) {
 		width *= 2;
 		height *= 2;
 		pitch *= 2;
-		scaleBuffer = new byte[width * height * bytesPerPixel];
+		numPixels = width * height * bytesPerPixel;
+		scaleBuffer->allocate(numPixels, videoState.fileName + " scale buffer");
 	}
 
 	uint16 x = (screenWidth - width) / 2;
@@ -84,9 +88,10 @@ void playVideo(Video::VideoDecoder *videoDecoder, VideoState videoState) {
 
 			if (frame) {
 				if (scaleBuffer) {
+					const SciSpan<const byte> input((const byte *)frame->getPixels(), frame->w * frame->h * bytesPerPixel);
 					// TODO: Probably should do aspect ratio correction in KQ6
-					g_sci->_gfxScreen->scale2x((const byte *)frame->getPixels(), scaleBuffer, videoDecoder->getWidth(), videoDecoder->getHeight(), bytesPerPixel);
-					g_system->copyRectToScreen(scaleBuffer, pitch, x, y, width, height);
+					g_sci->_gfxScreen->scale2x(input, *scaleBuffer, videoDecoder->getWidth(), videoDecoder->getHeight(), bytesPerPixel);
+					g_system->copyRectToScreen(scaleBuffer->getUnsafeDataAt(0, pitch * height), pitch, x, y, width, height);
 				} else {
 					g_system->copyRectToScreen(frame->getPixels(), frame->pitch, x, y, width, height);
 				}
@@ -111,7 +116,6 @@ void playVideo(Video::VideoDecoder *videoDecoder, VideoState videoState) {
 		g_system->delayMillis(10);
 	}
 
-	delete[] scaleBuffer;
 	delete videoDecoder;
 }
 
--- a/engines/sci/engine/object.cpp
+++ b/engines/sci/engine/object.cpp
@@ -51,24 +51,25 @@ static bool relocateBlock(Common::Array<reg_t> &block, int block_location, Segme
 	return true;
 }
 
-void Object::init(byte *buf, reg_t obj_pos, bool initVariables) {
-	byte *data = buf + obj_pos.getOffset();
+void Object::init(const SciSpan<const byte> &buf, reg_t obj_pos, bool initVariables) {
+	const SciSpan<const byte> data = buf.subspan(obj_pos.getOffset());
 	_baseObj = data;
 	_pos = obj_pos;
 
 	if (getSciVersion() <= SCI_VERSION_1_LATE) {
-		_variables.resize(READ_LE_UINT16(data + kOffsetSelectorCounter));
-		_baseVars = (const uint16 *)(_baseObj + _variables.size() * 2);
-		_methodCount = READ_LE_UINT16(data + READ_LE_UINT16(data + kOffsetFunctionArea) - 2);
+		const SciSpan<const byte> header = buf.subspan(obj_pos.getOffset() - kOffsetHeaderSize);
+		_variables.resize(header.getUint16LEAt(kOffsetHeaderSelectorCounter));
+		_baseVars = _baseObj.subspan<const uint16>(_variables.size() * sizeof(uint16));
+		_methodCount = data.getUint16LEAt(header.getUint16LEAt(kOffsetHeaderFunctionArea) - 2);
 		for (int i = 0; i < _methodCount * 2 + 2; ++i) {
-			_baseMethod.push_back(READ_SCI11ENDIAN_UINT16(data + READ_LE_UINT16(data + kOffsetFunctionArea) + i * 2));
+			_baseMethod.push_back(data.getUint16SEAt(header.getUint16LEAt(kOffsetHeaderFunctionArea) + i * 2));
 		}
 	} else if (getSciVersion() >= SCI_VERSION_1_1 && getSciVersion() <= SCI_VERSION_2_1_LATE) {
-		_variables.resize(READ_SCI11ENDIAN_UINT16(data + 2));
-		_baseVars = (const uint16 *)(buf + READ_SCI11ENDIAN_UINT16(data + 4));
-		_methodCount = READ_SCI11ENDIAN_UINT16(buf + READ_SCI11ENDIAN_UINT16(data + 6));
+		_variables.resize(data.getUint16SEAt(2));
+		_baseVars = buf.subspan<const uint16>(data.getUint16SEAt(4), _variables.size() * sizeof(uint16));
+		_methodCount = buf.getUint16SEAt(data.getUint16SEAt(6));
 		for (int i = 0; i < _methodCount * 2 + 3; ++i) {
-			_baseMethod.push_back(READ_SCI11ENDIAN_UINT16(buf + READ_SCI11ENDIAN_UINT16(data + 6) + i * 2));
+			_baseMethod.push_back(buf.getUint16SEAt(data.getUint16SEAt(6) + i * 2));
 		}
 	} else if (getSciVersion() == SCI_VERSION_3) {
 		initSelectorsSci3(buf);
@@ -77,9 +78,9 @@ void Object::init(byte *buf, reg_t obj_pos, bool initVariables) {
 	if (initVariables) {
 		if (getSciVersion() <= SCI_VERSION_2_1_LATE) {
 			for (uint i = 0; i < _variables.size(); i++)
-				_variables[i] = make_reg(0, READ_SCI11ENDIAN_UINT16(data + (i * 2)));
+				_variables[i] = make_reg(0, data.getUint16SEAt(i * 2));
 		} else {
-			_infoSelectorSci3 = make_reg(0, READ_SCI11ENDIAN_UINT16(_baseObj + 10));
+			_infoSelectorSci3 = make_reg(0, _baseObj.getUint16SEAt(10));
 		}
 	}
 }
@@ -89,20 +90,20 @@ const Object *Object::getClass(SegManager *segMan) const {
 }
 
 int Object::locateVarSelector(SegManager *segMan, Selector slc) const {
-	const byte *buf = 0;
+	SciSpan<const byte> buf;
 	uint varnum = 0;
 
 	if (getSciVersion() <= SCI_VERSION_2_1_LATE) {
 		const Object *obj = getClass(segMan);
 		varnum = getSciVersion() <= SCI_VERSION_1_LATE ? getVarCount() : obj->getVariable(1).toUint16();
-		buf = (const byte *)obj->_baseVars;
+		buf = obj->_baseVars.subspan<const byte>(0);
 	} else if (getSciVersion() == SCI_VERSION_3) {
 		varnum = _variables.size();
-		buf = (const byte *)_baseVars;
+		buf = _baseVars.subspan<const byte>(0);
 	}
 
 	for (uint i = 0; i < varnum; i++)
-		if (READ_SCI11ENDIAN_UINT16(buf + (i << 1)) == slc) // Found it?
+		if (buf.getUint16SEAt(i << 1) == slc) // Found it?
 			return i; // report success
 
 	return -1; // Failed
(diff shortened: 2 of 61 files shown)


csnover:
I had broken the SCI16-only build (this always happens only when I don’t check it before pushing stuff, somehow…) so the link and commits have been revised. I also forgot to put the KQ games I tested on the list. It has been updated.

csnover:
This PR is now rebased on top of #886.

engines/sci/resource.cpp
```diff
@@ -2226,26 +2232,25 @@ static const byte detectSci21NewStringSignature[] = {
...
-	const byte *resourceData = resource->data;
 
 	signature++; // skip over size byte
 	if (signatureSize < 4)
 		error("resource signature is too small, internal error");
-	if (signatureSize > resource->size)
+	if (signatureSize > resource->size())
 		return false;
 
 	const uint32 signatureDWord = *((const uint32 *)signature);
```
wjp: Not caused by your patch, but this read violates alignment requirements and should be a `READ_UINT32`.

engines/sci/engine/savegame.cpp
```diff
@@ -437,40 +437,41 @@ void HunkTable::saveLoadWithSerializer(Common::Serializer &s) {
...
 			assert(blockSize > 0);
 
 			if (blockType == SCI_OBJ_STRINGS)
-				s.syncBytes(buf, blockSize);
+				s.syncBytes(buf.getUnsafeDataAt(0, blockSize), blockSize);
 
 			buf += blockSize;
 
-			if (_buf - buf == 0)
+			if (*_buf - buf == 0)
```
wjp: I've removed this in cdfe5a3107397ca39b7638d0d44c43c3be42d6df .

engines/sci/resource.cpp
wjp: I've fixed this in 905feaa8c24fc03d47df040dc1fc62cac464261e.

engines/sci/engine/script.h
```diff
@@ -192,10 +191,10 @@ class Script : public SegmentObj {
 	void setLockers(int lockers);
 
 	/**
-	 * Retrieves a pointer to the exports of this script
-	 * @return	pointer to the exports.
+	 * Retrieves the offset of the export table in the script
+	 * @return	the exports offset.
 	 */
-	const uint16 *getExportTable() const { return _exportTable; }
+	uint getExportsOffset() const { return (SciSpan<const byte> &)_exports - *_buf; }
```
wjp: Can this be replaced by `sourceByteOffset` to avoid a const-ness warning?

engines/sci/engine/object.cpp
```diff
@@ -136,14 +137,14 @@ int Object::propertyOffsetToId(SegManager *segMan, int propertyOffset) const {
 	}
 
 	if (getSciVersion() < SCI_VERSION_1_1) {
-		const byte *selectoroffset = ((const byte *)(_baseObj)) + kOffsetSelectorSegment + selectors * 2;
-		return READ_SCI11ENDIAN_UINT16(selectoroffset + propertyOffset);
+		const SciSpan<const byte> selectoroffset = _baseObj.subspan<const byte>(0).subspan(kOffsetSelectorSegment + selectors * 2);
```
csnover: Nope! It is gone now.

engines/sci/engine/object.h
```diff
@@ -74,21 +76,48 @@ class Object {
...
+	Object &operator=(const Object &other) {
+		_baseObj = other._baseObj;
+		_baseMethod = other._baseMethod;
+		_variables = other._variables;
+		_methodCount = other._methodCount;
+		_flags = other._flags;
+		_offset = other._offset;
+		_pos = other._pos;
+
+		if (getSciVersion() == SCI_VERSION_3) {
```
csnover: In 9aaf1a008b5e1e75fa6c5b23cc8993a34e0680b9 (SCI32: Fix up Object to use initializers and more containers) and 0aa44aaf85f39295b2de2eb1dde4b7436311eb43 (SCI32: Fix SCI3 save games), `_propertyOffsetsSci3` and `_baseVars` are changed to use `Common::Array`.

engines/sci/engine/script.cpp
```diff
@@ -125,74 +126,83 @@ void Script::load(int script_nr, ResourceManager *resMan, ScriptPatcher *scriptP
...
-			_numExports = READ_SCI11ENDIAN_UINT16(_exportTable + 1);
-			_exportTable += 3;	// skip header plus 2 bytes (_exportTable is a uint16 pointer)
+		SciSpan<const uint16> exportTable = findBlockSCI0(SCI_OBJ_EXPORTS).subspan<const uint16>(0);
+		if (exportTable) {
+			// The export table is after the block header (4 bytes / 2 uint16s)
+			// and the number of exports (2 bytes / 1 uint16).
+			// The exports span does not need to be explicitly sized since the
+			// maximum size was already determined by findBlockSCI0
+			_exports = exportTable.subspan(3);
+			_numExports = exportTable.getUint16SEAt(2);
```
csnover: Yes, the old code was incorrectly reading the byte size of the export block, not the export count.

engines/sci/engine/script.cpp
```diff
@@ -955,14 +933,19 @@ void Script::syncLocalsBlock(SegManager *segMan) {
 }
 
 void Script::initializeClasses(SegManager *segMan) {
-	const byte *seeker = 0;
+	SciSpan<const byte> seeker;
 	uint16 mult = 0;
 
 	if (getSciVersion() <= SCI_VERSION_1_LATE) {
-		seeker = findBlockSCI0(SCI_OBJ_CLASS);
+		seeker = _script;
```
csnover: Yes, `findBlockSCI0` now returns a Span that is the size of the found block, whereas this seeker needs to be able to iterate over the entire script. As far as I could tell, the only thing the `findBlockSCI0` call was doing here was fast forwarding to the first class block.

engines/sci/engine/kvideo.cpp
```diff
@@ -84,9 +88,10 @@ void playVideo(Video::VideoDecoder *videoDecoder, VideoState videoState) {
 
 			if (frame) {
 				if (scaleBuffer) {
+					const SciSpan<const byte> input((const byte *)frame->getPixels(), frame->w * frame->h * bytesPerPixel);
 					// TODO: Probably should do aspect ratio correction in KQ6
-					g_sci->_gfxScreen->scale2x((const byte *)frame->getPixels(), scaleBuffer, videoDecoder->getWidth(), videoDecoder->getHeight(), bytesPerPixel);
-					g_system->copyRectToScreen(scaleBuffer, pitch, x, y, width, height);
+					g_sci->_gfxScreen->scale2x(input, *scaleBuffer, videoDecoder->getWidth(), videoDecoder->getHeight(), bytesPerPixel);
+					g_system->copyRectToScreen(scaleBuffer->getUnsafeDataAt(0, width * height), pitch, x, y, width, height);
```
csnover: Yes, thanks. Fixed.

engines/sci/engine/script.cpp
```diff
@@ -125,74 +126,83 @@ void Script::load(int script_nr, ResourceManager *resMan, ScriptPatcher *scriptP
...
+	bufSize += extraLocalsWorkaround * 2;
 
-	_buf = (byte *)malloc(_bufSize);
-	assert(_buf);
-
-	assert(_bufSize >= script->size);
-	memcpy(_buf, script->data, script->size);
+	SciSpan<byte> outBuffer = _buf->allocate(bufSize, script->name() + " buffer");
+	script->copyDataTo(outBuffer);
+	_script = _buf->subspan(0, scriptSize, script->name());
```
csnover: Good catch. Fixed.

engines/sci/engine/script.cpp
```diff
@@ -125,74 +126,83 @@ void Script::load(int script_nr, ResourceManager *resMan, ScriptPatcher *scriptP
...
-		const byte* localsBlock = findBlockSCI0(SCI_OBJ_LOCALVARS);
-		if (localsBlock) {
-			_localsOffset = localsBlock - _buf + 4;
-			_localsCount = (READ_LE_UINT16(_buf + _localsOffset - 2) - 4) >> 1;	// half block size
+
+		SciSpan<const byte> localsTable = findBlockSCI0(SCI_OBJ_LOCALVARS);
+		if (localsTable) {
+			// skip header (4 bytes)
+			_localsOffset = localsTable - *_buf + 4;
+			_localsCount = (_buf->getUint16LEAt(_localsOffset - 2) - 4) >> 1; // half block size
```
csnover: It seems that way. Updated.

engines/sci/engine/script.h
csnover: Yeah. That makes way more sense.

engines/sci/engine/script.cpp
csnover: Looks like more work needs to be done for this beyond just using the unpadded size—changing only `scriptSize` to `script->size()` causes at least EQ2 to get a bogus locals size on some scripts.

engines/sci/engine/script.cpp
```diff
@@ -125,74 +126,83 @@ void Script::load(int script_nr, ResourceManager *resMan, ScriptPatcher *scriptP
...
-		if (_numExports % 2)
-			_localsOffset = 22 + _numExports * 2;
-		else
-			_localsOffset = 24 + _numExports * 2;
+		_localsCount = _buf->getUint16LEAt(12);
+		_numExports = _buf->getUint16LEAt(20);
+		if (_numExports) {
+			_exports = _buf->subspan<const uint16>(22, _numExports * sizeof(uint16));
+			// SCI3 local variables always start dword-aligned
+			// TODO: Isn't this condition inverted from what it should be?
```
csnover: Yeah. Not sure what I was thinking.

csnover:
I’ve updated this PR based on all comments received thus far. The only thing I didn’t change is the size of the script span; instead, I added a comment about why the padded size is used, which I hope should suffice for now.

engines/sci/graphics/view.cpp
```diff
@@ -820,12 +795,12 @@ void GfxView::draw(const Common::Rect &rect, const Common::Rect &clipRect, const
 	const int16 width = MIN(clipRect.width(), celWidth);
 	const int16 height = MIN(clipRect.height(), celHeight);
 
-	bitmap += (clipRect.top - rect.top) * celWidth + (clipRect.left - rect.left);
+	const byte *bitmapData = bitmap.getUnsafeDataAt((clipRect.top - rect.top) * celWidth + (clipRect.left - rect.left), width * height);
```
wjp: This `width * height` doesn't seem correct for (horizontally) clipped views.

I'm thinking `celWidth * height - (clipRect.left - rect.left)` ?

wjp:
I've now also gone through all the graphics code. Only audio code remaining now.

engines/sci/graphics/view.cpp
wjp: Or I suppose `celWidth * (height - 1) + width`.

wjp: Except that would break for `height == 0`. Never mind...

engines/sci/sound/drivers/fmtowns.cpp
```diff
@@ -461,12 +461,12 @@ int MidiDriver_FMTowns::open() {
...
-	if (data) {
-		data += 6;
-		for (int i = 0; i < 128; i++) {
-			_intf->callback(5, 0, i, data);
-			data += 48;
+void MidiDriver_FMTowns::loadInstruments(const SciSpan<const uint8> &data) {
+	if (data.size()) {
+		SciSpan<const uint8>::const_iterator it = data.cbegin() + 6;
+		for (int i = 0; i < 128; i++, it += 48) {
+			// TODO: Unsafe use of va_args
```
wjp: This TODO is confusing. By `unsafe use`, do you mean the pointer is used without bounds checking (in `intf_loadInstrument` I guess?), or is there actual unsafe use of va_args itself somewhere?

engines/sci/sound/drivers/midi.cpp
```diff
@@ -656,13 +659,13 @@ void MidiPlayer_Midi::readMt32DrvData() {
 			f.seek(-2, SEEK_CUR);
 
 		// Send before-SysEx text
-		sendMt32SysEx(0x200000, static_cast<Common::SeekableReadStream *>(&f), 20);
```
wjp: Not your patch, but these static_casts are weird.

wjp:
I've gone through the whole PR now; no further comments. Looks great.

engines/sci/sound/drivers/midi.cpp
wjp: I've removed these in 1b1fae6494fb64840f83c509d35e9062abecce13.

engines/sci/sound/drivers/fmtowns.cpp
csnover: Yeah, that was a badly written message. I must have intended to go through the FMTowns code to figure out the safe way to do this, but forgot. The TODO is addressed in 93968a9e3f939653a47f1855bfe5e47d69c20821.

engines/sci/graphics/view.cpp
csnover: Good catch. Both these equations actually would break for clipped views if height is zero; the solution to me seems to be to just to return early if width or height are zero (since nothing gets drawn in that case anyway), and use `celWidth * (height - 1) + width`, which is more accurate (in a hypothetical edge case where view data is truncated mid-line, after the `width`). So, I’ve made that change.

engines/sci/graphics/font.cpp
```diff
@@ -87,13 +91,16 @@ void GfxFontFromResource::draw(uint16 chr, int16 top, int16 left, byte color, bo
 	int y = 0;
 	int16 greyedTop = top;
 
-	byte *pIn = getCharData(chr);
+	SciSpan<const byte> charData = getCharData(chr);
+	if (!charData) {
+		error("Character '%c' does not exist in font %d", chr, _resourceId);
+	}
```
m-kiewitz: This is not like it was before.
getCharData defaulted to the first char (which is afaik normally a space) in case an unknown character was asked for. Now it bombs out.

SCI16 behavior worked also exactly like that (using first character for unknown ones). It wasn't some coding error on our end, but intentionally.

engines/sci/graphics/view.cpp
```diff
@@ -933,15 +912,17 @@ void GfxView::drawScaled(const Common::Rect &rect, const Common::Rect &clipRect,
 	const int16 offsetY = clipRect.top - rect.top;
 	const int16 offsetX = clipRect.left - rect.left;
 
+	// TODO: Remove? This class is not for SCI32
 	// Happens in SQ6, first room
 	if (offsetX < 0 || offsetY < 0)
 		return;
```
m-kiewitz: Checks can be removed

m-kiewitz:
Went through. I don't like that there are so many changes and would have preferred it in case they were split up and in any case, I have quite a bad feeling about it - we really need to request full play throughs for basically all SCI games for the next release and even then...
I want to play through Colonel's Bequest + Dagger of Amon Ra. I guess I will wait for those until the PR was merged.

Well, everything except that font issue looks fine to me. Still I'm not an expert on the object + scripts code. A bug may lurk in there. Great work.

Need to try out KQ5 floppy when I have time. I hope the new code won't make the game error() out now. Back the it just played the credits music, then some tracks went bad and were muted off and that was it.

engines/sci/graphics/font.cpp
csnover: I don’t understand. `getCharData` previously returned a null pointer if the requested character was out of range, not a pointer to the first character. So as far as I can tell, this is the same as it was before, except without the potential of a null pointer dereference?

csnover: Oh, wait, I see now the difference. It would have not actually tried to read from the pointer at all in that case because the character dimensions would be zero.

engines/sci/graphics/view.cpp
csnover: Thanks for verifying! This check has been removed.

engines/sci/graphics/font.cpp
csnover: The draw methods are updated with a check on `chr` & warning instead of an error now.

csnover:
> I don't like that there are so many changes and would have preferred it in case they were split up

Me too; unfortunately, as I was working on this, there were increasingly difficult-to-resolve conflicts when trying to keep the changes split up across multiple commits (and ensuring those individual commits would actually compile). This happened primarily because the API of Span/SciSpan evolving during this work (to account for different access patterns used by different parts of the codebase), and so those changes required constantly going back through all the “finished” code and updating it to use the revised API. After some time of this, and after moving the Span class to Common, it became pretty much necessary to squash everything together to get it in shape for landing.

> we really need to request full play throughs for basically all SCI games for the next release and even then...

This is by far the largest change being made, so definitely has the largest potential regression surface, but I do have some additional changes queued (like the SCI0/1 object parsing stuff I have been talking about in the last week) that might also affect some SCI16 games, so I’d say to not expend effort quite yet on play testing (unless you really, really want to).

csnover:
Thank you for your thoughtful reviews! This has been landed. 1298762b7665dc1b7aeedf0271eadfb284309ef1...532828c143e0872525dadfe8b0ebe804f7071633

PR #883, opened 2017-01-04
Fullpipe: Plug some memory leaks

bluegr:
This is the first of a series of PRs to plug memory leaks in Full Pipe. There are more pending changes, which will need further testing and will be submitted in separate PRs


sev-:
Cool, thanks!

PR #884, opened 2017-01-05
I18N: Some improvements to Italian strings

tobiatesan:


sev-:
I'm merging it now, as it still cleanly merges, but these days it's best to go via https://translations.scummvm.org/

PR #885, opened 2017-01-05
BUILD: Speed up Travis-CI builds

csnover:
container-based infrastructure (by sudo: false) starts and runs
faster:
https://docs.travis-ci.com/user/migrating-from-legacy/

gcc and make are preinstalled by Travis-CI for C++ language
containers, so do not need to be listed as dependencies:
https://docs.travis-ci.com/user/trusty-ci-environment/#Compilers-%26-Build-toolchain

ccache caching prevents wasting time rebuilding unchanged code:
https://docs.travis-ci.com/user/caching#ccache-cache

This should work OK, though there is a [bug](https://github.com/travis-ci/travis-yaml/issues/58) in travis lint which means that there is a slight chance this change might break something. I am happy to monitor and unbreak the build if that happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language:
   - cpp
 
-sudo: required
+sudo: false
+cache: ccache
 
 addons:
   apt:
     packages:
-    - g++ make
     - libsdl2-dev
     - libsdl2-net-dev
     - libcurl4-openssl-dev


sev-:
Cool, Travis seems to be happy. Merging.

csnover:
Unfortunately looks like the compile time saved by ccache & quick start is perfectly balanced by the time it takes to send and receive the cache data. Travis say that they are working on backgrounding the cache transmission in the future (that is the slowest part), but right now it is treated as part of the job, so there is no decrease in build time. Plus, I’ve started seeing [this bug](https://github.com/travis-ci/travis-ci/issues/7103), so perhaps this should be backed out for the moment and tried again later, at least once travis-ci/travis-ci#7103 is fixed.

PR #886, opened 2017-01-05
SCI: Use strnlen instead of strlen to avoid buffer overflows

csnover:
Nothing terribly surprising here I hope, and I have been running with this patch for a long time with no issues, but since it makes changes to SCI16 I wanted to make sure everyone feels OK with these changes.

### DIFF
--- a/engines/sci/parser/vocabulary.cpp
+++ b/engines/sci/parser/vocabulary.cpp
@@ -207,8 +207,12 @@ bool Vocabulary::loadSuffixes() {
 	while ((seeker < resource->size - 1) && (resource->data[seeker + 1] != 0xff)) {
 		suffix_t suffix;
 
+		int maxSize = resource->size - seeker;
 		suffix.alt_suffix = (const char *)resource->data + seeker;
-		suffix.alt_suffix_length = Common::strnlen(suffix.alt_suffix, resource->size - seeker);
+		suffix.alt_suffix_length = Common::strnlen(suffix.alt_suffix, maxSize);
+		if (suffix.alt_suffix_length == maxSize) {
+			error("Vocabulary alt appears truncated for suffix %d in resource %d at %u", _parserSuffixes.size(), resource->getNumber(), seeker);
+		}
 		seeker += suffix.alt_suffix_length + 1; // Hit end of string
 
 		suffix.result_class = (int16)READ_BE_UINT16(resource->data + seeker);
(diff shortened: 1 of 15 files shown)


engines/sci/parser/vocabulary.cpp
```diff
@@ -208,7 +208,7 @@ bool Vocabulary::loadSuffixes() {
 		suffix_t suffix;
 
 		suffix.alt_suffix = (const char *)resource->data + seeker;
-		suffix.alt_suffix_length = strlen(suffix.alt_suffix);
+		suffix.alt_suffix_length = Common::strnlen(suffix.alt_suffix, resource->size - seeker);
```
m-kiewitz: We should maybe better error() out in case of this here when no terminating NUL is found.
Unless there is a game, that actually has such corrupted resources.

engines/sci/parser/vocabulary.cpp
```diff
@@ -288,12 +288,12 @@ bool Vocabulary::loadAltInputs() {
 		AltInput t;
 		t._input = data;
 
-		uint32 l = strlen(data);
+		uint32 l = Common::strnlen(data, data_end - data);
```
m-kiewitz: the strlen()s inside checkAltInput() inside vocabulary.cpp aren't changed too?

As I said - I would prefer to error() out on corrupted resources. Because those would cause issues down the line anyway (unless there are games w/ actual corrupted resources that are still playable). That way the user would know about it immediately.

csnover: Looks like I managed to miss quite a few places using strlen, not sure what happened there. I’ll get the other ones converted too and update this PR. I understand the rationale about wanting to error out too, I’m happy to make that change and see what happens.

csnover:
I added errors in vocabulary.cpp, and warnings in MessageReader instead of errors since truncated messages there should still be serviceable. I also updated the other places I saw where strlen was being used insecurely, changing GfxControl16, GfxText16, kFormat to use more Common::String. There are a couple places where strings are guaranteed to be well-formed because they are either coming from `fgets` or `.data`, and in those cases strlen is still used as-is.

bluegr:
Excellent work! :)

I agree with the rationale of throwing warnings instead of errors for game messages, as these are recoverable. I've only spotted a change where a +1 has been removed from strlen. Other than that, this is great work, +1 from me :)

wjp:
Looks good to me.

One potential addition would be strengthening the checks in the `kArrayTrimLeft` and `kArrayTrimCenter` cases in `SciArray::trim()`, since those would already run out of bounds before reaching the `strnlen`.

engines/sci/parser/vocabulary.cpp
```diff
@@ -288,12 +296,20 @@ bool Vocabulary::loadAltInputs() {
 		AltInput t;
 		t._input = data;
 
+		uint32 maxSize = data_end - data;
 		uint32 l = Common::strnlen(data, data_end - data);
```
m-kiewitz: You could use maxSize here instead of "data_end - data"

csnover:
Landed b1c3332fddbb16838f1a654d6fe35ddbe09bd051 and 10d97ce37947362e6fa7161cd85221f59ca49490. Thanks!

PR #887, opened 2017-01-08
DIRECTOR: Support for D4 cast and images

stevenhoefel:
Initial commit of work to get D4 cast members loading. Also work on
Sounds and Cursors. The Apartment and HandV now render much better.


sev-:
Supercool, thanks!

PR #888, opened 2017-01-09
IMAGE: Return 24bpp pixel format for Indeo3

csnover:
The YUVToRGBManager::convert410 function used by Indeo3 decoder requires at least 16bpp graphics mode. Engines that normally run at 8bpp but switch rendering modes for high-colour video (like SCI) rely on getting accurate pixel format information from the video player in order to mode switch to the best output mode for the video being played.

Indeo3’s current behaviour is to return the system screen pixel format, which means SCI engine is unable to play back Indeo3 video (GK2A.AVI from GK2 demo) because it is told to remain in its 8bpp mode.

~~I don’t know if not returning the system pixel format will cause problems for other engines that run in 16bpp mode, hence this PR. A potentially safer alternative would be to return the system pixel format unless the system pixel format is below 16bpp, at which point the 24bpp format could be used instead.~~ Now the PR follows the same pattern as other codecs that are encapsulated in AVI.

### DIFF
--- a/image/codecs/codec.cpp
+++ b/image/codecs/codec.cpp
@@ -210,7 +210,7 @@ Codec *createBitmapCodec(uint32 tag, int width, int height, int bitsPerPixel) {
 	case MKTAG('c','v','i','d'):
 		return new CinepakDecoder(bitsPerPixel);
 	case MKTAG('I','V','3','2'):
-		return new Indeo3Decoder(width, height);
+		return new Indeo3Decoder(width, height, bitsPerPixel);
 	case MKTAG('I', 'V', '4', '1'):
 	case MKTAG('I', 'V', '4', '2'):
 		return new Indeo4Decoder(width, height, bitsPerPixel);
--- a/image/codecs/indeo3.cpp
+++ b/image/codecs/indeo3.cpp
@@ -40,11 +40,24 @@
 
 namespace Image {
 
-Indeo3Decoder::Indeo3Decoder(uint16 width, uint16 height) : _ModPred(0), _corrector_type(0) {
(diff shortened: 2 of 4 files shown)


bgK:
The VideoDecoder class allows client code to override the output pixel format by calling VideoDecoder::setDefaultHighColorFormat. Maybe the proper fix is to make that work?

csnover:
Thanks for the pointer. That does sound more appropriate. I’ll try updating Indeo3 to use that instead and see how it goes.

dreammaster:
If you look at the constructor for IndeoDecoderBase in image/codecs/indeo/indeo.cpp, you'll see that I set up different pixel formats depending on what bit depth is specified for the movie itself, rather than the screen. This class is used as a base for Indeo 4 and 5, but not 3. You could do something similar, though.

csnover:
@dreammaster Yeah! I realised as I started looking more into how bit depths are handled by other codecs that the others receive a bit depth in the constructor. Giving Indeo3 the same information fixes the problem in the same manner as other codecs. I don’t love the code duplication, but I don’t know if it really makes sense to create another IndeoDecoderBase subclass just for these few lines of pixel format/surface creation code. Does it seem worth it to you?

csnover:
Oh, and since this codec is encapsulated by VMD too, I guess I need to check to make sure Urban Runner does not explode, since I suspect it relies on the codec giving 16bpp video.

csnover:
(I did break Urban Runner. This PR has been updated with a new patch.)

dreammaster:
No, it probably doesn't make sense to create another entire base class for
all three Indeo decoders to just duplicate those few lines. The way you've
done in your new commit is fine.

Paul.

csnover:
Thanks @bgK and @dreammaster for your feedback!

PR #889, opened 2017-01-15
SCI: Expand workaround for script bug #3537232

tsoliman:
In Space Quest 4 Floppy, I noticed that the time pod time travel sequences don't animate and sort of freeze.
I was able to fix this by applying the same sleep workaround as was being done for the intro by @bluegr 

Not sure if this is the right way to fix it, but it does fix it.

I also want to know if it is a good idea to maybe remove the room# condition. The rationale being that maybe different versions of the game might have different room numbers for the time pod?


tsoliman:
I just found the same problem as in the intro (sequence too short) in the ending cutscene time rip (room 21)

bluegr:
Nice work :)

It seems that this was an issue with the scripts in the floppy version of SQ4. Perhaps it would make more sense to enable this workaround in all rooms for that game version, so that it's active in all palette cycling calls?

tsoliman:
Ok I've done so and will just push it to master.

PR #891, opened 2017-01-19
BLADERUNNER: Add detection for German version

rootfather:
I received my original copy of the German "Blade Runner" release today, so here is the detection entry.

At the moment, it seems that the German release is unsupported by the engine. I did a full installation from my original CDs on a Windows XP VM and copied the full "BLADE" directory on the virtual HDD to my main machine.

After starting the game, it fails with the following error message:
```
Opening hashed: MUSIC.MIX
Opening hashed: SFX.MIX
Opening hashed: SPCHSFX.TLK
getResource: Resource ACTORS.TRE not found.
```

I added the detection information as a Pull Request, because I don't know if you want to merge the detection entry at the current state.

### DIFF
--- a/engines/bladerunner/detection_tables.h
+++ b/engines/bladerunner/detection_tables.h
@@ -39,6 +39,20 @@ static const ADGameDescription gameDescriptions[] = {
 		ADGF_NO_FLAGS,
 		GUIO0()
 	},
+	
+	// BladeRunner (German)
+	{
+		"bladerunner",
+		0,
+		{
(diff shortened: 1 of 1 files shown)


peterkohaut:
I don't have german version of this game yet, so i can't check it right now.
But we don't have language detection implemented, and it is needed, because last character in that ".tre"  is code for language - first letter of value of "Language=" in blade.ini

sev-:
I'll merge it, and the actual game will be looked upon later.

PR #893, opened 2017-01-27
Fix Bug : #7172

batman97:
Engine Sword 25: The bug "https://bugs.scummvm.org/ticket/7172" is fixed. 

#Bug's description: GCC's -Wint-to-pointer-cast (on amd64) complaints about the casting of an integer value into a pointer type in two places within Lua.
#reinterpret_cast<type> (expr) is used to fix it.

### DIFF
--- a/engines/sword25/util/lua_unpersist.cpp
+++ b/engines/sword25/util/lua_unpersist.cpp
@@ -118,7 +118,7 @@ static void registerObjectInIndexTable(UnSerializationInfo *info, int index) {
 	// Make sure there is enough room on the stack
 	lua_checkstack(info->luaState, 2);
 
-	lua_pushlightuserdata(info->luaState, (void *)index);
+	lua_pushlightuserdata(info->luaState, reinterpret_cast<void *>(index));
 	// >>>>> permTbl indexTbl ...... obj index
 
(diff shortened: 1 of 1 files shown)


wjp:
Could you document why this is in fact the right solution?

(Also please note our commit guidelines at http://wiki.scummvm.org/index.php/Commit_Guidelines )

batman97:
I think it is one of the possible ways to fix this bug.
The reinterpret_cast command allows casting from pointer to an integer type and vice versa.

wjp:
The question is if this makes the code behave *correctly* (not just warning-free) on all platforms. That requires a description about what the goal of this cast is, and how the result is used.

wjp:
Thank you for looking at this. I've done some research into this myself now, and it turns out Pluto itself had fixed this warning 7 years ago ( https://github.com/hoelzro/pluto/commit/a56f3affab84b6c765743564e918227ebf2069fb ), but a pluto alternative called Eris has approached this in a cleaner way IMO. I've adopted the Eris approach in PR #894.

sev-:
Closing this in favor of #894. In any case there is no feedback from the author and the commit log message is not following our commit guidelines.

PR #894, opened 2017-01-28
SWORD25: Use direct integer index into indexTbl

wjp:
This matches the behaviour of the Pluto rewrite Eris, and avoids
casting the integer indices to void *.

I verified this code by running this method in parallel with the current
lightuserdata/void* approach and asserting the retrieved objects were
identical when loading a savegame.

This fixes bug #7172 in an alternative way from #893.


wjp:
@RichieSams Does this seem sensible to you?

RichieSams:
Sorry for the delay in responding; I was at the PAX South Convention this weekend.

As for the change, we had done a similar change in https://github.com/scummvm/scummvm/pull/557. That is, replacing lua_pushlightuserdata with an actual int. 
Everything looks fine to merge by me.

sev-:
Thanks, merging.

PR #895, opened 2017-02-05
GUI: Add three new options for volume slider controls

csnover:
GUIO_NOSPEECHVOLUME can be used for games that allow toggling of speech but do not provide the ability for users to control speech volume. (e.g. GK2, which has a patch that can enable subtitles, but the game only allows control of music volume.)

GUIO_LINKMUSICTOSFX and GUI_LINKSPEECHTOSFX can be used for games that allow control of music, sfx, and speech in combinations, like games that provide control of digital audio separately from MIDI, or games that only control all three audio types through a single volume control. (e.g. Phant1 and KQ7, respectively.)

(And, before anyone asks: no, it’s not possible to cleanly split audio types in these games, since Robots sometimes contain speech or sound effects or both, and some games like GK2 put speech into non-Audio36 resources.)

### DIFF
--- a/common/gui_options.cpp
+++ b/common/gui_options.cpp
@@ -29,6 +29,8 @@ namespace Common {
 
 const struct GameOpt {
 	const char *option;
+	// Each description must be a unique identifier not containing a substring
+	// of any other description
 	const char *desc;
 } g_gameOptions[] = {
 	{ GUIO_NOSUBTITLES,  "sndNoSubs" },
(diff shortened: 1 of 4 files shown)


criezy:
This looks good to me.
And drascula is another engine that could benefit from this since it doesn't separate Speech and SFX (it has three controls: Master, Speech/SFX and Music - although currently the master volume is not correctly handled).

sev-:
Exactly! Thanks, merging.

PR #896, opened 2017-02-06
Create project fixes

criezy:
This PR contains a collection of fixes for issues I encountered using create_project. The commits are independent from each others so we could decide to only keep some of them.

I am not familiar with create_project. I don't often use it and I had never looked at its source code before today. So don't hesitate to point issues or suggest better ways to fix some of the issues.

All the fixes result from looking at an issue recently posted in the iPhone forum (http://forums.scummvm.org/viewtopic.php?t=13965&start=120#83734). Compiling the current master code requires using `--disable-engine=fullpipe`. The first commit fixes this specific error (that also occur when compiling for OS X and I assume Windows). The other commits result from issues I noticed when testing this (for example that using --disable-16bit did not disable engines that require 16 bit color such as fulllpipe or riven, while using configure does it).


sev-:
That is great! Thanks, merging as is.

PR #897, opened 2017-02-07
DREAMWEB: fixed detection of Italian CD release

darioscarpa:
The Italian CD release of Dreamweb (available on
https://www.scummvm.org/games/) was not properly
detected (the floppy version was detected, so
speech stayed disabled).
Added a proper entry to fix the issue.

### DIFF
--- a/engines/dreamweb/detection_tables.h
+++ b/engines/dreamweb/detection_tables.h
@@ -244,6 +244,24 @@ static const DreamWebGameDescription gameDescriptions[] = {
 		},
 	},
 
+	// Italian CD release
+	{
+		{
+			"dreamweb",
+			"CD",
+			{
+				{"dreamweb.exe", 0, "44d1708535cdb863b9cca372ad0b05dd", 65370},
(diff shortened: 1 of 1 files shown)


digitall:
@darioscarpa : Thanks for providing this. As there are no other detection entries for Italian CD version and we have this in the Freeware downloads page at https://www.scummvm.org/games/#dreamweb I assume it is that version that you have tested with? i.e. https://www.scummvm.org/frs/extras/Dreamweb/dreamweb-cd-it-1.1.zip

Overall, this looks like a good and simple change which should get merged quickly as no issue.

darioscarpa:
@digitall : Thanks for your quick feedback! Yes, that's the version I have tested with (succesfully). I also checked that I didn't break the detection of the IT floppy release (it was the only potential issue I could think of, being the hashes for the dreamweb.r00 and dreamweb.r02 the same for both versions, and not having looked into the detection algorithm).

digitall:
As a single self-contained commit, have used the newer "rebase and merge" option here to apply to the master branch without a merge commit. Thanks again for the patch. Closing.

PR #899, opened 2017-02-10
CRYO: Move all static engine data into cryo.dat

bluegr:
This series of commits repurposes the led.dat data file, changing its name to cryo.dat, and moves all of the static engine data into it. Since this pretty much changes the original concept of this file, I made a relevant pull request containing all the associated changes.

Furthermore, these commits add versioning to this file, so that future games supported by this engine will contain all of their static data in this data file.

PS Some of the commits in this pull request were mistakenly landed onto master. Apologies once again - these commits have been reverted


engines/cryo/eden.cpp
```diff
@@ -4729,29 +4729,30 @@ void EdenGame::convertMacToPC() {
...
+		headerId[8] = '\0';
+		if (strcmp(headerId, "CRYODATA"))
+			error("Invalid aux data file");
 
-			if (dataSize != kNumIcons * sizeof(Icon) + kNumRooms * sizeof(Room))
-				error("Mismatching data in aux data file");
+		if (f.readByte() != CRYO_DAT_VER)
+			error("Incorrect aux data version");
 
+		if (dataSize != kNumIcons * sizeof(Icon) + kNumRooms * sizeof(Room))
```
Retro-Junk: This check is very sketchy and may be incompatible across different platforms because sizeof(Icon) and Room may vary, even though they are declared as a byte-packed structures currently. It's probably better to use fixed constants here.

devtools/create_cryo/eden_static.h
```diff
@@ -20,12 +20,10 @@
...
-#include "cryo/cryolib.h"
+#pragma once
+#include "eden.h"
 
-namespace Cryo {
-
-Follower followerList[] = {
+Follower followerList[15] = {
 //            char,                 X,  sx, sy,  ex,  ey,bank,
 	{ PersonId::pidGregor,          5, 211,  9, 320, 176, 228,   0,  0 },
```
criezy: You will get an error here on Linux with gcc. This is a common issue I have seen  lot from developers using Visual Studio as it accepts this code, but using EnumName::EnumValue is not valid C++. So you should remove the PersonId:: prefix on all those here.

Note that the same code currently in engines/cryo/staticdata.cpp compiles because the enum definition in engines/cryo/defs.h is enclosed in a namespace of the same name, which is not the case for the one you added to devtools/create_cryo/eden.h. Adding the namespace in eden.h is another way to fix it.

And you will also need to fix the same errors below for the other enums (PersonMask, Object, Areas, AreaFlag).

criezy:
The cryo.dat (or led.dat currently) will need to be added to dist/engine-data and listed in DIST_FILES_ENGINEDATA (and in some platform specific install scripts).

This is not specific to this pull request, and will be needed even if this PR is not accepted since the file is already needed to run the engine. I mention it here because I noticed the issue while reviewing this PR.

I will let the engines authors decide if the PR hold be accepted (after the enum issue has been fixed) or not.

engines/cryo/eden.cpp
bluegr: There are other places where sizeof() is used, e.g. EdenGame::loadIconFile() and EdenGame::loadRoomFile()

Should these also be adapted accordingly?

devtools/create_cryo/eden_static.h
bluegr: Thanks for that, the namespaces around the enums looked strange, but now I understood why they are there. Strangely, this syntax for enums works fine for mingw64, so I assume that newer versions of GCC accept it. Anyway, I've wrapped all enums around namespaces, like the rest of the code

engines/cryo/eden.cpp
bluegr: I've removed all places where sizeof(struct) is used.

bluegr:
Thanks for the feedback! I've made some changes based on the comments. I've also added cryo.dat to engine-data.

bluegr:
Also, since @wjp expressed some concerns on IRC that a 1-byte version for cryo.dat isn't enough

    <wjp> one byte versioning, whee :-)

I've changed this to be a 32-bit integer

devtools/create_cryo/eden_static.h
criezy: Yes, that became valid in C++11, so recent enough version of g++ do accept it.

Retro-Junk:
No objections on my side. Thanks for your help.

bluegr:
Many thanks for the reviews!

Merging

PR #900, opened 2017-02-10
Fix Bug #6996 (Fixed jerky/laggy analog joystick mouse control)

rsn8887:
Fixes https://bugs.scummvm.org/ticket/6996

This small change to the source fixes the "stopping, jerky" analog joystick mouse pointer control. In the original source, the mouse pointer was only updated when the analog joystick axes _changed_. Now it is updated also when the joystick axes stay at the same, unchanged values. This fixes the "stop and go" behaviour of the pointer when it is controlled by an analog stick. 

This bug was experienced by me on the Vita, on Android (Nvidia Shield Portable), and reported by several users on Raspberry Pi (RetroPie) when they used their controllers to control the mousepointer instead of a real mouse (or keyboard).

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -179,11 +179,16 @@ void SdlEventSource::processMouseEvent(Common::Event &event, int x, int y) {
 	_km.y = y;
 }
 
-void SdlEventSource::handleKbdMouse() {
+void SdlEventSource::handleKbdMouse(Common::Event &event) {
+
 	// Skip recording of these events
 	uint32 curTime = g_system->getMillis(true);
 
 	if (curTime >= _km.last_time + _km.delay_time) {
+
+		int16 oldKmX = _km.x;
+		int16 oldKmY = _km.y;
+
 		_km.last_time = curTime;
 		if (_km.x_down_count == 1) {
 			_km.x_down_time = curTime;
@@ -247,6 +252,11 @@ void SdlEventSource::handleKbdMouse() {
 
 			if (_graphicsManager) {
 				_graphicsManager->getWindow()->warpMouseInWindow((Uint16)_km.x, (Uint16)_km.y);
+			}
+
+			if (_km.x != oldKmX || _km.y != oldKmY) {
+				event.type = Common::EVENT_MOUSEMOVE;
+				processMouseEvent(event, _km.x, _km.y);
 			}
 		}
 	}
@@ -425,7 +435,8 @@ Common::KeyCode SdlEventSource::SDLToOSystemKeycode(const SDLKey key) {
 }
 
 bool SdlEventSource::pollEvent(Common::Event &event) {
-	handleKbdMouse();
+	handleKbdMouse(event);
+
 
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	// In case we still need to send a key up event for a key down from a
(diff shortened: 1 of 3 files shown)


criezy:
Please have a look at the commit message formatting on http://wiki.scummvm.org/index.php/Commit_Guidelines
The commit message should look something like the following:
```
SDL: Fixed jerky/laggy analog joystick mouse control

This fixes bug 6996: Android: Mouse pointer control with
analog joystick is unusable
```

The change you propose looks fine, although I don't understand why this is needed. Calling warpMouseInWindow() results in a call to either SDL_WarpMouseInWindow (for SDL2) or SDL_WarpMouse (for SDL 1.2), which generates a SDL mouse motion event (at least if we believe the SDL documentation). So this should eventually result in processMouseEvent() being called (from pollEvent > dispatchSDLEvent > handleMouseMotion). Thus I am not sure why this additional call to processMouseEvent() is needed.

But since you indicate this change fixes the bug, it seems to me we should accept it. I will wait a bit to see if others have comment on this, and you should also fix your commit message before I can accept this.

For the record can you indicate here which version of SDL you are using?

rsn8887:
I am using SDL2

criezy:
Yes, SDL_WarpMouseInWindow() is indeed not called when the window doesn't have the mouse focus. And maybe my assumption that it has the focus was wrong.

There is one more change to do before this can be merged: the wince backend does call `handleKbdMouse()` in one place in a derived class (in backends/events/wincesdl/wincesdl-events.cpp). Please change this call to pass the event and squash the change with your commit so that it does not break compilation.

rsn8887:
Ok I made that change and squashed it with the commit. Please tell me if you need any other changes to be able to merge this. Thanks!

criezy:
I merged this change. Thank you for your contribution.

wjp:
This isn't quite right. `handleKbdMouse` now fills an `event` struct for an artificial `EVENT_MOUSEMOVE` event, but since `pollEvent` doesn't actually process or return this event, it gets ignored.

It seems like the expected path is that `warpMouseInWindow` generates a `MOUSEMOVE` event, but this apparently doesn't happen for you. So we could either see why not and if that could be changed without negative side effects, or we could have `pollEvent` actually return an artificial `EVENT_MOUSEMOVE` event itself if `handleKbdMouse` generates one (and check that doesn't have negative side effects too).

criezy:
Right. I completely missed that point. We could indeed have handleKbMouse() return a bool, with true if it calls processMouseEvent(), and in that case return from pollEvent() instead of continuing. Hopefully the time delay check in handleKbMouse() would mean SDL events get processed as well sometimes. That would need to be checked though (especially on slow systems).

It might be better first to check why we don't get a mouse move event from SDL. Is that because of that focus check in SdlWindow::warpMouseInWindow()? I wondered two days ago why it was added and why only for SDL2. Maybe sev who added that 6 months ago (in 23e2a61d4) remembers?

wjp:
I think that was because we were stealing back focus when the mouse is outside of the window. In particular SCI would do this when trying to restrict mouse movement to only part of the window.

wjp:
Semi-related: it also seems bad that `SdlEventSource::handleJoyAxisMotion` can return true without setting `event.type`

rsn8887:
I made a new pull request that fixes the remaining problems (handleKbdMouse now returns a bool etc) with analog joystick mouse pointer control.

rsn8887:
The new PR is https://github.com/scummvm/scummvm/pull/906

PR #903, opened 2017-02-14
SCUMM: Fix bug #4112 (MANIAC: Power never returns)

segrax:
Hi,

This fix is for a scripting oversight, 
have added the script analysis to the ticket: https://bugs.scummvm.org/ticket/4111

in the event Dr. Fred turns the power off and the player enters the lab, the script responsible for turning the power back on is stopped, resulting in the power never being turned back on.

This fix resolves it by setting the 'circuit breaker' object to on when termination of the script occurs

I'm not sure if it was the right choice, but decided to add 2 macros for the MM workarounds, as the continual IF/ELSE combinations with the various script/object numbers was rather unfriendly, have also applied it to the surrounding code... If this is ok, I'll back apply it to PR #902

### DIFF
--- a/engines/scumm/script_v2.cpp
+++ b/engines/scumm/script_v2.cpp
@@ -1192,7 +1192,28 @@ void ScummEngine_v2::o2_startScript() {
 	runScript(script, 0, 0, 0);
 }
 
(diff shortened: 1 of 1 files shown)


sev-:
Great work!

PR #905, opened 2017-02-15
ANDROIDSDL: added tab Control in main Options menu for switching some features

lubomyr:
Added tab 'Control' for switching some features in main 'Options' menu.
This tab is visible only if one of features is present. Currently it uses only in Android SDL port.
Implemented 3 checkboxes for switching:
1) Show On-screen control
2) Touchpad mouse mode
3) Swap Menu and Back buttons
screenshot https://dl.dropboxusercontent.com/u/47797448/scummvm/device-2017-02-15-204251.png


backends/events/androidsdl/androidsdl-events.cpp
```diff
@@ -43,16 +42,9 @@ bool AndroidSdlEventSource::handleMouseButtonDown(SDL_Event &ev, Common::Event &
...
-		if (show_onscreen == 0) {
-			SDL_ANDROID_SetScreenKeyboardShown(0);
-			show_onscreen++;
-		} else if (show_onscreen==1) {
-			SDL_ANDROID_SetScreenKeyboardShown(1);
-			show_onscreen++;
-		}
-		if (show_onscreen == 2)
-			show_onscreen = 0;
+		static bool show_onscreen = g_system->getFeatureState(OSystem::kFeatureOnScreenControl);
```
bluegr: I can't think of a good reason why this should be static

backends/platform/androidsdl/androidsdl-sdl.cpp
```diff
@@ -42,7 +44,101 @@ void OSystem_ANDROIDSDL::initBackend() {
 
 	if (!ConfMan.hasKey("gfx_mode"))
 		ConfMan.set("gfx_mode", "2x");
+	
+	if (!ConfMan.hasKey("swap_menu_and_back"))
```
bluegr: Hm... this sounds a bit confusing. Swap menu and back?

backends/platform/androidsdl/androidsdl-sdl.cpp
```diff
@@ -42,7 +44,101 @@ void OSystem_ANDROIDSDL::initBackend() {
...
+		swapMenuAndBackButtons(ConfMan.getBool("swap_menu_and_back"));
+	
+	if (!ConfMan.hasKey("touchpad_mouse_mode")) {
+		const bool enable = (SDL_ANDROID_GetMouseEmulationMode() == 0) ? false : true;
+		ConfMan.setBool("touchpad_mouse_mode", enable);
+	} else
+		touchpadMode(ConfMan.getBool("touchpad_mouse_mode"));
+	
+	if (!ConfMan.hasKey("onscreen_control")) {
+		const bool enable = (SDL_ANDROID_GetScreenKeyboardShown() == 0) ? false : true;
```
bluegr: You can just write this like:

    const bool enable = SDL_ANDROID_GetScreenKeyboardShown();

backends/platform/androidsdl/androidsdl-sdl.cpp
lubomyr: it's only short expression of 'swap menu and back buttons'
A lot of new device don't have physical menu button and for call menu i used back button.
But i received some feedback where users ask about possiblity to use menu (if it present on device) as menu and back as skip. That was implemented. Now it possible to switch it via main 'Options' menu.

backends/events/androidsdl/androidsdl-events.cpp
lubomyr: static - not neccesarily here.
No needs in const, because this variable changes value in next line.

bluegr: You could use const here, and change the next two lines to:

    g_system->setFeatureState(OSystem::kFeatureOnScreenControl, !show_onscreen);

lubomyr: Ok, I agree

sev-:
Looks good now, thanks, merging

PR #906, opened 2017-02-17
SDL: More fixes to jerky analog joystick pointer

rsn8887:
For some engines such as Dreamweb that a previous fix did not apply to, this fixes bug 6996: Android: Mouse pointer control with analog joystick is unusable.

The improvements are not limited to Android but apply to any host system that uses the SDL backend and allows analog joystick mouse pointer movement (Vita, Raspberry Pi etc.).

This PR fixes the remaining problems with analog joystick mouse pointer control.

Most importantly 

- handleKbdMouse now returns a bool so that the MouseEvents generated by it are not dropped anymore

Other analog joystick problems that are fixed in this PR are:

- increased resolution of analog joystick mouse speeds. It had only ~16 speed increments instead of the max possible 32768.

- improved smoothness of pointer movement. The position of the joystick-controlled mouse pointer is now updated at ~60 fps instead of the original, stuttery ~40 fps.

- improved the joystick deadzone by replacing the axial ("naive") deadzone with a proper radial scaled implementation. See http://www.third-helix.com/2013/04/12/doing-thumbstick-dead-zones-right.html for the differences between the two deadzone methods.

With the changes in this PR, the analog mouse pointer behavior is much improved in all engines.

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -270,8 +270,14 @@ bool SdlEventSource::handleKbdMouse(Common::Event &event) {
 			}
 
 			if (_km.x != oldKmX || _km.y != oldKmY) {
+				// keep hi-res coordinates since 
+				// processMouseEvent will overwrite them with lo-res numbers
+				oldKmX = _km.x;
+				oldKmY = _km.y;
 				event.type = Common::EVENT_MOUSEMOVE;
 				processMouseEvent(event, _km.x / MULTIPLIER, _km.y / MULTIPLIER);
+				_km.x = oldKmX;
+				_km.y = oldKmY;
 				return true;
 			}
 		}
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -179,8 +179,8 @@ void SdlEventSource::processMouseEvent(Common::Event &event, int x, int y) {
 	}
 
 	// Update the "keyboard mouse" coords
-	_km.x = x * _km.multiplier;
-	_km.y = y * _km.multiplier;
+	_km.x = x * MULTIPLIER;
+	_km.y = y * MULTIPLIER;
 }
 
 bool SdlEventSource::handleKbdMouse(Common::Event &event) {
@@ -207,36 +207,36 @@ bool SdlEventSource::handleKbdMouse(Common::Event &event) {
 			if (_km.x_down_count) {
 				if (curTime > _km.x_down_time + 300) {
 					if (_km.x_vel > 0)
-						_km.x_vel += _km.multiplier;
+						_km.x_vel += MULTIPLIER;
 					else
-						_km.x_vel -= _km.multiplier;
+						_km.x_vel -= MULTIPLIER;
 				} else if (curTime > _km.x_down_time + 200) {
 					if (_km.x_vel > 0)
-						_km.x_vel = 5 * _km.multiplier;
+						_km.x_vel = 5 * MULTIPLIER;
 					else
-						_km.x_vel = -5 * _km.multiplier;
+						_km.x_vel = -5 * MULTIPLIER;
 				}
 			}
 			if (_km.y_down_count) {
 				if (curTime > _km.y_down_time + 300) {
 					if (_km.y_vel > 0)
-						_km.y_vel += _km.multiplier;
+						_km.y_vel += MULTIPLIER;
 					else
-						_km.y_vel -= _km.multiplier;
+						_km.y_vel -= MULTIPLIER;
 				} else if (curTime > _km.y_down_time + 200) {
 					if (_km.y_vel > 0)
-						_km.y_vel = 5 * _km.multiplier;
+						_km.y_vel = 5 * MULTIPLIER;
 					else
-						_km.y_vel = -5 * _km.multiplier;
+						_km.y_vel = -5 * MULTIPLIER;
 				}
 			}
 
-			// The modifier key makes the mouse movement slower
-			// The extra factor of delay/25 makes velocities 
-			// independent of kbdMouse update rate
-			// all velovities were originally chosen
-			// at a delay of 25, so that is the reference
-			// operator order is important to avoid overflow
+			// - The modifier key makes the mouse movement slower
+			// - The extra factor "delay/25" ensures velocities 
+			// are independent of the kbdMouse update rate
+			// - all velocities were originally chosen
+			// at a delay of 25, so that is the reference used here
+			// - note: operator order is important to avoid overflow
 			if (_km.modifier) {
 				_km.x += ((_km.x_vel / 10) * ((int16)_km.delay_time)) / 25;
 				_km.y += ((_km.y_vel / 10) * ((int16)_km.delay_time)) / 25;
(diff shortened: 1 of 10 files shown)


rsn8887:
Note: I made this new pull request that fixes the remaining problems (most importantly handleKbdMouse now returns a bool) with analog joystick mouse pointer control. Other remaining problems that are fixed are 

-the ridiculously low resolution of analog joystick mouse movement (it had only 16 speed increments instead of the max possible 32768), 
- the non-smooth pointer movement that was only updating the analog pointer motion at 40 fps instead of ~60 fps, 
- the naive, axial, and hard-to-control dead zone implementation was replaced with a proper radial scaled joystick deadzone (http://www.third-helix.com/2013/04/12/doing-thumbstick-dead-zones-right.html) 

With that PR merged, the analog mouse pointer behavior should be finally acceptable.

backends/events/sdl/sdl-events.cpp
```diff
@@ -203,63 +207,70 @@ void SdlEventSource::handleKbdMouse(Common::Event &event) {
 			if (_km.x_down_count) {
 				if (curTime > _km.x_down_time + _km.delay_time * 12) {
 					if (_km.x_vel > 0)
-						_km.x_vel++;
+						_km.x_vel+=_km.multiplier;
```
sev-: It should be ```_km.x_vel += _km.multiplier;``` Pay attention to spaces around operator.

backends/events/sdl/sdl-events.cpp
```diff
@@ -203,63 +207,70 @@ void SdlEventSource::handleKbdMouse(Common::Event &event) {
...
-				_km.y_vel = 1;
+			} else if (_km.y > _km.y_max * _km.multiplier) {
+				_km.y = _km.y_max * _km.multiplier;
+				_km.y_vel = 1 * _km.multiplier;
 				_km.y_down_count = 1;
 			}
 
 			if (_graphicsManager) {
-				_graphicsManager->getWindow()->warpMouseInWindow((Uint16)_km.x, (Uint16)_km.y);
+				_graphicsManager->getWindow()->warpMouseInWindow((Uint16) (_km.x / _km.multiplier), (Uint16) (_km.y / _km.multiplier));
```
sev-: No space after the type cast, please.

backends/events/sdl/sdl-events.cpp
```diff
@@ -900,8 +938,10 @@ bool SdlEventSource::remapKey(SDL_Event &ev, Common::Event &event) {
 void SdlEventSource::resetKeyboardEmulation(int16 x_max, int16 y_max) {
 	_km.x_max = x_max;
 	_km.y_max = y_max;
-	_km.delay_time = 25;
+	_km.delay_time = 12;
```
sev-: Why did you make it twice faster?

backends/events/sdl/sdl-events.cpp
```diff
@@ -900,8 +938,10 @@ bool SdlEventSource::remapKey(SDL_Event &ev, Common::Event &event) {
...
 	_km.last_time = 0;
+	_km.multiplier = 16;
```
sev-: I don't see multiplier and modifier changed anywhere in the code. Am I missing something?

criezy:
I only had a quick look and will look at it in more details later today. I will however dump some thoughts I had two days ago following a remark from wjp: he noticed that SdlEventSource::handleJoyAxisMotion sometimes return true without setting the event type (when axis is in the dead zone and is set to 0). This is still the case in this PR.

My first though was that maybe it should return false just before calling processMouseEvent() when the axis is equal to 0.

My second though was that this function does not actually change the _km.x and _km.y value, as this is only done in handleKbMouse(), so why is it calling processMouseEvent(). My guess is that by creating an event and return true it ensures that pollEvent() is called again (as usually it is called in a while loop until there is no more event) and handleKbMouse() is thus called immediately after.

Another though related to this PR is that if pollEvent is not called more often than the `_km.delay_time` then after starting to move the mouse we never stop and no other event is processed. Can this happen on some slow systems?

So I was wondering if we should
- Change handleJoyAxisMotion to always return false (and never set the event type and call processMouseEvent)
- Move the call to handleKbMouse at the end of pollEvent instead of having it at the start.

That way handleKbMouse is called immediately after handleJoyAxisMotion in the same pollEvent call (if there is no other SDL event, otherwise on a following call to pollEvent). And it also ensures that we continue to process SDL events in priority, and in particular that we would stop the mouse if the joystick is moved back to the dead zone.

But this assume I didn't miss another good reason why handleJoyAxisMotion currently generates an event.

And overall I agree with sev that this looks like a good approach, but that maybe the multiplier should be defined as a constant.

wjp:
Regarding the timing: `EventDispatcher` always calls `pollEvent` in a tight loop until it returns false, so there shouldn't be a risk of this extra event crowding out others.

backends/events/sdl/sdl-events.cpp
rsn8887: multiplier: this should not be changed you are right. I will make a commit to define it as const. It is just used to get more fine-grained control over velocity, so that the mouse pointer velocity can be less than 1 pixel per frame without becoming exactly equal to zero.

modifier: I want to keep it open for future use. For example, in the forthcoming Vita port, the modifier is changed by holding down the R-Trigger on the gamepad. Any other gamepad-centric platform code could potentially make use of this modifier (e.g. the R-Trigger, so that the mouse is slowed when holding R+analog stick).

backends/events/sdl/sdl-events.cpp
rsn8887: will be fixed

backends/events/sdl/sdl-events.cpp
rsn8887: Original code updates joystick mouse at 40 fps which looks very rough to the eye. That was just  too slow. New code updates at 60 fps, which looks smooth to the eye. It is a visual difference like day and night. However it does make the mouse pointer twice as fast. I will fix this in next commit.

backends/events/sdl/sdl-events.cpp
rsn8887: ok will be done

backends/events/sdl/sdl-events.cpp
rsn8887: This makes the "unit of time" for velocities twice as small, e.g. velocities effectively became twice as fast.

I am fixing this by putting in an extra factor that will make velocities independent of _km.delay (as it should have always been), in the next commit.

rsn8887:
I made all the changes suggested: 

- Code formatting is correct
- Pointer speed issues are resolved
    update delay is still at 12 to ensure visually smooth ~60fps pointer motion, 
    but HandlekbMouse velocities are NOT twice as fast anymore, they are now independent of delay
-  handleJoyAxisMotion to always return false (and never set the event type and call processMouseEvent)
- The call to handleKbMouse is now at the end of pollEvent instead of having it at the start.

Everything works very nicely.

backends/events/sdl/sdl-events.cpp
```diff
@@ -205,40 +205,46 @@ bool SdlEventSource::handleKbdMouse(Common::Event &event) {
...
+			// operator order is important to avoid overflow
 			if (_km.modifier) {
-				_km.x += _km.x_vel / 10;
-				_km.y += _km.y_vel / 10;
+				_km.x += ((_km.x_vel / 10) * ((int16)_km.delay_time)) / 25;
+				_km.y += ((_km.y_vel / 10) * ((int16)_km.delay_time)) / 25;
 			} else {
-				_km.x += _km.x_vel;
-				_km.y += _km.y_vel;
+				_km.x += (_km.x_vel * ((int16)_km.delay_time)) / 25;
```
rsn8887: This ensures velocities are not twice as fast although delay is now 12 instead of 25

backends/events/sdl/sdl-events.cpp
rsn8887: Ok it is fixed, see last commit and my comment. the factor "delay/25" assures the pointer velocity is independent of delay, and will be the same as it was before this PR. But with a much smoother pointer movement.

backends/events/sdl/sdl-events.h
```diff
@@ -59,7 +59,9 @@ class SdlEventSource : public Common::EventSource {
 
 	struct KbdMouse {
 		int16 x, y, x_vel, y_vel, x_max, y_max, x_down_count, y_down_count;
+		static const int16 multiplier = 16;
```
sev-: Well, what I thinking that instead of ```_km.multiplier```, you could use ```#define MULTIPLIER 16``` Thoughts?

criezy: What I had in mind was actually using an enum:
```
enum {
	kKbdMouseMultiplier = 16
};
```

But a define or a const int is fine as well I think.

criezy:
Maybe you could squash the last two commits together. We don't really need a separate commit to fix typos in comments ;)

But otherwise I think this is good to be merged.

backends/events/sdl/sdl-events.h
rsn8887: I like the const int most. So I will keep it like that if you don't mind. I like having all the _km - relevant stuff together.

rsn8887:
I have made another change after I realized that very slow analog pointer motion was still capped at 1 pixel per delay interval.

This was because processMouseEvent overwrites the fine-grained hi-res _km.x coords with coarse-grained lo-res ones (basically modulo MULTIPLIER). A simple variable safe and restore fixes this.

backends/events/sdl/sdl-events.cpp
```diff
@@ -270,8 +270,14 @@ bool SdlEventSource::handleKbdMouse(Common::Event &event) {
 			}
 
 			if (_km.x != oldKmX || _km.y != oldKmY) {
+				// keep hi-res coordinates since 
+				// processMouseEvent will overwrite them with lo-res numbers
+				oldKmX = _km.x;
+				oldKmY = _km.y;
 				event.type = Common::EVENT_MOUSEMOVE;
 				processMouseEvent(event, _km.x / MULTIPLIER, _km.y / MULTIPLIER);
+				_km.x = oldKmX;
```
rsn8887: I have made another change after I realized that very slow analog pointer motion was still capped at 1 pixel per delay interval.

This was because processMouseEvent overwrites the fine-grained hi-res _km.x coords with coarse-grained lo-res ones (basically modulo MULTIPLIER). A simple variable safe and restore fixes this.

rsn8887:
It might be too late, but I have made another change after I realized that very slow analog pointer motion was still capped at 1 pixel per delay interval.

This was because processMouseEvent overwrites the fine-grained hi-res _km.x coords with coarse-grained lo-res ones (basically modulo MULTIPLIER). A simple variable safe and restore fixes this.

sev-:
Great work! Merging.

PR #907, opened 2017-02-20
SCUMM: MM V0: Fix two walk animation issues

segrax:
Hi,

This PR is for two animation issues,

1. During the cutscene in script-151 (Purple Tentacle chases Sandy), Sandy appears to slide as the walk animation begins durnig the same frame as she begins speaking, this sets the 'repeat' flag for all limbs. We fix this by specifically running the speakCheck function, if speaking has been enabled.

2. Some walkboxes contain a 1 pixel gap, which the original handles using a JMP (when we reach a situation with no valid walkbox), and then move the actor again. I believe on the C64 this is unnoticeable, as it  enough time passes for the interrupts which redraw the screen to fire. In ScummVM, the actor is only drawn once per frame, causing them to move 2 pixels or 'skip'

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -937,7 +937,6 @@ L2A33:;
 	}
 
 	if ((_moving & 0x0F) == 3) {
-L2C36:;
 		setTmpFromActor();
 
 		if (!_walkDirX) {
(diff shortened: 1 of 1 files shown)


sev-:
Thanks!

PR #908, opened 2017-02-22
SDL: Fix erratic analog pointer + control options

rsn8887:
Fixes erratic speeds in analog pointer motion
Implemented option to set analog/keyboard pointer speed
and control the analog joystick deadzone. The deadzone option appears
only if the build supports analog joystick (via JOY_ANALOG define)

- Even after PR#906, the analog joystick pointer sometimes moved at erratic speeds that were way too fast. The reason was that I used the scaled velocity _km.vel_x or _km.vel_y for the "missing" axis that was not part of the JOYAXISEVENT. The small differences amplified especially when the user kept the stick moving in a circle or in certain diagonal directions. I fixed this by introducing proper variables, called _km.joy_x and _km.joy_y for the last known, unscaled, axis readouts, and using only those values to calculate _km.vel_x and _km.vel_y.

- To allow the user some control over the analog/keyboard mouse pointer speed and the joystick deadzone, I added new menu options. I used @lubomyr 's recent commit as a basis for two new control menu options in the SDL backend: Mouse Speed and Joystick Deadzone. They appear in the dedicated "Controls" tab only when the features kFeatureKbdMouseSpeed and or kFeatureJoystickDeadzone are set, which is currently only the case for the SDL backend and the new Vita backend. In the SDL backend the Joystick Deadzone setting is disabled unless "JOY_ANALOG" is defined in the source (which has been used in the source since before I started my work on this).

- I also fixed the small problem, exposed by the new "controls" tab, where the first visible tab in the menu would "move" it the user selects an option in a drop down menu.


gui/options.cpp
```diff
@@ -120,6 +122,15 @@ static const char *savePeriodLabels[] = { _s("Never"), _s("every 5 mins"), _s("e
 static const int savePeriodValues[] = { 0, 5 * 60, 10 * 60, 15 * 60, 30 * 60, -1 };
 static const char *outputRateLabels[] = { _s("<default>"), _s("8 kHz"), _s("11 kHz"), _s("22 kHz"), _s("44 kHz"), _s("48 kHz"), 0 };
 static const int outputRateValues[] = { 0, 8000, 11025, 22050, 44100, 48000, -1 };
+static const char *kbdMouseSpeedLabels[] = { _s("3"), _s("5"), _s("8"), _s("10"), _s("13"), _s("15"), _s("18"), _s("20"), 0 };
+static const int kbdMouseSpeedValues[] = { 0, 1, 2, 3, 4, 5, 6, 7, -1 };
+#ifdef PSP2
+static const char *shaderLabels[] = { _s("None"), _s("2xLCD"), _s("2xSharp"), _s("2xScan"), _s("2xAAA"), _s("2xScale"), 0 };
```
sev-: This must be reworked. Please remove backend-specific code from the common code. Instead, provide this list via new call, just as we do for the scalers.

gui/options.cpp
```diff
@@ -78,7 +78,9 @@ enum {
 	kExtraPathClearCmd		= 'clex',
 	kChoosePluginsDirCmd	= 'chpl',
 	kChooseThemeCmd			= 'chtf',
-	kUpdatesCheckCmd		= 'updc'
+	kUpdatesCheckCmd		= 'updc',
+	kKbdMouseSpeedChanged = 'kmsc',
```
sev-: It would bake sense to indent these new ones as the existing entries.

backends/events/sdl/sdl-events.cpp
```diff
@@ -856,21 +895,25 @@ bool SdlEventSource::handleJoyAxisMotion(SDL_Event &ev, Common::Event &event) {
 	}
 #ifdef JOY_ANALOG
 	// radial and scaled analog joystick deadzone
-	float analogX = (float) (_km.x_vel * vel_to_axis);
-	float analogY = (float) (_km.y_vel * vel_to_axis);
+	float analogX = (float) _km.joy_x;
```
sev-: Extra space after type conversion

gui/options.cpp
rsn8887: I removed all backend-specific code.

gui/options.cpp
rsn8887: Of course, done.

backends/events/sdl/sdl-events.cpp
rsn8887: Sorry for that. Done.

rsn8887:
I have made all the suggested changes and squashed everything together. It should be ready to merge now. I removed all instances of shader code. This is now purely an improvement of the analog joystick pointer control.

sev-:
Thanks! Merging.

wjp:
I'm not happy with how this changes the Options dialog. It makes the new Controls tab the default tab even on desktop machines, even though it's almost empty. It only shows a single 'Mouse Speed' option, that is not understandable without context.

wjp:
Additionally, the code is confused about if the numeric labels for kbdMouse are translatable or not. They're marked as translatable with `_s`, but then never actually looked up at run-time by `_`.

wjp:
The `Controls` tab also disappears when changing video mode from within the Options dialog.

rsn8887:
I will open another PR to fix these things

wjp:
The disappearing tab when changing video mode might be unrelated to your code, by the way. I'm investigating that one.

rsn8887:
Ok wjp thanks for the suggestions. Here is the new pull request:
https://github.com/scummvm/scummvm/pull/912

Note for the disappearing tab: before my changes it disappeared even when I just changed a popup selection (without even clicking apply), because it triggered a reflowLayout.

That's why I added setFirstTab and getFirstTab methods to the tab control and used them to restore the first tab after popup selection, since the popup selection will not change screen layout. But that cannot be done in general, because screen layout might change so drastically that the currently active tab might not be visible, and first active has to change.

backends/events/sdl/sdl-events.cpp
```diff
@@ -231,18 +231,59 @@ bool SdlEventSource::handleKbdMouse(Common::Event &event) {
...
+				case 0:
+					speedFactor = 100;
+					break;
+				// 0.5 speed
+				case 1:
+					speedFactor = 50;
+					break;
+				// 0.75 speed
+				case 2:
+					speedFactor = 37;
```
wjp: `25 / 0.75` is roughly 33, not 37.

rsn8887: Fixed as part of proposed merge 
https://github.com/scummvm/scummvm/pull/912

PR #910, opened 2017-02-25
SDL: added shader menu (hidden if no shaders)

rsn8887:
A shader selection menu tab is added to the options. The menu is hidden unless the platform supports shaders (via the new feature kFeatureShader). The shader selection menu code makes use of new system functions getSupportedShaders, setShader, and getShader (mirrored after the functions used to change graphics modes), added in this PR, and adds no platform specific code to common namespace.

Currently the shader menu is only used by the yet-to-be-merged Vita backend, but I added a "UpdateShader" stub here to the surfaceSdlGraphicsManager. The stub is currently empty but will be filled by Vita code for that port, and can for example be filled by any other dev that wants to add shader support and would like a convenient shader selection menu.

The Idea is that the user could select hardware shaders that are applied in addition to the software scalers for example for overlays (scanlines) etc. Combining software scaling with hw shaders is quite powerful because of the different combinations, as shown by many emulators that support this.


rsn8887:
This will be part of the vita branch merge anyway, and after thinking about it some more, I think it will create merge problems if it is duplicated here.  I will remove it for now.

PR #911, opened 2017-02-26
GUI: Improve tab widget

wjp:
This does three things; see the individual commits for longer explanations:

* Clean up a previous hacky fix to GlobalOptionsDialog::reflow() changing which tabs are visible.
* Fix some invalid rects when radio/popup controls don't fully fit in their clip rects.
* Make tab width dynamic to make more tabs fit on-screen.

Since the last one is potentially disruptive, I would appreciate feedback on any unexpected side effects.


wjp:
I updated this PR to go quite a bit further than before: the individual tabs can now all have different widths, to make them even more space-efficient.

rsn8887:
Great work! I tested it on the Vita version, and it works great at all resolutions and with both standard layouts. One minor suggestion: There could be maybe one or two few pixels more width to each tab, because the text in the tabs looks a little bit crammed now (it gets very close to the tab border).

wjp:
Giving it a few extra pixels does look better. I've changed that.

sev-:
Great work! Merging.

PR #912, opened 2017-02-26
SDL: fix options control tab ordering and labels

rsn8887:
- control options tab is moved past graphics options tab
- control tab is not the default anymore
- changed label of "Mouse Speed" option to "Pointer Speed"
- changed Pointer Speed infotip description to make it more clear that it affects the keyboard/joystick pointer control only
- also made Pointer Speed labels translatable.


gui/options.cpp
```diff
@@ -244,7 +245,7 @@ void OptionsDialog::build() {
 			int value =  ConfMan.getInt("kbdmouse_speed", _domain);
 			if (_kbdMouseSpeedSlider && value < sizeof(kbdMouseSpeedLabels)) {
```
wjp: This shouldn't be `sizeof`, but `ARRAYSIZE(kbdMouseSpeedLabels)`. Or maybe `- 1` to avoid the 0 terminator. You should probably also check if `value >= 0`.

rsn8887: fixed.

wjp:
Thanks

PR #913, opened 2017-02-27
MT32: Update Munt to 2.0.3

wjp:
This update uses upstream commit 777c51cdb4dbb4e02a53c23edea9086f0b600e26.

The new SampleRateConverter is added, but not built as we don't use it.
Also, building it without source changes will need additional include
directories.

This update of Munt reduces the stack size, and thus fixes bug #9630.


wjp:
Merged as 867511d2d0a5d063e37b07115b8b98f5930d415b.

PR #914, opened 2017-03-01
SDL: Always use sub-pixel joystick pointer

rsn8887:
With the new joystick pointer speed adjustment in the SDL backend which can be less then one pixel per frame delay, all calls to 
ProcessMouseEvent(event, _km.x / MULTIPLIER, _km.y / MULTIPLIER) 
with those exact second and third arguments need to be prevented from updating _km.x and _km.y since that will destroy the sub-pixel resolution and make the joystick pointer control less precise. But calls to ProcessMouseEvent with other arguments should still update _km.x and _km.y.

Before this PR, the call to ProcessMouseEvent inside handleKbdMouse was made to work correctly by first saving and, after the call, re-loading the hi-res _km.x and _km.y coordinate values. But all the other calls still exhibit the problem of overwriting _km.x and _km.y with imprecise numbers.

To fix this, in this PR, I removed the update of _km.x and _km.y from the processMouseEvent routine, and instead added explicit _km updates to the sources where they are needed. Since most calls to ProcessMouseEvent are in the form ProcessMouseEvent(_km.x / MULTIPLIER, _km.y / MULTIPLIER) this did not require many changes, and I think the result is (IMO) in clearer code since processMouseEvent does not have this side-effect of changing _km.x and _km.y anymore.

### DIFF
--- a/backends/events/androidsdl/androidsdl-events.cpp
+++ b/backends/events/androidsdl/androidsdl-events.cpp
@@ -50,6 +50,9 @@ bool AndroidSdlEventSource::handleMouseButtonDown(SDL_Event &ev, Common::Event &
 		return false;
 
 	processMouseEvent(event, ev.button.x, ev.button.y);
+	// update KbdMouse
+	_km.x = ev.button.x * MULTIPLIER;
+	_km.y = ev.button.y * MULTIPLIER;
 
 	return true;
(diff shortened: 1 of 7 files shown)


sev-:
OK, thanks, merging.

PR #916, opened 2017-03-05
ANDROIDSDL: libcurl and sdl_net is enabled by default

lubomyr:


sev-:
Nice. Next step is to have buildbot build it :)

PR #918, opened 2017-03-08
SCUMM: MI2: Kill hanging tunes on start of next, Fixes bug #1410

BenCastricum:
Fix the "MI2: Two soundtracks playing at once" bug from the forum, as
well as the one the bug reporter experienced.

Both are caused by starting tunes as trigger from other tunes. This code suppresses the start or kills the hanging tune on start of the next. I left in some debug statements to see when the code kicks in so we can determine if there are any unexpected side effects. In the final version those probably should be removed.

### DIFF
--- a/engines/scumm/imuse/imuse.cpp
+++ b/engines/scumm/imuse/imuse.cpp
@@ -696,6 +696,23 @@ bool IMuseInternal::startSound_internal(int sound, int offset) {
 	if (_game_id == GID_MONKEY2 && (sound == 100) && (getSoundStatus_internal(107, true) == 1))
 		IMuseInternal::stopSound_internal(107);
 
+	// Workaround for monkey2 bug #1410 / Booty Island
+	//
+	// Tunes involved
+	// 100 - Captain Dread's map
+	// 113 - Guard Kiosk / Mardi Grass 
+	// 115 - Map of Booty Island / Ville de la Booty
+	// 118 - Ville de la Booty
+	//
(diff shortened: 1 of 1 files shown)


BenCastricum:
The debugging code has been removed. I think it's merge-able now.

engines/scumm/imuse/imuse.cpp
```diff
@@ -675,6 +675,44 @@ bool IMuseInternal::startSound_internal(int sound, int offset) {
...
+	// 101 and 101 moves to an outro and stop (triggering start of 107).
+	// Then at Captain Dread tune 107 is stopped and 100 is started.
+	//
+	// If you go quickly enough, the trigger occurs not at the Scabb Island
+	// map but at Captain Dread causing tune 107 not to be stopped.
+	// So we prevent starting 107 if 100 is already running.
+	if (_game_id == GID_MONKEY2 && (sound == 107) && (getSoundStatus_internal(100, true) == 1))
+		return false;
+
+	/* In some cases 107 is running and doesn't get killed at Dread's map */
```
bluegr: Why do you change comments to use /* .. */ here, when you're using // in the rest of your changes?

BenCastricum: Just an old habbit from other projects. I fixed it.

digitall:
@sev- : This seems fairly reasonable and self-contained. Any objections to just merging this?

sev-:
Right, merging

PR #919, opened 2017-03-09
GUI: Fix for #9711

Joefish:
Fix for [GUI: Changing 'GUI Language' in Global Options removes theme overlay](https://bugs.scummvm.org/ticket/9711).  
Also, _cbcf819_ fixes the theme label to display the correct theme.  
For more details see the commit messages.

### DIFF
--- a/gui/ThemeEngine.cpp
+++ b/gui/ThemeEngine.cpp
@@ -2420,19 +2420,32 @@ Common::String ThemeEngine::getThemeId(const Common::String &filename) {
 		return "builtin";
 
 	Common::FSNode node(filename);
-	if (!node.exists())
-		return "builtin";
+	if (node.exists()) {
+		if (node.getName().matchString("*.zip", true)) {
+			Common::String id = node.getName();
 
-	if (node.getName().matchString("*.zip", true)) {
-		Common::String id = node.getName();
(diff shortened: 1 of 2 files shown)


criezy:
Good catch!
And good work on following our conventions for the code style and commit messages.

Currently getThemeId(file) is indeed not working in all cases and as a result ThemeEngine can set _themeId to "builtin" instead of the correct ID. Somehow we did not notice this before, probably because this _themeId is not used in many place. Your first commit is OK to fix this issue.

The second issue is more minor (I didn't even notice it at first and had to read again your description to understand what was wrong - by the way good work on your commit messages!). But it's a nice one to get fixed nonetheless.

Your second commit will introduce a change of behaviour when changing both the language and the theme and some combination of old or new theme not supporting the charset for the old or new language. I don't see that as an issue though as I don't see one better than the other and neither is optimal. I have mapped all 7 combination of incompatibilities I can think of, and compared the old behaviour with the new ones. The old code behaves a bit better generally, but that is marginal, and both fail in some cases where the new language is compatible with the new theme (so they could succeed), but the new theme is not compatible with the old language and/or the new language is not compatible with the old theme. That might be the subject of another PR for a GSoC candidate :P

Thank you. I am merging this as it is.

PR #921, opened 2017-03-12
GUI: Restore previous settings if GUI cannot be rendered

Joefish:
This PR focuses on GlobalOptionsDialog::apply() to determine incompatible font/charset and fallback to previous settings on error.
'Greek' always triggers an error out of unknown reasons to me. It is the only language with charset iso-8859-7, so looking at ThemeEngine::init(), how fonts/charsets are loaded and why it fails and falls back to ASCII is the next step.

@criezy you mentioned in #919 that you have mapped out combinations that broke previously.
Would be great if you could reply and paste your list here :)

### DIFF
--- a/gui/options.cpp
+++ b/gui/options.cpp
@@ -1619,7 +1619,7 @@ void GlobalOptionsDialog::build() {
 
 	// Select the currently configured language or default/English if
 	// nothing is specified.
-	if (ConfMan.hasKey("gui_language"))
+	if (ConfMan.hasKey("gui_language") && !ConfMan.get("gui_language").empty())
 		_guiLanguagePopUp->setSelectedTag(TransMan.parseLanguage(ConfMan.get("gui_language")));
 	else
 #ifdef USE_DETECTLANG
@@ -1797,6 +1797,10 @@ void GlobalOptionsDialog::clean() {
 }
 
 void GlobalOptionsDialog::apply() {
+	OptionsDialog::apply();
+
+	bool isRebuildNeeded = false;
+
 	Common::String savePath(_savePath->getLabel());
 	if (!savePath.empty() && (savePath != _("Default")))
 		ConfMan.set("savepath", savePath, _domain);
@@ -1832,15 +1836,6 @@ void GlobalOptionsDialog::apply() {
 #endif
 
 	ConfMan.setInt("autosave_period", _autosavePeriodPopUp->getSelectedTag(), _domain);
-
-	GUI::ThemeEngine::GraphicsMode selected = (GUI::ThemeEngine::GraphicsMode)_rendererPopUp->getSelectedTag();
-	const char *cfg = GUI::ThemeEngine::findModeConfigName(selected);
-	if (!ConfMan.get("gui_renderer").equalsIgnoreCase(cfg)) {
-		// FIXME: Actually, any changes (including the theme change) should
-		// only become active *after* the options dialog has closed.
-		g_gui.loadNewTheme(g_gui.theme()->getThemeId(), selected);
-		ConfMan.set("gui_renderer", cfg, _domain);
-	}
 
 #ifdef USE_UPDATES
 	ConfMan.setInt("updates_check", _updatesPopUp->getSelectedTag());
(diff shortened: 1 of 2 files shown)


criezy:
I have not looked at this code yet. The mapping I wrote assumes that:
- If the new language charset is not compatible with the current theme, it fallback to English
- If the new theme is not compatible with the current language charset, it reverts to the previous theme.

Based on that I mapped what happens when changing both the language and theme, if the language is changed first (as was done before PR #919) and when the theme is changed first (as was done in PR #919). The 7 cases I considered are a combination of 3 incompatibilities:
1. The new language is not compatible with the old theme
2. The new theme is not compatible with the old language
3. The new language is not compatible with the new theme.

This gave me the following table:
```
Incompatibilities  |  Change language first         |  Change Theme first
--------------------------------------------------------------------------
        1          |  English + new theme           |  OK (new language + new theme)
        2          |  OK (new language + new theme) |  English + new theme
        3          |  new language + old theme      |  English + new theme
       1+2         |  English + new theme           |  English + old theme
       1+3         |  English + new theme           |  English + new theme
       2+3         |  new language + old theme      |  new language + old theme
      1+2+3        |  English + new theme           |  English + old theme
```

Note: The issue you see with greek is likely related to missing Helvetica font for charset iso-8859-7 in our code base.

Joefish:
The behavior described in the table of criezy's post has changed drastically with this PR.
Currently, the settings are reset to the previous state if the GUI fails to be rendered.
This means that when changing [classic theme, English] to [modern theme, Russian] and Russian
is not supported it falls back to [classic theme, English] although English may be supported by modern theme.
If this is not desired, please reply here and define the behavior.

What's currently left on my list:
- Clearing buffered fonts on theme change
- Track down segfault in BdfFont::~BdfFont on shutdown

gui/options.cpp
```diff
@@ -1878,39 +1878,60 @@ void GlobalOptionsDialog::apply() {
...
 	}
 #endif
 
 	GUI::ThemeEngine::GraphicsMode gfxMode = (GUI::ThemeEngine::GraphicsMode)_rendererPopUp->getSelectedTag();
-	const char *cfg = GUI::ThemeEngine::findModeConfigName(gfxMode);
-	if (!ConfMan.get("gui_renderer").equalsIgnoreCase(cfg)) {
-		ConfMan.set("gui_renderer", cfg, _domain);
+	Common::String oldGfxConfig = ConfMan.get("gui_renderer");
+	Common::String newGfxConfig = GUI::ThemeEngine::findModeConfigName(gfxMode);
+	if (newGfxConfig != oldGfxConfig) {
```
criezy: Is there any reason here to change the case-insensitive test into a case-sensitive one?

criezy:
The behaviour you propose in case of language/theme incompatibility seems fine to me. At least it is simple. I have not checked but I suspect we might get some confusing warning messages though (for example the one from ThemeEngine that says that it fails to load the localized font and use stye default language instead, when in the end we revert to the previous language, which might not be the default language). So I think it might be nice to expand a bit the error MessageDialog you have in GlobalOptionsDialog::apply() when reverting changes.

Another remark is that with your change it seems that it is rebuilding the dialog whenever redraw is true, which happens when changing the language or the theme or the renderer. But rebuild is only needed when changing the language. When changing the theme or renderer without changing the language calls to loadNewTheme() and draw() are sufficient. So it would be nice to avoid rebuilding the dialog in such a case.

gui/options.cpp
```diff
@@ -1619,7 +1619,7 @@ void GlobalOptionsDialog::build() {
 
 	// Select the currently configured language or default/English if
 	// nothing is specified.
-	if (ConfMan.hasKey("gui_language"))
+	if (ConfMan.hasKey("gui_language") && ConfMan.get("gui_language") != "")
```
criezy: That was indeed one of three easy way to fix the issue. I think this is the one I like the least, but it is also the less intrusive and safer. So this is fine with me.
However, rather than test equality with an empty string, you should use `bool Common::String::empty()`.

gui/options.cpp
```diff
@@ -1922,18 +1922,20 @@ void GlobalOptionsDialog::apply() {
...
+				MessageDialog error(_("Theme cannot be loaded!"));
 				error.runModal();
 #ifdef USE_TRANSLATION
 			}
 			TransMan.setLanguage(oldLang);
 			_guiLanguagePopUp->setSelectedTag(selectedLang);
 			ConfMan.set("gui_language", oldLang);
 #endif
+			MessageDialog error(_("Misc settings will be restored."));
+			error.runModal();
```
criezy: Rather that display two separate ErrorDialog, I think it would be better to concatenate the two messages and use a single dialog.

criezy:
Thank you for the update. I added two comments and it would be nice to do the two minor changes I pointed out.

Then, please squash all those commits together and I think it is ready to be merged (I will have one last look once everything is squashed).

gui/options.cpp
```diff
@@ -1885,52 +1880,73 @@ void GlobalOptionsDialog::apply() {
...
-
-		// Rebuild the Launcher and Options dialogs
-		g_gui.loadNewTheme(g_gui.theme()->getThemeId(), ThemeEngine::kGfxDisabled, true);
+	if (isRebuildNeeded) {
 		rebuild();
 		if (_launcher != 0)
 			_launcher->rebuild();
 	}
-#endif // USE_TRANSLATION
+	g_gui.loadNewTheme(_newTheme, gfxMode, true);
```
criezy: Why this call? It is already done a few lines above (on line 1912).

gui/options.cpp
```diff
@@ -1885,52 +1880,73 @@ void GlobalOptionsDialog::apply() {
...
-		// Activate the selected language
-		TransMan.setLanguage(selLang);
-
-		// Rebuild the Launcher and Options dialogs
-		g_gui.loadNewTheme(g_gui.theme()->getThemeId(), ThemeEngine::kGfxDisabled, true);
+	if (isRebuildNeeded) {
 		rebuild();
 		if (_launcher != 0)
 			_launcher->rebuild();
 	}
```
criezy: There is something missing in there. If you change the theme, but without changing the language, while there is no need to rebuild the dialog, there is still a need to redraw it. And unless I missed it, your change removed the call to draw(). So you might want to keep two booleans: isRebuildNeeded and isRedrawNeeded, and add a `else if (isRedrawNeeded) draw();` here.

gui/options.cpp
Joefish: The first loadNewTheme() is more of a test for rendering theme and language. If I were to call the second loadNewTheme() only on error, rebuild() would be needed before the first loadNewTheme() call too (if lang was changed).
It is a tradeoff. Either destroying/constructing the whole GUI twice in case of error, or only once but always calling loadNewTheme() twice.. But I guess it is more probable for loadNewTheme() to succeed (or at least it should be..) so my argument doesn't hold :D

gui/options.cpp
Joefish: draw() just sets the redraw status for redrawing the top dialog. In loadNewTheme() it is set to full redraw what does exactly the same (GuiManager::redraw() +222).

criezy: Right sorry. I mixed the draw() implementation of Dialog and Widget in my head.
Also the old code (from 11 years ago) was doing a draw() just after the loadNewTheme() and I didn't take the time to check it was (still) needed.

Here this is indeed not needed as loadNewTheme() does a full redraw.

gui/options.cpp
criezy: I understand that if the first loadNewTheme() fails you need to call it again to set back the old theme. But can you explain why you need to call it again if it succeeds. You seem to imply in your comment that it needs to be called after the rebuild() (or that rebuild() needs to be called before loadNewTheme()). But as far as I know that is not the case (and the current code doesn't do it - it set the language, then calls loadNewTheme() then calls rebuild() and doesn't call again loadNewTheme()). We need to set the language before loadNewTheme() but this is already done before the first one.

Joefish: If it succeeds I don't need to call loadNewTheme() again. The only reason the second loadNewTheme() call is outside the 'failure section' is so the GUI doesn't need to be destroyed twice.
rebuild() needs to be called before loadNewTheme() otherwise you get codepage mess as widget text, in some cases corrupted paths and what not. Just paste the if (isRebuildNeeded) clause before the first loadNewTheme(), go Options->Misc, select Russian and builtin theme, enjoy :)
On language change, especially if the charset changes rebuild() needs to be called, otherwise it's mayhem.

criezy:
Thank you. This looks all good to me now.
I am merging this.

PR #923, opened 2017-03-19
CMD: Detect and add game(s) with commandline, Feature request 703

yinsimei:
Pull request for GSoC 2017.


sev-:
Please squash these commits, so your SLUDGE-related code is not even introduced in this branch

yinsimei:
> Please squash these commits, so your SLUDGE-related code is not even introduced in this branch

Thank you for pointing it out. I didn't know I could squash commits.

Yes, I've removed SLUDGE-related code from my fork. I think it won't be able to be merged and it still needs some fixes. I will re-add them later.

base/commandLine.cpp
```diff
@@ -1039,8 +1175,15 @@ bool processSettings(Common::String &command, Common::StringMap &settings, Commo
 		return true;
 	} else if (command == "auto-detect") {
 		// If auto-detects succeed, we want to return false so that the game is started
-		return !autoDetect(settings["path"]);
+		return !detectGames(settings["path"], true);
 		//return true;
+	} else if (command == "detect") {
+		detectGames(settings["path"], false);
+		return true;
```
yinsimei: This function is called in the main function in main.cpp.
`	// TODO: deal with settings that require plugins to be loaded
	if (Base::processSettings(command, settings, res)) {
		if (res.getCode() != Common::kNoError)
			warning("%s", res.getDesc().c_str());
		return res.getCode();
	}
       // open scummvm interface/run game instructions...`
I guess the return value of this function means: true->do nothing except for displaying informations in command line, false-> open scummvm gui/run game/... next
So I choose to always return true for "detect" means, I only want the game informations displayed and nothing else whatever the detect result is. But it"s free to change this return value depending on what we need.

yinsimei: For example, for the --auto-detect command line. We want not only informations of detected games, but also to play the first one detected. 
So we `return !detectGames(settings["path"], true);` here, which means if we detect anything, return false->Run the game next. if we don't detect anything, return true->display information and do nothing else.

wjp:
Just FYI: all the commits you're pushing to your branch automatically show up on this PR. It's generally better to use a separate branch for a PR.

yinsimei:
> Just FYI: all the commits you're pushing to your branch automatically show up on this PR. It's generally better to use a separate branch for a PR.

Thank you. I was just wondering why this is happening.

PR #925, opened 2017-03-22
Fullpipe Common::String + memory leak fixes

bluegr:
In this pull request, I've changed a lot of engine strings in Fullpipe so that they are handled via Common::String. This results in cleaner and more manageable code.

I've tried splitting up the commits so that all relevant changes are grouped together

Update: I've also added some memory leak fixes to this branch, as part of fixing memory leaks in this engine. These are relevant, because the original intention of this branch was to optimize code and identify memory leaks.

### DIFF
--- a/engines/fullpipe/scene.cpp
+++ b/engines/fullpipe/scene.cpp
@@ -165,7 +165,7 @@ bool Scene::load(MfcArchive &file) {
 	_sceneId = file.readUint16LE();
 
 	_sceneName = file.readPascalString();
-	debug(0, "scene: <%s> %d", transCyrillic((byte *)_sceneName), _sceneId);
+	debug(0, "scene: <%s> %d", transCyrillic(_sceneName), _sceneId);
 
 	int count = file.readUint16LE();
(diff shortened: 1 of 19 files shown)


engines/fullpipe/scene.cpp
```diff
@@ -221,10 +210,10 @@ bool Scene::load(MfcArchive &file) {
 
 	_libHandle = g_fp->_currArchive;
 
-	if (_picObjList.size() > 0 && _bgname && strlen(_bgname) > 1) {
+	if (_picObjList.size() > 0 && _bgname.size() > 1) {
```
sev-: !_bgname.empty()

bluegr: Fixed in af874d4

sev-:
Sweet.

wjp:
Looks good, but I don't understand why the Common::String function arguments are a mix of by-value and by-(non-const-)reference, instead of the standard `const Common::String &`.

bluegr:
Good point @wjp. I'll adapt these accordingly on master

PR #703, opened 2016-03-10
GUI: Added filter to EditableWidget::setEditString

rupak0577:
As mentioned in the TODO for EditableWidget

PR #926, opened 2017-03-22
CMD: Detect and add game(s) with commandline, Feature request #7885

yinsimei:
GCoS 2017. Move the cmd commit to a branch to work in master without affecting the PR.

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -35,6 +35,7 @@
 #include "common/config-manager.h"
 #include "common/fs.h"
 #include "common/rendermode.h"
+#include "common/stack.h"
 #include "common/system.h"
 #include "common/textconsole.h"
 
@@ -68,6 +69,13 @@ static const char HELP_STRING[] =
 	"  -z, --list-games         Display list of supported games and exit\n"
 	"  -t, --list-targets       Display list of configured targets and exit\n"
 	"  --list-saves=TARGET      Display a list of saved games for the game (TARGET) specified\n"
+	"  -a, --add                Add a game from current or specified directory\n"
+	"                           Use --path=PATH before -a, --add to specify a directory.\n"
+	"  --massadd                Add all games from current or specified directory and all sub directories\n"
+	"                           Use --path=PATH before --massadd to specify a directory.\n"
+	"  -d, --detect             Display a list of games from current or specified directory\n"
+	"                           without adding it to the config. Use --path=PATH before --detect\n"
+	"                           to specify a directory.\n"
 	"  --auto-detect            Display a list of games from current or specified directory\n"
 	"                           and start the first one. Use --path=PATH before --auto-detect\n"
 	"                           to specify a directory.\n"
@@ -409,6 +417,15 @@ Common::String parseCommandLine(Common::StringMap &settings, int argc, const cha
 			END_COMMAND
 
 			DO_COMMAND('z', "list-games")
+			END_COMMAND
+
+			DO_COMMAND('a', "add")
+			END_COMMAND
+
+			DO_LONG_COMMAND("massadd")
+			END_COMMAND
+
+			DO_COMMAND('d', "detect")
 			END_COMMAND
 
 			DO_LONG_COMMAND("auto-detect")
@@ -784,12 +801,12 @@ static void listAudioDevices() {
 }
 
 /** Display all games in the given directory, or current directory if empty */
-static bool autoDetect(Common::String path) {
+static GameList getGameList(Common::String path) {
 	if (path.empty())
 		path = ".";
+
 	//Current directory
 	Common::FSNode dir(path);
-
 	Common::FSList files;
 
 	//Collect all files from directory
@@ -798,8 +815,15 @@ static bool autoDetect(Common::String path) {
 	GameList candidates(EngineMan.detectGames(files));
 	if (candidates.empty()) {
 		printf("ScummVM could not find any game in %s\n", path.c_str());
-		return false;
 	}
+	return candidates;
+}
+
+/** Display all games in the given directory, add it to config according to input */
+static bool detectGames(Common::String path, bool addToConfig) {
+	GameList candidates = getGameList(path);
+	if (candidates.empty())
+		return false;
 
 	// Print all the candidate found
 	printf("ID                   Description\n");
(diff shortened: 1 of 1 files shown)


base/commandLine.cpp
```diff
@@ -68,6 +69,13 @@ static const char HELP_STRING[] =
 	"  -z, --list-games         Display list of supported games and exit\n"
 	"  -t, --list-targets       Display list of configured targets and exit\n"
 	"  --list-saves=TARGET      Display a list of saved games for the game (TARGET) specified\n"
+	"  -a, --add                Add a game from current or specified directory\n"
+	"                           Use --path=PATH before -a, --add to specify a directory.\n"
+	"  --massadd                Add all games from current or specified directory and all sub directories\n"
+	"                           Use --path=PATH before --massadd to specify a directory.\n"
+	"  -d, --detect             Display a list of games from current or specified directory\n"
```
criezy: -d is already used for the debug level

base/commandLine.cpp
```diff
@@ -798,20 +815,139 @@ static bool autoDetect(Common::String path) {
...
+		return false;
+
+	int idx = 0;
+	// Pick one if there are several games
+	if (candidates.size() > 1) {
+		// Print game list
+		printf("Several games are detected. Please pick one game to add: \n");
+		int i = 1;
+		for (GameList::iterator v = candidates.begin(); v != candidates.end(); ++i, ++v) {
+			printf("%2i. %s\n", i, v->description().c_str());
```
criezy: Maybe print the gameid as well, for consistency with detectGames(), and because some users are used to those.

base/commandLine.cpp
```diff
@@ -798,20 +815,139 @@ static bool autoDetect(Common::String path) {
...
+			for (GameList::iterator v = candidates.begin(); v != candidates.end(); ++v) {
+
+				++ndetect;
+				Common::String domain = v->preferredtarget();
+				// If game was already added, don't add it
+				if (ConfMan.hasGameDomain(domain))
+					continue;
+
+				// Add game
+				ConfMan.addGameDomain(domain);
```
criezy: This is not sufficient, is it? Have you tested that you can actually start the game from the launcher after adding it this way. I would expect that at least the path would need to be added to ConfMan as well for the new domain.

base/commandLine.cpp
```diff
@@ -798,20 +815,139 @@ static bool autoDetect(Common::String path) {
...
+			candidates[idx].gameid().c_str(),
+			candidates[idx].description().c_str(),
+			Common::getLanguageDescription(candidates[idx].language()),
+			Common::getPlatformDescription(candidates[idx].platform()));
+	return true;
+}
+
+static bool massAddGame(Common::String path) {
+	if (path.empty())
+			path = ".";
```
criezy: Indentation is wrong for this line.

base/commandLine.cpp
```diff
@@ -1039,8 +1175,15 @@ bool processSettings(Common::String &command, Common::StringMap &settings, Commo
...
 	} else if (command == "auto-detect") {
 		// If auto-detects succeed, we want to return false so that the game is started
-		return !autoDetect(settings["path"]);
+		return !detectGames(settings["path"], true);
 		//return true;
+	} else if (command == "detect") {
+		detectGames(settings["path"], false);
+		return true;
+	} else if (command == "add") {
+		return !addGame(settings["path"]);
```
criezy: Do we actually want to start ScummVM after adding a game with the command line?
The same question also applies to mass add below.

I don't have a definite answer to this question. But I think I would not expect ScummVM to start after adding the game(s).

criezy:
Overall this looks quite good, although there are some changes that need to be made.

I am also wondering if we could avoid some code duplication between the GUI and the command line for adding games to the config file and for the mass add detection code.

base/commandLine.cpp
yinsimei: Ok, I will remove it

base/commandLine.cpp
yinsimei: Ok, I will add it.

base/commandLine.cpp
yinsimei: Yes, I've tested it. The game can be started from launcher. I think the path has already been taken care of in the function detectGames of metaengine or somewhere else. I will take a deeper look into how it works exactly.

base/commandLine.cpp
yinsimei: Yes, it's not specified in the feature request. I will change it.

wjp: For context, feature requests are just that: requests. They haven't necessarily been discussed or fully worked out.

wjp:
"Feature request #703" is old SF/Allura numbering. You should mention 7885 instead ( https://bugs.scummvm.org/ticket/7885 ).

yinsimei:
> I am also wondering if we could avoid some code duplication between the GUI and the command line for adding games to the config file and for the mass add detection code.

Yes, I've noticed the duplication. Do think it will be a good idea to have a massDetectGames in MetaEngine?

base/commandLine.cpp
yinsimei: No, you are right, it's not sufficient. The game can only be started at the first time and then there is a problem after re-opening the launcher. I will look into it

criezy:
Thank you.
I have now merged this change after squashing the two commits.

base/commandLine.cpp
```diff
@@ -68,6 +69,13 @@ static const char HELP_STRING[] =
 	"  -z, --list-games         Display list of supported games and exit\n"
 	"  -t, --list-targets       Display list of configured targets and exit\n"
 	"  --list-saves=TARGET      Display a list of saved games for the game (TARGET) specified\n"
+	"  -a, --add                Add a game from current or specified directory\n"
+	"                           Use --path=PATH before -a, --add to specify a directory.\n"
+	"  --massadd                Add all games from current or specified directory and all sub directories\n"
```
tobiatesan: This is very useful, but I find --massadd is super unusual as a command line switch. Could we change this to `--all` (perhaps with `-A`) or something before 1.10?

@yinsimei @criezy @sev-

tobiatesan: We've discussed the issue a bit with @criezy, see:

TL;DR, I don't find this very intuitive and I don't especially like the idea of interaction (if you want interaction, you'd use a GUI after all...)

The semantics I'd like to propose is:

```
-p <dir> --add                adds all games
-p <dir> --detect             enumerates dectected games in <dir>
-p <dir> --game <id> --add     adds just game <id>
```

How do you like it?

tobiatesan: (If you like the current semantics, I think the help should at least mention that `-a` is interactive -- especially since it's interactive only some of the time; that might give an headache to some script writer and/or cabinet modder.)

bgK: That sounds good and actually usable by third party launchers.

Maybe the `addGameToConf` function could additionally display the target of added games so that it can be used for subsequent calls to scummvm.

yinsimei: So, if I understand well, we'll use --add in the place of --massadd and add a --game option for --add ?

tobiatesan: @yinsimei yes, that's essentially the idea. And `--add` would error out if there are multiple games with no `--game` option specified (possibly enumerating them or just advising to use `--detect`)
@bgK good idea, I like it.

criezy: > And --add would error out if there are multiple games and no --game option is passed

@tobiatesan I though we agreed during the discussion on IRC today that --add without --game would add all the games found (like the current --massadd)? And that is also what one of your earlier comment above says.

tobiatesan: @criezy yes, sorry, that's correct. I honestly have no idea why I wrote the bit you quoted.

PR #927, opened 2017-03-24
WINDOWS: Flow of taking screenshots on Windows fix #9701

NewBoy22:
This pull request fixes the silent fail when making screenshots if ScummVM was installed with Windows installer to the default directory ["system partition"\Program Files (x86)].

As suggested in the bug ticket the screenshot directory has been changed to %AppData%\ScummVM\Screenshots (or Windows directory on older Windows). README has also been updated as requested.

https://bugs.scummvm.org/ticket/9701#no1

### DIFF
--- a/README
+++ b/README
@@ -81,8 +81,9 @@ Table of Contents:
 8.0) Configuration file
  * 8.1 Recognized configuration keywords
  * 8.2 Custom game options that can be toggled via the GUI
-9.0) Compiling
-10.0) Credits
+9.0) Screenshots (SDL backend only)
+10.0) Compiling
+11.0) Credits
 
 
 1.0) Introduction:
--- a/README
+++ b/README
@@ -2631,23 +2631,7 @@ configuration of each entry, allowing the custom options to be shown.
 9.0) Screenshots (SDL backend only):
 ---- -------------------------------
 By default screenshots are put into the current directory, however on Windows
-it has a preset directory:
-
-    Windows Vista/7:
-    \Users\username\AppData\Roaming\ScummVM\Screenshots\
-
-    Windows 2000/XP:
-    \Documents and Settings\username\Application Data\ScummVM\Screenshots\
-
-    Windows NT4:
-    <windir>\Profiles\username\Application Data\ScummVM\Screenshots\
-    
-    Windows 95/98/ME:
-    <windir>\
-
-Screenshots are stored under a hidden area in Windows NT4/2000/XP/Vista/7,
-which can be accessed by running "%APPDATA%\ScummVM\Screenshots\" or by
-enabling hidden files in Windows Explorer.
+the directory for this purpose is set to "Users\username\My Pictures\ScummVM Screenshots".
 
 
 10.0) Compiling:
--- a/README
+++ b/README
@@ -2627,7 +2627,29 @@ once or readded in the ScummVM launcher's game list. This will update the
 configuration of each entry, allowing the custom options to be shown.
 
 
-9.0) Compiling:
+9.0) Screenshots (SDL backend only):
+---- -------------------------------
+By default screenshots are put into the current directory, however on Windows
+it has a preset directory:
+
+    Windows Vista/7:
+    \Users\username\AppData\Roaming\ScummVM\Screenshots\
+
+    Windows 2000/XP:
+    \Documents and Settings\username\Application Data\ScummVM\Screenshots\
+
+    Windows NT4:
+    <windir>\Profiles\username\Application Data\ScummVM\Screenshots\
+    
+    Windows 95/98/ME:
+    <windir>\
+
+Screenshots are stored under a hidden area in Windows NT4/2000/XP/Vista/7,
+which can be accessed by running "%APPDATA%\ScummVM\Screenshots\" or by
+enabling hidden files in Windows Explorer.
+
+
+10.0) Compiling:
 ---- ----------
 For an up-to-date overview on how to compile ScummVM for various
 platforms, please consult our Wiki, in particular this page:
@@ -2736,7 +2758,7 @@ debug messages (see <https://technet.microsoft.com/en-us/sysinternals/debugview.
       <http://wiki.scummvm.org/index.php/Compiling_ScummVM/Symbian>
 
 
-10.0) Credits
+11.0) Credits
 ----- -------
 Please refer to our extensive Credits list at:
 
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -623,55 +623,34 @@ bool OpenGLSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 				return true;
 			}
 
-#ifdef WIN32
 			// Alt-s creates a screenshot
 			if (event.kbd.keycode == Common::KEYCODE_s) {
 				Common::String filename;
 
-				Common::String screenshotsPath = OSystem_Win32::getScreenshotsPath();
-
-				char fullpath[MAXPATHLEN];
-
-				for (int n = 0;; n++) {
-					SDL_RWops *file;
-
-					filename = Common::String::format("scummvm%05d.bmp", n);
-
-					strcpy(fullpath, screenshotsPath.c_str());
-					strcat(fullpath, filename.c_str());
-
-					file = SDL_RWFromFile(fullpath, "r");
-					if (!file)
-						break;
-					SDL_RWclose(file);
-				}
-
-				saveScreenshot(fullpath);
-				debug("Saved screenshot '%s'", filename.c_str());
-
-				return true;
-		}
+#ifdef WIN32
+				Common::String screenshotsPath = OSystem_Win32().getScreenshotsPath();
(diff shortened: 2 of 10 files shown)


raziel-:
The TOC in Readme needs to be changed accordingly

backends/platform/sdl/win32/win32.cpp
```diff
@@ -145,6 +145,46 @@ bool OSystem_Win32::openUrl(const Common::String &url) {
...
+	win32OsVersion.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
+	GetVersionEx(&win32OsVersion);
+	// Check for non-9X version of Windows.
+	if (win32OsVersion.dwPlatformId != VER_PLATFORM_WIN32_WINDOWS) {
+		// Use the Application Data directory of the user profile.
+		if (win32OsVersion.dwMajorVersion >= 5) {
+			if (!GetEnvironmentVariable("APPDATA", screenshotsPath, MAXPATHLEN))
+				error("Unable to access application data directory");
+		}
+		else {
```
bluegr: Formatting: } else {

Please refer to: http://wiki.scummvm.org/index.php/Code_Formatting_Conventions

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -620,6 +623,35 @@ bool OpenGLSdlGraphicsManager::notifyEvent(const Common::Event &event) {
...
+						break;
+					SDL_RWclose(file);
+				}
+
+				saveScreenshot(fullpath);
+				debug("Saved screenshot '%s'", filename.c_str());
+
+				return true;
+		}
+#else
```
criezy: You can factorise a lot of code between the windows case and the else case. Actually only the line where you get the screenshot path needs to be different.

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -2522,6 +2525,38 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
...
+			}
+
+			if (saveScreenshot(fullpath))
+				debug("Saved screenshot '%s'", filename.c_str());
+			else
+				warning("Could not save screenshot");
+
+			return true;
+		}
+#else
```
criezy: Again, you can factorise a lot of code between the windows case and the else case.

backends/platform/sdl/win32/win32.h
```diff
@@ -38,6 +38,8 @@ class OSystem_Win32 : public OSystem_SDL {
 
 	virtual bool openUrl(const Common::String &url);
 
+	static Common::String getScreenshotsPath();
```
criezy: Maybe this could be made a virtual function in OSystem_SDL so that if needed we can specify a location on other OS's as well.

mgerhardy:
Wouldn't it be better to return the home directory or the scummvm home directory as a default? Because otherwise you can't save a screenshot if you launched scummvm from /usr/bin (or do you change working directory inside of scummvm?). Saving to current working dir won't work in most cases on *nix.

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -2522,23 +2525,33 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
...
 		}
 
-		// Alt-S: Create a screenshot
+		// Alt-S: Create a screenshot - Win32
 		if (event.kbd.hasFlags(Common::KBD_ALT) && event.kbd.keycode == 's') {
-			char filename[20];
+			Common::String filename;
+
+#ifdef WIN32
+			Common::String screenshotsPath = OSystem_Win32().getScreenshotsPath();
```
criezy: You want to access the unique OSystem instance here, from which the OSystem_SDL (and thus OSystem_Win32) derives. This instance is accessed by using g_system. Then you can cast it to OSystem_SDL to get access to the getScreenshotsPath() function.

And since you now have a virtual function there is no need to use OSystem_Win32 explicitly and you can use the same code on all systems.

(the same comment applies to openglsdl-graphics.cpp)

README
```diff
@@ -2627,7 +2628,29 @@ once or readded in the ScummVM launcher's game list. This will update the
...
+    Windows 2000/XP:
+    \Documents and Settings\username\Application Data\ScummVM\Screenshots\
+
+    Windows NT4:
+    <windir>\Profiles\username\Application Data\ScummVM\Screenshots\
+    
+    Windows 95/98/ME:
+    <windir>\
+
+Screenshots are stored under a hidden area in Windows NT4/2000/XP/Vista/7,
```
mgerhardy: also keep in mind that you can just use SDL_GetPrefPath if sdl > 2.0 is used (and it looks like you are talking about the sdl backend here anyway).

NewBoy22:
@csnover Thank you for the instructions!
I decided to save the Screenshots to AppData because I thought that all the files that ScummVM creates should be together in one place. It can be annoying for a Windows user when a software leaves all kinds of traces of itself everywhere.

The final decision is of course not mine. So do you want me to move the screenshots to the My Pictures directory ? Also on Windows 95 there is no My Pictures only My Documents or at least that is what the Wikipedia says. So I could separate Windows 95 from every other Windows version and do it like this: Windows 95 -> My Documents\ScummVM Screenshots, Windows 98 and above -> My Pictures\ScummVM Screenshots. And of course when it is clear where to put it, I will fix every other cconcerns about the code.

Kirben:
Your are right, that AppData would be the best place to store ScummVM screenshots on Windows, so all the ScummVM data is stored in one location. The 'My Documents' area is really only meant for files created specifically by the user.

I'm not sure that Windows 95/98/ME even matters anymore , as there was no release of ScummVM 1.9.0. for Windows 9x.

criezy:
It seems to me that the screenshot is created by the user (by pressing Alt+s) and that he will want to have easy access to them. Thus on OS X I am thinking of saving them on the Desktop. This is where the OS saves screenshots when using the Crtl+Shift+3 (or 4) shortcuts, so it would be familiar to users. An alternative would be to use the user Documents directory (where we save ScummVM savegames by default on OS X) or the user Pictures directory. In any case I will implement the OS X side once this PR has been merged (unless NewBoy22 wants to do it as part of this PR).

I am not sure that where I decide to save the screenshots on OS X in the end has any relevance for the discussion of where they should be saved on Windows though. I think it partly depends on what the users are used to on a given OS (and I am not a Windows user, so I don't know what a Windows user would expect).

I am also thinking that maybe we could read a screenshotpath setting from the config file, without having a way to set/change it through the GUI, but to allow advanced users to change it anyway (by editing the config file) if they don't like the default. And in that case it might make sense to support it on all platforms using the SDL backend.

Kirben:
Yes, Windows is a quite different, Microsoft have altered various locations and naming several times over the years. Microsoft actually discourages leaving unused items on the desktop these days.

It would definitely be worth adding a Start Menu item to the Windows installer script, for easy access to the location of screenshots, when this is eventually merged too.

NewBoy22:
On Windows there is no concept by Microsoft about where to or how to deal with these kind of application datas. The new Windows infrastructure the UWP tries to create a convention by sandboxing the UWP apps. It kinda works like Android. To use the OS's shared folders you have to ask permission in the manifest file, but the application have full permission to it's AppData folder. Also on Windows there is no default place for the OS screenshots since there is no shortcut for it. The PrintScreen and the Snipping Tool only use the clipboard, the user himself has to save it somewhere manually.

I think the best is to stick with AppData and create a shortcut to it with the installer script into the Start Menu as @Kirben said.

The My Picture folder is I think a place where most of the users put theit family photos and private stuff like that. It is annoying to see random application's photos between personal pictures, even if it has a separate folder.

csnover:
> Also on Windows there is no default place for the OS screenshots since there is no shortcut for it. The PrintScreen and the Snipping Tool only use the clipboard, the user himself has to save it somewhere manually.

While I do not have access to personally verify this, [several](https://www.howtogeek.com/226280/how-to-take-screenshots-in-windows-10/) [sources](http://www.wikihow.com/Screenshot-in-Windows-10) [online](http://www.laptopmag.com/articles/how-to-take-a-screenshot-in-windows-8) indicate that Windows 8 and 10 will create and place screenshots in the user’s `Pictures\Screenshots` directory if you use Win+PrtScr.

NewBoy22:
I can't verify that either, I use Windows 7 but it seems true. Still, my personal opinion is that it is not the best solution. The config file is also a good ide, so it could be changed but it still needs a default place.

Forgot to mention that I don't have access to OSX so I can't implement that part.

criezy:
Thank you for your work.
I have squashed all the commits and merged this in master. I have also implemented the `getScreenshotsPath()` for macOS X and added the possibility to specify the path in the config file.

criezy:
I am reopening this has this actually broke the compilation on built:
`fatal error: ShlObj.h: No such file or directory`

Thus I commented out that part of the code for now and this should be reworked.

criezy:
Kirben fixed the compilation already (thanks Kirben!). So I am closing this again.

PR #928, opened 2017-03-26
DIRECTOR: Add detection for Arnie and its demo

VelocityRa:
My versions are in Swedish but I'm pretty sure there aren't any other versions, that's why I chose `Common::EN_ANY`.

The game doesn't currently boot as `loadCastData` eventually tries to read non-Cast data and asserts [here](https://github.com/scummvm/scummvm/blob/master/engines/director/score.cpp#L495).

### DIFF
--- a/engines/director/detection_tables.h
+++ b/engines/director/detection_tables.h
@@ -538,7 +538,7 @@ static const DirectorGameDescription gameDescriptions[] = {
 			GUIO1(GUIO_NOASPECT)
 		},
 		GID_GENERIC,
-		2
+		5
 	},
 
 	{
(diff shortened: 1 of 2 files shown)


engines/director/detection_tables.h
```diff
@@ -527,6 +527,34 @@ static const DirectorGameDescription gameDescriptions[] = {
 		2
 	},
 
+	{
+		{
+			"ernie",
+			"Ernie - Demo",
```
sev-: Since you're adding another target, you need to register it in the detection.cpp file as well.

sev-:
D5 support is quite far away in current codebase. We haven't even looked much into booting it.

VelocityRa:
Ah, alright. Issue is fixed anyway.

sev-:
Thanks

henke37:
English versions existing or not, the language is Swedish and nothing else.

PR #929, opened 2017-03-28
Do not hardcode resolutions for aspect ratio correction

criezy:
The changes in this PR aim at correcting missing aspect ratio correction for some SCI games that should support it. This was reported in the forum for King's Quest V mac, which uses a resolution of 320x190: http://forums.scummvm.org/viewtopic.php?t=14303

The issue is that in both the SurfaceSDL and OpenGL backends the code checks the resolutions to decide whether to apply the correction or not, and only apply it for 320x200 and 640x400 resolutions.

My initial fix was to simply add more resolutions to the tests, and this can be found here: https://github.com/scummvm/scummvm/compare/master...criezy:sci-ar

On IRC yesterday snover asked: "can AR setting be retrieved from the engine instead of hardcoding numbers into the backend?" So I looked at a different approach and this is what I am proposing here. It checks whether the GUIO_NOASPECT is defined by the current target and applies the aspect ratio correction if it isn't.

Do you see issues with this approach?

### DIFF
--- a/backends/graphics/opengl/opengl-graphics.cpp
+++ b/backends/graphics/opengl/opengl-graphics.cpp
@@ -36,8 +36,11 @@
 #include "common/tokenizer.h"
 #include "common/rect.h"
 #endif
+#include "common/gui_options.h"
+#include "common/config-manager.h"
 
 #include "graphics/conversion.h"
+#include "graphics/scaler/aspect.h"
 #ifdef USE_OSD
 #include "graphics/fontman.h"
 #include "graphics/font.h"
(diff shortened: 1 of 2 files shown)


bgK:
Are you sure all the engines set GUIO_NOASPECT when aspect ratio correction does not apply?

criezy:
@bgK: No, I am not sure. I did a grep for it and it is used by quite a lot of engines, but I didn't check in details.

If there is no other reason to reject this PR, I will update the PR when I find some time to add the GUIO_NOASPECT where it is missing.

csnover:
@criezy With the ongoing discussion we have had about the AR stuff over the past while, is this PR still relevant, or should it be closed and a new one opened later with the changed AR handling code?

criezy:
I thought some of the discussion was in there, but it looks like it all happened somewhere else. To summarise we found out that:
- Some engines, such as ADL, need an aspect ratio correction that is different from the 5:6 correction we currently do.
- Some engines, such as supernova, have dual mode 320x200 and 640x480 and only the former should be corrected.
- Other engines such as Dreamweb also have dual mode 320x200 and 640x480 but neither should be corrected.

The work done by @csnover in PR #1007 and PR #1009 will help to properly handle all those cases. For this reason I am waiting for those PRs to be merged before continuing to work on AR correction.

I will probably work in a different branch and open a new PR then as it might be simpler than rebasing this one, but for now I will keep this PR open as a reminder (mostly to myself) I am working on this and close it when I open the other one.

rootfather:
@criezy, may I ask what's the current state of this now that #1007 and #1009 both are merged?

criezy:
I have started working on a different approach in a separate branch (https://github.com/criezy/scummvm/tree/ar2) and since I am unlikely to forget about it at this point I am closing this.

PR #930, opened 2017-03-30
GUI: Display game title at the top of the GMM

ashutosh97:


base/main.cpp
```diff
@@ -209,12 +208,14 @@ static Common::Error runGame(const EnginePlugin *plugin, OSystem &system, const
 	Common::String caption(ConfMan.get("description"));
 
 	if (caption.empty()) {
-		caption = EngineMan.findGame(ConfMan.get("gameid")).description();
+	    caption = EngineMan.findGame(ConfMan.get("gameid")).description();
```
criezy: We use tabs and not spaces for indentation.

Actually it doesn't look like any of the changes in this file do anything, so I would suggest to just drop all these changes from your commit.

engines/dialogs.cpp
```diff
@@ -20,13 +20,25 @@
  *
  */
 
+
+#include "base/version.h"
```
criezy: This header file is already included a few lines below.

engines/dialogs.cpp
```diff
@@ -47,6 +60,9 @@
 #include "gui/KeysDialog.h"
 #endif
 
+
+		
+
```
criezy: Why did you add those three blank lines?

engines/dialogs.cpp
```diff
@@ -78,9 +102,10 @@ MainMenuDialog::MainMenuDialog(Engine *engine)
 	GUI::StaticTextWidget *title = new GUI::StaticTextWidget(this, "GlobalMenu.Title", "ScummVM");
 	title->setAlign(Graphics::kTextAlignCenter);
 #endif
-
-	GUI::StaticTextWidget *version = new GUI::StaticTextWidget(this, "GlobalMenu.Version", gScummVMVersionDate);
+     
+	GUI::StaticTextWidget *version = new GUI::StaticTextWidget(this, "GlobalMenu.Version", caption.c_str()) ;
```
criezy: The description of the task we have on our wiki is "In addition to the ScummVM logo and version, how about showing the engine name and game title at the top of the GMM?" It doesn't say anything about removing the version.

Can you explain why you took the decision to remove the version? Is there is good reason for doing it?

base/main.cpp
ashutosh97: Done.

engines/dialogs.cpp
ashutosh97: Removed the duplicate one.

engines/dialogs.cpp
```diff
@@ -20,13 +20,25 @@
...
 #include "common/system.h"
 #include "common/translation.h"
+#include "common/archive.h"
+#include "common/debug.h"
+#include "common/debug-channels.h" 
+#include "gui/EventRecorder.h"
+#include "common/fs.h"
+#include "common/textconsole.h"
+#include "common/tokenizer.h"
+#include "common/osd_message_queue.h"
```
ashutosh97: Removed all the unnecessary header files.

engines/dialogs.cpp
ashutosh97: Got added by mistake. Removed them too.

engines/dialogs.cpp
ashutosh97: I removed the version name in the GMM and instead of that i added the game name. The reason behind doing so was that the version name is already displayed in the beginning when we launch ScummVM.exe , hence i thought that it would be useful if the game title is displayed when the user pauses the game by pressing Ctrl+F5. Displaying the game title would be useful in identifying the game which has been paused.

hpvb:
Have we come to a decision on whether or not removing the ScummVM version from GMM header is a good idea or not? 

In the way it is done here adding the version also on the non-scaled gmm everything becomes super-cramped. If we want to keep the version and the game name in there I think a more thorough redesign of the gmm dialog is in order. 

Does anyone have any ideas on that?

hpvb:
After discussion on IRC it seemed that:
 * Users are likely to know what game they are playing
 * Users are less likely to know what version of ScummVM they are on (particularly prepackaged ScummVM versions through retailers like GOG)
 * Most game names won't fit inside that space so it would look bad regardless

We're going to remove that task from the wiki.

Thanks for your submission though, please do keep contributing!

sev-:
You could add the full name to a tooltip

PR #932, opened 2017-04-01
MARIO: New engine

Retro-Junk:
Worst game ever for them, just another engine for ScummVM!

This is the long-awaited engine for a game that is not allowed to be named.
But today we break that rule and speak it aloud:

Plumbers Don't Wear Ties

... don't you agree? Either way, here you have it.

At this moment PC version of the game is playable few with minor glitches. Most noticeable one is broken mouse cursor. 3DO version will be supported in the future.

PS. By the way, this is not a joke.


dreammaster:
Really, there's not that much to review for this engine, given it's absurd simplicity. Code seems reasonably formatted. Probably best to remove the readme.txt

bluegr:
At first, I thought this was a joke, not only because of the size of the engine, but because of a total lack of video support or scene hotspots.... but then Strangerke told me that the game itself is a series of images, and you get multiple choice questions in each scene, which lead to other scenes.

There's not much to say, apart from the fact that this is the smallest engine ever! :P Best non-April fool's joke! :D 

There are some leftover debug statements scattered within mario.cpp, which should be removed. And the Mario & Luigi joke is priceless :)

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+			case Common::EVENT_LBUTTONDOWN:
+				if (_leftButtonDownFl) {
+					Common::Point mousePos = g_system->getEventManager()->getMousePos();
+					for (_curChoice = 0; _curChoice < _scenes[_curSceneIdx]._decisionChoices; _curChoice++) {
+						if (_scenes[_curSceneIdx]._choices[_curChoice]._region.contains(mousePos))
+							break;
+					}
+					if (_curChoice < kMaxChoice) {
+						debug("Accepting mouse click at %d : %d , choice = %d", mousePos.x, mousePos.y, _curChoice);
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+}
+
+void MarioGame::loadImage(const Common::String &dirname, const Common::String &filename) {
+	if (_image) {
+		delete _image;
+		_image = nullptr;
+	}
+
+	Common::String name = dirname + "/" + filename;
+	debug("%s : %s", __FUNCTION__, name.c_str());
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+		g_system->unlockScreen();
+		g_system->getPaletteManager()->setPalette(_image->getPalette(), 0, 256);
+		g_system->updateScreen();
+	}
+}
+
+void MarioGame::playSound() {
+	Common::String name = _scenes[_curSceneIdx]._sceneName + "/" + _scenes[_curSceneIdx]._waveFilename;
+	debug("%s : %s", __FUNCTION__, name.c_str());
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+		error("unable to load sound %s", name.c_str());
+
+	Audio::RewindableAudioStream *audioStream = Audio::makeWAVStream(file, DisposeAfterUse::YES);
+	Audio::AudioStream *stream = audioStream;
+	stopSound();
+	_mixer->playStream(Audio::Mixer::kSFXSoundType, &_soundHandle, stream, -1, Audio::Mixer::kMaxChannelVolume);
+}
+
+void MarioGame::stopSound() {
+	debug("%s", __FUNCTION__);
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+	if (_mixer->isSoundHandleActive(_soundHandle))
+		_mixer->stopHandle(_soundHandle);
+}
+
+void MarioGame::showScene() {
+	debug("%s : %d", __FUNCTION__, _curSceneIdx);
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+	_curBitmapIdx = _scenes[_curSceneIdx]._startBitmap;
+	loadImage(_scenes[_curSceneIdx]._sceneName, _bitmaps[_curBitmapIdx]._filename);
+	_actions.push(Redraw);
+	_setDurationFl = true;
+	_actions.push(PlaySound);
+}
+
+void MarioGame::updateScene() {
+	debug("%s : %d", __FUNCTION__, _curBitmapIdx);
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+			loadImage(_scenes[_curSceneIdx]._sceneName, _scenes[_curSceneIdx]._decisionBitmap);
+		}
+	} else {
+		loadImage(_scenes[_curSceneIdx]._sceneName, _bitmaps[_curBitmapIdx]._filename);
+		_setDurationFl = true;
+	}
+}
+
+void MarioGame::changeScene() {
+	debug("%s : %d", __FUNCTION__, _curChoice);
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+			g_system->getTimerManager()->removeTimerProc(onTimer);
+		} else {
+			_curSceneIdx = getSceneNumb(_scenes[_curSceneIdx]._choices[_curChoice]._sceneIdx);
+			_actions.push(ShowScene);
+		}
+	}
+}
+
+void MarioGame::processTimer() {
+	debug("%s", __FUNCTION__);
```
bluegr: Leftover debug code

engines/mario/mario.cpp
```diff
@@ -0,0 +1,332 @@
...
+	for (int i = 0; i < kMaxBitmaps; i++) {
+		_bitmaps[i]._duration = file.readSint16LE();
+		file.read(buf, kMaxName);
+		_bitmaps[i]._filename = Common::String(buf);
+	}
+	file.close();
+}
+
+int MarioGame::getSceneNumb(int sNo) {
+	debug("%s : %d", __FUNCTION__, sNo);
```
bluegr: Leftover debug code

bluegr:
Regarding the debug code: a debug level could be used (e.g. 1), so that the debug messages do not appear all the time in debug builds

criezy:
I have no issue with this engine being merged, but maybe it should be renamed first? Mario is fun as a joke, but I would prefer something that makes more sense, such as 'plumbers'.

PaulGilbert:
Seems like it should be ready to merge in.. the engine was renamed by Strangerke a week ago to something a bit more sensible.

bluegr:
IMHO, all of the commits should be renamed with the PLUMBERS prefix, otherwise this would look bad on commit history. I guess that this can be squashed into one commit, since the engine itself is really really (really...) small

dreammaster:
md5's idea is a good one.. in this case, given the engine simplicity, I too think it a good idea to squash the history into a single PLUMBERS: commit.

criezy:
This PR has now been around for almost a month and I have seen no objections to it being merged. If nobody object or ask for more time to discuss it within the next two days I will squash all the commits as suggested by @bluegr and @dreammaster and merge it on Sunday.

bluegr:
No objections from me, if the commits are squashed/renamed appropriately :)

wjp:
I was hoping this PR would die on its own. I don't want to be associated with this game, think it's in incredibly poor taste, and think it would be a mistake to merge it.

bluegr:
Well, this game is meant to be a spoof. People who would be interested in this game, would be able to play it. I find the game itself wacky and weird, but people who would like to play it, are free to do so.

I don't believe that engines should be rejected just because some of the ScummVM developers don't like the games that they support, unless there is a valid reason for the games themselves to be extremely out of scope of ScummVM (e.g. shoot'em'up games).

dreammaster:
I have to agree with md5. Whilst I concur with wjp that the game sounds pretty simplistic and silly, I don't think peoples' tastes should be a criteria for whether an engine is accepted. I personally dislike some of the games already on ScummVM.. I don't handle severe/pyschological horror very well, and from what little I've heard of IHNM, playing it would probably scar me for life. But I accept that it has it's place, and some people seem to enjoy it.

In fact, in the case of this engine, it really bears some significant similarities in genre to Voyeur.. after all, Voyeur is more or less just static images and movies simply keyed at different times and places in the mansion, with what you select/see determining the story (albeit with a little randomized selection at the start of killer). PLUMBERS is more simplistic and doesn't have FMV, but I don't see any other major fundamental differences that would allow Voyeur but preclude PLUMBERS.

rootfather:
In my opinion, ScummVM is all about preserving game history and make sure that the games are playable in the future.

I agree that PLUMBERS is a really bad game. But it is a game and thus should be preserved, even if it might be one of the worst games ever made (which might make the game somewhat interesting). But it's still a game that might even have some fans and we have an engine for it now. 

If you don't like it, that's perfectly fine. I don't think that it has a bad effect on the ScummVM project if we offer support for "bad" games, it more likely shows the diversity of the project.

In general, I support merging this engine to the project.

Strangerke:
I have fixed this morning the mouse cursor, which means the engine is now complete. I didn't add save/load, despite obviously it would be pretty easy, because the game may be finished in a reasonable time if you enable the skip screen option in the debugger. Reasonable = less than 2 minutes. The history still has to be fixed, which I guess would take me a matter of minutes if I was a bit more at ease with git.

Now I'd like to add a couple of notes about this engine.

The game supported is the (super-super rare) PC version, which is fully censored (at the opposite of the more recent (rare) 3DO version which shows a nipple in the censorship screen (I don't remember more than that, and I still wonder the use of the censor screen, maybe it's yet another of their jokes), and the ONLY animation played in the 3DO version isn't present in this version (thus the lack of video player).

Now, the negative point: even I have very mixed feelings about a merge.

On one hand, I love this game because it's completely surrealist and because it's clearly a joke. I told that thousands of times and I sincerely believe it, as it makes me laugh each time I play it. So obviously, now that an engine is complete, I would hate to see it rot.

Now on the other hand, despite I'm a crazy inclusionist, I have to admit that this game is *not* an adventure game. It's clearly a visual novel and it really doesn't fit our current genre scope.

So my mixed feelings are easy to understand: I'd love to see it merged so it doesn't rot, and I sincerely don't want it to be supported, or at least not as long as a game genre scope exists.

criezy:
As you may have noticed, despite what I wrote last week I have no yet merged the engine. This is because I wanted to give myself and others the time to think and comment about it. For several reasons I also have mixed feeling about a merge.

When I discussed merging the engine with sev this weekend he mentioned that scummvm-misc might be a better place for it, and I agree. One issue though is that scummvm-misc was apparently never migrated to git and still resides on sf.net/SVN. Also this would not prevent the engine from rotting (I would not be surprised for example if the Gargoyle engine doesn't work with the current ScummVM source code). Having those engines in one place though might help to increase visibility and find developers to keep them up to date. And despite the expanded scope of ScummVM i feel there is still place for a separate repository to house engines that do not quite fit into the ScummVM scope.

digitall:
Can we take a pragmatic view here... An pretty much complete engine has been offered for merge, it is tiny and is disabled by default so will only be built for nightly builds or if users wish to build it themselves. 

Artistically, it is a awful game, in the same way that "Plan 9 From Outer Space" is an awful movie, but there are cult fans of this and tastes vary.

While I tend to agree that PDWT could be considered a Visual Novel or some similar genre, this is a "No True Scotsman" argument and Myst and several other games with minimal inventory or FMV could be considered similar.

Personally, I would suggest merging and leaving it disabled from release builds for now.

criezy:
Despite some reservations, I have decided to merge the engine. Not merging it would have meant letting it die slowly (scummvm-misc is a graveyard at this point), and there seemed to be a majority wanting it to be merged.

At the very least it will serve as a very simple example of a full engine.

PR #935, opened 2017-04-02
GUI: fix for bug #9714: unnecessary tab in options

overlockeer:
In option menu the Control tab was showing. 
The only available option in the control tab was pointer speed, which does not affect users of mice.
Therefore the control tab was unnecessary

### DIFF
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -181,8 +181,8 @@ bool OSystem_SDL::hasFeature(Feature f) {
 #endif
 #ifdef JOY_ANALOG
 	if (f == kFeatureJoystickDeadzone) return true;
-#endif
 	if (f == kFeatureKbdMouseSpeed) return true;
+#endif
 	return ModularBackend::hasFeature(f);
 }
 


sev-:
The commit does not conform to our commit guidelines.

criezy:
Indeed, the commit message does not conform to our guidelines and needs to be changed. See http://wiki.scummvm.org/index.php/Commit_Guidelines

But beyond that I am not convinced this change is correct. Looking quickly at the code, it seems to me that the "keyboard mouse" code that uses this setting can be invoked even when JOY_ANALOG is not defined. I am not sure what would be a good indicator of whether the option is relevant or not. In the generic SDL case maybe checking the value of `ConfMan.getInt("joystick_num")` might be a better option since the SDL Joystick subsystem is only initialised when it is null or positive. But we might need to check derived classes to check if they use other inputs (for example keyboard) that might use this code as well.

And maybe rather than try to hide it a better solution would be to make the label of the option more precise than the current "Pointer Speed" so that we know it does not apply to the pointer moved by the mouse but only when the pointer is controlled by a joystick or keyboard.

criezy:
To expand a bit on my previous comment, I think the way to go here might be to modify OSystem_SDL::hasFeature for the kFeatureJoystickDeadzone and kFeatureKbdMouseSpeed to check if the SdlEventSource has a joystick setup (for example check if the _joystick pointer is not null).

csnover:
@overlockeer are you planning on working on this patch any further?

overlockeer:
I´m not planing on working on this any further, my interests have shifted since then.

PR #937, opened 2017-04-14
SDL: Only recreate SDL2 window when necessary

csnover:
Destroying and recreating the SDL window whenever the video mode changes in SDL2 is not necessary and causes several problems:

1. In windowed mode, the game window shifts position;
2. In fullscreen mode in macOS, every time the window is recreated, it causes the OS to play its switch-to-fullscreen animation again and emit system alert noises;
3. The window content flickers; and
4. The engine loses events from the old destroyed window.

This patch changes the SDL backend code to avoid destroying and recreating the SDL window when using SDL2, except when switching between OpenGL modes, which cannot be done without recreating the window.

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -489,13 +489,9 @@ bool OpenGLSdlGraphicsManager::setupMode(uint width, uint height) {
 		_glContext = nullptr;
 	}
 
-	_window->destroyWindow();
-
-	uint32 flags = SDL_WINDOW_OPENGL;
+	uint32 flags = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE;
 	if (_wantsFullScreen) {
 		flags |= SDL_WINDOW_FULLSCREEN;
-	} else {
-		flags |= SDL_WINDOW_RESIZABLE;
 	}
 
 	// Request a OpenGL (ES) context we can use.
@@ -503,11 +499,11 @@ bool OpenGLSdlGraphicsManager::setupMode(uint width, uint height) {
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, _glContextMajor);
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, _glContextMinor);
 
-	if (!_window->createWindow(width, height, flags)) {
+	if (!_window->createOrUpdateWindow(width, height, flags)) {
 		// We treat fullscreen requests as a "hint" for now. This means in
 		// case it is not available we simply ignore it.
 		if (_wantsFullScreen) {
-			_window->createWindow(width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE);
+			_window->createOrUpdateWindow(width, height, SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE);
 		}
 
 		if (!_window->getSDLWindow()) {
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -214,6 +214,10 @@ SurfaceSdlGraphicsManager::SurfaceSdlGraphicsManager(SdlEventSource *sdlEventSou
 
 SurfaceSdlGraphicsManager::~SurfaceSdlGraphicsManager() {
 	unloadGFXMode();
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+	if (_window)
+		_window->destroyWindow();
+#endif
 	if (_mouseSurface)
 		SDL_FreeSurface(_mouseSurface);
 	_mouseSurface = 0;
@@ -2644,13 +2648,11 @@ void SurfaceSdlGraphicsManager::notifyVideoExpose() {
 	_forceFull = true;
 }
 
-#ifdef USE_SDL_RESIZABLE_WINDOW
 void SurfaceSdlGraphicsManager::notifyResize(const uint width, const uint height) {
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	setWindowResolution(width, height);
 #endif
 }
-#endif
 
 void SurfaceSdlGraphicsManager::transformMouseCoordinates(Common::Point &point) {
 #if SDL_VERSION_ATLEAST(2, 0, 0)
@@ -2683,9 +2685,6 @@ void SurfaceSdlGraphicsManager::deinitializeRenderer() {
 
 	SDL_DestroyRenderer(_renderer);
 	_renderer = nullptr;
-
-	if (_window)
-		_window->destroyWindow();
 }
 
 void SurfaceSdlGraphicsManager::setWindowResolution(int width, int height) {
@@ -2746,7 +2745,7 @@ SDL_Surface *SurfaceSdlGraphicsManager::SDL_SetVideoMode(int width, int height, 
 		createWindowFlags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
 	}
 
-	if (!_window->createWindow(width, height, createWindowFlags)) {
+	if (!_window->createOrUpdateWindow(width, height, createWindowFlags)) {
 		return nullptr;
 	}
 
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.h
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.h
@@ -156,9 +156,7 @@ public:
 
 	// SdlGraphicsManager interface
 	virtual void notifyVideoExpose();
-#ifdef USE_SDL_RESIZABLE_WINDOW
 	virtual void notifyResize(const uint width, const uint height);
-#endif
 	virtual void transformMouseCoordinates(Common::Point &point);
 	virtual void notifyMousePos(Common::Point mouse);
 
--- a/backends/platform/sdl/sdl-window.cpp
+++ b/backends/platform/sdl/sdl-window.cpp
@@ -28,9 +28,12 @@
 
 #include "icons/scummvm.xpm"
 
+static const uint32 fullscreenMask = SDL_WINDOW_FULLSCREEN_DESKTOP | SDL_WINDOW_FULLSCREEN;
+
 SdlWindow::SdlWindow()
 #if SDL_VERSION_ATLEAST(2, 0, 0)
-	: _window(nullptr), _inputGrabState(false), _windowCaption("ScummVM")
+	: _window(nullptr), _inputGrabState(false), _windowCaption("ScummVM"),
+	_lastFlags(0), _lastX(SDL_WINDOWPOS_UNDEFINED), _lastY(SDL_WINDOWPOS_UNDEFINED)
 #endif
 	{
 }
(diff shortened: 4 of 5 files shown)


csnover:
I’m not sure why it didn’t occur to me earlier, but after talking a bit with @tsoliman I realised that recreating the window in SDL2 should be avoidable in all cases except switching to/from OpenGL. So, I will be updating this patch as a general enhancement to the game window behaviour with SDL2.

csnover:
This PR has been updated per the previous comment and is ready for review/landing. Thanks!

bgK:
I like this idea a lot, but there seems to be some bugs left, at least on Linux.
* In the launcher, in software mode (2x), when switching from the windowed mode to fullscreen and back again, the game list stays scaled and does not fit in the window. Perhaps the SDL Renderer needs to be updated to match the new window size.
capture d ecran de 2017-04-18 21-00-23
* In the launcher, when using OpenGL, when going back to the windowed mode from fullscreen, the window is maximized instead of going back to its previous size.

criezy:
I also like the idea.

But there is another issue: it does not compile when using SDL 2.0.4. After checking the documentation it appears that SDL_SetWindowResizable has been introduced in SDL 2.0.5. So if we want to change the resizable flag in SDL 2.0.4 or older it seems that we might need to recreate the window.

csnover:
Thanks for the feedback!

> * In the launcher, in software mode (2x), when switching from the windowed mode to fullscreen and back again, the game list stays scaled and does not fit in the window.
> * In the launcher, when using OpenGL, when going back to the windowed mode from fullscreen, the window is maximized instead of going back to its previous size.

These issues have been fixed by changing the order of operations so that the window size is set *after* changing the fullscreen mode.

When testing on Linux, I also noticed problems with SDL resetting the position of the window when it was programmatically resized (this does not happen on macOS), so `SdlWindow` now manages the window positioning as well. An unintended positive side-effect of this is that the window will reposition itself back to where it was after being recreated when switching to/from OpenGL mode, so the mode switch is less jarring.

> But there is another issue: it does not compile when using SDL 2.0.4. After checking the documentation it appears that SDL_SetWindowResizable has been introduced in SDL 2.0.5. So if we want to change the resizable flag in SDL 2.0.4 or older it seems that we might need to recreate the window.

I have added a version check so making windows resizable will now recreate the window on platforms with older SDL. (I sure wish they would follow semver and not add new features on patch releases!)

This should be working better now, so please give it another go and let me know how it is working.

bgK:
Thanks, the software path now works well for me. The OpenGL path still has an issue on Linux. When switching from windowed mode to fullscreen, the screen resolution changes to the size of the window. What is expected is the size of the window to change to match the desktop resolution.

criezy:
I am wondering if having a different code path for SDL >= 2.0.5 and for SDL <= 2.0.4 is a good idea. I got the feeling following some discussions on IRC that Windows users tend to hotswap the SDL DLL because of different bugs in different versions, and if a user tries to use the SDL 2.0.4 DLL (or lower) with a ScummVM released compiled with SDL 2.0.5, this is probably going to have an interesting behaviour. And if because of that the release is compiled with SDL 2.0.4 then we would not benefit from the change when using SDL 2.0.5 DLL.

As far as I can see the `SDL_WINDOW_RESIZABLE` is currently not used at all in surfacesdl, and is used in openglsdl when in windowed mode only. So the main issue is switching between windowed and fullscreen mode when using OpenGL. But what happens if we use `SDL_WINDOW_RESIZABLE` together with `SDL_WINDOW_FULLSCREEN`? If there is no adverse effects, then we would not need to recreate the window when switching between windowed and fullscreen mode, even with SDL 2.0.4 or lower, and we could use the same code for SDL 2.0.4 and 2.0.5 (always recreate the window if we need to change the flag). Recreating the window would then only occur when switching between surfacesdl and openglsdl (but it might already be the case anyway?).

The SDL documentation indicate that 
> You can't change the resizable state of a fullscreen window.

However I tried always using the SDL_WINDOW_RESIZABLE flag (so both for windowed and fullscreen modes) with openglsdl and everything seems to work perfectly. Creating the window with both seems to keep the resizable flag, and creating the window with SDL_WINDOW_RESIZABLE  and then switching to fullscreen also keeps it. So in windowed mode the window is resizable (as it should be) and in fullscreen mode it simply seems to be ignored.

criezy:
Note also that there seems to be an issue with the window positioning. On macOS X, when switching to windowed from fullscreen, the window ends up being in the top left corner (whatever the position of the window was before going to fullscreen). This is a bit annoying.

criezy:
The issue with window positioning is due to SdlWindow::destroyWindow() always saving the  position of the window before destroying it. It should only do it when the window is not fullscreen:
```
if (!(_lastFlags & (SDL_WINDOW_FULLSCREEN_DESKTOP | SDL_WINDOW_FULLSCREEN)))
	SDL_GetWindowPosition(_window, &_lastX, &_lastY);
```

This fixes the issue for me.

I should have mentioned I am still using SDL 2.0.4, and thus when switching between fullscreen and windowed it recreates the window (since the `SDL_WINDOW_RESIZABLE` changes).

bgK:
Using `SDL_SetWindowDisplayMode` to set the fullscreen window size instead of `SDL_SetWindowSize` fixes the fullscreen related issues for me on Linux, and removes the need to restore the window position: https://github.com/bgK/scummvm/commit/276732b5d1e5d958ced773fd2c2f9920b0d084ec.

csnover:
> I am wondering if having a different code path for SDL >= 2.0.5 and for SDL <= 2.0.4 is a good idea. […]

These are excellent points, so I will be updating this to get rid of the separate code paths.

> The issue with window positioning is due to SdlWindow::destroyWindow() always saving the position of the window before destroying it.

Oops :)

> Using SDL_SetWindowDisplayMode to set the fullscreen window size instead of SDL_SetWindowSize fixes the fullscreen related issues for me on Linux, and removes the need to restore the window position: bgK@276732b.

This seems to still cause the window to crawl down and to the right. Also, switching out of fullscreen in software mode without calling SDL_SetWindowSize afterwards causes SDL to render with the wrong size in the window, but then calling SDL_SetWindowSize at that point causes the window to move to (0,0). I will provide a little more information later on what I am seeing in each mode and API.

bgK:
> This seems to still cause the window to crawl down and to the right.

Even without calling `SDL_SetWindowPosition`? Saving and restoring the window position should not be necessary with `SDL_SetWindowDisplayMode`. The window manager should remember where the window was before entering fullscreen and put it back at the same place when leaving fullscreen. (If I understand the issue correctly, calling `SDL_SetWindowSize` while the window is leaving fullscreen messes up its position)

> Also, switching out of fullscreen in software mode without calling SDL_SetWindowSize afterwards causes SDL to render with the wrong size in the window, but then calling SDL_SetWindowSize at that point causes the window to move to (0,0). I will provide a little more information later on what I am seeing in each mode and API.

I think this is because the call to `SDL_GetWindowSize` in `SurfaceSdlGraphicsManager::SDL_SetVideoMode` happens before the window finishes going out of fullscreen.  In https://github.com/bgK/scummvm/commit/276732b, the `SurfaceSdlGraphicsManager::notifyResize` event handler is enabled to fix this issue.

csnover:
@bgK Oops (again), sorry! I had some tunnel vision and didn’t pay attention to everything in your patch (sheesh). With the resize event stuff, I also see the software mode working correctly on Linux (and continuing to work OK on macOS). Thank you for figuring that out. What you are saying makes sense.

There are a couple more problems I am struggling to communicate clearly (please ask questions if unclear), when using the OpenGL renderer:

1. When exiting fullscreen, the `SDL_SetWindowSize` call applies on macOS (so the window ends up being resized back to its “natural” size), but is ignored on Linux (so the window ends up staying whatever size it was prior to the fullscreen switch); this ought to be decided one way or the other and fixed to be consistent. The second one seems intuitively more correct to me (and is probably easier to implement by just not calling `SDL_SetWindowSize` when exiting fullscreen), but I’m open to whatever.
2. When a game calls to `initGraphics`, the window gets reset to whatever size the game requested, even if the user had previously resized their window, and even if the new size is the same as the previous size passed to `initGraphics`. It seems to me like there should be some better behaviour than now, like:
   * only resizing the window if the game content is larger than the current window
   * only resizing when the engine first starts up
   * only resizing if the user hasn’t changed the window size (and then restores automatic resizing if they hit some hotkey to reset it, like the ctrl-alt-± hotkeys)
   * something else

   In the case of SCI32, `initGraphics` is called during games but only to change the bit depth, so at minimum I’d like to get rid of the behaviour that resets the window size in that particular case, by:
   * passing a special value like `0` to `initGraphics` for the width/height parameters
   * having either the renderer or SdlWindow ignore resizing to a size that it already is
   * some better idea someone else comes up with

   This seems like a bigger problem with e.g. DreamWeb DOS CD that has a 640x480 start screen but then runs at 320x200 (apparently with no PAR correction).

That last one is kind of a lot of stuff, and maybe should get hashed out in a separate place, to avoid feature creep here. Let me know your thoughts.

I will update this PR in the morning to reflect the changes from @bgK.

Thanks,

csnover:
The PR has been updated now with the latest code.

bgK:
Thanks!

> When exiting fullscreen, the SDL_SetWindowSize call applies on macOS (so the window ends up being resized back to its “natural” size), but is ignored on Linux (so the window ends up staying whatever size it was prior to the fullscreen switch); this ought to be decided one way or the other and fixed to be consistent. The second one seems intuitively more correct to me (and is probably easier to implement by just not calling SDL_SetWindowSize when exiting fullscreen), but I’m open to whatever.

I like the idea of removing the call to `SDL_SetWindowSize` to keep the manual resizes when leaving fullscreen. However we probably need to keep it in software mode so that going out of fullscreen and changing the window size at the same time using the options dialog works.

> When a game calls to initGraphics, the window gets reset to whatever size the game requested, even if the user had previously resized their window, and even if the new size is the same as the previous size passed to initGraphics.

I have no strong opinion regarding your proposals or if it should be done here or as a separate change. The following one seems to me to be the least surprising for both users and engine authors:
> * having either the renderer or SdlWindow ignore resizing to a size that it already is

tsoliman:
Any objections to merging this?

bgK:
> Any objections to merging this?

The changes look good to me except the following use case that is currently broken:

> going out of fullscreen and changing the window size at the same time using the options dialog

Some testing on Windows would be nice as well.

hpvb:
Looks good to me, but I haven't actually ran it. Would be a welcome change for debugging too, no more rechecking 'always on top' :)

bgK:
While testing on Windows, I noticed the following use case is not working:
* Have ScummVM configured for OpenGL + Fullscreen
* Start ScummVM, press Alt+Enter to switch to windowed mode
* The window stays in fullscreen mode. This is because, SDL was never told the window size to use when not in fullscreen mode.

Otherwise, this PR works well for me on Windows.

csnover:
That problem on Windows with default OpenGL fullscreen was a problem on macOS too. The `SDL_SetWindowSize` call was being avoided when switching away from fullscreen to keep from resetting the dimensions the user had previously set the window to, but the real problem is in the OpenGL backend, not this window code. So, I have updated this PR to remove that fullscreen check, and if it still seems to be working at least as well as the current master branch, it can finally be landed.

With regards to that remaining OpenGL issue, I guess it needs to be dealt with separately in another ticket/PR, since right now it is completely ignoring the user’s requested window size (it seems to be using only the gameWidth/gameHeight from `_currentState`, which is only ever updated by `initSize`), and even drops out of fullscreen on any new `initGraphics` call. This could be due to incomplete support of SDL2 in that code, since it seems like in `OpenGLSdlGraphicsManager::notifyResize` that there is stuff in the SDL1 branch intending to pay attention to user-initiated resize, but no such similar code in the SDL2 branch.

bgK:
So .. unfortunately starting ScummVM in fullscreen + OpenGL mode and then togging to windowed mode still does not work on Linux/X11 :sob:.

The call to `SDL_SetWindowSize` seems to be ignored, the window keeps the fullscreen size. The (ugly) workaround I've found is to first create a window at the default dimensions and then toggle it to fullscreen. I don't know if there is a better solution, but I couldn't find one when fiddling with the code.

Here is a patch for the workaround. Feel free to use it, ignore it or amend it as you like: https://gist.github.com/bgK/c4afdbfa41a1171f2f4cae17e1dff1e2

Otherwise, this is now working very well :+1:

criezy:
That looks good and works well for me now.

That workaround from bgK for Linux might not look very pretty, but it also looks quite small and localised, so I think I would go with it. But it's up to you.

csnover:
I like @bgK’s patch. It is small and well-documented. There is a lot of complexity in the graphics backend code so it is really nice that it ended up being so simple. I’m going to pull those changes into the branch and then push to master in a few moments. Thanks!

csnover:
Landed in 332fabcb8a0ff3957f0198ad3b93b4f3861f7eba. Thank you both again for your suggestions, fixes, and testing.

PR #938, opened 2017-04-29
TESTBED: Fix building with --disable-cloud

tsoliman:
The linker fails when building with `--disable-cloud` and not passing `--disable-libcurl`


sev-:
Thanks!

PR #939, opened 2017-04-29
Clean up renderText

tobiatesan:
**NOTE: This has a conflict with #940, I have opened #941 which is #940 + #939 with the conflicts resolved.**
If you have no objections to either please just merge #941.
----------------------

While implementing preloading of script text I've done a minuscule bit of housekeeping on renderText (more to come later).

This should be almost exactly semantically equivalent, except that now the block with ``warning("Requested to draw on an empty surface:...)`` uses the ``width``, ``height`` values that are going to be used in the rest of the method body, which I suppose is actually desired.

### DIFF
--- a/engines/director/frame.cpp
+++ b/engines/director/frame.cpp
@@ -742,6 +742,38 @@ void Frame::renderText(Graphics::ManagedSurface &surface, uint16 spriteId, Commo
 		return;
 
 	TextCast *textCast = _sprites[spriteId]->_buttonCast != nullptr ? (TextCast*)_sprites[spriteId]->_buttonCast : _sprites[spriteId]->_textCast;
+	int x = _sprites[spriteId]->_startPoint.x; // +rectLeft;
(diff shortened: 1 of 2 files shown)


tobiatesan:
Ah, nevermind, wait a second.

tobiatesan:
Seconds after opening the PR I noticed that what is being read from the stream, besides the actual text, appears to be basically a `struct MacFontRun`, so maybe that's what it should return...

tobiatesan:
Or, in fact, a MacFontRun, since it has a `Common::String text`, althought I'm not sure what it's for.

tobiatesan:
No, probably `MacFontRun` should not be involved after all.

tobiatesan:
**NOTE: This has a conflict with #940, I have opened #941 which is #940 + #939 with the conflicts resolved.**
If you have no objections to either please just merge #941.

tobiatesan:
I'm closing this one so as not to create confusion, since I do believe you want just #941.

PR #943, opened 2017-04-30
DIRECTOR: Preload Stxts data in textCast

tobiatesan:
***EDIT*: I've opened #944, which includes this, which, in turn, includes #942, and adds caching of MacTexts.
If you like all changes please merge that one.**
------------------------------
This is #942, with, additionally, loading of Stxts into textCasts rightaway in Score.

If you like it please merge this one.

Cached MacTexts are coming soon.

PR #944, opened 2017-05-01
DIRECTOR: Cache Mactexts

tobiatesan:
This includes #943 and adds caching of MacTexts (one per textCast) on top.
See how you like it; consider merging this one directly if you like it.


sev-:
Merging this, however, you're using #include directives incorrectly. Please take a look at other files in the engine or watch my commit which will follow

PR #945, opened 2017-05-02
SCI32: Set a savegame description if none was entered

tsoliman:
This fixes the the SCI32 incarnation of Trac#5343
(defect#3061964): Savegames with no name can't be restored


tsoliman:
Updated to fix Torin

tsoliman:
merged as f30f34cbde

PR #946, opened 2017-05-02
GUI: remove addDirtyRect() that kept clearing objects above a list

Joefish:
The partial invalidation of the screen cut off objects that were located
above a list, like the build time / git hash on the main screen. Removing
addDirtyRect() fixed the bug and I didn't find any case where artifacts
were left behind.

BUG: #6394 (GUI: "List View" save page draws over font)

### DIFF
--- a/gui/ThemeEngine.cpp
+++ b/gui/ThemeEngine.cpp
@@ -298,10 +298,6 @@ void ThemeItemDrawDataClip::drawSelf(bool draw, bool restore) {
 			_engine->renderer()->drawStepClip(_area, _clip, *step, _dynamicData);
 		}
 	}
-
-	extendedRect.clip(_clip);
-
-	_engine->addDirtyRect(extendedRect);
 }
 
 void ThemeItemTextData::drawSelf(bool draw, bool restore) {


Joefish:
While it fixes the bug it also removes functionality that other ::drawSelf() implementation have (ThemeEngine.cpp:270 - 360) without corrupting already drawn elements on screen. Removing it I expected artifacts when selecting games in the list, but I couldn't observe any difference..
I most likely missed some use case that my commit breaks so I would appreciate comments on this change, anything that helps me understand the system better.

criezy:
Are you sure this is a good idea? I am guessing you will be missing updates now.
Actually I just tried it to verify and now the scrollbar doesn't move when scrolling in the game list!

I remember I had a quick look at this bug in the past and I think I tracked it down to the shadow not being correctly handled. I think I also made some quick test to check that splitting the _backgroundOffset into shadow and bevel offsets and applying the shadow offset only to the bottom and right or the rect was fixing this bug (or more precisely extending the rect by the bevel offset and then extending the bottom and right edges only by the shadow - bevel offset). I never had time to check that this was the proper fix and didn't have side effects.

Joefish:
damn you 24" for giving my programs too much space so I won't see scrollbars most of the time!

@criezy I was quite sure that I missed something obvious as those updates were most likely not added for fun.. Crawling up the callstack and seeing just getBossClipRect() (and seeing nothing wrong with it) getting passed I wasn't sure where to look next without something that my commit breaks.
I'll look into your approach. Thanks.

Also I hope it isn't a problem that I try to fix this bug now (it annoyed me for quite a while already) as sev told maxyodedara5 to fix it iirc but I haven't seen him on IRC for ~3 weeks..

criezy:
I have now opened PR #959 with the suggestion I made in my previous comment.

criezy:
Closing since this was not a good approach and an alternative one was committed a month ago.

PR #948, opened 2017-05-12
BACKENDS: Compress screenshots using PNG if available

csnover:
Following a discussion on IRC about compressing screenshots, I decided to see how challenging it would be to implement PNG compression in builds that include libpng. The answer seems to be: not very.

For the moment, the function for writing to a PNG stream accepts only RGB888 surfaces, which should be fine since this is what was already happening with BMPs.

For builds without PNG support, this code continues to fall back to saving as bitmaps.

I do not have a lot of faith that this code works properly with big endian systems, so would really appreciate it if someone smarter could look at that and/or test on a BE system. I’m also not sure that the shift values for the expected pixel format are correct (they don’t seem to match SDL’s SDL_PixelFormat shift values for the converted surface—could this be due to my running on an LE system and having that endianness exposed by SDL and not by Graphics::PixelFormat?).

I also had to change the dependency order between libgraphics and libimage to get gcc to link successfully, which might break other things (I’m hoping Travis will tell me that when I open this PR). If it does break, there seem to be some alternative approaches that could be used.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -24,8 +24,8 @@ MODULES += \
 	backends \
 	engines \
 	video \
-	graphics \
 	image \
+	graphics \
 	audio \
 	common \
 	po
--- a/backends/graphics/opengl/opengl-graphics.cpp
+++ b/backends/graphics/opengl/opengl-graphics.cpp
@@ -28,6 +28,7 @@
 #include "backends/graphics/opengl/pipelines/shader.h"
 #include "backends/graphics/opengl/shader.h"
 
+#include "common/array.h"
 #include "common/textconsole.h"
 #include "common/translation.h"
 #include "common/algorithm.h"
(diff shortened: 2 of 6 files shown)


wjp:
Tested on Linux on a BE PPC G4; png output seems fine.

criezy:
I had a quick look and the changes look good to me. This also look like a feature we might want. So I have no objections for this PR being merged.

I also tested with a PPC mac build and the result is correct both for OpenGL and SurfaceSDL graphics modes.

bgK:
Looks good to me 👍 

Is it useful to keep the bitmap version of the code around? Most ports should have libpng.

csnover:
> Is it useful to keep the bitmap version of the code around? Most ports should have libpng.

The bitmap writing code is very trivial, so there is not much (or any?) maintenance burden keeping it around. I would think that only if libpng is changed to a mandatory dependency that the bitmap code should be removed.

It sounds like there is a general positive consensus about this change; I will keep this PR open for responses until Sunday, and if there are no new objections, I will plan to merge it then.

criezy:
I agree that it is probably best to keep the BMP code around as long as libpng remains optional.

PR #949, opened 2017-05-21
ENGINES: Allow detection entries to match on full paths

csnover:
This allows an engine to match files that exist multiple times
in the same game directory with the same basename.

For example, different releases of Torin's Passage in SCI engine
come with zero or more GERMAN, FRENCH, ENGLISH, etc. directories,
all containing files with the same basenames but with different
contents per language. Because the allFiles map used only the
basename of a file as a key, it could not match more than one of
these localization directories, which made it impossible to select
from all the possible languages.

See https://github.com/csnover/scummvm/commit/425dd7ed19a32611b3cba1b0f812aac496602b61 for an example implementation.

This all came about because there are multiple releases of Torin with different available localizations but with the same main resource bundles, so I could not figure out any other way to differentiate these releases without making the detector capable of matching paths instead of just basenames:

* The GOG.com release has the same multi-language EN/FR/DE resource bundles as the multi-language English audio CD release, but it does not include any of the translated message directories for some reason, so without subdirectory matching (to exclude the non-existing localizations), people adding the GOG release would end up getting options to choose languages that do not exist in this release
* Based on the detection entries that were already in the detection table, the Spanish and Italian releases also use the same EN/FR/DE resource bundles and English audio as the English multilanguage release, and just include some new message resources with the ES/IT translations. Without subdirectory matching, the same problem occurs here where people with both the GOG release *and* the EN/FR/DE release will get options for ES/IT translations which do not exist

Refs Trac#9772.

### DIFF
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -345,7 +345,7 @@ void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFileP
 	g_system->logMessage(LogMessageType::kInfo, report.c_str());
 }
 
-void AdvancedMetaEngine::composeFileHashMap(FileMap &allFiles, const Common::FSList &fslist, int depth) const {
+void AdvancedMetaEngine::composeFileHashMap(FileMap &allFiles, const Common::FSList &fslist, int depth, const Common::String &parentName) const {
 	if (depth <= 0)
 		return;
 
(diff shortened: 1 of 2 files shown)


hpvb:
Tested by merging the PR locally and running a mass add on my scummvm game collection. Everything was detected as usual.

PR #951, opened 2017-05-22
SCI: Add 'music/' subdir to file path for LSL7 GOG

hpvb:
LSL7 originally came with music in two qualities in subdirectories of
the 'music' directory. 22s8 and 22s16. The gog.com release doesn't do
this and only ships the hq music in the root of 'music/'. This PR adds
it to the search path.


csnover:
Legit. Thanks!

PR #952, opened 2017-05-25
GUI: Fully hide GMM save/load buttons if not supported by engine

wjp:
Previously, they were only greyed out. According to the discussion on
patch #8925, this was a limitation of the GUI layout engine when these
buttons were introduced in 2008. This no longer seems to be a problem.

Are there are any potential issues with changing this that I haven't considered? I don't remember the exact situation from 9 years ago.


bluegr:
IIRC, the only reason why these were greyed out instead of hiding then was a GUI limitation. I think that the GUI was unable to dynamically hide widgets once they were drawn.

If there's no issue with the GUI itself, I believe this is the correct way to go, and it makes much more sense GUI-wise for the users

bgK:
The buttons probably need to be both made invisible and disabled, or otherwise fixed, because it is still possible to click them even if they are not visible.

wjp:
@bgK: thanks, changed that.

hpvb:
Tested with dreamweb, works. Maybe we want to do a separate PR to change the spacing of the dialog though?

PR #953, opened 2017-05-26
SCI: Fix kArrayFill

hpvb:
The ScummVM implementation of class SciArray::fill() has a bug where it
will overwrite the array[index] with the value count times, rather than
fill the array starting from index count times.

This patch fixes that behavior. This was noticed because the LSL7 dice
game was broken, it was impossible to lose. After this patch the dice
game works as expected.


csnover:
I tweaked the patch to increment the target pointer instead of changing the loop. Thanks for finding and fixing the bug, great work!

PR #954, opened 2017-05-26
SCI: Various breakpoint improvements

wjp:
This improves breakpoints in a number of ways:

* Fix regression from 0f9c33e02f1cb2c740c1eb0dcaad96dd22ec29e7 that broke wildcard selector breakpoints
* Add `bpw`/`bpr` handling to property access opcodes (`aTop` and such). (Catching property reads/writes inside kernel functions is still not done.)
* Add new `bp_action` (alias `bpact`) command to set the action to perform when an individual breakpoint is triggered:
  * `break`: break into debugger (the default)
  * `log`: only log the breakpoint, but don't break
  * `bt`: show a backtrace, but don't break
  * `inspect`: show the current obj (`bpw`, `bpr`, `bpx` only), but don't break
  * `none`: ignore this breakpoint

### DIFF
--- a/engines/sci/console.cpp
+++ b/engines/sci/console.cpp
@@ -3722,54 +3722,55 @@ bool Console::cmdLogKernel(int argc, const char **argv) {
 	return true;
 }
 
+void Console::printBreakpoint(int index, const Breakpoint &bp) {
+	debugPrintf("  #%i: ", index);
+	const char *bpaction;
+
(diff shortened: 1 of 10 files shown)


bluegr:
Overall, these are very welcome changes to the SCI debugger. Nice work :)

wjp:
Merged this into master manually.

PR #956, opened 2017-06-01
ENGINES: Display savestate date as YYYY-MM-DD

tsoliman:
This avoids confusion in locales that use MM/DD/YYYY vs DD/MM/YYYY

thanks @csnover for suggesting dashes (like ISO-8601)


bgK:
You may want to change the default save description to use dashes as well for consistency:
https://github.com/scummvm/scummvm/blob/e8b70a4686916774a1bf499c9c3136e3052b6b43/gui/saveload.cpp#L70

tsoliman:
I plan to merge this sometime this weekend unless anyone objects

PR #957, opened 2017-06-02
GUI: Fix modern theme savegame thumbnail label layout

tsoliman:
The Save/Load screen has a preview area with labels for date and time.
The bottom label was drawing past the bottom of the rect.

Before:
image

After:
image


tsoliman:
This is wrong. There's a problem in `SaveLoadChooserSimple::reflowLayout()`: it doesn't realize there are 2 lines of text coming from `_saveDateSupport`

PR #958, opened 2017-06-02
GUI: Fix savegame thumbnail label layout

tsoliman:
The Save/Load screen has a preview area with labels for date and time.
The bottom label was drawing past the bottom of the rect.

There's 2 sets of optional labels based on whether or not the engine supports saving playtime and/or savedates

This gives four cases:

|_saveDateSupport|_playTimeSupport|Before|After|
|----|----|----|----|
|true|true|before1|after1|
|false|false|before2|after2|
|true|false|before3|after3|
|false|true|before4|after4|


tsoliman:
The last commit further improves both the look and the logic a little bit. The theme rect is now rigged for no labels and is expanded in code based on how many labels are added (instead of shrinking based on how many labels aren't added).

This possibly removes assumptions in the theme about line height.

Here's the new comparisons (fine tuning can be done in the theme based on the label-less case)

|Before|After(shrink)|After(expand)|
|----|----|----|
|before1|after1|after1a|
|before2|after2|after2a|
|before3|after3|after3a|
|before4|after4|after4a|

bgK:
Just tried the changes. Looks good :+1:

tsoliman:
I plan to merge this sometime this weekend unless anyone objects

PR #959, opened 2017-06-05
GUI: Separate bevel and shadow effect when extending widget rect

criezy:
When widget::draw() is called it asks the ThemeEngine to redraw the background first and then the widget gets redrawn in drawWidget(). The ThemeEngine uses an extended rect to restore the background to include bevel and shadow effects. However if this extended rect overlaps with other widgets, since those other widgets are not redrawn, a part of those will be missing. See for example bug #6394: GUI: List View save page drawns over font.

In case we get overlap we might need to change the way widgets are drawn so that all widgets intersecting the area where the backgroud is restored are redrawn. This commit simply seperate the bevel and shadow effects, and uses the shadow offset only to extend the bottom and right sides of the rectangle (while the bevel offset is still used to extend all four sides). This results in a smaller extended rectangle (if the shadow offset is bigger than the bevel offset, which is the case of the list view) and thus decrease the risk of the issue happening. The particular cases described in bug #6394 are all fixed with this change.

Do you see potential issue with this change? In particular is this OK to assume that the shadow is only drawn on the bottom and right side of widgets?

### DIFF
--- a/gui/ThemeEngine.cpp
+++ b/gui/ThemeEngine.cpp
@@ -87,6 +87,7 @@ struct WidgetDrawData {
 	    E.g. when taking into account rounded corners, drop shadows, etc
 	    Used when restoring the widget background */
 	uint16 _backgroundOffset;
+	uint16 _shadowOffset;
 
 	bool _buffer;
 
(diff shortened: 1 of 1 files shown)


criezy:
I have been using this for the past week and have not noticed any regression. So unless there are objections I am planning to merge this early next week.

criezy:
I have now pushed this commit to master.

PR #960, opened 2017-06-09
Windows: Use GetUserDefaultUILanguage() instead of GetThreadLocale().

res2k:
The thread locale concerns display options (e.g. date formatting), not
the display language. They are typically, but not necessarily the same,
as Windows allows them to be configured separately.

### DIFF
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -432,7 +432,7 @@ Common::String OSystem_SDL::getSystemLanguage() const {
 	char langName[9];
 	char ctryName[9];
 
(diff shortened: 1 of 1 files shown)


digitall:
@Kirben : Any comments on this?

Since this is a pretty small and self-contained change and looks reasonable, can this just be merged?

digitall:
@sev- : Since this is a small change, it would be good to get this reviewed and merged. Is there anyone apart from Kirben who could look at this?

Kirben:
Yes, the language code changes look fine, and make sense.

sev-:
Thanks, merged

PR #961, opened 2017-06-21
DIRECTOR: Refactor lingo-events.cpp so that 1-ary processEvent is the unique entry point to events

tobiatesan:
This also does slight changes to semantics that follow from the docs and stubs primaryEvent and MovieScript handling.


sev-:
Great, it all makes sense, even with the slightly limited functionality. Mergining it for impovements.

sev-:
But there are conflicts now. Could you please rebase?

PR #962, opened 2017-06-21
JANITORIAL: Hug net-specific stuff in module.mk in ifdef USE_CLOUD

tobiatesan:
I'm not the most familiar person so this *could* be wrong, but things like networking/sdl_net/handlers/filesajaxpagehandler.o look positively like cloud stuff to me.

### DIFF
--- a/backends/module.mk
+++ b/backends/module.mk
@@ -56,7 +56,6 @@ MODULE_OBJS += \
 	cloud/onedrive/onedrivelistdirectoryrequest.o \
 	cloud/onedrive/onedriveuploadrequest.o
 endif
-endif
 
 ifdef USE_LIBCURL
 MODULE_OBJS += \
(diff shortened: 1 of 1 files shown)


Tkachov:
Well, `configure` looks like it defines `USE_CLOUD` if either libcurl or sdl_net are used. As networking/sdl_net, obviously, is included only when sdl_net is used (== `USE_SDL_NET` is defined), `USE_CLOUD` is also defined, so this change won't break anything.

The thing why these networking/ classes were not under `USE_CLOUD` is because they are not exactly cloud classes (which are in cloud/). curl requests might be used for something else, not cloud integration, and webserver might be used without cloud to manage files via LAN.

tobiatesan:
> Well, configure looks like it defines USE_CLOUD if either libcurl or sdl_net are used.

Sorry, this is not really necessary, I had broken create_project (https://github.com/scummvm/scummvm/pull/963#discussion_r123338610) and I drew the wrong conclusion altogether.

PR #963, opened 2017-06-21
CREATE_PROJECT: Add fix for nested ifs

tobiatesan:
This adds a quick ~~workaround~~ fix so that any if blocks nested inside a if block with an unmet condition are handled with push(false) even if their condition is satisfied.

For example, before adding this, upon running create_project.exe --msvc --disable-cloud the block inside

ifdef USE_CLOUD
ifdef USE_LIBCURL
... 
endif
endif

would get evaluated (with unpredictable results) since it was inside USE_LIBCURL (=1).

It now works for me ~~(after applying #962).~~

### DIFF
--- a/devtools/create_project/create_project.cpp
+++ b/devtools/create_project/create_project.cpp
@@ -1770,18 +1770,20 @@ void ProjectProvider::createModuleList(const std::string &moduleDir, const Strin
 			if (std::find(defines.begin(), defines.end(), *i) == defines.end())
 				shouldInclude.push(false);
 			else
-				shouldInclude.push(true);
+				shouldInclude.push(true && shouldInclude.top());
 		} else if (*i == "ifndef") {
 			if (tokens.size() < 2)
 				error("Malformed ifndef in " + moduleMkFile);
 			++i;
 
 			if (std::find(defines.begin(), defines.end(), *i) == defines.end())
-				shouldInclude.push(true);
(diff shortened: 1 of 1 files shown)


devtools/create_project/create_project.cpp
```diff
@@ -1770,18 +1770,20 @@ void ProjectProvider::createModuleList(const std::string &moduleDir, const Strin
...
 			if (std::find(defines.begin(), defines.end(), *i) == defines.end())
-				shouldInclude.push(true);
+				shouldInclude.push(true && shouldInclude.top());
 			else
 				shouldInclude.push(false);
 		} else if (*i == "else") {
-			shouldInclude.top() = !shouldInclude.top();
+			bool last = shouldInclude.top();
+			shouldInclude.pop();
+			shouldInclude.push(last && shouldInclude.top());
```
tobiatesan: Whops, missing a !

tobiatesan: Fixed.

wjp:
This looks fine to me. I'm just not sure why you call it a workaround?

tobiatesan:
Yes, that goes without saying; as for `true`, I kind of like that the logic is made more explicit that way.

Works for me, if no-one complains I'd merge...

criezy:
The logic looks OK to me. I haven't tried either, but I think you can merge (after squashing).

PR #964, opened 2017-06-22
SCI32: Proper implementation of kObjectIntersect

bluegr:
This a rarely used kernel function, which is needed for the arcade
games in PQ4.

Fixes Trac #9855

### DIFF
--- a/engines/sci/engine/kgraphics32.cpp
+++ b/engines/sci/engine/kgraphics32.cpp
@@ -245,9 +245,8 @@ reg_t kSetPalStyleRange(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kObjectIntersect(EngineState *s, int argc, reg_t *argv) {
-	Common::Rect objRect1 = g_sci->_gfxCompare->getNSRect(argv[0]);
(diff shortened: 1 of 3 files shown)


bluegr:
This implementation is what was done in the original, it has been tested, and the bug reporterin #9855 reported that all the arcade sequences in PQ4 are working with this.

If nobody objects, I'll merge this on Friday

wjp:
At first glance the error handling here doesn't seem to be correct for games >= SCI21MIDDLE. I'm guessing you didn't check the full range of sci interpreters?

bluegr:
That is quite interesting. I haven't seen any difference in error checking in this code between SCI2.1 early and SCI3.

Could you be a bit more specific? Is something missing? Should there be extra checks for newer SCI versions, which I've missed?

wjp:
Since I've now done the necessary research myself anyway, I've just directly implemented this as 820caf370e2c24e6aefd8b1b90227ca29564bab6.

bluegr:
Interesting. I did not check kSetNowSeen for this. Thanks for your work! :)

PR #966, opened 2017-07-02
MOHAWK: Overhaul the Riven engine

bgK:
This pull request overhauls the Riven engine. The most notable changes are:
- Restructured the engine to more closely match the original. A lot of the code is still there, just moved around and tweaked.
- Added the transition effects between cards
- Added the flies / fireflies effect used in some locations
- Reworked script execution to use a queue so that scripts are run in the same order as the original
- Checked (and amended when necessary) the hardcoded external commands against the original engine
- Fixed minor bugs

With these changes I'm confident Riven can enter testing soon.


engines/mohawk/configure.engine
```diff
@@ -2,5 +2,5 @@
 # add_engine [name] [desc] [build-by-default] [subengines] [base games] [deps]
 add_engine mohawk "Mohawk" yes "cstime myst riven" "Living Books" "highres"
 add_engine cstime "Where in Time is Carmen Sandiego?" no
-add_engine riven "Riven: The Sequel to Myst" no "" "" "16bit"
+add_engine riven "Riven: The Sequel to Myst" yes "" "" "16bit"
```
sev-: Build by default? Is it intended or just a leftover?

bgK: This is on purpose. I'd like Riven to be part of the 1.10 release, possibly publishing the testing announcement a few weeks after Starship Titanic.

sev-: I see. Though, nothing prevents you from marking it right before the announcement.

Merging.

PR #967, opened 2017-07-03
COMMON: Change way the Singleton instances are instantiated

criezy:
When compiling the current code with clang from a recent XCode version on macOS we get tons of warnings like the following one:
```
./common/singleton.h:70:8: warning: instantiation of variable 'Common::Singleton<GUI::EventRecorder>::_singleton' required here, but no definition is available     [-Wundefined-var-template]
                if (!_singleton)
                    ^
backends/platform/sdl/sdl.cpp:122:9: note: in instantiation of member function 'Common::Singleton<GUI::EventRecorder>::instance' requested here
        delete g_eventRec.getTimerManager();
              ^
./gui/EventRecorder.h:51:41: note: expanded from macro 'g_eventRec'
#define g_eventRec (GUI::EventRecorder::instance())
                                        ^
./common/singleton.h:88:12: note: forward declaration of template entity is here
        static T *_singleton;
                  ^
```

One way to fix those warnings is to replace all the specialisation instantiations (using the DECLARE_SINGLETON() macro) with instantiation of the template variable in singleton.h directly.

This is what this PR does, and I have tested that it compiles and works with gcc as old as 4.0 as well as clang. Somebody should check that Visual Studio is happy with this code as well before the PR gets merged.

### DIFF
--- a/backends/cloud/cloudmanager.cpp
+++ b/backends/cloud/cloudmanager.cpp
@@ -32,11 +32,6 @@
 #include "backends/networking/sdl_net/localwebserver.h"
 #endif
 
-namespace Common {
-
-DECLARE_SINGLETON(Cloud::CloudManager);
-
-}
 
 namespace Cloud {
 
--- a/backends/fs/ds/ds-fs-factory.cpp
+++ b/backends/fs/ds/ds-fs-factory.cpp
@@ -28,9 +28,6 @@
 #include "backends/fs/ds/ds-fs.h"
 #include "dsmain.h" //for the isGBAMPAvailable() function
 
-namespace Common {
-DECLARE_SINGLETON(DSFilesystemFactory);
-}
 
 AbstractFSNode *DSFilesystemFactory::makeRootFileNode() const {
 	if (DS::isGBAMPAvailable()) {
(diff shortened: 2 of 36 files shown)


criezy:
On IRC tsoliman has reported that it works with gcc 3.4.4 and dreammaster has reported that it works with Visual Studio 2015.

If there is no objection I would like to land this. Or maybe somebody wants to also check with an older version of Visual Studio? I will wait a few more days, and unless I hear objections will merge this.

sev-:
OK, I'm merging this. Let's give the buildbot a spin.

Thierry, great work!

sev-:
Nope, it did not work. Sorry, I had to revert.

ConfigManager is not a singleton anymore, and is losing values within the engines.

criezy:
The reason it didn't work for @sev is because he was using plugins, and because we are restricting the list of exported symbols for plugins on macOS: https://github.com/scummvm/scummvm/blob/master/configure#L3650
Removing `-exported_symbol_list "$(srcdir)/plugin.exp"` makes it work properly.

A possible issue here is that it looks like we do the same for elf with `--retain-symbols-file,$(srcdir)/backends/plugins/elf/plugin.syms` but I don't know if this has the same consequences.

I am wondering if it is worth pursuing this singleton cleanup and hoping we won't see similar issues on exotic platforms or if it should be abandoned.

PR #969, opened 2017-07-08
fix music in "Draci Historie" (use MIDI channel 16)

kmar:
DH uses MIDI channel 16 for percussion instead of the common 10

### DIFF
--- a/engines/draci/music.cpp
+++ b/engines/draci/music.cpp
@@ -53,7 +53,7 @@ MusicPlayer::MusicPlayer(const char *pathMask) : _pathMask(pathMask), _isGM(fals
 
 void MusicPlayer::sendToChannel(byte channel, uint32 b) {
 	if (!_channelsTable[channel]) {
-		_channelsTable[channel] = (channel == 9) ? _driver->getPercussionChannel() : _driver->allocateChannel();
+		_channelsTable[channel] = (channel == 15) ? _driver->getPercussionChannel() : _driver->allocateChannel();
 		// If a new channel is allocated during the playback, make sure
 		// its volume is correctly initialized.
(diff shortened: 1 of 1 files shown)


kmar:
anybody competent around here?

tobiatesan:
@kmar thanks for contributing. 
If I understand, it is the case that _without_ this patch music is played back incorrectly, right?

kmar:
You're welcome.
Yes that's correct. It can be verified easily by installing Anvil studio, load "hudba2.mid" and change channel 16 to channel 10 (and compare how both sound).
Actually I know this because back in '95 or so I helped the guys with music playback using OPL
and a friend of mine actually wrote the music for the game so I know how it should sound.
I also rebuilt scummvm and verified that the patch indeed works.

rspalek:
hi,

On Mon, Jul 10, 2017 at 12:17 PM, kmar <notifications@github.com> wrote:

> Actually I know this because back in '95 or so I helped the guys with
> music playback using OPL
> and a friend of mine actually wrote the music for the game so I know how
> it should sound.
>
thank you for the patch. nice to hear from a friend of a friend :))  I'm
one of the co-authors of both the original game and its scummvm
reimplementation.

-- 
Robert Špalek | software engineer | rspalek@gmail.com | +1 (408) 359-7434

kmar:
hi, I believe we haven't met (so nice to meet you), but I remeber Lukas and Pavel (and his orange Skoda 120:)

digitall:
Since this is a pretty small and self-contained change and appears to be uncontroversial, can this now be merged?

sev-:
Merged, thanks a lot!

PR #970, opened 2017-07-10
CMD: Fix batch adding of games

tobiatesan:
This changes the behaviour of batch adding as discussed in PR926:
https://github.com/scummvm/scummvm/pull/926#discussion_r126132411

Moreover, it adds a --recursive option that controls whether the search is recursive or not.

New semantics is as follows:
    
```
  [-p <dir>] --add                 adds all games in <dir> or working 
                                   dir

  [-p <dir>] --detect              enumerates dectected games in <dir> 
                                   with their ids

  [-p <dir>] --game <id> --add     adds just game <id> if found in <dir>
                                   and not already added

  [-p <dir>] --recursive --add     adds all games in <dir> and subdirs
                                   if not already added

  [-p <dir>] --recursive --game <id> --add    
                                   adds just game <id> if found in <dir>
                                   or its subdirs and not already added

  [-p <dir>] --recursive --detect  enumerates games in <dir> and subdirs

  [-p <dir>] --auto-detect
                                   launches the first game found in <dir>

  [-p <dir>] --recursive --auto-detect
                                   displays error message

```

1. I would appreciate feedback, especially concerning the very last one.

The reason for the displaying an error message when attempting to do 
autodetection on a whole tree is mainly one of UX, IMO it *might* get
confusing on a sufficiently large/deep tree.

The relevant if() can be removed safely if you guys believe it's not the case.

2. Do you have any comments re: message strings, either?

3. As for the rest, if works well enough for me, but I would love if someone could give it a quick spin, especially on a Mac or some exotic platform.

Thanks

### DIFF
--- a/README
+++ b/README
@@ -1361,10 +1361,19 @@ arguments -- see the next section.
   -z, --list-games         Display list of supported games and exit
   -t, --list-targets       Display list of configured targets and exit
   --list-saves=TARGET      Display a list of saved games for the game (TARGET) specified
-  --auto-detect            Display a list of games from current or specified
-                           directory and start the first one found. Use
-                           --path=PATH before--auto-detect to specify a
-                           directory.
+  -a, --add                Add all games from current or specified directory.
+                           If --game=ID is passed only the game with id ID is
+                           added. See also --detect.
+                           Use --path=PATH before -a, --add to specify a directory.
+  --detect                 Display a list of games with their ID from current or
+                           specified directory without adding it to the config.
+                           Use --path=PATH before --detect to specify a directory.
+  --game=ID                In combination with --add or --detect only adds or attempts to
+                           detect the game with id ID.
+  --auto-detect            Display a list of games from current or specified directory
+                           and start the first one. Use --path=PATH before --auto-detect
+                           to specify a directory.
+  --recursive              In combination with --add recurse down all subdirectories
   --console                Enable the console window (default: enabled) (Windows only)
 
   -c, --config=CONFIG      Use alternate configuration file
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -878,9 +878,9 @@ static Common::String detectGames(Common::String path, Common::String recursiveO
 	GameList candidates = recListGames(dir, recursive);
 
 	if (candidates.empty()) {
-		printf("ScummVM could not find any game in %s\n", dir.getPath().c_str());
+		printf("WARNING: ScummVM could not find any game in %s\n", dir.getPath().c_str());
 		if (!recursive) {
-			printf("Consider using --recursive to search inside subdirectories\n");
+			printf("WARNING: Consider using --recursive *before* --add or --detect to search inside subdirectories\n");
 		}
 		return Common::String();
 	}
@@ -1167,15 +1167,18 @@ bool processSettings(Common::String &command, Common::StringMap &settings, Commo
 		// If we get a non-empty ID, we store it in command so that it gets processed together with the
 		// other command line options below.
 		if (settings["recursive"] == "true") {
-			printf("Autodetection not supported with --recursive; are you sure you didn't want --detect?\n");
+			printf("ERROR: Autodetection not supported with --recursive; are you sure you didn't want --detect?\n");
+			err = Common::kUnknownError;
 			return true;
 			// There is not a particularly good technical reason for this.
 			// From an UX point of view, however, it might get confusing.
 			// Consider removing this if consensus says otherwise.
 		} else {
 			command = detectGames(settings["path"], settings["recursive"]);
-			if (command.empty())
+			if (command.empty()) {
+				err = Common::kNoGameDataFoundError;
 				return true;
+			}
 		}
 	} else if (command == "detect") {
 		detectGames(settings["path"], settings["recursive"]);
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -69,11 +69,11 @@ static const char HELP_STRING[] =
 	"  -z, --list-games         Display list of supported games and exit\n"
 	"  -t, --list-targets       Display list of configured targets and exit\n"
 	"  --list-saves=TARGET      Display a list of saved games for the game (TARGET) specified\n"
-	"  -a, --add                Add a game from current or specified directory\n"
+	"  -a, --add                Add all games from current or specified directory.\n"
+	"                           If --game=ID is passed only the game with id ID is added. See also --detect\n"
 	"                           Use --path=PATH before -a, --add to specify a directory.\n"
-	"  --massadd                Add all games from current or specified directory and all sub directories\n"
-	"                           Use --path=PATH before --massadd to specify a directory.\n"
-	"  --detect                 Display a list of games from current or specified directory\n"
+	"  --game=ID                In combination with --add, only adds the game with id ID. See also --detect\n"
+	"  --detect                 Display a list of games with their ID from current or specified directory\n"
 	"                           without adding it to the config. Use --path=PATH before --detect\n"
 	"                           to specify a directory.\n"
 	"  --auto-detect            Display a list of games from current or specified directory\n"
@@ -267,6 +267,7 @@ void registerDefaults() {
 	ConfMan.registerDefault("gui_saveload_last_pos", "0");
 
 	ConfMan.registerDefault("gui_browser_show_hidden", false);
+	ConfMan.registerDefault("game", "");
 
 #ifdef USE_FLUIDSYNTH
 	// The settings are deliberately stored the same way as in Qsynth. The
@@ -420,9 +421,6 @@ Common::String parseCommandLine(Common::StringMap &settings, int argc, const cha
 			END_COMMAND
 
 			DO_COMMAND('a', "add")
-			END_COMMAND
-
-			DO_LONG_COMMAND("massadd")
 			END_COMMAND
 
 			DO_LONG_COMMAND("detect")
@@ -601,6 +599,9 @@ Common::String parseCommandLine(Common::StringMap &settings, int argc, const cha
 			END_OPTION
 
 			DO_LONG_OPTION("gui-theme")
+			END_OPTION
+
+			DO_LONG_OPTION("game")
 			END_OPTION
 
 			DO_LONG_OPTION("themepath")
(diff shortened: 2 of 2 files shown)


base/commandLine.cpp
```diff
@@ -855,11 +851,39 @@ static bool addGameToConf(const GameDescriptor &gd) {
 	return true;
 }
 
+static GameList recListGames(Common::FSNode dir, bool recursive) {
+	GameList list = getGameList(dir);
+
+	if (recursive) {
+		Common::FSList files;
+		if (dir.getChildren(files, Common::FSNode::kListDirectoriesOnly)) {
```
tobiatesan: Only now I notice this is probably a useless leftover.

tobiatesan:
I probably also need to update README.txt with the new options, if there are complaints.

criezy:
> I probably also need to update README.txt with the new options, if there are no complaints.

@tobiatesan You certainly do :)

At first quick glance, I have no complaint with the suggested change.
But I am wondering what to do with ` [-p <dir>] --game <id> --auto-detect`. Show an error indicating the user probably wants to do `[-p <dir>] <id>`?

Also I have not yet looked at the code, and not tried it, but I am planning to do both this weekend.

tobiatesan:
@criezy thanks. On a further thought, I would *guess* I'd need to also update the `err` parameter with `kNoGameDataFoundError` in case `--auto-detect` is launched and no games are found?

I'm assuming there isn't a profound reason why this isn't the case already.

README
```diff
@@ -1361,10 +1361,19 @@ arguments -- see the next section.
...
+                           Use --path=PATH before -a, --add to specify a directory.
+  --game=ID                In combination with --add, only adds the game with
+                           id ID. See also --detect
+  --detect                 Display a list of games with their ID from current or
+                           specified directory without adding it to the config.
+                           Use --path=PATH before --detect to specify a directory.
+  --auto-detect            Display a list of games from current or specified directory
+                           and start the first one. Use --path=PATH before --auto-detect
+                           to specify a directory.
+  --recursive              In combination with --add recurse down all subdirectories
```
criezy: or with --detect

base/commandLine.cpp
```diff
@@ -69,16 +69,17 @@ static const char HELP_STRING[] =
...
-	"                           Use --path=PATH before --massadd to specify a directory.\n"
-	"  --detect                 Display a list of games from current or specified directory\n"
+	"  --game=ID                In combination with --add, only adds the game with id ID. See also --detect\n"
+	"  --detect                 Display a list of games with their ID from current or specified directory\n"
 	"                           without adding it to the config. Use --path=PATH before --detect\n"
 	"                           to specify a directory.\n"
 	"  --auto-detect            Display a list of games from current or specified directory\n"
 	"                           and start the first one. Use --path=PATH before --auto-detect\n"
 	"                           to specify a directory.\n"
+	"  --recursive              In combination with --add recurse down all subdirectories\n"
```
criezy: or with --detect

base/commandLine.cpp
```diff
@@ -871,97 +894,46 @@ static Common::String detectGames(Common::String path) {
...
-		}
-		--idx;
-		if (idx < 0 || idx >= (int)candidates.size()) {
-			printf("Invalid index. No game added.\n");
-			return false;
+static int recAddGames(Common::FSNode dir, Common::String game, bool recursive) {
+	int count = 0;
+	GameList list = getGameList(dir);
+	for (GameList::iterator v = list.begin(); v != list.end(); ++v) {
+		if (v->gameid().c_str() != game && game != "") {
```
criezy: Consider using `!game.empty()` instead of `game != ""`

criezy:
Generally it looks good to me.
There are a couple of behavioural changes we might want to make though:
- Handle `--recursive` when we want to start a game without adding it after having run a recursive --detect: `./scummvm --path=<path> --recursive <id>`. Currently we have to pass the full path and cannot use `--recursive`, but sadly `--recursive --detect` doesn't provide that path.
 - We might want to print the path in addition to game ID and description when using --detect with --recursive.

I am also wondering if for consistency we might not want to change the way we can start a game from `./scummvm --path=<path> <id>` to `./scummvm --path=<path> --game=<id> --start` (optionally with the use of `--recursive` as mentioned above).

README
tobiatesan: Whops, thanks

tobiatesan:
@criezy thanks for taking the time to try it out.

> Handle --recursive when we want to start a game without adding it after having run a recursive --detect: ./scummvm --path=<path> --recursive <id>. Currently we have to pass the full path and cannot use --recursive, but sadly --recursive --detect doesn't provide that path.

If I am understanding this correctly, with `./scummvm --path=<path> --recursive <id>`you essentially want an `--auto-detect` that can recurse *and* will run *one* specific game (`<id>`)?

I suppose it's doable in two or three lines, but something about it feels a *little bit* too magic-y to me.
I'll throw it in in case I'm the only one.

> We might want to print the path in addition to game ID and description when using --detect with --recursive.

Makes sense, you've got it.

RobLoach:
Did some testing...

1. Compiled `tobiatesan/scummvm#fix_massadd` without any problems
2. Ran `./scummvm --path=~/Games/ScummVM/monkey2/MONKEY2.000 --auto-detect`
    - Found that the`[GAME]` option needs to be a directory, not a file.
3. Ran `./scummvm --path=~/Games/ScummVM/monkey2/ --auto-detect`
    - Played Monkey Island without any issue.
4. Ran `./scummvm --path ~/Games/ScummVM --recursive --add`
    - It successfully added 2 games
5. Ran the same command
    - It successfully doesn't add the same game twice:
        - "Found monkey2, but has already been added, skipping"

Overall, this is looking really nice! Will make managing the game library much easier. Thank you for getting this up and running!

Need any help with anything?

criezy:
@tobiatesan If `./scummvm --recursive --detect` provides the full path to each detected game that's good enough for me as this would allow to run any one of the detected game afterward with `./scummvm --path=<fullpath> <id>` (and thus without the need to make that last one recursive).

Once you have done that and updated the README and help I think it can be merged.

RobLoach:
Would something like the below be possible?

```
./scummvm --path=~/Games/toonstruck/MISC/MAIN.EXE --auto-detect
```

Note that the executable is in a MISC subfolder of `toonstruck`.

tobiatesan:
@RobLoach thanks for trying it out.

@criezy as soon as I have a free minute I'll add the full path and fix the README; sorry for taking so long

@RobLoach:

> > Would something like the below be possible?
>
> `./scummvm --path=~/Games/toonstruck/MISC/MAIN.EXE --auto-detect`

Essentially traversing the dir tree upward instead of downward if `--recurse` is not passed?
I haven't thought much about it, but I don't see any obvious obstacle.
If you mean recognizing _the executable_, might get painful.

Does anyone else like or dislike the idea?

tobiatesan:
@RobLoach may I ask what's your use case for this? Is it by chance some form of "drag'n'drop"?

i30817:
Ah, another thing comes to mind. Does scummvm already have a 'list all game variants' cmd line switch? It would be valuable in *building* automatically updated databases that can map the --detect, as more engines get added.

Well if the output can be json with another switch or something.

tobiatesan:
@i30817 thank you a whole lot for your interest.
I believe your ideas are both very reasonable, thank you for those.
I suppose #2 is doable (probably it's faster do do it by hand than to find and add as a dependency a portable C++ json lib); so is #1, limited to games that have detection entries.

However, I feel they are a bit outside the scope of this PR; we can tackle this at a later date (not _so_ far away in the future).

If you have a second could you open a feature request (well, two) on `http://bugs.scummvm.org` to facilitate discussion with the rest of the team? Thanks.

RobLoach:
> may I ask what's your use case for this? Is it by chance some form of "drag'n'drop"?

Correct, different launching platforms, etc. Two examples of where launching from a data file/sub-folder may be helpful:

1. Windows' "Open With" functionality. Right click on a data file, Open With ScummVM
2. [RetroArch](http://retroarch.com) will be able to pick up the data files and launch from a data file. There are things we can do in the ScummVM core to make it execute on the folder, just need a way to discover which folder to use.

Again, these are not absolute requirements, I'm just curious of a way we could make the whole process easier for discovering/launching games. You've made it so much more seamless! :+1:

i30817:
what about just   ` [-p <dir>] --game <id> `. For launching individual games that have multiple versions in the same dir?

criezy:
@i30817: you can already do that without the `--game=` (and it has been available for a long time). Just use `[-p <dir>] <id>`

I can see an issue with supporting`--path=<file> --auto-detect`: our detector works on directories and not on files. We could strip the file name and detect games in the directory, but then there is a possibility that it could detect several games and not starting the one we expect (i.e. we would expect to start the one that uses the file we pass). We could modify the detection code to only return games that use the file when a file is passed instead of a directory, but that would still not behaves as expected since we only use a subset of the files to detect games and have no way to know which other files are part of the game (and in particular we almost never use the original executable since often ScummVM doesn't need it).

RobLoach:
> our detector works on directories and not on files

Thanks, I'd opt to ignore the `-p=<file>` feature for now, we can work on that that post-merge, or figure out alternatives.

i30817:
> @i30817: you can already do that without the --game= (and it has been available for a long time). Just use [-p <dir>] <id>

The 'id' is not the variant right? If the dir hypothetically contains both lsl6 and lsl6highres for example, and we just use scummvm -p <dir> lsl6 it will 'chose the first', which is semi random. I was told that when RetroArch tried to use the variant name (the ones displayed with `scummvm -t`, they wouldn't start (probably because the scummvm.ini file doesn't exist). The only <id> that worked was the bare game, which means to me that at least variants aren't well supported from the cmd line, without adding them first.

criezy:
@i30817 Indeed, the 'id' is not the variant, but it can be any target already added to ScummVM. This means however that you cannot currently start the variant without adding it first.

Using `scummvm -t` will list all the games added to scummvm. It will not list variants that you have not yet added. Thus you can use targets listed by `scummvm -t` on the command line to start the corresponding game, and it will work.

Also `scummvm --detect` and `scummvm --game=<id> --add` have the same problem in that they list or add ids and not variant names. That means you can use the output of `scummvm --detect` for either `scummvm --game=<id> --add` or `scummvm <id>`, but that will not let you select a specific target. For example the adds will just add all the target that match the id, while running the game will use the first one. The current code (without this PR) behaves slightly differently in that respect as when adding games, if there is more than one variant it asks you which variant you want to add. But the idea of this pull request was in part to remove the interactive aspect of the current ScummVM behaviour.

i30817:
Well, Retroarch will continue to have its problem of not running variants correctly then. I know this is not your problem but i'm pretty sure they don't add games and don't keep a scummvm.ini anywhere. 

Granted this is because they currently hack around the need for it with the id.scummvm files that contain the game id that the users are required to place on the game dir, but even then, scummvm not allowing boot of variants means these ids are just the main game, so the variant that gets launched is just the 'first'.

It's pretty much a mess.

tobiatesan:
@criezy 
> @tobiatesan If ./scummvm --recursive --detect provides the full path to each detected game that's good enough for me as this would allow to run any one of the detected game afterward with ./scummvm --path=<fullpath> <id> (and thus without the need to make that last one recursive).

Added in `f8be31aa`. However, it is very hard to make everything fit in 80 cols, so users of small terminals might be penalized.

> Once you have done that and updated the README and help I think it can be merged.

README also up to date.
No changes in behaviour were introduced and the retroarch folks generously volunteering as guinea pigs for this PR makes me feel reasonably safe; if it doesn't make your computer explode see if you want to merge it.

@RobLoach @i30817 I really really care about doing whatever it takes to make you guys' lives easier, but I'd say we're straying a bit from the original purpose of this PR, which, as @criezy mentioned, was mainly removing the interactive console.
If you could please file a separate bug or come discuss requirements in our channel, I would be delighted.

criezy:
This is now merged.

I amended a couple of commits to fix the command line help and README (to indicate that `--recursive` works with `--detect` as well as with `--add`). I also left in those commits the indication that `--game=<ID>` works with `--detect`, despite the fact that this was not implemented, and made an additional commit to implement it.

PR #971, opened 2017-07-10
VIDEO: Allow setting the mixer sound type used to play audio tracks

bgK:
While trying to fix bug Trac#9938, I noticed it was not possible to set the mixer sound type for the audio tracks handled by VideoDecoder. For Riven, it makes sense to set the sound type to SFX since the audio tracks of videos are mostly used for effects. I decided to fix that in VideoDecoder.

Then I saw some subclasses of VideoDecoder already allow to set the sound type. I went on and changed them to get the sound type from VideoDecoder. After doing that, I unexpectedly had no compilation failures. No engine actually sets the mixer sound type for videos. Am I missing something?

### DIFF
--- a/engines/access/video/movie_decoder.cpp
+++ b/engines/access/video/movie_decoder.cpp
@@ -152,7 +152,7 @@ bool AccessVIDMovieDecoder::loadStream(Common::SeekableReadStream *stream) {
 				soundblasterRate = _stream->readByte();
 				audioSampleRate = 1000000 / (256 - soundblasterRate);
 
-				_audioTrack = new StreamAudioTrack(audioSampleRate);
+				_audioTrack = new StreamAudioTrack(audioSampleRate, getSoundType());
 				addTrack(_audioTrack);
 
 				_stream->seek(chunkStartOffset); // seek back
@@ -194,7 +194,7 @@ bool AccessVIDMovieDecoder::loadStream(Common::SeekableReadStream *stream) {
 
 	// If sample rate was found, create an audio track
 	if (audioSampleRate) {
-		_audioTrack = new StreamAudioTrack(audioSampleRate);
+		_audioTrack = new StreamAudioTrack(audioSampleRate, getSoundType());
 		addTrack(_audioTrack);
 	}
 
@@ -641,7 +641,8 @@ bool AccessVIDMovieDecoder::StreamVideoTrack::hasDirtyPalette() const {
 	return _dirtyPalette;
 }
 
-AccessVIDMovieDecoder::StreamAudioTrack::StreamAudioTrack(uint32 sampleRate) {
+AccessVIDMovieDecoder::StreamAudioTrack::StreamAudioTrack(uint32 sampleRate, Audio::Mixer::SoundType soundType) :
+		AudioTrack(soundType) {
 	_totalAudioQueued = 0; // currently 0 milliseconds queued
 
 	_sampleRate  = sampleRate;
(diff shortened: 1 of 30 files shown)


csnover:
Here are some engines that set the sound type for videos:

```
./gob/videoplayer.cpp:		video = new ::Video::IMDDecoder(_vm->_mixer, Audio::Mixer::kSFXSoundType);
./gob/videoplayer.cpp:		video = new ::Video::PreIMDDecoder(properties.width, properties.height, _vm->_mixer, Audio::Mixer::kSFXSoundType);
./gob/videoplayer.cpp:		video = new ::Video::VMDDecoder(_vm->_mixer, Audio::Mixer::kSFXSoundType);
./gob/videoplayer.cpp:		video = new ::Video::VMDDecoder(_vm->_mixer, Audio::Mixer::kSFXSoundType);
./sci/graphics/video32.cpp:	VideoPlayer(eventMan, new Video::AVIDecoder(Audio::Mixer::kSFXSoundType)),
./sci/graphics/video32.cpp:	VideoPlayer(eventMan, new Video::AdvancedVMDDecoder(Audio::Mixer::kSFXSoundType)),
./sci/graphics/video32.cpp:	VideoPlayer(eventMan, new Video::AVIDecoder(Audio::Mixer::kSFXSoundType)),
./titanic/support/avi_surface.cpp:	_decoder = new AVIDecoder(Audio::Mixer::kPlainSoundType);
./titanic/support/avi_surface.h:	AVIDecoder(Audio::Mixer::SoundType soundType = Audio::Mixer::kPlainSoundType) :
./titanic/support/avi_surface.h:	AVIDecoder(const Common::Rational &frameRateOverride, Audio::Mixer::SoundType soundType = Audio::Mixer::kPlainSoundType) :
./voyeur/animation.h:	RL2Decoder(Audio::Mixer::SoundType soundType = Audio::Mixer::kPlainSoundType);
./zvision/video/zork_avi_decoder.h:	ZorkAVIDecoder(Audio::Mixer::SoundType soundType = Audio::Mixer::kPlainSoundType) :

```

bgK:
Thanks, I indeed missed SCI32. Audio::Mixer::SoundType was being accepted as a Common::Rational when calling AVIDecoder's constructor.
The rest is fine. Gob is not using VideoDecoder. The other engines are using the default value. My own greps did not find any other cases either.

csnover:
Just curious, is there a particular reason why you want to remove the constructor signature that accepts the sound type?

With regards to the implicit conversion, have you tried doing something like temporarily replacing `Audio::Mixer::SoundType` with something like this, to prevent a successful conversion?:

```
class SoundType {};
const SoundType kPlainSoundType = {};
const SoundType kMusicSoundType = {};
const SoundType kSFXSoundType = {};
const SoundType kSpeechSoundType = {};
```

Maybe someone knows a similar but better technique that they could share with me. (Making the `Rational(int)` constructor temporarily explicit would work too but also cause a lot of false positives.)

csnover:
(I suppose `SoundType` could also be changed permanently to use the “type safe enum” idiom since I don’t think there is any valid way this enum should be converted to int.)

bgK:
I just tried changing the definition of SoundType to something strongly typed and caught an issue in some unrelated code in Starship Titanic. Thanks for the suggestion.
Too bad we can't use C++11's strongly typed enums.

I did not add the SoundType as a constructor argument to the VideoDecoder base class because some decoders don't support audio tracks. I didn't feel the sound type was more important than any of the other settings and deserved a special place in the constructor. I removed the sound type from the constructor arguments of the other video decoders for consistency.

bgK:
I've pushed the unrelated Titanic commit to master, and squashed together the VideoDecoder commits.

bgK:
I'll merge this Wednesday unless somebody wants to further review/discuss it.

bgK:
This was merged as ec49730

PR #972, opened 2017-07-14
Discworld 2: Bug in the showing of "string sets"

PetrCejpek:
Greetings,
 I am working on the Czech translation for Discworld 2: Missing presumed. I came across to following bug.

Every text quote in the ENGLISH1.txt and ENGLISH2.txt files is introduced by character which ascii number corresponds to the length of the following string. Some sets of strings are introduced by two characters where the value of the first one is 8x (hexadecimal) and x corresponds to the number of strings in the set. I am guessing it wants to tell that the whole string set has only one sound quote or something like that.

Example:
**‚X**I mean - it seemed we might have
a few mutual talking points -
information to exchange -
**8**pleasantries -
something in common -
that sort of thing.

Here, **‚X** what is 82 58 in hexadecimal means, that the set has two strings and the first one has length of 88 (58 hexadecimaly).

Unfortunately, I found out that ingame this works well only within interviews with other NPCs. When I try to talk about items in the inventory, it shows only the first string.

Example:
**‚l**I'd taste-test this stuff, but the
animation budget won't stretch to a
proper demonstration of the results -
**B**so you're just going to have to
take it on faith for this one, OK?

Again, it begins with **,** (82), therefore it should show two-string quote. But as the result the game shows:

> I'd taste-test this stuff, but the
> animation budget won't stretch to a
> proper demonstration of the results -

and then Rincewind suddenly stops talking.

**Details of the configuration:**
Scummvm 1.9.0.2
Discworld 2: Missing Presumed (English)
Windows 7 or 10


wjp:
Thank you for the report.

For bug reports, please use https://bugs.scummvm.org/ .

If this was meant to include a fix, you may have selected the wrong branch when creating this Pull Request.

PR #974, opened 2017-07-23
GUI: Stop processing input events from macOS file browser dialog

tsoliman:
This fixes a bug I observed that is related to the macOS file browser.

After the macOS browser is closed, all events that happened while the macOS browser was open were being sent to ScummVM.

Here's the basic repeatable test case:
- From the launcher, Add game, the macOS-specific browser dialog opens
- press down arrow 4 times
- press cancel (and maybe move the mouse over the launcher to force the events to be processed)
- the down arrow events are passed to the launcher and the game screen (bad)

A variant of this is if you press some keys and then press Esc to cancel out of the launcher. This will cause the entire launcher to quit when it receives the Esc key.

A more devious variant of this test case: When pressing Up/Down and then Enter/Return to select a game you want to add and that game happens to have variants (e.g. PQ2). That variant picking dialog will be rapidly confirmed as it gets the arrow keys and enter key. The user may not even see it appear and disappear.

### DIFF
--- a/gui/browser_osx.mm
+++ b/gui/browser_osx.mm
@@ -29,6 +29,7 @@
 #include "common/system.h"
 #include "common/algorithm.h"
 #include "common/translation.h"
+#include "common/events.h"
 
 #include <AppKit/NSNibDeclarations.h>
 #include <AppKit/NSOpenPanel.h>
@@ -176,6 +177,15 @@ int BrowserDialog::runModal() {
 		g_system->setFeatureState(OSystem::kFeatureFullscreenMode, true);
 		g_system->endGFXTransaction();
 	}
+
+	// While the native macOS file browser is open, any input events (e.g. keypresses) are
+	// still received by the NSApplication. With SDL backend for example this results in the
+	// events being queued and processed after we return, thus dispatching events that were
+	// intended for the native file browser. For example: pressing Esc to cancel the native
+	// macOS file browser would cause the application to quit in addition to closing the
+	// file browser.
+	// Clear the event sources' event queue.
+	g_system->getEventManager()->getEventDispatcher()->clearEvents();
 
 	return choiceMade;
 }
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -992,6 +992,10 @@ void SdlEventSource::resetKeyboardEmulation(int16 x_max, int16 y_max) {
 	_km.joy_y = 0;
 }
 
+void SdlEventSource::clearEvents() {
+	SDL_FlushEvents(SDL_FIRSTEVENT, SDL_LASTEVENT);
+}
+
 bool SdlEventSource::handleResizeEvent(Common::Event &event, int w, int h) {
 	if (_graphicsManager) {
 		_graphicsManager->notifyResize(w, h);
--- a/backends/events/sdl/sdl-events.h
+++ b/backends/events/sdl/sdl-events.h
@@ -51,6 +51,11 @@ public:
 	 */
 	virtual void resetKeyboardEmulation(int16 x_max, int16 y_max);
 
+	/**
+	 * Flush the SDL event queue
+	 */
+	virtual void clearEvents();
+
 protected:
 	/** @name Keyboard mouse emulation
 	 * Disabled by fingolfin 2004-12-18.
--- a/common/EventDispatcher.cpp
+++ b/common/EventDispatcher.cpp
@@ -70,6 +70,12 @@ void EventDispatcher::dispatch() {
 	}
 }
 
+void EventDispatcher::clearEvents() {
+	for (List<SourceEntry>::iterator i = _sources.begin(); i != _sources.end(); ++i) {
+		i->source->clearEvents();
+	}
+}
+
 void EventDispatcher::registerMapper(EventMapper *mapper, bool autoFree) {
 	if (_autoFreeMapper) {
 		delete _mapper;
--- a/common/events.h
+++ b/common/events.h
@@ -154,6 +154,11 @@ public:
 	 * By default we allow mapping for every event source.
 	 */
 	virtual bool allowMapping() const { return true; }
+
+	/**
+	 * Clear any pending events that have been queued up in the event source backend
+	 */
+	virtual void clearEvents() { }
 };
 
 /**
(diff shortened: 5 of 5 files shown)


gui/browser_osx.mm
```diff
@@ -177,6 +178,9 @@ - (IBAction) showHiddenFiles : (id) sender {
 		g_system->endGFXTransaction();
 	}
 
+	// Clear the event queue so that keyboard input (for example) isn't sent to ScummVM right after the browser closes
```
wjp: This needs a much more extensive comment explaining this behaviour, since it isn't something expected.

criezy:
Wouldn't using the event manager be cleaner, to avoid exposing SDL to the gui code?
The code below seems to work for example:
```
// While the native macOS file browser is open, any input events (e.g. keypresses) are
// still received by the NSApplication. With SDL backend for example this results in the
// events beeing queued and processed after we return, thus dispatching events that were
// intended for the native file browser. for example pressing Esc to cancel the native
// macOS file browser would cause the application to quit in addition to closing the
// file browser.
// Poll all available event snow to clear the event queue.
Common::Event e;
while(g_system->getEventManager()->pollEvent(e)) {}
```

As it is, your change is going to break compilation for any backend not using SDL on macOS (for example my WIP Qt backend).

tsoliman:
That was my initial approach too but I couldn't get that code to compile in Objective C++, hence why the initial hack (https://github.com/tsoliman/scummvm/commit/c781658ba4554538ecd112a2a883720ee3bac09f) is in the calling C++ code (in `launcher.cpp` rather in `browser_osx.mm`).

I think a hack that exists inside browser_osx.mm is cleaner since the whole native macOS browser is the source of the problem.

I don't feel too strongly about either way.

The 3rd alternative is completely removing the special handling for macOS and have it use the regular file browser like most other ports do. This removes the problem and the need for a hack. It also makes ScummVM behave the same way on all platforms, but is slightly less user-friendly perhaps? Again I don't think it is too unfriendly.

tsoliman:
After talking to @criezy and @csnover on IRC, I have updated the PR to what @criezy suggested above.

tsoliman:
The `g_system->getEventManager()->pollEvent(e)` approach seems to be actually notifying all observers about the event. (try putting some debugs in `SurfaceSdlGraphicsManager::notifyEvent`)

criezy:
@tsoliman: right, but I expect this is also how the non-native file browser on other platforms behaves when it poll events. Although we the native file browser we get all those notifications just after closing it and not while it is running, and it might seem strange to have those delayed actions (e.g. taking snapshot, switching between windowed and fullscreen).

As snover suggested, an alternative is to expose a flushEvents() API in the events manager.

tsoliman:
> right, but I expect this is also how the non-native file browser on other platforms behaves when it poll events. Although we the native file browser we get all those notifications just after closing it and not while it is running, and it might seem strange to have those delayed actions (e.g. taking snapshot, switching between windowed and fullscreen).

That's not really a fair comparison IMHO. The non native browser is regular ScummVM GUI code, it doesn't have its own set of inputs bound for other purposes.

I think the goal of this PR is remove the insanity related to the macOS browser. Having macOS key presses and other input shortcuts propagate to current (or future) observers after also doing their initial purpose is part of that insanity. The fact that it is delayed and batched contributes to the insanity.

> As snover suggested, an alternative is to expose a flushEvents() API in the events manager.

I'll update the PR with that approach.

Once again I appreciate your guidance on this and I hope I didn't come across as argumentative or combative. 😄

tsoliman:
I've updated the PR with the EventManager API approach. The way EventManager reaches the input backends is: EventManager->EventDispatcher->EventSource(s).

@criezy This approach should work for both backends (assuming the problem happens in Qt as well).

Feedback welcome (doubly so with regards to naming)

criezy:
I have not tried this yet, but I suspect that flushing events will bring its own range of issues. This could cause observers to get unbalanced events (for example a key pressed without a key release). One such use case that comes to mind is the Mass Add:
1. Press (and do not release Shift to display the Mass Add button
2. Click on Mass Add
3. Then release the shift key while the file browser is active
4. Close the file browser

With the pollEvent() approach, the Mass Add button should correctly be back to a normal Add. With a flush of events I suspect you will still have a Mass Add button even though shift is no longer pressed.

So both approaches will have their lot of issues. The decision is on which ones are worse.

criezy:
Also I have now tested using Alt-Enter or Alt-S while the file browser is open with the pollEvent() approach, and it actually almost behaves correctly: it does not switch to full screen or take a screenshot when closing the file browser, unless you keep the Alt key pressed until after you close the file browser.

criezy:
To be clear, I don't know which approach is best. I was just pointing out that neither is perfect ;)

In the clearEvents() approach, you could implement `EventDispatcher::clearEvents()` as follow so that you do not have to expose it to EventSource:
```
void EventDispatcher::clearEvents() {
	Event event;

	for (List<SourceEntry>::iterator i = _sources.begin(); i != _sources.end(); ++i) {
		while (i->source->pollEvent(event)) {}
	}

	List<Event> delayedEvents = _mapper->getDelayedEvents();
}
```
So basically it does the same as dispatch(), but without actually dispatching the events.

If you prefer to have the EventSource do the work, then you would need to implement it in the other EventSource as well, and not just the SDL one, so that they work as expected.
Also one possible issue with your current code is that clearEvents() does not clear the delayed events in the event mapper (but this is easy to fix if we decide it should be done).

tsoliman:
> Also I have now tested using Alt-Enter or Alt-S while the file browser is open with the pollEvent() approach, and it actually almost behaves correctly ...

IMHO is a sort of 2 wrongs making a right. The EventMapper should be converting alt-s to something like `Common::EVENT_SCREENSHOT` but instead, the GUI code has platform/backend-specific hacks in there. 

At this I have to say that native browser never really worked right. I imagine the flush behavior is closer to what was originally expected, as I was astonished to see that ScummVM is processing events that were not meant for it. While the poll approach kinda-sorta-seems to work, you can say the same about the current master without polling or flushing.

tsoliman:
> To be clear, I don't know which approach is best. I was just pointing out that neither is perfect ;)

Thanks for this clarification. My previous reply was written without that in mind :)

bgK:
Did you check the value of SDL_KeyboardEvent::windowID for the events being received while the native dialog is open? If I understand correctly, it should not be set for events occurring while no SDL window has the focus (https://wiki.libsdl.org/SDL_KeyboardEvent). This is for SDL 2 only.

tsoliman:
@bgK `SDL_KeyboardEvent::windowID` is 1 normally and 0 for the events coming from the native browser. 👍

tsoliman:
After discussions, this is probably the wrong implementation since it seems to be SDL2-only.
I'll defer to @criezy's as he has pending changes that include a fix for this issue.

PR #975, opened 2017-07-24
Rework and optimize Common::BitStream

wjp:
This optimizes Common::BitStream, removing one layers of virtual function calls for reading bits from a stream, and two layers for reading from a memory buffer.

A benchmark of reading 100 megabytes 5 bits at a time goes from 204ms to 88ms on my machine.

The effect on indeo frame decoding in the titanic engine is a speedup of about 20%. (And significantly more when running valgrind.)


csnover:
I am running through Starship Titanic now with this patch applied. So far so good.

PR #976, opened 2017-07-30
TITANIC: Right Click = Shift + Left Click

dafioram:
This maps the right click in Titanic to be shift + left click. This allows chevrons and fast-transitions to happen just by right clicking. 

The shift happens on right click down and the shift is undone when right click is released. This is done to not conflict with the shift state of the left click.

When the shift key is down the right click still works as shift + left click. The left click still works as it always has.

Some code cleanup might be done since there are now unused variables (e.x., MK_RBUTTON), but I believe this implementation is behaviorally correct.

### DIFF
--- a/engines/titanic/events.cpp
+++ b/engines/titanic/events.cpp
@@ -72,14 +72,16 @@ void Events::pollEvents() {
 			eventTarget()->middleButtonUp(_mousePos);
 			return;
 		case Common::EVENT_RBUTTONDOWN:
-			_specialButtons |= MK_RBUTTON;
+			_specialButtons |= MK_LBUTTON;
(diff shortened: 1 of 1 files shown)


dreammaster:
Thanks for the excellent suggestion. I expanded on your changes to remove all the right mouse button handling methods, so there won't be any confusion in the future why they aren't being called. Also as a little reward of appreciation for your hard work testing.. the initial "Continue" dialog at the start of the game has an Easter Egg.. if you hold down the right mouse button on either of the Doorbot's eyes, it will show you his "Evil Twin". :) I've also updated the code as part of this commit so that it will still work.

PR #977, opened 2017-07-30
COMMON: Migrate Various Endian Functions to Common from Engine Code.

digitall:
These changes are a subset of a commit in the external BOLT engine tree used for Merlin's Apprentice:
https://github.com/beholdnec/scummvm-bolt/commit/def568f05f600c87b10a89ca83d40fe80b09ea57

Since this commit mixes engine changes with making changes to common code and other engines, including removal of a WinCE specific hack in the Queen engine, this should be reviewed and merged prior to that engine's future PR discussions.

This will allow the external engine to be rebased and cleaned up in an easier manner without possible conflicts due to work on those other engines in the master tree in the meantime.

Also, the proposed change here to common/endian.h looks reasonable and will avoid other duplications in future of these kind of READ* and WRITE* functions.

### DIFF
--- a/common/endian.h
+++ b/common/endian.h
@@ -608,4 +608,36 @@ inline uint32 READ_BE_UINT24(const void *ptr) {
 #define READ_UINT24(a) READ_BE_UINT24(a)
 #endif
 
+inline int16 READ_LE_INT16(const void *ptr) {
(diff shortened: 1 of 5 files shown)


sev-:
Looks fine to me. Merging

PR #979, opened 2017-08-03
GRAPHICS: Consider flip mode when handling off-screen clipping

yinsimei:
Hi, I meet a bug concerning transparent surface when importing the sludge engine, which is, when a horizontally flipped sprite leaves the scene from the left side, it begins to disappear from the right side. So I guess maybe we need to take the flip mode in to account when handling off-screen clipping in transparent_surface.cpp and it works for Sludge games.
But I'm not sure if this is the right level to fix the bug as this part of code has existed for a long time without anyone else noticing this need, so I made a PR to discuss the problem.

### DIFF
--- a/graphics/transparent_surface.cpp
+++ b/graphics/transparent_surface.cpp
@@ -400,19 +400,31 @@ Common::Rect TransparentSurface::blit(Graphics::Surface &target, int posX, int p
 	// Handle off-screen clipping
 	if (posY < 0) {
 		img->h = MAX(0, (int)img->h - -posY);
-		img->setPixels((byte *)img->getBasePtr(0, -posY));
+		if (!(flipping & FLIP_V))
+			img->setPixels((byte *)img->getBasePtr(0, -posY));
 		posY = 0;
 	}
 
 	if (posX < 0) {
 		img->w = MAX(0, (int)img->w - -posX);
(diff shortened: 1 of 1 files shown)


tobiatesan:
Thanks for addressing this with a PR.

- I believe, all things considered, this can be the appropriate place to add this behaviour.

- My brain is not the most sound of static analyzers, but at a glance the changes appear reasonable (as in, correct) to me.

- At a glance I don't see an obvious performance cost either. Good.

- `transparent_surface.h` is used in sword25, wintermute, fullpipe, wage and director.
   Wintermute seems unaffected by the change, ie. it doesn't apparently break anything in there. Also good.

- `flipping` seems to be used exclusively in `sword25` and `fullpipe`.
   I don't have a copy to test if it works in there. @sev- appears to be the main author of `fullpipe` and one of the largest contributors to `sword25` along with @bluegr, so I'd ask them if this matches the semantics they have/had in mind just to be sure. Skimming the code in `engines/sword25` and `engines/fullpipe` I don't see any obvious inconsistencies.

- @csnover and the SCI gang might also be interested in these changes.

sev-:
I tested both full pipe and sword25, both seem to not have any introduced glitches with this patch.

So, I am merging it.

PR #981, opened 2017-08-04
VIDEO: Support old-style stereo VMDs

csnover:
I still need to quickly run this through Valgrind to make sure I didn’t screw something up with the buffer ownership, and I didn’t test Urban Runner yet to make sure this changeset doesn’t affect its videos (if someone can do this so I do not have to dig out my copy from offline storage again, that would be nice). Otherwise, cursory testing seems to indicate things are working, so I wanted to get this up right away to give people some time to look at it and send feedback.

So far I ran this updated code against a few VMDs from SQ6 (all the videos accessible from the movies easter egg in the shuttle), Torin (intro), GK2 (chapter 6 intro videos), Phant1 (intro, chapter 1 intro, some of the chase VMDs), Lighthouse, Shivers (intro), and Shivers 2 (Sierra logo), with no noted issues so far.

I made some light modifications to MemoryReadWriteStream to implement SeekableReadStream so it can be passed directly to DPCMStream because I didn’t feel like writing another ring buffer implementation. Happy to accept feedback on possible better alternatives to this.

With my random testing, I did not run across a video with `_soundSlicesCount > 32`, so I am not sure if that will still work OK with the clamping of `extraSize` to zero. If someone does not know offhand of a sample video with this value, I will probably just write a quick script to check all the VMDs I have to see if I can find one.

Thanks,


video/coktel_decoder.cpp
```diff
@@ -1543,6 +1543,93 @@ Graphics::PixelFormat IMDDecoder::getPixelFormat() const {
 	return Graphics::PixelFormat::createFormatCLUT8();
 }
 
+class DPCMStream : public Audio::AudioStream {
+public:
+	DPCMStream(Common::SeekableReadStream *stream, int rate, int channels, bool oldStereo) {
+		_stream = stream;
```
8BitPimp: Small thing, you could perhaps
```
assert(stream);
```
here as the rest of the code expects it to be valid?

8BitPimp: could also:
```
assert(channels <= 2);
```
As _buffer is hard-coded with a length of two and indexed up to the channel count.

csnover: Thanks for your review!

`stream` could be asserted, though if it’s null here you’re probably having a bad day already since it means that a critical memory allocation failed. `channels` is never going to be any value other than 1 or 2 since that argument is always hard-coded and the original data format doesn’t support more than 2 channels, so I don’t think there is any practical reason to check that input.

FWIW the only thing that actually changed here was the addition of the `_oldStereo` flag and an extra sanity check on the audio buffer when reading samples since the buffer can be filled with an odd number of samples. All this code just had to be moved up a bit so the class could be used by functions defined earlier in the source.

csnover:
I’d like to land this by Sunday, but if anyone wants more time than that to look at this, just let me know.

PR #985, opened 2017-08-11
Graphics:: Add multiply blend mode

yinsimei:
Add multiply blend mode to blitting of transparent surface.

### DIFF
--- a/graphics/transparent_surface.cpp
+++ b/graphics/transparent_surface.cpp
@@ -62,6 +62,7 @@ void doBlitBinaryFast(byte *ino, byte *outo, uint32 width, uint32 height, uint32
 void doBlitAlphaBlend(byte *ino, byte *outo, uint32 width, uint32 height, uint32 pitch, int32 inStep, int32 inoStep, uint32 color);
 void doBlitAdditiveBlend(byte *ino, byte *outo, uint32 width, uint32 height, uint32 pitch, int32 inStep, int32 inoStep, uint32 color);
 void doBlitSubtractiveBlend(byte *ino, byte *outo, uint32 width, uint32 height, uint32 pitch, int32 inStep, int32 inoStep, uint32 color);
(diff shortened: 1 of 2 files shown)


graphics/transparent_surface.cpp
```diff
@@ -326,6 +327,72 @@ void doBlitSubtractiveBlend(byte *ino, byte *outo, uint32 width, uint32 height,
...
+	byte *out;
+
+	if (color == 0xffffffff) {
+		for (uint32 i = 0; i < height; i++) {
+			out = outo;
+			in = ino;
+			for (uint32 j = 0; j < width; j++) {
+
+				if (in[kAIndex] != 0) {
+					out[kRIndex] = MIN((in[kRIndex] * in[kAIndex] >> 8) * out[kRIndex] / 225, 255);
```
8BitPimp: Should these really be `/ 225` or rather `/ 255`?

sev-: If you want to ' / 255' (not 225 as marked here, apparently a typo), it is better to use ' >> 8'

yinsimei: True, I'll correct it

csnover: `>> 8` is equivalent of `/ 256`, not `/ 255`.

csnover: Also, I would challenge the statement that using a bit shift is better. I may be incorrect, but my understanding is that all compilers will perform peephole optimisation of unsigned division with a 2^n denominator to turn it into a bit shift (or whatever instruction is optimal on a given platform) automatically. However, whereas arithmetic division of negative numerators is well-defined, bit shifting negative integers is either UB (left shift) or implementation-defined (right shift). So you can get into UB territory more easily when bit shifting manually, without actually having any performance improvement since the compiler will optimise it anyway if it is possible.

sev-: Actually, you're right about 256, that should be the actual value. As of the bitshift, we're dealing here is 'byte' types which are unsigned.

sev-:
Thanks for addressing the notes, merging.

PR #988, opened 2017-08-12
MOHAWK: Myst and Riven saves and controls readme additions

dafioram:
Add basic controls for riven and myst.

List that riven and myst can have saves from original game and how to rename the saves.

The controls are pretty basic for myst/riven, but I was already adding details for the saved games so I thought I would add that also. Skipping the cutscenes (with the escape key) will be really useful to someone if they don't know that is an option.


bgK:
Thanks

PR #993, opened 2017-08-17
SCUMM: Prevent darkenPallete() in Room 0 in FT, fixes bug #9871

BenCastricum:
Room 0 is a blank screen, fading out has no effect. So simply skipping the call is a reasonable way to prevent the crash mentioned in the bug report.


sev-:
Nice, thank you!

PR #995, opened 2017-08-19
TITANIC: CStarCamera Refactoring and CStarCrosshairs logic fix

dafioram:
1. Added an enum for star_camera that is more informative about number of stars locked on. Many other classes still use it as an int so this could be expanded to those, later.
2. Reduce some file dependencies of files related to the star_camera class.
3. Some function renames and more comments
4. Fix for #10126. There were a lot of ways it crashed then mentioned in the ticket so I fixed all the different ways and tested the puzzle a lot. That section of code may benefit from case statements or more functions.


dafioram:
Hold on to merging this. I found another bug that I want to fix first and then I will do a --force (update) this PR.

dafioram:
Okay got my changes in.
5. Allow star_camera functions and the like to do something else if the normalization is unsuccessful.
6. Fix #10147. A new bug that caused the star puzzle to crash.

PR #997, opened 2017-08-20
TITANIC: Change ship view and position even if not moved

dafioram:
The code was preventing the position and view from changing
when the distance between the current and new position for
a marked auto mover was zero. This happens if you lock the
2nd or 3rd star and then unlock and relock again.

It was prevented this with asserts and if statement checks
and I removed them all.

This section of code isn't doing any inverses based on the
reciprocal of the distance so there is no issue with allowing
transition speeds/distances of zero.

Fixes #10148.

### DIFF
--- a/engines/titanic/star_control/camera_auto_mover.cpp
+++ b/engines/titanic/star_control/camera_auto_mover.cpp
@@ -50,10 +50,8 @@ void CCameraAutoMover::proc2(const FVector &oldPos, const FVector &newPos,
 	_destPos = newPos;
 	_posDelta = _destPos - _srcPos;
        float temp = 0.0;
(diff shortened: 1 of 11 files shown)


dafioram:
I added a 2nd commit that prevents the awkward transition from happening for #9961. It was happening for large distances as now I just don't allow the lock to be attempted when it will happen.

For now it can be added as a todo to get the transitions to lock correctly for two stars when the distances are >1e8.

I will continue to work on the code, but that bug is now no longer visible, so it is minor now.

dafioram:
This is only a partial fix for #9961 since the not locking issue can happen for 3 stars, but it doesn't look as bad then.

dafioram:
Added some minor additions.

PR #999, opened 2017-08-24
TITANIC: Reduce header includes

dafioram:
1. Reduced header includes in titanic.h a lot and in closely related files.  Lots of files where including titanic.h just to get its includes and not its API.
2. Move debug enums and the like in titanic.h to the debugger class. Several of the star control files where including titanic.h just so they could use the debugger.
3. The random float function in titanic.h can give 0 to 35537 since the documentation says it only gives to 35535 I changed the numbers so it does that.
4. I especially reduced the header includes in the star control files (because I am more familiar with those).

This overall change should make it easier to identifying why a file is included in that particular file and if one greps for an included header they are more likely to find it in the class that uses it. This also reduces the binary by about 200 kB. The build time is about the same.


csnover:
Now that the engine is out of the unstable development phase, this PR makes it sound like it would be a good idea to run [IWYU](https://include-what-you-use.org/) against the whole engine.

It would be nice too if the behaviour change of `getRandomFloat` could be separated in the commit history since it is usually good to keep such intentional changes away from housekeeping stuff when possible.

dafioram:
The getRandomFloat change is now commit 2.

How does titanic get added to [doxygen](http://doxygen.scummvm.org/)?

dafioram:
I can expand this PR further, but I think it is a higher priority for me to spend time on refactoring the start_control/ section of the code (see if I can figure out #9961).

PR #1000, opened 2017-08-24
IMAGE: Implement Indeo4 transparency

csnover:
This fixes the various SGT stateroom furniture and Titania parts (and maybe a couple other miscellaneous places using Indeo4 transparency) in TITANIC engine.

There are still a few visual artefacts remaining for some transparent videos:

* The Major Horizontally Mobile Storage Compartment (b14.avi) renders a block of half-shelf, half-black to the right of the top shelf whenever it is opened (or closed when you leave and re-enter the stateroom) which gets stuck on the screen.
* The Personal Maintenance Hub (b7.avi) seems to have its top-left edges cut off on some frames, and on other frames the lower-right draws pixels that should be transparent.

I don’t *think* these are Indeo4 transparency problems any more since the validation checks pass (the code reads exactly the right number of pixels and the right number of bytes), and the RLE value output from the reference decoder now appears to match the reference decoder. However, I suppose it is possible that the current implementation may be doing something bad that is causing trouble with B/P frames, or interacting with other Indeo code in a way that leads to these glitches.

Relatedly, there are some problems with fringe artefacts on some other Indeo4 videos (like the Major Semi-Recumbent Relaxation Device, b18.avi) but these videos do not use Indeo4 transparency, so there is at least one other problem remaining with video transparency in TITANIC which is not related to this Indeo4 feature.

### DIFF
--- a/image/codecs/indeo/indeo.cpp
+++ b/image/codecs/indeo/indeo.cpp
@@ -444,7 +444,8 @@ IVI45DecContext::IVI45DecContext() : _gb(nullptr), _frameNum(0), _frameType(0),
 		_bRefBuf(0), _rvmapSel(0), _inImf(false), _inQ(false), _picGlobQuant(0),
 		_unknown1(0), _gopHdrSize(0), _gopFlags(0), _lockWord(0), _hasBFrames(false),
 		_hasTransp(false), _usesTiling(false), _usesHaar(false), _usesFullpel(false),
-		_gopInvalid(false), _isIndeo4(false), _pFrame(nullptr), _gotPFrame(false) {
+		_gopInvalid(false), _isIndeo4(false), _transKeyColor(0), _pFrame(nullptr),
+		_gotPFrame(false) {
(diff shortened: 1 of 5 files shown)


csnover:
I developed this with PR#975 applied to my working branch, so the getVLC2 call is causing compilation failure on Travis. (Land PR#975 first. :))

csnover:
I spent a little more time on this today and continue to see transparency artefacts during playback outside of TITANIC engine, so unfortunately for me there is still a problem in the decoder. z132.avi is another one that is doing some weird stuff. In either case, I guess I am still working on this. Updates forthcoming…

csnover:
I’ve fixed the remaining transparency problems which were a simple error in handling the case where RLE runs ended in the padding area of a line. All the Indeo4 transparency rendering seems to be working correctly now. The stuck block of pixels of the Major Horizontally Mobile Storage Compartment is actually the Minor Horizontally Mobile Storage Compartment being stuck, which is an unrelated bug. I will land this PR shortly, once I have done a bit more testing to verify no other weird stuff is happening.

PR #1002, opened 2017-08-30
AUDIO: Import micromod code, xm/s3m/mod decoder

yinsimei:
This commit is based on https://github.com/martincameron/micromod
The principal is to convert these tracker sound formats into wav and feed it to wav decoder to make an audio stream.

### DIFF
--- a/audio/mods/mod_xm_s3m.cpp
+++ b/audio/mods/mod_xm_s3m.cpp
@@ -1,0 +1,1350 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+/*
+ * This code is based on IBXM mod player
+ *
+ * Copyright (c) 2015, Martin Cameron
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or
+ * without modification, are permitted provided that the
+ * following conditions are met:
+ *
+ * * Redistributions of source code must retain the above
+ * copyright notice, this list of conditions and the
+ * following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the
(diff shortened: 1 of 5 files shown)


audio/mods/mod_xm_s3m.cpp
```diff
@@ -0,0 +1,1353 @@
...
+	memset(_rampBuf, 0, 128 * sizeof(int));
+	tick();
+}
+
+} // End of namespace Modules
+
+namespace Audio {
+
+AudioStream *makeModXmS3mStream(Common::SeekableReadStream *stream, int rate, int interpolation) {
+	Modules::ModXmS3mConvertor *modXmS3mConvertor = new Modules::ModXmS3mConvertor(stream, rate, interpolation);
```
bgK: `modXmS3mConvertor` never freed, resulting in a memory leak.

yinsimei: True, I'll correct it.

yinsimei:
@bgK Yes, you are right, this code can be very slow for some devices and needs optimisation indeed. I'm not very familiar with how AudioStream inheritance works, but I'll have a look at it. I think it's the "xmToWav" part in the converter that I need to split and put into that child AudioStream class.

I guess I'll make something like "Paula" but simpler and with more channels? It's been set to 4 in paula but I may need more of them for .s3m and .xm: https://github.com/scummvm/scummvm/blob/master/audio/mods/paula.h#L39

Thanks for your note!

bgK:
I'd say your new AudioStream subclass can have a ModXmS3mConvertor member, which is then used in `AudioStream::readBuffer` to decode ticks as required with `ModXmS3mConvertor::getAudio`. Your AudioStream class probably also needs a mix buffer to keep a partially consumed decoded tick.

yinsimei:
Hi, what's your opinion about the recent commit that changes the converter to an audio stream?
Find sth else to be improved?

audio/mods/mod_xm_s3m.cpp
```diff
@@ -63,20 +63,18 @@
...
 #include "audio/mods/module2wav.h"
 
-#define DEBUG_DUMP_FILE 0
-
 namespace Modules {
 
-class ModXmS3mConvertor {
+class ModXmS3mStream : public Audio::AudioStream {
 private:
+	static const int NUM_MIX_BUFFER = 16384;
```
bgK: Are you sure about the mix buffer size? The original computes it dynamically in `calculate_mix_buf_len`.

yinsimei: You're right, it's better to use calculate_mix_buf_len for it.

audio/mods/mod_xm_s3m.cpp
```diff
@@ -0,0 +1,1352 @@
...
+		for (int idx = 0; idx < samples; ++idx) {
+			int ampl = mixBuf[idx];
+			if (ampl > 32767) {
+				ampl = 32767;
+			}
+			if (ampl < -32768) {
+				ampl = -32768;
+			}
+			*ptr++ = ampl & 0xFF;
+			*ptr++ = (ampl >> 8) & 0xFF;
```
bgK: The AudioStream documentation says the samples must be in native endianness. So you should just copy the 16 bits sample to the output buffer without explicitly storing the bytes in little endian order.

yinsimei: Ok, I see

yinsimei:
Thanks a lot for all the code review.
I've squashed them into one commit. It's ready to be merged, I guess.

sev-:
nice! thank you!

PR #1005, opened 2017-09-02
TITANIC: DVectorDAffineDestruction

dafioram:
1. Remove and replace all DVector usage with FVector and all DAffine usage with FPose.
This required adding in new functionality that the D versions did that the F versions didn't.
This mostly affected the Star Camera, Matrix Transform, Marked Mover, and Orientation classes. The F version was much more widely spread throughout the code then the D version so thats why float Vectors/Matrices were kept over doubles. Also the precision of float is accurate enough. This change didn't greatly improve maintainability. E.x., the exact process that lockMarker2 of StarCamera is doing is not well understood (but it works well enough).
2. The FPose inverse() function now uses matrix4Inv internally (which is more general) this fixes a drift once locked onto only one star.
3. Templated matrix4Inv input matrix type. 

Not near term:
matrix4Inv may be a good candidate for common/ and more generally common matrix classes could also be put there since other engines (like bladerunner) also use them. If this was done it would be more desirable to do a general matrix class with opengl matrix functionality, a 4x4 homogeneous matrix class, or quaternions. The current way of using a 4x3 matrix to store a 3x3 rotation and a 1x3 position/direction vector requires too many separate functions and can all be handled much simpler in a 4x4 structure (homogeneous transform).


dafioram:
4. Added fix for case of star3 locking overshooting (#9961). I hadn't observed star3 overshooting in a while, but now that I could reproduce and I've added the workaround it should be gone for good.

PR #1006, opened 2017-09-03
COMMON: Add helper macro for printing rects

csnover:
I’ve been using this macro fairly frequently in all sorts of code that uses rects during debugging, so it seems like something that would be generally useful. Since this changes `common/rect.h` it will invalidate pretty much every object cache so I didn’t want to just commit this and tie up buildbot for hours unexpectedly. Unless there is some objection to its inclusion in common, please feel free to land this whenever it won’t be too disruptive to others (or just let me know when that would be and I will do it then).

### DIFF
--- a/common/rect.h
+++ b/common/rect.h
@@ -27,6 +27,8 @@
 #include "common/util.h"
 #include "common/debug.h"
 
+#define PRINT_RECT(x) (x).left,(x).top,(x).right,(x).bottom
+
 namespace Common {
 
 /**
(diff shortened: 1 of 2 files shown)


sev-:
What about Common::Rect::debugPrint() method?

csnover:
This is a little more flexible since it allows the rect to be written into an arbitrary formatting string, e.g. `warning("Draw rect (%d, %d, %d, %d) is out of bounds of the screen; clipping it", PRINT_RECT(_drawRect));`

sev-:
OK

csnover:
Thanks!

PR #1008, opened 2017-09-03
SDL: List supported 32bpp pixel formats when using SDL2

csnover:
_hwscreen is always initialized to 16bpp so the supported 32bpp
pixel formats would never be put into the list of supported pixel
formats, making it useless for engines to query for usable 32bpp
pixel formats.

This patch changes things so that the native desktop pixel format
is at the top of the supported formats list, and all pixel formats
<= the default desktop pixel format will now show up in the list
of supported formats. ("Supported" is somewhat of a misnomer here
since there is no hardware querying beyond checking the default
desktop pixel format. SDL generally accepts a wide variety of pixel
formats and tries to convert them to whatever the hardware
supports.)

Hopefully this is pretty uncontroversial but right now I am pushing the desktop format as the top format in the list of supported formats instead of the `_hwSurface` 16bpp format, and I am not sure if this is OK or will break compatibility or is otherwise undesirable. I also didn’t bother to do anything for SDL1 since I’m considering that a legacy target at this point. Please let me know your thoughts. Thanks!

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -464,10 +464,63 @@ Common::List<Graphics::PixelFormat> SurfaceSdlGraphicsManager::getSupportedForma
 	return _supportedFormats;
 }
 
-void SurfaceSdlGraphicsManager::detectSupportedFormats() {
+#if SDL_VERSION_ATLEAST(2, 0, 0)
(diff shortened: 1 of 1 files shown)


csnover:
The documentation for `OSystem::getSupportedFormats` says that the first item in the list “must […] provide the largest color space of those formats with direct hardware support” so I guess this new behaviour is more correct than the old behaviour of returning the `_hwSurface` format, and the OpenGL backends return a 32bpp format first so it’s unlikely that this could break something since OpenGL mode would then also be broken. So I’ve pushed this.

PR #1009, opened 2017-09-03
BACKENDS: Refactor OpenGL & SDL graphics backends

csnover:
The first commit in this PR is https://github.com/scummvm/scummvm/pull/1007 which was needed for the other code to be applied cleanly (I’m being a little lazy here), so if you can, make comments on that other PR for that commit.

This patchset refactors the OpenGL and SDL graphics backends,
primarily to unify window scaling and mouse handling, and to
fix coordinate mapping between the ScummVM window and the
virtual game screen when they have different aspect ratios.

Unified code for these two backends has been moved to a new
header-only WindowedGraphicsManager class, so named because it
contains code for managing graphics managers that interact with
a windowing system and render virtual screens within a larger
physical content window.

The biggest behavioral change here is with the coordinate
system mapping:

Previously, mouse offsets were converted by mapping the whole
space within the window as input to the virtual game screen
without maintaining aspect ratio. This was done to prevent
'stickiness' when the mouse cursor was within the window but
outside of the virtual game screen, but it caused noticeable
distortion of mouse movement speed on the axis with blank
space.

Instead of introducing mouse speed distortion to prevent
stickiness, this patch changes coordinate transformation to
show the system cursor when the mouse moves outside of the virtual
game screen when mouse grab is off, or by holding the mouse inside
the virtual game screen (instead of the entire window) when mouse
grab is on.

This patch also improves some other properties of the
GraphicsManager/PaletteManager interfaces:

* Nullipotent operations (getWidth, getHeight, etc.) of the
  PaletteManager/GraphicsManager interfaces are now const
* Methods marked `virtual` but not inherited by any subclass have
  been de-virtualized
* Extra unnecessary calculations of hardware height in
  SurfaceSdlGraphicsManager have been removed
* Methods have been renamed where appropriate for clarity
  (setWindowSize -> handleResize, etc.)
* C++11 support improved with `override` specifier added on
  overridden virtual methods in subclasses (primarily to avoid
  myself accidentally creating new methods in the subclasses
  by changing types/names during refactoring)

Additional refactoring can and should be done at some point to
continue to deduplicate code between the OpenGL and SDL backends.
Since the primary goal here was to improve the coordinate mapping,
full refactoring of these backends was not completed here.

Please let me know your comments and feedback on these changes. I tested this with…

* SDL1/2 Surface/OpenGL on macOS
* SDL2 Surface/OpenGL on Linux VM

…using SCI engine, EcoQuest floppy (uses backend mouse & AR correction), GK1 CD (mode switches when playing intro with high-quality video on) lo-res & hi-res (no AR correction, forced normal1x scaler).

Hopefully this is not breaking compilation of any of the port-specific graphics managers, but I will probably need to run this patch on buildbot to verify that (and fix whatever might be broken in compilation). Looking forward to your feedback. Thanks!

### DIFF
--- a/backends/events/default/default-events.cpp
+++ b/backends/events/default/default-events.cpp
@@ -254,4 +254,30 @@ void DefaultEventManager::pushEvent(const Common::Event &event) {
 		_artificialEventSource.addEvent(event);
 }
 
+void DefaultEventManager::purgeMouseEvents() {
+	_dispatcher.dispatch();
+
+	Common::Queue<Common::Event> filteredQueue;
+	while (!_eventQueue.empty()) {
+		Common::Event event = _eventQueue.pop();
+		switch (event.type) {
+		case Common::EVENT_MOUSEMOVE:
+		case Common::EVENT_LBUTTONDOWN:
+		case Common::EVENT_LBUTTONUP:
+		case Common::EVENT_RBUTTONDOWN:
+		case Common::EVENT_RBUTTONUP:
+		case Common::EVENT_WHEELUP:
+		case Common::EVENT_WHEELDOWN:
+		case Common::EVENT_MBUTTONDOWN:
+		case Common::EVENT_MBUTTONUP:
+			// do nothing
+			break;
+		default:
+			filteredQueue.push(event);
+			break;
+		}
+	}
+	_eventQueue = filteredQueue;
+}
+
 #endif // !defined(DISABLE_DEFAULT_EVENTMANAGER)
--- a/backends/events/default/default-events.h
+++ b/backends/events/default/default-events.h
@@ -80,6 +80,7 @@ public:
 	virtual void init();
 	virtual bool pollEvent(Common::Event &event);
 	virtual void pushEvent(const Common::Event &event);
+	virtual void purgeMouseEvents() override;
 
 	virtual Common::Point getMousePos() const { return _mousePos; }
 	virtual int getButtonState() const { return _buttonState; }
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -75,7 +75,7 @@ static uint32 convUTF8ToUTF32(const char *src) {
 #endif
 
 SdlEventSource::SdlEventSource()
-    : EventSource(), _scrollLock(false), _joystick(0), _lastScreenID(0), _graphicsManager(0)
+    : EventSource(), _scrollLock(false), _joystick(0), _lastScreenID(0), _graphicsManager(0), _queuedFakeMouseMove(false)
 #if SDL_VERSION_ATLEAST(2, 0, 0)
       , _queuedFakeKeyUp(false), _fakeKeyUp()
 #endif
@@ -505,6 +505,12 @@ bool SdlEventSource::pollEvent(Common::Event &event) {
 	if (screenID != _lastScreenID) {
 		_lastScreenID = screenID;
 		event.type = Common::EVENT_SCREEN_CHANGED;
+		return true;
+	}
+
+	if (_queuedFakeMouseMove) {
+		event = _fakeMouseMove;
+		_queuedFakeMouseMove = false;
 		return true;
 	}
 
@@ -1001,6 +1007,12 @@ void SdlEventSource::resetKeyboardEmulation(int16 x_max, int16 y_max) {
 	_km.modifier = false;
 	_km.joy_x = 0;
 	_km.joy_y = 0;
+}
+
+void SdlEventSource::fakeWarpMouse(const int x, const int y) {
+	_queuedFakeMouseMove = true;
+	_fakeMouseMove.type = Common::EVENT_MOUSEMOVE;
+	_fakeMouseMove.mouse = Common::Point(x, y);
 }
 
 bool SdlEventSource::handleResizeEvent(Common::Event &event, int w, int h) {
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -158,7 +158,7 @@ int SdlEventSource::mapKey(SDLKey sdlKey, SDLMod mod, Uint16 unicode) {
 	} else if (key >= Common::KEYCODE_UP && key <= Common::KEYCODE_PAGEDOWN) {
 		return key;
 	} else if (unicode) {
-		// Return unicode in case it's stil set and wasn't filtered.
+		// Return unicode in case it's still set and wasn't filtered.
 		return unicode;
 	} else if (key >= 'a' && key <= 'z' && (mod & KMOD_SHIFT)) {
 		return key & ~0x20;
@@ -968,9 +968,8 @@ bool SdlEventSource::remapKey(SDL_Event &ev, Common::Event &event) {
 		event.kbd.keycode = Common::KEYCODE_F5;
 		event.kbd.ascii = mapKey(SDLK_F5, ev.key.keysym.mod, 0);
 	}
-	// Nap center (space) to tab (default action )
+	// Map center (space) to tab (default action)
 	// I wanted to map the calendar button but the calendar comes up
-	//
 	else if (ev.key.keysym.sym == SDLK_SPACE) {
 		event.type = Common::EVENT_KEYDOWN;
 		event.kbd.keycode = Common::KEYCODE_TAB;
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -169,13 +169,15 @@ int SdlEventSource::mapKey(SDLKey sdlKey, SDLMod mod, Uint16 unicode) {
 	}
 }
 
-void SdlEventSource::processMouseEvent(Common::Event &event, int x, int y) {
+bool SdlEventSource::processMouseEvent(Common::Event &event, int x, int y) {
 	event.mouse.x = x;
 	event.mouse.y = y;
 
 	if (_graphicsManager) {
-		_graphicsManager->notifyMousePosition(event.mouse);
+		return _graphicsManager->notifyMousePosition(event.mouse);
 	}
+
+	return true;
 }
 
 bool SdlEventSource::handleKbdMouse(Common::Event &event) {
@@ -307,8 +309,7 @@ bool SdlEventSource::handleKbdMouse(Common::Event &event) {
 
 			if (_km.x != oldKmX || _km.y != oldKmY) {
 				event.type = Common::EVENT_MOUSEMOVE;
-				processMouseEvent(event, _km.x / MULTIPLIER, _km.y / MULTIPLIER);
-				return true;
+				return processMouseEvent(event, _km.x / MULTIPLIER, _km.y / MULTIPLIER);
 			}
 		}
 	}
@@ -548,7 +549,9 @@ bool SdlEventSource::dispatchSDLEvent(SDL_Event &ev, Common::Event &event) {
 		// with a mouse wheel event. However, SDL2 does not supply
 		// these, thus we use whatever we got last time. It seems
 		// these are always stored in _km.x, _km.y.
-		processMouseEvent(event, _km.x / MULTIPLIER, _km.y / MULTIPLIER);
+		if (!processMouseEvent(event, _km.x / MULTIPLIER, _km.y / MULTIPLIER)) {
+			return false;
+		}
 		if (yDir < 0) {
 			event.type = Common::EVENT_WHEELDOWN;
 			return true;
@@ -739,12 +742,12 @@ bool SdlEventSource::handleKeyUp(SDL_Event &ev, Common::Event &event) {
 
 bool SdlEventSource::handleMouseMotion(SDL_Event &ev, Common::Event &event) {
 	event.type = Common::EVENT_MOUSEMOVE;
-	processMouseEvent(event, ev.motion.x, ev.motion.y);
+
 	// update KbdMouse
 	_km.x = ev.motion.x * MULTIPLIER;
 	_km.y = ev.motion.y * MULTIPLIER;
 
-	return true;
+	return processMouseEvent(event, ev.motion.x, ev.motion.y);
 }
 
 bool SdlEventSource::handleMouseButtonDown(SDL_Event &ev, Common::Event &event) {
@@ -765,12 +768,11 @@ bool SdlEventSource::handleMouseButtonDown(SDL_Event &ev, Common::Event &event) 
 	else
 		return false;
 
-	processMouseEvent(event, ev.button.x, ev.button.y);
 	// update KbdMouse
 	_km.x = ev.button.x * MULTIPLIER;
 	_km.y = ev.button.y * MULTIPLIER;
 
-	return true;
+	return processMouseEvent(event, ev.button.x, ev.button.y);
 }
 
 bool SdlEventSource::handleMouseButtonUp(SDL_Event &ev, Common::Event &event) {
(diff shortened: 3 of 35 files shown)


bgK:
Thanks a lot. After a cursory look, the changes look good.

After some testing using Riven on Linux/X11, I have some thoughts regarding the new behavior of the mouse cursor. Those reflect personal preferences and are not requests for changes:
1. Some platforms don't have a system cursor (I know for sure the PS3 doesn't). When calling `SDL_ShowCursor(SDL_ENABLE)`, no cursor is drawn. Those platforms will have 'sticky borders' with this change as is.
2. When the current engine hides the cursor, it feels wrong to have the system cursor show when hovering the black borders, but the game cursor not show when hovering the game area.
3. When hovering the black borders, the system cursor shows, but the game cursor is still visible. Having two mouse cursors visible at the same time in the same window does not feel right. Maybe the graphics backend should skip drawing the game cursor when the system cursor is visible.
4. Maybe at some later point it would be nice to have a way for engines to tell the backend "this is my neutral cursor" so the backend can use it to draw when hovering the black borders, to have seamless transitions with the game area.

csnover:
> Some platforms don't have a system cursor (I know for sure the PS3 doesn't). When calling SDL_ShowCursor(SDL_ENABLE), no cursor is drawn. Those platforms will have 'sticky borders' with this change as is.

If these platforms don’t have any system cursor, shouldn’t their ports be setting mouse grab immediately on start-up of a game? It doesn’t make sense to me to allow the cursor to go outside the game area if there’s never anything to interact with outside of it.

> When the current engine hides the cursor, it feels wrong to have the system cursor show when hovering the black borders, but the game cursor not show when hovering the game area.

That is an interesting point, and not one that immediately occurred to me since SCI tends to use a busy cursor instead of hiding the cursor completely (and I implemented software cursor rendering in SCI32 instead of using CursorMan so I didn’t think about integration there either). This sounds to me like a good change to make.

> When hovering the black borders, the system cursor shows, but the game cursor is still visible. Having two mouse cursors visible at the same time in the same window does not feel right. Maybe the graphics backend should skip drawing the game cursor when the system cursor is visible.

This is achievable for engines that use CursorMan to draw the mouse; other ones would either need some extra mouse-exit/mouse-enter event signals to stop rendering the mouse cursor internally, or I guess could just end up being a little inconsistent with the rest of the engines. I don’t know if anything other than SCI32 currently does software compositing of mouse cursors; at least SCI32 would need to stop calling `CursorMan.showMouse(false)` on startup, but that’s fine, it seems it doesn’t need to be doing that anyway.

Personally I feel like the current behaviour is pretty reasonable, since right now it communicates that the game is still receiving mouse inputs even though the system cursor is outside of the game area (though, do see the additional thoughts I have about this below).

Idly, I wonder if there would be any value in having engines that don’t use CursorMan communicate this with a feature flag.

> Maybe at some later point it would be nice to have a way for engines to tell the backend "this is my neutral cursor" so the backend can use it to draw when hovering the black borders, to have seamless transitions with the game area.

This is complicated. I don’t want to give a false impression to users that the game’s area of interactivity extends out to the edges of the window, though in reality right now it does. Extending interactivity into the black areas can be nice for games that require you to click exactly on the edge of the screen to perform an action, without requiring the user to enable mouse grab, though obviously it doesn’t work for both axes at the same time, so it’s a pretty terrible “solution” to that particular problem. (Maybe a better solution there would be to let users specify an extra click border in game options or something?) As such, I’m feeling like the current behaviour should probably be changed so that the mouse movement out to the edge of the screen should be sent to the game, but any additional mouse movement or actions (clicks, etc.) in the black areas should not be sent to the game. Thoughts/feelings?

Thank you very much for your excellent feedback!

rsn8887:
Great work. The changes present here will certainly break the Playstation Vita port compilation, because the Vita port inherits some functions from SurfaceSdlGraphicsManager in surfacesdl-graphics.cpp . 

I haven't done any testing yet, but I know the main challenge will be to update the file

- /backends/graphics/psp2sdl/psp2sdl-graphics.cpp

to work again with the newly changed parent class in

- /backends/graphics/surfacesdl/surfacesdl-graphics.cpp

It ranges from simple things such as renaming the variable _hwscreen to _hwScreen, to slightly more complicated things such fixing the hardware-based aspect ratio correction. The hardware can do any stretching, but is currently hard-coded to only either do a vertical stretch by a factor of 240/200, if setAspectRatioCorrection(true) is called, or not, if setAspectCorrection(false) is called. This stretch is done before blitting to the screen in the function PSP2SdlGraphicsManager::PSP2_UpdateRects , to minimize computational effort.

bgK:
> > Some platforms don't have a system cursor (I know for sure the PS3 doesn't). When calling SDL_ShowCursor(SDL_ENABLE), no cursor is drawn. Those platforms will have 'sticky borders' with this change as is.
> 
> If these platforms don’t have any system cursor, shouldn’t their ports be setting mouse grab immediately on start-up of a game? It doesn’t make sense to me to allow the cursor to go outside the game area if there’s never anything to interact with outside of it.

That's a good point. There does not seem to be an easy way to set the default mouse grab state on startup currently. I'll take a look at it once this is merged.

> > When hovering the black borders, the system cursor shows, but the game cursor is still visible. Having two mouse cursors visible at the same time in the same window does not feel right. Maybe the graphics backend should skip drawing the game cursor when the system cursor is visible.
> This is achievable for engines that use CursorMan to draw the mouse; other ones would either need some extra mouse-exit/mouse-enter event signals to stop rendering the mouse cursor internally, or I guess could just end up being a little inconsistent with the rest of the engines. I don’t know if anything other than SCI32 currently does software compositing of mouse cursors; at least SCI32 would need to stop calling CursorMan.showMouse(false) on startup, but that’s fine, it seems it doesn’t need to be doing that anyway.
> Personally I feel like the current behaviour is pretty reasonable, since right now it communicates that the game is still receiving mouse inputs even though the system cursor is outside of the game area.
> > Maybe at some later point it would be nice to have a way for engines to tell the backend "this is my neutral cursor" so the backend can use it to draw when hovering the black borders, to have seamless transitions with the game area.
> This is complicated. I don’t want to give a false impression to users that the game’s area of interactivity extends out to the edges of the window, though in reality right now it does. Extending interactivity into the black areas can be nice for games that require you to click exactly on the edge of the screen to perform an action, without requiring the user to enable mouse grab, though obviously it doesn’t work for both axes at the same time, so it’s a pretty terrible “solution” to that particular problem. (Maybe a better solution there would be to let users specify an extra click border in game options or something?) As such, I’m feeling like the current behaviour should probably be changed so that the mouse movement out to the edge of the screen should be sent to the game, but any additional mouse movement or actions (clicks, etc.) in the black areas should not be sent to the game. Thoughts/feelings?

IMO, the behavior should be all or nothing. Either as currently where the engine is still receiving events and the game cursor is visible while the mouse is in the black borders. Or having the game cursor not show and preventing the mouse events from reaching the engine, optionally using a neutral game cursor from the game. I prefer the latter solution, but the current one seems fine too. I don't think anything in between would be intuitive for the users.

csnover:
I’ve rebased this against current master and added support for making sure the system cursor is appropriately shown/hidden when the game changes the mouse visibility with OSystem. In doing this I also discovered that the backends were not handling empty cursors correctly, so there are some additional commits for fixing that and making sure that CursorManager actually sends blank cursors to the backends when they’re needed.

Incidentally, I’m wondering if CursorManager is operating correctly when it gets to the top its cursor stack. If I am reading this correctly it seems that it will end up changing the mouse visibility state to false once all the cursors are gone, but then won’t set it back to true once a new cursor is pushed and the first pushed cursor will have its visibility flag set to false. Is this really the desired behaviour? It feels to me like this is less a feature and more a workaround for the issue of backends previously not handling empty cursors correctly, but maybe I just have tunnel vision.

I am still working on hiding the in-game cursor when the system cursor goes outside the bounds of the game screen and stopping those events from entering the event queue to see how that feels in practice, and will comment again when I have some more thoughts on how this works out (and some demo code other people can try and give feedback with).

csnover:
OK, I’ve updated the patch to drop mouse events that are received outside of the active screen area, and added a separate commit to hide the in-game mouse, so these behaviours can be compared separately.

As I’ve realised during this process, one reason why I do not care for this cursor disappearing behaviour is because it doesn’t match the learned behaviour I have of ScummVM, where the in-game cursor doesn’t disappear just because I’ve moved the mouse outside of the ScummVM window. So this new behaviour doesn’t match how I am used to ScummVM working, and that’s part of why I am having a negative gut reaction to the change. I’m not sure if I have a more concrete reason based in actual user experience logic for disliking it at this point, other than perhaps that it could confuse users into thinking the game is currently not accepting inputs or that ScummVM is broken because the cursor is disappearing and reappearing.

csnover:
Sorry if you guys are getting a lot of noise from the rebases. I am hoping GitHub is smart enough to filter those out in the emails. I’m also hoping most of this can be landed in the next few days, if others’ schedules permit the ability to give feedback this weekend.

I added a couple of additional refactoring-related changes: 

* Added support for 32bpp cursors in the SurfaceSdl backend (previously these were being drawn as garbage because they were being written into a 16bpp surface designed for use by scalers);
* Fixed missing mouse motion events when a game would warp the mouse but the SDL window did not have mouse focus so SDL would not generate a mouse motion event for the warp (this was a problem in the previous implementation too, just maybe worse with SDL2 which seems to be better about calculating window mouse focus);
* Remove mouse events from the event queue after any mouse event which triggered a warp, so the additional queued events won’t make their way back to the game and jump the mouse right back to where it was before the warp (which would just re-trigger the warp action for games that warped based on mouse position, like PQ4)

That’s it from me for now.

criezy:
I played with both 84d4258 and with 7a2c150 to compare the behaviour regarding the cursor when the mouse is outside the game area. I think I prefer the behaviour where the game cursor remains visible and the engine still receives mouse events. I did get the impression at first ScummVM stopped working when the game cursor suddenly disappeared because the mouse moved just outside the game area.

But I don't have a strong opinion on that one.

criezy:
I cannot reproduce it consistently, but three times now this evening when testing the code I have seen the same issue: when starting ScummVM, I get both the GUI cursor and the system cursor visible inside the ScummVM window and I need to double click for a click to register.

criezy:
Great work overall.
However the change of signature (making function const) need to be reported to all derived classes (NullGraphicsManager and all the classes that derive from the OpenGLGraphicsManager and SurfaceSdlGraphicsManager for the various sdl-based ports). Only the `getWidth` and `getHeight` were changed, which means there is a signature mismatch for (at least) `hasFeature`, `getFeatureState`, `grabPalette`, `grabOverlay`, `getOverlayHeight` and `getOverlayWidth`.

If our oxygen documentation is up to date you would need to check `TizenGraphicsManager`, `DINGUXSdlGraphicsManager`, `GPHGraphicsManager`, `LinuxmotoSdlGraphicsManager`, `OPGraphicsManager`, `SymbianSdlGraphicsManager` and `WINCESdlGraphicsManager`.

Also your change are likely to have an impact beyond constification of some functions on those various sdl-based ports. For example you removed the virtual function `transformMouseCoordinates(Common::Point &)` from the SurfaceSdlGraphicsManager class, but this function is still reimplemented in several derived classes (`DINGUXSdlGraphicsManager`, `GPHGraphicsManager`, `LinuxmotoSdlGraphicsManager` and `WINCESdlGraphicsManager`). It would be good to have an idea of what changes will be required in all those ports so that they still work after your changes.

csnover:
> I cannot reproduce it consistently, but three times now this evening when testing the code I have seen the same issue: when starting ScummVM, I get both the GUI cursor and the system cursor visible inside the ScummVM window and I need to double click for a click to register.

I’ve noticed this a couple times, but only since upgrading to macOS 10.12 last weekend, where the window manager occasionally fails to fully give the ScummVM window focus even though it should be focused until I ⌘+tab out and back. Are you using Xcode? Has this happened at all on any other OS?

criezy:
I am using configure and the Xcode command line tools to compile. And I also upgraded this computer to macOS 10.12 last week. I have had macOS 10.12 on my laptop for a while though, and never noticed this issue. I am not using my laptop very often though, and I may just have been lucky. it's possible this issue is not related to the changes in this PR. I also never had this issue with OS X 10.9, which was what I had until last week on this computer.

I will do some more tests with master on 10.12 and with your PR on either 10.6 or 10.9 (since I still have access to those) to see if I can reproduce this issue.

criezy:
I can now confirm that the "I get both the GUI cursor and the system cursor visible inside the ScummVM window and I need to double click for a click to register" issue is not related to the changes in this PR. I tried for a while to reproduce without it, and just when I was giving up and started testing something else this happened with the current master.

csnover:
> I think I prefer the behaviour where the game cursor remains visible and the engine still receives mouse events.

All mouse events, including button events, or just the movement events?

> However the change of signature (making function const) need to be reported to all derived classes (NullGraphicsManager […]

Thank you for doing the initial research on this. This will be very helpful when I go through the ports stuff. I hope to get the final cursor behavioural issues nailed down and working to everyone’s satisfaction before I expend effort on the other ports.

> […] `transformMouseCoordinates` […]

All of the `transformMouseCoordinates` stuff should just go away entirely, since coordinate translation is now handled within the base class in the `convertVirtualToWindow/convertWindowToVirtual` methods in a way that is generalisable. All the ports should have to do is set up SDL with the correct hardware dimensions (and maybe call `handleResize` if SDL’s support for resize events is broken on a given platform).

criezy:
> All mouse events, including button events, or just the movement events?

Probably all of them. It might be strange if it handles only some of the mouse events. And that way it makes it easy to click on something that is on the edge of the game area (without having to grab the mouse).

csnover:
> Probably all of them. It might be strange if it handles only some of the mouse events. And that way it makes it easy to click on something that is on the edge of the game area (without having to grab the mouse).

I do feel like it would be surprising to click outside of the interactive area of the game and experience an interaction with an edge, especially since that edge might be very far away depending upon how stretched the window is. The thing about edge areas is something I did mention a bit earlier in this thread, and something where I believe that a more sane solution for this problem would be a separate feature that allows a definable extra edge threshold to be drawn around the game. I guess if you feel strongly about wanting to enable that here, I could add some threshold for the stretched edges, so clicks *near* the interactive area work as-if they were at the edge of the interactive area. In the interest of getting this work finished and landed, though, I would probably prefer to hold off on that for a separate PR unless this felt super-critical.

So, reviewing the feedback so far, I am thinking that probably the best middle ground for the mouse behaviour is to have the cursor remain visible, but to not respond to cursor movements outside of the interactive area, which matches how the cursor generally works now with the ScummVM window. Let me know if anyone has a strong opinions against this, otherwise I will move forward with that design for the moment and then go ahead and review the ports for changes so this feature can be landed for 1.10. (Of course, this is always something that can be revised in the future if additional feedback is received once the feature lands, since these tweaks are trivial to make in this refactored code.)

csnover:
I have updated the PR with the necessary signature changes to the other graphics backends, and updated the out-of-bounds mouse behaviour to match what I described in my previous comment.

While I was fixing up the signatures in ports, I took a quick look at what is going on, and it seems like most of these graphics manager ports are copy-pasting a bunch of code to add a half-size scaler. This approach is not good for obvious reasons, and future code to implement integer scaling modes would solve the same problem of wanting to reduce the game’s display area without all of this mess. In the interim, either the scale factor could be changed to a `Common::Rational` or `frac_t` and the ports set that instead, or the `handleResize` function can be used to reduce the virtual window size. The current approach of copy-pasting and tweaking code should not be accepted from porters.

csnover:
This has landed and I am tracking some ports compile failures. Once those are all fixed up I will close this PR.

csnover:
Compilation failures on Buildbot have been fixed, so I am closing this now. An email will be sent momentarily to the mailing list for porters about the changes.

PR #1010, opened 2017-09-04
GUI: Support double quoted debugger parameters

dreammaster:
This implements splitting of lines entered in the debugger into parameters with support for double quoted strings. In the Titanic engine, this allows for easier specification of specific views to jump to. In other engines, it may make it easier to specify things like item names that include spaces.

### DIFF
--- a/gui/debugger.cpp
+++ b/gui/debugger.cpp
@@ -290,7 +290,8 @@ bool Debugger::parseCommand(const char *inputOrig) {
 	const char *param[256];
 
 	// Parse out any params
-	char *input = strdup(inputOrig);	// One of the rare occasions using strdup is OK (although avoiding strtok might be more elegant here).
+	// One of the rare occasions using strdup is OK, since splitCommands needs to modify it
+	char *input = strdup(inputOrig);
 	splitCommand(input, num_params, &param[0]);
 
 	// Handle commands first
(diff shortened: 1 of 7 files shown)


dreammaster:
Another thing I'd like to raise, is whether we really need to specially intercept the ~ and # keys in ConsoleDialog::defaultKeyDownHandler. In Titanic, at least, resources in files are specified using the filename, a hash separator, and the resource name, like y#457.avi. Currently, I use '!' as an alternative, and do a search and replace of ! with # in the debugger commands. But it would be nicer to just be able to enter # without it automatically closing the window.

gui/debugger.cpp
```diff
@@ -288,17 +288,10 @@ bool Debugger::handleCommand(int argc, const char **argv, bool &result) {
...
 	// Parse out any params
-	char *tok = strtok(input, " ");
-	if (tok) {
-		do {
-			param[num_params++] = tok;
-		} while ((tok = strtok(NULL, " ")) != NULL);
-	} else {
-		param[num_params++] = input;
-	}
+	char *input = strdup(inputOrig);	// One of the rare occasions using strdup is OK (although avoiding strtok might be more elegant here).
```
bgK: This comment should probably be updated to say the string is duplicated because `splitCommand` modifies it.

bgK:
That seems useful! I'm all for removing the special handling of `~`  and `#` to close the debug console since those keys aren't used by default to open it anyway.

PR #1014, opened 2017-09-05
SCI: Omit SysEx delays when sending to MT-32 emulator

csnover:
Real MIDI devices, and MT-32 in particular, need delays between
SysEx messages to ensure sufficient time to receive and process
the incoming data buffer. Sending too much data too quickly to
these devices can cause them to crash with a buffer overflow.

The MT-32 emulator, on the other hand, has no problem receiving
SysEx data instantly, so skipping the delays means that games that
send lots of data to the MT-32 will start up much faster.

This seemed to work OK when I did some quick tests with SQ1VGA, SQ3, and ICEMAN, but I wanted to run this by anyone with specific knowledge in this area before committing it in case there is a risk of this causing a problem which I do not know about.

### DIFF
--- a/engines/sci/sound/drivers/midi.cpp
+++ b/engines/sci/sound/drivers/midi.cpp
@@ -134,6 +134,12 @@ public:
 		kMaxSysExSize = 264
 	};
 
+	enum Mt32Type {
+		kMt32TypeNone,
+		kMt32TypeReal,
+		kMt32TypeEmulated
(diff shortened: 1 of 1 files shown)


csnover:
I talked to @waltervn a bit about this on IRC and there weren’t any immediate concerns, so let’s give it a go. It can always be backed out later if some runs into any problems during testing.

PR #1018, opened 2017-09-11
AGOS: Fix value truncation in AGOSEngine::animateSprites

csnover:
In commit f0581bab4a4b1b4102a7fbdd4a3d54d4397e00e9 the index value
for the array was changed from a constant to a variable, but uint8
is too small to contain 293, so this value was ending up as 37.


sev-:
I wonder how it worked...

PR #1019, opened 2017-09-11
BBVS: Fix UB shifting negative signed integers

csnover:
These shifted values are replaced with their literal equivalents
as would be calculated on an x86.


sev-:
Nice!

PR #1020, opened 2017-09-16
SCI: Shivers1: Fix bad play call in room 23090

dafioram:
In the the newRoom function of script rm23v090 (room 23090),  
it was calling a play function with 5 arguments, but the play
function only takes 4 arguments. Since it looks like a fade
call it has been placed with that.

snover helped.


csnover:
Oh, I also forgot to mention that the commit message formatting is different than ticket subject formatting, so it should be something like “SCI32: Fix bad play call in Shivers room 23090”, and in the description it should have a reference to the ticket, like “Fixes Trac#10200”. As with the patch formatting, look to some of the old commits for more guidance in this regard.

dafioram:
The request changes have been made.

csnover:
Landed in 817c9e77aa6f2c40571e807888bace46b08d0142. Once again, great work!

PR #1022, opened 2017-09-18
SCI: Play MIDI version of SCI0 sound resource if user prefers it (updated)

angstsmurf:
To commemorate the anniversary of pull requests #831 and #832 by @rklaver, I've created updated versions.

It is a minimal change that fixes the conflicts. Unfortunately I don't have a real MT-32 to test it on, but at least the sound of leaving the escape pod at the start of Space Quest III is different and in stereo when "Prefer digital sound effects" is unchecked.

See the description of #831 for further information.

### DIFF
--- a/engines/sci/sound/music.cpp
+++ b/engines/sci/sound/music.cpp
@@ -350,8 +350,27 @@ void SciMusic::soundInitSnd(MusicEntry *pSnd) {
 	pSnd->time = ++_timeCounter;
 
 	if (track) {
+		bool playSample;
+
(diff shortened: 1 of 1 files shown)


rklaver:
Happy anniversary! 🎉

Seriously though, unfortunately I have not been able to fix the issues with these PRs as life got in the way... So thanks a lot for picking up the slack!

csnover:
Thank you for your help improving the SCI sound code! It is very much appreciated.

PR #1023, opened 2017-09-18
SCI: Set default MT-32 reverb before each sound (updated)

angstsmurf:
Similar to #1022, this is a minimally updated version of #832 by @rklaver.

Again, I have not really been able to test it.

See the description of #832 for further information.


csnover:
Thank you for your help improving the SCI sound code! It is very much appreciated.

PR #1024, opened 2017-09-18
Clean up use of SearchMan in fallback detection

wjp:
This makes the engines cge, cge2, sludge and directory leave `SearchMan` unchanged by their fallback detection. This fixes https://bugs.scummvm.org/ticket/10173 .


criezy:
I can confirm that the fallback detection still works for both CGE and CGE2 with those changes.

wjp:
Thanks for testing!

Since the changes to director are pretty much identical to those in sludge, I'm going to assume they are ok too.

PR #1026, opened 2017-09-20
SCI32: Fix pointer color when quitting for phant1

dafioram:
When quitting the game at the main menu and hitting no
or quitting the game while playing the pointer color
over the buttons will be black the first time rather
than red. This fix makes it so they will be red.

This was done by removing a check in HotSpot::doit()
which checks the global193 value. Removing this check
fixes the problem.

Fixes Trac#9977.

snover helped.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -3684,8 +3684,37 @@ static const uint16 phant1RatPatch[] = {
 	PATCH_END
 };
 
+// In Phantasmagoria the cursor's hover state will not trigger on any of the
+// buttons in the main menu after returning to the main menu from a game, or
+// when choosing "Quit" on the main menu and then cancelling the quit in the
+// confirmation dialogue, until another button has been hovered and unhovered
+// once.
+// This happens because the quit confirmation dialogue creates its own
+// event handling loop which prevents the main event loop from handling the
+// cursor leaving the button (which would reset global 193 to 0), and the
+// dialogue does not reset global193 itself, so it remains at 2 until a new
+// button gets hovered and unhovered.
+// There is not enough space in the confirmation dialogue code to add a reset
+// of global 193, so we just remove the check entirely, since it is only used
+// to avoid resetting the cursor's view on every mouse movement, and this
+// button type is only used on the main menu and the in-game control panel.
+//
+// Applies to at least: English CD
+static const uint16 phant1RedQuitCursorSignature[] = {
+	SIG_MAGICDWORD,
+	0x89, 0xc1,                   // lsg $c1
+	0x35, 0x02,                   // ldi 02
+	SIG_END
+};
+
+static const uint16 phant1RedQuitCursorPatch[] = {
+	0x33, 0x05,                   // jmp [past global193 check]
+	PATCH_END
+};
+
 //          script, description,                                      signature                        patch
 static const SciScriptPatcherEntry phantasmagoriaSignatures[] = {
+	{  true,    23, "make cursor red after clicking quit",         1, phant1RedQuitCursorSignature,    phant1RedQuitCursorPatch },
 	{  true,   901, "fix invalid array construction",              1, sci21IntArraySignature,          sci21IntArrayPatch },
 	{  true,  1111, "ignore audio settings from save game",        1, phant1SavedVolumeSignature,      phant1SavedVolumePatch },
 	{  true, 20200, "fix broken rat init in sEnterFromAlcove",     1, phant1RatSignature,              phant1RatPatch },


dafioram:
Here is a diff of the disasm function.

https://gist.github.com/dafioram/f955d323145f7565508c4087550d1d36/revisions?diff=split

For the diff I was expecting it to remove lines 61-66 by replacing lines 61-70 with 67-70. Instead it replaced 61-64 with 67-70, but it fixes the bug.

wjp:
The script patcher puts your bytes at the location of the signature, but it can't just remove any other bytes, because that would move everything around and break all memory offsets.

You can add some random instructions to overwrite the remaining bytes too. (Search for "waste bytes" in `script_patches.cpp` for examples.)

dafioram:
Thanks wjp thats what I needed. I decided to go with a jump instead of a bunch of negations (nots are used as waste bytes mostly) for filler since I am skipping a bunch of code I think a jump reflects this. This change now looks the way I expect it too and it works.

The new disasm difference is given here:
https://gist.github.com/dafioram/3699831914705eef1a8e6dbb3b53def0/revisions?diff=split

You will see that I have left most of the original code the same and just put a jump at the beginning. Line 62 is different from before because the jump call on line 61 is taking the first byte of the next line (bnt) so that is why the next line now shows up as a le? that makes sense since that is the 2nd byte of the previous line 62, but rest assured that will not get executed because of the jump.

engines/sci/engine/script_patches.cpp
```diff
@@ -3684,12 +3684,42 @@ static const uint16 phant1RatPatch[] = {
...
+	0x31, 0x12,                   // bnt 12 [00dd]
+	0x39, 0x0f,                   // pushi 0f
+	0x78,                         // push1
+	0x76,                         // push0
+	0x38, 0x4c, 0x01,             // pushi 014c
+	SIG_END
+};
+
+static const uint16 phant1RedQuitPointerPatch[] = {
+	0x33, 0x08,                   // jmp [past global193 check]
```
csnover: This patch is overwriting the end of the `((not (gDButton respondsTo: #exitDir))` check, because the signature starts at `not, bnt`, not at `lsg c1`. Either use the PATCH_ADDTOOFFSET macro to forward to the correct place to apply the patch (relative to the start of the signature), or remove the unnecessary extra bytes in the signature prior to the actual place we want to rewrite.

engines/sci/engine/script_patches.cpp
```diff
@@ -3684,12 +3684,42 @@ static const uint16 phant1RatPatch[] = {
...
+// Applies to at least: English CD
+static const uint16 phant1RedQuitPointerSignature[] = {
+	0x18,                         // not
+	SIG_MAGICDWORD,
+	0x31, 0x25,                   // bnt 25 [00e9]
+	0x89, 0xc1,                   // lsg c1
+	0x35, 0x02,                   // ldi 02
+	0x1c,                         // ne?
+	0x31, 0x12,                   // bnt 12 [00dd]
+	0x39, 0x0f,                   // pushi 0f
```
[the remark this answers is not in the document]
dafioram: I am not sure what this selector is here is the relevant code:
			((not (gDButton respondsTo: #exitDir))
				(if (!= global193 2)
					(gGGInvCursor loop: 0 setCel: 0)
					(= global193 2)
				)
				(gUser message: theVerb)
			)
maybe it is  loop? Still new to lisp.

engines/sci/engine/script_patches.cpp
```diff
@@ -3684,12 +3684,42 @@ static const uint16 phant1RatPatch[] = {
...
+	0x78,                         // push1
+	0x76,                         // push0
+	0x38, 0x4c, 0x01,             // pushi 014c
```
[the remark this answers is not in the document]
dafioram: I am not sure what this selector is here is the relevant code:
			((not (gDButton respondsTo: #exitDir))
				(if (!= global193 2)
					(gGGInvCursor loop: 0 setCel: 0)
					(= global193 2)
				)
				(gUser message: theVerb)
			)
maybe it is  setCel?

dafioram:
Done. I didn't need to make the signature as long as I thought so there is no need to identify selectors. I am still uncertain when to use a _UINT16/_UINT8 vs not when it is a number (not a selector).  Should my bnt, lsg, and ldi all use the _UINT8 on their numbers?

engines/sci/engine/script_patches.cpp
```diff
@@ -3684,8 +3684,38 @@ static const uint16 phant1RatPatch[] = {
...
+	0x31, 0x25,                   // bnt $25
+	0x89, 0xc1,                   // lsg $c1
+	0x35, 0x02,                   // ldi 02
+	0x1c,                         // ne?
+	SIG_END
+};
+
+static const uint16 phant1RedQuitCursorPatch[] = {
+	PATCH_ADDTOOFFSET(+1),        // skip not
+	0x33, 0x07,                   // jmp [past global193 check]
```
csnover: This patch is still overwriting the jump instruction for the _previous_ condition, so this earlier advice still applies:

> This patch is overwriting the end of the `((not (gDButton respondsTo: #exitDir))` check, because the signature starts at `not, bnt`, not at `lsg c1`. Either use the `PATCH_ADDTOOFFSET` macro to forward to the correct place to apply the patch (relative to the start of the signature), or remove the unnecessary extra bytes in the signature prior to the actual place we want to rewrite.

`bnt $25` is the end of the previous check. It is not the start of the global 193 check.

Nowhere else in this script is there an `lsg $c1, ldi 2`, and the patch requires only 2 bytes, so the signature does not need to be any longer than those two instructions. Therefore, the forwarding with `PATCH_ADDTOOFFSET` becomes unnecessary here since the patch will start exactly where we want to replace with a jump.

engines/sci/engine/script_patches.cpp
```diff
@@ -3684,8 +3684,38 @@ static const uint16 phant1RatPatch[] = {
...
+// when choosing "Quit" on the main menu and then cancelling the quit in the 
+// confirmation dialogue, until another button has been hovered and unhovered once.
+// This happens because the quit confirmation dialogue creates its own event handling loop 
+// which prevents the main event loop from handling the cursor leaving the button 
+// (which would reset global 193 to 0), and the dialogue does not reset global 193 itself, 
+// so it remains at 2 until a new button gets hovered and unhovered. 
+// There is not enough space in the confirmation dialogue code to add a reset of global 193, 
+// so we just remove the check entirely, since it is only used to avoid resetting the cursor’s 
+// view on every mouse movement, and this button type is only used 
+// on the main menu and the in-game control panel.
```
csnover: Please make sure these comments are wrapped at 80 columns.

engines/sci/engine/script_patches.cpp
```diff
@@ -3684,8 +3684,38 @@ static const uint16 phant1RatPatch[] = {
 	PATCH_END
 };
 
+// In Phantasmagoria the cursor’s hover state will not trigger on any of the buttons
```
csnover: Others have complained at me in the past when I sneaked some Unicode into the source code, so please make sure these beautiful, perfect, typographically correct apostrophes are turned into abominable ASCII typewriter apostrophes. :)

dafioram:
Done.

csnover:
My apologies about the length of time it took to get back around on this. I missed your update here that you had finished, and then got wrapped up with some other work. Everything looks great now, so I’ve merged this patch! Thank you again for your significant effort on this work, it is appreciated.

PR #1027, opened 2017-09-20
COMMON: Remove the DisposeAfterUse default value from dynamic write memory streams constructors

bgK:
The default value was DisposeAfterUse::NO, which made it easy to accidentally leak memory by omitting to specify a value.

Also fix various memory leaks due to not setting DisposeAfterUse::YES when constructing dynamic memory streams.
Some non trivial changes were needed in the XEEN engine that @dreammaster may want to check.

I was intending to mark the `MemoryReadWriteStream` and `MemoryWriteStreamDynamic` classes as final because subclassing those seems like a bad idea to me (And they have a non-virtual destructor making it easy to leak memory in a creative way). However there are two remaining subclasses in LASTEXPRESS and SCI. Thoughts?

### DIFF
--- a/audio/decoders/quicktime.cpp
+++ b/audio/decoders/quicktime.cpp
@@ -344,7 +344,7 @@ bool QuickTimeAudioDecoder::QuickTimeAudioTrack::isOldDemuxing() const {
 
 AudioStream *QuickTimeAudioDecoder::QuickTimeAudioTrack::readAudioChunk(uint chunk) {
 	AudioSampleDesc *entry = (AudioSampleDesc *)_parentTrack->sampleDescs[0];
-	Common::MemoryWriteStreamDynamic *wStream = new Common::MemoryWriteStreamDynamic();
+	Common::MemoryWriteStreamDynamic *wStream = new Common::MemoryWriteStreamDynamic(DisposeAfterUse::NO);
 
 	_decoder->_fd->seek(_parentTrack->chunkOffsets[chunk]);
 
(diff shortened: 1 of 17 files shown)


dreammaster:
I reviewed the Xeen changes, and they all look good to me

criezy:
This looks like a good change to me. It was indeed all to easy to introduce memory leaks with a default of DisposeAfterUse::No in those stream constructors (as I am well aware since I introduced the one in sword1 only two months ago :P).

And it looks like the change to the networking code should fix [bug #9718 CLOUD: memory leak](https://bugs.scummvm.org/ticket/9718) as well :)

csnover:
This is fantastic. Thank you for doing this.

As far as the usage in SCI goes, I don’t think there would be a problem changing from inheritance to composition, so I will work on that shortly and push the work to this PR once it’s ready.

csnover:
So sorry. I have one of those brains that doesn’t think smart things until after the right time to think them.

> (And they have a non-virtual destructor making it easy to leak memory in a creative way)

The base `Common::Stream` destructor is virtual, so all these ones are too. So I don’t think there is any real benefit of making these final-classes.

bgK:
> The base Common::Stream destructor is virtual, so all these ones are too. So I don’t think there is any benefit of making these final-classes, at least from the perspective of avoiding memory problems.

Right!

I've updated the NetworkReadStream commit to include a reference to bug #9718.

Thanks to you three for your reviews. I'll merge this during the week end.

PR #1029, opened 2017-09-21
COMMON: Add new standard methods to Array

csnover:
This PR adds:

1. New [constructor signatures (2) and (3)](http://en.cppreference.com/w/cpp/container/vector/vector), for common constructions to create arrays of a specific size.
2. New [data method](http://en.cppreference.com/w/cpp/container/vector/data), for retrieving a pointer to the underlying storage of a Common::Array, even when the array is empty.

I don’t feel like these should be at all controversial since they are just imports from the C++ standard library, and there are tests, but let me know if there are any objections.

### DIFF
--- a/common/array.h
+++ b/common/array.h
@@ -87,10 +87,10 @@ public:
 	 * Construct an array by copying data from a regular array.
 	 */
 	template<class T2>
-	Array(const T2 *data, size_type n) {
+	Array(const T2 *array, size_type n) {
 		_size = n;
 		allocCapacity(n);
-		uninitialized_copy(data, data + _size, _storage);
+		uninitialized_copy(array, array + _size, _storage);
 	}
 
 	~Array() {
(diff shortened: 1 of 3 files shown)


dreammaster:
It all seems to be fairly straightforward and clean. Though I would suggest fixing the shadowing issue:
./common/array.h:90:37: warning: declaration of ‘data’ shadows a member of 'this'

As Travis CI demonstrates, it's not fun, considering the number of files throughout the ScummVM that include common/array.h

csnover:
Sorry about that; clang (and GCC 5+) don’t complain about shadows that can’t possibly conflict so I didn’t see those. That’s been fixed. I also fixed the CI failure, which was caused by using TS_ASSERT_EQUALS to compare pointer to nullptr. (For whatever reason Apple Clang defines nullptr even in C++98 mode so this usually happens when I write new tests.)

bgK:
Thanks! I've been missing the count constructor frequently.

csnover:
I haven’t heard anything else so I will land these at the weekend unless anyone asks me to hold off.

sev-:
All looks good, merging.

PR #1030, opened 2017-09-24
SCI: implement strncpy with memmove

moralrecordings:
The C spec says that for strcpy and strncpy, the source and destination cannot have overlapping buffers. If memory serves, [sometime back in 2011](https://sourceware.org/bugzilla/show_bug.cgi?id=12518) glibc was changed on x86 to copy memory backwards, because of some alleged infinitessimal speed boost. Since then, any code that assumed a strcpy with overlapping buffers would work because the source was before the destination will very subtly break.

There's at least one instance of a SCI game expecting strcpy to work with overlapping buffers, so this patch replaces the underlying call with a strlen() + memmove(), which is guaranteed to work.

Fixes the flamingo puzzle in Island of Dr. Brain v1.1 on Linux.

### DIFF
--- a/engines/sci/graphics/palette32.cpp
+++ b/engines/sci/graphics/palette32.cpp
@@ -531,18 +531,18 @@ void GfxPalette32::updateHardware() {
 #endif
 
 #ifdef ENABLE_SCI32_MAC
-	if (g_sci->getPlatform() != Common::kPlatformMacintosh) {
-		// The last color must always be white
-		bpal[255 * 3    ] = 255;
-		bpal[255 * 3 + 1] = 255;
-		bpal[255 * 3 + 2] = 255;
+	if (g_sci->getPlatform() == Common::kPlatformMacintosh) {
+		bpal[255 * 3    ] = 0;
+		bpal[255 * 3 + 1] = 0;
+		bpal[255 * 3 + 2] = 0;
 	} else {
 #else
 	{
 #endif
-		bpal[255 * 3    ] = 0;
-		bpal[255 * 3 + 1] = 0;
-		bpal[255 * 3 + 2] = 0;
+		// The last color must always be white
+		bpal[255 * 3    ] = 255;
+		bpal[255 * 3 + 1] = 255;
+		bpal[255 * 3 + 2] = 255;
 	}
 
 	// If the system is in a high color mode, which can happen during video


moralrecordings:
For kicks I wrote up the troubleshooting process here - https://moral.net.au/writing/2017/09/23/sierra_bug/

wjp:
Great catch, thanks! Very interesting write-up too.

By the way, our data file hashes are only of the first 5000 bytes, which may be why it didn't appear to match for you.

Could you add a reference to the SCI script method that depends on this behaviour (`flamingo::init` in Island of Dr Brain, it seems?) in the commit message? That'll make it easier to test in the future if we ever want to change these functions.

This might also affect the `memcpy` function in a similar way.

A complicating matter is that `memmove` might not always do what the original Sierra interpreters did for overlapping memory areas. From this flamingos case, it seems to do the right thing when copying overlapping strings forward, but we need to test (or reverse) if this approach matches SSCI when copying strings backward too. (From a quick glance, KQ5 copies bytes in a forward loop.)

moralrecordings:
Can do. I was actually about to panic and drop the request, as after rebasing restoring save games made the SCI interpreter crash, but that appears to have been caused by an unrelated commit chain from yesterday?

I thought about memcpy, but was hesitant to go ahead and change things that there wasn't a concrete test case for. Easy enough to do, just replace ::memcpy with ::memmove. As for the reverse case... offhand I can't think of a scenario where you'd *want* the weird inconsistent strncpy output, but such a requirement would still appear as a bug on (what I assume is) the majority of platforms that do forward copy.

csnover:
> Can do. I was actually about to panic and drop the request, as after rebasing restoring save games made the SCI interpreter crash, but that appears to have been caused by an unrelated commit chain from yesterday?

I’m not able to reproduce a crash when restoring Island of Dr Brain save games when building from master. Could you provide more detail on what you are experiencing?

csnover:
Sorry about the save game problem, the serialiser was losing its version information. It should be OK again, though you will need to regenerate any save game made in the last half-day or so. :\

csnover:
Oops. I was trying to update the PR with an alternate implementation using an explicit forward copy, and GitHub accepted most of the push but then decided the replacement commit wasn’t valid and closed the PR. I can’t reopen this now, so I guess I will just open a new one with the different patch. Thanks, @moralrecordings, for your great write-up, work on isolating the problem, and the initial patch!

PR #1031, opened 2017-09-26
SCI: Replace uninitalized call when shoplifting in SQ4CD

dafioram:
When in Galaxy Galleria if you go into the Software
store and try to shoplift the SQ4 Hintbook then the game will
crash after you leave and are electricuted.

This crash was happening because the temporary variable 0 was
trying to have its value put onto the stack, but it is
uninitalized. A workaround for this has been done.

Fixes Trac#10229.

### DIFF
--- a/engines/sci/engine/workarounds.cpp
+++ b/engines/sci/engine/workarounds.cpp
@@ -412,6 +412,7 @@ const SciWorkaroundEntry uninitializedReadWorkarounds[] = {
 	{ GID_SQ1,            -1,   703,  0,         "firePulsar", "changeState",     sig_uninitread_sq1_1,     0, { WORKAROUND_FAKE,   0 } }, // export 1, but called locally (when shooting at aliens)
 	{ GID_SQ4,            -1,   398,  0,            "showBox", "changeState",                     NULL,     0, { WORKAROUND_FAKE,   0 } }, // CD: called when rummaging in Software Excess bargain bin
 	{ GID_SQ4,            -1,   928, -1,           "Narrator", "startText",                       NULL,  1000, { WORKAROUND_FAKE,   1 } }, // CD: happens in the options dialog and in-game when speech and subtitles are used simultaneously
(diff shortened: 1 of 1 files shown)


csnover:
There is an entire system in workarounds specifically for addressing uninitialized temp reads. Look to `uninitializedReadWorkarounds` in workarounds.cpp.

dafioram:
I updated the commit and comments above.

engines/sci/engine/workarounds.cpp
```diff
@@ -412,6 +412,7 @@ const SciWorkaroundEntry uninitializedReadWorkarounds[] = {
 	{ GID_SQ1,            -1,   703,  0,         "firePulsar", "changeState",     sig_uninitread_sq1_1,     0, { WORKAROUND_FAKE,   0 } }, // export 1, but called locally (when shooting at aliens)
 	{ GID_SQ4,            -1,   398,  0,            "showBox", "changeState",                     NULL,     0, { WORKAROUND_FAKE,   0 } }, // CD: called when rummaging in Software Excess bargain bin
 	{ GID_SQ4,            -1,   928, -1,           "Narrator", "startText",                       NULL,  1000, { WORKAROUND_FAKE,   1 } }, // CD: happens in the options dialog and in-game when speech and subtitles are used simultaneously
+	{ GID_SQ4,           395,   395,  -1,   "fromStoreScript", "changeState",                     NULL,    -1, { WORKAROUND_FAKE,   0 } }, // CD: happens, when shoplifting in Galaxy Galleria  - bug #10229
```
csnover: There is a whitespace error in the comment before the hyphen, and the index should be 0, not -1, unless this function is using many uninitialised temps.

dafioram: vv t shows at least 3 uninitialised temps at that point, but I don't know if it uses the others or not. Should I still change the index to 0?

csnover: Temps are always uninitialised at the beginning of a function. This is normal. If an uninitialised read occurs, the engine will raise an error just like the one from the original ticket, which then needs to get evaluated and the correct fake value entered into the workarounds table.

dafioram:
Commit updated.

csnover:
Thanks! This has landed.

PR #1032, opened 2017-09-26
SCI: Workaround for uninit reads during wordsearch for castlebrain

dafioram:
During the wordsearch puzzle (room 320 click left door) the
game will crash because of an uninitalized read of temp
variables in word::dispatchEvent (which gets called a lot),
if the player clicks the same letter or different letters
aggressively.

A workaround has been added to set all uninitalized temp variables
to 0 in room 320 script 325 to prevent this crash. The index for
the workaround was set to -1 to rather than 14 (temp var 14)
because if just var 14 is fixed then temp var 15 will crash with
an uninitialized read so there is atleast 2 uninitialized temp
vars that are attempted to be read during this time leading to
the bug.

Fixes Trac#9783.


csnover:
Thanks for the patch! This has landed in ec1cfcbf47593545357ca3730fdb36b345f669de.

PR #1038, opened 2017-10-02
MACOSX: Allow selecting device for CoreMidi

criezy:
After reading this [forum post](http://forums.scummvm.org/viewtopic.php?t=14422) I decided to take a look at the CoreMIDI API documentation, and since it seemed easy enough changed the code to try to list available devices so that the user can select which one to use.

I tried this with a software synthesiser and it seems to work correctly. However since I know absolutely nothing about MIDI and don't have any MIDI hardware I decided to open this pull request so that others can have a look, and hopefully try it, before this gets merged.

### DIFF
--- a/backends/midi/coremidi.cpp
+++ b/backends/midi/coremidi.cpp
@@ -53,7 +53,7 @@ http://lists.apple.com/archives/coreaudio-api/2003/Jul/msg00137.html
  */
 class MidiDriver_CoreMIDI : public MidiDriver_MPU401 {
 public:
-	MidiDriver_CoreMIDI();
+	MidiDriver_CoreMIDI(ItemCount device);
 	~MidiDriver_CoreMIDI();
 	int open();
 	bool isOpen() const { return mOutPort != 0 && mDest != 0; }
@@ -62,13 +62,14 @@ public:
 	void sysEx(const byte *msg, uint16 length);
 
 private:
+	ItemCount mDevice;
 	MIDIClientRef	mClient;
 	MIDIPortRef		mOutPort;
 	MIDIEndpointRef	mDest;
 };
 
-MidiDriver_CoreMIDI::MidiDriver_CoreMIDI()
-	: mClient(0), mOutPort(0), mDest(0) {
+MidiDriver_CoreMIDI::MidiDriver_CoreMIDI(ItemCount device)
+	: mDevice(device), mClient(0), mOutPort(0), mDest(0) {
 
 	OSStatus err;
 	err = MIDIClientCreate(CFSTR("ScummVM MIDI Driver for OS X"), NULL, NULL, &mClient);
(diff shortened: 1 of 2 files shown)


digitall:
@criezy : Thanks for looking at this. Do you think you can do the same for the CoreAudio API as well?
https://github.com/scummvm/scummvm/blob/master/backends/midi/coreaudio.cpp#L325

criezy:
Yes, I saw that this was also missing from the CoreAudio driver. This one seemed more complicated though, so I decided to start with CoreMIDI. I may look at the CoreAudio tomorrow and see if I can make sense of it.

digitall:
Ah OK. Thanks for looking at it anyway. Hopefully we can find a Max OSX user to test this, maybe the original forum poster.

criezy:
After another look at this, and unless somebody can give me a good argument why this would be needed, I have decided to leave the CoreAudio MIDI driver as it is and not attempt to list available devices.

CoreMIDI was not functional without a MIDI device, thus the changes in this PR have two benefits:
- It is not listed when no MIDI device is available.
- It allows selecting the MIDI device to use when several are available.

However with CoreAudio we ca usen, and are currently using, the builtin Apple DLS softsynth. This provides a way to play MIDI on any macOS computer (and we even support using a custom soundfont with it). I think it is fine keeping our CoreAudio MIDI code as it is to use the builtin softsynth, and use  CoreMIDI when we want to use a hardware MIDI device.

csnover:
That seems fine. As far as I can tell, glancing at the Apple documentation, Core MIDI is just the hardware MIDI part of the superset of Core Audio services. I wonder if it would be worth simply changing the label from “Core Audio” to something like “Apple DLS softsynth”?

criezy:
> I wonder if it would be worth simply changing the label from “Core Audio” to something like “Apple DLS softsynth”?

I was thinking the same thing yesterday, and I have now made the change. I also removed the TODO about listing devices.

backends/midi/coremidi.cpp
```diff
@@ -198,17 +199,40 @@ class CoreMIDIMusicPlugin : public MusicPluginObject {
...
-	MusicDevices devices;
 	// TODO: Return a different music type depending on the configuration
-	// TODO: List the available devices
-	devices.push_back(MusicDevice(this, "", MT_GM));
+
+	MusicDevices devices;
+	char buffer[256];
+	ItemCount deviceCount = MIDIGetNumberOfDestinations();
+	for (ItemCount i = 0 ; i < deviceCount ; ++i) {
+		MIDIEndpointRef dest = MIDIGetDestination(i);
```
csnover: De-duplicate this common name retrieval with a `Common::String getDeviceName(ItemCount deviceIndex) const` or `bool getDeviceName(ItemCount deviceIndex, Common::String &outName) const`?

criezy:
Thank you for your review. I followed your suggestions and also made a few additional small changes.
And since the user on the forum also confirmed that this works I have now merged these commits.

PR #1040, opened 2017-10-05
GUI: Rework formatting of the Updates Dialog

rootfather:
Currently, the "Updates Dialog" that appears when ScummVM is run for the first time uses some text that has the tooltip attribute and is therefore shown in a smaller font.

In my opinion, it looks better if the fonts stay the same within the dialog. Additionally, I aligned everything to the left and removed the brackets around the "you can always..." message.

This is solely my own opinion. If you think it's good as it was before, feel free to discard this PR.

**Before**
scummvm winsparkle tooltip

**After**
scummvm winsparkle no tooltip leftalign

In case this PR gets accepted, is it reasonable to merge the two messages into one?

### DIFF
--- a/gui/updates-dialog.cpp
+++ b/gui/updates-dialog.cpp
@@ -52,7 +52,7 @@ UpdatesDialog::UpdatesDialog() : Dialog(30, 20, 260, 124) {
 		"which requires access to the Internet.\n"
   		"\n"
 		"Would you like to enable this feature?");
-	const char *message2 = _("(You can always enable it in the options dialog on the Misc tab)");
+	const char *message2 = _("You can always enable it in the options dialog on the Misc tab.");
 
 	// First, determine the size the dialog needs. For this we have to break
 	// down the string into lines, and taking the maximum of their widths.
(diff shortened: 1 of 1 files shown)


criezy:
I am not sure what is the best here. Looking at your second screenshot I think it might look more natural without line breaks between the first two sentences: "ScummVM now supports automatic check for updates which requires access to the internet. Would you like to enable this feature?" and keep the third sentence separate to have a break after the question so that it remains visible. Or maybe even move the third sentence below the checkbox and popup widget.

And that is not really related, but maybe the wording of the third sentence could be changed to something like "You can change this setting later in the Misc tab of the Options dialog."

And since I started to hijack this PR, let me continue ;)
I remember there was some discussion about usability when this was implemented and I have never been fully happy about it. I don't really like for example that the dialog asks a question for which the expected reply would be `Yes` or `No`, but there is no Yes and No buttons. Also the popup has a `Never` entry, and it feels a bit weird to be able to toggle on the `Check for updates automatically` and then select `Never`. And that this means there are two different ways to not enable automatic checks. And I don't think the purpose of the `Cancel` button is necessarily obvious either. Would it bring back the dialog the next time we start ScummVM because the user did not make a choice? Or does it mean the choice is to not enable automatic checks? (I know the answer is the latter, but I don't think this is obvious for a normal user).

I am thinking it might be best to simplify it, maybe along the line of:
```
ScummVM now supports automatic check for updates
which requires access to the internet. Would you
like to enable this feature?

Update check: [ Never                         v]

You can change this setting later in the Misc tab
in the Options dialog.

                                   [ Proceed ]
```

The default would be Never, meaning the user needs to make a change to opt in into the automatic checks.

However the button label still does not match the question asked. Maybe the wording could be improved here? Maybe the question could even be replaced with something like "You can enable this feature below."

csnover:
I agree with @criezy’s comments on this. His proposed dialogue change seems like the best route forward to me. For the wording I would say something like “Please choose how often you would like to check for updates:” instead of “Would you like to enable this feature?”. And of course then it can be collapsed into a single message.

rootfather:
Thank you @criezy and @csnover for your input. What you guys propose is by far a better solution than the current approach.

Unfortunately, it is far beyond my knowledge to address this issue properly, so I have to rely on you to get this implemented.

sev-:
I like Thierry's proposal and will implement it once I have some time. Please keep this PR open for my reference.

rootfather:
Thank you @sev-!

sev-:
Implemented

rootfather:
Thanks!

PR #1041, opened 2017-10-06
Fix undefined behaviour in variadic functions

criezy:
Passing an argument of a type that undergoes default argument promotion as last argument of a variadic function results in an undefined behaviour. This includes bool, char, short, and enumerations.

See http://en.cppreference.com/w/cpp/language/variadic_arguments#Default_conversions

In ScummVM some code was using an enum or an uint16 as last argument of variadic functions, and they have been replaced by int and uint respectively in this PR to avoid the undefined behaviour.

Note: they were reported by clang when compiling with -Wvarargs

### DIFF
--- a/engines/tsage/core.cpp
+++ b/engines/tsage/core.cpp
@@ -2358,8 +2358,11 @@ int SceneObject::checkRegion(const Common::Point &pt) {
 	return regionIndex;
 }
 
-void SceneObject::animate(AnimateMode animMode, ...) {
-	_animateMode = animMode;
+// The parameter to the function below should really be an AnimateMode value.
+// However passing an enum type as last argument of a variadic function may
+// result in undefined behaviour.
+void SceneObject::animate(int animMode, ...) {
+	_animateMode = (AnimateMode)animMode;
 	_updateStartFrame = g_globals->_events.getFrameNumber();
 	if (_numFrames)
 		_updateStartFrame += 60 / _numFrames;
--- a/engines/hugo/display.cpp
+++ b/engines/hugo/display.cpp
@@ -332,7 +332,7 @@ int16 Screen::mergeLists(Rect *list, Rect *blist, const int16 len, int16 blen) {
  * Process the display list
  * Trailing args are int16 x,y,dx,dy for the D_ADD operation
  */
-void Screen::displayList(Dupdate update, ...) {
+void Screen::displayList(int update, ...) {
 	debugC(6, kDebugDisplay, "displayList()");
 
 	int16         blitLength = 0;                   // Length of blit list
--- a/engines/hugo/display.h
+++ b/engines/hugo/display.h
@@ -61,7 +61,7 @@ public:
 
 	void     displayBackground();
 	void     displayFrame(const int sx, const int sy, Seq *seq, const bool foreFl);
-	void     displayList(Dupdate update, ...);
+	void     displayList(int update, ...);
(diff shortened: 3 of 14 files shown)


engines/tsage/core.cpp
```diff
@@ -2358,8 +2358,11 @@ int SceneObject::checkRegion(const Common::Point &pt) {
 	return regionIndex;
 }
 
-void SceneObject::animate(AnimateMode animMode, ...) {
-	_animateMode = animMode;
+// The parameter to the function below should really be an AnimateMode value.
+// However passing an enum type as last argument of a variadic function may
+// results in undefined behaviour.
```
csnover: “may results”

criezy:
I am not 100% sure either that all of those changes are needed, nor that they are the best way to fix the issue. This is why I opened this pull request to seek feedback rather than directly push to master.

csnover:
OK, so having a discussion with someone more knowledgeable than me tonight, this does appear to be one of these hidden sharp corners of the specification where you have to get away from the concrete “how things actually work in compilers now” and think about things in the abstract “what the standard says compilers are allowed to do if they want”. And so Clang is right about this.

The C++ specification states in [cstdarg.syn](http://eel.is/c++draft/cstdarg.syn#1):

> If the parameter parmN is […] of a type that is not compatible with the type that results when passing an argument for which there is no parameter, the behavior is undefined.

I am told that a “type that results when passing an argument for which there is no parameter” refers to a rest (...) argument. e.g. in `void foo(int, ...); foo(1, 2);`, `2` is “an argument for which there is no parameter”.

This means that whatever `parmN` is, it has to be compatible to one of those types used for a rest argument. [expr.call](http://eel.is/c++draft/expr.call#9) tells us what these final rest types are in two steps. First, by first narrowing the valid types:

> […] if the argument does not have arithmetic, enumeration, pointer, pointer to member, or class type, the program is ill-formed.

Then, by describing the promotions to get from these argument types to their final rest types:

> If the argument has integral or enumeration type that is subject to the [integral promotions](http://eel.is/c++draft/conv.prom), or a floating-point type that is subject to the [floating-point promotion](http://eel.is/c++draft/conv.fpprom), the value of the argument is converted to the promoted type before the call.

Following those promotions, and the previous list of types, this leaves us with int, double, pointer, pointer-to-member, trivial class, and scoped enumeration as the non-implementation-defined compatible types for `parmN`, which is exactly how Clang warns us here.

In C++11 and later, none of this matters because variadic templates allow us to pass all types through a function parameter pack which gets expanded such that there is *never* an unnamed parameter, so there are no promotions and no incompatibilities related to promotions. So it is another example of a nice feature of C++11 that could be useful to us.

criezy:
@snover: Thank you for taking the time to chase information on the meaning of what is written in the C++ standard. It was confusing me, so I was relying on on the interpretation of it by unknown people on the internet. It's good to see a clear explanation of that c++ standard and to see that it is consistent with what I previously saw.

So to me it seems that the conclusion is that we should merge this to be on the safe side. Since one of the feature of ScummVM is to be very portable, it is not impossible that somebody out there will use a compiler whose behaviour is not the one we want. Variadic template functions is a nice way to fix this, but unfortunately not one we can use as it would break some toolchains that are using compilers too old to support it.

Do you agree?

csnover:
Yes, this should be merged. The variadic templates can come later.

PR #1042, opened 2017-10-08
ENGINES: Improve output of unknown game variant detection

csnover:
When a user tries to add a game expecting it to be a particular game for a
particular engine, but a detector from another engine happens to match some
files that exist in the game directory and reports on those files instead,
this can cause a lot of confusion because the detector doesn't say what
engine or game it thought it matched.

This PR makes the following changes:

1. Adds the name of the matching engine as well as any matching game IDs
   (if applicable) to the detector's logged output, and more specific guidance
   about where to send the detection information (to the bug tracker);
2. Wraps the first part of the report to 80 columns, taking into account the
   dynamic formatting of the string, instead of statically placing newlines
   in ways that doesn't actually work for wrapping dynamic content. To avoid
   anyone else needing to reimplement this kind of primitive word wrapping of
   monospaced text, this functionality was added to Common::String, though it
   could be moved to free function or just kept within the advanced detector if
   it is not desired for it to be part of Common::String;
3. Fixes a buggy-looking behaviour of the detector, where it stopped looking
   for missing files if a hash/size did not match even though there could be
   more missing files in the future. This was most evident when listing out the
   detected game IDs in SCI engine against a synthetic test directory
   containing non-matching files, where every game entry with a RESOURCE.MAP was
   getting matched even though other resource files for those games didn't exist
   in the test directory.

Refs Trac#10272.

### DIFF
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -466,6 +466,7 @@ ADGameDescList AdvancedMetaEngine::detectGame(const Common::FSNode &parent, cons
 
 		bool allFilesPresent = true;
 		int curFilesMatched = 0;
+		bool hashOrSizeMismatch = false;
 
 		// Try to match all files for this game
 		for (fileDesc = g->filesDescriptions; fileDesc->fileName; fileDesc++) {
@@ -477,16 +478,21 @@ ADGameDescList AdvancedMetaEngine::detectGame(const Common::FSNode &parent, cons
 				break;
 			}
 
+			if (hashOrSizeMismatch)
+				continue;
+
 			if (fileDesc->md5 != NULL && fileDesc->md5 != filesProps[tstr].md5) {
 				debug(3, "MD5 Mismatch. Skipping (%s) (%s)", fileDesc->md5, filesProps[tstr].md5.c_str());
 				fileMissing = true;
-				break;
+				hashOrSizeMismatch = true;
+				continue;
 			}
 
 			if (fileDesc->fileSize != -1 && fileDesc->fileSize != filesProps[tstr].size) {
 				debug(3, "Size Mismatch. Skipping");
 				fileMissing = true;
-				break;
+				hashOrSizeMismatch = true;
+				continue;
 			}
 
 			debug(3, "Matched file: %s", tstr.c_str());
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -325,17 +325,30 @@ Common::Error AdvancedMetaEngine::createInstance(OSystem *syst, Engine **engine)
 		return Common::kNoError;
 }
 
-void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFilePropertiesMap &filesProps) const {
-	// TODO: This message should be cleaned up / made more specific.
-	// For example, we should specify at least which engine triggered this.
-	//
-	// Might also be helpful to display the full path (for when this is used
-	// from the mass detector).
+void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFilePropertiesMap &filesProps, const ADGameIdList &matchedGameIds) const {
 	Common::String report = Common::String::format(
-			_("The game in '%s' seems to be unknown.\n"
-			  "Please, report the following data to the ScummVM team along with name\n"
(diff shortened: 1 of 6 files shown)


engines/advancedDetector.cpp
```diff
@@ -325,17 +325,27 @@ Common::Error AdvancedMetaEngine::createInstance(OSystem *syst, Engine **engine)
...
 	Common::String report = Common::String::format(
-			_("The game in '%s' seems to be unknown.\n"
-			  "Please, report the following data to the ScummVM team along with name\n"
-			  "of the game you tried to add and its version, language, etc.:"), path.getPath().c_str()) + "\n";
-	report += "\n";
+			_("The game in '%s' seems to be an unknown %s engine game variant.\n\n"
+			  "Please report the following data to the ScummVM team at "
+			  "https://bugs.scummvm.org along with the name of the game "
+			  "you tried to add and its version, language, etc.:"), path.getPath().c_str(), getName(), "https://bugs.scummvm.org/");
+
```
criezy: It looks like you half intended to use a %s for the bug tracker path?

Also it seem strange to translate this part of the message, but not the part added to it below (the `"Matched game IDs:"` bit).

criezy:
This seems like a useful change.

Note that we don't handle well non-ASCII characters in `logMessage()`, but we also discussed making the unknown variant message more visible by using a GUI dialog, which incidentally would result in translations being correctly displayed. And since we have a `kFeatureOpenUrl` feature in `OSystem`, we could even have a button in the dialog to open the bug tracker on ports where this feature is supported. I was also thinking of adding copyTextToClipboard in `OSystem` (currently we only have a way to get text from clipboard) so that the relevant detection message could be automatically copied.

csnover:
> we also discussed making the unknown variant message more visible by using a GUI dialog, which incidentally would result in translations being correctly displayed.

That seems like a good idea in the case of a single unknown variant detection, once we can easily get the information out of the GUI (like by a `copyTextToClipboard` API).

Things get a little more complicated to me when it’s a situation where multiple engines detect in the same directory, like with Trac#10272 where MGDX is detected and Playtoons1 is also detected. To me, the ideal would be to show two detection entries and then if the unknown variant is selected to show the extra information as if a single unknown variant was detected. But maybe that’s just too much effort for this for now, and we should just focus on improving things incrementally at the moment.

> And since we have a kFeatureOpenUrl feature in OSystem, we could even have a button in the dialog to open the bug tracker on ports where this feature is supported.

Automation definitely crossed my mind here. It is something where I am not entirely sure if there is a good reason to require *any* manual action, other than an opt-in for privacy reasons, except that it requires more work/infrastructure to fully automate. (Ideally there would just be some system that sends the report along with an optional email address to mail back when it’s addressed, but this… idealistic at this moment.)

FWIW, it *is* possible to prepopulate the Trac newticket page just by passing fields in the query string, but the form is only available to logged in users. Doing that would reduce friction and eliminate the need to put anything into the clipboard (though of course that ought to be a clipboard fallback option for systems without kFeatureOpenUrl support regardless), but only seems feasible if we can either use a separate page that redirects with a valid login or improve the 403 message that gets displayed currently to be actually helpful. Since Trac already memorises the return URL if you go to log in from the newticket page, even if a user was not logged in, we would still get the form auto-population when they returned from login/registration, so it would be nice to make this work.

>[…]automatically copied.

To be clear: automatically copied *after* clicking a button that says something like “copy report to clipboard”?

I’ll take care of that other… oversight… with the double URLs shortly. :)

criezy:
> To be clear: automatically copied after clicking a button that says something like “copy report to clipboard”?

Yes, I think that is what I had in mind. Automatically copying something to the clipboard without the user having to do anything would I think be doubly confusing: the user might not want to change the content of his clipboard, and he might not even know that it is happening. With a button he could do it only if he wants to, and it would be obvious to him it gets copied to the clipboard.

I think all this can go into a feature request on the tracker though (if there is not already one). And this PR can be merged as soon as you have fixed the double URL issue and either changed the`"Matched game IDs:"` bit to translate it, or changed the first part of the message to not translate it (having a half translated message would be a bit strange I think).

csnover:
OK, I rebased this and fixed up the missing translation and the missing string format parameter. I have some other things to tie up tonight and then I’ll land this tomorrow when I’m fresh.

csnover:
This is landed now. Thank you for your feedback!

PR #1043, opened 2017-10-08
SCUMM: Check for and flag demos correctly (PR#931 Fixup)

digitall:
Rewrote the detection matching to use the extra field like the other
two.

This requires extensive testing due to touching code shared for a lot of
games and being sensitive to individual versions of games.

### DIFF
--- a/engines/scumm/detection.cpp
+++ b/engines/scumm/detection.cpp
@@ -489,42 +489,45 @@ static void computeGameSettingsFromMD5(const Common::FSList &fslist, const GameF
 
 	// Compute the precise game settings using gameVariantsTable.
 	for (const GameSettings *g = gameVariantsTable; g->gameid; ++g) {
-		if (g->gameid[0] == 0 || !scumm_stricmp(md5Entry->gameid, g->gameid)) {
-			// The gameid either matches, or is empty. The latter indicates
-			// a generic entry, currently used for some generic HE settings.
-			if (g->variant == 0 || !scumm_stricmp(md5Entry->variant, g->variant)) {
-				// Perfect match found, use it and stop the loop
-				dr.game = *g;
-				dr.game.gameid = md5Entry->gameid;
+		if (g->gameid[0] != 0 && scumm_stricmp(md5Entry->gameid, g->gameid)) continue;
+		// The gameid either matches, or is empty. The latter indicates
+		// a generic entry, currently used for some generic HE settings.
 
-				// Set the platform value. The value from the MD5 record has
-				// highest priority; if missing (i.e. set to unknown) we try
-				// to use that from the filename pattern record instead.
-				if (md5Entry->platform != Common::kPlatformUnknown) {
-					dr.game.platform = md5Entry->platform;
-				} else if (gfp->platform != Common::kPlatformUnknown) {
-					dr.game.platform = gfp->platform;
-				}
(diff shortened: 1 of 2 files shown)


engines/scumm/detection.cpp
```diff
@@ -489,42 +489,45 @@ static void computeGameSettingsFromMD5(const Common::FSList &fslist, const GameF
...
-				if (md5Entry->platform != Common::kPlatformUnknown) {
-					dr.game.platform = md5Entry->platform;
-				} else if (gfp->platform != Common::kPlatformUnknown) {
-					dr.game.platform = gfp->platform;
-				}
+		if (g->gameid[0] != 0 && scumm_stricmp(md5Entry->gameid, g->gameid)) continue;
+		// The gameid either matches, or is empty. The latter indicates
+		// a generic entry, currently used for some generic HE settings.
+
+		if (g->variant != 0 && !!scumm_stricmp(md5Entry->variant, g->variant)) continue;
```
bluegr: The two not operators here (!!) are superfluous, and unneeded

engines/scumm/detection.cpp
```diff
@@ -489,42 +489,45 @@ static void computeGameSettingsFromMD5(const Common::FSList &fslist, const GameF
...
+		// Variant matches too
+
+		if (g->preferredTag !=0 && !!scumm_stricmp(md5Entry->extra, g->preferredTag)) continue;
```
bluegr: Same here, with !!

engines/scumm/detection.cpp
```diff
@@ -489,42 +489,45 @@ static void computeGameSettingsFromMD5(const Common::FSList &fslist, const GameF
...
+		//OH, and do check the extra field in the entry too, otherwise we miss demos
```
bluegr: Nitpicking: this comment needs formatting to be the same as the rest

engines/scumm/detection.cpp
bluegr: Also, this line is the only change to the detector code in this PR :)

digitall:
@bluegr : I have addressed all your comments and rebased the commit in this PR onto the latest master commit.

One point, is that the "!!" is not just spurious, but seems to be a practice to avoid compiler warnings and sanitize the value to a boolean. See:
http://stupefydeveloper.blogspot.co.uk/2009/09/c-double-exclamation.html

I am not sure, but I think this is not necessary here unless @henke37 did it for some specific reason?

bluegr:
Thanks!

Yeah, the usage of !! does help sanitize a return value to a boolean, but it's not really needed here, as the condition above these does not use it.

Have you tried these changes? Are all the SCUMM games detected properly?

As mentioned, the only real change to the detector is the additional check for the preferredTag field

digitall:
@bluegr No problem. No, I haven't tried these changes yet, beyond checking they compile without warnings.

I can try moving my .scummvmrc file off, and doing a Mass Add with and without this patch to see what difference it makes, but I don't think I have most of the SCUMM game demos so this may be pointless. I suggest that a SCUMM engine dev who has all of the relevant demos tests this and sees if the detection is better as they are better placed to know if this is correct. Not sure who that should be? @sev- ?

sev-:
OK, merging this. Will do some detection testing after the merge.

PR #1044, opened 2017-10-19
IOS7: Updated app icons and splash screens to be more contemporary and clean

maikudou:
Default icons and splash screens looks alien and outdated on current iOS devices


janisozaur:
The submitted images are not optimised at all.

wjp:
More importantly, this changes our logo itself, which will need discussion, and more motivation than that they look alien and outdated on iOS devices.

PR #1046, opened 2017-10-26
SAGA: Add Code to Support French Fan Translation.

digitall:
This code was supplied by Darknior (hugues.fabien@gmail.com).

### DIFF
--- a/engines/saga/introproc_ite.cpp
+++ b/engines/saga/introproc_ite.cpp
@@ -128,6 +128,8 @@ EventColumns *Scene::queueIntroDialogue(EventColumns *eventColumns, int n_dialog
 		textEntry.rect.top = INTRO_DE_CAPTION_Y;
 	} else if (_vm->getLanguage() == Common::IT_ITA) {
 		textEntry.rect.top = INTRO_IT_CAPTION_Y;
+	} else if (_vm->getLanguage() == Common::FR_FRA) {
+		textEntry.rect.top = INTRO_IT_CAPTION_Y;
 	} else {
 		textEntry.rect.top = INTRO_CAPTION_Y;
 	}
@@ -379,6 +381,8 @@ int Scene::ITEIntroCaveCommonProc(int param, int caveScene) {
 		lang = 1;
 	else if (_vm->getLanguage() == Common::IT_ITA)
 		lang = 2;
+	else if (_vm->getLanguage() == Common::FR_FRA)
+		lang = 3;
 
 	int n_dialogues = 0;
 
--- a/engines/saga/introproc_ite.cpp
+++ b/engines/saga/introproc_ite.cpp
@@ -42,6 +42,7 @@ namespace Saga {
 #define INTRO_CAPTION_Y 170
 #define INTRO_DE_CAPTION_Y 160
 #define INTRO_IT_CAPTION_Y 160
+#define INTRO_FR_CAPTION_Y 160
 #define INTRO_VOICE_PAD 50
 #define INTRO_VOICE_LETTERLEN 90
 
@@ -129,7 +130,7 @@ EventColumns *Scene::queueIntroDialogue(EventColumns *eventColumns, int n_dialog
 	} else if (_vm->getLanguage() == Common::IT_ITA) {
 		textEntry.rect.top = INTRO_IT_CAPTION_Y;
 	} else if (_vm->getLanguage() == Common::FR_FRA) {
-		textEntry.rect.top = INTRO_IT_CAPTION_Y;
+		textEntry.rect.top = INTRO_FR_CAPTION_Y;
 	} else {
 		textEntry.rect.top = INTRO_CAPTION_Y;
 	}
--- a/engines/saga/itedata.cpp
+++ b/engines/saga/itedata.cpp
@@ -446,6 +446,27 @@ const char *ITEinterfaceTextStrings[][53] = {
 		NULL,
 		NULL,
 		"       \xC7\xE0\xE3\xF0\xF3\xE7\xEA\xE0     "
+	},
+	// French fan translation : ACTIONS
+	{
+		"Aller vers", "Examiner", "Prendre", "Parler \205", "Ouvrir",
+		"Fermer", "Utiliser", "Donner", "Options", "Test",
+		"Demo", "Aide", "Quitter le Jeu", "Vite", "Lent",
+		"On", "Off", "Continuer \205 jouer", "Charger", "Sauver",
+		"Options du Jeu", "Vitesse texte", "Musique", "Son", "Annuler",
+		"Quitter", "OK", "Mid", "Click", "10%",
+		"20%", "30%", "40%", "50%", "60%",
+		"70%", "80%", "90%", "Max", "Quitter le jeu?",
+		"Chargement OK!", "Nommer la sauvegarde", "Donner %s \205 %s", "Utiliser %s avec %s",
+		"[Nouvelle sauvegarde]",
+		"Je ne peux pas prendre ça.",
+		"Je ne vois rien de sp\202cial.",
+		"Il n'y a pas la place pour l'ouvrir.",
+		"Il n'y a pas d'ouverture \205 fermer.",
+		"Je ne sais pas comment l'utiliser.",
+		"Sous-titre",
+		"Que r\202pond Rif?",
+		"Charger une sauvegarde"
 	}
 };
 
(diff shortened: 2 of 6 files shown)


digitall:
I made two small whitespace changes from the files supplied to remove trailing tabs on following lines, otherwise these are as supplied.

Darknior:
Thanks a lot it will save my time ;)
I will try to do it fine next time if i have an other patch to add ScummVM.

engines/saga/itedata.cpp
```diff
@@ -860,45 +999,57 @@ const IntroCredit creditsTreeHouse2[] = {
...
 const IntroCredit creditsFairePath1[] = {
 	{Common::EN_ANY, kITECreditsAny, kITECreditsHeader, "Programming"},
 	{Common::DE_DEU, kITECreditsAny, kITECreditsHeader, "Programmiert von"},
 	{Common::IT_ITA, kITECreditsAny, kITECreditsHeader, "Programmazione"},
+	{Common::FR_FRA, kITECreditsAny, kITECreditsHeader, "Programmeurs"},
 	{Common::UNK_LANG, kITECreditsAny, kITECreditsText, "Talin, Walter Hochbrueckner,"},
 	{Common::EN_ANY, kITECreditsAny, kITECreditsText, "Joe Burks and Robert Wiggins"},
 	{Common::DE_DEU, kITECreditsAny, kITECreditsText, "Joe Burks und Robert Wiggins"},
-	{Common::IT_ITA, kITECreditsAny, kITECreditsText, "Joe Burks e Robert Wiggins"},
+	{Common::IT_ITA, kITECreditsAny, kITECreditsText, "Joe Burks et Robert Wiggins"},
```
Darknior: oups you must switch the e / et please
IT_ITA : Joe Burks e Robert Wiggins
FR_FRA : Joe Burks et Robert Wiggins
Thanks

digitall: Have corrected and force-pushed a revised version.

engines/saga/itedata.cpp
```diff
@@ -482,9 +504,15 @@ const char *pieceNames[][PUZZLE_PIECES] = {
 		"cacciavite", "pinze", "morsa", "morsa da legno", "livella",
 		"spago", "pialla", "martello", "metro a nastro", "accetta",
 		"cesoie", "righello", "sega", "mazza", "pennello"
+	},
+	{
+		"tournevis", "pinces", "\202taux \205 vis", "\202taux en bois", "niveau",
```
bgK: `pinces` and `étaux` are plural, the other languages seem to use singular.

engines/saga/itedata.cpp
```diff
@@ -590,6 +643,12 @@ const char *optionsStr[][4] = {
 		"\"Si, grazie. Ne avrei bisogno.\"",
 		"\"No, grazie, voglio provare a risolverlo da solo.\"",
 		"Penso che la tessera %s sia nel posto sbagliato."
+	},
+	{
+		"\"Je r\202soudrais cette \202nigme plus tard.\"",
```
bgK: `résoudrais` is conditional tense, the other languages seem to use future.

engines/saga/itedata.cpp
Darknior: Yes it is an error from the original translation i don't see :(
It is better singular.

engines/saga/itedata.cpp
Darknior: I think it's good because it say he will do it later, it is future.
Or now he want help to resolve now.

engines/saga/itedata.cpp
Darknior: "tournevis", "pince", "\202tau \205 vis", "pince \205 bois", "niveau",

engines/saga/itedata.cpp
bgK: Ok, then `résoudrai` is the proper future.

bluegr:
The changes to the code look to be correct. Nice work! :)

I can't comment on the translation, but once everything is reviewed by our French-speaking members of the team, it looks OK to be merged.

digitall:
@bgK : Have made the changes you and @Darknior discussed. OK now?

engines/saga/introproc_ite.cpp
```diff
@@ -128,6 +128,8 @@ EventColumns *Scene::queueIntroDialogue(EventColumns *eventColumns, int n_dialog
 		textEntry.rect.top = INTRO_DE_CAPTION_Y;
 	} else if (_vm->getLanguage() == Common::IT_ITA) {
 		textEntry.rect.top = INTRO_IT_CAPTION_Y;
+	} else if (_vm->getLanguage() == Common::FR_FRA) {
+		textEntry.rect.top = INTRO_IT_CAPTION_Y;
```
criezy: Wouldn't it be cleaner to add a INTRO_FR_CAPTION_Y define?

engines/saga/itedata.cpp
```diff
@@ -569,6 +615,13 @@ const char *whineStr[][NUM_WHINES] = {
 		"Sigh...",
 		"Credo che questa faccenda ti abbia dato alla testa, Sakka!",
 		"Beh, non ricordo regole specifiche contro i suggerimenti."
+	},
+	{
+		"Rohh, allez Sakka!",
+		"Un indice ne peut pas faire de mal.",
+		"Pfff...",
+		"Je crois que ton carnet te monte \205 la tête, Sakka!",
```
criezy: Should that be 't\210te' rather than 'tête'?

engines/saga/itedata.cpp
```diff
@@ -590,6 +643,12 @@ const char *optionsStr[][4] = {
 		"\"Si, grazie. Ne avrei bisogno.\"",
 		"\"No, grazie, voglio provare a risolverlo da solo.\"",
 		"Penso che la tessera %s sia nel posto sbagliato."
+	},
+	{
+		"\"Je r\202soudrai cette \202nigme plus tard.\"",
+		"\"Oui, j'aimerais un indice s'il vous plait.\"",
+		"\"Non merci, je voudrais r\202soudre cela par moi même.\"",
```
criezy: Should that be 'm\210me' rather than 'même'?

engines/saga/itedata.cpp
```diff
@@ -545,6 +586,11 @@ const char *sakkaStr[][NUM_SAKKA] = {
 		"Hey, non si dovrebbero aiutare i candidati!",
 		"Ragazzi! Questo dovrebbe essere un test!",
 		"Forza ragazzi, non si pu\225!"
+	},
+	{
+		"H\202, vous n'êtes pas suppos\202s aider les concurrents!",
```
criezy: Shouldn't that be '\210tes' rather than 'êtes'?

criezy:
I am wondering if spaces should be added before the ! and ? in the French text. Usually you would need a thin non-breakable space, which you might not have in the font used, but maybe a normal space is better than no space at all? But that is questionable and it could be left as it is now. In France we would use a space instead of the thin non-breakable space, but wikipedia indicates that in Canada it would simply be omitted - see https://fr.wikipedia.org/wiki/Espace_fine_ins%C3%A9cable (in French).

engines/saga/itedata.cpp
Darknior: Yes you're right i forgot this one :( Thanks

engines/saga/itedata.cpp
Darknior: yes same, i forgot to replace the two ê :(

engines/saga/itedata.cpp
Darknior: same

Darknior:
If you want you can add space between ! and ? ... it not choking me ;)

engines/saga/introproc_ite.cpp
Darknior: #define INTRO_DE_CAPTION_Y 160
#define INTRO_IT_CAPTION_Y 160

DE and IT are already same, and it works perfectly like that in French.
But if you think it's better, you can also add a FR one with same setting.

#define INTRO_FR_CAPTION_Y 160

bgK:
@Darknior, @digitall thanks! Everything looks fine on the grammar side. I'll let the typography to criezy's capable hands.

digitall:
@criezy : Have made the changes which you suggested, apart from adding the spaces between the !/? characters as I wasn't sure about that...

Darknior:
For me it's good like that, the spaces between the !/? characters is not really fine.

criezy:
I don't feel strongly either way regarding the spaces before exclamation and question marks. And since all other issues have been fixed I will merge this now.

PR #1047, opened 2017-10-31
SCUMM: Prevent false detection of Steam games

BenCastricum:
Ignore steam files of which we don't know the MD5. Currently this causes a [crash (assert)](https://bugs.scummvm.org/ticket/10290) on trying to read the (non-existing) index file from the executable. Unknown steam variants can't be handled anyway because we need to hardcode the offset to the game index file in ScummVM.


sev-:
Good catch!

PR #1048, opened 2017-11-05
Add detection checksum for indy3 FM-TOWNS German

h3xx:
Source: Emuparadise.


raziel-:
taken from here: https://www.emuparadise.me/dmca.php

quote:
By downloading or emulating ANY titles from this website you are agreeing to the following:

You are responsible for checking your local laws regarding the use of this site
You are authorized to / own a license to use any file you download or emulate
You will not distribute any of these files without the appropriate permissions
These files are hosted for archival purposes on a best effort basis and we do not guarantee any level of quality as far as emulation is concerned.

Read more at https://www.emuparadise.me/dmca.php#BGLIVSehOWaGk4fU.99
/quote

That does sound fishy

rootfather:
I don't think we can accept this until we get a valid source (and not some pirated stuff from a random website).

h3xx:
Closing until I can verify legitimacy.

PR #1049, opened 2017-11-08
LURE: Stop taking address of unaligned pointer

csnover:
While usage of these pointers was technically safe because they
were read through an alignment-aware API, taking the address of an
unaligned pointer was generating warnings in Clang, and is not
strictly necessary here. This change solves the warning and also
protects this code from any future change that might cause it to
start reading unsafely.

I think this change should be safe and not cause any trouble, but
I do not have this game so cannot test it myself to be sure. So
if someone else can either test it or look and make sure I did
not do something stupid, I would appreciate it.

### DIFF
--- a/engines/lure/res.cpp
+++ b/engines/lure/res.cpp
@@ -87,7 +87,7 @@ void Resources::freeData() {
 }
 
 struct AnimRecordTemp {
-	uint16 *offset;
+	uint16 offset;
 	MovementDataList *list;
 };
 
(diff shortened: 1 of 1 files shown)


digitall:
@csnover : Lure of the Temptress was released as freeware and can be downloaded from our downloads page, so you can test with this: http://www.scummvm.org/games/#lure

csnover:
This would actually also turn out OK even without the use of the alignment-safe APIs since all the preceding fields are short-sized, so this really is mostly just about silencing the warning from Clang since it’s pretty unlikely someone will come around later and decide to stick a char-sized field at the start of the struct or change the memory allocator to create oddly-aligned memory.

PR #1050, opened 2017-11-09
SDL: fix OSD message artifacts

Risca:
On the OpenPandora handheld, the OSD message would not render unless you moved the cursor in the area where it was supposed to show. Additionally, the OSD message was not transparent like in v1.8.

This fixes both these issues, but to be honest, I not exactly sure why.

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -2293,6 +2293,11 @@ void SurfaceSdlGraphicsManager::displayMessageOnOSD(const char *msg) {
 	// Init the OSD display parameters, and the fade out
 	_osdMessageAlpha = SDL_ALPHA_TRANSPARENT + kOSDInitialAlpha * (SDL_ALPHA_OPAQUE - SDL_ALPHA_TRANSPARENT) / 100;
 	_osdMessageFadeStartTime = SDL_GetTicks() + kOSDFadeOutDelay;
+	// Enable alpha blending
+	SDL_SetAlpha(_osdMessageSurface, SDL_RLEACCEL | SDL_SRCALPHA, _osdMessageAlpha);
+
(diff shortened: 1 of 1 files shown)


criezy:
Indeed the missing `_forceRedraw = true;` and `SDL_SetAlpha` in `SurfaceSdlGraphicsManager::displayMessageOnOSD` are bugs.

I think we did not notice the first one before because most of the OSD messages we currently have on desktop platforms result from a change that would redraw the screen anyway (for example toggling fullscreen or aspect ratio correction on/off). But I did reproduce the bug on macOS with a message displayed from a piece of code that does not otherwise request a screen update.

And indeed with the SurfaceSDL backend OSD messages are fully opaque and do not respect the `kOSDInitialAlpha` and your commit fixes that.

I do have one question before merging though: why did you remove the `SDL_SRCCOLORKEY`. With SDL2 this flag is not even passed to SDL so it doesn't matter, and with SDL 1.2 it seems to work on macOS wether the flag is here or not. I find the SDL 1.2 documentation confusing in that respect, so I don't know if removing that flag is correct or not.

Risca:
I removed the SDL_SRCCOLORKEY for two reasons:
* the _osdMessageSurface is no longer created with that flag
* SDL_SetAlpha() doesn't care about it

From the documentation:
"flags is used to specify whether alpha blending should be used (SDL_SRCALPHA) and whether the surface should use RLE acceleration for blitting (SDL_RLEACCEL). flags can be an OR'd combination of these two options, one of these options or 0."

https://www.libsdl.org/release/SDL-1.2.15/docs/html/sdlsetalpha.html

criezy:
That documentation is pretty clear; I don't know why I was confused by it. I will blame my imaginary pet zombie who ate my imaginary brain yesterday.

Thank you for your contribution. I will merge this now.
I did notice some issues with SDL2 with the transparent OSD message, but I will fix those later tonight.

PR #1051, opened 2017-11-11
BASE: Remove bad casts between incompatible Plugin types

csnover:
Previously, a C-style cast was used to convert a
Common::Array<Plugin *>, populated with pointers to StaticPlugin
and DynamicPlugin instances, to a
Common::Array<PluginSubclass<T> *>, but PluginSubclass<T> is a
*sibling* class to StaticPlugin/DynamicPlugin, so this cast was
invalid and the results undefined. The methods for retrieving
subclasses of plugins can't be easily changed to just generate an
array of temporary wrapper objects that expose an identical API
which dereferences to the preferred PluginObject subclass because
pointers to these objects are retained by other parts of ScummVM,
so the wrappers would needed to be persisted or they would need to
just re-expose the underlying Plugin object again. This indicated
that a way to solve this problem is to have the callers receive
Plugin objects and get the PluginObject from the Plugin by
explicitly stating their desired type, in a similar manner to
std::get(std::variant), so that the pattern used by this patch to
solve the problem.

An alternative approach might be to do something like subclass
Common::Array to return appropriate objects, but this again seems
to run into the problem with needing some kind of extra wrapper,
more allocations, and additional changes to the callers to use the
wrapper correctly.

I am interested in learning if anyone else has some suggested
alternative approaches that would work better here.

### DIFF
--- a/audio/mididrv.cpp
+++ b/audio/mididrv.cpp
@@ -92,9 +92,9 @@ MusicType MidiDriver::getMusicType(MidiDriver::DeviceHandle handle) {
 		return MT_MT32;
 
 	if (handle) {
-		const MusicPlugin::List p = MusicMan.getPlugins();
-		for (MusicPlugin::List::const_iterator m = p.begin(); m != p.end(); m++) {
-			MusicDevices i = (**m)->getDevices();
+		const PluginList p = MusicMan.getPlugins();
+		for (PluginList::const_iterator m = p.begin(); m != p.end(); m++) {
+			MusicDevices i = (*m)->get<MusicPluginObject>().getDevices();
 			for (MusicDevices::iterator d = i.begin(); d != i.end(); d++) {
 				if (handle == d->getHandle())
 					return d->getMusicType();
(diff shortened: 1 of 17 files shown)


csnover:
If nobody knows of a better way to handle this, I will aim to land this on Sunday.

wjp:
I haven't looked at the full PR in detail, but the core changes make sense. (And I see no real benefits in e.g. moving the variant get functionality up a level to the container.)

I've tested both static and dynamic engine plugins with this PR on 64 bit linux without any problems.

csnover:
Landed. Thank you for your review and testing!

rsn8887:
These changes cause the PSP port to crash when scumm engine is enabled and the user uses the launcher to
a) select "add game" -> "choose", regardless of directory it crashes as soon as "choose" is selected
or
b) selecting "start" (crashes immediately after pressing start, on any game)
or
c) selection "load game" (crashes immediately)

Reverting this commit fixes all the crashes in the PSP port.

Has this been tested on any other platforms that use plugins? The problem might not be limited to PSP.

PR #1054, opened 2017-11-12
SCUMM: Fix race condition in MOD player

csnover:
Fixes Trac#6272.


csnover:
Thank you for your review! Landed in 0eda63bed1b2d7fe0dba1360cad880b826248eaa.

PR #1055, opened 2017-11-13
COMMON: Add support for arrays & limited custom deleters to smart pointers

csnover:
This enables ScopedPtr to be used for all cases where raw pointers
are currently used to manage owned memory (assuming C malloc APIs
are replaced with C++ APIs), and brings ScopedPtr closer to
std::unique_ptr in terms of functionality. DisposablePtr also comes
along for the ride, for the moment.

Custom deleters of ScopedPtr are not currently fully conforming to
std::unique_ptr for the sake of simplicity of implementation.
Unlike in the standard library, plain functions and lvalue
references are not supported, nor may custom deleters be passed
to the constructor at runtime. This can be improved in the future,
if necessary, by creating a Pair class that uses the Empty Base
Optimization idiom to avoid extra storage overhead of the deleter
instance when it is not needed, as in typical standard library
implementations, plus some additional type traits to support the
necessary metaprogramming for the different type overloads.

Use of array types in SharedPtr also do not conform to C++17
std::shared_ptr, which checks for convertability between the
argument and an array type instead of having an explicit array
specialization. This should be an OK compromise for now, and
also might even be preferable (earlier C++ standards did not have
any explicit handling of array types in std::shared_ptr, so now
the C++ standard library needs to maintain backwards-compatibility
where we do not).

This patch unfortunately introduces some code duplication due to
the fact that C++98 does not allow default arguments in function
template parameters. Using C++11, there would be no code
duplication.

### DIFF
--- a/common/ptr.h
+++ b/common/ptr.h
@@ -26,6 +26,7 @@
 #include "common/scummsys.h"
 #include "common/noncopyable.h"
 #include "common/safe-bool.h"
+#include "common/type-traits.h"
 #include "common/types.h"
 
 namespace Common {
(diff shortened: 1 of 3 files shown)


csnover:
FWIW, I think there is also a reasonable argument to be made that `Common::Array` should just always be used for owning arrays, in which case the array specialisations could be removed again which would save most of the duplication (except for `DisposablePtr`, which is really just a duplication of `ScopedPtr` with a special hard-coded non-deleting deleter…). Interested in hearing others’ thoughts in that regard as well.

csnover:
After trying to use this to fix Full Pipe memory ownership/leaks, I ended up always running into the problem that, without C++11 move semantics, there is very limited application for this enhancement since these things can’t live in containers.

`DefaultDeleter` was helpful to combine with `Common::for_each` to get rid of a bit of boilerplate when I had to fall back to using raw pointers, but that’s no justification for the rest of this PR. The couple cases of `new[]` were served appropriately with the new `Common::Array(n)` constructor too. So, never mind about this for now.

PR #1056, opened 2017-11-15
COMPOSER: Remove leading dots from file paths read from book.ini

bgK:
book.ini in the French version of Gregory has file paths with this format: '../data/page99.rsc'.

According to the data structure described in bug #6548, removing the leading '../' should allow the engine to find the resource file where it is located, in the data folder.

I can't test this fix because I don't have the game. I would be glad if someone could test it or at least verify it does not break the other games.


sev-:
Actually, the game works well for me without this patch. With the patch the game asserts on:

```failed to open 'page99.rsc'!```

sev-:
...and the original bugreport is incorrect.

PR #1058, opened 2017-11-18
NEVERHOOD: fix musical hut crash in DR, bug #7085

w23:
Patched size was wrong too.

[https://bugs.scummvm.org/ticket/7085](https://bugs.scummvm.org/ticket/7085)


sev-:
Thanks!

PR #1061, opened 2017-11-18
FULLPIPE: Fix various memory issues in Full Pipe

csnover:
This PR contains changes necessary to stop Full Pipe from
leaking memory regularly, fixes the broken RTL, and fixes the
broken in-game save/load screens (mostly; I don't know how to
find a hyphen glyph for drawing the hyphens in the dates so
they are drawn with dot separators instead).

I did not run through the entire game with these changes, just
the first few areas which are immediately accessible without
solving any puzzles. Because these memory leaks were due to
systemic design issues around memory ownership rather than
isolated instances, I expect that there are additional leaks
in other parts of the engine which are not fixed. These
additional leaks can be tracked using separate tickets as they
are discovered.

Fixes Trac#10316, Trac#9654, Trac#9657.

### DIFF
--- a/common/ptr.h
+++ b/common/ptr.h
@@ -40,12 +40,7 @@ class SharedPtrDeletionImpl : public SharedPtrDeletionInternal {
 public:
 	SharedPtrDeletionImpl(T *ptr) : _ptr(ptr) {}
 	~SharedPtrDeletionImpl() {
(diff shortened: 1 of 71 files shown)


sev-:
I checked the game, everything works well, though there are still few leaks.

PR #1063, opened 2017-11-19
SDL2: Improve handling of keyboard repeat events

bgK:
Unlike its predecessor, SDL2 always generates keydown repeat events. ScummVM was not updated to account for it. As a result, key repeat events received by the engines are not properly flagged.

This pull request makes sure the repeat events are flagged and disables ScummVM's own repeat event generation when using SDL2. This allows to use the operating system's configured timings for key repeats instead of ScummVM's hardcoded values.

### DIFF
--- a/backends/events/default/default-events.cpp
+++ b/backends/events/default/default-events.cpp
@@ -40,7 +40,8 @@ DefaultEventManager::DefaultEventManager(Common::EventSource *boss) :
 	_modifierState(0),
 	_shouldQuit(false),
 	_shouldRTL(false),
-	_confirmExitDialogActive(false) {
+	_confirmExitDialogActive(false),
+	_shouldGenerateKeyRepeatEvents(true) {
 
 	assert(boss);
 
@@ -84,7 +85,9 @@ void DefaultEventManager::init() {
 bool DefaultEventManager::pollEvent(Common::Event &event) {
 	_dispatcher.dispatch();
 
-	handleKeyRepeat();
+	if (_shouldGenerateKeyRepeatEvents) {
+		handleKeyRepeat();
+	}
 
 	if (_eventQueue.empty()) {
 		return false;
--- a/backends/events/default/default-events.cpp
+++ b/backends/events/default/default-events.cpp
@@ -247,7 +247,7 @@ void DefaultEventManager::handleKeyRepeat() {
 		}
 	} else {
 		// Check if event should be sent again (keydown)
-		if (_currentKeyDown.keycode != Common::KEYCODE_INVALID && _keyRepeatTime < time) {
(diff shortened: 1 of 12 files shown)


backends/events/default/default-events.cpp
```diff
@@ -86,163 +83,181 @@ void DefaultEventManager::init() {
...
-
-			break;
 
 		default:
 			break;
 		}
 	} else {
 		// Check if event should be sent again (keydown)
-		if (_currentKeyDown.keycode != 0 && _keyRepeatTime < time) {
+		if (_currentKeyDown.keycode != Common::KEYCODE_INVALID && _keyRepeatTime < time) {
```
csnover: I think there is (and has always been) an off-by-one in this time check, which should be `<=` or else this requires advancing 1ms past the next repeat time to repeat.

common/events.h
```diff
@@ -97,10 +97,12 @@ typedef uint32 CustomEventType;
 struct Event {
 	/** The type of the event. */
 	EventType type;
-	/** Flag to indicate if the event is real or synthetic. E.g. keyboard
-	  * repeat events are synthetic.
-	  */
-	bool synthetic;
+	/**
```
csnover: Since this header is being modified anyway, would you mind fixing these doxygen comments in a separate commit to have appropriate whitespace between the members and at the start of each comment line? My inner perfectionist thanks you in advance.

backends/platform/sdl/sdl.cpp
```diff
@@ -207,6 +208,16 @@ void OSystem_SDL::initBackend() {
 	if (_eventSource == 0)
 		_eventSource = new SdlEventSource();
 
+	if (_eventManager == nullptr) {
+		DefaultEventManager *eventManager = new DefaultEventManager(_eventSource);
+#if SDL_VERSION_ATLEAST(2, 0, 0)
```
csnover: Is this going to cause problems for (possibly hypothetical at this point) backends that use SDL2 events but don’t use OSystem_SDL? Maybe this should be an `#if/else` on the member initialiser in the DefaultEventManager constructor instead to avoid such an issue.

backends/events/default/default-events.cpp
bgK: Thanks, fixed

common/events.h
bgK: My OCD now feels better as well

backends/platform/sdl/sdl.cpp
bgK: I'm reluctant to introduce platform specific code in `DefaultEventManager`. And introducing `Sdl2EventManager` just for this feels like over-engineering.

csnover: That seems like a fair balance.

PR #1065, opened 2017-11-21
TSAGE: Various fixes

csnover:
This is a small collection of fixes to TsAGE engine.
There will probably be at least one more, but this clears
three of the release blockers so I wanted to get it in
front of eyes as quickly as possible for review.


dreammaster:
Looks good to me.

PR #1069, opened 2017-11-26
NEVERHOOD: fix late game notes crash in DR

w23:
After player gets to the castle using a robot, there are no Willie's notes any more. Instead, there are notes from "Windows'" (in "Dyadyushka Risech" Russian version) on red paper. These notes require patching.


sev-:
Thanks!

PR #1070, opened 2017-11-26
ZVISION: Workaround for bug #6770

eriktorbjorn:
Since no one has been able to suggest a fix for any ScummVM bug responsible for bug #6770 ("ZVISION: Control lever still visible after opening Perils of Magic time tunnel"), here's my attempt at fixing it as a script bug in Zork: Grand Inquisitor.

NB: I've only tried the puzzle itself, and I've only tested with the English DVD version of the game. Oh, and I'm also not that familiar with the game engine itself. What could go wrong? :-)

### DIFF
--- a/engines/zvision/scripting/scr_file_handling.cpp
+++ b/engines/zvision/scripting/scr_file_handling.cpp
@@ -141,6 +141,24 @@ bool ScriptManager::parseCriteria(Common::SeekableReadStream &stream, Common::Li
 		criteriaList.back().push_back(entry);
 	}
 
+	// WORKAROUND for a script bug in Zork: Grand Inquisitor, room me2j
+	// (Closing the Time Tunnels). When the time tunnel is open the game
+	// shows a close-up of only the tunnel, instead of showing the entire
+	// booth. However, the scripts that draw the lever in its correct
+	// state do not test this flag, causing it to be drawn when it should
+	// not be. This bug does not happen in the original game, suggesting
+	// a ScummVM bug. But I'm not aware of any other such glitches, and it
+	// still seems like a correct way of working around bug #6770.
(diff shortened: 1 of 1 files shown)


bluegr:
Thanks for your work on this!

The workaround you made looks OK. I'm a bit worried about the fact that this issue only occurs in ScummVM, suggesting that something isn't implemented as it should be. However, since this is an isolated case, and it only causes a small graphics glitch, I think we can use the workaround for now.

+1 from me

RichieSams:
The patch looks ok to me.

dafioram:
This bug is present in the original version. I tested on gog (cd/windows). 

I also tested eriktorbjorn's PR on linux and it resolves the graphic issue for the gog/cd version of zgi.

eriktorbjorn:
So does anyone know if it will work for all languages/versions or not? None of the other workarounds needed language/version checks, as far as I can tell. (Other than to tell Nemesis and Grand Inquisitor apart, of course.)

RichieSams:
I can test with the German version of ZGI when I get home tonight

RichieSams:
Shucks. Turns out I have the German version of Zork Nemesis, not ZGI...

digitall:
@eriktorbjorn : Happy to merge now?

eriktorbjorn:
Well, there's nothing more I can do with it at any rate, so feel free to merge it if you think it looks ok.

digitall:
@eriktorbjorn : Well, the scope of the code change is very much limited to the ZVISION engine and ZGI and the code is quite self-contained, so I think it is fine to merge and fixup later as required. Merging...

bonki:
I can confirm that the fix also works in the German version of ZGI.

PR #1071, opened 2017-11-27
AUDIO: Update MAME OPL to latest upstream & add dual-OPL2 + OPL3

csnover:
Based on upstream revision
02a4917f30ace7de14d561ad09e2b4579facaf13.

Fixes Trac#7884.

Obviously, this shouldn’t land until after the release is
branched. 😎

I tried to keep this as updatable as possible by separating
the core ScummVM code (previously everything was just in
mame.cpp), and then making any necessary adjustments within
the MAME code using some obvious #ifdefs. Since MAME uses
C++14 now, I had to do some hacking to make it compile in
C++98 compilers. (I imagine this will become increasingly
difficult as time goes on…)

I tested this very lightly with EcoQuest & PQ3 for dual-OPL2
and IHNM for OPL3. Neither EQ nor PQ3 are quite right, but
they are not quite right in both the MAME and DOSBox OPLs,
so it is probably a SCI bug. (One of the reasons why I
wanted to get this stuff up-to-date was because I’m going to
start on rewriting SCI MIDI stuff soon and want to make sure
the underlying synths are as accurate as possible.)

Please bang on this and report any issues that didn’t
exist before! Thanks!

### DIFF
--- a/audio/fmopl.cpp
+++ b/audio/fmopl.cpp
@@ -59,7 +59,7 @@ OPL::OPL() {
 
 const Config::EmulatorDescription Config::_drivers[] = {
 	{ "auto", "<default>", kAuto, kFlagOpl2 | kFlagDualOpl2 | kFlagOpl3 },
-	{ "mame", _s("MAME OPL emulator"), kMame, kFlagOpl2 },
+	{ "mame", _s("MAME OPL emulator"), kMame, kFlagOpl2 | kFlagDualOpl2 | kFlagOpl3 },
 #ifndef DISABLE_DOSBOX_OPL
 	{ "db", _s("DOSBox OPL emulator"), kDOSBox, kFlagOpl2 | kFlagDualOpl2 | kFlagOpl3 },
 #endif
@@ -167,11 +167,14 @@ OPL *Config::create(DriverId driver, OplType type) {
 
 	switch (driver) {
 	case kMame:
-		if (type == kOpl2)
-			return new MAME::OPL();
-		else
-			warning("MAME OPL emulator only supports OPL2 emulation");
-		return 0;
+		switch (type) {
+		case kOpl2:
+			return new MAME::OPL<kOpl2>();
+		case kDualOpl2:
+			return new MAME::OPL<kDualOpl2>();
+		case kOpl3:
+			return new MAME::OPL<kOpl3>();
+		}
 
 #ifndef DISABLE_DOSBOX_OPL
 	case kDOSBox:
--- a/audio/module.mk
+++ b/audio/module.mk
@@ -46,6 +46,9 @@ MODULE_OBJS := \
 	softsynth/opl/dbopl.o \
 	softsynth/opl/dosbox.o \
 	softsynth/opl/mame.o \
+	softsynth/opl/mame/emu/attotime.o \
+	softsynth/opl/mame/sound/fmopl.o \
+	softsynth/opl/mame/sound/ymf262.o \
 	softsynth/fmtowns_pc98/towns_audio.o \
 	softsynth/fmtowns_pc98/towns_euphony.o \
 	softsynth/fmtowns_pc98/towns_midi.o \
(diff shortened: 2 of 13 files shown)


waltervn:
I wonder about two things here. 1) Is this new implementation fast enough for our low-end ports like Dreamcast? 2) Assuming this new implementation is comparable to DOSBox OPL in terms of accuracy and speed, is there any real benefit to adding it?

csnover:
> 1) Is this new implementation fast enough for our low-end ports like Dreamcast?

This is an interesting question, and I don’t think those low-usage ports should be deciding factors for improvements to ScummVM. If this patch achieves its goal of improving the OPL emulation quality/accuracy, the overwhelming majority of users will benefit from this change; if a port needs some downgraded thing, porters may maintain a faster-but-broken derivative.

Some data which I can provide is that on the PQ3 theme it’s using about 30% more time than DOSBox OPL in the new dual OPL2 mode and about 63% more time than the old one in OPL2 mode. In real-world terms this is still a tiny amount of CPU time, just 4–5% (including locking and mixing overhead) of a single core of a 2.3GHz Core i7, which is about the same amount of CPU time it takes VLC on this machine to play back a 256kbit Vorbis file.

> 2) Assuming this new implementation is comparable to DOSBox OPL in terms of accuracy and speed, is there any real benefit to adding it?

That’s an excellent question. I don’t know what is the difference in accuracy. I hope that others can provide more information about this through testing. I am assuming that MAME will be more accurate, especially since it looks like dual OPL2 in DOSBox OPL is actually running an OPL3, which is close but not quite compatible. If they have truly equivalent accuracy, then there’s really not even a reason to have two different OPL emulators, so it would be good to get more data here since then we could simplify this a bit for users by just removing one of them.

waltervn:
> This is an interesting question, and I don’t think those low-usage ports should be deciding factors for improvements to ScummVM.

In that case we should probably have a discussion (on the mailing list) about whether or not these ports should remain to be officially supported. However, as long as they are supported, I don't think we should remove code that might be essential to those ports. I say "might" here, as it's possible this new code works fine for them. It would be great if some of the porters could try out this patch.

> if a port needs some downgraded thing, porters may maintain a faster-but-broken derivative.

The old MAME OPL currently is our faster-but-broken option. One could argue that if this new MAME OPL implementation is better than DOSBox OPL, it is DOSBox OPL that should be removed instead.

> I am assuming that MAME will be more accurate, especially since it looks like dual OPL2 in DOSBox OPL is actually running an OPL3, which is close but not quite compatible.

My apologies in advance for going a little off-topic here. I'm sure you're aware of this, but Dual OPL2 is not a sound chip, it's a sound card configuration with two OPL2 chips. One routed to the left audio channel, and one routed to the right audio channel. IMHO the most obvious way to implement this would be to use two instances of an OPL2 emulator. This can be done even with the old MAME code (in fact we did exactly that at FreeSCI).

However, AFAIK only SCI is currently using that Dual OPL2 mode. There's quite a bit of Dual OPL2 code all over the place just to accommodate that small bit of SCI code that could easily be rewritten to use the OPL3 interface instead. As I've argued in the past, we may want to consider removing the Dual OPL2 interface entirely.

sev-:
I propose to rename this driver to "MAME OPL emulator (accurate)" and still keep the old one as "MAME OPL emulator (fast)".

lskovlun:
It may be of interest to note that the DGDS games (Dynamix) borrowed Sierra's sound code (it was around the time Sierra bought Dynamix), so the SCI MIDI code may be useful in that context. Sierra conversely got the scaler code from Dynamix. There was some work being done on DGDS in the past, but nothing came of it.

wjp:
@lskovlun vasc has been working on dgds recently at https://github.com/vcosta/scummvm (and he has talked to snover about midi code on IRC).

sev-:
Any comments on my proposal to keep both?

bonki:
I think it's reasonable to keep both for the time being (as long as we avoid maintaining "dead code") and if it turns out that all low-end ports are able to handle the new implementation without problems and have switched over we can always drop "Mame (fast)" then.

csnover:
MAME’s OPL3 emulation is not working correctly. It deterministically randomly fails to play notes on one of the channels even though the OPL data stream is correct and works on Dual OPL2 and DOSBox OPL3. I haven’t been able to tell a difference in accuracy versus the faster DOSBox softsynth, so I am not planning to move forward with this.

PR #1072, opened 2017-11-28
dists: enforce SDL_MOUSE_RELATIVE=0 on XDG systems

yegortimoshenko:
Mouse is very sensitive when ScummVM is fullscreen. This fixes that problem.

This probably won't get accepted in this form, setting some option set in SDL frontend probably will be more appropriate.


csnover:
Thanks for your patch! You’re right that this will not be accepted in its current form. There is no `SDL_MOUSE_RELATIVE` anywhere in SDL2, this was only a thing in SDL1.2, and SDL1.2 is quite old and deprecated. If you are having trouble with current master builds using SDL2, please open a ticket [on the bug tracker](https://bugs.scummvm.org/) with a more detailed description of what is the problem, including expected and actual behaviour. Thanks!

PR #1074, opened 2017-11-29
SDL: Don't use the window size hint for fullscreen

bgK:
This prevents the OpenGL backend from changing mode when entering fullscreen for games with a window size hint.

When in fullscreen with OpenGL, unlike with the Renderer API, SDL2 does not automatically scale the window contents to fit the screen. Which means the window size must be the screen size. Passing windowed mode dimensions to SDL_CreateWindow in that case seems incorrect.

Fixes #10335.


wjp:
I can confirm that this fixes Trac10335 for me. I've tested Dreamweb, Freddy Pharkas and Space Quest 6 in OpenGL and non-OpenGL modes, windowed and fullscreen. All behave fine now. (64 bit Gentoo linux, with Xorg, nvidia binary drivers and SDL2 2.0.7)

PR #1076, opened 2017-12-01
SCI: Sabotage qfg1vga's speed test

wjp:
This ensures the detected speed will end up at the highest level. This
improves the detail in Yorick's room (96), and slightly changes the
timing in other rooms.

Fixes bug #10296.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -5183,6 +5183,27 @@ static const uint16 qfg1vgaPatchBrutusScriptFreeze[] = {
 	PATCH_END
 };
 
+// Speed up the speed test by a factor 50, ensuring the detected speed
+// will end up at the highest level. This improves the detail in
+// Yorick's room (96), and slightly changes the timing in other rooms.
+//
+// Method changed: speedTest::changeState
+static const uint16 qfg1vgaSignatureSpeedTest[] = {
+	0x76,                               // push0
+	0x43, 0x42, 0x00,                   // callk GetTime 0
+	SIG_MAGICDWORD,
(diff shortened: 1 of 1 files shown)


csnover:
This seems fine if you don’t want to spend any more time thinking about this, but I think it would probably be better to just pave over the end of `speedRoom::init`, have it set local0 to 0 and then just call `startGame::doit`, replacing the `self.setScript(speedTest)`, since that would guarantee the maximum setting regardless of any temporary CPU load at game startup (I don’t think these speed tests are valid for the way ScummVM/modern systems work anyway). If that screws up something else that doesn’t expect room init to immediately change rooms, the paving could be done in `speedTest::changeState` instead, to make state 0 set local0 and then jump to the `startGame.doit()` call.

wjp:
I started writing that patch initially, but this seemed more likely to work with all game versions.

csnover:
That’s a fair rationale. I still forget about how many more derivatives there were in the SCI16 era (SCI32 had only 6 versions of the speed test room).

csnover:
There is not urgency on either of these SCI PRs, so I would suggest to defer to @wjp once he has time to take care of it and not worry about landing them yourself.

digitall:
@csnover : I am happy to defer to @wjp to merge this, but my point was he was the submitter of the PR and appears to be happy with this as-is and you are OK as well... so is anything stopping this just being merged? I am trying to make sure that we do not have too many PRs just hanging around in limbo, partly as newer changes to the tree end up breaking clean merge and then they need rework...

csnover:
Don’t worry about it. @wjp and I both know how to use git rebase. This isn’t going to break and require rework due to tree changes.

PR #1079, opened 2017-12-01
ALL: Remove obsolete register keyword

csnover:
The register keyword was deprecated from the C++11 standard,
<http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4193.html#809>,
and removed from the C++17 standard,
<http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4340>, so
cannot exist in a well-formed C++17 program.

It has never done anything in GCC
<https://gcc.gnu.org/ml/gcc/2010-05/msg00113.html>
and because of the way it is specified in the standard, it is “as
meaningful as whitespace”
<http://www.drdobbs.com/keywords-that-arent-or-comments-by-anoth/184403859>.

The one remaining use of the register keyword is in the DS backend,
where it is used to create a local register variable using the
non-standard GCC Extended Asm feature.

### DIFF
--- a/audio/rate.h
+++ b/audio/rate.h
@@ -46,7 +46,7 @@ enum {
 };
 
 static inline void clampedAdd(int16& a, int b) {
-	register int val;
+	int val;
 #ifdef OUTPUT_UNSIGNED_AUDIO
 	val = (a ^ 0x8000) + b;
 #else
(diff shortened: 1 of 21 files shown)


wjp:
Looks good.

csnover:
Landed. Thank you for your review!

PR #1080, opened 2017-12-02
Revert "WAGE: Fix crashes for poorly initialized objects"

csnover:
This reverts commit ea0fb987e042a86b8da683cafa7b9cf04d1636e6.

This issue was discovered during a full build of ScummVM using Clang 5,
which was generating a warning about this code since it is not valid.

It is UB to call a null pointer. Whatever caller is trying to do so needs
to be fixed instead.

My opinion is that keeping this workaround in the codebase instead of
fixing the bad calls runs the risk of allowing more bad calls to slip in
during development of the engine, and may make it more difficult (if it
is not already very difficult already since this engine has not been
worked on for a while) remember what the condition was where this invalid
call happened and fix it.

### DIFF
--- a/engines/wage/entities.h
+++ b/engines/wage/entities.h
@@ -125,7 +125,7 @@ public:
 
 	void setDesignBounds(Common::Rect *bounds);
 
-	Common::String toString() { if (!this) return "<NULL>"; return _name; }
+	Common::String toString() { return _name; }
 };
 
 class Chr : public Designed {
--- a/engines/wage/entities.h
+++ b/engines/wage/entities.h
@@ -125,7 +125,7 @@ public:
 
 	void setDesignBounds(Common::Rect *bounds);
 
-	Common::String toString() { return _name; }
+	Common::String toString() const { return _name; }
 };
 
 class Chr : public Designed {
--- a/engines/wage/script.cpp
+++ b/engines/wage/script.cpp
@@ -56,6 +56,13 @@
 
 namespace Wage {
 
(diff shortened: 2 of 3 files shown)


wjp:
I agree.

sev-:
OMG, so we again are fixing compiler warnings, and even doing it blindly?

It will be tons of code to fix, as _value in script.cpp is a union, and in number of cases it has value 0.

This code perfectly fixes that, e.g. constructs in form of ```return _value.obj->toString()```

wjp:
Calling methods on null pointers leads to very confusing code and unclear conventions, and avoiding that seems very desirable.

Might I suggest a function similar to the following for handling these cases?

```
Common::String toString(const Designed *d) {
	if (!d)
		return "<NULL>";
	else
		return d->toString();
}
```

hpvb:
Calling methods on null pointers also crashes on clang 5+ and GCC 6+. It's explicitly forbidden and null checks on the this pointer are removed as dead code.

I had to deal with this problem on Godot.

wjp:
Since this method only turned out to be called in one place, I've taken the liberty of updating the calls and pushing that to this PR's branch.

wjp:
By the way, `Script::Operand::toString()` also dereferences another union member `_value.string` via another code path. If this can also be null, we should catch that too.

sev-:
Yes, it could be NULL, and I like the approach you suggested.

sev-:
Cool, thanks!

PR #1081, opened 2017-12-02
CRYO: Stop packing struct containing pointers

csnover:
This struct packing causes the pointer-to-member-function to become
unaligned, and does not seem necessary in any way.

This was added in commit 7bf6e1e8502 by @Strangerke, and I asked
about this a few weeks ago but received no more information, so
I’m opening this as a reminder (and also as a friendly threat 😇)
that I’ll go ahead with removing the pragma soon unless more
clarity can be added around why it is needed. Thanks!

### DIFF
--- a/engines/cryo/eden.cpp
+++ b/engines/cryo/eden.cpp
@@ -3126,13 +3126,11 @@ void EdenGame::tyranDies(perso_t *perso) {
 }
 
 void EdenGame::specialObjects(perso_t *perso, char objid) {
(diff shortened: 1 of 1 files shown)


wjp:
The struct in question is only used in a *very* limited way, and I see no way that packing could have any effect on it. Removing the packing sounds good to me.

sev-:
Looks good to me too.

csnover:
Thanks for your feedback! This is landed now.

PR #1082, opened 2017-12-02
SKY: Fix unaligned pointer dereferences

csnover:
Dereferencing an unaligned pointer will cause a memory fault on
at least older ARM and SPARC architectures, and is warned about
starting in Clang 4.


wjp:
The changes look good.

The code duplication between get/set could probably be removed using READ_UINT16//WRITE_UINT16 macros, but on the other hand this engine is very stable by now, so that's probably not worth it.

sev-:
Looks good and appropriate. Merging. Thanks!

PR #1083, opened 2017-12-02
SCUMM: Remove dead code

csnover:
Clang optionally warns about this dead code. If there is
some reason not to delete it all outright (which is why I
open this as a PR instead of just committing this deletion
directly), `#if 0` should be added around the dead code to
suppress the warning.

### DIFF
--- a/engines/scumm/object.cpp
+++ b/engines/scumm/object.cpp
@@ -152,10 +152,6 @@ void ScummEngine::setOwnerOf(int obj, int owner) {
 		if (ss->where == WIO_INVENTORY) {
 			if (ss->number < _numInventory && _inventory[ss->number] == obj) {
 				error("Odd setOwnerOf case #1: Please report to Fingolfin where you encountered this");
-				putOwner(obj, 0);
-				runInventoryScript(arg);
-				stopObjectCode();
-				return;
 			}
 			if (ss->number == obj)
 				error("Odd setOwnerOf case #2: Please report to Fingolfin where you encountered this");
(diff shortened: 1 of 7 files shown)


engines/scumm/object.cpp
```diff
@@ -152,10 +152,6 @@ void ScummEngine::setOwnerOf(int obj, int owner) {
 		if (ss->where == WIO_INVENTORY) {
 			if (ss->number < _numInventory && _inventory[ss->number] == obj) {
 				error("Odd setOwnerOf case #1: Please report to Fingolfin where you encountered this");
-				putOwner(obj, 0);
```
sev-: This must be left intact until we catch it. Then we study the conditions and remove the error(). This code matches disassembly.

sev-: If you comment out this (I assume you wanted to shut the warnings), instead of removing, the patch could be merged.

sev-:
Any updates?

csnover:
This error has never been reported in the bug tracker in the 12 years since this code was committed, so it doesn’t make sense to me to retain it in any form, since all evidence suggests that it is never used even if it existed in the original interpreter. I don’t really care either way, I was just fixing up some dead code warnings, so if you would like to not remove it then that’s fine by me and you can just close this.

sev-:
Good point, but then I suggest to turn them into comments. Could you do that and merge?

bonki:
I would not go as far as saying that just because something has never been reported it has never occurred, that's committing a logical fallacy. I don't expect every user of ScummVM to be technically interested or know who Fingolfin is. Also, because you can just ignore the error and continue playing the game, I would say it lowers the chance of someone actually reporting it in the unlikely case of it triggering.

That being said, I agree that it seems unlikely by now that it's used anywhere but I'd also say that leaving it in place commented out seems best.
I also suggest updating both error messages with more appropriate ones (maybe with a link to http://wiki.scummvm.org/index.php/User_Manual/Reporting_bugs ?).

sev-:
Closing. I'll do manual merge of some features.

PR #1084, opened 2017-12-02
SCUMM: Fix stack overflows

csnover:
These two fixes should be good to go as-is, but I wanted
to open this as a PR anyway since I don’t think these are the
only two places where there is a stack overflow in SCUMM, didn’t
want to commit these and not check all the other places first,
but I don’t have the time right now to do research on all the
other places where I found some `int args[N]` in the codebase.
So maybe someone else can look for me? It could be an interesting
small project taking an hour or two, and I would appreciate it.

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -768,7 +768,7 @@ void Actor::startWalkAnim(int cmd, int angle) {
 	 * work as usual
 	 */
 	if (_walkScript) {
-		int args[16];
+		int args[NUM_SCRIPT_LOCAL];
 		memset(args, 0, sizeof(args));
 		args[0] = _number;
 		args[1] = cmd;
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -2844,7 +2844,7 @@ void Actor::runActorTalkScript(int f) {
 
 	if (_talkScript) {
 		int script = _talkScript;
-		int args[16];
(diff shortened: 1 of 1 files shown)


raziel-:
Do you just need the location of those "int args[N]" (file and line) or do you also need them to be fixed in the correct manner?

Because the first i could provide...the second not, i'm afraid.

raziel-:
Can they ALL be changed to [NUM_SCRIPT_LOCAL]?

raziel-:
I tried, please don´t be mad @csnover  :-)

csnover:
> Can they ALL be changed to [NUM_SCRIPT_LOCAL]?

This is the question which I was hoping to have answered, but don’t have time now to research myself. I assume that not all of these instances are actually being used to initialise locals and so some of them are appropriately sized and don’t need to be changed, but it requires looking at the call graph starting from initializeLocals and seeing who ends up calling initializeLocals with one of these incorrectly sized buffers.

raziel-:
OK, so i did understand it wrong, sorry

Nothing i can help with

csnover:
No problem. Thanks for your willingness to give it a go!

sev-:
That's a good catch, however, generally there are only a few ops which use all 25 arguments. Still I'll look more into the code for other places like this.

sev-:
I am merging this, and checked other usages (well, where size of args[] was < 16). All other places seems to be fine. In most cases it has nothing to di with the scripts, just local list of arguments or even a temporary variable.

PR #1085, opened 2017-12-02
SCUMM: Fix stack overflow

raziel-:
Due to @csnover ´s request on his PR #1084, i created what i think he asked for.

But i don´t know for sure.
I also don´t know how to add to your PR, so i created my own.

I *also* don´t know how to fix the spelling mistakes in the changes and the one change in debugger.cpp i accidently didn´t name correct, wanted to revert and instead added a change, reverted it and added it again.

I have done this all on the website, so please don´t be too harsh...

So, if this is not what you wanted just scap it, if you can work with this, i´ll add you as buddy (or whatever that is called when you can change something in my fork) :-)

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -768,7 +768,7 @@ void Actor::startWalkAnim(int cmd, int angle) {
 	 * work as usual
 	 */
 	if (_walkScript) {
-		int args[16];
+		int args[NUM_SCRIPT_LOCAL];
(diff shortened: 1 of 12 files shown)


csnover:
Thank you again for your willingness to look at this! Per the discussion on gh-1084, it’s necessary to actually investigate each call site to see whether or not the size of the stack-allocated array is appropriate or not, so just doing a blanket replacement is probably not what is needed here.

PR #1086, opened 2017-12-03
SDL: Fix memory corruption when switching to/from 32-bit cursors

csnover:
When a 32-bit cursor has the same size as an 8- or 16-bit cursor,
the mouse surfaces were not being regenerated even though the
32-bit cursors have a different memory requirement. This lead to
memory corruption as an inappropriate surface would be used for
the other type of cursor.

The shoe-horned 32-bit cursor support is clearly showing its
scrappy nature here and probably ought to be revisited in the
future if the SurfaceSdl graphics manager sticks around.

Fixes Trac#10349, Trac#10350, Trac#10351.

I am super tired so there may be a better way to do this which
I don’t see right now. I also did not do any double-checking that
this is not leading to leaks of surfaces, hoping someone can do
that or at least give me some feedback that I can use to fix this
and land this fix first thing tomorrow. If someone wants to just
take this fix and make it better and land it before I get to it,
that is fine too.

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -1836,12 +1836,20 @@ void SurfaceSdlGraphicsManager::copyRectToOverlay(const void *buf, int pitch, in
 #pragma mark -
 
 void SurfaceSdlGraphicsManager::setMouseCursor(const void *buf, uint w, uint h, int hotspotX, int hotspotY, uint32 keyColor, bool dontScale, const Graphics::PixelFormat *format) {
+	bool formatChanged = false;
+
 	if (format) {
 #ifndef USE_RGB_COLOR
 		assert(format->bytesPerPixel == 1);
 #endif
+		if (format->bytesPerPixel != _cursorFormat.bytesPerPixel) {
+			formatChanged = true;
+		}
 		_cursorFormat = *format;
 	} else {
+		if (_cursorFormat.bytesPerPixel != 1) {
+			formatChanged = true;
+		}
 		_cursorFormat = Graphics::PixelFormat::createFormatCLUT8();
 	}
 
@@ -1856,13 +1864,9 @@ void SurfaceSdlGraphicsManager::setMouseCursor(const void *buf, uint w, uint h, 
 
 	_cursorDontScale = dontScale;
 
-	if (_mouseCurState.w != (int)w || _mouseCurState.h != (int)h) {
+	if (_mouseCurState.w != (int)w || _mouseCurState.h != (int)h || formatChanged) {
 		_mouseCurState.w = w;
 		_mouseCurState.h = h;
-
-		if (!w || !h) {
-			return;
-		}
 
 		if (_mouseOrigSurface) {
 			SDL_FreeSurface(_mouseOrigSurface);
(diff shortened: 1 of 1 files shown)


bgK:
Sorry, but bug #10349 seems not to be fixed in Myst ME:

screenshot from 2017-12-03 13-51-51

The rest looks ok to me. I'm not sure it's worth bothering to check whether the format changed to / from 4bpp. Maybe checking the format changed at all is good enough.


> If the SurfaceSdl graphics manager sticks around.

IMO, it will have to stay around for the platforms that don't work with OpenGL - PS3, Vita, ... But maybe it can be simplified by splitting the SDL 1 and SDL 2 code into different classes, letting the SDL 1 version slowly die.

dafioram:
bgK, I cannot reproduce what you are seeing in Myst ME. The cursors work for me in Myst ME before and after this PR. Besides the fact that you are using classic theme what else is your setup (SDL2 version)? I am on 2.0.7 and it looks fine.

With this PR applied I tested urban runner, myst me, rama, and titanic and the cursors looked as expected in these three games.

bgK:
I'm using SDL 2.0.7. It seems to happen only when not in "normal 1x" mode, and when hovering that linking book hotspot.
A quick debugging session shows it's because `_mouseSurface` still is 4bpp from the in game mouse cursor while it should be 2bpp. The code in `setMouseCursor` should be resetting `_mouseSurface` to null in that case, but is not.

csnover:
Thank you! I went ahead and made the format change check less specific, and now `_mouseSurface` is being cleared in `setMouseCursor` so things should be OK, but I do not have Myst ME to check that one. If you could, give it another go and let me know if it’s fixed there. Once that’s confirmed then I will land this. I did also do a check and it looks like everything is being cleaned up appropriately without making any leaks.

bgK:
Now Myst ME fails when switching from a 4bpp cursor to another:
```
scummvm: ../backends/graphics/surfacesdl/surfacesdl-graphics.cpp:1891: virtual void SurfaceSdlGraphicsManager::setMouseCursor(const void*, uint, uint, int, int, uint32, bool, const Graphics::PixelFormat*): Assertion `!_mouseSurface' failed.

Thread 1 "scummvm" received signal SIGABRT, Aborted.
0x00007ffff40eb8a0 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff40eb8a0 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff40ecf09 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff40e40dc in __assert_fail_base () from /usr/lib/libc.so.6
#3  0x00007ffff40e4153 in __assert_fail () from /usr/lib/libc.so.6
#4  0x00005555559548d4 in SurfaceSdlGraphicsManager::setMouseCursor (this=0x61d000030c80, buf=0x6190000c1780, w=16, h=16, hotspotX=8, hotspotY=8, 
    keyColor=16777215, dontScale=false, format=0x7fffffffb6b0) at ../backends/graphics/surfacesdl/surfacesdl-graphics.cpp:1891
#5  0x00005555559186c2 in ModularBackend::setMouseCursor (this=0x611000000180, buf=0x6190000c1780, w=16, h=16, hotspotX=8, hotspotY=8, 
    keycolor=16777215, dontScale=false, format=0x7fffffffb6b0) at ../backends/modular-backend.cpp:225
#6  0x0000555555a69683 in Graphics::CursorManager::replaceCursor (this=0x60400007fe10, buf=0x6190000c1280, w=16, h=16, hotspotX=8, hotspotY=8, 
    keycolor=16777215, dontScale=false, format=0x7fffffffb6b0) at ../graphics/cursorman.cpp:147
#7  0x000055555580663b in Mohawk::MystCursorManager::setCursor (this=0x602000094f10, id=200) at ../engines/mohawk/cursors.cpp:138
#8  0x000055555573d68e in Mohawk::MohawkEngine_Myst::checkCursorHints (this=0x614000019040) at ../engines/mohawk/myst.cpp:1078
#9  0x00005555557393b2 in Mohawk::MohawkEngine_Myst::checkCurrentResource (this=0x614000019040) at ../engines/mohawk/myst.cpp:750
#10 0x000055555573591d in Mohawk::MohawkEngine_Myst::doFrame (this=0x614000019040) at ../engines/mohawk/myst.cpp:439
#11 0x0000555555734883 in Mohawk::MohawkEngine_Myst::run (this=0x614000019040) at ../engines/mohawk/myst.cpp:347
```

That condition is not true, and `_mouseSurface` is not reset to nullptr:
`if (_mouseSurface == _mouseOrigSurface) {`

bgK:
In the failure case, the cursor changes size but keeps the same pixel format.

csnover:
I’m glad I put those assertions in there. I’ve pushed an update so that the `_mouseSurface` will also get released if the new Bpp is 4, which I think should resolve that final problem. Let me know how it goes.

bgK:
Myst ME works fine now. Thanks a lot for taking care of this!

csnover:
No problem. Thank you for your help reviewing and testing so it actually got fixed correctly.

csnover:
In 20b2c1b7e156f0586799f7df9d6e93c757dabeac and 58cbd45ef5b6776b761fc3d51f531a8d12946c24.

PR #1087, opened 2017-12-07
SCI: Add RGB rendering mode and per-resource palette mods

wjp:
This branch adds an RGB rendering mode (16/32 bpp) to the SCI engine.

This rendering mode has three purposes:

* Avoid mode changes when showing RGB videos (Mac Quicktime)
* For screen transitions, allow the previous screen to keep its own palette. This avoids the ugly effect where the previous screen is rendered using the palette of the new screen.
* Open up the possibility to modify the palette of specific views. This was a feature of FreeSCI that we had removed when we merged the FreeSCI code into ScummVM. FreeSCI had (entirely optional) configuration files for a few EGA games that used this to improve the visuals.

For the third, I have added a python script in devtools/ that turns FreeSCI's configuration files into C++.

In the code, the `_paletteMapScreen` is solely used for modified view/pic palette support, and `engines/sci/graphics/scifx.cpp` contains the modifications. (These two actually form the bulk of the changed lines in this PR.)

---

Screenshots for the transition changes:

indexed transition

RGB transition

---

Screenshots for the proof of concept of configurable view changes:

SQ3 original

SQ3 modified


matthargett:
You are my hero for resurrecting this. I’ve shown dozens of people my before and after screenshots for the per resource palettes feature over the years. It was probably the last “major” work I did on FreeSCI, so it was a bit sad to me that it got lost in the import into scummvm. I had started similar work in Sarien back then (early 2000s) as well — is there any interest in a PR for that once this gets merged? Thanks again for reintroducing this feature!

sev-:
Yes, I'd like to see this for AGI as well!

wjp:
This is now in a state where it can be reviewed, and I've cleaned up the history of the branch.

Instead of directly parsing the config files, I've chosen to initially use an external python script to convert the scifx config files (currently two, for LSL2 and SQ3) from FreeSCI into C++ code. This could be changed in the future if there is a demand for it.

matthargett:
It would be nice if it was interpretable, rather than compiled.  The scifx syntax wasn't great, but maybe a simple JSON or CSV approach would work. In my Sarien patch, I didn't allow totally arbitrary palettes but allowed per-resource to use EGA, Amiga, or ][gs. This may be a simpler, but still very effective approach for subtle enhancement that would make the configuration files easier to read.

In a mobile game I was working on, we extended EGA palette by an additional 12 colors by taking the non-black/white/greay colors and adding/subtracting 0x55 from their RGB values. This gave us another strip of 6 darker colors and another strip of 6 brighter colors. As long as their never presented as a 4-step gradient, you'd never know it wasn't "pure" EGA. I mention this because using a darker-skewed palette for background pixmaps of dark interiors (like opening playable scene of SQ3) is something I did in the scifx file with a 25% reduction of the entire palette that was very effective for helping ego to not "blend in" too much. To simplify the config file format, and still subtly enhance scenes like those, defining another "dark" palette beyond Amiga, EGA, ][gs might be a good idea.

bluegr:
Overall, great work!

Commit 438d09ba7c has been merged as commit cf1785b23e. Commit 5333006636 will need to be adapted. The other two should be (almost) fine to merge.

This will require some testing, in case something got broken, but testing just a single game from each major SCI version should suffice

sev-:
@bluegr could you please rebase it and have this merged?

bluegr:
The SCI RGB changes have been adapted and committed. Still pending: the custom palette mods changes.

bluegr:
The custom palette mods changes have been merged, so this can be closed now. Thanks again for all the work!

PR #1088, opened 2017-12-17
SDL: Map SDL game controller events to ScummVM events

bgK:
SDL2 introduced the Game Controller API to complement the existing Joystick API. SDL game controllers are joysticks with a known button layout. SDL2 uses the Xbox 360 gamepad as the reference layout and maintains a database to map buttons of popular controllers to that layout.
xbox_360_controller_diagram

This pull request adds initial support for the game controller API and defines a button to ScummVM event mapping. The keymap is that of the Vita port:
- Left stick     => Mouse
- R Bumper + Left stick => Slow Mouse
- A          => Left mouse button
- B         => Right mouse button
- DPad           => Cursor Keys (useful for character motion)
- R Bumper + DPad       => Diagonal Cursor Keys
- L Bumper      => Game menu (F5)
- R Bumper      => Shift (used to enable Mass Add in menu)
- X         => Period '.' (used to skip dialog lines)
- R Bumper + X     => Space
- Y       => Escape (used to skip cutscenes)
- R Bumper + Y   => Return
- Start          => ScummVM's global in-game menu
- Select         => Toggle virtual keyboard
- R Bumper + Select     => AGI predictive input dialog

@rsn8887 Does the Vita port of SDL support the Game Controller API? It would be nice to remove the duplicate key mapping code in `psp2sdl-events.cpp`.


dafioram:
Would it make sense to also assign functionality to the bumper keys?

Since you are using a mapping that works for vita maybe we could keep that and add extra functionality for more keyboard keys. Like like Tab, F1, middle mouse click, and middle mouse scroll.

Maybe:
L bumper + D-Pad Vertical Axis  = Middle mouse scroll
R stick click = Middle mouse click
R bumper + select = tab
R bumper + start = F1

backends/events/sdl/sdl-events.cpp
```diff
@@ -867,70 +903,176 @@ bool SdlEventSource::handleJoyButtonUp(SDL_Event &ev, Common::Event &event) {
...
+	if (xAxis > JOY_DEADZONE) {
+		xAxis -= JOY_DEADZONE;
+	} else if (xAxis < -JOY_DEADZONE) {
+		xAxis += JOY_DEADZONE;
+	} else
+		xAxis = 0;
+	if (yAxis > JOY_DEADZONE) {
+		yAxis -= JOY_DEADZONE;
+	} else if (yAxis < -JOY_DEADZONE) {
+		YAxis += JOY_DEADZONE;
```
rsn8887: Typo

backends/events/sdl/sdl-events.cpp
```diff
@@ -630,6 +642,30 @@ bool SdlEventSource::dispatchSDLEvent(SDL_Event &ev, Common::Event &event) {
...
+			return handleControllerButton(ev, event, false);
+		case SDL_CONTROLLERBUTTONUP:
+			return handleControllerButton(ev, event, true);
+		case SDL_CONTROLLERAXISMOTION:
+			return handleControllerAxisMotion(ev, event);
+		}
+	}
+#endif
+
 	return false;
```
rsn8887: This looks like it duplicates the handling. What if both joystick and game controller is present at the same time?

rsn8887:
The Vita currently doesn't support this, but this can easily be changed by adding a game controller config at runtime. Since the GUID and button numbers are the same on the Vita port of SDL2 for the DS4, DS3 and Vita controllers, just one config needs to be added to support all the controllers.

However have you actually tested this? I am asking because there's a typo in the code (YAxis -> yAxis).

Make sure that when a game controller is present, the joystick is not still firing events in addition to the game controller. This might duplicate all the handling routines. and mess up the mouse speed.

On the Vita port, there's an example to handle pointer speed and deadzone GUI settings. Those are features that can be enabled for all devices that support game controllers. That way the user can set both deadzone AND pointer speed in the control menu of the gui.

In addition, there's some conversion in the code from analog joystick values to digital values that still uses the wrong method. It should use angular zones. I do not understand why it is there, since game controller joysticks are always analog? In fact it uses analog values from the joystick but then converts them to digital values.

backends/events/sdl/sdl-events.cpp
bgK: The `openJoystick` method makes sure `_joystick` and `_controller` are not both set at the same time.  The checks in this method ignore the joystick events when the device is recognized as a game controller, so all is fine.

bgK:
> Although the Vita currently doesn't support this, this can easily be changed by adding a single call to SDL_GameControllerAddMapping at runtime. Since the GUID and button numbers are the same on the Vita port of SDL2 for all the controllers (DS4, DS3 and Vita controllers), just one config line needs to be added to support all of them.

Are you interested in making the required change?

> There's a typo in the code (YAxis -> yAxis).

Good catch. That was embarrassing.

> JOY_ANALOG should always be defined with this to ensure that the kFeatureJoystickDeadzone is present and the control options tab will display the settings for both pointer speed AND deadzone.

> I do not understand why the new code branch for !defined(JOY_ANALOG) uses analog values reported from what seems to be an analog joystick(?). Then it converts these values to some digital values. This seems unnecessary. Also, since it starts from analog axis values and uses a deadzone, the conversion to digital values should be done correctly using angular zones. Angular zones give correct diagonals. I didn't bother with this on the Vita since it always has JOY_ANALOG defined. Angular zones are explained here.

> EDIT: It looks like this code !JOY_ANALOG code is only used on Symbian which is unsupported? Maybe it can be removed or at least left alone.

I've made some additional changes to address those points:
- The !JOY_ANALOG branch now lives in the Symbian port so it is removed when the Symbian port is declared dead. I didn't look to closely at how it works nor am I interested in fixing it. The new code should do exactly the same thing as before.
- I've fixed the deadzone setting not appearing in the controls tab by removing JOY_ANALOG altogether.
- I've changed the joystick settings to appear in the GUI only if the `joystick_num` config option is set.

I've also added some changes to allow joysticks to be hotplugged (Fixes https://bugs.scummvm.org/ticket/10366).

bgK:
> Would it make sense to also assign functionality to the bumper keys?

I mixed up the bumpers and triggers in the PR description (fixed now). Bumpers already have actions assigned. On the other hand, triggers are analog axes in the SDL model. For now we don't support mapping those to key presses. I'd prefer not to extend the scope of this pull request to add that feature.

rsn8887:
I added the controller configuration commands for Vita.

rsn8887:
It only works partially. Shift is pressed down when R Shoulder is up, and Shift if up when R shoulder is down. Everything else seems to work correctly.

But more importantly, the mapping of controller buttons according to https://wiki.libsdl.org/SDL_GameControllerButton uses an enum, e.g. 
SDL_CONTROLLER_BUTTON_A
SDL_CONTROLLER_BUTTON_B etc.
not numbers.

These enums can in principle change between SDL versions.

Same for the button state, it should query
SDL_CONTROLLERBUTTONDOWN
or
SDL_CONTROLLERBUTTONUP

We cannot assume that one of these is zero or one etc. So the mapping array must be indexed using the enum names, not implicitly using numbers.

rsn8887:
I fixed the problem with shift but found tons more problems:
Start/Select for menu and keyboard doesn't work either.

rsn8887:
It is all fixed.

bgK:
Very good, thanks!

I have integrated your fixes in the relevant commit and shortened the commit message so it follows the guidelines.

rsn8887:
Perfect, all good now.

dafioram:
1. I am not seeing the virtual keyboard when I hit select/back on my controller, besides compiling with vkeybd enabled is there something else I should be doing to get that to pop up?
2. The cursor can be left drifting when the player returns the controller joystick to the center (this is true before this PR as well). This is probably out of scope for this PR. Maybe something on SDL's end?
3. The hotplugging isn't working for me. I have to restart scummvm if I reconnect my controller.

rsn8887:
Concerning 1. and 3. 
- What controller do you have? Maybe it is not supported and ScummVM goes back to the regular joystick driver.
- Is your controller in the list of supported controllers [here](https://github.com/SDL-mirror/SDL/blob/master/src/joystick/SDL_gamecontrollerdb.h)? 
- Do the other button maps work? For example, does holding down the R shoulder button slow down the mouse pointer?
- If you select a different pointer speed in the controls tab in options, does your cursor speed actually change when you move it with the left analog stick?
Concerning 2. Increasing the deadzone setting in the control panel in options should stop the drifting cursor. 
Concerning 3. this might be related to 1.

dafioram:
I have an xbox 360 wireless controller connect to the 360 receiver. All buttons that bgK showed work for me.

Once I added a vkeybd pack the vkeybd worked fine with the back button.

If there are some settings I can change to prevent the drift then I will play with those.

I don't know why the hotplugging isnt working.

bgK:
Hotplugging works using a PS3 controller in all the cases I've been able to test:
- Linux, wired
- Linux, bluetooth
- Windows, wired

Perhaps joystick indices are not being reused on your setup causing ScummVM to believe another joystick has been plugged.

dafioram:
My testing was on a Linux VM so if the hotplugging works when not in a VM then that would be ok. 
I will test that later though. I don't want to hold up this PR.

If [Trac#10366](https://bugs.scummvm.org/ticket/10366) is still an issue after the PR is merged then I will reopen the ticket with the particulars.

bgK:
@rsn8887, @dafioram: thanks a lot for your very good testing, contributions and review!

dafioram:
Would it make sense for this feature to make its way into the 2-0 branch?

csnover:
The version branches are for bug fixes only; new features go into the major releases.

PR #1090, opened 2017-12-19
PORTS: Remove the Zaurus and Yopy ports

bgK:
* Yopy: https://en.wikipedia.org/wiki/Yopy
Production ceased in 2005. There was never an official release. I can't find traces of an unofficial release.

* Zaurus: https://fr.wikipedia.org/wiki/Zaurus
Production ceased in 2007. There was never an official release. The last unofficial release of ScummVM was 0.6.0.

I'm working on improving the keymapper. They look very obsolete, I'd rather remove them than port them to the new model.

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -320,17 +320,10 @@ void SdlEventSource::SDLModToOSystemKeyFlags(SDLMod mod, Common::Event &event) {
 
 	event.kbd.flags = 0;
 
-#ifdef LINUPY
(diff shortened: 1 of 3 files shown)


csnover:
I ran across these in the past and had similar thoughts, so this seems fine to me.

digitall:
I still have a Zaurus SL-5500 and used it to play ScummVM when 0.6.0 came out, but I agree. The QTOPIA / Zaurus specific code is limited to hardcoded key mappings and should be replaced in any case.

bgK:
Thanks for your approval.

PR #1091, opened 2017-12-21
MT32: Update to munt 2.3.0

tsoliman:
This fixes a few bugs including the sq1sci SpaceAlvin instrument shown here https://youtu.be/OLvsaJ4h-VY?t=7756
The bugs are caused by Sierra devs depending on MT-32 quirks that are only present in the old MT-32.

The newer Munt release depends on a problematic source file (`SampleRateConverter`) that we intentionally ignored before because it was unused in the C interface. Now it is used in the C interface and so I have 2 approaches in this PR. This file and its dependencies need files in an include directory declared because of the way the `#includes` are used.

The first approach is to fix the compilation with adding global include directories. This keeps the munt source unmodified and maintains the ease of future drop-in upgrades. The downside is that those directories are in the includes of the other modules.

The second approach is get rid of the global includes and just fix the `#include` directives in the munt source itself. This I think is the cleaner approach but it involves editing some munt source files.

I've divided the PR commits into 3 commits for ease of review of the different approaches.
The 1st commit is just the drop-in of the new source. It doesn't build.
The 2nd commit is the first approach.
The 3rd commit is the second approach.

### DIFF
--- a/audio/softsynth/mt32/Analog.cpp
+++ b/audio/softsynth/mt32/Analog.cpp
@@ -1,5 +1,5 @@
 /* Copyright (C) 2003, 2004, 2005, 2006, 2008, 2009 Dean Beeler, Jerome Fisher
- * Copyright (C) 2011-2016 Dean Beeler, Jerome Fisher, Sergey V. Mikayev
+ * Copyright (C) 2011-2017 Dean Beeler, Jerome Fisher, Sergey V. Mikayev
  *
  *  This program is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU Lesser General Public License as published by
@@ -24,8 +24,6 @@
 
 namespace MT32Emu {
 
-#if MT32EMU_USE_FLOAT_SAMPLES
-
 /* FIR approximation of the overall impulse response of the cascade composed of the sample & hold circuit and the low pass filter
  * of the MT-32 first generation.
  * The coefficients below are found by windowing the inverse DFT of the 1024 pin frequency response converted to the minimum phase.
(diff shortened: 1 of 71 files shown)


csnover:
Between these two options I think I would personally prefer to just modify the source, and maybe check with upstream to see if there’s a particular reason why they are requiring a separate include path, since there might be a good reason for that which is relevant to us, or they might just want to change it to resolve the problem.

If we do the code-patching approach, in #1071 I used some `#ifdef` guards to clarify ScummVM-specific code changes; I’m not sure if it is the best approach versus just including a patch file in the source tree, though I think either of these approaches would be better than integrating the changes invisibly. Hopefully some other folks can chime in with additional helpful feedback or better ideas (I’m a bit idea’d-out today). I’m also planning to run with this patch locally for a bit to see if I run into any runtime issues.

tsoliman:
I've talked to sergm and he made modifications in commits leading up to https://github.com/munt/munt/commit/948b32ae
(specifically https://github.com/munt/munt/commit/9e1d24518)
Now it doesn't depend on files in `srchelper/` by default so I've excluded those.

This is the best of both worlds I think.

----

I've collapsed the whole PR into a single commit with an appropriate message https://github.com/tsoliman/scummvm/commit/ac0d0328491806dcd97148f81000da574474e59a

Sometime in January I intend to close this PR and manually push the single buildable commit since I don't like unbuildable commits in the history (because of bisect)

tsoliman:
I changed to the title to "DO NOT MERGE" but please review :)

The current state is that it is unmodified munt code with no extra global includes.

tsoliman:
munt tagged the current changes as 2.3.0 - it is basically the exact same code but with the fall-through comments changed (and a new version in config.h)

I've updated this PR with the release version.

PR #1092, opened 2017-12-22
SHERLOCK: Disable cache for uncompressed resources

AReim1982:
Sherlock Holmes - Case Of The Rose Tattoo:
Low memory devices (for example the WII) throws an "Out Of Memory" error during the intro and the gameplay. The reason is that all resources are cached in the memory. It improves the performance when resources are compressed. But the cache shouldn't be used for uncompressed resources.

### DIFF
--- a/engines/sherlock/resources.cpp
+++ b/engines/sherlock/resources.cpp
@@ -57,24 +57,23 @@ void Cache::load(const Common::String &name, Common::SeekableReadStream &stream)
 
 	int32 signature = stream.readUint32BE();
 	stream.seek(0);
-
-	// Allocate a new cache entry
-	_resources[name] = CacheEntry();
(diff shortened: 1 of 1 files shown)


dreammaster:
The change tested okay to me. There wasn't any excessive disk thrashing that would require the resources to be cached in any case. Just one thing.. to conform to our commit standards, could you modify the commit message to capitalize the word 'SHERLOCK'. Then I'll merge it in.

AReim1982:
done... ;-)

Kroc:
I should test this on PSP because the game won't start and it could be related. Is there any way to get logs from PSP builds?

bonki:
Reverted in commit afc48cfb0ec2ce3e87b7121f5616bc01f6281284 because it broke Scalpel.

PR #1096, opened 2017-12-27
Proposed Fix for PSP Aspect Ratio Preservation in 4:3

digitall:
This patch is derived from dam-soft's information on bug #10239 "PSP port incorrect 4:3 aspect radio":
https://bugs.scummvm.org/ticket/10239

This should be reviewed by the PSP porter prior to merging.

### DIFF
--- a/backends/platform/psp/display_manager.cpp
+++ b/backends/platform/psp/display_manager.cpp
@@ -55,7 +55,8 @@ uint32 __attribute__((aligned(16))) MasterGuRenderer::_displayList[2048];
 const OSystem::GraphicsMode DisplayManager::_supportedModes[] = {
 	{ "Original Resolution", "Original Resolution", ORIGINAL_RESOLUTION },
 	{ "Keep Aspect Ratio", "Keep Aspect Ratio", KEEP_ASPECT_RATIO },
-	{ "Full Screen", "Full Screen", STRETCHED_FULL_SCREEN },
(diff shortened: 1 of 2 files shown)


rsn8887:
Now the only options are "stretch to screen width" or "4:3" but there's no option anymore to use the original aspect ration.

In fact there's no point for a "stretch to screen width option" since it distorts every game.

So the old keep aspect behavior should be used when keep aspect is off instead of stretch to fill.

digitall:
@rsn8887 : To be clear, I don't own a PSP and my only intent with this was to capture the code changes from the bug report into a Pull Request for comment / modification. Your comments looks reasonable, but I have no way of testing any of these changes, so will leave this up to the PSP porter / team to amend and apply as needed. Though am not sure who the current PSP porter is, though we do have a v1.9.0 release on the downloads page so hopefully they are still around.

rsn8887:
I pushed a change. Now the user has the choice between all three original PSP graphics modes (original, keep-aspect, and stretch-to–screen) plus a new mode called "Aspect Ratio Correction" in which the aspect ratio is corrected to 4:3.

digitall:
@rsn8887 : Thanks for this code. It looks reasonable. Are you the PSP porter?

rsn8887:
I guess I am now since Joost said on IRC that he isn't maintaining PSP anymore.

PR #1098, opened 2017-12-29
NEVERHOOD: Graceful shutdown of the engine

AReim1982:
Currently the Neverhood-Engine kills the mainprocess instead of breaking the gameloop on shutdown. The WII throws an dsi error after closing the game.


AReim1982:
@bgK: you are right... so I changed it. ;-)

bgK:
Very good! Thanks for your contribution.

PR #1100, opened 2017-12-31
SCUMM: really implement lipsync

rasky:
It looks like the code was there, but it was never fully implemented
because _curSoundPos was never being incremented. Experimentally,
it looks like it works if it is a 60FPS counter.

### DIFF
--- a/engines/scumm/sound.cpp
+++ b/engines/scumm/sound.cpp
@@ -448,14 +448,15 @@ void Sound::processSfxQueues() {
 			finished = !isSoundRunning(1);
 		} else {
 			finished = !_mixer->isSoundHandleActive(*_talkChannelHandle);
+			// calculate speech sound position simulating increment at 60FPS
+			_curSoundPos = (_mixer->getSoundElapsedTime(*_talkChannelHandle) * 60) / 1000;
 		}
 
 		if ((uint) act < 0x80 && ((_vm->_game.version == 8) || (_vm->_game.version <= 7 && !_vm->_string[0].no_talk_anim))) {
 			a = _vm->derefActor(act, "processSfxQueues");
 			if (a->isInCurrentRoom()) {
-				if (isMouthSyncOff(_curSoundPos) && !_mouthSyncMode) {
-					if (!_endOfMouthSync)
-						a->runActorTalkScript(a->_talkStopFrame);
+				if (isMouthSyncOff(_curSoundPos) && _mouthSyncMode) {
+					a->runActorTalkScript(a->_talkStopFrame);
 					_mouthSyncMode = 0;
 				} else  if (isMouthSyncOff(_curSoundPos) == 0 && !_mouthSyncMode) {
 					a->runActorTalkScript(a->_talkStartFrame);


rasky:
This has been lightly tested with DOTT. This video shows lip-sync working on a specific sentence:
https://www.dropbox.com/s/xdgcpykeihlwjs2/dott_lipsync.mov?dl=0

rsn8887:
Do these lip sync tags in .sou survive re-encoding to flac (.sof) or mp3 (.so3)? I am asking because those formats are used most often on PSP, to make use of hw-accelerated mp3 decoding, and on PS Vita, to save space for its limited memory cards.

sev-:
I'll check what the original is doing here.

rasky:
@rsn8887 yes, they get copied during compression, this is the code that takes care of them:
https://github.com/scummvm/scummvm-tools/blob/d4145d6d21c09bc497155e8e39c819129282c49b/engines/scumm/compress_scumm_sou.cpp#L117

@sev- thanks; I compared the output with what I see in dosbox, but if you can disasm the original binary, that's surely better.

rsn8887:
I tested this on the Vita using DOTT (english CDRom version). This is a huge improvement. The difference really is like day and night. With this patch, character lip movement stops as it should during pauses in the sentences, and they stop moving their lips at the correct times at the end of sentences.

Without this patch, characters lip movements talk through pauses in the speech and keep talking even if their sentence is over, making the game feel like a badly dubbed kung fu movie.

The difference is easily noticeable even in the beginning of the game, by simply skipping the intro (triangle button), and then watching the lip movements during the first few lines of dialog between the three characters in the mansion lobby. For example, without this patch, Laverne moves her mouth until the point when she starts walking up. With this patch, the mouth movement stops when her sentence ends, quite a bit earlier before she turns around and starts walking.

rasky:
@rsn8887 my patch is for SCUMM v6, which is used by DOTT and Sam&Max. I checked Sam&Max (both English and Italian version) but it doesn't have lip-sync tags in the audio files... too bad.

rsn8887:
Nice! I wonder can lip-sync be fixed for Full Throttle, too?

rsn8887:
Nice work! I wonder if lip-sync needs to be fixed for Full Throttle also?

rsn8887:
I couldn't find anything wrong with it so I would say this is ready for merge.

rsn8887:
Just for the record, lip syncing should be fixed for Full Throttle, too (in a different PR), here's proof.
Full Throttle Dosbox showing lip-sync at 10m 22s: https://www.youtube.com/watch?v=TWJ6QG_59Ko#t=10m22s

Full Throttle ScummVM showing no lipsync at 10m 05s: https://www.youtube.com/watch?v=c36OmC5UbME#t=10m05s

He says "I, uh, fixed your door," and there should be no lip movement during the commas before the "uh" pronounciation. But in ScummVM there's lip movement.

bonki:
Started looking into this. Could anyone provide a FT savegame to this scene (or any other with prominent lip sync markers)?

rsn8887:
I don't have one, but it takes 2 minutes to get there. Start game press esc esc esc esc to skip all cutscenes, until you are in the first gameplay scene. Click the trash container a few times to jump out, walk right one screen, kick the door to the bar, save. The scene plays as soon as you walk into the now open door.

rsn8887:
I made a ticket ( https://bugs.scummvm.org/ticket/10407 ) to prevent cluttering up this PR too much, which is doing what it should, e.g. fixing lipsync in DOTT.

rsn8887:
I thought for a second that maybe adding 
```
_curSoundPos = (_vm->_imuseDigital->getPosInMs(kTalkSoundID) * 60) / 1000;
```
at line #446 in sound.cpp, similar to this PR, might fix Full Throttle, but nope. It seems to be not that extremely simple. But it might be quite simple. Note since getPosInMs is a private function for some reason, I had to switch it to public for my tests.

bonki:
Can you confirm that https://github.com/bonki/scummvm/commit/4cd11c3fb80fab490722928123240b93ed2fdb6c works (works for me)?
Obviously this shouldn't be merged because I messed with `getPosInMs()` which is used in other places and probably breaks other parts of iMUSE. We could just create a new public method but there already seems to be code for lip-syncing stuff in there which is used by the v8 script interpreter (i.e. MI3) so this could/should probably be fixed entirely differently in a much cleaner way.

rsn8887:
Awesome, it works perfect, thanks! I tried something similar today, but didn't think of changing that function. Maybe we could add a public function to the imuse class that simply returns the mixer handle associated with a soundID? The code in sound.cpp would then be quite clean.

rsn8887:
@Rasky: I just made a PR https://github.com/rasky/scummvm/pull/1 to your branch based on Bonki's work. That PR fixes v7 games also. It was tested with FT. Feel free to rebase and merge that PR if you agree so we can get this complete lipsync fix into one PR.

bonki:
> Why not add a public function "int32 getSoundElapsedTimeInMs(int soundId);" to imuse and leave getPosInMs() as it is?

Because it feels hacky given that there is already iMUSE code for lip syncing (that's why I didn't do it in my branch which was just meant to be a PoC). Without knowing how any of the digital iMUSE stuff works it feels as if the existing code should be fixed to support both v7 and v8 games if that's possible to avoid introducing unnecessary redundancy.
But if someone with better knowledge on SCUMM thinks this looks good I'd be happy to see it merged of course :)

rsn8887:
The existing iMuse lipsync code are just function called by the v8 kernel. I think it makes sense to implement v7 lipsync in the way here, since it works the same as v6. But a further review would be welcome.

rsn8887:
@rasky: I initially made that or to your master branch (Now closed). I opened a new one to your correct branch.

rasky:
Thanks, I'll have a look this week.

rsn8887:
@rasky: Did you ever get a chance to look at my PR to your branch? If you agree to merge it into your branch, then this PR could serve to address all lipsync fixing in one place.

sev-:
I checked with the original. This replicates the functionality, though in original this is done via timers, but same effect and same frequency, 60fps.

I'm merging this and merging v7 code too.

PR #1102, opened 2018-01-01
GUI: Rework the frame limiter to actually reach the target framerate

bgK:
Fixes the mouse cursor not drawing at regular intervals in the GUI,
and feeling laggy when moved at constant speed such as when using
a game controller.

The previous combination of a fixed 10 milliseconds delay and time since
last update checks meant that in most cases 20 milliseconds elapsed
between two calls to updateScreen resulting in a 50 fps framerate. On
systems with wait for vsync enabled that meant that some frames were
missed.

The new frame limiter waits for a variable delay equal to the non
consumed time in the slot allocated to the frame before updating
the screen.

### DIFF
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -592,8 +592,6 @@ void GuiManager::processEvent(const Common::Event &event, Dialog *const activeDi
 
 void GuiManager::doFullRedraw() {
 	_redrawStatus = kRedrawFull;
-	redraw();
-	_system->updateScreen();
 }
 
 void GuiManager::giveFocusToDialog(Dialog *dialog) {
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -208,7 +208,7 @@ bool GuiManager::loadNewTheme(Common::String id, ThemeEngine::GraphicsMode gfx, 
 void GuiManager::redraw() {
 	ThemeEngine::ShadingStyle shading;
 
-	if (_redrawStatus == kRedrawDisabled || _dialogStack.empty())
+	if (_dialogStack.empty())
 		return;
 
 	shading = (ThemeEngine::ShadingStyle)xmlEval()->getVar("Dialog." + _dialogStack.top()->_name + ".Shading", 0);
@@ -241,8 +241,11 @@ void GuiManager::redraw() {
 			break;
 
 		default:
-			return;
+			break;
 	}
+
+	// Redraw the widgets that are marked as dirty
+	_dialogStack.top()->drawWidgets();
 
 	_theme->updateScreen();
 	_redrawStatus = kRedrawDisabled;
@@ -304,8 +307,6 @@ void GuiManager::runLoop() {
 	while (!_dialogStack.empty() && activeDialog == getTopDialog() && !eventMan->shouldQuit()) {
 		uint32 frameStartTime = _system->getMillis(true);
 
-		redraw();
-
 		// Don't "tickle" the dialog until the theme has had a chance
 		// to re-allocate buffers in case of a scaler change.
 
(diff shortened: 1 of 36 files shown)


gui/gui-manager.cpp
```diff
@@ -379,8 +365,14 @@ void GuiManager::runLoop() {
 			}
 		}
 
-		// Delay for a moment
-		_system->delayMillis(10);
+		_theme->updateScreen();
+
+		// Delay until the allocated frame time is elapsed to match the target frame rate
+		uint32 frameEndTime = _system->getMillis(true);
```
csnover: `actualFrameDuration = _system->getMillis(true) - frameStartTime;` instead of storing the end time and subtracting twice.

gui/gui-manager.cpp
```diff
@@ -299,10 +299,11 @@ void GuiManager::runLoop() {
 	}
 
 	Common::EventManager *eventMan = _system->getEventManager();
-	uint32 lastRedraw = 0;
-	const uint32 waitTime = 1000 / 60;
+	const uint32 targetFrameDuration = 1000 / 60;
```
csnover: I’ve never really been clear on this, is this truly sufficiently accurate to not cause any rendering glitches even though the value is truncated?

bgK: My take on it is that it does not matter:
* If wait for vsync is enabled, the delay duration does not matter as long as it does not cross a frame boundary, updateScreen is going to block until the right time anyways. The target frame duration value being a little too small helps not waiting for too long.
* If wait for vsync is disabled ... there's going to be tearing regardless.

gui/gui-manager.cpp
bgK: done

bgK:
Further testing showed that this change induces a significant regression in responsiveness when scrolling the game list. This is due to each event causing a redraw of the list. The events are not consumed fast enough to empty the backlog and trigger a screen redraw.
I'll attempt to fix this by introducing dirty checking so that GUI elements are drawn at most once per frame.

This is now in a DO NOT MERGE state.

bgK:
The new commits implement dirty-checking for the GUI widget redraws. In locations where widgets were previously redrawn, they are now simply marked as needing to be redrawn.
At the end of the main loop, the top GUI dialog is traversed and its widgets are redrawn if needed.

This prevents widgets from being drawn needlessly multiple times per frame. The game list no longer lags on my system when it is scrolled.

This change opens the possibility to further rework the draw pipeline by removing the draw queues that make the draw calls hard to debug. Dialogs would need to be traversed twice. Once to draw the backbuffer, and a second time to draw the screen.

csnover:
My apologies for not noticing the scrolling problem earlier. I guess I should not have blown away my games list recently so it wasn’t scrolling any more. :)

This is really excellent work, thank you for working on this. The GUI runs so much better with this changeset; it is really a night and day difference. This is giving me some motivation to think about working on getting the GL shaders PR un-bitrotted and landed now that the GUI doesn’t feel bad in OpenGL. I’ll keep running with this patch for a bit to make sure I don’t run across anything else blowing up, and then try to do a more complete code review of the new changes.

Tangentially, the one place where the GUI is still noticeably lagging is when opening the Options/Edit Game panels. Have you happened to do any investigating on why the frame budget is being blown with these?

bgK:
> Tangentially, the one place where the GUI is still noticeably lagging is when opening the Options/Edit Game panels. Have you happened to do any investigating on why the frame budget is being blown with these?

The GUI seems to be doing a full (or mostly full) redraw. What stands out of the profile on my x64 system with a non optimized build is still soft shadows for rounded rectangles. The same code I messed with in https://github.com/scummvm/scummvm/commit/622adff628519c16991b4edacb7283660d1630e1. Perhaps a better approach is to pre-render the shadow in an off-screen 9-patch style surface, and blit it on the screen surface when needed. Or replace the shadow draw code with a better algorithm. I don't know.

bgK:
Thanks for your feedback. This has been merged into master.

PR #1103, opened 2018-01-01
CONFIGURE: Add a static builds switch

raziel-:
Configure will still build every target with shared objects by default.

Done:
- Added a configure switch (--enable-static) for static builds
- Right now only tested for SDL targets (checked working on AmigaOS4 with SDL1/2)

Missing:
- Add the platforms that need the switch for cross-compiling (macOS, iOS, probably more)
(I can´t do that myself as i have never cross-compiled and wouldn´t be able to check)
- Probably add a completely new "case" for static builds instead of abusing the sdl one
(Then again, it´s "just" that single change)

@csnover asked on the forums to create a PR for it, so, with my limited (non-existant is more like it) knowledge, i wrapped something up.
I *know* that this isn´t enough (and Colin is probably already running away), as there already is some kind of static builds flag in configure (_engines_built_static). As i don´t really know if this is the same or a different cause (static plugins?), i´m not touching it.

But please feel free to add or change what needs to be done (More likely close it right away)

### DIFF
--- a/configure
+++ b/configure
@@ -958,6 +958,7 @@ Game engines:
     The values of <engine name> for these options are as follows:
 $engines_help
 Optional Features:
+  --enable-static          build a static binary instead of using shared objects
   --enable-c++11           build as C++11 if the compiler allows that
   --disable-debug          disable building with debugging symbols
   --enable-Werror          treat warnings as errors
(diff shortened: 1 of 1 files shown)


sev-:
The _engines_built_static is indeed for plugins. The change is good and makes sense. Perhaps there will be additional things required for different platforms, but since this is not breaking the current behaviour, I'm merging this. Thank you!

PR #1104, opened 2018-01-01
ANDROIDSDL: config, build scripts and build instructions changes

lubomyr:


digitall:
@lubomyr : This all looks reasonable and limited to files specific to the ANDROIDSDL platform which you are supporting, so I doubt there would be any problem just merging this. OK to merge?

sev-:
Nice, merging

PR #1105, opened 2018-01-02
SYMBIAN: Build updates

fedor4ever:


digitall:
Since these changes only touch Symbian port specific code and @fedor4ever is the main porter / coder for this port, and they look reasonable and self-explanatory, I am going to merge these as-is.

PR #1107, opened 2018-01-05
SCI32: Fix PQ4 Barbie patch Selector

dafioram:
Fixes [Trac#10392.](https://bugs.scummvm.org/ticket/10392)

Previous version of patch was using a Selector that is
different  then what my disasm of showShoe::changeState
shows (I get "points" Selector instead of "flag"
Selector).

Also "flag" selector is not in selector.h, but Selector
"points" is. I think this is probably why the crash in the
report is happening.

I have played through this part and it work (no crash
and get correct points when giving barbie shoe right
after showing badge).

Here is my disasm output for showShoe::changeState of script 315

0084:0ad6: 89 9a          lsg  	9a		; 154
0084:0ad8: 35 02          ldi  	02
0084:0ada: 1e             gt?  
0084:0adb: 30 28 00       bnt  	0028  [0b06]
0084:0ade: 39 61          pushi	61		; 97, 'a', points

If m_kiewitz really saw the "flag" Selector when he did disasm on the floppy version will my patch signature (using points) not get recognized for a floppy player? As you can see we both agree that the hex value of the selector is 61.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -141,7 +141,7 @@ static const char *const selectorNameTable[] = {
 	"setScale",     // LSL6hires
 	"setScaler",    // LSL6hires
 	"readWord",     // LSL7, Phant1, Torin
-	"flag",         // PQ4
+	"points",       // PQ4
 	"select",       // PQ4
 	"handle",       // RAMA
 	"saveFilePtr",  // RAMA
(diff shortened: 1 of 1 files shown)


csnover:
Thank you for looking into this and writing a patch! Do you happen to have an English CD save game that you could send me so I can more easily reproduce this bug and test this patch? I was unable to use the German save game provided by the original reporter.

dafioram:
I have added  a saved game to the ticket.

csnover:
Thanks! Great work on researching and fixing this bug. You’re right that the correct name of the selector is “points”, not “flag”, so it broke when that got changed. This was my fault for not checking everything completely. The selector name for 0x61 is “points” in the floppy version too, so this fix is correct for all versions of the game.

PR #1109, opened 2018-01-08
AUDIO: Fixes Protracker Portamento effect

Joefish:
Mission Supernova uses the ProtrackerStream to play its intro and outro music.
For some time I noticed sound artifacts but I couldn't find the cause on my side.
So, after some digging I tracked down an overflow in the 'Portamento to note' effect.

Currently agos, gob, hopkins and parallaction are the only engines besides Supernova
that make use of a ProtrackerStream.
I have tested it only on Supernova so far but if anyone is aware of sound glitches in the
mentioned engines, it would be great if you could test it or create a bug report.

EDIT: I uploaded a before/after video [here](https://www.youtube.com/watch?v=9PsFNE_VlnM)

### DIFF
--- a/audio/mods/protracker.cpp
+++ b/audio/mods/protracker.cpp
@@ -101,15 +101,13 @@ private:
 
 	void doPorta(int track) {
 		if (_track[track].portaToNote && _track[track].portaToNoteSpeed) {
-			if (_track[track].period < _track[track].portaToNote) {
-				_track[track].period += _track[track].portaToNoteSpeed;
-				if (_track[track].period > _track[track].portaToNote)
-					_track[track].period = _track[track].portaToNote;
-			} else if (_track[track].period > _track[track].portaToNote) {
-				_track[track].period -= _track[track].portaToNoteSpeed;
-				if (_track[track].period < _track[track].portaToNote)
-					_track[track].period = _track[track].portaToNote;
-			}
+			int distance = _track[track].period - _track[track].portaToNote;
+			int sign = distance > 0 ? 1 : -1;
+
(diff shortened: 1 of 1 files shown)


criezy:
The change looks OK and should indeed ensure there is no overflow or underflow, and is otherwise functionally identical to the old code as far as I can see.

I am not however convince with the wrapping of the two cases (period smaller or greater than portaToNote) into a single one with the sign variable. To me this makes the code slightly harder to read. But that is my personal opinion and not an objection to merge.

audio/mods/protracker.cpp
```diff
@@ -101,15 +101,13 @@ class ProtrackerStream : public ::Audio::Paula {
...
-			if (_track[track].period < _track[track].portaToNote) {
-				_track[track].period += _track[track].portaToNoteSpeed;
-				if (_track[track].period > _track[track].portaToNote)
-					_track[track].period = _track[track].portaToNote;
-			} else if (_track[track].period > _track[track].portaToNote) {
-				_track[track].period -= _track[track].portaToNoteSpeed;
-				if (_track[track].period < _track[track].portaToNote)
-					_track[track].period = _track[track].portaToNote;
-			}
+			int distance = _track[track].period - _track[track].portaToNote;
```
criezy: I have a small doubt now: with period and pirtaToNote being both uint16, do we get integral promotion to int before they are substracted, or is the result computed as unsigned before being casted to the int with a potential issue in the case where portaToNote is greater than period?

Joefish: They are promoted to int before the the subtraction.
http://en.cppreference.com/w/cpp/language/operator_arithmetic#Conversions

PR #1111, opened 2018-01-13
SDL: Scale the joystick mouse speed with the vertical window size

bgK:
The mouse cursor now moves across the screen in a similar amount of time irrespective of the display resolution.


criezy:
This change seems reasonable to me.

bgK:
Thanks for your review!

PR #1112, opened 2018-01-19
IMAGE: Fix Indeo3 compiler warnings

bonki:
This gets rid of all

    warning: this statement may fall through [-Wimplicit-fallthrough=]

`gcc` warnings in `indeo3.cpp`.
I'm not familiar with the code but obviously the decoder is working and based on how the `case` statements are grouped it's obvious that these fallthroughs are intentional.

### DIFF
--- a/image/codecs/indeo3.cpp
+++ b/image/codecs/indeo3.cpp
@@ -550,6 +550,7 @@ void Indeo3Decoder::decodeChunk(byte *cur, byte *ref, int width, int height,
 										*cur_lp = FROM_LE_32(((READ_LE_UINT32(ref_lp) >> 1) + correction_lp[lp2 & 0x01][k]) << 1);
 										lp2++;
 										break;
+
 									case 1:
 										if (correction_type_sp[0][*buf1] != 1) {
 											// both correction types must be DYAD. If it's not the case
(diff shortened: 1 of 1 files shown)


dreammaster:
Thanks for the changes. Every little bit helps. However, I'm not sure that "FALLTHROUGH" is appropriate text for the comment. I'm vaguely recalling from prior discussions in the IRC channel that one of the tools being used for code analysis (maybe Coverity itself?) will only not show a fall through warning if specific comment text is used. I've used "// Intentional fall-through" as the text in the Titanic engine. Unless anyone has any better/appropriate form for the text, could you modify your commit to change the comment text accordingly?

bonki:
Fair point! According to [`gcc` documentation](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) any comment matching any of the following regular expressions silences the warning:

* `-fallthrough`
* @fallthrough@`
* `lint -fallthrough[ \t]*`
* `[ \t.!]*(ELSE,? |INTENTIONAL(LY)? )?<br>FALL(S | |-)?THR(OUGH|U)[ \t.!]*(-[^\n\r]*)?`
* `[ \t.!]*(Else,? |Intentional(ly)? )?<br>Fall((s | |-)[Tt]|t)hr(ough|u)[ \t.!]*(-[^\n\r]*)?`
* `[ \t.!]*([Ee]lse,? |[Ii]ntentional(ly)? )?<br>fall(s | |-)?thr(ough|u)[ \t.!]*(-[^\n\r]*)?`

I've checked the tree now and `// fall through` seems to be used in 90% of all cases so I guess this should be OK with other tools being used? Happy to change it to that if given the go.

dreammaster:
Good, it should be. I'd recommend waiting a few days for any other devs to chime in, and if not, if you could change it accordingly, and then I'll merge it in.

bonki:
Updated.

PR #1113, opened 2018-01-19
Documentation support!

MindfulMe:
Hey, I want to make more, since I am a student I want to join your team in further as a participant of GSoC 2018!

### DIFF
--- a/README
+++ b/README
@@ -16,7 +16,7 @@ Table of Contents:
  * 2.1 Reporting Bugs
 3.0) Supported Games
  * 3.1 Copy Protection
- * 3.2 Datafiles
+ * 3.2 Data files
  * 3.3 Multi-CD games notes
  * 3.4 Known Problems
  * 3.5 Beneath a Steel Sky notes
@@ -119,7 +119,7 @@ continually improving, so check back often.
 Among the systems on which you can play those games are regular desktop
 computers (running Windows, Linux, Mac OS X, ...), game consoles
 (Dreamcast, Nintendo DS & Wii, PS2, PSP, ...), smartphones (Android,
-iPhone, PocketPC, Symbian ...) and more.
+iPhone, PocketPC, Symbian ...) and even more.
 
 At this time ScummVM is still under heavy development. Be aware that
 whilst we attempt to make sure that many games can be completed with few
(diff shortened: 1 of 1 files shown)


criezy:
Thank you for the suggestions. The changes look good to me, but if you change `Quit start` to `Quickstart` in the body of the document, you also need to change it in the table of contents at the start of the document for consistency.

Also the commit message does not conform to our conventions. See http://wiki.scummvm.org/index.php/Commit_Guidelines (the subsystem here would be README for example).

README
```diff
@@ -554,12 +554,12 @@ ScummVM will skip copy protection in the following games:
...
 
-3.2) Datafiles:
+3.2) Data files:
 ---- ----------
-For a comprehensive list of required Datafiles for supported games
+For a comprehensive list of required Data files for supported games
 visit:
 
-  <http://wiki.scummvm.org/index.php/Datafiles>
+  <http://wiki.scummvm.org/index.php/Data files>
```
digitall: You should not add a space here as it will break the URL link unless the Wiki is updated as well.

I am not sure about the Team POV, but personally I would leave this as "datafiles" in all cases as these kind of compound nouns are common in English and while I agree that "data file" is more orthographically accurate, the term "datafile" or "data-file" are at least as common in general usage... plus changing this in all cases could prove to be a huge headache for little improvement.

digitall:
@criezy : I think you mean "Quickstart" to "Quick Start" ... Personally, I would refrain from breaking these two compound noun cases at all as this is perfectly acceptable in English usage and quite common, as per my previous line comment.

digitall:
@MindfulMe : If this is merged, I would suggest using rebase to squash both of your commits together into a single commit, you can use this to amend the commit message as well as per criezy's comment.

Good luck on your GSoC 2018 application and if you want to ask any questions to the developers / mentors, I suggest trying our IRC channel http://wiki.scummvm.org/index.php/IRC_Channel

MindfulMe:
Thank you guys  😸

criezy:
Than you again for your suggestions, and for considering applying for GSoC with us. In the end I have decided to reject those changes, except for the comma after `Actually` which I committed separately.

Also I have clarified the rules regarding GSoC applications (on http://wiki.scummvm.org/index.php/Summer_of_Code/Project_Rules) to explicitly indicate that we expect more than changes to the documentation for the patch submission as part of an application. Indeed we can't really be sure you can compile or understand the source code of the software if all you modify is the documentation ;-)

You are welcome to come and discuss with us on IRC (#scummvm channel on freenode) if you needs idea of a small task to work on. And it would be a good idea to discuss your application with us before the end of the application period anyway so that we can make suggestions to improve it.

PR #1114, opened 2018-01-25
AMIGA: Small rewrite of the readme to amigaguide converter scripts

raziel-:
Due to GitHub´s utf-8 converter i´m no longer able to use special chars like "°".
GitHub will add a char in front of it, rendering the following script useless.

(Not sure though if GitHub is the culprit or the sgit program i´m using for fetching or even the font {display} program on my platform)

I changed that special char to another marker which will use "<>" now instead.
This will make updating and keeping tracks of those files easier for me.


raziel-:
Thank you @digitall

PR #1117, opened 2018-01-26
TUCKER: Fix color cycling in submarine room

bonki:
Fixes Trac#6378.


sev-:
This, in fact, matches the original. The original was writing directly to 0x3C8 and 0x3C9 VGA ports which are 6-bit.

PR #1118, opened 2018-01-26
TUCKER: Fix walking Royston glitch

bonki:
Fixes Trac#6640.


bonki:
This happens because `xPos` is based on `s->_gfxBackgroundOffset` which *after* writing Royston's sprite to the buffer gets updated again by `s->_backgroundOffset`. In other words the sprite's location in the buffer and the dirty rect used to update the screen differ by `s->_backgroundOffset`, i.e. 1px. `_backgroundOffset` is used to let Royston walk across the room and used nowhere else in the code.

sev-:
Merging, thanks.

Looks like this could be also bug in the original

PR #1119, opened 2018-01-26
TUCKER: Fix bubbles in Plugs shop

bonki:
Fixes Trac#6641.

/cc @raziel- (you reported the bug)


sev-:
This now matches the original logic, thanks for fixing.

PR #1120, opened 2018-01-26
TUCKER: Cure Bud's and Billy's seasickness

bonki:
This ensures that the calculated offsets for Bud and Billy are [0,2] px
and follow the sequence 0px -> 1px -> 2px -> 1px -> 0px as opposed to
the earlier 0px -> 1px -> 2px -> 3px -> 0px which resulted in them
teleporting back to their original location and on top of that being
out of sync with the boat.

Fixes Trac#6643.

/cc @raziel- (you reported this bug)


sev-:
Thanks, merging

PR #1121, opened 2018-01-26
TUCKER: Fix alleyway cellar location hotspot bug

bonki:
Fixes Trac#6645.

I'm pretty confident about this one but this should be tested a bit more in other places.

/cc @raziel- (you reported this bug)


sev-:
Current code matches the original. Maybe a more thorough testing could be helpful.

bonki:
Thanks for checking, I'll look into it.

sev-:
Merging in the meantime.

bonki:
Just FYI, I checked a while ago and the bug does exist in the original so it's not surprising current code matches disassembly. I still believe this is the right fix and I plan on testing the whole game before next release so I'm OK with you having merged this now. Thanks!

PR #1122, opened 2018-01-26
TUCKER: Fix dentist music in mall being played incorrectly

bonki:
Sort of fixes Trac#6371 which is only half a bug.


sev-:
This now matches the original. Thank you!

PR #1123, opened 2018-01-26
TUCKER: Use existing enum for verbs

bonki:


sev-:
Thanks, it is a good improvement.

PR #1124, opened 2018-01-27
GUI: Remove the ThemeItem draw queues

bgK:
Pull request #1102 added dirty checking as a level of indirection when drawing
the GUI. This PR removes a previously existing level of indirection, the draw
queues. With this, we are back to one level only. This makes it easier to debug why
a particular low level draw method was called, by inspecting the call stack.

Draw calls now happen directly instead of being queued to the backbuffer or
screen queues. The background and foreground dialog layers are now drawn
to the appropriate surface by doing a rendering pass for each layer.

The goal here is only code cleanup. There should be no significant performance
changes.


sev-:
This looks all great to me. Good code simplification. Merging this, and will do extensive testing, so please prepare for the bug squashing.

PR #1125, opened 2018-01-27
TUCKER: Fix dirty rects drawing bug in drawStringInteger()

bonki:
This fixes a bug in `drawStringInteger()` (used for drawing savegame slots in the original game UI) where the x-coordinate for the dirty rect was wrongly calculated.


sev-:
Thanks! Merging

PR #1126, opened 2018-01-27
TUCKER: Fix sentence bar offsets

bonki:
Fixes Trac#10413 and Trac#10414.

### DIFF
--- a/engines/tucker/tucker.cpp
+++ b/engines/tucker/tucker.cpp
@@ -1971,9 +1971,9 @@ void TuckerEngine::clearItemsGfx() {
 }
 
 void TuckerEngine::drawPausedInfoBar() {
-	int len = getStringWidth(36, _infoBarBuf);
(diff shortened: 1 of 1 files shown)


bonki:
Note that in the original interpreter the pause information (and probably the hints as well, but I haven't checked) are not horizontally centered but off by 6 px. This is why in our initial implementation (which used to work at some point) there's an offset of 326 = 320 + 6 - this shifted the localized "PAUSED" text one pixel down and six pixels to the right. Since English text was never vertically centered in our implementation this rendered at least the pause information at the correct location.
However, in this fix I both vertically and horizontally center all text (sans the 6px offset to the right) since I see no reason to replicate this, it just looks ugly and wrong.

sev-:
Yes, the original had this hardcoded at least in English localization. This approach is better. Thank you, merging.

PR #1127, opened 2018-01-30
TUCKER: Print infobar string if no savegame exists

bonki:
Before, the ingame load dialog pretended to allow loading savegames from all possible slots even if no savegames existed.
This introduces the original interpreter's behavior which instead loads a resource string from `infobar.txt` informing the user that no savegames are available.

Fixes Trac#10417.


sev-:
Very good, thank you! Merging.

PR #1128, opened 2018-01-30
Build system updates for Buildbot

csnover:
This PR contains changes to work with the new Buildbot service,
which is currently at <https://github.com/csnover/scummvm-buildbot/>,
and to modernise small portions of the build system.

More information about the new Buildbot service can be found in the 
[http://lists.scummvm.org/pipermail/scummvm-devel/2017-December/012041.html](mailing list post).

Here is a summary of the most notable changes:

* Common & engine code using non-standard C/POSIX/GNU APIs which
  do not exist in standard C++, like strdup and sqrtf, have been
  updated to replace these non-standard APIs with standard ones.
  This is necessary because newer platform compilers with stricter
  ANSI C++ compliance fail to compile this code.
* The default optimisation modes have been improved. With
  optimisations enabled, the default is -Os with auto-vectorisation,
  and without, -Og. Feature detection is used to fall back to older
  optimisation modes -O2 and -O0 if these options are not
  available.
* split-dwarf and Gold linker are now used where available to try
  to improve linker performance.
* Platform-specific default optimisation mode overrides have been
  removed everywhere since the new defaults should be appropriate
  for all platforms[1]. `--enable-optimizations` has been enhanced
  to accept an optional flag, so users can choose to force an
  optimisation mode that way if needed.
* Simple auto-detection for compiler standards support up to C++14
  has been added, so the latest supported standard C++ version is
  now used by default instead of defaulting to C++98/03 and
  requiring users to opt in to the newer standards modes. The
  `--enable-c++11` flag is now a no-op and `--force-c++98` can be
  used to force compilation using the legacy standard.
* An `--enable-static-libs` flag is added to use `--static-libs`
  instead of `--libs` when collecting linker information from
  third-party scripts. This is needed for macOS/iOS. Eventually
  this should be expanded to actually link libraries statically
  when enabled, so static builds can be generated even in build
  environments which also have shared libraries, but this is not
  needed right now for Buildbot so is not fully implemented. (This
  would be most useful for the Linux builds, which right now will
  only work against particular versions of Debian due to dynamic
  linking.)
* Various port compilation breakages due to using deprecated,
  removed, and non-standard system/library APIs have been fixed to
  use the correct APIs.
* The build system now respects all standard environment variable
  compiler overrides (e.g. STRINGS, STRIP, etc.).
* Forced overrides of features have been removed for some platforms
  in lieu of the normal feature detection sytem.
* Some new warnings emitted by newer compilers have been fixed.
* Obsolete iOS 3 build code is gone, as are special targets for
  macOS static builds. (The macOS static builds situation is still
  not ideal, but it does function well enough in conjunction with
  the Buildbot image to generate working macOS builds.)

The engines with code changes are AGI, Bladerunner, Director, Gob,
Lure, Parallaction, SCUMM, Sword2. I did some basic testing of all
the engines except Bladerunner to try to check that these code
changes haven’t broken anything obviously, but as always
additional review/testing on these changes is welcome/encouraged.

The ports with changes are 3DS, AmigaOS, Android, Caanoo, Dingux,
Dreamcast, FreeMiNT, GameCube, GCW0, iOS, macOS, Maemo, Open
Pandora, PSP, Raspberry Pi, RISC OS, Wii, Win32, WinCE. (Most of
these changes are just the optimisation mode override removals.)

Builds from the new Buildbot were tested over a couple-week period
a few weeks ago to verify that they do at least run on these
platforms, which cover >98% of our users by downloads:

* Android
* Debian
* macOS
* Maemo
* PS3
* PSP
* Raspberry Pi
* Vita[1]
* Wii
* Windows

I will leave it up to others to test/fix any of the other ports.

[1]: Vita doesn’t have enough memory to load a binary with all
     engines statically compiled without -Os, so the default
     non-optimised build doesn’t work. rsn8887 is aware of this and
     is aware that the Vita port needs dynamic plugin support (since
     eventually -Os won’t be able to remove enough code to fit in
     memory either). For consistency, and to ensure we are getting
     more exercise of the dynamic plugins system, we should probably
     start enabling plugins by default at this point for all
     platforms, but that can be changed & discussed separately.


criezy:
I only took a very quick look, focussing on the macOS build. I will take a better look in a couple of weeks (I have little time to do it before I will be traveling).

I suppose the `--enable-static-libs` replaces the use of the `scummvm-static` target we had on macOS? If so is there a mechanism in place to ensure that this is used when we do a bundle (to avoid having to remember one more option that has to be used to generate a version we want to share)?

And I might be wrong but to me it looks like something is missing to generate a static build of scummvm. I didn't see how the changes in configure fully replace the scummvm-static target. I am mostly noting this here as a remember to myself to check this when I get more time.

I also see the osxsnap target has been removed. Is there something else in place to generate a macOS package easy to distribute? The parts where it sets the extended attributes on the UTF-8 files is useful for example as otherwise they are not displayed correctly when you open those on some versions of macOS. If we have to do all that manually this is prone to errors.

By the way if you remove the osxsnap target, then `dists/macosx/DS_Store` and `dists/macosx/background.jpg` should probably be removed as well since they are only used for the disk image generated by osxsnap.

I also see you removed the packaging for the old iPhone backend. Is there actually a value in keeping this old backend around or should the ios7 backend be a requirement now? If we want the old backend to stay around, shouldn't we also keep the packaging part of it? Or was there a good reason to remove it?

csnover:
Thanks for your feedback, these are great questions and definitely show some areas which could be improved.

> I suppose the --enable-static-libs replaces the use of the scummvm-static target we had on macOS? If so is there a mechanism in place to ensure that this is used when we do a bundle (to avoid having to remember one more option that has to be used to generate a version we want to share)?

That is an ultimate goal; this PR is a little short of it though. Right now a regular build links statically only because the Buildbot image has no shared libraries so the linker is forced to do it that way. More work would be needed to get this to work correctly everywhere since the Apple linker *really* doesn’t want developers to link anything statically.

The static targets went away because libraries were hand-coded for the alternative target, so were fragile and got out-of-sync easily with the main LIBS list generated by `configure`, and it didn’t work well for generating automated builds from Buildbot because it performed a full recompilation step instead of using the scummvm target. My thinking was that for users, this isn’t an issue since users don’t typically need to create static builds, and for us it is now not an issue since the Buildbot worker can generate these static builds now. Obviously, I forgot about the PPC builds again, so sorry about that oversight and thanks for pointing it out. It can be restored for now for those if there isn’t a better alternative. (I don’t have the energy to make this do the right thing and simply filter the `-l` flags into absolute paths to `.a` files, and would be happy for someone else to make that change for when the `--enable-static-libs` flag is on.)

> I also see the osxsnap target has been removed. Is there something else in place to generate a macOS package easy to distribute? The parts where it sets the extended attributes on the UTF-8 files is useful for example as otherwise they are not displayed correctly when you open those on some versions of macOS. If we have to do all that manually this is prone to errors.

This is a good question. I honestly didn’t give a ton of thought to it. It seemed to me at the time that giving a zip file with the app bundle was sufficient, since this is how nearly all of the non-Mac-store apps I download these days are distributed. I haven’t seen apps distributed with extra files in the DMG like this quite some time, and the limitations of such an approach are fairly substantial (i.e. these extra files are only available so long as the installer image is still on the computer and mounted). I think we’d be much better off spending time adding a Help menu which links to the user manual instead, since that would give users access to this information at any time after installation, in the officially blessed OS-standard manner, and would eliminate the need for extra DMG stuff just for docs. This would be needed if we were ever to publish to the Mac App Store anyway. What do you think about doing this instead of trying to restore DMG packaging?

> I also see you removed the packaging for the old iPhone backend. Is there actually a value in keeping this old backend around or should the ios7 backend be a requirement now? If we want the old backend to stay around, shouldn't we also keep the packaging part of it? Or was there a good reason to remove it?

IIRC, I thought that the packaging for that was just broken due to the way the macOS stuff was commingled with it, but I may have been just confused. In any case I don’t think there is any reason to retain that old backend, there haven’t been any downloads (literally zero non-bot downloads) for any iPhoneOS packages in a year and devices going all the way back to iPhone 4 should work with the newer iOS backend according to [iOS support matrix](http://iossupportmatrix.com/). So I guess not keeping any of it around any more makes sense to me.

criezy:
> The static targets went away because [...] and it didn’t work well for generating automated builds from Buildbot because it performed a full recompilation step instead of using the scummvm target.

That is not actually correct. The static target does not do a full recompilation. If all the objects are already compiled then it only does one link to generate the statically-linked executable.
Your other point is correct though and it would indeed be better not to duplicate the list of libraries.

> I haven’t seen apps distributed with extra files in the DMG like this quite some time

The majority of the dmg images indeed come with just the app bundle and a shortcut to the Applications folder (which we don't even have). It is not uncommon though to also have a ReadMe file.

> I think we’d be much better off spending time adding a Help menu which links to the user manual

That is something that has been on my list for a while actually. But even assuming the user manual on the wiki has all the content currently in the ReadMe file, it would not replace the distribution of the NEWS file and translated QuickStart/ReadMe (unless they are all moved to the wiki as well). Although those could be included inside the app Bundle and accessed from the Help menu (I have just added that idea to my list). And then there is the licenses (GPL, LGPL, BSD and FREEFONT), copyright, and author files. Is that fine to also have those hidden the bundle with access from the menu?

csnover:
> That is not actually correct.

You’re right, I only remembered that it wasn’t using the `all` or `scummvm` target as a dependency. Thanks for pointing that out.

> That is something that has been on my list for a while actually. But even assuming the user manual on the wiki has all the content currently in the ReadMe file, it would not replace the distribution of the NEWS file and translated QuickStart/ReadMe (unless they are all moved to the wiki as well). Although those could be included inside the app Bundle and accessed from the Help menu (I have just added that idea to my list). And then there is the licenses (GPL, LGPL, BSD and FREEFONT), copyright, and author files. Is that fine to also have those hidden the bundle with access from the menu?

Until we can all sit down and figure out how to get to a single documentation source, it’s completely reasonable to me to have the Help menu items just open the same documentation files, which would now live in the bundle. As far as licensing goes, there is no prescribed way in which licences must be made visible, the only requirement is that they must be included with the program, so a link on the Help menu is fine. If this all sounds OK, I can work on making this change, unless you’d like to do it first. Let me know. (As an aside, I’d also really like the About 
window to not be an auto-scroller; it gives all this license and author information, but is unusable.)

configure
```diff
@@ -3642,28 +3649,41 @@ echo_n "Checking whether to have a verbose build... "
...
+		echo_n -- "$_optimization_level (forced)"
+	fi
+	echo ""
 	append_var CXXFLAGS "$_optimization_level"
+
+	# Enabling optimizations also makes it possible to use -Wuninitialized, so
+	# let's do that.
 	append_var CXXFLAGS "-Wuninitialized"
+elif test "$_optimizations" != no && cc_check_no_clean -Og; then
+	echo "-Og"
```
bgK: Are you sure about setting `-Og` by default ? It's good for us for building faster debug builds, but won't linux packages maintainers hate it? For example the Gentoo people like setting very specific compiler flags via environment variables.

configure
```diff
@@ -1272,10 +1292,9 @@ for ac_option in $@; do
 		_backend=`echo $ac_option | cut -d '=' -f 2`
 		;;
 	--enable-c++11)
-		_use_cxx11=yes
 		;;
-	--disable-c++11)
-		_use_cxx11=no
+	--disable-c++11|--force-c++98)
```
bgK: I'm not sure I like the idea of adding an option to `configure` to support a C++ standard version that will likely no longer be accepted in the near future. 
What do you think about adding something like `--with-c++-dialect=c++98` instead?

bgK: No sure it's better actually because then we need to add checks for the allowed version...

configure
csnover: I’m not very confident. I’ve waffled back and forth on the best non-release optimisation strategy, and on whether or not Buildbot should just always generate fully optimised builds since users today use these builds as rolling releases. It seemed like *something* needed to be done differently since `-O0` isn’t good enough for Munt, and `-Og` seemed like it could be a decent compromise between runtime speed, compilation speed, and avoiding symbols optimised away. As a result of my doubt, I added the user override so it is possible to force the strategy with e.g. `--enable-optimizations=-O0` if needed.

configure
csnover: I don’t have any strong preference either way. I guess the arguments for keeping it this way instead of using something like `--with-c++-dialect=` are that (1) it has a clearly identifiable end-of-useful-life where it can be removed, and (2) it doesn’t offer any easy avenue for users to attempt use of incompatible language standards.

configure
bgK: Sorry, my comment was not clear enough. `-Og` seems like a good compromise to me for debug builds.
What I'm not sure about is if configure should set any optimization flags by default. Linux packaging tools usually set the optimisation flags in the environment themselves. If we override that with `-Og`, people will expect to have release builds but instead get debug-optimized builds.

bonki: I've looked at the ScummVM Gentoo ebuild and it doesn't seem to use the system's `C[XX]FLAGS` specified in `/etc/make.conf` right now. Usually, it is expected that ebuilds honor the environment variables to an extent that makes sense for the given package. When upstream doesn't use `autoconf` and "unnecessarily" enforces or hardcodes certain compiler or linker flags Gentoo usually strips these away (either on-the-fly as part of the ebuild or via a separate patch file). If we don't have this at the moment, we should allow overriding `C[XX]FLAGS` via the commandline. As long as that is possible I see no problem defaulting to `-Og`.

bonki: Let me clarify: I see no problem defaulting to `-Og` for *debug builds*, just as bgK says.

csnover: OK, I understand now, thank you for clarifying. So is the best approach then to not do this at all without an opt-in, not do this if `CXXFLAGS` is set, try to test for an `-O` flag in `CXXFLAGS` and not do it if one is found, something else?

bgK: I'm very unsure what's the best option. We seem to enable debug builds by default. Perhaps when `--disable-debug` is passed to configure then `-Og` should not be set. That way it's possible not to have optimization flags without editing configure.

configure
bgK: Ok, sounds good.

configure
bgK: Oh, but it's also possible to pass `--disable-optimizations` to disable `-Og`. Then I guess it's good.

bonki: Sounds good.

configure
bonki: Personally I'd prefer `--with-c++-dialect=foo` but I'm fine with either one.

rootfather:
Any updates on this PR?

csnover:
I added the Help menu for macOS. It doesn’t include selection of the appropriate internationalised readme file at the moment; I added a [new ticket](https://bugs.scummvm.org/ticket/10464#ticket) for tracking that separately with some information, since I will not be able to do that work.

The new Help menu includes a link to the user manual which does not exist currently, https://www.scummvm.org/manual, which should just redirect to the wiki. I did it this way because that link should always work into the future regardless of where the user manual might end up, which may not be the case if the manual stops being hosted on a separate subdomain or stops being a MediaWiki or any other eventuality.

To avoid having to have 3 different menu items for licenses, and to stop any future license violations in ports caused by not including all the licenses, I also collapsed all the license files into a single COPYING file. The other ports’ file lists should be adjusted appropriately.

I added code in `configure` to make the `--enable-static-libs` flag convert `-lfoo` in `$LIBS` to the appropriate absolute library on Darwin, so that should make your PPC static builds OK again without needing the extra targets (and anyone else who really wants to do a static build from macOS for some reason).

So, I think that is all the critical feedback addressed. Let me know if not.

I would like to be able to land this now so this can be finally done, but I never received the access that I requested to move the other repositories into the organisation and to make sure it is possible to recover if there is a contingency during the required server updates. As such, there is nothing left for me to do here, and someone else with access will need to do that I guess.

ports.mk
```diff
@@ -277,167 +259,14 @@ endif
...
-iphone: $(OBJS)
-	$(CXX) $(LDFLAGS) -o scummvm $(OBJS) \
-		$(OSX_STATIC_LIBS) \
-		-framework UIKit -framework CoreGraphics -framework OpenGLES \
-		-framework CoreFoundation -framework QuartzCore -framework Foundation \
-		-framework AudioToolbox -framework CoreAudio -lobjc -lz
-
-# Special target to create a snapshot disk image for Mac OS X
-# TODO: Replace AUTHORS by Credits.rtf
-osxsnap: bundle
```
sev-: I am also not sure why we are suddenly dropping this.

ports.mk
```diff
@@ -450,7 +279,7 @@ scummvmwinres.o: $(srcdir)/icons/scummvm.ico $(DIST_FILES_THEMES) $(DIST_FILES_N
 	$(QUIET_WINDRES)$(WINDRES) -DHAVE_CONFIG_H $(WINDRESFLAGS) $(DEFINES) -I. -I$(srcdir) $(srcdir)/dists/scummvm.rc scummvmwinres.o
 
 # Special target to create a win32 snapshot binary (for Inno Setup)
-win32dist: $(EXECUTABLE)
+win32dist: all
```
sev-: Does it mean that it builds the devtools too?

PR #1129, opened 2018-01-31
Fix broken compilation on up-to-date msys2/mingw-w64

csnover:
It’s questionable that the SDL backend code is trying so hard to
maintain forbidden symbols at all since backend code is exactly
where such things are allowed, and it is a game of whack-a-mole
to keep system API changes from breaking this fragile system of
symbol redefinitions. Probably this should all just get replaced
with a FORBIDDEN_SYMBOL_ALLOW_ALL, but for the time being this at
least gets builds working again on up-to-date msys2/mingw-w64
compilers.

Fixes Trac#10405.

### DIFF
--- a/backends/platform/sdl/sdl-sys.h
+++ b/backends/platform/sdl/sdl-sys.h
@@ -31,15 +31,8 @@
 
 #include "common/scummsys.h"
 
-// Remove FILE override from common/forbidden.h, and replace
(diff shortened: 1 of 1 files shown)


sev-:
Makes sense to me. Originally it was written to catch SDL.h FILE usage.

sev-:
merging.

PR #1130, opened 2018-01-31
NEWS: Mention touch controls in PS Vita port

rsn8887:
I have never done News updates, and am not even sure if I should. So I am doing this as a PR just to be safe.


sev-:
We enlist ports before the engines. I'll merge this and move, so the history is kept.

sev-:
Ah, no, I'm wrong. This is the correct order. See release 1.3.0, for example.

PR #1132, opened 2018-02-02
Allow games to specify their pixel aspect ratio - DO NOT MERGE

criezy:
**DO NOT MERGE**: This pull request is incomplete and is here to ask for feedback.

The way the aspect ratio correction works in ScummVM is currently too restrictive for some games. Backends hardcode a 6:5 correction designed to go from 320x200 to 320x240 or 640x400 to 640x480. When aspect ratio correction is on, the SDL backends (both surface and OpenGL) for example systematically apply this 6:5 correction when the game original resolution is either 320x200 or 640x400 and do no correction otherwise.

Here are some cases where it fails:
  - Dreamweb, despite having a resolution of 320x200, was designed with square pixels in mind (as can be seen with the circle  in the introduction, or with the rotating fan in Eden's bedroom).
  - ADL games with an original resolution of 560x384 need to be corrected to 560x420, which means a 35:32 correction (see conversation starting at 07:47 at http://logs.scummvm.org/log.php?log=scummvm.log.21Jun2017).
  - Some mac SCI games have a native resolution of 320x190 and apparently need to be corrected to 320x228 (see http://forums.scummvm.org/viewtopic.php?t=14303)

The approach that I have taken here is to modify Graphics::Mode to contain a pixel aspect ratio in addition to the width and height of the game. I then modified `OSystem::initSize(int width, int height, const Graphics::PixelFormat *format = NULL)` to `OSystem::initSize(const Graphics::Mode &mode, const Graphics::PixelFormat *format = NULL)` and adapted the OpenGL and SurfaceSdl graphics backend as well as the dreamweb and ADL engines.

All other backends are broken since I did not report the change in the OSystem API to them yet.

Another limitation of this pull request is that the SurfaceSdl backend only supports 1:1 or 6:5 corrections. So any other correction is ignored. To test those currently you have to use the OpenGL mode.

Note that in theory this approach allows the backend to choose the correction based on the given game pixel aspect ratio and the current display device (i.e screen) pixel aspect ratio. In the changes I made the backends assume the display uses square pixel when doing the correction. 

At this points this allows to see that the approach works. Before I continue further I would like some feedback:
- Does this approach look good to you? Do you have a better suggestion?
- Currently the aspect ratio is stored as a frac_t, which has some precision issue. See commit 109b772. Should Common::Rational be used instead?
- At some point while working on this I got confused between `OSystem::GraphicsMode` and `Graphics::Mode`. Maybe the later should be renamed?
- The aspect correction code for the SurfaceSdl backend, which is in [graphics/scaler/aspect.cpp](https://github.com/scummvm/scummvm/blob/master/graphics/scaler/aspect.cpp) assumes a 6:5 correction. Rewriting this code seems like a big task. Maybe we could let SDL2 do the scaling (I think it can do it? And it might even use hardware acceleration?) ? But then what of SDL 1? Can the support be dropped? Can we just leave it with the current code and no correction other than 6:5? Do you have any other suggestion?

### DIFF
--- a/backends/graphics/graphics.h
+++ b/backends/graphics/graphics.h
@@ -58,7 +58,7 @@ public:
 	virtual Graphics::PixelFormat getScreenFormat() const = 0;
 	virtual Common::List<Graphics::PixelFormat> getSupportedFormats() const = 0;
 #endif
-	virtual void initSize(uint width, uint height, const Graphics::PixelFormat *format = NULL) = 0;
+	virtual void initSize(const Graphics::Mode &mode, const Graphics::PixelFormat *format = NULL) = 0;
 	virtual void initSizeHint(const Graphics::ModeList &modes) {}
 	virtual int getScreenChangeID() const = 0;
 
--- a/backends/graphics/graphics.h
+++ b/backends/graphics/graphics.h
@@ -27,7 +27,7 @@
 #include "common/noncopyable.h"
 #include "common/keyboard.h"
 
-#include "graphics/mode.h"
+#include "graphics/video_mode.h"
 #include "graphics/palette.h"
 
 /**
@@ -58,8 +58,8 @@ public:
 	virtual Graphics::PixelFormat getScreenFormat() const = 0;
 	virtual Common::List<Graphics::PixelFormat> getSupportedFormats() const = 0;
 #endif
-	virtual void initSize(const Graphics::Mode &mode, const Graphics::PixelFormat *format = NULL) = 0;
-	virtual void initSizeHint(const Graphics::ModeList &modes) {}
+	virtual void initSize(const Graphics::VideoMode &mode, const Graphics::PixelFormat *format = NULL) = 0;
+	virtual void initSizeHint(const Graphics::VideoModeList &modes) {}
 	virtual int getScreenChangeID() const = 0;
 
 	virtual void beginGFXTransaction() = 0;
--- a/backends/graphics/null/null-graphics.h
+++ b/backends/graphics/null/null-graphics.h
@@ -48,7 +48,7 @@ public:
 		list.push_back(Graphics::PixelFormat::createFormatCLUT8());
 		return list;
 	}
-	void initSize(const Graphics::Mode &mode, const Graphics::PixelFormat *format = NULL) override {}
+	void initSize(const Graphics::VideoMode &mode, const Graphics::PixelFormat *format = NULL) override {}
 	virtual int getScreenChangeID() const override { return 0; }
 
 	void beginGFXTransaction() override {}
--- a/backends/graphics/null/null-graphics.h
+++ b/backends/graphics/null/null-graphics.h
@@ -48,7 +48,7 @@ public:
 		list.push_back(Graphics::PixelFormat::createFormatCLUT8());
 		return list;
 	}
-	void initSize(uint width, uint height, const Graphics::PixelFormat *format = NULL) override {}
+	void initSize(const Graphics::Mode &mode, const Graphics::PixelFormat *format = NULL) override {}
 	virtual int getScreenChangeID() const override { return 0; }
 
 	void beginGFXTransaction() override {}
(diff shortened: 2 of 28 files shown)


csnover:
> Here are some cases where it fails: […]

Do you have a more complete list of cases anywhere? I believe NES Maniac Mansion also belongs on this list since it is supposedly around 8:7 PAR, and @rsn8887 mentioned that at least some of Monkey Island 1’s assets appeared to be designed for square pixels. Maybe [this ticket](https://bugs.scummvm.org/ticket/10376) can be repurposed for this AR work if you don’t have one already since it discusses this a bit.

> Does this approach look good to you? Do you have a better suggestion?

So far the changes are in line with what I was expecting when `Graphics::Mode` was introduced, so seems good to me so far.

The dimensions-only constructor is convenient, and after thinking about it, I’m still leaning toward wanting to always have engines give a PAR instead of having a “guess the PAR from the dimensions” magic. Adding some predefined constants for the common PARs would be helpful so developers have an easier time picking the right one. I’ve also been wondering whether a helper function to generate the PAR from a width + height + display aspect ratio might be convenient (or maybe it just ends up being confusing and unnecessary, I’m not sure).

> Currently the aspect ratio is stored as a frac_t, which has some precision issue. See commit 109b772. Should Common::Rational be used instead?

Probably so. In absence of a specific performance problem requiring reduced precision, it makes more sense to me to default to `Common::Rational`. I think I only continued to use `frac_t` in some of the graphics code because it was there before, but my memory is not great right now so I could be making that up. The reasons against using `frac_t` are certainly not just with precision, it is also not type safe since it will implicitly convert to `int` which makes it way too easy to introduce programming errors by applying wrong operations to a `frac_t`. (If `frac_t` needs to continue to exist at all it should probably be turned into a proper class so it isn’t possible to misuse in this way, but that’s out of scope of this review.)

> At some point while working on this I got confused between OSystem::GraphicsMode and Graphics::Mode. Maybe the later should be renamed?

I don’t have a good sense of how to address this. While theoretically generic, it seems like the OSystem “graphics mode” is used in practice to define different ways of scaling the video output, so maybe _it_ should have a different name using a word like “scaler” or “filter” or “effect” or “shader” instead. I’m not sure how to successfully avoid confusion by changing `Graphics::Mode`, the only other names I can think of are similarly ambiguous (like VideoMode). I am open to any idea in this regard.

> The aspect correction code for the SurfaceSdl backend, which is in graphics/scaler/aspect.cpp assumes a 6:5 correction. Rewriting this code seems like a big task. Maybe we could let SDL2 do the scaling (I think it can do it? And it might even use hardware acceleration?) ?

SDL2’s own scaling does seem preferable for the final pass for the reasons you’ve mentioned. There is also some arbitrary scaling code in TransparentSurface which I’ve wanted moved out for some time now which could also be used if needed. I had some trouble when trying to use the SDL scaling code for 32bpp cursors (IIRC 2.0.4 on Windows would not render anything), though that trouble may have been also from the broken RLE optimisation since it happened around the same time.

> But then what of SDL 1? Can the support be dropped? Can we just leave it with the current code and no correction other than 6:5? Do you have any other suggestion?

As much as I would like to eliminate the legacy baggage of SDL1, I think it needs to stick around until we don’t have any more platforms without an SDL2 port. That said, I don’t think those legacy platforms need to receive first-class support at this point, so accurate scaling only with SDL2 & OpenGL seems like a reasonable compromise.

Finally, mentioning this at the end only since you added a docblock on it: the `operator<` on Mode probably should be comparing AR-corrected dimensions and then `initSizeHint` would use it for finding the largest dimensions (there is a TODO in there about this).

bluegr:
@criezy Any news on this?

sev-:
I think, this now could be closed as way too outdated. Though I like the idea and I think, this is a good improvement.

criezy:
Right, I was only keeping this open as a reminder to myself that I should get back to it at some point. But that reminder has not really worked 😞
I will close this now.

PR #1133, opened 2018-02-06
Documentation support!

goops17:
I added some improvement in documentation. GSoC 2018 applicant.

### DIFF
--- a/README
+++ b/README
@@ -2,7 +2,7 @@ ScummVM README
 ------------------------------------------------------------------------
 
 For more information, compatibility lists, details on donating, the latest
-release, progress reports and more, please visit the ScummVM home page
+release, progress reports and more, please visit the ScummVM homepage
 at: <http://www.scummvm.org/>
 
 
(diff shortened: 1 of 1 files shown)


criezy:
Thank you for your suggestions, and for considering applying for GSoC with us. Also thank you for providing me with an opportunity to refresh my knowledge of comma usage in English ;-) I have committed separately some of your suggestions together with a few additional grammatical corrections. I have rejected however the changes to `home page` and `web page` (both the current forms and your proposed changes are correct, and I didn't see any strong reason to change it).

Also note that your commit message does not follow [our commit guidelines](http://wiki.scummvm.org/index.php/Commit_Guidelines). Please remember to follow our conventions for your next pull request.

I have also clarified the rules regarding GSoC applications (on http://wiki.scummvm.org/index.php/Summer_of_Code/Project_Rules) to explicitly indicate that we expect more than changes to the documentation for the patch submission as part of an application. Indeed we can't really be sure you can compile or understand the source code of the software if all you modify is the documentation ;-)

You are welcome to come and discuss with us on IRC (#scummvm channel on freenode) if you needs idea of a small task to work on. And it would be a good idea to discuss your application with us before the end of the application period anyway so that we can make suggestions to improve it.

PR #1134, opened 2018-02-06
KYRA: Added detection entry for EOB1 v1.9

r-lind:


bonki:

Can you provide a little more information on this? From what I understand this is about the unofficial 1.9 version from 2006 floating around the internet? Do you know if this exists as a patch? I've googled a bit and it looks as if this only available as a full game (i.e. pirated) download, which means we cannot support this.

r-lind:
Sorry! Since I own the original I just looked upon it as a patched version. But certainly the improvements should be made available in a patch like manner. Probably just a question of working through the list of changes (cut and pasted):

Eye of the Beholder version 1.9, based on Demo version 1.0
Release - July 11, 2006

- FIXED: All faulty Items and identifications. This includes the STONE GEM
- FIXED: Maps. All rooms unfinished in 1.4 are good for visitation
- FIXED: Joining NPCs. Beohram IS a paladin etc.
- FIXED: Some minor bugs.
- Feature: Poisons are NOT clearly marked "poison" :)
- Feature: No problems with Bones, disappearing or left after resurrection

The 1.0 DEMO it is based on is available here at the end: https://eob.wikispaces.com/share/view/16393369

r-lind:
Don't know much more I think.

PR #1136, opened 2018-02-17
AGI: Fix black box in bottom-right corner

Drenn1:
Future GSoC applicant here with a bugfix.

This fixes a minor bug in Space Quest 1 AGI, probably other AGI games; when you load a savegame after dying (or presumably anytime when the text cursor is not shown), a black square appears on the bottom right of the screen.

sq1_bug

This was caused by the text engine "forgetting" that there is no cursor being displayed; it then proceeds to try to delete the cursor, causing the text to leak onto the previous line.

Interestingly, there is a similar bug in the DOS version, but it only occurs if you type some text after dying. ScummVM had the exact opposite problem - it only occured if you didn't type any text...

### DIFF
--- a/engines/agi/text.cpp
+++ b/engines/agi/text.cpp
@@ -59,6 +59,7 @@ TextMgr::TextMgr(AgiEngine *vm, Words *words, GfxMgr *gfx) {
 
 	_inputEditEnabled = false;
 	_inputCursorChar = 0;
+	_inputCursorCharDrawn = false;
 
 	_statusEnabled = false;
 	_statusRow = 0;
@@ -629,8 +630,9 @@ bool TextMgr::inputGetEditStatus() {
 void TextMgr::inputEditOn() {
 	if (!_inputEditEnabled) {
 		_inputEditEnabled = true;
-		if (_inputCursorChar) {
+		if (_inputCursorCharDrawn) {
 			displayCharacter(0x08); // backspace
+			_inputCursorCharDrawn = false;
 		}
 	}
 }
@@ -640,6 +642,7 @@ void TextMgr::inputEditOff() {
 		_inputEditEnabled = false;
 		if (_inputCursorChar) {
 			displayCharacter(_inputCursorChar);
+			_inputCursorCharDrawn = true;
 		}
 	}
 }
@@ -887,7 +890,7 @@ void TextMgr::stringEdit(int16 stringMaxLen) {
 	_inputStringRow = _textPos.row;
 	_inputStringColumn = _textPos.column;
 
-	if (_inputCursorChar) {
+	if (_inputCursorCharDrawn) {
 		// Cursor character is shown, which means we are one beyond the start of the input
 		// Adjust the column for predictive input dialog
 		_inputStringColumn--;
--- a/engines/agi/text.h
+++ b/engines/agi/text.h
@@ -162,6 +162,7 @@ public:
 
 	bool  _inputEditEnabled;
 	byte  _inputCursorChar;
+	bool  _inputCursorCharDrawn;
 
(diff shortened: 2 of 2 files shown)


m-kiewitz:
Hi bonki,

Thanks for finding this bug. I'm pretty sure that the original interpreter checked for _inputCursorChar though, so that should be kept. I will have to dig through all of the different AGI versions and check what's going wrong first of all, because our behavior should be like in the original interpreter.
And after that we can look into a way of fixing this bug then.
AGI isn't that complicated, but there are tons of interpreter versions and one tiny change may introduce a problem in a game somewhere, so the first objective is to be as accurate as possible.

I had very busy weeks at work. Maybe I will take a day off in the next few days. At the very least I should be able to have some time to look into this during the next weekend. Will contact you again.

Kind Regards

bonki:
Credit goes to @Drenn1, I just pushed the review button :)

Drenn1:
If you're looking for the difference from the original interpreter, it might have something to do with "AgiEngine::loadGame". The crux of the problem is that it sets the cursor char to the value in the savefile, then calls "TextMgr::promptRedraw()", which in turn calls "TextMgr::inputEditOn()". At this point the cursor has been loaded but not drawn; however the game thinks it's drawn, which is why it tries to delete it.

I can only assume that maybe the save/loading routines differ from the original interpreter, causing this bug.

sev-:
@m-kiewitz Any chance for the review?

m-kiewitz:
@sev-
I have already said that this is definitely not the right way to go.

First the actual problem for the difference between ScummVM/AGI and Sierra's AGI must be found and fixed including verifying that all AGI versions have the same bug. Afterwards the AGI bug needs to get investigated, and after that one can check for any potential fix against that AGI bug.

I will have 2 weeks off from work in one week.

bonki:
@m-kiewitz Have you had a chance to start looking into this?

m-kiewitz:
@bonki 
I'm on it, I extended my holiday for a week as well.

m-kiewitz:
Sorry for abusing this conversation, but I don't have any other way of doing it.
@dreammaster 
I replied to your e-mail 10/11 days ago, right after I received it. I also replied yesterday.
It seems the e-mails do not go through, maybe got detected as spam on your end, i don't know. Please look for them. I'm currently not at work, so I can't send you an e-mail from work. I hope this goes through.

m-kiewitz:
Sorry, have to abuse it once more
@dreammaster 
I sent you 2 e-mails after your one from yesterday. It seems both of those were marked as spam once again. At least I haven't gotten an answer yet :(

bonki:
What's the status of this?

m-kiewitz:
@bonki 
As I said, I'm on it. Still.
I had to do workplace work even during the last few weekends and I can't prioritize this over my workload of course. Patience please.
I won't forget about it, even without reminders.

bonki:
@m-kiewitz 
ACK and apologies if it sounded pushy (it wasn't meant to), I was just curious if you had forgotten about this so I won't remind you again.
PS: I genuinely hope you get to recover from work even when working weekends. Take care!

m-kiewitz:
@bonki 
No problem.
I hope it all somewhat cleans up soon.

digitall:
For cross reference, this is bug Trac 10444:
https://bugs.scummvm.org/ticket/10444

bluegr:
So, the solution proposed for solving this bug is incorrect. There hasn't been any update on this for over a year, so there's no reason to keep it open. Any further work and comments can be made on bug 10444, or in a new pull request with a different approach.

As it stands, this pull request does not offer any added value, thus it shall be closed.

PR #1137, opened 2018-02-17
AUDIO: add support for OPL2LPT

vincentbernat:
The OPL2LPT is an OPL2 chip plugged on a parallel port. It is
write-only but otherwise acts as a classic AdLib. This commit adds
support for this device.

User is expected to have the right permissions on the parallel
port. By default, the first suitable parallel port is used. It is
possible to override that with the hidden configuration setting
"opl2lpt_parport".

It depends on the presence of the libieee1284 library which abstracts
a bit parallel port handling. An alternative would be to access
directly /dev/parportX on Linux. This would amount of code but it
would be Linux-only.

Tested with Indy 3 and SOMI.


Kroc:
Would this also give the potential to support the Covox Speech Thing further down the line? There is also a modern re-creation of this out there.

csnover:
Could you explain the benefit of this to users so that I can understand why this is desirable? Is there some problem with the OPL emulator that this is trying to solve?

vincentbernat:
This is only of interest for people having this particular device. Usually, they don't want to rely on emulation. After testing both the emulator and the real device, I can hear a slight difference, but this may also be a placebo effect.

Note I have no problem if this feature is not merged. People would find this PR and maybe comment on it if they are interested.

bonki:
I'm not commenting on whether we need this but if this gets merged I'd rather the `configure` options be `--{enable,disbale}-opl2lpt` and have that conditionally depend on `libieee1284` (also, `USE_OPL2LPT` instead of `USE_IEEE1284`).

vincentbernat:
That's something I can modify if there is a chance of merging.

audio/opl2lpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+			if (ieee1284_open(_pport, F1284_EXCL, &caps) != E1284_OK) {
+				warning("cannot open parallel port %s", _pport->name);
+			}
+			if (ieee1284_claim(_pport) != E1284_OK) {
+				warning("cannot claim parallel port %s", _pport->name);
+				ieee1284_close(_pport);
+				continue;
+			}
+			reset();
+			ieee1284_free_ports(&parports);
```
waltervn: At first glance the _pport assignment seems peculiar as parports is later freed. The answer there seems to be refcounting inside libieee1284? If so, a comment to explain that would be useful there, I think. Also, ieee1284.h states that "after calling <ieee1284_free_ports>, only ieee1284_find_ports may be used", so perhaps this port list should just be kept?

vincentbernat: The manual page `libieee1284` says the list list of ports can be freed after `ieee1284_open()`. Manual page of `ieee1284_free_ports()` says the structures are reference counted. Source code confirms, so it should be fine.

audio/opl2lpt.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "common/scummsys.h"
+
+#include "common/config-manager.h"
+#include "common/debug.h"
+#include "common/str.h"
+#include "common/textconsole.h"
+#include "audio/fmopl.h"
+
+#include <unistd.h>
+#include <sys/ioctl.h>
```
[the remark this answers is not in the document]
vincentbernat: I don't need it either, so I removed it.

libieee1284 says it supports Windows, but only certain modes. We are using the most basic features, but I don't know exactly what Windows support covers.

vincentbernat:
@waltervn PR amended with your remarks.

@bonki I tried to add an `--{enable,disable}-opl2lpt`, but I think there is currently no option depending on another option, except for engine features. I can come up with a code of my own but should I just rename the current option (since it's the only user of libieee1284) or do something more future proof with a way for an option to depend on another one?

bonki:
Sorry for the late reply. Something more future-proof would definitely be desirable.

sev-:
My suggestion is to name the configure option '--enable-opl2lpt', have it turned off by default, and check the library presence only when it is enabled. Then, have USE_IEEE1284 renamed to ENABLE_OPL2LPT, and use everywhere.

The reasoning is that the ieee1284 library anyway will be disabled only for this option, and we could just merge two concepts into one.

vincentbernat:
I have updated the PR for that. I only kept `IEEE1284_CFLAGS` and `IEEE1284_LIBS`.

PR #1138, opened 2018-02-19
XEEN: Add support for OPL to map to music sound type

dreammaster:
I'd like opinions on this before I merge it into master. Basically, World of Xeen has SFX using raw data files that can be passed directly to playStream, and a complicated music system that uses the ScummVM OPL. The default FMOPL implementation, though, is hardcoded to use the kPlainSoundType for the resultant audio stream. This commits adds in an optional sound type parameter to the start method, which allows the Xeen engine to specify it to output to a music stream, whose volume is then automatically properly controlled by the music volume slider.

### DIFF
--- a/audio/fmopl.cpp
+++ b/audio/fmopl.cpp
@@ -191,9 +191,9 @@ OPL *Config::create(DriverId driver, OplType type) {
 	}
 }
 
-void OPL::start(TimerCallback *callback, int timerFrequency) {
+void OPL::start(TimerCallback *callback, int timerFrequency, Audio::Mixer::SoundType soundType) {
 	_callback.reset(callback);
-	startCallbacks(timerFrequency);
+	startCallbacks(timerFrequency, soundType);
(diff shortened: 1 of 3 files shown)


waltervn:
Just a couple of notes. A few years ago we refactored the OPL system to allow it to drive OPL hardware rather than it being limited to just emulators. The reason why kPlainSoundType is used is that a hardware OPL will not pass through the ScummVM mixer. In order to have working volume control in both scenarios, volume control can be performed by adjusting the voice output levels. Games that support the AdLib card would have done volume control this way, since the AdLib card has no mixer or volume control.

dreammaster:
I was afraid of that. Plus, reviewing the sound code I'd previously implemented, I remembered that the decoder actually has code handling for both Music and some SFX, so I wouldn't be able to have it strictly outputing to kMusicSoundType anyway. Guess this change is a bust then.

PR #1139, opened 2018-03-09
README: Add notes about Quest for Glory character import

raziel-:
Taken mostly from the wiki and from criezy´s latest additions

### DIFF
--- a/README
+++ b/README
@@ -35,19 +35,20 @@ Table of Contents:
  * 3.14 Mickey's Space Adventure notes
  * 3.15 Myst notes
  * 3.16 Nippon Safes Inc. Amiga notes
- * 3.17 Riven notes
- * 3.18 Simon the Sorcerer notes
- * 3.19 Starship Titanic notes
- * 3.20 The Curse of Monkey Island notes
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -935,8 +937,32 @@ In addition, you will need to rename disk image 2 to disk1,
...
+For example, a character can be imported from Quest for Glory I directly
+to Quest for Glory III without having to necessarily play
+Quest for Glory II.
+
+Characters cannot be imported from future games to past games, nor can
+a character be imported to the same game that was just completed.
+In other words, a character from Quest for Glory II cannot be imported
+into Quest for Glory II.
+
+If you want to use a saved character from the original sierra interpreter,
```
bonki: s/sierra/Sierra/ ?

raziel-: Yes, there is always one thing i miss :-)

kintera:
Please remember quest for glory 4 has a bug in some versions that a qfg4
character can he imported back in to quest for glory 4, and 3's import is
buggy and will not always select the correct class on import.

sev-:
Thanks, merging. Good stuff.

PR #1140, opened 2018-03-13
PLUMBERS: optimization fixes

whiterandrek:


criezy:
Thank you. The changes are simple, but at first glance it looks like they all make sense.

Your last commit (3e2d2d4) does not comply with our conventions for commit messages though. But since it only fixes an indentation issue introduced in the previous commit I would suggest just to squash it with the previous commit anyway.

sev-:
Now it conflicts with the base. Could you please update the patch and I'll merge?

sev-:
Thanks!

PR #1141, opened 2018-03-16
SCI: High-quality "LarryScale" cel scaler for LSL7

DanielSWolf:
LSL7 uses nearest-neighbor scaling both for *downscaling* (if the actor moves to the back) and for *upscaling* (if the actor moves to the front). Neither looks good.

This PR implements a custom cel scaler called "LarryScale" that's optimized for cartoon-style graphics. Right now, I've only activated it in LSL7.

For details, see [issue #10196](https://bugs.scummvm.org/ticket/10196).

Here's a comparison between the default scaler and LarryScale:

larryscale


i30817:
Why does the scaler change the red thing in salad in the lower right if it's supposed to apply to actors perspective changes? Is it because that's actually a interactable object?

Just making sure that this doesn't affect backgrounds (even though this is none of my business and i'm just a luser).

bonki:
Out of curiosity, are there other games which could potentially benefit from this which meet LarryScaler's [limitations](https://github.com/scummvm/scummvm/pull/1141/commits/af0434cf091c868ad55ce8885872a147fbd22262#diff-412eefaaed2f72b98311717fdde56342R41)?

DanielSWolf:
Here's a quick list off the top of my head. I'm sure there are many more games that use the same art style and might benefit from LarryScale.

* Day of the Tentacle
* King's Quest VII
* Sam & Max Hit the Road
* Space Quest 6
* Toonstruck (except for Drew's sprite)
* Torin's Passage

Note, however, that some games rely on the exact implementation of their cel scaler. As [mentioned](https://bugs.scummvm.org/ticket/10196#comment:1) by @csnover, this might be the case for many SCI games. Replacing the cel scaler for these games might result in visual artifacts. So whether a certain game tolerates replacing its scaler would have to be tested.

bonki:
From the looks of it we can probably add *The Prince and the Coward* to the list, not sure in what state/shape our engine is, though.

DanielSWolf:
I don't think so. Judging by the screenshots I found, the cels in The Prince and the Coward are, at least partially, anti-aliased.

prince

rootfather:
Might The Curse of Monkey Island be a candidate too?

DanielSWolf:
Unfortunately, no. COMI uses beautifully antialiased graphics. Smoothly downscaling them would require a different algorithm.

bonki:
You seem to be right, forget I ever mentioned PRINCE :)

criezy:
This looks great.
After looking at the changes I only have one question for others in the team: shouldn't this option be turned off by default so that the default provides the original experience and enhancements are opt in? I don't really know what our policy is regarding this point, which is why I ask.

rootfather:
@criezy: Since it alters the appearance of the original game, I suggest to disable it by default and add an engine option for LarryScale.

bonki:
In this very specific case I would really like to see it turned on by default but I don't have a super strong of an opinion about it.

I think we all can agree on that this looks amazing and that we would like to merge it as soon as possible.
As far as I understand it, the "only" (hah) issue is properly testing it to make sure that we don't break anything which might require pixel-perfect alignment as explained by csnover in the original ticket.

sev-:
I would turn it on by default as well. We did the graphics enhancements in the past already.

rsn8887:
I agree with criezy that this enhancement should be opt-in since it looks completely different from the original. Also it seems to me that the dildo in the front has equal thickness red and white zebra stripes in the original, and much thinner white stripes in the scaled version? Can the scaled be tested with an object that has a simple zebra stripe pattern (equal thickness black and white stripes) to make sure it doesn’t diminish one of the stripes to one pixel thickness when zooming?

DanielSWolf:
I took a closer look at the dildo and found out why its red stripes seem to dominate. This picture shows the dildo in question (zoomed in to 200%). The rightmost image is the original, unscaled version; the others are scaled via LarryScale.

image

LarryScale works by trying to preserve lines, while sacrificing fills. You'll notice that the original image contains lots of dark-red lines. LarryScale carefully preserves these dark-red lines in the downscaled versions, which makes the resulting image look redder than intended.

I believe this is one of very few cases where the LarryScale'd version of an image doesn't look better than the nearest-neighbor-scaled version.

rsn8887:
Interesting! Looking at your images, it doesn’t really look all that bad. Even in this case, one could argue that the original image had slightly bigger red than white stripes. It all looks fine to me.

criezy:
The feedback so far is positive. I will leave a few more days for others to comment, and unless there are objections I will merge this this coming weekend.

bonki:
While I'd love to see this merged ASAP, I'm not sure it is a good idea to merge it as-is without any sort of testing from our side given that it has been said that this could in theory cause rendering errors if pixel-perfect alignment is required somewhere. Should we not at least do a single playthrough with this enabled first or can we rely upon proper testing before the next release?

criezy:
My though was that it would be easier for it to be tested once this is merged.

If the consensus is that we should wait until somebody has done a full play through before merging, then this is fine with me. I am just worried it will rot. I actually own this game and I never played it, so this might be an opportunity for me to play through it.

bonki:
Good point(s), then by all means go ahead :)

DanielSWolf:
I played about one hour of the game after implementing the scaler. During this time, there was a single occurrence where I felt that Larry's sprite jumped one pixel to the left at the beginning of an animation. This was hardly noticeable, however, and I'm not even sure that this didn't happen in the original game.

@criezy If you haven't played LSL7 before, by all accounts, give it a go! It's a great game if you don't mind its somewhat silly humor.

If you do find any problems, let me know. I'm rather busy at the moment, but I'll try to look into it!

Kroc:
Could SCUMMVM anti-alias the resulting image? (to give an effect such as in https://github.com/scummvm/scummvm/pull/1141#issuecomment-375994827)

DanielSWolf:
Not without making assumptions about the game's palette that might break some games. What you can do, however, is apply an antialiasing 2x scaler to the entire game, which gives quite impressive results.

criezy:
I have not forgotten about this, but since I am currently playing through the game with LarryScale I decided I might as well wait a bit more before merging.

DanielSWolf:
👍

criezy:
@DanielSWolf It looks like the change introduced a regression in Phantasmagoria (see https://bugs.scummvm.org/ticket/10568). Could you have a look and see if you can shed light on this?

DanielSWolf:
@criezy Sure, I'll look into it.

PR #1142, opened 2018-03-18
LILLIPUT: Merge of Lilliput Engine

Joefish:
This PR adds support for 'The Adventures of Robin Hood' by Millenium Interactive.
The game is in a playable state but input handling / engine structure needs some more love
as it is still quite laggy. Also audio is not working yet.
Despite those shortcomings, its core, the scripting engine seems to be completely reversed
and implemented. Also during my play testing I haven't had any issues except those
aforementioned. I must add that I haven't played it through as I'm not that familiar with the
game yet (and the controls are quite frustrating) so there may be problems I am not aware of.
@Strangerke may be able to correct me and go into more details.

Tested on Linux 4.15.7-1-ARCH
[Playtest video (7a978b6)](https://www.youtube.com/watch?v=019tv7Dea9k)

Looking forward to your suggestions :)

### DIFF
--- a/engines/lilliput/detection.cpp
+++ b/engines/lilliput/detection.cpp
@@ -172,7 +172,6 @@ SaveStateList LilliputMetaEngine::listSaves(const char *target) const {
 	pattern += "-##.SAV";
 
 	filenames = saveFileMan->listSavefiles(pattern);
-	sort(filenames.begin(), filenames.end());   // Sort (hopefully ensuring we are sorted numerically..)
 
 	SaveStateList saveList;
 	char slot[3];
@@ -210,6 +209,7 @@ SaveStateList LilliputMetaEngine::listSaves(const char *target) const {
 		}
 	}
 
+	Common::sort(saveList.begin(), saveList.end(), SaveStateDescriptorSlotComparator());
 	return saveList;
 }
 
--- a/engines/lilliput/detection.cpp
+++ b/engines/lilliput/detection.cpp
@@ -65,7 +65,7 @@ static const LilliputGameDescription gameDescriptions[] = {
 			},
 			Common::EN_ANY,
 			Common::kPlatformDOS,
-			ADGF_NO_FLAGS,
+			ADGF_UNSTABLE,
 			GUIO0()
 		},
 		kGameTypeRobin
@@ -81,7 +81,7 @@ static const LilliputGameDescription gameDescriptions[] = {
 			},
 			Common::FR_FRA,
 			Common::kPlatformDOS,
-			ADGF_NO_FLAGS,
+			ADGF_UNSTABLE,
 			GUIO0()
 		},
 		kGameTypeRobin
@@ -97,7 +97,7 @@ static const LilliputGameDescription gameDescriptions[] = {
 			},
 			Common::DE_DEU,
 			Common::kPlatformDOS,
-			ADGF_NO_FLAGS,
+			ADGF_UNSTABLE,
 			GUIO0()
 		},
 		kGameTypeRobin
@@ -113,7 +113,7 @@ static const LilliputGameDescription gameDescriptions[] = {
 			},
 			Common::IT_ITA,
 			Common::kPlatformDOS,
-			ADGF_NO_FLAGS,
+			ADGF_UNSTABLE,
 			GUIO0()
 		},
 		kGameTypeRobin
--- a/engines/lilliput/detection.cpp
+++ b/engines/lilliput/detection.cpp
@@ -169,7 +169,7 @@ SaveStateList LilliputMetaEngine::listSaves(const char *target) const {
 	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
 	Common::StringArray filenames;
 	Common::String pattern = target;
-	pattern += "-??.SAV";
+	pattern += "-##.SAV";
 
 	filenames = saveFileMan->listSavefiles(pattern);
 	sort(filenames.begin(), filenames.end());   // Sort (hopefully ensuring we are sorted numerically..)
--- a/engines/lilliput/detection.cpp
+++ b/engines/lilliput/detection.cpp
@@ -127,11 +127,11 @@ public:
 	}
 
 	const char *getName() const {
-		return "Robin Hood";
+		return "Lilliput";
 	}
 
 	const char *getOriginalCopyright() const {
-		return "Robin Hood Engine copyright S.L.Grand, Brainware, 1991-1992";
+		return "Lilliput Engine copyright S.L.Grand, Brainware, 1991-1992";
 	}
 
 	bool createInstance(OSystem *syst, Engine **engine, const ADGameDescription *gd) const;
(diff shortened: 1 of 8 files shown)


engines/robin/script.cpp
```diff
@@ -0,0 +1,3390 @@
...
+	debugC(1, kDebugScript, "OC_sub17BA5()");
+	
+	byte *tmpArr = getCharacterVariablePtr();
+	uint16 oper = _currScript->readUint16LE();
+	int16 var3 = getCharacterVariablePtr()[0];
+	computeOperation(tmpArr, oper, var3);
+}
+
+void RobinScript::OC_setByte18823() {
+	debugC(1, kDebugS
```
sev-: This is a noop. Should be ```x = CLIP(x, 0, 56)```. Check other cases too.

engines/robin/script.h
```diff
@@ -0,0 +1,311 @@
...
+class RobinEngine;
+
+enum kActionType {
+	kActionNone = 0,
+	kActionTalk = 3,
+	kActionGoto = 4,
+	kCodeEntered = 6
+};
+
+enum KValueType {
```
sev-: should be kValueType

Joefish:
Thanks for your comments @sev- I will look into it.

engines/lilliput/script.cpp
```diff
@@ -0,0 +1,3390 @@
...
+	index += 4;
+	_speechTimer = index;
+	_vm->displaySpeechBubble();
+	_vm->displaySpeech(_vm->_displayStringBuf);
+}
+
+void LilliputScript::decodePackedText(char *buf) {
+	debugC(2, kDebugScript, "decodePackedText(buf)");
+
+	// All the languages use the English dictionnary
```
bonki: Typo

engines/lilliput/script.cpp
```diff
@@ -0,0 +1,3390 @@
...
+}
+
+void LilliputScript::disasmScript(ScriptStream script) {
+	
+	while (!script.eos()) {
+		uint16 val = script.readUint16LE();
+		if (val == 0xFFF6) // end of script
+			return;
+
+		bool hasIf = false;
```
bonki: `hasIf` is unused - is this intentional?

engines/lilliput/script.cpp
```diff
@@ -0,0 +1,3390 @@
...
+	case 0xFA:
+		str = Common::String::format("(_array10999PosX[currentCharacter], _array109C1PosY[currentCharacter])");
+		break;
+	case 0xF9:
+		str = Common::String::format("(_currentCharacterVariables[4], _currentCharacterVariables[5])");
+		break;
+	case 0xF8: {
+		int index = curWord & 0xFF;
+		assert((index >= 0) && (index < 40));
+		str = Common::String::format("_vm->_rulesBuffer12Pos3[%d]", index);
```
bonki: Looks like a `break` is missing here. If this fallthrough really is intentional it should be marked as such with `// fall through`.

dreammaster:
Nice work on updating the history. Just a minor overlook.. the header file defines should ideally be updated as well from names like ROBIN_CONSOLE_H to use LILLIPUT_ instead

engines/lilliput/lilliput.h
```diff
@@ -0,0 +1,396 @@
...
+#include "engines/engine.h"
+#include "graphics/palette.h"
+#include "graphics/surface.h"
+
+namespace Common {
+class RandomSource;
+}
+
+/**
+ * This is the namespace of the Robin engine.
```
bonki: s/Robin/Lilliput/

i30817:
These two games (ROME AD&92 and Robin) are great candidates for wider resolutions and better pathfinding. Can't wait.

engines/lilliput/detection.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+SaveStateList LilliputMetaEngine::listSaves(const char *target) const {
+	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
+	Common::StringArray filenames;
+	Common::String pattern = target;
+	pattern += "-??.SAV";
+
+	filenames = saveFileMan->listSavefiles(pattern);
+	sort(filenames.begin(), filenames.end());   // Sort (hopefully ensuring we are sorted numerically..)
```
bgK: `Common::sort(saveList.begin(), saveList.end(), SaveStateDescriptorSlotComparator());` can now be used so replace hope with certainty.

engines/lilliput/detection.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+int LilliputMetaEngine::getMaximumSaveSlot() const {
+	return 99;
+}
+
+SaveStateList LilliputMetaEngine::listSaves(const char *target) const {
+	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
+	Common::StringArray filenames;
+	Common::String pattern = target;
+	pattern += "-??.SAV";
```
bgK: Since this has been written, `#` has been added to match specifically numbers.

engines/lilliput/lilliput.cpp
```diff
@@ -0,0 +1,2897 @@
...
+			break;
+		case Common::EVENT_KEYUP:
+		case Common::EVENT_KEYDOWN: {
+			if ((event.type == _lastKeyPressed.type) && (event.kbd == _lastKeyPressed.kbd))
+				break;
+
+			_lastKeyPressed = event;
+			int nextIndex = (_keyboard_nextIndex + 1) % 8;
+			if (_keyboard_oldIndex != nextIndex) {
+				_
```
bgK: Setting the seed to a fixed value will make the random source always return the same sequence.

bonki: @bgK I believe you meant to comment on another line? :)

bonki: EDIT: If someone else is wondering, this regards line 2847. For some reason the preview points to the wrong piece of code.

Joefish:
Thanks for the feedback everyone!

@i30817 This would be something we can take a look at once the base is adequarely stable, but definitely worth exploring later.
@bonki It seems save game support was partially implemented (or maybe even working at one point in the past) but right now it does not.
@dreammaster Actually I had it written down on my TODO list but somehow slippeed my mind.. Fixed it already though :)

dreammaster:
Another minor thing.. is there any particular reason a "_shouldQuit" variable is being set up and used in lilliput.cpp? It would seem easier not to bother with it, and simply use the base engine shouldQuit() method instead

sev-:
It is time to merge it and continue work in-tree.

rsn8887:
This has some narrowing conversions from int to char, and causes PSP2 build to fail:
http://buildbot.scummvm.org/builders/master-psp2/builds/2354/steps/compile/logs/stdio/text

Can this be fixed by someone who knows the engine, changing those char to uint8 or some such? Thanks!

PR #1143, opened 2018-03-21
AUDIO: Add Nuked OPL3 core

nukeykt:
This commit adds Nuked OPL3 core.
This core emulates OPL3 chip nearly perfectly as it is based on chip's internal logic reverse engineering and extensive hardware tests.
On downside this core requires more powerful device than MAME and DOSBox cores.

### DIFF
--- a/audio/softsynth/opl/nuked.cpp
+++ b/audio/softsynth/opl/nuked.cpp
@@ -1,0 +1,1488 @@
+//
+// Copyright (C) 2013-2018 Alexey Khokholov (Nuke.YKT)
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


sev-:
This is a very cool addition.

However, I noticed that you do not maintain it as a separate repository.

Would it be possible that you have it on your github account, and we could add a link to it in the comments?

In the meantime, I'm merging this. I will do couple adjustments after the merge.

audio/softsynth/opl/nuked.cpp
```diff
@@ -0,0 +1,1488 @@
...
+        }
+        else
+        {
+            chip->eg_add = shift + 1;
+        }
+    }
+
+    if (chip->eg_timerrem || chip->eg_state)
+    {
+        if (chip->eg_timer == 0xfffffffff)
```
sev-: Here must be 0xfffffffffUL

bonki: It really should be `0xfffffffffULL` because `long` is only guaranteed to be 32 bits.

nukeykt:
Thanks for merging.
At the moment, opl3.c/h in the Chocolate Doom's repository are considered as official. Links:
https://github.com/chocolate-doom/chocolate-doom/blob/master/opl/opl3.c
https://github.com/chocolate-doom/chocolate-doom/blob/master/opl/opl3.h

PR #1144, opened 2018-03-21
DS: Fix compilation and touch input

Drenn1:
This continues off of PR #1057.

The commits from there would allow it to build, but the .nds file produced wasn't working; it would whitescreen on load. My first commit tweaked the arguments to ndstool to get it working. I'm not certain why it was using objcopy in the first place, perhaps ndstool itself changed?

The second commit fixed the touch screen, which wasn't working at all. At first I thought it just wasn't working on my 3ds due to the removal of pressure sensitivity from the original DS; but, my original DS had the same issue, so I suppose that wasn't the problem.

There are still some outstanding issues:
* The binaries are larger than the old ScummVM DS releases. I'm not sure if the core is just bigger now or there's some other reason for it. I can't put as many engines into a single build anymore (4MB limit). Some engines might not fit even by themselves...
* Some menus aren't being drawn properly, particularly when you select "edit game" and try to change tabs.

Tested this with the AGI engine, which seems to work fine.

### DIFF
--- a/backends/platform/ds/arm9/source/osystem_ds.cpp
+++ b/backends/platform/ds/arm9/source/osystem_ds.cpp
@@ -23,6 +23,7 @@
 
 // Allow use of stuff in <time.h>
 #define FORBIDDEN_SYMBOL_EXCEPTION_time_h
+#define FORBIDDEN_SYMBOL_ALLOW_ALL
 
 #include "common/scummsys.h"
 #include "common/system.h"
--- a/backends/platform/ds/ds.mk
+++ b/backends/platform/ds/ds.mk
@@ -146,11 +146,8 @@ dsclean:
 
 # TODO: Add a 'dsdist' target ?
 
-%.bin: %.elf
-	$(OBJCOPY) -S -O binary $< $@
-
-%.nds: %.bin $(ndsdir)/arm7/arm7.bin
-	ndstool -c $@ -9 $< -7 $(ndsdir)/arm7/arm7.bin -b $(srcdir)/$(ndsdir)/$(LOGO) "$(@F);ScummVM $(VERSION);DS Port"
+%.nds: %.elf $(ndsdir)/arm7/arm7.elf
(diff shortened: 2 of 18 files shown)


sev-:
Thanks! Would you like to join the team and continue maintaining the port? Also, could you provide build for 2.0.0?

backends/platform/ds/ds.mk
```diff
@@ -170,10 +171,10 @@ dsclean:
 
 # HACK/FIXME: C compiler, for cartreset.c -- we should switch this to use CXX
 # as soon as possible.
-CC := $(DEVKITPRO)/devkitARM/bin/arm-eabi-gcc
+CC := $(DEVKITPRO)/devkitARM/bin/arm-none-eabi-gcc
```
bonki: This change broke buildbot because `arm-none-eabi-gcc` is not available. I guess we need to update our toolchain?

configure
```diff
@@ -2566,7 +2566,8 @@ case $_host_os in
 		append_var DEFINES "-DARM"
 		append_var DEFINES "-DNONSTANDARD_PORT"
 		append_var CXXFLAGS "-isystem $DEVKITPRO/libnds/include"
-		append_var CXXFLAGS "-isystem $DEVKITPRO/devkitARM/arm-eabi/include"
+		append_var CXXFLAGS "-I$DEVKITPRO/portlibs/nds/include"
```
bonki: Is there a particular reason why you're not using `-isystem` here?

backends/platform/ds/arm9/source/osystem_ds.cpp
```diff
@@ -23,6 +23,7 @@
 
 // Allow use of stuff in <time.h>
 #define FORBIDDEN_SYMBOL_EXCEPTION_time_h
+#define FORBIDDEN_SYMBOL_ALLOW_ALL
```
bonki: Why?

backends/plugins/ds/ds-provider.cpp
```diff
@@ -20,6 +20,8 @@
  *
  */
 
+#define FORBIDDEN_SYMBOL_ALLOW_ALL
```
bonki: Same here, why?

sev-:
Yes, currently buldbot is broken. Could you please fix it, or we will have to revert.

PR #1145, opened 2018-03-22
COMMON: Fix UB, left shift of negative value

bonki:


sev-:
Thanks! The change makes sense. Merging.

PR #1146, opened 2018-03-22
Remove unnecessary calls to setSeed()

bonki:


bonki:
Merging then, thanks!

PR #1148, opened 2018-03-25
SDL: Can load game controller DB file + joystick button for virtual keyboard

lmerckx:
1) Allow to load a custom game controller mapping file 
- Either define a "gcdbfile" property in your configuration file (pointing to desired file)
- Or copy a file named "gamecontrollerdb.txt" in your extra path
This file completes the standard SDL one to support more game controllers.
Can be found here, for instance: https://github.com/gabomdq/SDL_GameControllerDB

2) Add a specific joystick button to open virtual keyboard (button 7), only for joysticks with no game controller support or SDL1

### DIFF
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -73,8 +73,8 @@ static uint32 convUTF8ToUTF32(const char *src) {
 
 void SdlEventSource::loadGameControllerMappingFile() {
 	bool loaded = false;
-	if (ConfMan.hasKey("gcdbfile")) {
-		Common::FSNode file = Common::FSNode(ConfMan.get("gcdbfile"));
+	if (ConfMan.hasKey("controller_map_db")) {
+		Common::FSNode file = Common::FSNode(ConfMan.get("controller_map_db"));
 		if (file.exists()) {
 			if (SDL_GameControllerAddMappingsFromFile(file.getPath().c_str()) < 0)
 				error("File %s not valid: %s", file.getPath().c_str(), SDL_GetError());	
@@ -82,7 +82,8 @@ void SdlEventSource::loadGameControllerMappingFile() {
 				loaded = true;
 				debug("Game controller DB file loaded: %s", file.getPath().c_str());
 			}
-		}
+		} else
+			warning("Game controller DB file not found: %s", file.getPath().c_str());
 	}
 	if (!loaded && ConfMan.hasKey("extrapath")) {
 		Common::FSNode dir = Common::FSNode(ConfMan.get("extrapath"));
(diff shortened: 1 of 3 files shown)


backends/events/sdl/sdl-events.cpp
```diff
@@ -63,6 +70,31 @@ static uint32 convUTF8ToUTF32(const char *src) {
 
 	return utf32;
 }
+
+void SdlEventSource::loadGameControllerMappingFile() {
+	bool loaded = false;
+	if (ConfMan.hasKey("gcdbfile")) {
```
bgK: `gcdbfile` is not a very explicit name. What about `controller_mapping_db`?

Also please document the new setting in the README, and mention `gamecontrollerdb.txt` will be loaded from `extrapath` otherwise.

backends/events/sdl/sdl-events.cpp
```diff
@@ -63,6 +70,31 @@ static uint32 convUTF8ToUTF32(const char *src) {
...
+		Common::FSNode file = Common::FSNode(ConfMan.get("gcdbfile"));
+		if (file.exists()) {
```
bgK: If the user has configured a mapping file and it was not found, it would be nice to print a warning saying so.

backends/events/sdl/sdl-events.cpp
```diff
@@ -63,6 +70,31 @@ static uint32 convUTF8ToUTF32(const char *src) {
...
+			if (SDL_GameControllerAddMappingsFromFile(file.getPath().c_str()) < 0)
+				error("File %s not valid: %s", file.getPath().c_str(), SDL_GetError());	
+			else {
+				loaded = true;
+				debug("Game controller DB file loaded: %s", file.getPath().c_str());
+			}
+		}
+	}
+	if (!loaded && ConfMan.hasKey("extrapath")) {
```
bgK: I'm not too fond of using `extrapath` directly. Usually ScummVM looks up files using the SearchManager which also includes platform specific paths. However SearchMan cannot give the full path for the files (as they are not necessarily directly on the filesystem). That being said I have no idea how to improve this while keeping the code simple. I guess this is fine and can always be improved later if needed.

backends/events/sdl/sdl-events.cpp
```diff
@@ -85,6 +117,8 @@ SdlEventSource::SdlEventSource()
 		if (SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER) == -1) {
 			error("Could not initialize SDL: %s", SDL_GetError());
 		}
+		else
```
bgK: That else statement is not necessary as `error` quits ScummVM if `SDL_InitSubSystem` fails.

lmerckx:
Thank you for your review bgK. Following changes have been made:
- Use `controller_map_db` property for configuration file (controller_mapping_db was too long for README file) ;-)
- README file has been updated
- Add warning message if `controller_map_db` file doesn't exist
- Indeed, my first idea was to use the SearchManager, but I can't get the complete path of the file found. So, I left this code unchanged for the moment.
- The else has been removed after InitSubSystem

Don't hesitate if some improvements are still needed !

PR #1150, opened 2018-03-26
ALL: Load savegame thumbnail only when necessary

bonki:
The rationale behind this is that it's better to only load the thumbnail when actually used (in most cases this is `FooMetaEngine::querySaveMetaInfos()`, but there are exceptions which I hope I have all caught) than to always load it whenever a savegame's header is read and throw it away again immediately.

A few engines already do this.

### DIFF
--- a/common/recorderfile.cpp
+++ b/common/recorderfile.cpp
@@ -608,7 +608,8 @@ Graphics::Surface *PlaybackFile::getScreenShot(int number) {
 		if (screenCount == number) {
 			screenCount++;
 			_readStream->seek(-4, SEEK_CUR);
-			return Graphics::loadThumbnail(*_readStream);
+			Graphics::Surface *thumbnail;
+			return Graphics::loadThumbnail(*_readStream, thumbnail) ? thumbnail : NULL;
 		} else {
 			uint32 size = _readStream->readUint32BE();
 			_readStream->skip(size-8);
--- a/engines/access/access.cpp
+++ b/engines/access/access.cpp
@@ -488,11 +488,6 @@ Common::Error AccessEngine::loadGameState(int slot) {
 	if (!readSavegameHeader(saveFile, header))
 		error("Invalid savegame");
 
-	if (header._thumbnail) {
-		header._thumbnail->free();
-		delete header._thumbnail;
-	}
-
 	// Load most of the savegame data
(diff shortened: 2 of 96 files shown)


sev-:
Good idea, but what about doing it differently? Instead of writing everywhere 
```
        if (loadThumbnail) {
		header._thumbnail = Graphics::loadThumbnail(*in);
		if (!header._thumbnail)
			return false;
	} else {
		Graphics::skipThumbnail(*in);
	}
```

You write:

```
        header._thumbnail = Graphics::loadThumbnail(*in, loadThumbnail);
        if (!header._thumbnail)
	    return false;
```

And at the start of `loadThumbnail()` you do:

```
       if (!loadThumbnail)
           return true;
```

That will make code compact everywhere, and will ensure that the engine authors will not miss it in the future.

sev-:
Right, I did not check the prototype, and for some reason thought that loadThumbnail return a boolean. And yes, somebody could skip the bytes there.

But since it is not boolean, what about returning `nullptr`, but then modify the receiving code to correctly process nulls, and not returning early, as it is now?

bonki:
`loadThumbnail()` already returns null to indicate failure in case of a broken thumbnail header/data in which case it is necessary to return early from the caller as that indicates a broken savegame so that is not a good idea either (especially as the caller would still have to additionally check if it actually wanted to load or skip the thumbnail when seeing null).

Instead, I propose to actually let it return a boolean and take a boolean as per your earlier suggestion. Thus,

    Graphics::Surface *loadThumbnail(Common::SeekableReadStream &in);

becomes

    bool loadThumbnail(Common::SeekableReadStream &in, Graphics::Surface *&thumbnail, bool skipThumbnail);

The caller now only does a

        if (!Graphics::loadThumbnail(*file, header._thumbnail, skipThumbnail))
            return false;

See https://github.com/scummvm/scummvm/compare/master...bonki:bonki-all-ondemand-thumbnails2 for the actual changes using TITANIC as example.

Note that with POLA in mind I chose to let `loadThumbnail()`s `skipThumbnail` default to `false` while `readSavegameHeader()`s defaults to `true`. That way they do the right thing(tm) in their respective default case.

sev-:
Time to merge. Thanks!

PR #1151, opened 2018-03-26
[WIP] SUPERNOVA: Refactoring of the engine

Joefish:
This is an ongoing effort to improve and refactor the engine. The goal is to create less convoluted classes and a clear separation of responsibilities (rendering, logic, sound, ..). With this I also hope to fix user input stalling calls (e.g. no quit during paletteFade) and remove wrappers like [exitOnEscape](https://github.com/scummvm/scummvm/blob/master/engines/supernova/rooms.cpp#L373).  
Right now, there are member functions in SupernovaEngine that forward their calls to the appropriate class instance, e.g. renderImage to Screen::renderImage that rooms.cpp and state.cpp (the game logic) depend on.

My initial idea was writing a manager to defer render/sound/.. calls being issued from game logic so user input can be polled at certain intervals, screen redrawn, .. between those calls. As I was sketching it out I realized I was basically writing a virtual machine what is waay overkill on the problem (not to mention it would entail a complete rewrite of the game logic). That's how I defined the goals mentioned above.  
There is still work to be done, as seen here:
- [ ] seperate game logic from render calls in Screen
- [ ] Substitute render calls for ImageId variant
- [ ] Input Abstraction
- [ ] Cutscene Abstraction
- [ ] Maybe extend Cutscene for Room::animation (TBD)
- [ ] Refactor terminal screen ([SleepCabin](https://github.com/scummvm/scummvm/blob/master/engines/supernova/rooms.cpp#L679), [AxacussOffice](https://github.com/scummvm/scummvm/blob/master/engines/supernova/state.cpp#L745))

After this, it will be easier to determine what the blockers are for refactoring rooms and state.
  
Just saw cd7b85dbc6 is broken. Will be fixed later today. :heavy_check_mark:


criezy:
Overall it looks good, but I would suggest squashing some commits to clean the history.
- Commit 1523d2e adds a bunch of function declarations without implementing them. I don't like it and I would prefer if it was squashed with commit 4206c5e that implements those functions.
- Commit 41367d8 fixes commit d0947cf and they could probably be squashed?
- Commits baa8fb7 and 2d7c214 fix commit cd5b932 and they could also probably be squashed?

Another questions concerns the "buffering" of the images. I remember that loading all the images when starting the engine is how you implemented it initially. I changed it to load images on demand. The reason for this was to limit memory usage and make the engine starts faster. And loading one file appeared to be fast enough to be done on the fly. What is the reason to change it back to loading all the images at the start?

And the follow up question would be: is there any reason to keep loading all the sounds at the start, or could it be done when we need them?

And finally from the work done on the ImageId and renderImage, I am guessing you have more work planned to use those?

engines/supernova/detection.cpp
```diff
@@ -60,13 +60,22 @@ static const ADGameDescription gameDescriptions[] = {
...
-		NULL,
+		nullptr,
 		AD_ENTRY1s("ms2_data.000", "e595610cba4a6d24a763e428d05cc83f", 24805),
 		Common::DE_DEU,
 		Common::kPlatformDOS,
 		ADGF_UNSTABLE,
 		GUIO1(GUIO_NONE)
 	},
+	{
+		"msn2",
```
sev-: why this duplicate entry is here?

Joefish: huh.. Guess there went something wrong when I resolved a conflict.

Joefish:
@criezy I agree with the squashing of the commits you named.
Concerning the buffering of images, I didn't remember that it was like that before, but it seems unnecessary to me to create/destroy the image on every call (if images are in different files). While it increases the mem usage (by ~5.5MB incl. audio), it reduces code complexity as it now just returns a pointer to the loaded image in resource manager instance.
Personally, I don't mind both ways. I just thought it to be more convenient doing file IO at the beginning once and having everything residing in RAM instead of 'poking' the hdd every few seconds, especially since the footprint is so small.
I haven't tested it on a low-end machine if the startup is unnecessarily delayed by it though.

Same reasoning with Audio. I didn't see a reason to destroy it if we have already loaded it once in a SeekableReadStream that can be easily rewinded.

And yes there's more work to come that uses ImageId (todo: Substitute render calls for ImageId variant). I was talking with @Strangerke a few weeks ago and he mentioned that it would be nice to have an Object list per Room for better debuggability. That's when I thought I wish I had mapped out all images when I was originally implementing it as it was quite confusing from time to time what image the render call actually referred to.

criezy:
@Joefish Then I think this can be merged once the relevant commits have been squashed. The work can continue in-tree, and we can change the image and sound loading policy later if we want to (and anyway this should have no impact to the code outside of the resource manager class you introduced, so should be fairly easy to change).

We also need to fix the global constructor issue flagged by @sev in the current master (and not fixed by this PR as far as I can see) with the nullObject. I can do that once this PR has been merged.

Joefish:
The commits were rebased on the latest master commit and the global ctor call for nullObject removed.
My intention for using nullObject instead of nullptr is that I don't have to litter the rendering and input code with nullptr checks (although in some cases GameManger::isNullObject takes over that part) as the nullObject represents a valid Object and doesn't need any additional care.
@criezy, if you a better/cleaner way of solving it don't hesitate to throw out 138311f and do it your way.
I also recompiled every commit without issues, so it can be merged and continue work in-branch.

criezy:
Thank you. I merged this with a few changes to fix warnings that were introduced by some of those commits.

PR #1154, opened 2018-03-28
COMMON: Add WARN_UNUSED_RESULT to scummsys.h

bonki:
Compatibility based on the information provided [here](https://stackoverflow.com/questions/4226308/msvc-equivalent-of-attribute-warn-unused-result/49468398#49468398).

### DIFF
--- a/common/scummsys.h
+++ b/common/scummsys.h
@@ -398,6 +398,18 @@
 	#endif
 #endif
 
+#ifndef WARN_UNUSED_RESULT
+	#if __cplusplus >= 201703L
+		#define WARN_UNUSED_RESULT [[nodiscard]]
+	#elif GCC_ATLEAST(3, 4)
+		#define WARN_UNUSED_RESULT __attribute__((__warn_unused_result__))
+	#elif defined(_Check_return_)
+		#define WARN_UNUSED_RESULT _Check_return_
+	#else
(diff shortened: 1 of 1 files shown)


sev-:
This is a very good thing.

So far, we are relying on Coverity for reporting these, but having them reported right away during compilation would help a lot.

Curious to hear what other devs think.

dreammaster:
Seems okay to me. At first, I was slightly concerned that it might cause issues with methods whose return value doesn't have to be used, such as pause methods that return whether it was interrupted with a keypress or not. But on review, it looks like this is simply a new attribute that we can apply to specific function declarations to ensure their return value is used. Correct?

bonki:
That is correct. In fact, you can't even enable this globally for gcc (I don't know about MSVC/clang), you have to apply the attribute per function. So while this doesn't give us anything for free, it at least provides us with a means to enable this for functions where it makes sense.

bgK:
Good idea! Do you plan to apply this attribute to some functions in `common/`?

criezy:
This seems like a good idea to me as well.

bonki:
Thanks for the feedback, merging then! @bgK Not right away and I have not really looked into it, but `File::*` is probably a good candidate to begin with?

PR #1155, opened 2018-03-28
LILLIPUT: Fix narrowing conversion compiler errors on psp2 build

rsn8887:
conversion from int to char causes psp2 build to fail, see log here:
http://buildbot.scummvm.org/builders/master-psp2/builds/2356/steps/compile/logs/stdio

This is my attempt at fixing that problem


rsn8887:
Review requested.

rsn8887:
It seems to just work... maybe I should merge?

PR #1156, opened 2018-04-04
GUI: Unify clip and non-clip draw calls

bgK:
This is a clean up for the GUI code. The main change is that the draw clipping rect is now part of the ThemeEngine state, allowing to unify the clip and non-clip variants methods, reducing code duplication.

This also contains some unused code removal.

### DIFF
--- a/gui/widget.cpp
+++ b/gui/widget.cpp
@@ -411,19 +411,17 @@ void ButtonWidget::setUnpressedState() {
 
 PicButtonWidget::PicButtonWidget(GuiObject *boss, int x, int y, int w, int h, const char *tooltip, uint32 cmd, uint8 hotkey)
 	: ButtonWidget(boss, x, y, w, h, "", tooltip, cmd, hotkey),
-	  _alpha(255), _transparency(false), _showButton(true), _isAlpha(false) {
+	  _alpha(255), _transparency(false), _showButton(true) {
 
 	setFlags(WIDGET_ENABLED/* | WIDGET_BORDER*/ | WIDGET_CLEARBG);
 	_type = kButtonWidget;
-	_mode = ThemeEngine::kAutoScaleNone;
 }
 
 PicButtonWidget::PicButtonWidget(GuiObject *boss, const Common::String &name, const char *tooltip, uint32 cmd, uint8 hotkey)
 	: ButtonWidget(boss, name, "", tooltip, cmd, hotkey),
-	  _alpha(255), _transparency(false), _showButton(true), _isAlpha(false) {
+	  _alpha(255), _transparency(false), _showButton(true) {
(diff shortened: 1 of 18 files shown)


gui/widget.cpp
```diff
@@ -451,23 +426,6 @@ void PicButtonWidget::setGfx(const Graphics::Surface *gfx, int statenum) {
 	_gfx[statenum].copyFrom(*gfx);
 }
 
-void PicButtonWidget::setAGfx(const Graphics::TransparentSurface *gfx, int statenum, ThemeEngine::AutoScaleMode mode) {
```
sev-: Why is this removed? It is not used in the current code, but I added transparent PNG usage to the themes, and it could be used for the buttons now.

gui/widget.cpp
```diff
@@ -411,19 +388,17 @@ void ButtonWidget::setUnpressedState() {
 
 PicButtonWidget::PicButtonWidget(GuiObject *boss, int x, int y, int w, int h, const char *tooltip, uint32 cmd, uint8 hotkey)
 	: ButtonWidget(boss, x, y, w, h, "", tooltip, cmd, hotkey),
-	  _alpha(255), _transparency(false), _showButton(true), _isAlpha(false) {
+	  _alpha(255), _transparency(false), _showButton(true) {
 
 	setFlags(WIDGET_ENABLED/* | WIDGET_BORDER*/ | WIDGET_CLEARBG);
 	_type = kButtonWidget;
-	_mode = ThemeEngine::kAutoScaleNone;
```
sev-: Why are you removing picture scaling? It is not used right now, but was used as part of Simon 20th Anniversary project, and thus, could be easily reused by our themes now.

gui/widget.cpp
```diff
@@ -809,24 +719,7 @@ void GraphicsWidget::drawWidget() {
 		const int x = _x + (_w - _gfx.w) / 2;
 		const int y = _y + (_h - _gfx.h) / 2;
 
-		g_gui.theme()->drawSurfaceClip(Common::Rect(x, y, x + _gfx.w,  y + _gfx.h), getBossClipRect(), _gfx, _state, _alpha, _transparency);
-	} else if (_agfx.getPixels()) {
```
sev-: Why are you making this simplification? This will limit the PNG usage.

gui/widget.cpp
bgK: I removed that code because:
* It was never used since its introduction 4 years ago
* It was not updated to work with the clipping rect (meaning it is broken)

The same reasoning applies to your other comments.

sev-: Valid point. Then it is fine to merge, and once I work on real adding of the PNG, I can re-add those.

Still, in my opinion, these two commits need to be separated, as this has nothing to do with the unification.

PR #1159, opened 2018-04-15
MOHAWK: MYST: Fix Observatory displayed hour from 0:XX to 12:XX

dafioram:
Fixes #Trac10486.

In ScummVM when the hour was set to what should be 12:XX it was
displayed as 0:XX. This differs with the original which displays
the 12:xx.

I verified this fix in both Myst ME and Myst English CD.

### DIFF
--- a/engines/mohawk/myst_stacks/myst.cpp
+++ b/engines/mohawk/myst_stacks/myst.cpp
@@ -633,25 +633,43 @@ uint16 Myst::getVar(uint16 var) {
 			return 10;
 	case 79: // Stellar Observatory Date - Year #4 (Right)
 		return (_state.observatoryYearSetting / 1) % 10;
-	case 80: // Stellar Observatory Hour #1 - Left ( Number 1 (0) or Blank (10))
+	case 80: // Stellar Observatory Hour #1 - Left ( Hour digits can be 10 (Blank), or 0-2)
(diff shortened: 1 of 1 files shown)


engines/mohawk/myst_stacks/myst.cpp
```diff
@@ -634,24 +634,32 @@ uint16 Myst::getVar(uint16 var) {
 	case 79: // Stellar Observatory Date - Year #4 (Right)
 		return (_state.observatoryYearSetting / 1) % 10;
 	case 80: // Stellar Observatory Hour #1 - Left ( Number 1 (0) or Blank (10))
+		uint32 observatoryMinutes,observatoryHour;
```
bgK: Please don't share variables between switch cases, it's confusing.

dafioram:
Done.

dafioram:
Does it need to be clarified why line 651 is returning a 2?

bgK:
> Does it need to be clarified why line 651 is returning a 2?

Sorry, I just tested with my French version that uses the 24h format, and this is not right. The first hour of the day now displays as 12:00 -> 12:59 while it should be 00:00 -> 00:59.

dafioram:
See if my latest change works better in the 24 Format/french version.

bgK:
All good, thanks!

PR #1160, opened 2018-04-16
GUI: Implement an "Unknown game dialog"

rootfather:
Thanks to the great help of @criezy, here's my implementation of an GUI dialog that appears when an unknown game is detected.

Features:
 - Allows copying the data collected by game detector to the clipboard
 - Allows opening the bug tracker in the default web browser and pre-fills the form fiels

Feedback is greatly appreciated! As I'm still learning how to C++ and have little experience, don't hesitate to tell me if I did something awfully stupid. :)

@wjp: The option to automatically report the collected data unfortunately will fail ungracefully if the user is not logged in to ScummVM's Trac instance. There's also something wrong with https://bugs.scummvm.org/login which to my understanding would handle this case by a redirect. 

scummvm-unknown-game-dialog
scummvm-unknown-game-dialog-bugtracker

This closes https://bugs.scummvm.org/ticket/10435.

### DIFF
--- a/backends/platform/sdl/macosx/macosx_wrapper.h
+++ b/backends/platform/sdl/macosx/macosx_wrapper.h
@@ -27,6 +27,7 @@
 
 bool hasTextInClipboardMacOSX();
 Common::String getTextFromClipboardMacOSX();
+bool setTextInClipboardMacOSX(const Common::String&);
 Common::String getDesktopPathMacOSX();
 
 #endif
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -33,6 +33,7 @@
 #include "gui/EventRecorder.h"
 #include "common/taskbar.h"
 #include "common/textconsole.h"
+#include "common/translation.h"
 
 #include "backends/saves/default/default-saves.h"
 
@@ -502,11 +503,18 @@ Common::String OSystem_SDL::getTextFromClipboard() {
 
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	char *text = SDL_GetClipboardText();
+#ifdef USE_TRANSLATION
+	// The string returned by SDL is in UTF-8. Convert to the
+	// current TranslationManager encoding.
+	char *conv_text = SDL_iconv_string(TransMan.getCurrentCharset().c_str(), "UTF-8", text, SDL_strlen(text) + 1);
+	if (conv_text) {
+		SDL_free(text);
+		text = conv_text;
+	}
+#endif
 	Common::String strText = text;
 	SDL_free(text);
 
-	// FIXME: The string returned by SDL is in UTF-8, it is not clear
-	// what encoding should be used for the returned string.
 	return strText;
 #else
 	return "";
@@ -515,7 +523,16 @@ Common::String OSystem_SDL::getTextFromClipboard() {
 
 bool OSystem_SDL::setTextInClipboard(const Common::String& text) {
 #if SDL_VERSION_ATLEAST(2, 0, 0)
-	// FIXME: The string we get from SDL is in UTF-8 and should probably be converted.
+#ifdef USE_TRANSLATION
+	// Assume the encoding is the current TranslationManager encoding
+	// and attempt to convert to UTF-8 (which is what SDL takes).
+	char *utf8_text = SDL_iconv_string("UTF-8", TransMan.getCurrentCharset().c_str(), text.c_str(), text.size() + 1);
+	if (utf8_text) {
+		int status = SDL_SetClipboardText(utf8_text);
+		SDL_free(utf8_text);
+		return status == 0;
+	}
+#endif
 	return SDL_SetClipboardText(text.c_str()) == 0;
 #else
 	return false;
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -513,6 +513,15 @@ Common::String OSystem_SDL::getTextFromClipboard() {
 #endif
 }
 
+bool OSystem_SDL::setTextInClipboard(const Common::String& text) {
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+	// FIXME: The string we get from SDL is in UTF-8 and should probably be converted.
+	return SDL_SetClipboardText(text.c_str()) == 0;
+#else
+	return false;
+#endif
+}
+
 uint32 OSystem_SDL::getMillis(bool skipRecord) {
 	uint32 millis = SDL_GetTicks();
 
--- a/backends/platform/sdl/sdl.h
+++ b/backends/platform/sdl/sdl.h
@@ -72,6 +72,7 @@ public:
 	// Clipboard
 	virtual bool hasTextInClipboard();
 	virtual Common::String getTextFromClipboard();
+	virtual bool setTextInClipboard(const Common::String&);
 
 	virtual void setWindowCaption(const char *caption);
 	virtual void addSysArchivesToSearchSet(Common::SearchSet &s, int priority = 0);
--- a/base/plugins.cpp
+++ b/base/plugins.cpp
@@ -514,7 +514,7 @@ GameDescriptor EngineManager::findGameInLoadedPlugins(const Common::String &game
 	return result;
 }
 
-GameList EngineManager::detectGames(const Common::FSList &fslist) const {
+GameList EngineManager::detectGames(const Common::FSList &fslist, bool useUnknownGameDialog) const {
 	GameList candidates;
 	PluginList plugins;
 	PluginList::const_iterator iter;
@@ -524,7 +524,7 @@ GameList EngineManager::detectGames(const Common::FSList &fslist) const {
 		// Iterate over all known games and for each check if it might be
 		// the game in the presented directory.
 		for (iter = plugins.begin(); iter != plugins.end(); ++iter) {
-			candidates.push_back((*iter)->get<MetaEngine>().detectGames(fslist));
+                       candidates.push_back((*iter)->get<MetaEngine>().detectGames(fslist, useUnknownGameDialog));
 		}
 	} while (PluginManager::instance().loadNextPlugin());
 	return candidates;
--- a/base/plugins.cpp
+++ b/base/plugins.cpp
@@ -524,7 +524,7 @@ GameList EngineManager::detectGames(const Common::FSList &fslist, bool useUnknow
 		// Iterate over all known games and for each check if it might be
 		// the game in the presented directory.
 		for (iter = plugins.begin(); iter != plugins.end(); ++iter) {
-                       candidates.push_back((*iter)->get<MetaEngine>().detectGames(fslist, useUnknownGameDialog));
+			candidates.push_back((*iter)->get<MetaEngine>().detectGames(fslist, useUnknownGameDialog));
 		}
 	} while (PluginManager::instance().loadNextPlugin());
 	return candidates;
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -374,7 +374,7 @@ void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFileP
 
 	// Check if the GUI is running, show the UnknownGameDialog and print the translated unknown game information
 	if (GUI::GuiManager::hasInstance() && g_gui.isActive()) {
-		GUI::UnknownGameDialog dialog(report, reportTranslated, bugtrackerAffectedEngine);
+		UnknownGameDialog dialog(report, reportTranslated, bugtrackerAffectedEngine);
 		dialog.runModal();
 	}
 }
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -150,7 +150,7 @@ bool cleanupPirated(ADGameDescList &matched) {
 }
 
 
-GameList AdvancedMetaEngine::detectGames(const Common::FSList &fslist) const {
+GameList AdvancedMetaEngine::detectGames(const Common::FSList &fslist, bool useUnknownGameDialog) const {
 	ADGameDescList matches;
 	GameList detectedGames;
 	FileMap allFiles;
@@ -162,7 +162,7 @@ GameList AdvancedMetaEngine::detectGames(const Common::FSList &fslist) const {
 	composeFileHashMap(allFiles, fslist, (_maxScanDepth == 0 ? 1 : _maxScanDepth));
 
 	// Run the detector on this
-	matches = detectGame(fslist.begin()->getParent(), allFiles, Common::UNK_LANG, Common::kPlatformUnknown, "");
+       matches = detectGame(fslist.begin()->getParent(), allFiles, Common::UNK_LANG, Common::kPlatformUnknown, "", useUnknownGameDialog);
 
 	if (matches.empty()) {
 		// Use fallback detector if there were no matches by other means
@@ -327,7 +327,7 @@ Common::Error AdvancedMetaEngine::createInstance(OSystem *syst, Engine **engine)
 		return Common::kNoError;
 }
 
-void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFilePropertiesMap &filesProps, const ADGameIdList &matchedGameIds) const {
+void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFilePropertiesMap &filesProps, const ADGameIdList &matchedGameIds, bool useUnknownGameDialog) const {
 	const char *reportCommon = "The game in '%s' seems to be an unknown %s engine game "
 							   "variant.\n\nPlease report the following data to the ScummVM "
 							   "team at %s along with the name of the game you tried to add and "
@@ -373,7 +373,7 @@ void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFileP
 	g_system->logMessage(LogMessageType::kInfo, reportLog.c_str());
 
 	// Check if the GUI is running, show the UnknownGameDialog and print the translated unknown game information
-	if (GUI::GuiManager::hasInstance() && g_gui.isActive()) {
+       if (GUI::GuiManager::hasInstance() && g_gui.isActive() && useUnknownGameDialog == true) {
 		UnknownGameDialog dialog(report, reportTranslated, bugtrackerAffectedEngine);
 		dialog.runModal();
 	}
@@ -449,7 +449,7 @@ bool AdvancedMetaEngine::getFileProperties(const Common::FSNode &parent, const F
 	return true;
 }
 
-ADGameDescList AdvancedMetaEngine::detectGame(const Common::FSNode &parent, const FileMap &allFiles, Common::Language language, Common::Platform platform, const Common::String &extra) const {
+ADGameDescList AdvancedMetaEngine::detectGame(const Common::FSNode &parent, const FileMap &allFiles, Common::Language language, Common::Platform platform, const Common::String &extra, bool useUnknownGameDialog) const {
 	ADFilePropertiesMap filesProps;
 
 	const ADGameFileDescription *fileDesc;
@@ -574,7 +574,7 @@ ADGameDescList AdvancedMetaEngine::detectGame(const Common::FSNode &parent, cons
 	// We didn't find a match
 	if (matched.empty()) {
 		if (!filesProps.empty() && gotAnyMatchesWithAllFiles) {
-			reportUnknown(parent, filesProps, matchedGameIds);
+                       reportUnknown(parent, filesProps, matchedGameIds, useUnknownGameDialog);
 		}
 
 		// Filename based fallback
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -358,17 +358,18 @@ void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFileP
 	reportTranslated += "\n\n";
 
 	reportTranslated.wordWrap(65);
-
-	for (ADFilePropertiesMap::const_iterator file = filesProps.begin(); file != filesProps.end(); ++file) {
-		report += Common::String::format("  {\"%s\", 0, \"%s\", %d},\n", file->_key.c_str(), file->_value.md5.c_str(), file->_value.size);
-		reportTranslated += Common::String::format("  {\"%s\", 0, \"%s\", %d},\n", file->_key.c_str(), file->_value.md5.c_str(), file->_value.size);
-	}
-	report += "\n";
-	reportTranslated += "\n";
-
-	// Write the original message about the unknown game to the log file
 	Common::String reportLog = report;
 	reportLog.wordWrap(80);
+
+	Common::String unknownFiles;
+	for (ADFilePropertiesMap::const_iterator file = filesProps.begin(); file != filesProps.end(); ++file)
+		unknownFiles += Common::String::format("  {\"%s\", 0, \"%s\", %d},\n", file->_key.c_str(), file->_value.md5.c_str(), file->_value.size);
+
+	report += unknownFiles;
+	reportTranslated += unknownFiles;
+	reportLog += unknownFiles + "\n";
+
+	// Write the original message about the unknown game to the log file
 	g_system->logMessage(LogMessageType::kInfo, reportLog.c_str());
 
 	// Check if the GUI is running, show the UnknownGameDialog and print the translated unknown game information
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -30,6 +30,8 @@
 #include "common/textconsole.h"
 #include "common/translation.h"
 #include "gui/EventRecorder.h"
+#include "gui/gui-manager.h"
+#include "engines/unknown-game-dialog.h"
 #include "engines/advancedDetector.h"
 #include "engines/obsolete.h"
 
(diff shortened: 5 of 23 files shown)


backends/platform/sdl/macosx/macosx_wrapper.h
```diff
@@ -27,6 +27,7 @@
 
 bool hasTextInClipboardMacOSX();
 Common::String getTextFromClipboardMacOSX();
+bool setTextInClipboardMacOSX(const Common::String&);
```
bonki: While optional, I believe we list parameter names in all prototypes.

backends/platform/sdl/sdl.h
```diff
@@ -72,6 +72,7 @@ class OSystem_SDL : public ModularBackend {
 	// Clipboard
 	virtual bool hasTextInClipboard();
 	virtual Common::String getTextFromClipboard();
+	virtual bool setTextInClipboard(const Common::String&);
```
bonki: Same here.

engines/advancedDetector.cpp
```diff
@@ -327,34 +328,57 @@ Common::Error AdvancedMetaEngine::createInstance(OSystem *syst, Engine **engine)
 
 void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFilePropertiesMap &filesProps, const ADGameIdList &matchedGameIds) const {
 	Common::String report = Common::String::format(
+			"The game in '%s' seems to be an unknown %s engine game "
+			"variant.\n\nPlease report the following data to the ScummVM "
+			"team at %s along with the name of the game you tried to add and "
+			"its version, language, etc.:",
+			path.getPath().c_str(), getName(), "https://bugs.scummvm.org/");
+	
+	Common::String reportTranslated = Common::String::format(
```
bonki: You can avoid duplication by having the format string in a separate variable.

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,133 @@
...
+		break;
+	}
+	case kClose:
+	{
+		close();
+		break;
+	}
+	case kOpenBugtrackerURL:
+	{
+		g_system->openUrl("https://bugs.scummvm.org/newticket?summary=[UNK] Unknown game for engine " + _bugtrackerAffectedEngine + ": &description=" + _bugtrackerGameData + "&component=Engine%3A+"+_bugtrackerAffectedEngine+"&type=enhancement"+"&keywords=unknown-game "+_bugtrackerAffectedEngine);
```
bonki: Using `String::format()` would make this a bit more readable.

engines/advancedDetector.cpp
```diff
@@ -327,34 +328,57 @@ Common::Error AdvancedMetaEngine::createInstance(OSystem *syst, Engine **engine)
...
 			_("The game in '%s' seems to be an unknown %s engine game "
 			  "variant.\n\nPlease report the following data to the ScummVM "
 			  "team at %s along with the name of the game you tried to add and "
 			  "its version, language, etc.:"),
-			path.getPath().c_str(), getName(), "https://bugs.scummvm.org/");
-
+			  path.getPath().c_str(), getName(), "https://bugs.scummvm.org/");
+	
+	Common::String bugtrackerAffectedEngine = Common::String::format(getName());
```
bonki: No need for `format()` here.

rootfather:
@bonki You are right about the button positions, this was an oversight be me. What do you think about the following?
```
	//Check if we have clipboard functionality
	if (g_system->hasFeature(OSystem::kFeatureClipboardSupport)) {
	   new ButtonWidget(this, copyToClipboardButtonPos, _h - buttonHeight - 8, copyToClipboardButtonWidth, buttonHeight, _("Copy to clipboard"), 0, kCopyToClipboard);
	} else {
		   //Set the bug tracker position to the one of the disabled copy-to-clipboard button
		   openBugtrackerURLButtonPos = copyToClipboardButtonPos;
	}
```

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,133 @@
...
+	int copyToClipboardButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Copy to clipboard")) + 10);
+	int openBugtrackerURLButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Report game")) + 10);
+	int closeButtonPos = _w - closeButtonWidth - 10;
+	int copyToClipboardButtonPos = _w - copyToClipboardButtonWidth - closeButtonWidth - 15;
+	int openBugtrackerURLButtonPos = _w - copyToClipboardButtonWidth - closeButtonWidth - openBugtrackerURLButtonWidth - 20;
+	_closeButton = new ButtonWidget(this, closeButtonPos, _h - buttonHeight - 8, buttonWidth, buttonHeight, _("Close"), 0, kClose);
+
+	//Check if we have clipboard functionality
+	if (g_system->hasFeature(OSystem::kFeatureClipboardSupport)) {
+	   new ButtonWidget(this, copyToClipboardButtonPos, _h - buttonHeight - 8, copyToClipboardButtonWidth, buttonHeight, _("Copy to clipboard"), 0, kCopyToClipboard);
```
bonki: Sneaky spaces; should be a tab :)

bonki:
@rootfather Not particularly pretty but it gets the job done as long as no new buttons are added to the dialog.

engines/advancedDetector.cpp
rootfather: I don't know if this could be avoided since the message contains strings that could not be translated if I simply tried to do something like reportTranslated = _(report)...

engines/unknown-game-dialog.cpp
rootfather: fixed.

engines/advancedDetector.cpp
bonki: I am talking about the format string only, something along the lines of:

    const char *reportFormat = Common::String("The game in '%s' ...");
    Common::String report           = Common::String::format(  reportFormat,  ...);
    Common::String reportTranslated = Common::String::format(_(reportFormat), ...)

engines/advancedDetector.cpp
```diff
@@ -327,34 +328,57 @@ Common::Error AdvancedMetaEngine::createInstance(OSystem *syst, Engine **engine)
...
-
+		reportTranslated += Common::String::format("  {\"%s\", 0, \"%s\", %d},\n", file->_key.c_str(), file->_value.md5.c_str(), file->_value.size);
+	}
 	report += "\n";
-
+	reportTranslated += "\n";
+	
+	// Show the UnknownGameDialog and print the translated unknown game information
+	GUI::UnknownGameDialog dialog(report, reportTranslated, bugtrackerAffectedEngine);
+	dialog.runModal();
```
bgK: I wonder what happens when invoking the detector from the command line using say `--detect`.

bonki: Hm, apparently it quits without printing anything. Unless I'm doing something wrong that's not quite ideal.

rootfather: Erm... it segfaults...

engines/advancedDetector.cpp
rootfather: Thanks! Fixed.

engines/advancedDetector.cpp
rootfather: To be precise, it segfaults when the detector is invoked from the command line. @bgK do you know how to check if the GUI is currently running? I tried to check for `GUI::GuiManager::instance()._launched == true` and make this the condition to invoke GUI::UnknownGameDialog, but that doesn't work...

engines/unknown-game-dialog.cpp
rootfather: Fixed!

engines/advancedDetector.cpp
rootfather: Fixed in b711c1e (code by @criezy)

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,144 @@
...
+	const int screenH = g_system->getOverlayHeight();
+
+	int buttonWidth = g_gui.xmlEval()->getVar("Globals.Button.Width", 0);
+	int buttonHeight = g_gui.xmlEval()->getVar("Globals.Button.Height", 0);
+
+	//Calculate the size the dialog needs
+	Common::Array<Common::String> lines;
+	int maxlineWidth = g_gui.getFont().wordWrapText(_reportTranslated, screenW - 2 * 20, lines);
+
+	_w = MAX(MAX(maxlineWidth, 0), (2 * buttonWidth) + 10) + 20;
```
criezy: You probably want to use `3 * buttonWidth` since you can have up to three buttons here. Or you could adapt it depending on the available features to not force having space for a button that will not be created.

Also you probably want to move the button widths computation above so that you can use it here since it can be wider than buttonWidth. This would end up looking like this:
```
	
	int closeButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Close")) + 10);
	int copyToClipboardButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Copy to clipboard")) + 10);
	int openBugtrackerURLButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Report game")) + 10);
	int totalButtonWidth = closeButtonWidth;
	if (g_system->hasFeature(OSystem::kFeatureClipboardSupport))
		totalButtonWidth += 10 + copyToClipboardButtonWidth;
	if (g_system->hasFeature(OSystem::kFeatureOpenUrl))
		totalButtonWidth += 10 + openBugtrackerURLButtonWidth;

	_w = MAX(MAX(maxlineWidth, 0), totalButtonWidth) + 20;
```

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include "common/translation.h"
+#include "common/str-array.h"
+#include "gui/gui-manager.h"
+#include "gui/message.h"
+#include "gui/ThemeEval.h"
+#include "gui/widgets/popup.h"
+#include "engines/unknown-game-dialog.h"
+#include "backends/platform/sdl/sdl.h"
+
+namespace GUI {
```
criezy: I know you initially added the file to `gui/`, and it made sense to use the GUI namespace then, but since now the file is in `engines/` I don't think you should use the GUI namespace. Looking at the other dialogs in `engines/dialogs.h` they are in the global namespace.

So I would suggest to do the same and remove this line, then remove the closing bracket at the end of this file, and do the same in unknown-game-dialog.h.

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,144 @@
...
+			openBugtrackerURLButtonPos = copyToClipboardButtonPos;
+	}
+
+	//Check if we have support for opening URLs
+	if (g_system->hasFeature(OSystem::kFeatureOpenUrl)) {
+		new ButtonWidget(this, openBugtrackerURLButtonPos, _h - buttonHeight - 8, openBugtrackerURLButtonWidth, buttonHeight, _("Report game"), 0, kOpenBugtrackerURL);
+		//Formatting the reportData for bugtracker submission [replace line breaks]...
+		_bugtrackerGameData = _reportData;
+		while (_bugtrackerGameData.contains("\n")) {
+			replace(_bugtrackerGameData, "\n", "%0A");
```
criezy: I am confused as I would not expect this to compile (unless there are `using namespace Common;` directives scattered around). Shouldn't this line be:
```
	Common::replace(_bugtrackerGameData, "\n", "%0A");
```

criezy:
Since I created the ticket on our bug tracker, you can guess that I think this is a good idea :-P

In addition to the comments I made in the review, I am also wondering if we should use a ScrollContainerWidget in the dialog in case the report message is too big to be displayed on the screen.

Kirben:
There are a few problems with the feature:
1. ScummVM still generates way too many false unknown versions, have you actually tried a mass add with this feature enabled? I currently get around 25 false positives in console window.
2. Auto generated bug report are a bad idea, as we need contact details to confirm the game is actually working after details for that specific game version have been added.
3. If the bug report system formatting changes in the future, the report option gets broken.

backends/platform/sdl/sdl.cpp
```diff
@@ -502,17 +503,42 @@ Common::String OSystem_SDL::getTextFromClipboard() {
 
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	char *text = SDL_GetClipboardText();
+#ifdef USE_TRANSLATION
```
bgK: It is my understanding we use `ISO-8859-1` when `USE_TRANSLATION` is not defined. It's probably better to convert to that rather than using raw UTF-8 data which is bound to fail to render properly.
(the same applies to `setTextInClipboard`)

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,144 @@
...
+		break;
+	}
+	case kClose:
+	{
+		close();
+		break;
+	}
+	case kOpenBugtrackerURL:
+	{
+		Common::String bugtrackerURL = Common::String::format((
```
bgK: This code will make its way into ResidualVM which does not use a Trac bugtracker. I would appreciate if the code used to build the bugtracker URL was moved to a separate function so it can be overridden easily.

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+	_w = MAX(MAX(maxlineWidth, 0), (2 * buttonWidth) + 10) + 20;
+	int lineCount = lines.size() + 1;
+
+	_h =  3 * kLineHeight;
+	_h += lineCount * kLineHeight;
+
+	// Center the dialog
+	_x = (screenW - _w) / 2;
+	_y = (screenH - _h) / 2;
```
bgK: Centering the dialog needs to happen in the `reflowLayout` method so the dialog stays centered when resizing the window.

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+	// Buttons
+	int closeButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Close")) + 10);
+	int copyToClipboardButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Copy to clipboard")) + 10);
+	int openBugtrackerURLButtonWidth = MAX(buttonWidth, g_gui.getFont().getStringWidth(_("Report game")) + 10);
+	int closeButtonPos = _w - closeButtonWidth - 10;
+	int copyToClipboardButtonPos = _w - copyToClipboardButtonWidth - closeButtonWidth - 15;
+	int openBugtrackerURLButtonPos = _w - copyToClipboardButtonWidth - closeButtonWidth - openBugtrackerURLButtonWidth - 20;
+	_closeButton = new ButtonWidget(this, closeButtonPos, _h - buttonHeight - 8, buttonWidth, buttonHeight, _("Close"), 0, kClose);
```
bgK: There's a bunch of manual GUI widget placement in this method. Did you consider describing the dialog in a layout file to automate this?

engines/advancedDetector.cpp
```diff
@@ -326,36 +328,54 @@ Common::Error AdvancedMetaEngine::createInstance(OSystem *syst, Engine **engine)
...
+	reportTranslated += "\n";
 
-	g_system->logMessage(LogMessageType::kInfo, report.c_str());
+	// Write the original message about the unknown game to the log file
+	Common::String reportLog = report;
+	reportLog.wordWrap(80);
+	g_system->logMessage(LogMessageType::kInfo, reportLog.c_str());
+
+	// Check if the GUI is running, show the UnknownGameDialog and print the translated unknown game information
+	if (GUI::GuiManager::hasInstance() && g_gui.isActive()) {
```
bgK: Using `g_gui` here will instantiate the GuiManager singleton. I don't think it's intended.

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+	y += kLineHeight;
+}
+
+void UnknownGameDialog::handleCommand(CommandSender *sender, uint32 cmd, uint32 data) {
+	switch(cmd) {
+	case kCopyToClipboard:
+	{
+		g_system->setTextInClipboard(_reportData);
+		MessageDialog dialog(_("All necessary information about your game has been copied into the clipboard"), _("OK"));
```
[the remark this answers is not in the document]
criezy: Good idea!

engines/unknown-game-dialog.cpp
rootfather: I haven't looked at the layouting/theming yet...

engines/unknown-game-dialog.cpp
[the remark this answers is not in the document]
rootfather: Indeed!

rootfather:
@Kirben 
> Auto generated bug report are a bad idea, as we need contact details to confirm the game is actually working after details for that specific game version have been added.

I see your point, so we should keep the requirement for the user to be logged in into the Trac instance. This ensures that we have a proper way to contact the submitter. This also means that just the /login redirect needs to be fixed.

> If the bug report system formatting changes in the future, the report option gets broken.

Yep, that's true - maintaining this feature requires to change the URL building method if required.

@bgK 
> ScummVM invokes the detector each time a user starts a game. I think this dialog should only be shown when adding games.

I don't know *why*, but at the moment, the dialog is only shown when adding the games, not when starting them. After starting a game from the launcher, the game information is shown in the log window, but the dialog is not invoked...

> Multiple engines may report unknown files in a single folder. This means the user will be presented with the dialog multiple times.
The dialog interrupts the mass detector instead of presenting the results once all the directories have been recursively scanned.

That's true, this should be refined.

> Games not using the AdvancedDetector don't show the new dialog which may be confusing for our users.

This means that some of the changes made to AdvancedDetector have to be applied to the engines that rely on own detectors instead of AdvancedDetector?

 > In my opinion there is a separation of concerns issue with the implementation. The detector should return the detection results to the caller (maybe it found known games, maybe it found unknown files, maybe it found both, maybe it found nothing). The caller (the add button, mass add, launch game, the command line, ...) should deal with the result in the appropriate way. Show a dialog if the GUI is active; accumulate results for the mass add button, show a console report for the command line.

This might actually be out of scope for me know...

engines/unknown-game-dialog.cpp
rootfather: That can be done. Simply moving the URL building part into a separate function and then calling this function in case of kOpenBugtrackerURL?

engines/unknown-game-dialog.cpp
rootfather: It compiled, but I changed this to your convenience ;)

engines/unknown-game-dialog.cpp
```diff
@@ -0,0 +1,147 @@
...
+	_y = (g_system->getOverlayHeight() - _h) / 2;
+	GUI::Dialog::reflowLayout();
+}
+
+void UnknownGameDialog::handleCommand(CommandSender *sender, uint32 cmd, uint32 data) {
+	switch(cmd) {
+	case kCopyToClipboard:
+	{
+		g_system->setTextInClipboard(_reportData);
+		g_system->displayMessageOnOSD(_("All necessary information about your game has been copied into the clipboard"));
```
bgK: `setTextInClipboard` returns a boolean to indicate success or failure. It would be better to only show the success message if putting the text in the clipboard actually worked (An maybe add a failure message).

engines/unknown-game-dialog.cpp
bgK: That would be fine.

rootfather: Does https://gist.github.com/rootfather/222ff1079263ea8073d39b5ac4659a8a matches with what you proposed?

bgK: Yes, sounds good.

sev-:
I recommend to limit this functionality only when 'Add game' button is pressed. Not even Mass Add. Then it will make sense even creating tickets. For creating tickets, however, I would add another field: username, or e-mail address, or both, so we could trace the report back.

backends/platform/sdl/sdl.cpp
criezy: Right. And that would be a very simple change to do (here, and in the macOS backend).

criezy:
@Kirben @sev- There are some good points there.

1. Too many false unknown version
Hopefully limiting this dialog to the case where we Add a game (not when starting a game, and not for Mass Add) as suggested by sev should help with that. And I agree this is a good idea.

2. User contact
Yes, anonymous report is a bad idea. But since the bug tracker requiers users to be logged in to create a new ticket, this PR will not allow that. I am wondering here if requiring a GitHub account is not stopping a lot of reports, and maby we could accept other login bridges (for example using a Facebook account as we have on the translations website) but that is a separate issue. But we should not allow reports without contact information.

3. Bur report system future changes
That is something I am also worried about. I know nothing about websites, but would it be feasible and easy to have a redirect page on the web site that takes a couple of query items (engine and description) and reformat and redirect to whatever is our current bug tracker system? So the dialog would open for example `https://www.scummvm.org/unknowngame?engine=Foo&description=Bar`. Then if we change the bug tracker in the future we would just need to update the website redirect as well?

rootfather:
@criezy:
regarding 1: I'll see what I can do about the Mass Add feature. Currently, the dialog is not shown when a game is started, even if I don't understand *why*.

regarding 3: I know nothing about websites neither, but this seems like a good approach. Then, we won't render the current version of ScummVM unable to send bugreports and only would have to change the website redirect.

base/plugins.cpp
```diff
@@ -524,7 +524,7 @@ GameList EngineManager::detectGames(const Common::FSList &fslist) const {
 		// Iterate over all known games and for each check if it might be
 		// the game in the presented directory.
 		for (iter = plugins.begin(); iter != plugins.end(); ++iter) {
-			candidates.push_back((*iter)->get<MetaEngine>().detectGames(fslist));
+                       candidates.push_back((*iter)->get<MetaEngine>().detectGames(fslist, useUnknownGameDialog));
```
criezy: It looks like you mistakenly replaced tab indentation with spaces on this line.

engines/advancedDetector.cpp
```diff
@@ -373,7 +373,7 @@ void AdvancedMetaEngine::reportUnknown(const Common::FSNode &path, const ADFileP
 	g_system->logMessage(LogMessageType::kInfo, reportLog.c_str());
 
 	// Check if the GUI is running, show the UnknownGameDialog and print the translated unknown game information
-	if (GUI::GuiManager::hasInstance() && g_gui.isActive()) {
+       if (GUI::GuiManager::hasInstance() && g_gui.isActive() && useUnknownGameDialog == true) {
 		UnknownGameDialog dialog(report, reportTranslated, bugtrackerAffectedEngine);
```
criezy: Here as well the indentation seems to have been changed to spaces.

Also the test on `GUI::GuiManager::hasInstance() && g_gui.isActive()` is no longer necessary since I don't see a case where they would fail while `useUnknownGameDialog` is true. But I don't see any issue with keeping those to be extra-safe.

Note that the `== true` is not necessary here either, but since our code formatting conventions don't say anything about this, I guess this is fine.

engines/advancedDetector.cpp
```diff
@@ -574,7 +574,7 @@ ADGameDescList AdvancedMetaEngine::detectGame(const Common::FSNode &parent, cons
 	// We didn't find a match
 	if (matched.empty()) {
 		if (!filesProps.empty() && gotAnyMatchesWithAllFiles) {
-			reportUnknown(parent, filesProps, matchedGameIds);
+                       reportUnknown(parent, filesProps, matchedGameIds, useUnknownGameDialog);
```
criezy: More tabs that magically transformed into spaces...

engines/advancedDetector.h
```diff
@@ -278,7 +278,7 @@ class AdvancedMetaEngine : public MetaEngine {
 
 	virtual GameDescriptor findGame(const char *gameId) const;
 
-	virtual GameList detectGames(const Common::FSList &fslist) const;
+       virtual GameList detectGames(const Common::FSList &fslist, bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

engines/advancedDetector.h
```diff
@@ -313,7 +313,7 @@ class AdvancedMetaEngine : public MetaEngine {
 	 * @param extra		restrict results to specified extra string (only if kADFlagUseExtraAsHint is set)
 	 * @return	list of ADGameDescription pointers corresponding to matched games
 	 */
-	virtual ADGameDescList detectGame(const Common::FSNode &parent, const FileMap &allFiles, Common::Language language, Common::Platform platform, const Common::String &extra) const;
+       virtual ADGameDescList detectGame(const Common::FSNode &parent, const FileMap &allFiles, Common::Language language, Common::Platform platform, const Common::String &extra, bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

engines/advancedDetector.h
```diff
@@ -333,7 +333,7 @@ class AdvancedMetaEngine : public MetaEngine {
 	 * Log and print a report that we found an unknown game variant, together with the file
 	 * names, sizes and MD5 sums.
 	 */
-	void reportUnknown(const Common::FSNode &path, const ADFilePropertiesMap &filesProps, const ADGameIdList &matchedGameIds = ADGameIdList()) const;
+       void reportUnknown(const Common::FSNode &path, const ADFilePropertiesMap &filesProps, const ADGameIdList &matchedGameIds = ADGameIdList(), bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

engines/metaengine.h
```diff
@@ -79,7 +79,7 @@ class MetaEngine : public PluginObject {
 	 * (possibly empty) list of games supported by the engine which it was able
 	 * to detect amongst the given files.
 	 */
-	virtual GameList detectGames(const Common::FSList &fslist) const = 0;
+       virtual GameList detectGames(const Common::FSList &fslist, bool useUnknownGameDialog = false) const = 0;
```
criezy: Another indentation issue.

engines/metaengine.h
```diff
@@ -269,7 +269,7 @@ class EngineManager : public Common::Singleton<EngineManager> {
 public:
 	GameDescriptor findGameInLoadedPlugins(const Common::String &gameName, const Plugin **plugin = NULL) const;
 	GameDescriptor findGame(const Common::String &gameName, const Plugin **plugin = NULL) const;
-	GameList detectGames(const Common::FSList &fslist) const;
+       GameList detectGames(const Common::FSList &fslist, bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

engines/scumm/detection.cpp
```diff
@@ -961,7 +961,7 @@ class ScummMetaEngine : public MetaEngine {
 	virtual bool hasFeature(MetaEngineFeature f) const;
 	virtual GameList getSupportedGames() const;
 	virtual GameDescriptor findGame(const char *gameid) const;
-	virtual GameList detectGames(const Common::FSList &fslist) const;
+       virtual GameList detectGames(const Common::FSList &fslist, bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

engines/sky/detection.cpp
```diff
@@ -79,7 +79,7 @@ class SkyMetaEngine : public MetaEngine {
 	virtual GameList getSupportedGames() const;
 	virtual const ExtraGuiOptions getExtraGuiOptions(const Common::String &target) const;
 	virtual GameDescriptor findGame(const char *gameid) const;
-	virtual GameList detectGames(const Common::FSList &fslist) const;
+       virtual GameList detectGames(const Common::FSList &fslist, bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

engines/sword1/detection.cpp
```diff
@@ -89,7 +89,7 @@ class SwordMetaEngine : public MetaEngine {
 	virtual bool hasFeature(MetaEngineFeature f) const;
 	virtual GameList getSupportedGames() const;
 	virtual GameDescriptor findGame(const char *gameid) const;
-	virtual GameList detectGames(const Common::FSList &fslist) const;
+       virtual GameList detectGames(const Common::FSList &fslist, bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

engines/sword2/sword2.cpp
```diff
@@ -95,7 +95,7 @@ class Sword2MetaEngine : public MetaEngine {
 	virtual GameList getSupportedGames() const;
 	virtual const ExtraGuiOptions getExtraGuiOptions(const Common::String &target) const;
 	virtual GameDescriptor findGame(const char *gameid) const;
-	virtual GameList detectGames(const Common::FSList &fslist) const;
+       virtual GameList detectGames(const Common::FSList &fslist, bool useUnknownGameDialog = false) const;
```
criezy: Another indentation issue.

gui/launcher.cpp
```diff
@@ -573,7 +573,7 @@ bool LauncherDialog::doGameDetection(const Common::String &path) {
 
 	// ...so let's determine a list of candidates, games that
 	// could be contained in the specified directory.
-	GameList candidates(EngineMan.detectGames(files));
+       GameList candidates(EngineMan.detectGames(files, true));
```
criezy: Another indentation issue.

criezy:
I will try to summarise ideas and issues that were discussed above.

The short version of it is that it seems to me that once the incorrect indentation have been fixed, and maybe the creation of the `Report game` button has been disabled, this code is good to be merged. Is there any objections? Do you see any other pending issues not listed below?

1. **Formatting of clipboard-related commits:**

I have cleaned the clipboard-related commits in my fork (you can see them at https://github.com/criezy/scummvm/commits/unknown-game-dialog). I am planning to use those to replace the ones in this pull request.

2. **ScummVM still generates way too many false unknown versions**
**ScummVM invokes the detector each time a user starts a game**
**The dialog interrupts the mass detector**

With the latest changes from @rootfather the dialog only appears when using the `Add Game` button. This means it is not used with `Mass Add`, when detecting from the command line, or when starting a game. This I think fixes the main issue that was highlighted by several people.

We could consider using the dialog for Mass Add as well if we can get it to only popup once at the end with all the unknown version found. But I think this is out of scope of this PR.

3. **we need contact details to confirm the game is actually working**
**If the bug report system formatting changes in the future, the report option gets broken**

As previously noted, the bug tracker forces users to be logged in to create a ticket, so we do get contact details.

I am wondering however if we should `#if 0` the code that creates the `Report game` button and add a comment for now until we can figure out how to handle the URL in a nice way to be more future proof.

4. **Multiple engines may report unknown files in a single folder. This means the user will be presented with the dialog multiple times**
**The detector should return the detection results to the caller**

Yes, I think it would be a good idea to return unknown version information to the caller and not just detected games, and let the caller handle this information as it sees fit instead of reporting those from the detection code. However I think this is out of scope of this PR.

5. **Use a ScrollContainerWidget**

I propose to work on that once this PR has been merged.

criezy:
I have now merged manually those changes after cleaning some commits and squashing most of them.
I have also added a couple of commits to use a ScrollContainerWidget and to properly resize and relayout the dialog (and not just center it) when the overlay size changes.

Thank you for your work on this.

PR #1162, opened 2018-04-19
BASE: Allow options to come after the command on the command line

bgK:


sev-:
It is not clear what exactly it does. Could you please elaborate with examples?

bgK:
@criezy Good catch, I fixed `--list-saves` to behave more like the other commands with regards to entering multiple commands. The rest of your proposed changes sound right but are, I think, out of the scope of this pull request.

base/commandLine.cpp
```diff
@@ -455,7 +458,7 @@ Common::String parseCommandLine(Common::StringMap &settings, int argc, const cha
 				// FIXME: Need to document this.
 				// TODO: Make the argument optional. If no argument is given, list all saved games
 				// for all configured targets.
-				return "list-saves";
```
criezy: For consistency you should probably check that `command` is empty before setting it here and if it is not display an error as done in the DO_COMMAND.

bgK:
Thanks for your reviews, I merged this manually.

PR #1163, opened 2018-04-21
MOHAWK: Myst: Selentic receiver improvements

dafioram:
1. Slow down sound receiver turn rate at maximum speed so it can actually be possible to stop close to a desired angle.
2. The receiver direction hint is now given when < 5 degrees away rather than when <= 5 degrees away. This has it line up with the original.

I think the sound receiver rotating could probably be improved further as it is jumpy, but at least now it is more usable. 

I've also seen the directional hint behavior sporadic and I will try to see if I can repo that. Sometimes the hint stop blinking and such.


bgK:
Thanks for your contribution. I merged this manually with an additional fix.

PR #1164, opened 2018-04-22
Improve command line --list-saves command

criezy:
This include two changes:
- Change the syntax of the command to use `--game=TARGET` to specify the target, and thus be consistent with other commands.
- Allow listing save games for all known targets (when no target is specified by the user).


criezy:
@bgK Thank you for your comments. I think all the issues you noticed should be fixed now.

The one thing I am wondering is whether it is worth having some additional code to still parse correctly `--list-saves=TARGET` (without documenting that it still works) for backward compatibility with tools that might use it.

bonki:
Personally, I don't think that's necessary.

PR #1165, opened 2018-04-23
MOHAWK: MYST: Turn held page state into an enum

dafioram:
Previously, the held page stage was an unsigned int 16
with values 0-13. The enum will make its state more clear.

Because enums are implicitly ints in c++03 and _global->heldPage
is an uint16 I have to typecast to uint16 when ever I compare
the enum. This is not ideal, but I still think the enum with
the cast is more clear.

I have also noted interesting behavior as TODO's.

### DIFF
--- a/engines/mohawk/myst.cpp
+++ b/engines/mohawk/myst.cpp
@@ -1196,8 +1196,8 @@ bool MohawkEngine_Myst::canSaveGameStateCurrently() {
 }
 
 void MohawkEngine_Myst::dropPage() {
-	uint16 page = _gameState->_globals.heldPage;
-	bool whitePage = page == 13;
+	HeldPage page = _gameState->_globals.heldPage;
+	bool whitePage = page == kWhitePage;
 	bool bluePage = page - 1 < 6;
 	bool redPage = page - 7 < 6;
 
@@ -1205,24 +1205,24 @@ void MohawkEngine_Myst::dropPage() {
 	_sound->playEffect(800);
 
 	// Drop page
-	_gameState->_globals.heldPage = 0;
+	_gameState->_globals.heldPage = kNoPage;
 
 	// Redraw page area
 	if (whitePage && _gameState->_globals.currentAge == 2) {
 		_scriptParser->toggleVar(41);
 		redrawArea(41);
 	} else if (bluePage) {
-		if (page == 6) {
+		if (page == kBlueFirePlacePage) {
 			if (_gameState->_globals.currentAge == 2)
 				redrawArea(24);
 		} else {
 			redrawArea(103);
 		}
 	} else if (redPage) {
-		if (page == 12) {
+		if (page == kRedFirePlacePage) {
 			if (_gameState->_globals.currentAge == 2)
 				redrawArea(25);
-		} else if (page == 10) {
+		} else if (page == kRedStoneshipPage) { //TODO: Should this also happen for the blue page above?
 			if (_gameState->_globals.currentAge == 1)
 				redrawArea(35);
 		} else {
--- a/engines/mohawk/myst.cpp
+++ b/engines/mohawk/myst.cpp
@@ -1222,7 +1222,7 @@ void MohawkEngine_Myst::dropPage() {
 		if (page == kRedFirePlacePage) {
 			if (_gameState->_globals.currentAge == 2)
 				redrawArea(25);
-		} else if (page == kRedStoneshipPage) { //TODO: Should this also happen for the blue page above?
+		} else if (page == kRedStoneshipPage) {
 			if (_gameState->_globals.currentAge == 1)
 				redrawArea(35);
 		} else {
--- a/engines/mohawk/myst_stacks/myst.cpp
+++ b/engines/mohawk/myst_stacks/myst.cpp
@@ -1138,16 +1138,14 @@ void Myst::o_bookGivePage(uint16 var, const ArgumentsArray &args) {
 	debugC(kDebugScript, "Card Id (Book Cover): %d", cardIdBookCover);
 	debugC(kDebugScript, "SoundId (Add Page): %d", soundIdAddPage);
 
-	// No page or white page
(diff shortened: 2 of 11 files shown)


bonki:
Our code formatting conventions require that there be no whitespace after a cast.

Then again, why not change `_globals.heldPage` to be of type `HeldPage`?

dafioram:
I can make that change in /engines/mohawk/myst_state.h#L84, but how would I make that work with syncAsUint16LE() in engines/mohawk/myst_state.cpp#L313?

bonki:
IIRC the serializer's `syncAs`-methods are templated so it should just work.

engines/mohawk/myst_stacks/myst.cpp
```diff
@@ -1148,40 +1148,40 @@ void Myst::o_bookGivePage(uint16 var, const ArgumentsArray &args) {
 	uint16 mask = 0;
 
 	switch (_globals.heldPage) {
```
bonki: This now needs an empty default case or will otherwise generate a compiler warning because of unhandled enum values.

bonki: Indeed, maybe L1142-44 should be merged into the switch.

dafioram:
Done.

engines/mohawk/myst_state.cpp
```diff
@@ -121,13 +121,13 @@ bool MystGameState::load(int slot) {
 
 	// Set our default cursor
 	_vm->_cursor->showCursor();
-	if (_globals.heldPage == 0 || _globals.heldPage > 13)
+	if (_globals.heldPage == kNoPage || _globals.heldPage > kWhitePage) // TODO: This second condition can never happen
```
bgK: Please remove the condition since it can't happen.

engines/mohawk/myst.cpp
```diff
@@ -1196,33 +1196,33 @@ bool MohawkEngine_Myst::canSaveGameStateCurrently() {
...
 		} else {
 			redrawArea(103);
 		}
 	} else if (redPage) {
-		if (page == 12) {
+		if (page == kRedFirePlacePage) {
 			if (_gameState->_globals.currentAge == 2)
 				redrawArea(25);
-		} else if (page == 10) {
+		} else if (page == kRedStoneshipPage) { //TODO: Should this also happen for the blue page above?
```
bgK: This code is checked against the disassembly, so no, it does not need to happen in other cases.

engines/mohawk/myst_stacks/myst.cpp
```diff
@@ -1138,50 +1138,52 @@ void Myst::o_bookGivePage(uint16 var, const ArgumentsArray &args) {
...
-		_vm->changeToCard(cardIdBookCover, kTransitionDissolve);
-		return;
-	}
-
 	uint16 bookVar = 101;
 	uint16 mask = 0;
 
 	switch (_globals.heldPage) {
-	case 7:
+	case kNoPage:
```
bgK: Here you can collapse the kNoPage and kWhitePage cases. Also the `break` can be removed since `return` already exits the function.

dafioram:
Done. 

Ignoring the fireplace pages, any idea why the kRedStoneshipPage is redrawn differently then the other red pages when it is dropped and the current age == 1? While the blue page in the stoneship age redrawArea is handled like the other blue pages being dropped?

engines/mohawk/myst_stacks/myst.cpp
```diff
@@ -1138,50 +1138,49 @@ void Myst::o_bookGivePage(uint16 var, const ArgumentsArray &args) {
...
+		// fallthrough
```
bgK: No need to put a `fallthrough` marker if there is no code in the case. It's used to tell the compiler we didn't forget a break. Here it's obvious the fallthrough is intentional.

engines/mohawk/myst.cpp
```diff
@@ -1196,33 +1196,34 @@ bool MohawkEngine_Myst::canSaveGameStateCurrently() {
...
 		} else {
 			redrawArea(103);
 		}
 	} else if (redPage) {
-		if (page == 12) {
+		if (page == kRedFirePlacePage) {
 			if (_gameState->_globals.currentAge == 2)
 				redrawArea(25);
-		} else if (page == 10) {
+		} else if (page == kRedStoneshipPage) { // This code was checked against the disassembly and
```
bgK: The whole thing is checked. It does not make sense to add that comment here.

bgK:
> Ignoring the fireplace pages, any idea why the kRedStoneshipPage is redrawn differently then the other red pages when it is dropped and the current age == 1? While the blue page in the stoneship age redrawArea is handled like the other blue pages being dropped?

The red stoneship page case is different because there are three cases. The drawer is closed. The drawer is open but there is no page, the drawer is open and there is a red page. See https://github.com/scummvm/scummvm/blob/c1a12c2475ca90708b8c85ae3157f742870b0614/engines/mohawk/myst_stacks/stoneship.cpp#L268.

engines/mohawk/myst_state.cpp
```diff
@@ -121,13 +121,13 @@ bool MystGameState::load(int slot) {
...
+	if (_globals.heldPage == kNoPage)
 		_vm->setMainCursor(kDefaultMystCursor);
-	else if (_globals.heldPage < 7)
+	else if (_globals.heldPage < kRedLibraryPage) //A blue page is held
 		_vm->setMainCursor(kBluePageCursor);
-	else if (_globals.heldPage < 13)
+	else if (_globals.heldPage < kWhitePage) //A red page is held
 		_vm->setMainCursor(kRedPageCursor);
-	else // if (globals.heldPage == 13)
+	else // if (globals.heldPage == kWhitePage), i.e., the white page is held
```
bonki: Are these comments necessary?

dafioram:
Done.

bgK:
Very good, thanks

PR #1168, opened 2018-04-25
MOHAWK: MYST: Make global currentAge an enum

dafioram:
This global basically tracks what age/book/stack
the player is in.

E.x., instead of _global.currentAge = 2
now _global.currentAge = kMystLibrary.

Interestingly, two integer values (5,8) are
never used.

Also two of the endings are considered Ages
instead of being apart of _global.ending.

### DIFF
--- a/engines/mohawk/myst.cpp
+++ b/engines/mohawk/myst.cpp
@@ -550,18 +550,18 @@ void MohawkEngine_Myst::changeToStack(uint16 stack, uint16 card, uint16 linkSrcS
 
 	switch (_curStack) {
 	case kChannelwoodStack:
-		_gameState->_globals.currentAge = 4;
+		_gameState->_globals.currentAge = kChannelwood;
 		_scriptParser = new MystStacks::Channelwood(this);
 		break;
 	case kCreditsStack:
 		_scriptParser = new MystStacks::Credits(this);
 		break;
 	case kDemoStack:
-		_gameState->_globals.currentAge = 0;
+		_gameState->_globals.currentAge = kSelenitic;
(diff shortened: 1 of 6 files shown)


bgK:
> Can we just call it Age instead.

Well it's not really an age. More like an linking point into an age.

dafioram:
@bgK I have added the two missing states. I know that those 2 states can be explicitly selected for changing stacks, but is this actually done? I think setting currentAge = 5 || 8 is never done directly, but instead something like arg2[0] = 5 || 8 in changeToStack(arg1,arg2) is done directly? I wasn't getting any issues when I didn't have that enum for the intro.

@bonki for variables that track the current age there are some closely related variables like: zipMode, ZipDests, MystScriptParser::_stackMap, MystStack, others? I named the enum to differentiate them from currentAge. That being said it would probably make sense to merge some of these together (the currentAge enum with values of _startCard?). In fact I think MystStack is the same as ActiveAge.

For Page vs. HeldPage there aren't that many page variables. It would probably also be useful to change HeldPage/Pages values to the mask values (mutually exclusive binary) since that gets used to add pages to the various books see globals.redPagesInBook and globals.bluePagesInBook usage in Myst::o_bookGivePage.

bgK:
You're right that the engine does not directly use cases 5 and 8. However the scripts can ask the engine to change to stack 5 or 8 via o_changeStack. The first argument of that command is a value of the new enum type when the current game is Myst (as opposed to the Myst Demo).

MystStack an enum that does not exist in the original engine. It's only used to have the stacks used by all three supported games (Myst, Demo, Making of) in a single location.

PR #1169, opened 2018-04-26
MOHAWK: MYST: Make Dini/Atrus endings an enum

dafioram:
The global variable "ending" tracks various
states that Atrus can be in.

This variable determines what video of
Atrus will be playing and if the brother
books have been destroyed.

Despite this variable being called ending
there are other endings not captured by
this variable (like the brother endings).

Additionally, the enum could also be called DiniState. 

I originally had the 4rth enum called kAtrusReturns since that state is also associated with him returning, but because of it usage in places like Myst::toggleVar were the brother book states were being checked it made sense to give it the kBooksDestroyed name. So I could have a 5th enum value that is called kAtrusReturns and just map that to kBooksDestroyed so that when it is more clear to use one form, kAtrusReturns in Dini, and, kBooksDestroyed in the myst library.

### DIFF
--- a/engines/mohawk/myst_state.h
+++ b/engines/mohawk/myst_state.h
@@ -79,11 +79,21 @@ enum ActiveAge {
 	kMechanical    = 3,
 	kChannelwood   = 4,
 	kIntro         = 5,
-	kDini          = 6,
+	kDni           = 6,
 	kMystStart     = 7,
 	kCredits       = 8,
 	kSirrusEnding  = 9,
 	kAchenarEnding = 10
+};
+
+// Various states that Atrus can be in when in Dni
+enum DniEnding {
+	kDniNotVisited    = 0, // Player hasn't been to Dni/K'veer yet
+	kAtrusWantsPage   = 1, // Player is in Dni with the white page
+	kAtrusLeaves      = 2, // Atrus leaves Dni after receiving the white page
+	kForgotPage       = 3, // Player has entered Dni without bringing the white page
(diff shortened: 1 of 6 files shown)


engines/mohawk/myst_state.h
```diff
@@ -86,6 +86,15 @@ enum ActiveAge {
 	kAchenarEnding = 10
 };
 
+// Various states that Atrus can be in when in Dini
+enum DiniEnding {
```
Deledrius: Typo.  The location is called D'ni, not Dini.  `DniEnding` would be sufficiently readable.

dafioram:
Good catch I made that change as well as a previously commited Dini usage in ActiveAge (kDini->kDni).

engines/mohawk/myst_state.h
```diff
@@ -79,13 +79,22 @@ enum ActiveAge {
...
+	kDni           = 6,
 	kMystStart     = 7,
 	kCredits       = 8,
 	kSirrusEnding  = 9,
 	kAchenarEnding = 10
 };
 
+// Various states that Atrus can be in when in Dni
+enum DniEnding {
+	kAtrusWantsPage   = 1, // Player is in Dni with the white page
```
bgK: Value 0 used when the played has not yet entered K'veer is missing.

dafioram:
Done. 

I'm guessing _globals.ending was set to 0 (its first state) originally through the memset in MystGameState. Now I have explicitly set it to its first state (kDniNotVisited) as well as _globals.heldPage = kNoPage; in the MystGameState constructor.

PR #1171, opened 2018-04-27
Update README.md

JohnnyWalkerDesign:
Added necessary switches otherwise you get a curl not found error

### DIFF
--- a/backends/platform/ios7/README.md
+++ b/backends/platform/ios7/README.md
@@ -20,7 +20,7 @@ The next step is to compile the **create_project** tool. Open the Xcode project 
 Execute the following commands in a terminal:
 ```
 $ cd path_to_the_build_directory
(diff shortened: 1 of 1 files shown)


criezy:
The change is correct.

We do however have  more up to date instructions on our wiki at http://wiki.scummvm.org/index.php/Compiling_ScummVM/iPhone with a link to a more recent libraries package that allows to not disable jpeg, and for example play Myst (which was actually the main reason I updated the package and wrote those instructions on the wiki 5 months ago). I never noticed we have that README.md file as well.

We could either update the content of the file to be in sync with those newer instructions, but it might fall out of sync again. So maybe we should replace them with a link to the wiki.

JohnnyWalkerDesign:
Ah. Yes, I've been struggling with those instructions for an hour or so. If there's better ones elsewhere, best to just link to them I think!

JohnnyWalkerDesign:
FYI: With those other instructions I had to run `sudo xcodebuild` for it to work, otherwise I got errors.

criezy:
I have now updated the instructions on the wiki with additional information and changed the README.md file to point to those.

Thank you for pointing out the issues in the build instructions we had in the README.md file and wiki.

PR #1172, opened 2018-04-28
MOHAWK: MYST: Autosave to Slot 0

dafioram:
The game will autosave to slot 0 if the
player quits using F5. It will not autosave
if the player quits using ctrl+q, ctrl+F5,
or closes the ScummVM window.

This will override any saves the player
puts in save slot 0. It would be preferable
to have slot 0 blocked out when the player
tries to save. I'm not sure how to do that
without also blocking the save from being
loadable.

It would also be good to label the autosave
in the description when the save slots are
pulled up for loading and their is no save
in slot 0.

Its possible the player may quit at a bad spot
and the game will save a bad state. Possibly
more conditions could be added to autoSave()
so that it would not autosave when a bad state
is detected.

It would be even better to add more Autosaves,
but the conditions would be more complicated.
E.x., Autosave 1 could be a save right before
the player went to the next age. Autosave 2
could be a save at the beginning of an Age.

### DIFF
--- a/engines/mohawk/detection.cpp
+++ b/engines/mohawk/detection.cpp
@@ -283,7 +283,8 @@ void MohawkMetaEngine::removeSaveState(const char *target, int slot) const {
 SaveStateDescriptor MohawkMetaEngine::querySaveMetaInfos(const char *target, int slot) const {
 #ifdef ENABLE_MYST
 	if (strstr(target, "myst")) {
-		return Mohawk::MystGameState::querySaveMetaInfos(slot);
+		SaveStateDescriptor desc = Mohawk::MystGameState::querySaveMetaInfos(slot);
+		return desc;
 	}
 #endif
 #ifdef ENABLE_RIVEN
--- a/engines/mohawk/myst.cpp
+++ b/engines/mohawk/myst.cpp
@@ -72,6 +72,7 @@ MohawkEngine_Myst::MohawkEngine_Myst(OSystem *syst, const MohawkGameDescription 
 	_showResourceRects = false;
 	_curStack = 0;
 	_curCard = 0;
+	_lastSaveTime = 0;
 
 	_hoverResource = nullptr;
 	_activeResource = nullptr;
@@ -395,6 +396,11 @@ void MohawkEngine_Myst::doFrame() {
 		_waitingOnBlockingOperation = false;
 	}
 
+	if (shouldPerformAutoSave(_lastSaveTime) && canSaveGameStateCurrently() && _gameState->isAutoSaveAllowed()) {
+		autoSave();
+		_lastSaveTime = _system->getMillis();
+	}
+
 	Common::Event event;
 	while (_system->getEventManager()->pollEvent(event)) {
 		switch (event.type) {
@@ -446,6 +452,9 @@ void MohawkEngine_Myst::doFrame() {
 						}
 
 						if (_needsShowCredits) {
+							if (canSaveGameStateCurrently() && _gameState->isAutoSaveAllowed())
+								// Attempt to autosave before exiting
+								autoSave();
 							if (isInteractive()) {
 								_cursor->hideCursor();
 								changeToStack(kCreditsStack, 10000, 0, 0);
@@ -1159,6 +1168,11 @@ Common::Error MohawkEngine_Myst::loadGameState(int slot) {
 
 Common::Error MohawkEngine_Myst::saveGameState(int slot, const Common::String &desc) {
 	return _gameState->save(slot, desc) ? Common::kNoError : Common::kUnknownError;
+}
+
+void MohawkEngine_Myst::autoSave() {
+	if (!_gameState->save(Mohawk::kAutoSaveSlot, Mohawk::kAutoSaveName))
+		warning("Attempt to autosave has failed.");
 }
 
 bool MohawkEngine_Myst::hasGameSaveSupport() const {
--- a/engines/mohawk/myst_state.cpp
+++ b/engines/mohawk/myst_state.cpp
@@ -41,10 +41,11 @@ MystSaveMetadata::MystSaveMetadata() {
 	saveHour = 0;
 	saveMinute = 0;
 	totalPlayTime = 0;
+	autoSave = false;
 }
 
 bool MystSaveMetadata::sync(Common::Serializer &s) {
-	static const Common::Serializer::Version kCurrentVersion = 1;
+	static const Common::Serializer::Version kCurrentVersion = 2;
 
 	if (!s.syncVersion(kCurrentVersion)) {
 		return false;
@@ -57,6 +58,7 @@ bool MystSaveMetadata::sync(Common::Serializer &s) {
 	s.syncAsByte(saveMinute);
 	s.syncString(saveDescription);
 	s.syncAsUint32LE(totalPlayTime);
+	s.syncAsByte(autoSave, 2);
 
 	return true;
 }
@@ -185,7 +187,9 @@ bool MystGameState::save(int slot, const Common::String &desc) {
 		return false;
 	}
 
-	updateMetadateForSaving(desc);
+	bool isAutoSave = isAutoSaveAllowed() && (slot == Mohawk::kAutoSaveSlot);
+
+	updateMetadateForSaving(desc, isAutoSave);
 
 	return saveMetadata(slot);
 }
@@ -216,7 +220,7 @@ Common::String MystGameState::buildMetadataFilename(int slot) {
 	return Common::String::format("myst-%03d.mym", slot);
 }
 
-void MystGameState::updateMetadateForSaving(const Common::String &desc) {
+void MystGameState::updateMetadateForSaving(const Common::String &desc, bool autosave) {
 	// Update save creation info
 	TimeDate t;
 	g_system->getTimeAndDate(t);
@@ -227,6 +231,7 @@ void MystGameState::updateMetadateForSaving(const Common::String &desc) {
 	_metadata.saveMinute = t.tm_min;
 	_metadata.saveDescription = desc;
 	_metadata.totalPlayTime = _vm->getTotalPlayTime();
+	_metadata.autoSave = autosave;
 }
 
 bool MystGameState::saveMetadata(int slot) {
@@ -251,12 +256,41 @@ bool MystGameState::saveMetadata(int slot) {
 	return true;
 }
 
+bool MystGameState::isAutoSaveAllowed() {
+	// Open autosave slot and see if it an autosave
+	// Autosaving will be enabled if it is an autosave or if there is no save in that slot
+	Common::String filename = buildMetadataFilename(Mohawk::kAutoSaveSlot);
+	Common::InSaveFile *metadataFile = g_system->getSavefileManager()->openForLoading(filename);
+	if (!metadataFile) { // There is no save in the autosave slot, enable autosave
+		return true;
+	}
+
+	Common::Serializer m(metadataFile, nullptr);
+
+	// Read the metadata file
+	Mohawk::MystSaveMetadata metadata;
+	if (!metadata.sync(m)) { // the save in the autosave slot is corrupted, enable autosave
+		delete metadataFile;
+		return true;
+	}
+
+	if (metadata.autoSave == true) {
+		return true;
+	}
+
+	delete metadataFile;
+	return false;
+}
+
 SaveStateDescriptor MystGameState::querySaveMetaInfos(int slot) {
 	// Open the metadata file
 	Common::String filename = buildMetadataFilename(slot);
 	Common::InSaveFile *metadataFile = g_system->getSavefileManager()->openForLoading(filename);
+	SaveStateDescriptor desc;
+	desc.setWriteProtectedFlag(slot == Mohawk::kAutoSaveSlot);
+
 	if (!metadataFile) {
-		return SaveStateDescriptor();
+		return desc;
 	}
 
 	Common::Serializer m(metadataFile, nullptr);
@@ -265,19 +299,20 @@ SaveStateDescriptor MystGameState::querySaveMetaInfos(int slot) {
 	Mohawk::MystSaveMetadata metadata;
 	if (!metadata.sync(m)) {
 		delete metadataFile;
-		return SaveStateDescriptor();
+		return desc;
 	}
 
 	// Set the save description
-	SaveStateDescriptor desc;
 	desc.setDescription(metadata.saveDescription);
 	desc.setSaveDate(metadata.saveYear, metadata.saveMonth, metadata.saveDay);
 	desc.setSaveTime(metadata.saveHour, metadata.saveMinute);
 	desc.setPlayTime(metadata.totalPlayTime);
+	
+	desc.setDeletableFlag(slot != Mohawk::kAutoSaveSlot);
 	Graphics::Surface *thumbnail;
 	if (!Graphics::loadThumbnail(*metadataFile, thumbnail)) {
 		delete metadataFile;
-		return SaveStateDescriptor();
+		return desc;
 	}
 	desc.setThumbnail(thumbnail);
 
--- a/engines/mohawk/myst_state.h
+++ b/engines/mohawk/myst_state.h
@@ -47,6 +47,8 @@ struct MystSaveMetadata {
 
 	uint32 totalPlayTime;
 
+	bool autoSave;
+
 	Common::String saveDescription;
 
 	MystSaveMetadata();
@@ -106,6 +108,7 @@ public:
 
 	bool load(int slot);
 	bool save(int slot, const Common::String &desc);
+	bool isAutoSaveAllowed();
 	static void deleteSave(int slot);
 
 	void addZipDest(uint16 stack, uint16 view);
@@ -340,7 +343,7 @@ private:
 	bool loadState(int slot);
 	void loadMetadata(int slot);
 	bool saveState(int slot);
-	void updateMetadateForSaving(const Common::String &desc);
+	void updateMetadateForSaving(const Common::String &desc, bool autosave);
 	bool saveMetadata(int slot);
 
 	// The values in these regions are lists of VIEW resources
--- a/engines/mohawk/myst.h
+++ b/engines/mohawk/myst.h
@@ -96,6 +96,8 @@ enum TransitionType {
 };
 
 const uint16 kMasterpieceOnly = 0xFFFF;
+const int kAutoSaveSlot = 0;
+const Common::String kAutoSaveName = "Autosave";
 
 struct MystCondition {
 	uint16 var;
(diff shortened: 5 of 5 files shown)


engines/mohawk/myst.cpp
```diff
@@ -1161,6 +1164,12 @@ Common::Error MohawkEngine_Myst::saveGameState(int slot, const Common::String &d
 	return _gameState->save(slot, desc) ? Common::kNoError : Common::kUnknownError;
 }
 
+void MohawkEngine_Myst::autoSave() {
+	int autoSaveSlot = 0;
+	const Common::String autoSaveName = "Autosave - Current";
```
bonki: Most engines just use "Autosave" or "Autosave <slotnumber>".

engines/mohawk/myst.cpp
```diff
@@ -1161,6 +1164,12 @@ Common::Error MohawkEngine_Myst::saveGameState(int slot, const Common::String &d
...
+	Common::Error saveError = saveGameState(autoSaveSlot, autoSaveName);
```
bonki: Why save `saveError` and not use it?

engines/mohawk/myst.cpp
```diff
@@ -1161,6 +1164,12 @@ Common::Error MohawkEngine_Myst::saveGameState(int slot, const Common::String &d
 	return _gameState->save(slot, desc) ? Common::kNoError : Common::kUnknownError;
 }
 
+void MohawkEngine_Myst::autoSave() {
+	int autoSaveSlot = 0;
```
bonki: This probably should be a constant so it can be used in other places when necessary.

bonki:
> It would be preferable to have slot 0 blocked out when the player tries to save. I'm not sure how to do that without also blocking the save from being loadable.

You can write-protect a save slot in `querySaveMetaInfos` via `SaveStateDescriptor`'s `setWriteProtectedFlag()`. You can also protect it from user deletion via `setDeletableFlag()`.

E.g.

    if (slot == kAutoSaveSlot)
        desc.setWriteProtectedFlag(slot == kAutoSaveSlot);
        desc.setDeletableFlag(slot != kAutoSaveSlot);
    }

bonki:
What you need to do is keep track of when the last autosave occurred (`_lastSaveTime = _system->getMillis()`) and somewhere appropriate inside the main game loop call

    if (shouldPerformAutoSave(_lastSaveTime)) {
        autoSave();
    }

`shouldPerformAutoSave()` respects the `autosave_period` config setting (the autosave interval from Misc->Autosave in the GUI) and returns `true` if and only if the time difference is big enough.

You should also check `MohawkEngine_Myst::canSaveGameStateCurrently()` before attempting to write a savegame to make sure that the game is in a state where it is legal to save.

In TUCKER I also check if there exists a savegame in the autosave slot which is not an actual autosave but other than that all engines which autosave do pretty much the same.

dafioram:
I'm not quite sold on using time based autosave vs. just saving when the player quits (sort of like a checkpoint autosave). 

If we use the time based autosave then the player may have to solve some puzzles they have already solved if the autosave was a while back.  With the quiting autosave it is possible to save at an undesirable time (middle of a animation?) and I only have it working when the player quits using F5.

In myst their are very few states for the save (so maybe a checkpoint system would make the most sense), I think it is mostly the current age, number of pages in each book and if they are holding a page or not.

bonki:
> If we use the time based autosave then the player may have to solve some puzzles they have already solved if the autosave was a while back.

I'm not proposing to replace exit-based autosave with interval-based autosave but to have both (mostly because interval-based autosave is what all other engines use and there's a GUI setting for that). Can you explain how having interval-based autosaves would force the player to re-solve puzzles (serious question, I never played Myst)?
If that really is an issue I agree that a checkpoint system might be better. @bgK?

> With the quiting autosave it is possible to save at an undesirable time (middle of a animation?)

That's exactly why `canSaveGameStateCurrently()` exists, unless I misunderstand you.

dafioram:
If there is a save in slot 0 how do we know if it was a previous save by the player or created by an autosave?

bonki:
In TUCKER I added a `flags` field to the savegame metadata and mark autosaves as such, see https://github.com/scummvm/scummvm/blob/master/engines/tucker/saveload.cpp#L252 and https://github.com/scummvm/scummvm/blob/master/engines/tucker/saveload.cpp#L281.

dafioram:
So this would bump up the savedgame version (header.version) since new saves would have a flag indicating whether or not it is an autosave and we would assume saved games without that flag are not. Correct?

I think autosaving every few seconds is more than good enough, perhaps that is too frequent (frequent disk IO)?

If the user has a previous save in slot0 are they are prevented from saving over it after the autosave feature? I think it makes sense to lock that slot to saving. In my current implementation the autosave slot is not labeled as an autosave slot until a saved game is there, It just says "Untitled Saved Game" What am I missing?

bonki:
> So this would bump up the savedgame version (header.version) since new saves would have a flag indicating whether or not it is an autosave and we would assume saved games without that flag are not. Correct?

Exactly!

> I think autosaving every few seconds is more than good enough, perhaps that is too frequent (frequent disk IO)?

As I said, this is not something you need and should keep track of yourself, `shouldPerformAutoSave()` does that for you and respects the user setting.

> If the user has a previous save in slot0 are they are prevented from saving over it after the autosave feature? I think it makes sense to lock that slot to saving. In my current implementation the autosave slot is not labeled as an autosave slot until a saved game is there, It just says "Untitled Saved Game" What am I missing?

Could you rephrase your question?

dafioram:
Alright new update. It now autosaves every 5 secs rather than when the player quits. There are still quite a few things.

1. I haven't updated the save file version yet nor do I check to see what version the file is. Any ideas on how I check an old version that has no version?
2. How do I use the user set autosave time rather than the 5secs I currently have hard coded (autosave_period)?
3. How do I label the save slot 0 as Autosave before the first autosave is done?
4. Should I move the autosave call into doFrame()?
5. Why can't I use _system->getMillis()?

digitall:
To answer 5., I would take an educated guess that the method / function you are calling this from is not a method of the Engine Object (or direct subclass)... hence it does not inherit access to the class variable _system for OSystem access. See:
https://github.com/scummvm/scummvm/blob/master/engines/engine.h#L58

You can either use the hacky global pointer, g_system->getMillis or you would need to call this via the pointer back to the engine object. This is quite common for debug consoles, which need a pointer to the engine object passed to them which by convention is _vm ... thus you would need something like _vm->_system->getMillis() . If you do a search on the rest of the engines, you should see examples of this...

digitall:
I think 2. has a pretty simple answer as well... Just delete your implementation of shouldPerformAutoSave() and the default implementation in engine.cpp will do that for you. See:
https://github.com/scummvm/scummvm/blob/master/engines/engine.cpp#L460

digitall:
I think 3. is pretty simple as well... You write a savegame with description "Autosave" into the slot if empty. Not sure if we have a good "standard" solution for this, but the Tucker engine might be worth a look to see how that deals with this:
https://github.com/scummvm/scummvm/blob/master/engines/tucker/saveload.cpp#L287

engines/mohawk/myst_state.h
```diff
@@ -96,6 +96,11 @@ enum DniEnding {
 						   // and destroying the books of his sons
 };
 
+enum SavegameVersions {
```
bgK: Save version handling is already implemented there: https://github.com/dafioram/scummvm/blob/8bc6e3d3b69a874d88baa5e8dcd83fb2d9e82653/engines/mohawk/myst_state.cpp#L49

bgK:
> How do I label the save slot 0 as Autosave before the first autosave is done?

Wouldn't it be misleading to the user if the slot read Autosave before anything was actually saved?

> Should I move the autosave call into doFrame()?

Yes, please.

dafioram:
I am now using the default shouldPerformAutoSave() so that just grabs the config based period (issue 2), and I have also moved the polling into the doFrame function (issue 4). I removed the redundant enum for the version.

@digitall should I prefer using getMillis() over getTotalPlayTime() the latter is what I am currently using and it seems to work.

After we have discussed the autosave operation a bit I am not sure how useful/used the autosave will be. With this feature the user will only have enabled autosaving if
1.) They don't already have a save in slot 0
2.) They don't save to slot 0 before the first config set autosave period. The default period is 300 secs for this value, which most users will probably not change. 

If slot 0 is not marked off as Autosave then they can save in this slot and hence disable autosave. I was hoping to block off slot 0 to future user saves so as to prevent the case of the user saving in slot 0 and then playing for a while and then forgetting to save and exiting. If the slot is reserved then the game will continue to autosave at what ever interval is in the config.

bonki:
The user cannot save in this slot if it is write protected (`setWriteProtectedFlag()`) so the issue doesn't exist. It is a minor UX problem in the sense that it might not be obvious to the user why they can't save to slot 0 if there is no autosave yet, but that is an issue which almost all engines which implement autosave currently suffer from.
If I remember correctly some engines additionally save on e.g. room changes which results in an autosave being created as soon as you start a game so e.g. creating an autosave when you enter an Age in Myst would easily fix this.

bonki:
> should I prefer using getMillis() over getTotalPlayTime() the latter is what I am currently using and it seems to work.

That's a good question. Currently all engines use `getMillis()` whereas `getTotalPlayTime()` essentially does the same thing but respects [improved pause support](http://wiki.scummvm.org/index.php/Advanced_Engine_Features#Improved_pause_support), at the minimum this pauses the timer when you open GUI dialogs such as the GMM (which is why `getTotalPlayTime()` is used to record total (net) play time). For consistency's sake I would go for `getMillis()` for now but maybe we should change this for all engines later, unless this has side effects which I'm not thinking about/aware of right now.

engines/mohawk/myst_state.cpp
```diff
@@ -44,9 +44,10 @@ MystSaveMetadata::MystSaveMetadata() {
 }
 
 bool MystSaveMetadata::sync(Common::Serializer &s) {
-	static const Common::Serializer::Version kCurrentVersion = 1;
+	static const Common::Serializer::Version kMinimumVersion = 1;
+	static const Common::Serializer::Version kCurrentVersion = 2;
 
-	if (!s.syncVersion(kCurrentVersion)) {
+	if (!s.syncVersion(kMinimumVersion) || !s.syncVersion(kCurrentVersion)) {
```
bonki: I haven't used the serializer before but this looks wrong.

engines/mohawk/detection.cpp
```diff
@@ -290,7 +300,10 @@ void MohawkMetaEngine::removeSaveState(const char *target, int slot) const {
 SaveStateDescriptor MohawkMetaEngine::querySaveMetaInfos(const char *target, int slot) const {
 #ifdef ENABLE_MYST
 	if (strstr(target, "myst")) {
```
bonki: Is there a specific reason why you limit this to Myst?

dafioram:
Good points. 

I fixed the logic for syncVersion which was definitely wrong. I am now using getMillis(), and slot 0 is not writable (it was that way before).

I think the last things are save files as the new save version with a version number flag and autosave flag as well as detect old versions. Autosave will be disabled if the user has a save in slot 0. I will also try to see if I can get an autosave to happen when a user starts a game so that a save in slot 0 will happen immediately.

@bonki I've only done myst so far since one at a time is easier. I'm not opposed to doing riven next.

dafioram:
Adding that dialogue is a little bit much for now, but the slot is blocked off so that the user cannot save to it anyway.

@bonki I like the idea of doing a an autosave immediately when the game starts to get an autosave slot in slot 0, but one of the risks of doing that is that if a user starts a new game (maybe by accident) it will force that new game to be the current autosave, potentially erasing all their progress in the previous autosave.

I've seen other games solve that by having autosave(s) separate for each new game or just by warning the user that by starting a new game it will erase (override) the previous autosave. I think the latter would be simpler, but for now I will only autosave based on the config specified interval.

dafioram:
I have added a new commit. It is not great. I am currently running into some issues.

I cannot load an old save because the serializer is trying to read in the new data (Metadata) fields that I have added, autoSave and version. I probably don't need version since syncVersion() does that? I am not sure how the version is getting saved to the savefiles.

In order to enable autosaving to slot 0 I must know that the save is an autosave otherwise I have to assume it is not. Obviously if there is nothing in slot 0 then I can autosave there.

My other issues can be fixed after these are solved.

engines/mohawk/myst_state.cpp
```diff
@@ -41,12 +41,15 @@ MystSaveMetadata::MystSaveMetadata() {
 	saveHour = 0;
 	saveMinute = 0;
 	totalPlayTime = 0;
+	version = 2;
+	autoSave = 0;
 }
 
 bool MystSaveMetadata::sync(Common::Serializer &s) {
```
bgK: This should be something like that:

```java
	if (!s.syncVersion(kCurrentVersion)) {
 		return false;
 	}
 
 	s.syncAsByte(saveDay);
 	s.syncAsByte(saveMonth);
 	s.syncAsUint16LE(saveYear);
 	s.syncAsByte(saveHour);
 	s.syncAsByte(saveMinute);
 	s.syncString(saveDescription);
 	s.syncAsUint32LE(totalPlayTime);
 	s.syncAsByte(autoSave, 2);
```

The serializer methods take the version number since when to read the data as an optional parameter. With this code, if `autoSave` is initialized with the proper default value, it will only be read for saves with version 2 or more.

No need to add a version field. The serializer takes care of it internally.

engines/mohawk/myst_state.cpp
```diff
@@ -227,6 +234,8 @@ void MystGameState::updateMetadateForSaving(const Common::String &desc) {
 	_metadata.saveMinute = t.tm_min;
 	_metadata.saveDescription = desc;
 	_metadata.totalPlayTime = _vm->getTotalPlayTime();
+	_metadata.version = 2;
+	_metadata.autoSave = (uint8) Mohawk::kAutoSaveAllowed && (slot == Mohawk::kAutoSaveSlot);
```
bgK: You shouldn't use a global variable for knowing if this an autosave. It adds tight coupling to things that don't need to know much about each other.. There could be parameter `bool isAutosave` in this method.

bonki:
> @bonki I've only done myst so far since one at a time is easier. I'm not opposed to doing riven next.

Makes sense, I mistakenly thought that savegame handling was game-agnostic which apparently it isn't.

bgK:
> Makes sense, I mistakenly thought that savegame handling was game-agnostic which apparently it isn't.

Mohawk was more a platform on top of which game engines can be built (not unlike ScummVM or SDL). This is why little code is shared between the games.

bonki:
Thanks for the clarification.

engines/mohawk/myst_state.cpp
bgK: I don't think it's correct to decide whether autosave is allowed in querySaveMetaInfos. The game could be started without it being called. It's probably a lot easier and safer to have a method isAutoSaveAllowed that is called when trying to autosave that opens the save in slot 0 if there is one and checks if it's an autosave.

dafioram:
Alright, autosaving works if no autosave is in slot 0, but if there is a save in it (autosave or not) then autosave doesn't happen, not what I wanted.

engines/mohawk/myst_state.cpp
```diff
@@ -251,12 +257,42 @@ bool MystGameState::saveMetadata(int slot) {
...
+
+	// Read the metadata file
+	Mohawk::MystSaveMetadata metadata;
+	if (!metadata.sync(m)) { // the save in the autosave slot is corrupted, enable autosave
+		_isAutosave=true;
+		delete metadataFile;
+		return;
+	}
+
+	if (metadata.autoSave == (uint8)1) {
```
dafioram: metadata.autoSave never is true. I expect this to be true when I have saved an autosave. Doesn't matter if I hardcode _metadata.autoSave = 1 (even for all slots) in MystGameState::updateMetadateForSaving. So maybe MystSaveMetadata::sync isn't loading/syncing the autoSave value because I am not saving saves with the new version (2)?

dafioram:
metadata.autoSave line 280 in MystGameState::isAutoSaveAllowed is never is true. 

I expect this to be true when I have saved an autosave. Doesn't matter if I hardcode _metadata.autoSave = 1 (even for all slots) in MystGameState::updateMetadateForSaving. 

So maybe MystSaveMetadata::sync isn't loading/syncing the autoSave value because I am not saving saves with the new version (2)?

engines/mohawk/myst_state.cpp
bgK: It seems to be working for me.

dafioram: Does it work if you exit out of scummvm and restart it (with an autosave in slot 0)?

bgK: I think so. But as I said it would be simpler and safer to just check if the save in slot 0 is an autosave just before trying to autosave than doing it on engine startup and keeping the value around.

dafioram:
Whenever an autosave is requested it now checks slot 0 and enables autosave if slot 0 does not have an a save file or it is an autosave. 

I believe the code now works the way we want it and just needs some cleanup. Thoughts?

bgK:
> I believe the code now works the way we want it and just needs some cleanup. Thoughts?

Agreed.

engines/mohawk/myst.cpp
```diff
@@ -1161,6 +1167,11 @@ Common::Error MohawkEngine_Myst::saveGameState(int slot, const Common::String &d
 	return _gameState->save(slot, desc) ? Common::kNoError : Common::kUnknownError;
 }
 
+Common::Error MohawkEngine_Myst::autoSave() {
+	Common::Error saveError = saveGameState(Mohawk::kAutoSaveSlot, Mohawk::kAutoSaveName);
```
bgK: Please print a warning if autosave failed.

engines/mohawk/myst_state.cpp
```diff
@@ -227,6 +229,7 @@ void MystGameState::updateMetadateForSaving(const Common::String &desc) {
 	_metadata.saveMinute = t.tm_min;
 	_metadata.saveDescription = desc;
 	_metadata.totalPlayTime = _vm->getTotalPlayTime();
+	_metadata.autoSave = isAutoSaveAllowed() && (slot == Mohawk::kAutoSaveSlot);
```
bgK: I don't think it's necessary to call `isAutoSaveAllowed` here. The argument `int slot` could be changed to `bool autosave` and passed by the callers up to the autosave method (Calling  `_gameState->save` instead of `saveGameState` in `autosave`).

engines/mohawk/myst_state.cpp
```diff
@@ -251,12 +254,41 @@ bool MystGameState::saveMetadata(int slot) {
 	return true;
 }
 
+bool MystGameState::isAutoSaveAllowed(void) {
```
bgK: Please remove `void` here (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl25-dont-use-void-as-an-argument-type)

dafioram:
See if this looks good to you.

Autosaving (if possible) on exit has been added (back in).

bgK:
Thanks! I merged it with some minor changes (style and autosaving when quitting by closing the window / return to launcher)

PR #1175, opened 2018-05-02
Common: Template Math for angle conversion

dafioram:
Conversion between degrees and radians, used a lot through out the code. This turns the current functions into templates and allows input and output type to be different. 

These functions are only used by sci and wintermute currently, but a lot of engines have their own angle converts that can just use this.


criezy:
The change looks OK.
Have you identified other places already where you would like to use those template functions?

Also the first commit message does not quite match our conventions (`[COMMON]` vs `COMMON:`)

dafioram:
I updated the template usage and commit message. 

I am now reusing the 2 template methods in the 1 temple methods. I have decided to up convert the input (whatever it is) to a double and then do that math with M_PI and 180.0 and convert to whatever the output type is. 

Should I consider using the rounding function? If the output type is say an int then it will use the integer truncation, currently.

I am currently using a c style cast should I do that differently?

I know at least titanic and fullpipe could use it. Fullpipe has a #define ANGLE(X) function that would get replaced. How about I update the other engines in another PR?

criezy:
> Should I consider using the rounding function? If the output type is say an int then it will use the integer truncation, currently.

The SCI engine is currently using truncation, and I am not sure it would be wise to change it. It might break some subtle things in SCI. So I would leave it like that.

> I am currently using a c style cast should I do that differently?

That seems fine to me.

> How about I update the other engines in another PR?

That sounds fine as well.

I see no issue with this pull request now, so I will merge it tomorrow evening, unless somebody objects to it.

bonki:
> > Should I consider using the rounding function? If the output type is say an int then it will use the integer truncation, currently.

> The SCI engine is currently using truncation, and I am not sure it would be wise to change it. It might break some subtle things in SCI. So I would leave it like that.

Can't the same thing be said about always calculating with double-precision now internally? If it doesn't really make a difference then why always use double now, if it does then can we really be sure it doesn't break anything (I guess we can)?

Double-precision FP is also potentially much more expensive because it has to be emulated in software when the CPU doesn't support it in hardware (do we target such platforms)?

I guess that neither of the above is an actual issue in practice.

(Note that even without the cast the computation would be performed with double-precision because floating point literals without suffix are of type double.)

dafioram:
Yes, my change to using doubles internally is different from how sci did it (it went float to int so it upconverted everything to floats first). Previously, I was converting everything to what the input was, which would preserve the sci behavior, but would do some silly things in other situations. Consider a deg2rad call with an int input (degrees as an int is not unreasonable) and a float output (for the radian value), in that case 180 would be converted to an int and M_PI to an int and that would be very imprecise.

As a comprise I could convert everything to float internally (instead of double) and only for cases when a double is used as an input would that be slightly off (for a double to double it would go double ->float -> double). If that was a concern I could handle that specific case with a template specialization, but there might be a few (double to int, double to uint8).

bonki:
> Consider a deg2rad call with an int input (degrees as an int is not unreasonable) and a float output (for the radian value), in that case 180 would be converted to an int and M_PI to an int and that would be very imprecise.

That's not the case because of what I mentioned earlier. Both `180` and `M_PI` are double floating point literals (due to the lack of a suffix) which means that `deg` will always be implicitly promoted to `double`.

dafioram:
That comment you highlighted is for when I am considering the case when I have everything on the input converted to floats, see the previous sentence (Previously, I...). I am giving a counter example to why converting M_PI and 180.0 to be of the input type and why that can go wrong for an int to float call.

bonki:
Ah right, sorry.

dafioram:
Alright the default 2 parameter case in done by upconverting to floats and when the input is the double type it is done differently. 

I believe the 1 parameter <double> usage will find my double method instead of the other 1 parameter method that calls the two parameter as <double,double>. If the user does <double,double> it will probably call the float variant so hopefully the user knows to use <double> instead.

dafioram:
Since this is an inline function I've decided to calculate the scalar factors (pi/180 and 180/pi) upfront at double precision and just hard code them in the code.

bonki:
`scumm/he/logic/puttrace.cpp` and `titanic/star_control/fvector.cpp` could also benefit from this.

dafioram:
@bonki For this PR I just update the engines that were currently using rad2deg and deg2rad. Another PR will be done that updates engines that would benefit from it.

PR #1177, opened 2018-05-03
XEEN: Add an optional param to CD playback to specify sound type

dreammaster:
The current CD manager play method always internally plays the given track/file using the kMusicSoundType. This is a problem for the World of Xeen 2 CD  Talkie, since for it, the audio tracks are all exclusively voice samples. This would be an exacerbated problem for it, since the game keeps playing background music when the CD voice samples are played. And we don't want the voice playback to be reduced in volume when players try to reduce the music volume so they can hear the speech more clearly.

### DIFF
--- a/README
+++ b/README
@@ -929,17 +929,18 @@ is simpler and more straightforward than moving around using the menu.
 --------------------------------------
 To properly plan the World of Xeen CD Talkie using original discs,
 use LAME or some other MP3 encoder to rip the cd audio tracks to files.
-Name the files track1.mp3 track2.mp3 etc. ScummVM must be compiled with
+Name the files track2.mp3 track3.mp3 etc. ScummVM must be compiled with
 MAD support to use this option. You will need to rip the file from the
 CD as a WAV file, then encode the MP3 files in constant bit rate. This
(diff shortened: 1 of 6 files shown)


README
```diff
@@ -929,17 +929,18 @@ is simpler and more straightforward than moving around using the menu.
 --------------------------------------
 To properly plan the World of Xeen CD Talkie using original discs,
 use LAME or some other MP3 encoder to rip the cd audio tracks to files.
-Name the files track1.mp3 track2.mp3 etc. ScummVM must be compiled with
```
bonki: I'm just curious, why this change?

dreammaster: I thought it would avoid confusion, since my understanding is that on original discs, the first track is the data one, so the extracted tracks would start with track 2. So I wanted to avoid confusing the users into thinking they'd have to rename the 2nd track to track 1, 3rd to track 2, etc.

dreammaster: The GOG track list also starts with xeen02.ogg, so keeping extracted tracks as default has both methods matching.

bonki: Gotcha, thanks!

PR #1179, opened 2018-05-03
SCUMM: Improve autosave handling for v5+ games

bonki:
v5+ scripts can request saving/loading of savegames, this type of
savegame is internally called a "temporary" savegame
(_saveStateTemporary == true) which is invisible to the user (I'm not
sure whether this is by design or not).

Currently the savegame handling in scummLoop_handleSaveLoad() doesn't
distinguish between such temporary savegames and normal autosaves and
unconditionally resets _lastSaveTime (even after loading).
This has the unwanted side effect of potentially delaying the creation
of normal autosaves which are supposed to be created in accordance with
the autosave period setting in the GUI.

This commit makes sure that _lastSaveTime only gets updated if and only
if saving a (non-temporary) autosave.

### DIFF
--- a/engines/scumm/scumm.cpp
+++ b/engines/scumm/scumm.cpp
@@ -2435,6 +2435,9 @@ void ScummEngine::scummLoop_handleSaveLoad() {
 
 			if (success && _saveTemporaryState && VAR_GAME_LOADED != 0xFF && _game.version <= 7)
 				VAR(VAR_GAME_LOADED) = 201;
+
+			if (!_saveTemporaryState)
+				_lastSaveTime = _system->getMillis();
 		} else {
 			success = loadState(_saveLoadSlot, _saveTemporaryState, filename);
 			if (!success)
@@ -2458,7 +2461,6 @@ void ScummEngine::scummLoop_handleSaveLoad() {
 			clearClickedStatus();
 
 		_saveLoadFlag = 0;
-		_lastSaveTime = _system->getMillis();
 	}
 }
 


engines/scumm/scumm.cpp
```diff
@@ -2435,6 +2435,9 @@ void ScummEngine::scummLoop_handleSaveLoad() {
 
 			if (success && _saveTemporaryState && VAR_GAME_LOADED != 0xFF && _game.version <= 7)
 				VAR(VAR_GAME_LOADED) = 201;
+
+			if (!_saveTemporaryState)
+				_lastSaveTime = _system->getMillis();
```
bgK: Previously, `_lastSaveTime` was also reset when loading a save. Now not anymore. Is that change intentional?

bonki: This is indeed intentional (apparently my commit message isn't as unambiguous as I had hoped), but I'm not sure either if that is "correct"/OK (hence the PR) because I don't know how and when scripts actually use these opcodes.

Only v8 seems to have an opcode for loading, older versions down to v5 can can only save which is puzzing to me because that suggests that the user should be able to access these savegames, why else write them in the first place if scripts can't load them again? But as I mentioned this is currently not possible because the launcher/GMM doesn't list savegames with the temporary file extension.

My rationale was that since these savegames are not accessible by the user they should never interfere with our autosaves, not even on loading, which is why I didn't go with the more defensive approach (keep the instruction where it was and also reset on loading).

¯\\_(ツ)_/¯

bgK: Oh, right. Your change looks good to me, though I'm not SCUMM engine expert.

bonki: Thanks for looking at it!

bonki:
As I don't see how this could potentially break anything while it should improve user experience I'll go ahead and merge this as-is now. Thanks @bgK.

PR #1183, opened 2018-05-06
SAGA: Deduplicate kScriptTimeTicksPerSecond constant

bonki:
The underlying type of an enum cannot be a float so the following doesn't really make sense:

    enum ScriptTimings {
        kScriptTimeTicksPerSecond = (728L/10L),
        kScriptTimeTicksPerSecondIHNM = 72,
        kRepeatSpeedTicks = (728L/10L)/3,
        [...]
    };

Because of that, `kScriptTimeTicksPerSecond ` gets truncated to `72` and is therefore identical to `kScriptTimeTicksPerSecondIHNM`.

This commit removes `kScriptTimeTicksPerSecondIHNM`, but if the engine actually requires a higher precision then this is not correct and the constant(s) need(s) to be made floating point.

If it's OK to truncate I suggest adding a comment to the enum.

### DIFF
--- a/engines/saga/saga.h
+++ b/engines/saga/saga.h
@@ -311,7 +311,6 @@ enum GameObjectTypes {
 
 enum ScriptTimings {
 	kScriptTimeTicksPerSecond = (728L/10L),
-	kScriptTimeTicksPerSecondIHNM = 72,
 	kRepeatSpeedTicks = (728L/10L)/3,
 	kNormalFadeDuration = 320, // 64 steps, 5 msec each
 	kQuickFadeDuration = 64,  // 64 steps, 1 msec each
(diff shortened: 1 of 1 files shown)


dafioram:
It was changed to an enum in commit ea3b0d11003. Before that commit (ad0b8f1de57e0c21) it was 72.8 ticks per second, at least thats what the comment said. 

At that time the timing stuff was handled in Script::SF_sleep in source/saga/sfuncs.cpp

dafioram:
I played Inherit the earth a bit with those enums set as floats as well as TicksToMSec to take in a float and operate internally with floats. It played the same as before the changes.

sev-:
Ok, merging.

PR #1185, opened 2018-05-07
Prefer pkg-config over freetype-config if possible

Polynomial-C:
As of freetype-2.9.1 the freetype-config script no longer gets installed
by default.

See also [Make installation of `freetype-config' optional](http://git.savannah.gnu.org/cgit/freetype/freetype2.git/commit/?id=a7833f26c4ac45cafe1dffdcd7f7dcfd6493161c) commit by freetype upstream.

### DIFF
--- a/configure
+++ b/configure
@@ -204,6 +204,7 @@ _sparklepath=
 _sdlconfig=sdl2-config
 _libcurlconfig=curl-config
 _freetypeconfig=freetype-config
+_freetype_found="false"
 _sdlpath="$PATH"
 _freetypepath="$PATH"
 _libcurlpath="$PATH"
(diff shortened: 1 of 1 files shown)


Polynomial-C:
Calling pkg-config directly might break cross-compilation but I didn't bother to work on this as the configure script already calls pkg-config directly somewhere else.

bgK:
At one point, in residualvm we used pkg-config to detect GLEW. Users were confused because they installed GLEW using macports / homebrew and it was still not found. What we didn't tell them is that they also needed to install pkg-config. In my opinion, there should be a check / warning / something to help users know they are missing pkg-config.

bgK:
> Calling pkg-config directly might break cross-compilation but I didn't bother to work on this as the configure script already calls pkg-config directly somewhere else.

Won't that be a problem on buildbot as soon as this is merged? I don't have access to the server to check, but `configure` will likely use the host `pkg-config`, find the host version of freetype, and use incorrect flags to build the cross compiled targets.

eli-schwartz:
pkg-config supports PKG_CONFIG_PATH for prioritizing the directory to read freetype.pc from. People who are cross-compiling software and wish to use dependencies that are not from the default, native system path should be setting this variable.

It also transparently supports using custom versions of things, e.g. for openssl 1.1 Arch Linux ships an openssl-1.0 compat package for old software, which means by setting (or prepending to) PKG_CONFIG_PATH, it becomes trivial to swap in a different library for basically anything, in **one** place, as long as you support pkg-config. No need for hacks like mucking around with $PATH and probably overriding lots more than just the *-config scripts.

There is a **reason** pkg-config is the standard. Because it **works**. It lets you define where you want to get your XXX from, it orders flags for you, it lets you transparently add recursive library dependencies for both `--static` and non-static builds as defined by the very simple, clear .pc grammar, etc. etc. etc.

Currently this software definitively breaks on Arch Linux, and users are reporting bugs to us. But pkg-config support has been around for a long time, it's the ideal way to detect dependencies, and we don't plan on re-enabling the legacy shellscript. So downstream projects need to (finally!) migrate to pkg-config. 

Other distros will almost certainly do the same thing, sooner or later.

Note that linux distros will usually have pkg-config installed by default:
- as part of a "building packages metapackage", e.g. on Arch Linux it is part of base-devel
- on Gentoo which is a source-based distro it should be part of the system profile
- on Debian and Fedora it seems to be a dependency of the various -devel packages containing headers, if the devel package has .pc files

For the rest, this is what wiki instructions are for. The current macOS instructions even explain the need for manually installing pkg-config *if using fluidsynth*.

bgK:
I tried this with the PlayStation 3 toolchain I have installed. I used to configure the build using `../configure --host=ps3`. The script would find the PS3 libraries using the environment variables defined by the toolchain: https://github.com/ps3dev/ps3toolchain/blob/71e9c0222ca4f0d3041a45b6821c05f390b27fa3/readme.txt#L29.
With this change I need to do `PKG_CONFIG_PATH=/usr/local/ps3dev/portlibs/ppu/lib/pkgconfig/ ../configure --disable-all-engines --host=ps3`.

What do you think is the proper approach?
* Change `configure` to append the toolchain pkg-config path to the user defined `PKG_CONFIG_PATH` environment variable before passing it to `pkg-config`.
* Tell the porters to properly define `PKG_CONFIG_PATH` for their build environment and alter the buildbot configuration for all the toolchains.

eli-schwartz:
If your configure script already sets up the toolchain for them, it makes sense to set `PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/path/to/libdir/pkgconfig"` too.

Polynomial-C:
So is there anything I can do to speed up merging of this PR?

rsn8887:
Unless I misunderstand something, I am pretty sure this will break freetype support on Vita on the current buildbot config. I think this should be thoroughly tested on all relevant platforms before merging.

Polynomial-C:
I have tried to make this not break on systems that don't have pkg-config installed. Just look at the `find_freetype` function in my patch. The only issue I can see is systems that cross-compile scummvm and thus need pkg-config from the target system rather than the host system. But I am quite sure this is already broken as the configure script already calls pkg-config to check for `libunity`.

eli-schwartz:
> The only issue I can see is systems that cross-compile scummvm and thus need pkg-config from the target system rather than the host system. But I am quite sure this is already broken as the configure script already calls pkg-config to check for libunity.

Properly supporting this is in fact the very reason Debian convinced Freetype to drop the freetype-config script. :D

The custom is to try running $PKG_CONFIG, defaulting to `pkg-config`, but on multi-arch/cross-compiler builds the builder should be setting `PKG_CONFIG=${CHOST}-pkg-config` which will be a wrapper script, see e.g. https://autotools.io/pkgconfig/cross-compiling.html

The pkg-config binary itself can come from anywhere as the only host vs. target dependent information is some default strings it uses for the sysroot/libdir, and it's all configurable via the environment variables documented in its manpage.

sev-:
Closing this for the sake of #1267

PR #1187, opened 2018-05-10
ENGINES: Return unknown game variants with the list of detected games

bgK:
This refactoring work decouples unknown game variants reporting from the advanced detector.
* Unknown game variants are now part of the results returned by the detectors. This allows the caller to have different handling. e.g. the launcher displays the unknown game dialog, the command line displays a text report.
* The unknown game report now aggregates unknown game matches from all the engines. This ensures the unknown game dialog is shown at most once when trying to add a game.
* The non file-based AD fallback detectors can now add a list of unknown files to the detected game entry. Those unknown files will be displayed in the unknown files dialog. I've implemented it for the sludge engine.

This also enables implementing the following features (none are actually implemented as this pull request is already very large):
* Showing the unknown game dialog when using the mass add button.
* Ignoring unknown game detection results from one engine when there is an exact match from another (removing false positives when two games from different engines use files with the same name for detection)
* Allowing the user to cancel adding a game when on the unknown game dialog.

I changed a lot more code than I would have hoped to implement this. That includes changing the detection results from being map-based to regular structs. But in the end it seems to me the code is easier to understand that way.

### DIFF
--- a/backends/platform/dc/selector.cpp
+++ b/backends/platform/dc/selector.cpp
@@ -271,21 +271,21 @@ static int findGames(Game *games, int max, bool use_ini)
     }
 
     if (!use_ini) {
-      GameList candidates = EngineMan.detectGames(files);
+      DetectedGames candidates = EngineMan.detectGames(files);
 
-      for (GameList::const_iterator ge = candidates.begin();
+      for (DetectedGames::const_iterator ge = candidates.begin();
 	   ge != candidates.end(); ++ge)
 	if (curr_game < max) {
-	  strcpy(games[curr_game].filename_base, ge->gameid().c_str());
+	  strcpy(games[curr_game].filename_base, ge->gameId.c_str());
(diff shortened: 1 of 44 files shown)


criezy:
I have had a quick look at all the commits here and they all seem to make sense. And what you are trying to achieve here seems like a good idea to me. It will already nicely improve the behaviour of detecting games, and will allow implementing some nice additional feature in the future. 

Last week I did trigger the case of getting an unknown game report from an engine while the game was detected by another engine (because cge and cge2 engines have data files with the same name, so trying to detect Sfinx causes an unknown Soltys version report). So fixing this will be a nice follow up to this.

That will also allow us to move the UnknwonGameDialog code to gui instead of engine and fix circular dependency (and thus add back the use of a ScrollContainerWidget).

bgK:
Thanks @criezy for looking at the code.

I'm merging this now. This is a rather large change. If there are regressions, send them my way.

PR #1189, opened 2018-05-12
Add stretch modes

criezy:
This PR is a proof of concept to discuss and experiment with the possibility to scale the display to the window in various ways for backends that supports scaling (SDL 2 Surface or SDL 1.2 and 2 OpenGL).

This is different from the scaling of the graphics mode for the SDL Surface backend. This graphics mode causes the display to be scaled to a surface with a scaler of 1x, 2x or 3x. But then this surface is further scaled to fit the window  (or screen in fullscreen mode). The OpenGL backend similarly scales the texture to fit the window or screen.

In this PR I have implemented four different modes:

No scaling: no additional scaling is performed after the graphics mode scaling was applied
scummvm_scale_none

Integral scaling: scale by an integral amount as much as possible but not bigger than the window.
scummvm_scale_integer

Fit to window: scale to fit the window while respecting the aspect ratio. This is the current behaviour, and still the default behaviour after my changes.
scummvm_scale_fit

Fit and stretch to window: in this mode the aspecy ratio is not respected and there is no black bars.
scummvm_scale_fit_stretch

There is also a second commit that adds the Ctrl+Alt+s hotkey to cycle through the scale modes.

At this point there are two main pending questions.

1. Is there interest in this idea? Or is this a bad idea?
There seems to be since the question was asked on the forum a few time (see for example [here](http://forums.scummvm.org/viewtopic.php?t=14626) or [here](http://forums.scummvm.org/viewtopic.php?t=14604)). In my case I was mostly interested in the integral scaling, but while implementing this I decided to push the idea a bit further.

2. Assuming this is a good idea, should this be exposed in the GUI?
I think it probably should. The hotkeys are sufficient to test the idea, but most people will not be aware of it unless it appears in the Options dialog. On the other hand if we think few people will be interested in these options, then there is no point adding this to the Options dialog.
This should probably be added to the command line in any case though.

3. Do you see a potential issues with those changes? Or ways to make them better?

If there is interest I will spend some time to cleanup the code and finalise this pull request.


Kroc:
This may be confusing when there exists the 1x/2x/3x scaling beforehand -- perhaps this feature would be better described as "fit to screen", rather than "scaling"

tsoliman:
I really like this as a feature from the backend point of view, especially the integral scaling option. It really helps with odd-resolution monitors that aren't HiDPI (HiDPI monitors don't seem to suffer from the non-integral scaling artifacts as much)

I am not sure how to make it less confusing for the user if it is exposed as a GUI option. I like Kroc's "fit to screen" or maybe something like "further scaling"?
Other ideas: "scaling" vs "fit"

The current "Graphics Mode" seems to be a catch-all for two existing concepts: the existing Surface/OpenGL/etc in addition to the scaling. (OpenGL's submodes are handled by a "secret" keystroke). Maybe that itself should be broken down into "Graphics Mode" (surface/openGL/etc) and "Scaling/Filtering" and then the 3rd "Screen Fit" can be a 3rd dropdown?

Maybe this is outside the scope of the PR. Just my 2 cents.

lotharsm:
I really like the idea of this PR, especially the integer scaling for a pixel-perfect, crisp image. Not sure if we should expose this to the GUI since it might lead to confusion... *If* we decide to expose it to the GUI, I vote for a dropdown menu which allows selecting the different stretching modes.

sev-:
I like this too, and I think we have to expose it in GUI. Maybe not just call it stretching (vs scaling), but maybe rename scaling to 'discrete scaling' or 'integer scaling'.

lotharsm:
What's the current state of this PR?

criezy:
This is a timely question ;-)
My plan is to work on this this week-end since the weather forecast is that it will get even warmer and the room with my computer is in the coolest one (oriented to the north) and the smallest one (there is not much else to do that use the computer).

Somewhat related to the comment from @tsoliman I don't really like the way we currently handle graphics modes in the GUI, and [the ticket #7719](https://bugs.scummvm.org/ticket/7719) in our bug tracker is related to this. But I consider this is outside the scope of this PR.

criezy:
I have now cleaned the code, added the option to the Options dialog, and added a command line argument for it.

I think this PR is now in a state where it can be reviewed for a possible merge.

criezy:
Here is what the options currently look like. Any suggestion on a better wording is welcome.

scummvm_stretch_options

Kroc:
Wording suggestions:

s/Center/Center-in-window
s/Integral scaling/Stretch, but keep pixels whole
s/Fit to Window/Stretch, but keep aspect ratio
s/Stretch to Window/Stretch to fill whole window

lotharsm:
I tested this new feature and so far, it looks and works great for me!

The only suggestion I have is to reword the Integral scaling feature to something more descriptive like "Pixel-perfect scaling", but this is just my personal opinion.

Great job, looking forward to see this in master!

bonki:
I have tested this as well and it works as intended with minor annoyances when switching from fullscreen back to windowed mode but I assume these to not be related to this PR (I haven't thoroughly checked yet and I don't usually toggle between modes). I'll compare against a current vanilla build.

Other than that I agree with Lothar. My vote also goes to keeping *Center*, *Fit to window* and *Stretch to window* and only give *Integral scaling* another thought, even though it makes total sense once you realize what it does. I have no strong opinion against it, though.

As an aside, why is *Window* capitalized? I also notice that *Device* in *Audio -> Preferred Device* and *Speech* in *Audio -> Text and Speech* are capitalized while others are not. To me this looks wrong, is there some thought or system behind this?

As another aside, the cykling hotkey doesnt work on my Windows 10 test box because it seems to be system default and isn't captured by ScummVM.

lotharsm:
Since cycling through the different modes using the CTRL+ALT+S hotkey works for me on Windows 10, this might need some more investigation though.

criezy:
Some information on my choices:
- Ctrl+Alt+s seemed to be a logical choice for a shortcut since Ctrl+Alt+a is used for aspect ratio correctiopn, Ctrl+Alt+f is use to toggle filtering and Ctrl+Alt+plus/minus is used to increase or decrease the scale factor.
 - The `Center`, `Fit` and `Stretch` names come from what I have in Windows 7 for the background image options. It also has a `Fill` that cuts part of the image and I decided not to implement in ScummVM (missing part of the display doesn't seems like a god idea to me for a game). Using the same nomenclature as Windows seemed a good idea to me as I would hope Microsoft put some thoughs in those, and these would also be somewhat familiar to the user.
 - Microsoft does not specify `Window` though (so only `Fit` and not `Fit to Window` for example). To me this looked a bit confusing, which is why I have added `window`, but maybe that actually makes it more confusing since the option also impact fullscreen display, where this would be better named as `Fit to screen`.
 - And the main question mark for me is indeed on the remaining option (`Integral scaling`). `Pixel-perfect scaling` could indeed be an option.

And I agree with @bonki that we might want to make some changes so that we are consistent in the capitalization (probably only capitalize the first word since this seems to be what we do in most places).

bonki:
I don't know why it didn't work yesterday but I can confirm that the hotkey does work, so forget I ever said anything.

Ad `Fit to screen`: I'm not sure if this is better (or worse), one could also misinterpret that as the window being fit to the whole screen if one wants. I guess both can be interpreted in the wrong way so I'm totally fine with either.

Ad `Integral vs. Pixel-perfect scaling`: My stance is that both make sense but neither is self-explanatory so a user would need to consult a possible tooltip or the manual anyway, and then it doesn't really matter. So personally, I'm fine with either.

Ad capitalization: In my book only the first word should be capitalized except where it makes sense not to. I suggest lowercasing `window` (or `screen`, for that matter) here and I can do a PR for the rest of the GUI.

criezy:
I have sent an email to -devel about capitalisation and I will wait for the discussion there to reach a conclusion before updating this PR.

PR #1190, opened 2018-05-14
VIDEO: fixed decoding byteRun when there is more than 255 packets on line

whiterandrek:
We shouldn't rely on chunks count in FLC files. Instead, we should rely on width of the frame.

From https://www.compuphase.com/flic.htm
> Each line of the image is compressed separately, starting from the top of the image. The first byte of each line is the packet count. It is a holdover from the FLI format and it should be ignored, because it is now possible to have more than 255 packets on a line (for FLC files). Instead, the width of the frame image now is the criterion for decoding: continue decompressing until the number of uncompressed pixels becomes equal to the image width.

### DIFF
--- a/video/flic_decoder.cpp
+++ b/video/flic_decoder.cpp
@@ -265,9 +265,9 @@ void FlicDecoder::FlicVideoTrack::copyFrame(uint8 *data) {
 
 void FlicDecoder::FlicVideoTrack::decodeByteRun(uint8 *data) {
 	byte *ptr = (byte *)_surface->getPixels();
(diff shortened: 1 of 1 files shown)


bonki:
Please fix your indentation (should be tabs, not spaces).

bonki:
Looks reasonable and it seems other decoders do exactly just that. Anybody against me merging this?

bonki:
The only officially supported engine right now using the FLIC decoder is Tucker and I have checked that this does not break decoding of any of the FLIC files used by the game (there are no lines with >255 packets per line so in the Tucker case it makes no difference).

I've landed this in commit ff1b7ff89d03dce2e10bda136c954bb20c384c36 with an updated commit message. Thanks!

PR #1192, opened 2018-05-17
SCUMM: Improve 'imuse play' debugger command error handling

bonki:
This commit introduces the following (seemingly non-invasive) changes
to make the 'imuse play' debugger command more useful (i.e. don't crash
when trying to load the wrong kind of (sound) resource.

* ScummEngine::readSoundResource()
  Instead of fatally error()'ing upon hitting a non-sound resource type,
  e.g. a room header (0x524d4844 aka RMHD), we now only issue a warning().
  This enables the already existing dead code which properly returns 0
  (aka no resource loaded).

* ResourceManager::validateResource()
  Instead of fatally error()'ing upon hitting an illegal glob type we now
  only issue a warning() and return false (also existing dead code).
  All methods calling validateResource() check its return value so this
  seems like the right thing to do anyway.

* ScummDebugger::Cmd_IMuse()
  Instead of directly calling ensureResourceLoaded() we now call
  getResourceAddress() instead (which in turn calls ensureResourceLoaded()
  and handles other edge cases) and only attempt to play a sound if the
  returned pointer actually is valid.

Fixes Trac#10527.

### DIFF
--- a/engines/scumm/debugger.cpp
+++ b/engines/scumm/debugger.cpp
@@ -145,8 +145,8 @@ bool ScummDebugger::Cmd_IMuse(int argc, const char **argv) {
 					debugPrintf("Selecting from %d songs...\n", _vm->_numSounds);
 					sound = _vm->_rnd.getRandomNumber(_vm->_numSounds);
 				}
-				_vm->ensureResourceLoaded(rtSound, sound);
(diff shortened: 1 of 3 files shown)


sev-:
Thanks, it all makes sense to me. Merging.

PR #1195, opened 2018-05-19
MOHAWK: RIVEN: Add Autosave Support

dafioram:
Similar to myst autosave.

The game will autosave to slot 0 using the save
period given in the scummvm config file.

Autosaves are only allowed when an autosave is
in slot 0 or there is no save in slot 0.

This will not override any saves the player
has previously put in save slot 0. If there
is a save in slot 0 that is not an autosave
then there will be no autosaving.

Unlike quiting in myst using F5 and returning to launcher/qutiing I don't see where that happens in rivens doFrame() maybe check hasGameEnded()? in the Common::KEYCODE_F5 switch case of doFrame?


dafioram:
Hows this look?

bgK:
Very good, thanks!

PR #1198, opened 2018-05-22
MOHAWK: README: Minor autosave changes and notes for autosaving

dafioram:
1. Made the autosave section of the README to be a list of engines and games, added myst, riven, and tucker to it.
2. Prevented myst from autosaving over an original game save since it doesn't have a metadata and the autosave enable check looks at metadata file to determine if autosave should be enabled.
3. Allow non-autosave in slot 0 to be delete-able via the gui in myst and riven. Fixes Trac#10536

### DIFF
--- a/engines/mohawk/myst_state.cpp
+++ b/engines/mohawk/myst_state.cpp
@@ -259,10 +259,17 @@ bool MystGameState::saveMetadata(int slot) {
 bool MystGameState::isAutoSaveAllowed() {
 	// Open autosave slot and see if it an autosave
 	// Autosaving will be enabled if it is an autosave or if there is no save in that slot
-	Common::String filename = buildMetadataFilename(kAutoSaveSlot);
-	Common::ScopedPtr<Common::InSaveFile> metadataFile(g_system->getSavefileManager()->openForLoading(filename));
-	if (!metadataFile) { // There is no save in the autosave slot, enable autosave
+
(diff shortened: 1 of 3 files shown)


bonki:
Ad 1) KYRA (only v1 games?), COMPOSER and ZVISION also seem to implement autosaves. Can you verify these and add the necessary games them to the list?

dafioram:
Kyra autosaves only for game 1. It looks like composer autosaves, but I don't have any composer games to check that. ZGI autosaves, so nemesis probably autosaves also. SCI looks like it autosaves, but I think that needs to be checked individually for each sci game and can be added in a later commit (there may be issues like Trac#10538) .

engines/mohawk/myst_state.cpp
```diff
@@ -259,10 +259,14 @@ bool MystGameState::saveMetadata(int slot) {
 bool MystGameState::isAutoSaveAllowed() {
 	// Open autosave slot and see if it an autosave
 	// Autosaving will be enabled if it is an autosave or if there is no save in that slot
+	if (!loadState(kAutoSaveSlot)) { // Cannot load non-meta file, enable autosave
```
bgK: But if loading succeeds, we will have overwritten the current state with the save in slot 0.

dafioram:
Here it is.

bgK:
Thanks for your contribution!

PR #1199, opened 2018-05-22
BASE: Default joystick_num to 0 instead of -1

dafioram:
Fixes Trac#10366.

A default value of 0 will allow scummvm to
automatically use a controller that is
plugged in.

Most users that use a controller will like
that this just works and they won't have
to setup the config file.

Users not using a controller will just
get a WARNING: Invalid joystick: 0! in the cmd window.

This default behavior can be overriden
by including joystick_num in the config
file.

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -270,7 +270,7 @@ void registerDefaults() {
 #endif
 
 	// Miscellaneous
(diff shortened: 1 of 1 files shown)


tsoliman:
Does it make more sense for this defaulting to be a compile-time option? That way platforms that expect joysticks would compile for it, and others wouldn't get the warning.

PR #1200, opened 2018-05-23
FULLPIPE: fix URI

randomstuff:
The previous URI spawns many useless subshells on the POSIX backend
through (harmless) shell command injection.


sev-:
Thanks! merging

PR #1203, opened 2018-05-26
SAGA: Don't show paws message in saved thumbnail

dafioram:
Fixes Trac#10008. 

Previously, the game would show a paused game message in the thumbnail when the
game was saved using the gmm.


sev-:
Thanks! Merging

PR #1205, opened 2018-05-29
ALL: Use deg2rad and rad2deg for degree conversions

dafioram:
In several engines various M_PI/180.0, its reciprocal, and engine specific functions (radianToDegree...) have been replaced with these degree conversions functions from common math.

In sci I changed a few doubles to floats since the extra precision/expensiveness seemed unnecessary.

### DIFF
--- a/engines/fullpipe/scenes/scene18and19.cpp
+++ b/engines/fullpipe/scenes/scene18and19.cpp
@@ -33,6 +33,7 @@
 #include "fullpipe/interaction.h"
 #include "fullpipe/behavior.h"
 
+#include "common/math.h"
 
 namespace Fullpipe {
 
@@ -48,8 +49,6 @@ struct Swinger {
 	int sfield_24;
 };
 
-
-#define ANGLE(x) ((x) * M_PI / 180)
 
 void scene18_preload() {
 	g_fp->_scene3 = 0;
@@ -149,7 +148,7 @@ void scene18_setupSwingers(StaticANIObject *ani, Scene *sc) {
 	for (int i = 0; i < 8; i++) {
 		swinger = new Swinger;
 
-		swinger->angle = (double)i * ANGLE(45);
+		swinger->angle = (double)i * Common::deg2rad<double>(45.0);
 		swinger->sx = g_vars->scene18_wheelCenterX - (int)(cos(swinger->angle) * -575.0);
 		swinger->sy = g_vars->scene18_wheelCenterY - (int)(sin(swinger->angle) * -575.0) + 87;
 		swinger->ix = swinger->sx;
@@ -555,11 +554,11 @@ void sceneHandler18and19_drawRiders() {
 
 		double oldangle = swinger->angle;
 
-		swinger->angle += ANGLE(1);
+		swinger->angle += Common::deg2rad<double>(1.0);
 
-		if (swinger->angle > ANGLE(360)) {
-			swinger->angle -= ANGLE(360);
-			oldangle -= ANGLE(360);
+		if (swinger->angle > Common::deg2rad<double>(360.0)) {
+			swinger->angle -= Common::deg2rad<double>(360.0);
+			oldangle -= Common::deg2rad<double>(360.0);
 		}
 
 		int ix = g_vars->scene18_wheelCenterX - (int)(cos(swinger->angle) * -575.0);
@@ -568,7 +567,7 @@ void sceneHandler18and19_drawRiders() {
 		if (!g_vars->scene18_rotationCounter) {
 			ix = swinger->sx;
 			iy = swinger->sy;
-			swinger->angle = (double)i * ANGLE(45);
+			swinger->angle = (double)i * Common::deg2rad<double>(45.0);
 		}
 
 		if (swinger->ani->_movement)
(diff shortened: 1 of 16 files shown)


engines/sword25/util/lua/lmathlib.cpp
```diff
@@ -131,12 +131,12 @@ static int math_exp (lua_State *L) {
 }
 
 static int math_deg (lua_State *L) {
-  lua_pushnumber(L, luaL_checknumber(L, 1)/RADIANS_PER_DEGREE);
+  lua_pushnumber(L, Common::rad2deg<double>(luaL_checknumber(L, 1)));
```
criezy: Even though it doesn't change anything, maybe use `Common::rad2deg<lua_Number>(...)` here and below since the code in lua seems to use that everywhere (probably so that it can easily be changed to use single precision floating points instead of double). Although because of the define, the old code would have been using double precision arithmetic (before casting back to float) even if lua_Number had been defined to float.

engines/fullpipe/scenes/scene18and19.cpp
```diff
@@ -22,17 +22,16 @@
...
 #include "fullpipe/gameloader.h"
+#include "fullpipe/interaction.h"
 #include "fullpipe/motion.h"
+#include "fullpipe/objectnames.h"
 #include "fullpipe/scenes.h"
 #include "fullpipe/statics.h"
 
-#include "fullpipe/interaction.h"
-#include "fullpipe/behavior.h"
-
```
criezy: I am not sue about reordering the includes alphabetically here. There is sometimes a logic behind the order in which they are done.

engines/titanic/star_control/star_closeup.cpp
```diff
@@ -174,9 +175,8 @@ bool CStarCloseup::setup2(int val1, int val2) {
 				e->_pixel2 = (val >> 8) & 0xff;
 				e->_pixel3 = (val >> 16) & 0xff;
 				e->_field8 = g_vm->getRandomNumber(3) + 3;
-
 				e->_fieldC = g_vm->getRandomNumber(255);
-				e->_field10 = FACTOR * (float)g_vm->getRandomNumber(15);
+				e->_field10 = Common::deg2rad<double>((double)g_vm->getRandomNumber(15));
```
criezy: I see this one was changed from a single precision floating point arithmetic to double precision, but I can see why you did it. Also since this is on a random number, it is unlikely to break anything.

The whole file is a bit strange. The `_field10` is a double as far as I can see, it was using a float `FACTOR`, but all the floating point literal values were double. So it ended up doing double precision arithmetic, but with limited precision on the factor. This means your changes might result in a slightly different value for those as well, but it looks like that is what was intended. @dreammaster might want to check this though.

engines/mohawk/myst_stacks/myst.cpp
```diff
@@ -3246,9 +3247,9 @@ Common::Point Myst::towerRotationMapComputeCoords(uint16 angle) {
 	Common::Point end;
 
 	// Polar to rect coords
-	double radians = angle * M_PI / 180.0;
-	end.x = (int16)(_towerRotationCenter.x + cos(radians) * 310.0);
-	end.y = (int16)(_towerRotationCenter.y + sin(radians) * 310.0);
+	float radians = Common::deg2rad<uint16,float>(angle);
+	end.x = (int16)(_towerRotationCenter.x + cos(radians) * 310.0f);
+	end.y = (int16)(_towerRotationCenter.y + sin(radians) * 310.0f);
```
criezy: What is the reason for changing the code to use simple precision here? Because of the cast to int16 at the end? Due to the change of precision, this might in some edge case end up casting to a different integral value. This might be fine, but @bgK might want to check this change.

engines/sci/engine/kmath.cpp
```diff
@@ -206,7 +208,7 @@ reg_t kGetDistance(EngineState *s, int argc, reg_t *argv) {
 	int xdiff = (argc > 3) ? argv[3].toSint16() : 0;
 	int ydiff = (argc > 2) ? argv[2].toSint16() : 0;
 	int angle = (argc > 5) ? argv[5].toSint16() : 0;
-	int xrel = (int)(((float) argv[1].toSint16() - xdiff) / cos(angle * M_PI / 180.0)); // This works because cos(0)==1
+	int xrel = (int)(((float) argv[1].toSint16() - xdiff) / cos(Common::deg2rad<int,float>(angle))); // This works because cos(0)==1
```
criezy: I'd like for a developer familiar with the SCI engine to indeed confirm that the double precision is unnecessary here and in the other changes below.

engines/scumm/actor.cpp
```diff
@@ -460,7 +461,7 @@ void Actor::setActorWalkSpeed(uint newSpeedX, uint newSpeedY) {
 int getAngleFromPos(int x, int y, bool useATAN) {
 	if (useATAN) {
 		double temp = atan2((double)x, (double)-y);
-		return normalizeAngle((int)(temp * 180 / M_PI));
+		return normalizeAngle(Common::deg2rad<int>(temp));
```
criezy: This one should be `rad2deg`!

engines/scumm/he/moonbase/ai_main.cpp
```diff
@@ -20,14 +20,16 @@
...
+
+#include "scumm/he/intern_he.h"
+#include "scumm/he/moonbase/ai_pattern.h"
 #include "scumm/he/moonbase/ai_targetacquisition.h"
+#include "scumm/he/moonbase/ai_traveller.h"
 #include "scumm/he/moonbase/ai_types.h"
-#include "scumm/he/moonbase/ai_pattern.h"
+#include "scumm/he/moonbase/moonbase.h"
+
+#include "common/math.h"
```
criezy: Same comment as for fullpipe: I don't really see a reason to change the order of the includes.

engines/scumm/he/moonbase/ai_main.cpp
```diff
@@ -378,8 +380,8 @@ int AI::masterControlProgram(const int paramCount, const int32 *params) {
 		if (!getCoordinateVisibility(targetX, targetY, currentPlayer)) {
 			int closestHub = getClosestUnit(targetX, targetY, getMaxX(), currentPlayer, 1, BUILDING_MAIN_BASE, 1, 0);
 			int targetAngle = calcAngle(getHubX(closestHub), getHubY(closestHub), targetX, targetY);
-			int testX = static_cast<int>(getHubX(closestHub) + (500 * cos(degToRad(targetAngle))) + getMaxX()) % getMaxX();
-			int testY = static_cast<int>(getHubY(closestHub) + (500 * sin(degToRad(targetAngle))) + getMaxY()) % getMaxY();
+			int testX = static_cast<int>(getHubX(closestHub) + (500 * cos(Common::deg2rad<float>(targetAngle))) + getMaxX()) % getMaxX();
```
criezy: This doesn't change anything in the end, but for consistency you might want to use `Common::deg2rad<int,float>()` here and on the line below since this is what you are doing in the rest of the file.

engines/scumm/actor.cpp
dafioram: good catch.

engines/mohawk/myst_stacks/myst.cpp
bgK: It's fine. Using double precision is not needed here.

dafioram:
Done. To make this PR more manageable I have dropped SCUMM and SCI changes.

engines/sword25/util/lua/lmathlib.cpp
```diff
@@ -131,12 +131,12 @@ static int math_exp (lua_State *L) {
 }
 
 static int math_deg (lua_State *L) {
-  lua_pushnumber(L, luaL_checknumber(L, 1)/RADIANS_PER_DEGREE);
+  lua_pushnumber(L, Common::rad2deg<lua_Number>(luaL_checknumber(L, 1)));
```
sev-: This is an imported file. I would rather abstain from changing it at all.

sev-:
Good PR. Adds more uniformity, makes the code cleaner.

The only one precaution which I have are the changes to the Lua code. That code is imported, and I would try to not change it much for the sake of possible future upgrades. Though, I am not insisting.

dafioram:
Removed lua/sword25 changes.

bonki:
This looks ready to merge. @sev- @criezy?

criezy:
Thank you. I am merging this now as all the potentially debatable changes have been removed from the PR.

PR #1206, opened 2018-05-29
Fix RTF credits encoding inconsistencies

travismckay:
I happened to notice that the MacOS version of ScummVM had a few character encoding issues in the credits, where some HTML entities were not being properly translated.  I went through the credits.pl script and checked each of the entities from html_entities_to_ascii to ensure that they were also correctly defined in html_entities_to_rtf.

While I was in making changes, I also noticed that there is still a function containing translations for the TeX version - which seems to have been deprecated.  I took the liberty of removing that function, to reduce future confusion.

### DIFF
--- a/devtools/credits.pl
+++ b/devtools/credits.pl
@@ -155,31 +155,6 @@ sub html_entities_to_rtf {
 	return $text;
 }
 
-# Convert HTML entities to TeX codes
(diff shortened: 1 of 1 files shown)


criezy:
Good catch! Thank you for your contribution.

PR #1207, opened 2018-06-01
GUI: Grey out unlaunchable games in launcher

bonki:


gui/launcher.cpp
```diff
@@ -282,13 +286,26 @@ void LauncherDialog::updateListing() {
 
 			while (pos < size && (scumm_stricmp(description.c_str(), l[pos].c_str()) > 0))
 				pos++;
+
+			EngineMan.findGame(gameid, &plugin);
```
bgK: When using dynamic plugins, this causes the engine plugins to be repeatedly loaded and unloaded. Is that going to be too slow?

criezy:
The idea looks good to me, but I am also wondering what the performance impact might be when using dynamic plugins.

sev-:
It is going to be slow, especially on the lower end devices with plugins enabled.

What is the purpose of it? I mean, how somebody could get the game in the detector in the first place, and then have it non-runnable? Of course, besides when the game directory is removed, but then the user should know about it, right?

If, however, you're thinking about use case when only a few engines are compiled in during development, then I would recommend taking a look at the problem once again, once we have the PR with engine name landed. Then you could just iterate over the engine list.

bonki:
When testing I usually mass-extract games from my archive to a ramdrive so after rebooting I have dangling entries in the launcher and it's not immediately obvious which are on persistent storage and which don't exist anymore so this makes it much easier for me to spot games which have been removed from the FS. The case you mention when testing with only a few engines compiled in during development is the second use case, but less important to me personally.

I agree that this is not a good idea currently with plugins enabled. I can remove the engine-check for now and only leave the FS-check in place as I still think this would be useful to have. Thoughts?

criezy:
Maybe the same functionality could be provided as a command line command (such as `--check-targets`) to list invalid entries? Actually we could maybe even have a `--clean-targets` command line command to clean the list of games by removing all invalid entries? And maybe while we are at it a `--remove --game=TARGET` might be good as well?

bonki:
I do think the CLI should provide these switches but I don't see them replacing the proposed GUI feature.
I'd still like to have a visual indicator which entries are working, it's just good UX in my book (it can happen to any ordinary user that directories are missing or inaccessible for whatever reason). Just think of the case of having inaccessible NFS/SMB mounts - just because some entries are not working *right now* doesn't *necessarily* mean I want to remove any or all entries.

bgK:
> I agree that this is not a good idea currently with plugins enabled. I can remove the engine-check for now and only leave the FS-check in place as I still think this would be useful to have.

Sounds good to me. Most of my games are on a noisy mechanical hard drive I only plug when I need to access the data it contains. Having the unavailable entries grayed out sounds like an improvement to me.

bonki:
Done. Is anyone still against merging this?

bonki:
Merging since no objections to the latest update have been voiced, thanks for the feedback.

PR #1208, opened 2018-06-01
SCUMM: Work around distorted speech on submarine in Indy4

bonki:
The speech sample at offset 0x76ccbca ("Hey you!") which is used when
Indy gets caught on the German submarine seems to not be a VOC but plain
PCM s16be at 44.1 kHz with a bogus VOC header.
To work around this we play this sample using the raw PCM decoder.

Fixes [Trac#10559](https://bugs.scummvm.org/ticket/10559).

### DIFF
--- a/engines/scumm/sound.cpp
+++ b/engines/scumm/sound.cpp
@@ -492,6 +492,8 @@ void Sound::startTalkSound(uint32 offset, uint32 b, int mode, Audio::SoundHandle
 #endif
 	Common::ScopedPtr<ScummFile> file;
 
+	bool _sampleIsPCMS16BE44100 = false;
+
 	if (_vm->_game.id == GID_CMI) {
 		_sfxMode |= mode;
 		return;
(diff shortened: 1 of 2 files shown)


jepael:
Sample format in the file appears to be 16-bit signed PCM even if VOC header indicates 8-bit unsigned PCM.

Please check the [Track#10559](https://bugs.scummvm.org/ticket/10559) page for more info.

eriktorbjorn:
Should the workaround check explicitly that it's the English version, just in the unlikely case that some other version happens to have a sample at that exact offset?

bonki:
@jepael Good catch, the sample seems to be there. However, to me it appears to be plain PCM signed 16 bit big endian with a garbage VOC header. Can you confirm this? Let's continue the discussion in Trac until we know how to properly fix this.

@eriktorbjorn Doesn't the talkie only exist in English? Either way, you're right and I'll add this for now even though it looks like we will fix this in another way.

bonki:
I have now updated the PR with the changed approach (play the sample as raw signed 16 bit big endian 44100 Hz PCM) which seems to work fine. This means that we now probably also have to add a workaround to `compress_scumm_sou`.

Can we fix this in a less hacky way?

eriktorbjorn:
> Doesn't the talkie only exist in English?

I don't know. I just assumed there had to have been at least a German talkie version at some point, but when I looked around it seems there wasn't...?

criezy:
I checked both versions I have (original mac CD and Steam Mac, both English). They both exhibit the issue and in both cases this is fixed with the changes in this PR.

engines/scumm/sound.cpp
```diff
@@ -584,10 +586,23 @@ void Sound::startTalkSound(uint32 offset, uint32 b, int mode, Audio::SoundHandle
...
 #endif
 		} else {
-			offset += 8;
+			// WORKAROUND
+			// The speech sample at offset 0x76ccbca ("Hey you!") which is used when
+			// Indy gets caught on the German submarine seems to not be a VOC but plain
+			// PCM s16be at 44.1 kHz with a bogus VOC header.
+			// To work around this we play this sample using the raw PCM decoder.
+			// Fixes Trac#10559
+			if (mode == 2 && (_vm->_game.id == GID_INDY4) && (_vm->_language == Common::EN_ANY) && offset == 0x76ccbca) {
```
sev-: What about size check also, as a precaution?

bonki: Which size?

sev-: size of the in-game sample, the one which you're overriding with 86016.

bonki: I don't override it because it's not provided in the first place, the script only references an offset when playing a sample. `size` is otherwise only used when dealing with a compressed `monster.sou` where `size` comes from the hard-coded offset index table. I have to set it because the `SeekableSubReadStream` c-tor requires a size. `makeVOCStream` (normal code path) on the other hand doesn't because the VOC decoder has to parse the entire VOC to find the size of the audio data.

bonki: (in other words, there is no size to check)

bonki:
Pushing a slightly updated commit. Thanks for testing!

PR #1213, opened 2018-06-10
DISTS: Update appdata.xml file

AsciiWolf:
Add [OARS](https://hughsie.github.io/oars/) metadata.

### DIFF
--- a/dists/scummvm.appdata.xml
+++ b/dists/scummvm.appdata.xml
@@ -25,4 +25,33 @@
   </screenshots>
   <url type="homepage">https://www.scummvm.org</url>
   <update_contact>scummvm-devel@lists.scummvm.org</update_contact>
+  <content_rating type="oars-1.1">
+    <content_attribute id="violence-cartoon">none</content_attribute>
+    <content_attribute id="violence-fantasy">none</content_attribute>
+    <content_attribute id="violence-realistic">none</content_attribute>
+    <content_attribute id="violence-bloodshed">none</content_attribute>
+    <content_attribute id="violence-sexual">none</content_attribute>
+    <content_attribute id="violence-desecration">none</content_attribute>
+    <content_attribute id="violence-slavery">none</content_attribute>
+    <content_attribute id="violence-worship">none</content_attribute>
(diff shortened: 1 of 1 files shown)


AsciiWolf:
@criezy Ok, I have removed the version info part. Regarding the OARS metadata, I have asked [here](https://github.com/flathub/flathub/pull/425#discussion_r194269202), so let's wait for the answer.

AsciiWolf:
By the way, I am making a [Flatpak](https://flatpak.org) build of ScummVM for [Flathub](https://flathub.org). Feel free to let me know if you are interested in maintaining it.

AsciiWolf:
@criezy [Here](https://github.com/flathub/flathub/pull/425#discussion_r194396991) is the answer about OARS.

AsciiWolf:
So the current commit should be ok to merge. ;-)

criezy:
That looks OK to me. Also I slightly misunderstood what the version was in your initial commit and I have no objections in releases being listed if that is expected.

AsciiWolf:
Cool! But you were right that the version info should be updated every time a new stable version is released. So let's merge only the OARS metadata for now...

AsciiWolf:
@criezy Are there any other issues with this PR?

bonki:
If the release version needs to be part of the metadata we should add this in one go. What's required (@criezy correct me if I'm wrong) is to duplicate `dists/scummvm.appdata.xml` => `dists/scummvm.appdata.xml.in` using `@VERSION@` as placeholder and add `dists/scummvm.appdata.xml` to the list of files [here](https://github.com/scummvm/scummvm/blob/master/devtools/update-version.pl#L33). Would you like to add this change to your commit?

AsciiWolf:
@bonki I can do it. But this will only update the release version, not the release date.

AsciiWolf:
So the _update-version.pl_ script either needs to be updated to support replacing the `@DATE@` string with the correct date (I unfortunately do not have enough free time to do this) or this PR has to be merged in the current form (without the version part).

bonki:
Since the release metadata doesn't seem to be part of or required for OARS we'll merge this as-is for now, thanks!

PR #1216, opened 2018-06-15
SCI32: Only enable larry scaler for LSL7

dafioram:
Fixes Trac#10568.

### DIFF
--- a/engines/sci/graphics/celobj32.cpp
+++ b/engines/sci/graphics/celobj32.cpp
@@ -202,7 +202,7 @@ struct SCALER_Scale {
 
 		const CelScalerTable &table = CelObj::_scaler->getScalerTable(scaleX, scaleY);
 
-		const bool useLarryScale = ConfMan.getBool("enable_larryscale");
+		const bool useLarryScale = (g_sci->getGameId() == GID_LSL7) && ConfMan.getBool("enable_larryscale");
 		if (useLarryScale) {
 			// LarryScale is an alternative, high-quality cel scaler implemented
 			// for ScummVM. Due to the nature of smooth upscaling, it does *not*


DanielSWolf:
This MR fixes the immediate issue at hand, but I'm not convinced about the way it does that.

As discussed [in the original MR](https://github.com/scummvm/scummvm/pull/1141#issuecomment-375586622), the LarryScale algorithm is not exclusive to LSL7. It could be used to reduce scaling artifacts for a number of games, including Day of the Tentacle, Sam and Max hit the Road, and Toonstruck.

Because of this, the code in `celobj32.cpp` should *not* explicitly check whether the game ID is `GID_LSL7`, as this MR does. Instead, it should check whether the current game is *any* game that supports the LarryScale option. @eriktorbjorn and I discussed various approaches in the Trac issue, and I believe that he [found the correct solution](https://bugs.scummvm.org/ticket/10568#comment:9):

```cpp
const bool useLarryScale = Common::checkGameGUIOption(GAMEOPTION_LARRYSCALE, ConfMan.get("guioptions")) && ConfMan.getBool("enable_larryscale");
```

The only reason he and I haven't already submitted a MR with this line is that we're both not familiar with the implementation of ScummVM's options system, so we weren't sure whether this line does what we think it does.

rsn8887:
@DanielSWolf: I don't understand your point, since celobj32.cpp is SCI-engine specific. So if dafioram's patch works, it seems like it is a valid method and should be merged.

rsn8887:
I understand that once Larry scale is enabled for any other games, this should be re-visited. But for now, I think this is enough.

bonki:
I wholeheartedly agree with @DanielSWolf that this should be fixed in a cleaner way and don't understand why you felt the need to rush this while Daniel and @eriktorbjorn are working on a proper fix. Neither do I understand why this needs to be fixed "ASAP", it is understood and expected that master and daily builds are not stable and the ticket properly documents that the game is broken and needs to be fixed before the next release, so it's not going to be forgotten either. Yes, this needs to be addressed in a timely manner, but this *was* being worked on and merging without giving anyone else the chance to comment wasn't the right thing to do in my book.

dafioram:
Nothing about the commit current in master precludes a PR from being opened to apply the larry scaler in the correct way. 

Having this commit in allows phant1 to be played and if we never open the PR to do the larry scaler the right way we are still fine. The PR with the fix can even revert this commit as a part of the fix.

rsn8887:
I agree with @dafioram here. This commit does by no means exclude further work on the issue. I welcome @eriktorbjorn and/or @DanielSWolf to open another PR, once they have tested their solution properly.

I do not think this fix was rushed in any way. It was properly tested. It is a very minor and very clear and easily understood change. It fixes an enormous issue: the breakage, slow gfx and other gfx issues in many, if not all, SCI games (apart from LSL7).

My opinion is that the previous decision to make LarryScale explicitly default to `true` for all SCI games was a major change that was rushed and improperly tested or understood.

bonki:
> Nothing about the commit current in master precludes a PR from being opened to apply the larry scaler in the correct way.

This doesn't invalidate anything I wrote, however.

> Having this commit in allows phant1 to be played and if we never open the PR to do the larry scaler the right way we are still fine. The PR with the fix can even revert this commit as a part of the fix.

The game can still be played by any proper build, irrespective of this being applied or not.

> I do not think this fix was rushed in any way. It was properly tested. It is a very minor and very clear and easily understood change.

It is a workaround for something that didn't need a workaround. It was rushed because a proper fix was being worked on. Our commit guidelines state that touching code which one is not ultimately familiar with or responsible for or changes common code should result in a PR which should stay open for an appropriate amount of time to allow peer review and so other developers get the chance to comment on the changes being made. You merging this within a mere couple of hours of the PR being opened while others were discussing on the bug tracker about how to fix this in the correct way and in the spirit of the original PR is in violation of these guidelines. I would have welcome a contribution to the discussion in the ticket instead. Pulling the rug under someone else actively working on a ticket without warning is not very respectful.

> My opinion is that the previous decision to make LarryScale explicitly default to true for all SCI games was a major change that was rushed and improperly tested.

I do agree. And as you should know it was never intended to default to true for all SCI games, so this clearly was a bug in the earlier code which was merged. Maybe the original PR should have been reverted while working on the proper fix (?).

criezy:
Making LarryScale the default for all SCI games was a bug. It was never intended. Saying this was a rushed decision seems a bit strange to me considering the pull request remained open for two months and a half before being merged. I would have welcomed more reviews from others during that time, but I didn’t feel they were coming, which was one of the reason I decided to merge the PR. I can agree this was not properly tested though since I missed the issue. To my defense I didn’t expect any game other than LSL7 to be impacted, as I missed this while reviewing the code change, and thus didn’t test other games. But another reason I decided to merge was also to get this tested by more people with the unstable daily builds. And it seems that part worked, and the bug was found as a result ;-)

I don’t really see why it was so urgent to fix though (which is also why I decided not to revert the change when the issue was reported and while it was being worked on). It seems to me we could have waited a few more days to apply a proper fix. If the compilation was broken this would have been a very serious issue (actually merging the PR did break some builds, which fortunately I managed to fix quickly, and otherwise I would have reverted it). And I do agree with what Adrian wrote about communication. There was a discussion ongoing in the bug tracker, the PR was then open, and I have no issue about this, but I feel it should have been part of the discussion. But it was merge so quickly I, and probably many others, didn’t even have time to see it. I still feel merging the PR so quickly was unnecessary, but I also feel dwelling on it is unnecessary. While I felt we could have waited and that what was committed is not the most appropriate fix, I don’t see why we should turn this into a drama. I don’t feel the last commit needs to be reverted. But I do feel we need to work on a better fix (at first glance the one being discussed in the bug tracker ticket looks OK).

bonki:
I agree with everything you wrote so I'll leave it at that. One minor clarification because my quote was slightly misleading: As I did agree with you earlier on that merging the original PR would be a good idea to make it easier for people to test and the PR had been open for a while I can't call that one rushed. It *was*, however, a major change and improperly tested (which is what I meant above when I wrote that I agreed) which is why it got merged in the first place. Everybody who looked at the original PR missed that bug and I don't blame anyone for that and, as you said, it did get caught after all :-)).

I'd also like to clarify that my intention is not (and never was) to cause drama. I understand that the content of the commit per se is not a huge issue, but I still believe that this could have been handled better and shouldn't have been merged so quickly and I felt that leaving this uncommented on might have set a bad precedent, that's all.

rsn8887:
I apologize if I violated the guidelines.

I considered the bug to be of comparable severity as a broken compilation, since it affected potentially all SCI games, causing crashing, performance issues and/or gfx corruption. I considered the fix to be a very simple, logical and trivial fix that works which would not require much review. Those two considerations were the reason I merged this PR quickly. I saw it as a simple fix to a severe issue. 

My intention was simply to make all SCI games playable again in any daily builds or builds from master branch. I think the master branch is used by many users on many platforms, because it has by now quite significantly advanced over the last proper release. For example, there have been some major improvements to the smoothness of gamepad controls etc. since the last major release.

However, I fail to understand why this would pull the rug under anyone. This one commit does not stop the work by others. I was always under the impression that @DanielSWolf or @eriktorbjorn were planning to do their work as their own, separate, new PR anyways. The bug does not have to be present for anybody to keep working on improving the code related to LarryScale further.

PR #1217, opened 2018-06-15
SCUMM: fix erroneous costume animation in spyozon

svenmeier:
Don't fail when an error is encountered in an costume animation, but reset to start location instead.

This is essential to be able to complete Spy Fox 3 German, which contains an erroneous AKC_Jump location in an LED ticker animation in the hair-dresser room.

### DIFF
--- a/engines/scumm/akos.cpp
+++ b/engines/scumm/akos.cpp
@@ -1632,6 +1632,17 @@ bool ScummEngine_v6::akos_increaseAnim(Actor *a, int chan, const byte *aksq, con
 
 		case AKC_Jump:
 			curpos = GUW(2);
+
+			// WORKAROUND for bug #3813: In the German version of Spy Fox Ozone the the hair dresser room 21
+			// contains a costume animation 352 of an LED ticker with a jump to an erroneous position 846.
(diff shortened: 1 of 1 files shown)


Kirben:
I don't know changing the error to a warning is a good idea. If only a single AKOS error occurs a work around specific to that error would be better.

svenmeier:
@Kirben  Thanks for your feedback.

How should the workaround be added? Just a condition some where in the source?

Is there a workaround for this type of AKOS error already I could use a a reference? It's not clear to me how the faulty AKOS script should be identified.

Kirben:
Add only the the specific case required by the game, if possible. Also include comments explaining the game and version the problem occurs in.

svenmeier:
I've added a workaround specifically for the problematic costume animation.

bonki:
I can have a look and test this.

svenmeier:
I've updated the change so the workaround follows the structure of other workarounds.

raziel-:
Typo:
"the the hair dresser room 21" - double "the"

bonki:
This behaves like the original now and renders the German version completable as well, thanks! I landed this in commit 64c54591c356f41abe8df7de7afbd291279a99a5.

PR #1219, opened 2018-06-18
BLADERUNNER: Adding Subtitles support

antoniou79:
Attaching the external resource files for the subtitles and the optional font. This should be copied in the game installation folder. 
[BladeRunnerSubtitlesExtraResources180618.zip](https://github.com/scummvm/scummvm/files/2111259/BladeRunnerSubtitlesExtraResources180618.zip)


engines/bladerunner/actor.cpp
```diff
@@ -40,6 +40,9 @@
 #include "bladerunner/set.h"
 #include "bladerunner/slice_animations.h"
 #include "bladerunner/slice_renderer.h"
+#if SUBTITLES_SUPPORT
```
sev-: Why do you make it optional compilation at all?

engines/bladerunner/bladerunner.cpp
```diff
@@ -323,6 +329,20 @@ bool BladeRunnerEngine::startup(bool hasSavegames) {
 	_gameFlags = new GameFlags();
 	_gameFlags->setFlagCount(_gameInfo->getFlagCount());
 
+#if BLADERUNNER_RESTORED_CONTENT_GAME
+	// EDS flags
+	_extraGameFlagsForRestoredContent = new GameFlags(); //aux flags - custom
```
sev-: I think this is an overkill. You could just have it a normal GF_ and then set it at the presence of the subtitles file.

engines/bladerunner/bladerunner.cpp
```diff
@@ -323,6 +329,20 @@ bool BladeRunnerEngine::startup(bool hasSavegames) {
...
+	_extraGameFlagsForRestoredContent->setFlagCount(RESTORED_CONTENT_EXTRA_FLAGS + 1); // aux flags - custom // +1 since we don't assign something to 0 enum value
+#if SUBTITLES_SUPPORT
+#if SUBTITLES_ENABLED_BY_DEFAULT
```
sev-: Instead of making it compile-time, you could just read from our common GUI option, where we have subtitles settings. Just start using it for the engine.

engines/bladerunner/bladerunner.cpp
```diff
@@ -863,7 +894,9 @@ void BladeRunnerEngine::gameTick() {
 			if (_debugger->_viewObstacles) {
 				_obstacles->draw();
 			}
-
+			#if SUBTITLES_SUPPORT
+			_subtitles->tick(_surfaceFront);
+            #endif // SUBTITLES_SUPPORT
```
sev-: Indentation is weird here.

engines/bladerunner/bladerunner.h
```diff
@@ -126,12 +137,19 @@ class BladeRunnerEngine : public Engine {
 	EndCredits         *_endCredits;
 	ESPER              *_esper;
 	GameFlags          *_gameFlags;
+	#if BLADERUNNER_RESTORED_CONTENT_GAME
```
sev-: What is the restored content?

engines/bladerunner/font.cpp
```diff
@@ -37,6 +37,48 @@ Font::~Font() {
 	close();
 }
 
+#if SUBTITLES_SUPPORT
+#if SUBTITLES_EXTERNAL_FONT
+// for external FON font file / subtitles support
+bool Font::openFromStream(Common::ScopedPtr<Common::SeekableReadStream> &stream, int screenWidth, int screenHeight, int spacing1, int spacing2, uint16 color) {
+    reset();
```
sev-: Spaces instead of tabs.

engines/bladerunner/font.cpp
```diff
@@ -37,6 +37,48 @@ Font::~Font() {
...
+	_screenHeight = screenHeight;
+	_spacing1 = spacing1;
+	_spacing2 = spacing2;
+	_color = color;
+
+	if (!stream) {
+		return false;
+	}
+	_characterCount = stream->readUint32LE();
+    debug("Font's character count: %d", _characterCount);
```
sev-: Debug leftover? And indentation is wrong.

engines/bladerunner/font.cpp
```diff
@@ -68,6 +110,28 @@ bool Font::open(const Common::String &fileName, int screenWidth, int screenHeigh
 		_characters[i].width = stream->readUint32LE();
 		_characters[i].height = stream->readUint32LE();
 		_characters[i].dataOffset = stream->readUint32LE();
+		#if SUBTITLES_SUPPORT
```
sev-: The formatting for this whole block is all messed up.

engines/bladerunner/subtitles.cpp
```diff
@@ -0,0 +1,778 @@
...
+			_subsBgFont->draw(_subtitleLineQuote[i], s, _subtitleLineScreenX[i] - 1, _subtitleLineScreenY[i] + 1);
+		}
+		_subsFont->draw(_subtitleLineQuote[i], s, _subtitleLineScreenX[i],  _subtitleLineScreenY[i]);
+	}
+#endif // SUBTITLES_EXTERNAL_FONT
+}
+
+/**
+* Calculate the position (X axis - horizontal) where the current active subtitle text should be displayed/drawn
+* This also determines if more than one lines should be drawn and what text goes into each line; splitting into multiple lines is done here
```
sev-: Why not calculate the bounding box and then just use wordWrap() function plus aligned text rendering here? This could simplify things big time.

engines/bladerunner/ui/vk.cpp
```diff
@@ -456,9 +463,9 @@ void VK::init() {
 	_buttons->defineImage(2, Common::Rect(515, 368, 538, 398), nullptr,    _shapes[6],  _shapes[7],  nullptr);
 	_buttons->defineImage(3, Common::Rect(548, 368, 571, 398), nullptr,    _shapes[8],  _shapes[9],  nullptr);
 	_buttons->defineImage(4, Common::Rect(581, 368, 604, 398), nullptr,    _shapes[10], _shapes[11], nullptr);
-	_buttons->defineImage(5, Common::Rect( 31, 363,  65, 392), nullptr,    _shapes[0],  _shapes[1], _vm->_textVK->getText(0));
```
sev-: This was done on purpose, for the horizontal alignment.

engines/bladerunner/ui/vk.cpp
```diff
@@ -894,7 +901,7 @@ void VK::askQuestion(int intensity) {
 			// TODO: related questions are not used in game
 			// int relatedQuestion = -1;
 			// if (_questions[intensity][i].relatedSentenceId >= 0) {
-			// 	relatedQuestion = vk::findQuestionById(this, questions, relatedQuestionId);
+			//  relatedQuestion = vk::findQuestionById(this, questions, relatedQuestionId);
```
sev-: This is plainly wrong change

engines/bladerunner/ui/vk.cpp
```diff
@@ -683,10 +690,10 @@ void VK::drawNeedle(Graphics::Surface &surface) {
 	int g =  8 * colorIntensity;
 	int b = 12 * colorIntensity;
 
-	surface.drawLine(203, 324, x - 2, y,     ((7 - r    ) << 10) | ((18 - g    ) << 5) | (23 - b    ));
```
peterkohaut: some of whitespace removal is ok but some, like this, are reducing readibility

engines/bladerunner/actor.cpp
```diff
@@ -1091,10 +1098,18 @@ void Actor::speechPlay(int sentenceId, bool voiceOver) {
 		balance = CLIP<int>(balance, -127, 127);
 	}
 
+#if SUBTITLES_SUPPORT
+	_vm->_subtitles->getInGameSubsText(_id, sentenceId);
```
peterkohaut: this api doesnt look very future proof, for dialog between more actors api is implying will just show one sentence. maybe something like subtitles->queue(actorid, sentenceid, length)

engines/bladerunner/font.cpp
```diff
@@ -192,6 +270,14 @@ void Font::drawCharacter(const uint8 character, Graphics::Surface &surface, int
 
 	int endY = height + y - 1;
 	int currentY = y;
+
+#if BLADERUNNER_RESTORED_CONTENT_GAME
+	// Temp Bug fix - Return if w h unnaturally big: -- the INTERNAL tahoma18 font is corrupted so it could cause crashes
```
bonki: Can you make the comment be a bit more elaborate why this is needed and maybe even mark this as `FIXME`/`TODO` if it really is a temporary workaround?

engines/bladerunner/font.cpp
```diff
@@ -68,6 +110,27 @@ bool Font::open(const Common::String &fileName, int screenWidth, int screenHeigh
 		_characters[i].width = stream->readUint32LE();
 		_characters[i].height = stream->readUint32LE();
 		_characters[i].dataOffset = stream->readUint32LE();
+#if SUBTITLES_SUPPORT
+#if !SUBTITLES_EXTERNAL_FONT
+		// special explicit alignment fixes for TAHOMA18 (INTERNAL) font
+		if (fileName == "TAHOMA18.FON") {
+			// fix P -> i = 81 (ascii code 80  + 1)
```
bonki: A switch statement would make this a bit more readable in my opinion.

engines/bladerunner/font.h
```diff
@@ -23,6 +23,10 @@
 #ifndef BLADERUNNER_FONT_H
 #define BLADERUNNER_FONT_H
 
+#include "bladerunner/bladerunner.h" // needed for definition of Common::ScopedPtr (subtitles font external font file support) -- and for the subtitles relevant macro defines
```
bonki: If this remains a compile-time option (which has already been stated that it shouldn't) the include could be moved inside the following `if` if it's not needed otherwise.

engines/bladerunner/actor.cpp
antoniou79: I'm using multiple such defines on another local repository where I'm working on restoring dialogue and recreating some "deleted" scenes. These defines help quickly locate where changes for a relevant feature were and disable them if something malfunctions, so I kept them here. I agree that they are not essential of course.

engines/bladerunner/subtitles.cpp
antoniou79: I wasn't aware of the wordWrap method. Still, the purpose of this code here is to allow subtitle quotes to be split in multiple lines first by checking for new line characters, and then by doing the wrapping thing. However, the wrapping code also makes sure that the line segments are approximately even (or close to that) with regard to their width. Is this something I can do with wordWrap()?

engines/bladerunner/ui/vk.cpp
antoniou79: As noted, this is a side-effect of me running AStyle without checking for its side-effects.

engines/bladerunner/ui/vk.cpp
antoniou79: Another AStyle  side-effect.

engines/bladerunner/bladerunner.h
```diff
@@ -37,6 +37,14 @@
 #define BLADERUNNER_DEBUG_CONSOLE 0
 #define BLADERUNNER_DEBUG_GAME 0
 
+#define SUBTITLES_SUPPORT   1
```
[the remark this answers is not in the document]
antoniou79: I'll be removing it in the next commit.

engines/bladerunner/font.cpp
```diff
@@ -37,6 +37,48 @@ Font::~Font() {
 	close();
 }
 
+#if SUBTITLES_SUPPORT
+#if SUBTITLES_EXTERNAL_FONT
+// for external FON font file / subtitles support
+bool Font::openFromStream(Common::ScopedPtr<Common::SeekableReadStream> &stream, int screenWidth, int screenHeight, int spacing1, int spacing2, uint16 color) {
```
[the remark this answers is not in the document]
antoniou79: I'm not sure I understand here. Are you proposing I merged the openFromStream() with the open() method? They are re-using much of the same code, but I thought it cleaner to keep them separate.

engines/bladerunner/ui/vk.cpp
antoniou79: This was AStyle. I'm sorry it ended messing some stuff up. I'll try to revert all of the style changes.

engines/bladerunner/actor.cpp
antoniou79: I agree that a queue system would be nicer and I've had it in mind as a future possible task/upgrade. But maybe it's not really needed? It seems to be working quite ok as is.
Also, I'm not sure what you mean by "future proof". Are there cases in-game where the current implementation would not work?

engines/bladerunner/font.cpp
antoniou79: I'll try to make it clearer. This has to do with one specific internal font (TAHOMA18) being corrupted - so trying to use/show some of the fonts glyphs (eg for debugging purposes; the bug affects characters with ascii value past the 173 or so) caused the game to crash.

engines/bladerunner/font.cpp
antoniou79: I could try that. The current if clauses and line separation with comments reflect my groupings of characters that had similar behavior as far as their placement/kerning go.

engines/bladerunner/font.h
antoniou79: You mean the bladerunner.h include? This was needed to be outside, since I set the SUBTITLES_SUPPORT macro in there (similar to how BLADERUNNER_DEBUG_GAME, BLADERUNNER_DEBUG_CONSOLE work there).

antoniou79:
@sev.
> 1. What is the advantage of using TRE format for subtitles?

TRE is the native format for text resources that the game uses. This meant I could mostly reuse the existing implemented mechanism to access and display the text in-game. Also, by implementing access to external TRE files, it possible to easily expand this and override internal TRE files, for purposes of localization in other languages.
> 2. The disadvantage of using the FON instead of TTF is that there will be problems with supporting non-Latin subtitles. Also, it adds tons of extra code.

The FON format has similar advantages - it is native to the game, and external FON files can support localizations of the game. You can use the FON file for non-latin subtitles (I can do Greek subtitles for example) in a similar fashion that this is done for the LucasArts games (Monkey island, Full throttle).
The functions to access the font glyphs properties and show them were already implemented. 
I'm not sure where it adds tons of extra code. Is it for the dev tools needed to create FON files? I have written a tool for this (and also for converting subtitles to TRE format) in python and I'll be sharing on github.
> 3.   What is the purpose of making the whole thing switchable at compile time, not just at run-time?

No real purpose. Just me trying to keep my code isolated and quickly locate changes for a specific feature. This can be removed. 
> 4. Asytle was too aggressive (...)

I agree and sorry about that. I'm not sure how I can go about squashing commits, but I'll ask for help in the IRC chat.
> 5. You are using some sophisticated algorithm for word wrapping. I suggest taking a look at our wordWrap.

It's a custom algorithm. The purpose is to prioritize new line characters in the quote as split points, and if such don't exist then try to split the quote in as even parts as possible without overflowing out of the screen width. I could use wordWrap, but would it substitute this functionality?
> 6. A shortcut. Instead of keeping the list of videos in code, you could just check for the file presence. The check is cheap, and simplifies code and makes it more universal.

The list of the videos is kept mainly so that I can map them to an uint ID which is then used to index their respective subtitles file. It also makes it easier to pre-load all the available subtitles at startup, and then access them from memory.

engines/bladerunner/font.h
bonki: Ah right, makes sense then, of course.

engines/bladerunner/actor.cpp
sev-: It would be nice to remove them for ScummVM, as we prefer to have as less global defines as possible

engines/bladerunner/bladerunner.cpp
sev-: Then I suggest switching to the normal GF_. This will simplify the code here.

engines/bladerunner/bladerunner.cpp
sev-: No, please don't remove them. Rather synchronise both (see syncSoundSettings() method).

engines/bladerunner/subtitles.cpp
```diff
@@ -0,0 +1,778 @@
...
+/*
+ * Optional support for subtitles
+ * CHECK what happens in VQA where the audio plays separately (are the finales such VQAs ?)
+ *
+ * TODO in python script (FON from png glyphs) check if you can have semi-transparent pixels to better outline the fringe points of the glyphs - check what happens when MSB is set (transparency) and the rest of the color value is not all 0s.
+ * TODO Catch error for bad symbol in a quote (one that causes the font to crash) - this could happen with the corrupted internal font (TAHOMA18) -> font crash or bad font display / garbage character
+ * TODO add a keyboard shortcut key to enable / disable subtitles?
+ * TODO have a debug script to detect/report problematic lines (too long)
+ *
+ * TODO? put external FON and TRE in a new folder "SUBS" - case insensitive (?)
```
sev-: Our filesystem code takes care of case sensitivity. Everything is lowercased before string comparison, so you can use any case.

sev-:
> @SEV.
> 
> | What is the advantage of using TRE format for subtitles?
> TRE is the native format for text resources that the game uses. This meant I could mostly reuse the existing implemented mechanism to access and display the text in-game. Also, by implementing access to external TRE files, it possible to easily expand this and override internal TRE files, for purposes of localization in other languages.

What I see is that you're not using a oneliner to read the subtitles, but have a full blown parser in TextResource::openFromStream(). I see, that you could alter this code very slightly, but get support of the plain text subtitles, which will be easier to maintain, translate, etc.

E.g. I understand the engineering beauty of your solution "I am using the game formats", but then, you anyway have to write code around it, and making the maintenance harder, thus, voiding this beauty.

I would highly recommend to switch to plaintext.

> | The disadvantage of using the FON instead of TTF is that there will be problems with supporting non-Latin subtitles. Also, it adds tons of extra code.
> The FON format has similar advantages - it is native to the game, and external FON files can support localizations of the game. You can use the FON file for non-latin subtitles (I can do Greek subtitles for example) in a similar fashion that this is done for the LucasArts games (Monkey island, Full throttle).
> The functions to access the font glyphs properties and show them were already implemented.
> I'm not sure where it adds tons of extra code. Is it for the dev tools needed to create FON files? I have written a tool for this (and also for converting subtitles to TRE format) in python and I'll be sharing on github.

Right. Supposedly I dreamed this off. Part of my brain remembers glyph widths manipulation. This must be your word wrapping code.

Another complication which you have is that #ifdef'fed code about external font file. I recommend to simplify that, e.g. either make it run-time, or leave only one branch.

As of the tools, please, always have those as part of the distribution. The good place is devtools/bladerunner directory.

> | What is the purpose of making the whole thing switchable at compile time, not just at run-time?
> No real purpose. Just me trying to keep my code isolated and quickly locate changes for a specific feature. This can be removed.

Yes, please simplify that as denoted earlier.

> 
> | Asytle was too aggressive (...)
> I agree and sorry about that. I'm not sure how I can go about squashing commits, but I'll ask for help in the IRC chat.
> 
> | You are using some sophisticated algorithm for word wrapping. I suggest taking a look at our wordWrap.
> It's a custom algorithm. The purpose is to prioritize new line characters in the quote as split points, and if such don't exist then try to split the quote in as even parts as possible without overflowing out of the screen width. I could use wordWrap, but would it substitute this functionality?

Yes, wordWrap is doing this at the moment, however, you mentioned that you also have algo for making strings look even, and that would require enhancing the wordWrap code. I prefer to enhance the existing code (and make it as a parameter), thatn duplicating the functionality.

> | A shortcut. Instead of keeping the list of videos in code, you could just check for the file presence. The check is cheap, and simplifies code and makes it more universal.
> The list of the videos is kept mainly so that I can map them to an uint ID which is then used to index their respective subtitles file. It also makes it easier to pre-load all the available subtitles at startup, and then access them from memory.

I misread your code. I thought that you do have a string ID, but you have only integer one. Your code makes sense now.

engines/bladerunner/font.cpp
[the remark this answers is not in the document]
peterkohaut: yes, exactly that, it looks almost the same and one can call the other

engines/bladerunner/actor.cpp
peterkohaut: it will be needed as few dialogs a bit fast, but you dont have to change right now. also this state persistence in the new code is not so nice looking. but as i wrote, you dont have to do it now

peterkohaut:
I think you can remove lot of "external" code if you add all to a new mix file and load that file in same way how others are loaded. There are few tools on internet which can do that.

antoniou79:
I've packed all TRE resource files and the external optional font in a SUBTITLES.MIX file (now attached in my original comment above). Like peterkohaut suggested this allowed removal of all the extra code for accessing external files (text resources and font files) which also was duplicating some of the existing functionality of the engine methods. 
I looked into how to merge with ScummVM's wordwrap algorithm, which is itself simple enough and straightforward, but so far I haven't come up with an efficient way to do what I'm doing using (or extending) the existing wordwrap. That's because my text wrap code in calculatePosition assumes access to the width (in pixels) of the font glyphs and takes that into consideration as well as the boundaries (in pixels) which is basically the screen width.

sev-:
Font::getCharWidth() gives you width of a character, and second parameter to wordWrapText() is your bounding box width.

peterkohaut:
How is this PR going? I tried it out and the subtitles are awesome!

antoniou79:
@peterkohaut as far as the transcribing goes, I've completed one and a half passes of it. It's fairly ok, but I do want to complete the second pass which is done as I'm progressing with my let's play project (currently beginning of Act 4), double-checking and verifying the lines. Once in a while, I spot quotes where corrections or improvements are needed. There are also a few cases where I'm not entirely certain of what is actually said by the actors and those will probably need the help of someone for whom English is the native language. 

The transcribed parts (in-game and movie cutscenes), as well as the fonts for the subtitles are now all packed in the SUBTITLES.MIX file, so that's all the external files anyone needs to enable subtitles. 

The subtitles text is still in the game's TRE format and the fonts in the game's FON format using the BladeRunner engine's own existing classes and structures. 

I do see the benefit of eventually supporting (somehow) the common .srt format and utf-8 encoding, as well as using ScummVM's own Font class (and thus the wordWrapText() method) , but I don't think I'm up for doing that in this pull request, because it seems like that would negate almost all the work I've done so far -- and moreover I failed to come up with an elegant way (that I would be happy with) to do it., whereas I'm quite happy with my implementation as is now.

If the above is a hard requirement, then probably someone else will have to do it. If it's not, then please could you review the recent state of the PR and point out the outstanding issues with it so that I could focus on resolving them?

engines/bladerunner/font.cpp
```diff
@@ -68,6 +68,33 @@ bool Font::open(const Common::String &fileName, int screenWidth, int screenHeigh
 		_characters[i].width = stream->readUint32LE();
 		_characters[i].height = stream->readUint32LE();
 		_characters[i].dataOffset = stream->readUint32LE();
+		// Commented out. Explicit fixes don't really make sense if we can override the TAHOMA18 font
```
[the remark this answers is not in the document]
antoniou79: I kept the code supporting the internal font(s) (basically the TAHOMA18 which is the only one that looks appropriate, but is also corrupted and needs the explicit fixed for kerning and spacing) as a fallback (if the Font is missing from the SUBTITLES.MIX for some reason) and also as a way for anyone to check if that looks better or worse than the provided external font. 
I guess I can drop the support for the internal font entirely.

engines/bladerunner/text_resource.h
```diff
@@ -23,6 +23,8 @@
 #ifndef BLADERUNNER_TEXT_RESOURCE_H
 #define BLADERUNNER_TEXT_RESOURCE_H
 
+#include "bladerunner/bladerunner.h" // needed for definition of Common::ScopedPtr (subtitles font external font file support)
```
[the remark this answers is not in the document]
antoniou79: I'll be removing it in the next commit/ push.

engines/bladerunner/font.h
```diff
@@ -23,6 +23,8 @@
...
```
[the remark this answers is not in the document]
antoniou79: I'll be removing it in the next commit/ push. Also the common/utils.h is no longer needed there.

engines/bladerunner/ui/kia_section_settings.h
```diff
@@ -23,6 +23,7 @@
 #ifndef BLADERUNNER_KIA_SECTION_SETTINGS_H
 #define BLADERUNNER_KIA_SECTION_SETTINGS_H
 
+#include "bladerunner/bladerunner.h" // to get the macro defines
```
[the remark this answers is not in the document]
antoniou79: I'll be removing it in the next commit/ push.

engines/bladerunner/font.cpp
```diff
@@ -192,6 +233,18 @@ void Font::drawCharacter(const uint8 character, Graphics::Surface &surface, int
 
 	int endY = height + y - 1;
 	int currentY = y;
+
+	// FIXME/TODO
```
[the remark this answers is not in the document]
antoniou79: Not really needed. But it's a safety check; it won't really affect the external (custom) font, and it will prevent crashes with the internal font (mostly when debugging stuff with that font I guess).

engines/bladerunner/subtitles.h
```diff
@@ -0,0 +1,107 @@
...
+	Common::String       _currentSubtitleTextFull;
+	Common::String _subtitleLineQuote[kMaxNumOfSubtitlesLines];
+	int _subtitleLineScreenY[kMaxNumOfSubtitlesLines];
+	int _subtitleLineScreenX[kMaxNumOfSubtitlesLines];
+	int _subtitleLineSplitAtCharIndex[kMaxNumOfSubtitlesLines];
+	int _currentSubtitleLines;
+	bool _subtitlesQuoteChanged;
+
+	bool _gameSubsResourceEntriesFound[kMaxTextResourceEntries];    // false if a TRE file did not open successfully
+	bool _subsFontsExternal;                                        // true if we are using the external font, false otherwise
```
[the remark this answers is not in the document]
antoniou79: Probably only as a fallback subtitles font. If we don't want it at all, I can remove it.

engines/bladerunner/font.cpp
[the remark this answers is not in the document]
antoniou79: I'll remove that entirely. Those were quick fixes, and I still wasn't very happy with the end-result, so I guess they can be scrapped.

engines/bladerunner/bladerunner.h
antoniou79: This macro (remnant of old code of mine) has been removed

engines/bladerunner/font.cpp
antoniou79: This entire method has been removed.

engines/bladerunner/font.cpp
antoniou79: Removed

engines/bladerunner/bladerunner.cpp
antoniou79: This was changed entirely and integration with ScummVM configuration manager has been implemented.

engines/bladerunner/subtitles.h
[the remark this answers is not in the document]
antoniou79: Internal font support (for subtitles) has now been removed

devtools/blade_runner/subtitles/module.mk
```diff
@@ -0,0 +1,2 @@
+
+MODULE := devtools/blade_runner/subtitles
```
sev-: It looks that the building rules are missing here. E.g. we would need to have to demonstrate how to build it.

sev-:
Great! Let's merge it. What I'd like to see eventually is the automated way to build the font and subtitle files.

PR #1221, opened 2018-06-20
MOHAWK: MYST: Some code refactoring related to mechanical age rotation

dafioram:
No behavior changes.

1. Made the direction fortress is set to to be an enum

### DIFF
--- a/engines/mohawk/myst_stacks/mechanical.h
+++ b/engines/mohawk/myst_stacks/mechanical.h
@@ -57,6 +57,13 @@ private:
 	void fortressRotation_run();
 	void fortressSimulation_run();
 
+	enum Direction {
+		kSouth = 0, // Starting Island with Myst linking book
+		kEast  = 1, // Island with right half of code
+		kNorth = 2, // Island with left half of code
+		kWest  = 3  // No island, just water
+	};
+
 	DECLARE_OPCODE(o_throneEnablePassage);
 	DECLARE_OPCODE(o_birdCrankStart);
 	DECLARE_OPCODE(o_birdCrankStop);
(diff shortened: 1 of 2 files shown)


engines/mohawk/myst_stacks/mechanical.h
```diff
@@ -57,6 +57,13 @@ class Mechanical : public MystScriptParser {
 	void fortressRotation_run();
 	void fortressSimulation_run();
 
+	enum Direction {
+		kSOUTH = 0, // Starting Island with Myst linking book
```
bonki: Constants should either be kCamelCase or ALL_UPPERCASE as per our CFCs.

bonki:
Interesting! And it is not possible that the opcode is used while `_elevatorRotationLeverMoving` is `false` (`elevatorRotation_run()` doesn't get called in this case)?

dafioram:
They are very nearly the same. 

The question is what should happen when the user has used the rotator controls to set the new direction, but has not locked onto the new direction and they hit the button to do the rotation? The current behavior and the behavior in the original engine is that the direction should be set to the nearest direction. o_fortressRotationSetPosition is what is setting the controls to the nearest direction. The behavior with this PR is that the last locked on direction is the direction the fortress rotates to (since the code from o_fortressRotationSetPosition was removed), rather than the nearest. 

For now I am just going to retain the original behavior so I have dropped the o_fortressRotationSetPosition changes.

So this PR just adds the enum for the direction.

dafioram:
Another question is what should happen if the user sets the controls to a new direction, but does not lock them in and instead of hitting the button to do the rotation they instead step away from the controls and then return, what should be the direction that the controls are now set to? The nearest or the last locked onto?

The current behavior is that the controls will be what they were last locked onto for the step away and return scenario. This seems inconsistent with the behavior when the rotate button is pressed. 

My previous commit for this PR (that I have now removed) had the behavior when pressing the rotate button to have it be the last locked on which would make it consistent with the stepping away and returning behavior.

bgK:
Regarding your question, the button is for making the elevator go down, it's not meant to apply the tower rotation. IMO, the behavior should be the same when leaving and returning, and when making the elevator go down.

Regarding the code, I have merged the PR manually, leaving the type for the fortress position to an integer. We do calculations with it, it does not feel right it being an enum. I've kept the constants as documentation.

dafioram:
I agree, I have opened PR1226 so that hitting the button does not change the direction (from the one set in  fortressRotation_run()).

That looks much cleaner, I didn't realize uint16 could be set to enum values without type conversion.

PR #1222, opened 2018-06-21
MOHAWK: RIVEN: Change Ytram catch video behavior

dafioram:
1. Don't allow video 20 to play (it is paired with video 15).
This is because 20 has a graphical glitch.

2. Make the video that gets played be random for each catch.
This makes more sense to me and is simplier.
The original had the first two catches not be random then
subsequent catches played 2 other catch videos randomly.

For testing consider setting the random wait time for bspit.cpp#L264 to be something shorter/ not random. You may want to hard code the mlstId to one of the videos (bspit.cpp#L340).


dafioram:
Here is the Video15/Video20 combination. https://streamable.com/sasxj

PR #1224, opened 2018-06-23
MOHAWK: MYST: Warn instead of error for missing resources

dafioram:
Some missing resources are not fatal and the caller can decide
what to do if the resource is missing.

Have MystSound::playEffect skip playing the sound if it
is not present (most missing sound effects should be non-fatal).

It would be preferable to identify missing resources and address
them, but for unknown missing resources this isn't possible so
warning instead of erroring is okay in most cases.


bgK:
I'm sorry, but I would rather not have this change. All the calling functions would need to be updated to check for the result, adding unneeded complexity.

PR #1225, opened 2018-06-24
IOS: Add Keyboard and UI Fixes for Smartkeyboard on IPad Pro.

digitall:
These have not been compiled or otherwise tested, but are cleaned up
version of code changes from forum user "dottostring" in thread:
http://forums.scummvm.org/viewtopic.php?t=14524

This should add Fake ESC key support replacing the '`' key and add full
support for the Arrow keys.

### DIFF
--- a/backends/platform/ios7/ios7_keyboard.mm
+++ b/backends/platform/ios7/ios7_keyboard.mm
@@ -46,6 +46,9 @@
 	[self setAutocorrectionType:UITextAutocorrectionTypeNo];
 	[self setAutocapitalizationType:UITextAutocapitalizationTypeNone];
 	[self setEnablesReturnKeyAutomatically:NO];
+	UITextInputAssistantItem* item = [self inputAssistantItem];
+	item.leadingBarButtonGroups = @[];
+	item.trailingBarButtonGroups = @[];
 
 	return self;
 }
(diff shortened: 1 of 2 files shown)


sev-:
Let's merge it and see.

bgK:
Looks like this broke the build.

digitall:
@bgK : Damn... I did say this was untested, though we are lacking a IOS maintainer currently so doing the best I can based on user code feedback in the forums. I will look over the buildbot build failure stderr and warnings and see if I can fix the build. If not, we may need to revert, or update the buildbot or code to fix :|

MWumpusZ:
There are other issues (for me) probably related to this PR; "Sometimes" scummvm stops reacting to the bluetooth keyboard.  I can reproduce it by pressing the "search" button on my bluetooth keyboard, the ios search thing comes up and can be used, but if i cancel (or otherwise later return to scummvm), the normal entry of text via the bluetooth keyboard no longer works.

Nothing I tried short of restarting scummvm completely made the keyboard work again :-/

Interestingly, the arrow keys (if the change from my PR is applied) do still work, which strongly indicates that its some internal state of scummvm breaking.  Possibly the "connection to the input field" - no idea if that makes sense.

I will investigate further in the next days.

Observed in KQ2 - an AGI game.  Others not yet checked :)

MWumpusZ:
And more issues: after this PR, I can‘t see any way of getting the virtual keyboard at all, even when I don‘t have an external keyboard connected.

I‘ll be looking at all of these things and none of the problems look like they are insurmountable, but at this point I‘m wondering if it might not be better to revert this PR and digitalls fix to it, and then redo them better?

MWumpusZ:
@bgK how can I see what broke in the build?  I realise digitall has fixed it since, but even before his fix, travis apparently ran through without issue (unless I'm misreading above), so evidently there is something additional performing build tests that I am not seeing?
Just trying to understand what all I can check myself before submitting possible PRs :-)

bgK:
Unfortunately travis does not build the iOS specific code. It just does Linux builds.
Our buildbot on the other hand compiles for a lot of platforms including iOS: http://buildbot.scummvm.org/. The build was broken there. However it only builds the code after it is merged into master.

digitall:
@bgK , @MWumpusZ : Normally you could have retrieved the error messages from the buildbot build logs, but the bot was reconfigured in a major way on 29th July i.e. day after the IOS build issues and the history restarts from there:
http://buildbot.scummvm.org/builders/master-ios?numbuilds=115

Sorry about that.

criezy:
I am guessing the issue was with `UITextInputAssistantItem` being only available in the iOS 9.0+ SDKs (see https://developer.apple.com/documentation/uikit/uitextinputassistantitem?language=objc) and buildbot using an older one (possibly 7.x).

digitall:
@MWumpusZ : Thanks for taking a look at this code and seeing about fixing it up. I am not an iOS expert and don't even own any Apple or iOS devices, so could not test this, but though it best to get the code into the tree and worked on. I tried to fix it up as far as possible where logical, but I am afraid I am not an expert here so can't offer more guidance / knowledge.

@criezy : Will check the buildbot iOS SDK versions and report back...

digitall:
Yes, SDK is 7.1 as per:
buildbot> ls /opt/toolchains/apple-ios/sdks
iPhoneOS4.2.sdk  iPhoneOS7.1.sdk

MWumpusZ:
@criezy @digitall Thanks for the information, it‘s all coming together now :-). I had already noticed the low buildbot build numbers and suspected that something major had happened recently.

Regarding this PR and it‘s consequences: I consider two things should be done in the short term; for one, the arrow key functionality should be made to work, and #1290 should do that.
Further, the behaviour when there is NO external keyboard has been messed up a bit by this PR, and I‘ll look into tidying that up next.
The whole UITextInputAssistant thing - currently commented out due to the SDK version thing - is a nice to have, but almost certainly less important feature.  I‘ll look at it, but it‘s lower priority as far as I‘m concerned.

PR #1226, opened 2018-06-24
MOHAWK: MYST: Improve Mechanical Fortress Rotator Direction State

dafioram:
1. Hitting the rotator button without locking on will not lock onto a new direction, but the previous direction locked onto will be used
2. Have fortress facing direction be the same when returning to the controls. Previously west direction was resetting to south.

### DIFF
--- a/engines/mohawk/myst_stacks/mechanical.cpp
+++ b/engines/mohawk/myst_stacks/mechanical.cpp
@@ -637,16 +637,7 @@ void Mechanical::o_elevatorTopMovie(uint16 var, const ArgumentsArray &args) {
 }
 
 void Mechanical::o_fortressRotationSetPosition(uint16 var, const ArgumentsArray &args) {
-	VideoEntryPtr gears = _fortressRotationGears->getVideo();
-	uint32 moviePosition = Audio::Timestamp(gears->getTime(), 600).totalNumberOfFrames();
-
-	// Myst ME short movie workaround, explained in o_fortressRotation_init
-	if (_fortressRotationShortMovieWorkaround) {
-		moviePosition += 3600 * _fortressRotationShortMovieCount;
-	}
-
-	_fortressDirection = (moviePosition + 900) / 1800 % 4;
-
+	// The fortress direction is already set in fortressRotation_run() so we don't do it here
 	// Stop the gears video so that it does not play while the elevator is going up
 	_fortressRotationGears->getVideo()->stop();
 }
--- a/engines/mohawk/myst_stacks/mechanical.cpp
+++ b/engines/mohawk/myst_stacks/mechanical.cpp
@@ -815,7 +815,11 @@ void Mechanical::o_fortressRotation_init(uint16 var, const ArgumentsArray &args)
 
 	VideoEntryPtr gears = _fortressRotationGears->playMovie();
 	gears->setLooping(true);
-	gears->seek(Audio::Timestamp(0, 1800 * _fortressDirection, 600));
+	if (!_fortressRotationShortMovieWorkaround) {
+		gears->seek(Audio::Timestamp(0, 1800 * _fortressDirection, 600));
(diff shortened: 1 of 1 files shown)


dafioram:
@bgK I think this PR is really worth while. The 2nd commit lines up the rotator behavior the same as the original Myst and is really a follow on to the workaround that was added to the code to be able to visit the north and west areas.

The first commit is a deviation from the originals, but I think a sensible one. If the user uses the right stick to turn the direction but doesn't lower the left stick then the fortress direction will remain the way it was pointing. The original has it change if the button pressed even when the left lever is not lowered.

I have also tested this a bunch and it is very solid. All the directions can still be visited.

engines/mohawk/myst_stacks/mechanical.cpp
```diff
@@ -846,8 +841,6 @@ void Mechanical::o_fortressRotation_init(uint16 var, const ArgumentsArray &args)
 	uint32 movieDuration = gears->getDuration().convertToFramerate(600).totalNumberOfFrames();
 	if (movieDuration == 3680) {
 		_fortressRotationShortMovieWorkaround = true;
-		_fortressRotationShortMovieCount = 0;
```
bgK: Is removing those two lines necessary? It would seem preferable to me if the puzzle was fully reset when accessing it.

dafioram: movieDuration == 3680 is only true when the fortress was last set to West. So what the setting of zero for these two variables does it causes the direction that the rotator is set to to be reset to the default/starting direction (South), but again this only happens when the direction was last West. 

So the behavior before this PR is that the Rotator direction will be the last location it was unless it was West then it will be South when the user steps back up to the rotator.

We can have the puzzle reset everytime the user returns to the puzzle, but that is not what these two lines of code (with the if check) does and it is a deviation from what regular Myst does. The West direction is currently behaving differently then all the other directions.

The only deviation that this PR makes is that if the user changes the direction, but does not lower the left stick (to lock on) and hits the button the direction the fortress is pointing will not be this new direction, were as the original would have it be this new direction.

bgK: > movieDuration == 3680 is only true when the fortress was last set to West.

Are you sure about that ? That test is meant to detect the broken video file that shipped with Myst ME. It should be true regardless of the position.

dafioram: Okay, sorry let me clarify.  

Mechanical::o_fortressRotation_init is called whenever the player steps back up to the rotator puzzle controls. For Myst ME the condition is true so _fortressRotationShortMovieCount = 0; this variable being set to zero causes the moviePosition value in Mechanical::fortressRotation_run() to map to South since the gears are not looping (since the player just stepped up to the controls). 

In this case the moviePosition should of been retained since the gears were not just turning, but since _fortressRotationShortMovieCount is set to zero this direction state is lost. I am not sure why this is not also an issue for north.

If you set the direction to west. step away from the controls and back to it the controls will be still set to west (can be checked by raising and lowering the left control stick) with this PR, but it will reset to south before this PR.

bgK: Ok. The catch with `_fortressRotationShortMovieCount` and `_fortressRotationShortMovieLast` is that they are not persisted in the saved state. So, if you save and restore, the same problem as before your fix should happen. Perhaps those variables need to be initialized in `o_fortressRotation_init` to values that will make `fortressRotation_run` work.

dafioram: The player always starts on the south island when loading a save in mech age so the direction always being set to south on init (and not stored in save state) makes sense.

bgK:
Thanks for fixing this! I've tweaked the code a bit until my OCD was satisfied before merging.

PR #1227, opened 2018-06-25
SDL: Clip mouse range in convertWindowToVirtual

Drenn1:
When the graphics scale was 2x or higher, it was possible for the mouse to go too far right or down. If the game's resolution was 320x200, it could range from 0-320 or 0-200, giving it a 321x201 range; it should really range from 0-319 or 0-199, which is what this does.

It looks like only the SDL backend uses this function.

### DIFF
--- a/backends/graphics/windowed.h
+++ b/backends/graphics/windowed.h
@@ -120,8 +120,11 @@ protected:
 		x = CLIP<int>(x, sourceX, sourceMaxX);
 		y = CLIP<int>(y, sourceY, sourceMaxY);
 
-		return Common::Point(((x - sourceX) * targetWidth + sourceWidth / 2) / sourceWidth,
-		                     ((y - sourceY) * targetHeight + sourceHeight / 2) / sourceHeight);
+		int virtualX = ((x - sourceX) * targetWidth + sourceWidth / 2) / sourceWidth;
+		int virtualY = ((y - sourceY) * targetHeight + sourceHeight / 2) / sourceHeight;
+
+		return Common::Point(CLIP<int>(virtualX, 0, targetWidth - 1),
+				             CLIP<int>(virtualY, 0, targetHeight - 1));
 	}
 
 	/**
--- a/backends/graphics/windowed.h
+++ b/backends/graphics/windowed.h
@@ -94,8 +94,11 @@ protected:
 			error("convertVirtualToWindow called without a valid draw rect");
 		}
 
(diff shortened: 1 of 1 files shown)


dafioram:
Seems like a simple change. From a user standpoint I don't notice any change, but I guess I wouldn't for a 1-pixel change.

backends/graphics/windowed.h
```diff
@@ -120,8 +120,11 @@ class WindowedGraphicsManager : virtual public GraphicsManager {
 		x = CLIP<int>(x, sourceX, sourceMaxX);
 		y = CLIP<int>(y, sourceY, sourceMaxY);
 
-		return Common::Point(((x - sourceX) * targetWidth + sourceWidth / 2) / sourceWidth,
-		                     ((y - sourceY) * targetHeight + sourceHeight / 2) / sourceHeight);
+		int virtualX = ((x - sourceX) * targetWidth + sourceWidth / 2) / sourceWidth;
+		int virtualY = ((y - sourceY) * targetHeight + sourceHeight / 2) / sourceHeight;
+
+		return Common::Point(CLIP<int>(virtualX, 0, targetWidth - 1),
+				             CLIP<int>(virtualY, 0, targetHeight - 1));
```
criezy: The indentation is a bit off die to the use of two additional tabs (after the first two tabs) instead of spaces.

Drenn1:
OK, I made the requested changes.

backends/graphics/windowed.h
```diff
@@ -94,8 +94,11 @@ class WindowedGraphicsManager : virtual public GraphicsManager {
 			error("convertVirtualToWindow called without a valid draw rect");
 		}
 
-		return Common::Point(targetX + (x * targetWidth + sourceWidth / 2) / sourceWidth,
-		                     targetY + (y * targetHeight + sourceHeight / 2) / sourceHeight);
+		int windowX = targetX + (x * targetWidth + sourceWidth / 2) / sourceWidth;
+		int windowY = targetY + (y * targetHeight + sourceHeight / 2) / sourceHeight;
+
+		return Common::Point(CLIP<int>(windowX, 0, targetWidth - 1),
```
criezy: There is a catch here: it should be clipped between `targetX` and `targetX + targetWidth - 1` ;-)
And the same thing on the next line as well.

criezy:
Thank you. Can you please squash the last two commits? After that it will be good to merge.

PR #1228, opened 2018-06-25
SCI: Use LarryScale in games that support it, not just LSL7

eriktorbjorn:
This is the solution discussed in bug #10568 ("SCI: PHANT1: Crash on startup"). A temporary-maybe-permanent fix was committed recently (fab1894f21, "ISCI32: Only enable larry scaler for LSL7") but I believe that this is the correct solution, and that ScummVM will do the right thing... but I'm not 100% sure.

The idea is that since games can declare that they will use an engine-specific setting (in this case the Larry scaler), it's better to test for that than to just assume that only LSL7 will ever use it.

### DIFF
--- a/engines/sci/graphics/celobj32.cpp
+++ b/engines/sci/graphics/celobj32.cpp
@@ -33,6 +33,7 @@
 #include "sci/util.h"
 #include "graphics/larryScale.h"
 #include "common/config-manager.h"
+#include "common/gui_options.h"
 
 namespace Sci {
 #pragma mark CelScaler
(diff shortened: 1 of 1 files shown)


criezy:
Thank you for creating this PR.
I also think this is the correct way to fix the issue from bug #10568.

DanielSWolf:
👍

bonki:
It certainly looks as if this is the correct way to fix this.

While we're at it, is there a particular reason why `GAMEOPTION_LARRYSCALE` currently isn't enabled for the demo (apologies if this has been mentioned before)?

criezy:
Thank you. Merging.
I have also added the flag to the LSL7 demo to my working copy and will test a bit before pushing.

PR #1235, opened 2018-06-30
ZVISION: Add workaround for ZGI bug #10604

eriktorbjorn:
This is one possible way of working around ZGI bug #10604. It was possible to render the game unplayable simply by looking at Jack's cigar box while waiting for him to return to examine the lamp. This makes sure the pp1f_video_flag is properly reset in that case as well.

Though doing it this way will do nothing to repair old savegames. Only prevent the bug from happening again. On ther other hand, this is at the start of the game so it's not that much of a burden, I think.


dafioram:
s/assignmetn/assignment

dafioram:
Since your second commit is a small one and you haven't pushed to master you could combine the 2 commits together with `git rebase HEAD~2 -i` and then do a force push.

dafioram:
It commit adds a potential blocker to the CD version.

 If you take the steps to trigger the bug with the CD version on this commit then the game stops right before playing the video that is supposed to play after you are holding the cigar.

eriktorbjorn:
I've restricted the workaround to the DVD version. Hopefully both versions should work fine now.

eriktorbjorn:
One worrying thought that just struck me... It would be good if someone could check if the bug was introduced with the DVD version, or if it was introduced with the 1.2 patch for the CD version.

dafioram:
How do I tell if my CD version is patched? The version I have is the old GOG version before they changed to the DVD.

eriktorbjorn:
> How do I tell if my CD version is patched?

Unfortunately, I don't know. I wouldn't even know about the patch if it wasn't mention on the [the ScummVM Wiki](http://wiki.scummvm.org/index.php/Datafiles#Zork:_Grand_Inquisitor).

dafioram:
The CD version that used to ship with GOG did have the patch applied. 

When running ZGI through the original interpreter the version can be checked in game with "ctrl+v". This also means I cannot check the behavior of ZGI CD 1.0 since my files are already patched.

One of the fixes said to be patched in the 1.2 notes is "Fixes the Card Pick Up Bug in the Port Foozle Bar".

dafioram:
@eriktorbjorn you can probably squash these two commits together since it is just one fix. It is still useful to have the commit contains the details of the two separate commits.

dafioram:
As expected I can't reproduce this bug in the original engine for CD 1.2. Not sure about DVD version.

sev-:
OK, merging. Will look for users feedback.

PR #1236, opened 2018-06-30
ZVISION: Add workaround for ZGI bug #10605

eriktorbjorn:
This is one possible way of working around ZGI bug #10605. It was possible for the "Fire! Fire!" timer to get stuck if you brought up the inventory screen before it had timed out. (Unless you brought it up _very_ quickly so that there was still more than 10 units left on the clock.)

I have a sneaking suspicion that there may be other timer-related script bugs, but I haven't looked for them. See the bug report for some further thoughts.

### DIFF
--- a/engines/zvision/scripting/scr_file_handling.cpp
+++ b/engines/zvision/scripting/scr_file_handling.cpp
@@ -205,6 +205,28 @@ bool ScriptManager::parseCriteria(Common::SeekableReadStream &stream, Common::Li
 			entry.argumentIsAKey = false;
 		}
 
+		// WORKAROUND for a script bug in Zork: Grand Inquisitor. If the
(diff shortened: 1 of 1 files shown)


criezy:
The change looks fine to me.

eriktorbjorn:
I should mention that I haven't been able to get ZGI to run under Wine (I know it should be possible, but it didn't work for me), so I haven't been able to verify that it also happens in the original.

If it doesn't happen in the original, then it's probably a bug in how ScummVM handles ZVision's timers.

dafioram:
I have verified that this bug does happen in the original engine for CD 1.2. 

My DVD version is from GOG and runs on ScummVM so I can't check the DVDs behavior, but I think it's safe to assume that it is the same.

sev-:
Thanks, merging

PR #1239, opened 2018-07-04
CONFIGURE: Disable -Wpragma-pack for >=clang-6.0

bonki:
Clang 6 enables `-Wpragma-pack` which warns when leaving an included file which changes the current alignment.
Our `common/pack-{start,end}.h` trigger this and it cannot easily and portably be disabled inside `pack-{start,end}.h` so we disable it globally for now.

### DIFF
--- a/configure
+++ b/configure
@@ -1914,6 +1914,14 @@ fi
 
 if test "$have_clang" = yes; then
 	add_line_to_config_mk 'HAVE_CLANG = 1'
+	_clang_major=`gcc_get_define __clang_major__`
+	#_clang_minor=`gcc_get_define __clang_minor__`
+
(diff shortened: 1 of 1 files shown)


bonki:
Clarification:

It would be preferable to strictly disable the warning in `common/pack-start.h` and re-enable it again in `common/pack-end.h` using `#pragma clang diagnostic push/pop` and `#pragma clang diagnostic ignore "-Wpragma-pack"`. However, popping from within the same include re-enables the warning before leaving the file, resulting in warnings again. This means that we would have to add these before/after every use of `common/pack-{start,end}.h`, which isn't feasible.

We could also define macros doing everything and use those but you cannot use `#pragma` in a macro. There is `_Pragma()` but that doesn't seem to be portable.

PR #1240, opened 2018-07-06
TINSEL: Show saved game creation time in load/save gui

dafioram:
Saved games inspected via the ScummVM load or save
gui will now show the year, month, day, hour, and minute
of its creation. This was already being saved in the
saved game header so no version bump is necessary.

This required adding kSavesSupportMetaInfo and
kSavesSupportCreationDate features. I also had to
add kSavesSupportThumbnail or the saved date is not
shown.

It was necessary to write querySaveMetaInfos.

### DIFF
--- a/engines/tinsel/detection.cpp
+++ b/engines/tinsel/detection.cpp
@@ -102,6 +102,7 @@ public:
 	virtual bool hasFeature(MetaEngineFeature f) const;
 	virtual SaveStateList listSaves(const char *target) const;
 	virtual int getMaximumSaveSlot() const;
+	virtual SaveStateDescriptor querySaveMetaInfos(const char *target, int slot) const;
 	virtual void removeSaveState(const char *target, int slot) const;
 };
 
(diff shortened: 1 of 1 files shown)


engines/tinsel/detection.cpp
```diff
@@ -110,7 +111,10 @@ bool TinselMetaEngine::hasFeature(MetaEngineFeature f) const {
 		(f == kSupportsListSaves) ||
 		(f == kSupportsLoadingDuringStartup) ||
 		(f == kSupportsDeleteSave) ||
-		(f == kSimpleSavesNames);
+		(f == kSimpleSavesNames) ||
+		(f == kSavesSupportMetaInfo) ||
+		(f == kSavesSupportThumbnail) ||
```
dreammaster: Was kSavesSupportThumbnail incorrectly added, or is that something you still want to experiment with?

dafioram: If I remove that flag then the SaveDate info does not get shown.

dreammaster: Huh. You're right. I'll merge this in then.

PR #1242, opened 2018-07-07
GUI: Show saveload date/time, or playtime even if thumbnail not enabled

dafioram:
Previously, if playtime or date/time support was added it would not be displayed
unless thumbnail support was also valid.

Now they will be shown even if the thumbnail is not valid.

No container is shown if only metainfo support is enabled.

Removed thumbnail support from tinsel.

The behavior is the same as before if thumbnail support is enabled.

Currently, Tinsel supports date time, but not playtime nor thumbnail, but it must enable thumbnail to show the date time and it looks like this:
scummvmdatetimeprev

After this PR it will look like this:
scummvmdatetime

And if playtime support is added:
scummvmdatetimeplay

For this PR I have kept the container top to be level with the save list.

### DIFF
--- a/gui/saveload-dialog.cpp
+++ b/gui/saveload-dialog.cpp
@@ -463,13 +463,14 @@ void SaveLoadChooserSimple::handleCommand(CommandSender *sender, uint32 cmd, uin
 }
 
 void SaveLoadChooserSimple::reflowLayout() {
-	if (g_gui.xmlEval()->getVar("Globals.SaveLoadChooser.ExtInfo.Visible") == 1 && _thumbnailSupport) {
+	if (g_gui.xmlEval()->getVar("Globals.SaveLoadChooser.ExtInfo.Visible") == 1 && (_thumbnailSupport || _saveDateSupport || _playTimeSupport)) {
 		int16 x, y;
 		uint16 w, h;
 
 		if (!g_gui.xmlEval()->getWidgetData("SaveLoadChooser.Thumbnail", x, y, w, h))
 			error("Error when loading position data for Save/Load Thumbnails");
 
+		// Even if there is no thumbnail support, getWidgetData() will provide default thumbnail values
 		int thumbW = kThumbnailWidth;
 		int thumbH = kThumbnailHeight2;
 		int thumbX = x + (w >> 1) - (thumbW >> 1);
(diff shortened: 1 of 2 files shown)


gui/saveload-dialog.cpp
```diff
@@ -463,17 +463,22 @@ void SaveLoadChooserSimple::handleCommand(CommandSender *sender, uint32 cmd, uin
...
 
 void SaveLoadChooserSimple::reflowLayout() {
-	if (g_gui.xmlEval()->getVar("Globals.SaveLoadChooser.ExtInfo.Visible") == 1 && _thumbnailSupport) {
+	if (g_gui.xmlEval()->getVar("Globals.SaveLoadChooser.ExtInfo.Visible") == 1 && (_thumbnailSupport || _saveDateSupport || _playTimeSupport)) {
 		int16 x, y;
 		uint16 w, h;
+		int thumbW;
+		int thumbH;
+		int thumbX;
+		int thumbY;
```
criezy: Why did you change the place where those variables are declared?

gui/saveload-dialog.cpp
```diff
@@ -463,17 +463,22 @@ void SaveLoadChooserSimple::handleCommand(CommandSender *sender, uint32 cmd, uin
...
 		int16 x, y;
 		uint16 w, h;
+		int thumbW;
+		int thumbH;
+		int thumbX;
+		int thumbY;		
 
 		if (!g_gui.xmlEval()->getWidgetData("SaveLoadChooser.Thumbnail", x, y, w, h))
-			error("Error when loading position data for Save/Load Thumbnails");
+				error("Error when loading position data for Save/Load Thumbnails");
```
criezy: Why the change of indentation here? It looks like the original one was correct?

gui/saveload-dialog.cpp
```diff
@@ -482,29 +487,37 @@ void SaveLoadChooserSimple::reflowLayout() {
...
 
 		if (_saveDateSupport) {
-			_date->resize(thumbX, height, kThumbnailWidth, kLineHeight);
+			_date->resize(thumbX, height, thumbW, kLineHeight);
 			height += kLineHeight;
-			_time->resize(thumbX, height, kThumbnailWidth, kLineHeight);
+			_time->resize(thumbX, height, thumbW, kLineHeight);
 			height += kLineHeight;
+			_date->setVisible(_saveDateSupport);
+			_time->setVisible(_saveDateSupport);
```
criezy: By moving those inside the if block aren't you now missing a setVisible(false) when _saveDateSupport is false?

gui/saveload-dialog.cpp
```diff
@@ -482,29 +487,37 @@ void SaveLoadChooserSimple::reflowLayout() {
...
 		}
 
-		if (_playTimeSupport)
-			_playtime->resize(thumbX, height, kThumbnailWidth, kLineHeight);
+		if (_playTimeSupport) {
+			_playtime->resize(thumbX, height, thumbW, kLineHeight);
+			_playtime->setVisible(_playTimeSupport);
```
criezy: Same comment here: are you now now missing a setVisible(false) when _playTimeSupport is false?

dafioram:
Updated.

gui/saveload-dialog.cpp
```diff
@@ -483,28 +484,42 @@ void SaveLoadChooserSimple::reflowLayout() {
 		if (textLines > 0)
 			textLines++; // add a line of padding at the bottom
 
-		_container->resize(x, y, w, h + (kLineHeight * textLines));
-		_gfxWidget->resize(thumbX, thumbY, thumbW, thumbH);
+		if (_thumbnailSupport) {
+			_gfxWidget->resize(thumbX, thumbY, thumbW, thumbH);	
+			_gfxWidget->setVisible(true);
+		}
+		else { // choose sensible values for displaying playtime and date/time when a thumbnail is not being used
```
criezy: Our [code formatting conventions](http://wiki.scummvm.org/index.php/Code_Formatting_Conventions#Hugging_braces) indicate the else should not be on a new line (and same thing for the other two else below.

dafioram:
Updated.

PR #1243, opened 2018-07-08
NEWS: Add MOHAWK changes

dafioram:
There are more changes than this, but these are the more noticeable changes.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -19,9 +19,14 @@ For a more comprehensive changelog of the latest experimental code, see:
    - Fixed a bug that leads to enormous memory consumption in scene 22.
 
  MOHAWK:
-   - Fixed clicking through the books in the library in Myst 1.0.
-   - Fixed video playback if videos are skipped in Riven.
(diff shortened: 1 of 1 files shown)


sev-:
Are we going to enlist every bugfix in the NEWS file? We never did it before, and it is quite verbose.

I would write just something like: "fixed numerous bugs and general stability improvements", then I would describe "support for 25th Anniversary rereleases".

criezy:
I agree with @sev- 
I think we might want to be more selective in what is listed in the NEWS file or group multiple items together so that important changes are not drown under a big number of changes.

Kroc:
If a particular crash was stopping a player from completing the game, they may be very happy to be notified of they're particular pain point.

bgK:
> If a particular crash was stopping a player from completing the game, they may be very happy to be notified of their particular pain-point. By all means merge some items, but spell out show-stoppers directly.

There were no progress blocking bugs.

Thanks for updating the news file! I've added the new features of the 25th anniversary edition and reworded some of the news items before merging.

PR #1245, opened 2018-07-09
TINSEL: Add playTime to saved game and display it

dafioram:
Also bumps the saved game format version to 3.

I changed the order of how the data is saved so that
the playtime could be read in without skipping as much data.

This is because querySaveMetaInfos only needs the metafields
where as saveload::DoRestore() needs more of the header.

A crash was happening when trying to delete a saved game from the
launcher. It is because the engine is not initalized. I assign a
dummy value to resolve the issue.

Loading saved games from previous versions works.

When an old version saved game is loaded it will start with
zero playtime.

Old saved games are shown as not having playtime data.


sev-:
Good stuff.

dafioram:
Even though this has been merged it would probably still be good for it to be glanced at by @dreammaster.

PR #1248, opened 2018-07-10
XEEN: fix crashes in "Clouds of Xeen" while resting

Henne:
When dreamsequences are triggerd in "Clouds of Xeen", some files are missing
("scene1.raw", "dreams2.voc") and the engine exits.

If there are dreamsequences in "Clouds of Xeen", then this fix is not the right thing.

### DIFF
--- a/engines/xeen/interface.cpp
+++ b/engines/xeen/interface.cpp
@@ -1056,7 +1056,7 @@ void Interface::rest() {
 			party.changeTime(map._isOutdoors ? 380 : 470);
 		}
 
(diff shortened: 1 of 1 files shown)


dreammaster:
It'll actually make more sense to put the if check in the engine dream method, which is already virtual and has a different implementation for Clouds/Darkside vs. Swords. Originally Clouds & Dark Side on their own were also going to get separate engine subclasses, and only Dark Side would have a dream implementation. But it proved too difficult to implement World, which would be using both classes functionality.

I'll close this pull request and commit in a check in dream accordingly.

PR #1249, opened 2018-07-12
LAB: Fix thumbnail issues

dafioram:
Fixes Trac#10619.

The thumbnail loading for the saved games was defaulting to
disabled so no thumbnail was being loaded and this caused trying
to show the thumbnail to crash scummvm.

So I have set the thumbnail pointer to be null if the thumbnail
doesn't get loaded and I've set the thumnail to not be skipped
so that thumbnails get shown when using the gui.

### DIFF
--- a/engines/lab/lab.h
+++ b/engines/lab/lab.h
@@ -502,7 +502,7 @@ private:
 	void handleTrialWarning();
 };
 
-WARN_UNUSED_RESULT bool readSaveGameHeader(Common::InSaveFile *in, SaveGameHeader &header, bool skipThumbnail = true);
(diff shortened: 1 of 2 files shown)


sev-:
Thanks! Merging

digitall:
@dafioram : I have a copy of Cruise For A Corpse and have tested these changes with that engine. Thumbnail loading worked fine without changing the default value on the function call, but I have added the nullptr assignment as 2a4fa26d5241f7febbb47e4419e6ca1c3f6925d9, partly as good practice in any case.

dafioram:
Thanks for checking.

PR #1250, opened 2018-07-13
LURE: Don't report failed to save/load when it is sucessful

dafioram:
This required switching the ternary true and false operands.

### DIFF
--- a/engines/lure/lure.h
+++ b/engines/lure/lure.h
@@ -121,11 +121,11 @@ public:
 	bool isEGA() const { return (getFeatures() & GF_EGA) != 0; }
 
 	virtual Common::Error loadGameState(int slot) {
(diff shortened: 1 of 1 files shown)


tsoliman:
This has been broken for 9+ years?

dafioram:
It doesn't actual crash the game just issues an annoy GUI message that says the read failed. The GUI message can be closed so not major.

tsoliman:
I deleted my original message that said "This has been broken for 9+ years" (as it is not constructive), but then realized you already replied to it. So here it is for the record :)

tsoliman:
Yes I just tested, ScummVM makes me think the save failed but it actually works. I was able to load it with your patch. Thanks for this.

PR #1251, opened 2018-07-13
PINK: Set engine playtime to playtime in saved game

dafioram:
1. Add playtime in msecs to SaveStateDescriptor (SaveStateDescriptor is in the Graphics namespace)
2. Use first commit to have pink load playtime from saved game into engine.


sev-:
Thanks!

PR #1252, opened 2018-07-14
GROOVIE: Allow saving via GMM and increase save slots to 25

dafioram:
If the user enters in an invalid description for the save then it will be
simplified by the saved name cache process in Script::savegame.

I checked this with t7g, but not with the 11th guest or any other Groovie game.

The original in-game loading/saving can still only access the first 10 save
slots (0-9).

Its possible to saving via the GMM when the game is interactive.

Loading a game saved via the gui using the original in-game load menu
works fine.

Saving via the GMM with timestamp works, but only only saves the characters
up to the first non-alpha non-letter characer.


sev-:
That's a great change. And the 11th hour is not supported at this moment.

PR #1256, opened 2018-07-20
AGOS: Add in game loading/Saving via GMM

dafioram:
Once the engine has been started and the game is at a safe place to
save/load (as determined by quickLoadOrSave()) the player can save/load
their game using the GMM.

Currently, the save descriptor shown in the listsaves is the filename,
but the actual correct descriptor is correct in the save when saving via
the GMM. The descriptor saved using the GMM can be seen using the original
game save/load menus. The displaying of the correct  descriptor can be added
later.

The filenames shown by listsaves is generated from the input target
which can be one of several games. Most of them are just the basename so
the target simon2-cd-win is just simon2.xxx for the savefile, but for
waxworks and elvira2 they differ on whether the platform is dos
or amiga. Since the target does not contain this information I am
currently not differentiating the two, so I defaulted it to generate
names based on the dos version. This can also be changed later. For
amiga waxworks and elvira2 this has the effect that saved games cannot
be loaded/seen using listsaves. They will be correctly saved when saving via
the GMM since that does not use the savefile name generated from the target.

I've tested this with waxworks dos, simon1 dos and simon2 dos.

### DIFF
--- a/engines/agos/agos.h
+++ b/engines/agos/agos.h
@@ -216,6 +216,10 @@ protected:
 	virtual bool hasFeature(EngineFeature f) const;
 	virtual void syncSoundSettings();
 	virtual void pauseEngineIntern(bool pause);
+	virtual bool canLoadGameStateCurrently() { return true; };
+	virtual bool canSaveGameStateCurrently() { return true; };
+	virtual Common::Error loadGameState(int slot);
+	virtual Common::Error saveGameState(int slot, const Common::String &desc);
 
 	virtual void setupOpcodes();
 	uint16 _numOpcodes, _opcode;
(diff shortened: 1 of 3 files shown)


Kirben:
The current quick load and save code in the AGOS engine is a hack at best, and I still don't consider it a safe method to use.  It was only added long ago for console ports, lacking the keyboard.

We currently just copy various script segments and hope for the best when loading and saving. I'm rejecting this pull request, as these changes are just too risky.

Kirben:
If anyone is interested in looking further at GMM support for AGOS game engine, the quick load and save code must be perfected first. Specifically by examining the load the save segments of scripts in each game, to work out exactly what each line is actually doing. Then mirror those changes exactly (with comments) in the AGOS game engine, to perfect the quick loading and saving code.

PR #1257, opened 2018-07-21
MOHAWK: RIVEN: Don't allow autosaving during credits

dafioram:
Fixes Trac#10635.

### DIFF
--- a/engines/mohawk/riven.cpp
+++ b/engines/mohawk/riven.cpp
@@ -737,7 +737,7 @@ void MohawkEngine_Riven::saveGameStateAndDisplayError(int slot, const Common::St
 }
 
 void MohawkEngine_Riven::tryAutoSaving() {
-	if (!canSaveGameStateCurrently() || !isGameStarted()) {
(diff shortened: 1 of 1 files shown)


bgK:
Maybe we can disable saving when `_gameEnded` is true instead to avoid duplicating the state.

dafioram:
setGameEnded is called at the end of RivenStack::runCredits so it can be called earlier, now the while loop that plays the credits won't play the credits because of the endgame check so we could remove that check.

The endgame setting to true looks like it only happens in a few places in the code only associated with the demo (besides runCredits).

MohawkEngine::run also checks it to keep running, but I think that won't be checked until the credits loop is done.

bgK:
Saving is already disabled during the credits because they are started by a script. It's just for a frame after the credits that autosaving has a chance to trigger. During that frame, `_gameEnded` is true.

dafioram:
Let me see what happens if I use _gameEnded instead...

dafioram:
Yeah that worked.

bgK:
Thanks !

PR #1259, opened 2018-07-24
MOHAWK: RIVEN: Only use x mouse position to move dome sliders

dafioram:
Fixes Trac#10642.

The original engine will move the dome sliders whenever the player is
dragging a dome slider to the left or right regardless of y position.

In ScummVM the dome slider position would only change to the players x
mouse position when the y value was also in the slider hotspot. This
change removes the y check by making the point to be checked always
have a y value in the hotspot rect.

### DIFF
--- a/engines/mohawk/riven_stacks/domespit.cpp
+++ b/engines/mohawk/riven_stacks/domespit.cpp
@@ -115,7 +115,7 @@ void DomeSpit::checkDomeSliders() {
 
 void DomeSpit::checkSliderCursorChange(uint16 startHotspot) {
 	// Set the cursor based on _sliderState and what hotspot we're over
-	int16 sliderSlot = getSliderSlotAtPos(startHotspot, getMousePosition());
+	int16 sliderSlot = getSliderSlotClosestToPos(startHotspot, getMousePosition());
 
 	if (sliderSlot >= 0 && isSliderAtSlot(sliderSlot)) {
 		_vm->_cursor->setCursor(kRivenOpenHandCursor);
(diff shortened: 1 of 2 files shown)


macca8:
I understand why you're doing this, but just a few thoughts...

The closed hand cursor implies that something is being held and can be manipulated.

Adopting the original engine's behaviour allows for an unnatural separation between this cursor and the slider it's supposed to be dragging, which isn't a good look.

On the other hand, the sight of this cursor wandering around without moving the slider isn't much better.

As an alternative, perhaps we could create a more natural look by changing the appearance of the cursor to match the situation.

Using the fix by bgK for #10525 as a guide,  why not force the mouse up event when the cursor leaves the area?

This changes the cursor from the closed hand view to the neutral view, which should make sense to the player as to why the slider isn't moving.

The player will need to grab the slider again, but also receives a clear message to stay within the slider panel's area, which isn't a hard thing to do.

This is only a suggestion. At the end of the day, it's your PR and your call.

dafioram:
For now I just want the behavior to line up with the original. 

My cursor has gone out of the slider area many times while dragging the slider in ScummVM and has lead me to think that ScummVM is slow or something, since the slider will get stuck. I think it's easy to do this when moving the sliders fast.

The closed hand cursor can mean that the slider is still being held onto and dragged even when outside of the area and no longer over the slider (until the user let's go).

dafioram:
The slider could still get stuck if the user went to the left or right of the area outside the slider view so I now also clip the x values to be in that area. 

Changing the slider should now feel very fluid now matter where the users cursor goes (as long as they first grab the slider and hold onto it).

dafioram:
@bgK I think users will like how much this improves the usability of the sliders.

dafioram:
I went through all the domes and found suitable x bounds and hard coded those so that the x bounds don't have to be searched for each time the user moves the sliders. The domes either had the minimum x values hotspot of the slider area be 210 or 211 and the max was either 407 or 408.

I checked the slider for each dome and they all work with the hard coded values.

engines/mohawk/riven_stacks/domespit.cpp
```diff
@@ -125,9 +125,25 @@ void DomeSpit::checkSliderCursorChange(uint16 startHotspot) {
...
 
 int16 DomeSpit::getSliderSlotAtPos(uint16 startHotspot, const Common::Point &pos) const {
+	// Emperically found min x and max x hotspot are used to bound mouse position into
+	// the slider area vertically. This and the y mouse position being directly put into
+	// the slider area allows the user to move the cursor out of the slider area and still
+	// be able to move the slider.
+	int16 minXHotspot = 211; // suitable min x value hotspot for all domes
+	int16 maxXHotspot = 407; // suitable max x value hotspot for all domes
+
+	// Now got find out which slider is being moved
```
bgK: Please fix / remove that comment. This function gets the slider slot closest to `pos`, which is not necessarily the slider being moved.

engines/mohawk/riven_stacks/domespit.cpp
```diff
@@ -125,9 +125,25 @@ void DomeSpit::checkSliderCursorChange(uint16 startHotspot) {
 }
 
 int16 DomeSpit::getSliderSlotAtPos(uint16 startHotspot, const Common::Point &pos) const {
```
bgK: Please rename this function to match its new role. `getSliderSlotClosestToPos` or something.

dafioram:
Done.

bgK:
Thanks!

dafioram:
Thanks bgK!

@macca8 feel free to check out the new sliders behavior, and see if there are any issues, but please compare it to the behavior of the original game since that is what players will be expecting.

PR #1260, opened 2018-07-26
BLADERUNNER: Add cos/sine table

dafioram:
This uses the one in Common.

The engine now contains a 10-bit cosine and sine table.
It used mostly for vector math.

This also allows two vector functions to be removed from vector.h.


sev-:
Thanks!

PR #1261, opened 2018-07-26
GRAPHICS: Set thumbnail to nullptr when loading the thumbnail is skipped

dafioram:
Since Graphics::loadThumbnail can return successfully when a thumbnail is skipped
the caller may want to check to the see that the thumbnail is not null before using
it. Setting it to null is just defensive in case the caller didn't do so.


dafioram:
I've changed the LAB default readSaveGameHeader behavior back to skipping the thumbnail (except for queryMeta) since thats how I think bonki intended it as a part of 00e59a3.

sev-:
Thank you, merging.

PR #1263, opened 2018-07-28
STARTREK: New engine (WIP)

Drenn1:
This adds support for Star Trek: 25th anniversary (only DOS version has been tested so far).

The first 6 of 7 away missions are playable. The sections on the enterprise with pseudo-3d ship combat are not yet implemented, so it currently jumps directly from one away mission to the next. Players will be missing some pre-mission context until this is implemented.


engines/startrek/common.h
```diff
@@ -0,0 +1,51 @@
...
+#include "common/textconsole.h"
+
+namespace Common {
+struct Rect;
+class Serializer;
+}
+
+namespace StarTrek {
+
+template<class T>
```
sev-: We have those, they're called MIN() and MAX()

engines/startrek/detection.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+
+namespace StarTrek {
+
+static const StarTrekGameDescription gameDescriptions[] = {
+	{
+		{
+			"st25",
+			"",
+			AD_ENTRY1("data.001", "57040928a0f374281aa86ba4e7db8444"),
```
sev-: I recommend to always add file size. Please switch to AD_ENTRY1s()

engines/startrek/detection.cpp
```diff
@@ -0,0 +1,381 @@
...
+#include "startrek/startrek.h"
+
+namespace StarTrek {
+
+struct StarTrekGameDescription {
+	ADGameDescription desc;
+
+	uint8 gameType;
+	uint32 features;
+	uint16 version;
```
sev-: Are you going to use this? It is set to 0 everywhere.

Drenn1:
Thanks for the feedback. I've replaced AD_ENTRY1 with AD_ENTRY1s for those versions that I have. Most were added by clone2727 a decade ago! I'll update the other versions as I add support for them.

sev-:
Then I recommend to remove/comment out those entries, so the users will be able to submit them. Also, I'll take a look at my archives in couple of weeks, perhaps I will be able to provide the file sizes.

engines/startrek/graphics.cpp
```diff
@@ -0,0 +1,759 @@
...
+	if (ConfMan.hasKey("render_mode"))
+		_egaMode = (Common::parseRenderMode(ConfMan.get("render_mode").c_str()) == Common::kRenderEGA) && (_vm->getGameType() != GType_STJR) && !(_vm->getFeatures() & GF_DEMO);
+
+	if (_vm->getGameType() == GType_ST25 && _vm->getPlatform() == Common::kPlatformDOS)
+		_font = new Font(_vm);
+
+	_numSprites = 0;
+	_pushedNumSprites = -1;
+
+	_palData = new byte[256 * 3];
```
sev-: I think, here we have a memory leak. You're not freeing it anywhere.

engines/startrek/graphics.cpp
```diff
@@ -0,0 +1,759 @@
...
+	return _font->getCharData(c & 0xff);
+}
+
+
+void Graphics::copyBackgroundScreen() {
+	drawDirectToScreen(_backgroundImage);
+}
+
+void Graphics::drawDirectToScreen(SharedPtr<Bitmap> bitmap) {
+	int xoffset = bitmap->xoffset;
```
sev-: What is the point of these variables, when you're doing other references to bitmap variables in the same call below?

engines/startrek/items.h
```diff
@@ -0,0 +1,260 @@
...
+	{  false,  0x00,  "",          0x00  },
+	{  false,  0x00,  "",          0x00  },
+	{  false,  0x00,  "",          0x00  },
+	{  false,  0x00,  "",          0x00  },
+	{  false,  0x00,  "",          0x00  },
+	{  false,  0x00,  "",          0x00  },
+	{  false,  0x00,  "",          0x00  }
+};
+
+const int NUM_ITEMS = sizeof(g_itemList) / sizeof(struct Item) - 64; // 0x49
```
sev-: There is ARRAYSIZE() macro for calculating that.

engines/startrek/iwfile.cpp
```diff
@@ -0,0 +1,78 @@
+
```
sev-: Extra line at beginning of the file.

engines/startrek/iwfile.h
```diff
@@ -0,0 +1,65 @@
+
```
sev-: Another extra line

engines/startrek/object.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "object.h"
+
+namespace StarTrek {
```
sev-: Why keep an empty file?

engines/startrek/object.h
```diff
@@ -0,0 +1,186 @@
...
+	char animationString[10];
+
+	// These might be part of "animationString"?
+	uint16 fielda2;
+	uint16 fielda4;
+	uint16 fielda6;
+
+public:
+	Actor() :
+		spriteDrawn(),
```
sev-: Shouldn't a proper initializarion go here?

engines/startrek/room.cpp
```diff
@@ -0,0 +1,539 @@
...
+#include "startrek/startrek.h"
+
+#include "rooms/function_map.h"
+
+// TODO: Delete this macro, replacing it with the next one.
+// New "[roomName]NumActions" variables need to be made before that.
+#define ADD_ROOM_OLD(ROOM) {\
+		if (name.equalsIgnoreCase(#ROOM)) {\
+			_roomActionList = ROOM##ActionList;\
+			_numRoomActions = sizeof(ROOM##ActionList) / sizeof(RoomAction);\
```
sev-: should be replaced with ARRAYSIZE()

engines/startrek/room.cpp
```diff
@@ -0,0 +1,539 @@
...
+	}
+
+	memset(&_roomVar, 0, sizeof(_roomVar));
+}
+
+Room::~Room() {
+	delete[] _rdfData;
+}
+
+uint16 Room::readRdfWord(int offset) {
```
sev-: replace this function with FROM_LE_16() macro

engines/startrek/saveload.cpp
```diff
@@ -0,0 +1,401 @@
...
+#include "startrek/startrek.h"
+
+namespace StarTrek {
+
+bool StarTrekEngine::showSaveMenu() {
+	GUI::SaveLoadChooser *dialog;
+	Common::String desc;
+	int slot;
+
+	dialog = new GUI::SaveLoadChooser(_("Save game:"), _("Save"), true);
```
sev-: Then you need to add file POTFILES to your engine

engines/startrek/startrek.cpp
```diff
@@ -0,0 +1,2229 @@
...
+	byte *data = txtFile->_data;
+	while (textIndex != 0) {
+		while (*(data++) != '\0');
+		textIndex--;
+	}
+
+	return (char *)data;
+}
+
+} // End of namespace StarTrek
```
sev-: This file is quite big. I recommend to move the methods to different files (leaving them in StarTreckEngine class): intro.cpp, inventory.cpp, animation.cpp, actors.cpp, or whatever else makes sense.

Drenn1:
In terms of localization: text is dumped with an external tool (https://github.com/drenn1/startrek25_rtools) which, in most cases, should also work with the other languages. Since all languages use the same english audio files (which the text is named by), it should be possible to simply put the other language strings into their own arrays and modify showText to choose the correct one, with just a bit of manual intervention for the strings that don't have audio files.

However, the text dumping program won't work on the floppy versions, since it relies on looking for the distinctive pattern of the audio filename to identify strings; if I want to support changes made between the floppy and CD versions, that will be a bit more work.

engines/startrek/filestream.cpp
```diff
@@ -0,0 +1,80 @@
...
+	delete stream;
+}
+
+FileStream::~FileStream() {
+	delete[] _data;
+}
+
+// ReadStream functions
+
+uint32 FileStream::read(void *dataPtr, uint32 dataSize) {
```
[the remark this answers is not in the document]
Drenn1: This was originally used like a stream, and still sometimes is, but in certain places I actually read from the byte array directly, so extending MemoryReadStreamEndian doesn't quite work. It's not purely a "stream" anymore so perhaps it should be renamed.

engines/startrek/object.h
Drenn1: It should be calling the default constructors / setting default values for primitive types (0), but I can make that explicit.

engines/startrek/room.cpp
Drenn1: Do you mean READ_UINT16()?

sev-: That could also work.

engines/startrek/filestream.cpp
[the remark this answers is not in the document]
sev-: You may extend MemoryReadStream() with getData() method. This is better than reimplementing the whole thing. Could you please point me to where do you read from the array directly?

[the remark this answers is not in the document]
Drenn1: Turns out it's only it one spot, the "StarTrekEngine::getLoadedText" function, since I'm reading from it as a string (char array).

In terms of a "getData" method, it seems that I'll need to make a second copy of the file in memory to access it that way, since there seems to be no direct way to access the MemoryReadStream's data. In this particular case I could just copy the portion of the string I need, which isn't much overhead, though still technically unnecessary.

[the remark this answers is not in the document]
Drenn1: Oh, nevermind, I can keep track of the initial byte array I pass to the MemoryReadStream. Yeah, I'll do that...

engines/startrek/filestream.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include "startrek/filestream.h"
+
+namespace StarTrek {
+
+FileStream::FileStream(byte *data, uint32 len, bool bigEndian) : Common::MemoryReadStreamEndian(data, len, bigEndian) {
+	_data = data;
+}
+
+FileStream::~FileStream() {
+	free(_data);
```
sev-: Instead of doing this, I suggest to extend MemoryReadStreamEndian class constructor with the optional "disposeMemory" flag, just as the MemoryReadStream does, pass it to MemoryReadStream, and kill this FileStream class altogether.

engines/startrek/detection.cpp
```diff
@@ -0,0 +1,394 @@
...
+		},
+		GType_ST25,
+		0,
+	},
+
+	{ // ST25 Amiga ? (EN)
+		{
+			"st25",
+			"Floppy",
+			AD_ENTRY1("data.000", "f0918b6d096455ce2ae6dd5ef973292e"),
```
sev-: Please, comment these out, so we could collect the filesizes. Users will be able to report them.

sev-:
Only couple things left to do: detection entries and that FileStream class, and I will merge the engine in.

engines/startrek/filestream.cpp
Drenn1: I can do that, but as I'm still using the "_data" field, I'd prefer to still keep this class unless you feel strongly about it.

sev-:
Merging. Please, continue your work in-tree

Strangerke:
I'll give a hand this evening, but I can't help while I'm at work.

Drenn1:
Thanks for merging. I'll try to get it fixed, but I don't have push access yet.

PR #1266, opened 2018-07-29
COMMON: Fix kerning issue in wordWrapText.

LubomirR:
Fix kerning issue in wordWrapText.

In certain situation, we have to recompute kerning for the current character because we deleted the previous one.


sev-:
Good catch! merging

PR #1272, opened 2018-08-08
MOHAWK: RIVEN: Delay less for slower systems

dafioram:
1. Move input handling from doFrame to another function
2. Delay less if doFrame loop takes longer.

### DIFF
--- a/engines/mohawk/riven.cpp
+++ b/engines/mohawk/riven.cpp
@@ -216,6 +216,30 @@ void MohawkEngine_Riven::doFrame() {
 		_stack->keyResetAction();
 	}
 
+	processInput();
+
+	_stack->onFrame();
+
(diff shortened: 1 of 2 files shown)


bgK:
I did a bit on testing on my system, with vsync enabled, and this seems to introduce missed frames when turning around.

Without the proposed change:
capture d ecran de 2018-08-11 09-47-39

With the proposed change:
capture d ecran de 2018-08-11 09-48-34

I believe waiting should happen before updating the screen so it does not interfer with vsync.

bgK:
Actually, my mistake. The dropped frames were always there. I messed up the delay while testing.

dafioram:
Interesting that you are getting 60fps and not 100fps for the 10ms delay is that due to the vsync or some other 60fps limitation?

PR #1273, opened 2018-08-10
GUI: Add editable path in "Add Game" tab

Sedictious:
Allows users to type-in target path when adding a game . I wasn't sure what to do with invalid paths (e.g. pop-up notifying the user), as it currently silently returns an empty directory.

### DIFF
--- a/gui/widgets/edittext.cpp
+++ b/gui/widgets/edittext.cpp
@@ -132,7 +132,7 @@ void EditTextWidget::startEditMode() {
 void EditTextWidget::endEditMode() {
 	releaseFocus();
 	
-	sendCommand(_ExitTxtCmd, 0);
+	sendCommand(kExitTxtCmd, 0);
 	sendCommand(_finishCmd, 0);
 }
 
@@ -143,7 +143,7 @@ void EditTextWidget::abortEditMode() {
 	releaseFocus();
 }
 
-Common::String EditTextWidget::getEditString(){
+Common::String EditTextWidget::getEditString() {
 	return _backupString;
 }
 
--- a/gui/widgets/edittext.cpp
+++ b/gui/widgets/edittext.cpp
@@ -64,7 +64,6 @@ void EditTextWidget::reflowLayout() {
 	EditableWidget::reflowLayout();
 }
 
-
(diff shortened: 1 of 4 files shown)


gui/widgets/edittext.cpp
```diff
@@ -132,14 +131,20 @@ void EditTextWidget::startEditMode() {
...
 }
 
 void EditTextWidget::abortEditMode() {
 	setEditString(_backupString);
 	sendCommand(_cmd, 0);
+	
 	releaseFocus();
 }
 
+Common::String EditTextWidget::getEditString(){
```
anandmehrotra: Missing space after brackets

dreammaster: Missing space after brackets

gui/widgets/edittext.h
```diff
@@ -25,9 +25,14 @@
 
 #include "gui/widgets/editable.h"
 #include "common/str.h"
+#include "gui/dialog.h"
 
 namespace GUI {
 
+enum {
+	_ExitTxtCmd = 'TXTE'
```
dreammaster: Enum values shouldn't start with underscores. Either use 'k' as a prefix, or use all uppercase like kExitTextCmd or EXIT_TXT_CMD

gui/widgets/edittext.cpp
dreammaster: The commit messages themselves don't conform to our guidelines.. you forgot to include the 'GUI: ' prefix in this case, since they're affecting the GUI code. Please rebase your commits to add in the prefix.

dreammaster:
I haven't had a chance to actually try running it yet, I'll try and do that tonight. Code-wise, though, you just need to fix the identified problems, and do the rebase of the commits to fix their messages.

dreammaster:
Tested it out, and it works well. The commit history just still needs to be rebased. I'd suggest squashing everything into a single commit with a 'GUI: ' prefix in the commit message. This would actually be cleaner than currently, where you've committed each file independently.

Sedictious:
@dreammaster Ahh yes, sorry about that. I was under the impression that you can squash upon merging, so I didn't give it much thought.

PR #1276, opened 2018-08-15
SCUMM : Add support for European Wii Pajama Sam (EAN3546430138939)

BenCastricum:
Playtested it a bit and it seems to work fine.


sev-:
Great, thanks!

PR #1290, opened 2018-08-16
iOS: Make arrow keys (e.g. from external keyboard) work on iOS

MWumpusZ:
Move keyCommands to iPhoneView (ios7_video.mm); here keyCommands is indeed being triggered, when it was in SoftKeyboard (ios7_keyboard.mm), it was never being called.

I only loosely understand why this should help.  Hopefully someone has more experience and can comment?

### DIFF
--- a/backends/platform/ios7/ios7_keyboard.mm
+++ b/backends/platform/ios7/ios7_keyboard.mm
@@ -53,6 +53,30 @@
 	return self;
 }
 
+- (NSArray *)keyCommands {
+	UIKeyCommand *upArrow = [UIKeyCommand keyCommandWithInput: UIKeyInputUpArrow modifierFlags: 0 action: @selector(upArrow:)];
+	UIKeyCommand *downArrow = [UIKeyCommand keyCommandWithInput: UIKeyInputDownArrow modifierFlags: 0 action: @selector(downArrow:)];
+	UIKeyCommand *leftArrow = [UIKeyCommand keyCommandWithInput: UIKeyInputLeftArrow modifierFlags: 0 action: @selector(leftArrow:)];
+	UIKeyCommand *rightArrow = [UIKeyCommand keyCommandWithInput: UIKeyInputRightArrow modifierFlags: 0 action: @selector(rightArrow:)];
+	return [[NSArray alloc] initWithObjects: upArrow, downArrow, leftArrow, rightArrow, nil];
+}
+
+- (void) upArrow: (UIKeyCommand *) keyCommand {
+	[softKeyboard handleKeyPress:273];
+}
+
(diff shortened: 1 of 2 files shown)


MWumpusZ:
See also:
https://github.com/scummvm/scummvm/pull/1225
https://github.com/scummvm/scummvm/commit/f353e466df2175f10e4328bce3f6a44e295ba305#diff-c3a1449120b5bbf18c95022cf9b77bd6

sev-:
It indeed does not look like a logical change. Maybe we will have someone with a better iOS knowledge to check it...

MWumpusZ:
Working on getting that knowledge :)

There's some other oddness to do with input not always being accepted correctly in the iOS version; it looks like inputs are not always being processed by the right "responder" - which I guess is the issue here too.  I have a workaround for that too, but similarly to here, I'm not convinced the workaround is "right".  It'll take a number of days until I get back to this topic though.

MWumpusZ:
Found some more time today after all ;)
The commit just now changes it to what the original forum poster presumably meant; this seems all the more likely as it obviates the need for one of digitalls fixes.

HOWEVER, I'm not yet happy with it.  Everything I've understood so far from the documentation of UIView and UIResponder seems to indicate that digitalls check in was where keyCommands SHOULD be.  I can make his version work by making the inputView a subview of keyboardView (itself a subview of the iPhoneView).  That corresponds to the class layout, and as such feels intuitively correct.  The code so far has both the keyboardView and the inputView directly as a subView of the iPhoneView, which is why the arrow keys not handled by inputView are not being passed to the keyboardView, but rather directly to the inputView.

digitall:
@MWumpusZ : Thanks for taking a look at this code and seeing about fixing it up. I am not an iOS expert and don't even own any Apple or iOS devices, so could not test this, but though it best to get the code into the tree and worked on. I tried to fix it up as far as possible where logical, but I am afraid I am not an expert here so can't offer more guidance / knowledge.

MWumpusZ:
@digitall understood, and thanks for the work!
As I noted, I‘m almost certain this revised version of the PR corresponds to dottostring (from the forum)s intention, and it works :). As such I request that it be considered for merging.
My misgivings regarding the views are secondary for now - there are a few things in the iOS-speific code I‘d consider refactoring, once I understand it better, but that‘s something for a separate PR.

digitall:
@MWumpusZ : Thanks for the work on this. As per your comment here and on #1225, have squashed these commits, made minor commit message fixups and committed as 82e8bc43e60ce5908097d62be1b2fbb59c4679e8 so will close. Any further work and discussion should go to #1225. Thanks.

PR #1299, opened 2018-08-16
STARTREK: atantable static and st25 detect

dafioram:
1.) Make atanTable static
2.) Detect another cd/dos variant of st25


sev-:
Thanks! merging

PR #1300, opened 2018-08-16
Sin/Cos Table update

dafioram:
1.) Sin/Cos table now takes number of points into the constructor rather than exponent of power of 2. It must now be divisible by 4.
2.) Update places that use sin/cos table.

### DIFF
--- a/common/cosinetables.cpp
+++ b/common/cosinetables.cpp
@@ -27,12 +27,11 @@
 
 namespace Common {
 
-CosineTable::CosineTable(int bitPrecision) {
-	assert((bitPrecision >= 4) && (bitPrecision <= 16));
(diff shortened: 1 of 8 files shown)


dafioram:
The current constructor requires that a power of 2 be specified so the number of radian points must be 16, 32, 64, ... 

This change allows the number of points to be specified which need not be a power of 2, so sizes of 16, 20, 24, 28, 32, ... Are possible. They must still be divisible by 4.

Some engines want sin(2 pi I / size) where size isn't a power of 2, I've seen 320 and 640 for size. Which this commit change can handle.

sev-:
Understood. Merging, thanks for the work and your explanation.

PR #1302, opened 2018-08-18
GUI: Add editable path in "Add Game" tab

Sedictious:
Allows users to type-in target path when adding a game . I wasn't sure what to do with invalid paths (e.g. pop-up notifying the user), as it currently silently returns an empty directory.

Essentially identical to  #1273 (didn't know you couldn't reopen PRs after a force-push)

### DIFF
--- a/gui/browser.cpp
+++ b/gui/browser.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "gui/browser.h"
+#include "gui/widgets/edittext.h"
 #include "gui/gui-manager.h"
 #include "gui/widgets/list.h"
 
@@ -53,10 +54,9 @@ BrowserDialog::BrowserDialog(const char *title, bool dirBrowser)
 	_showHidden = false;
 
 	// Headline - TODO: should be customizable during creation time
-	new StaticTextWidget(this, "Browser.Headline", title);
-
+	
 	// Current path - TODO: handle long paths ?
-	_currentPath = new StaticTextWidget(this, "Browser.Path", "DUMMY");
+	_currentPath = new EditTextWidget(this, "Browser.Path", "DUMMY");
 
 	// Add file list
 	_fileList = new ListWidget(this, "Browser.List");
@@ -94,6 +94,12 @@ void BrowserDialog::open() {
 
 void BrowserDialog::handleCommand(CommandSender *sender, uint32 cmd, uint32 data) {
 	switch (cmd) {
+	//Search for typed-in directory
+	case kExitTxtCmd:
+		_node = Common::FSNode(_currentPath->getEditString());
+		updateListing();
+		break;
+	//Search by text input
 	case kChooseCmd:
 		if (_isDirBrowser) {
 			// If nothing is selected in the list widget, choose the current dir.
@@ -157,7 +163,7 @@ void BrowserDialog::handleCommand(CommandSender *sender, uint32 cmd, uint32 data
 
 void BrowserDialog::updateListing() {
 	// Update the path display
-	_currentPath->setLabel(_node.getPath());
+	_currentPath->setEditString(_node.getPath());
 
 	// We memorize the last visited path.
 	ConfMan.set("browser_lastpath", _node.getPath());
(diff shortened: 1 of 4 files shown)


lotharsm:
Hi Sedictious, thank you for your contribution.

Could you please update your commit message to follow our commit guidelines? The correct message would be "GUI: Add editable path in Add Game tab".

Sedictious:
Since I'm pretty new to Git and don't want to mess up -again-, could you please point out how?
I know that I can use ``commit --amend`` but it seems like I must force push, which in my understanding would overwrite some recent changes. Am I missing something?

angstsmurf:
In this case a force push is fine.

Sedictious:
@angstsmurf Got it, thanks! I think everything's fine now.

gui/browser.cpp
```diff
@@ -53,10 +54,9 @@ BrowserDialog::BrowserDialog(const char *title, bool dirBrowser)
 	_showHidden = false;
 
 	// Headline - TODO: should be customizable during creation time
-	new StaticTextWidget(this, "Browser.Headline", title);
```
criezy: Why remove the title?
That change means the title no longer appear anywhere.

criezy:
I'd like to hear a good reason why the title was removed from the BrowserDialog before I can approve these changes. Other than that it looks good.

Sedictious:
@criezy You are completely right, thanks for catching that. At first, I commented it out to focus on the editable area and not worry about possible bugs and forgot about it later. Turns out, that was completely unnecessary and should now work as intended.

criezy:
Thank you for your work on this.
I have manually committed this change after squashing the two commits and making some minor changes (such as removing the tabs you introduced on otherwise empty lines).

LubomirR:
I think this PR breaks renaming game in the edit game dialog unless you change the focus to a different edit control (i.e. if you click OK right after changing the name).

The reason for that seems to be the added getEditString method. Now both EditTextWidget and its base class EditableWidget have a method called getEditString (not marked as a virtual/override).

EditableWidget::getEditString returns the real text, EditTextWidget::getEditString the "backup" text. The "backup" text in EditTextWidget is changed (commited) only when the control loses the focus. Since EditTextWidget has WIDGET_RETAIN_FOCUS and ButtonWidget returns false from wantsFocus, there's no focus change and the "backup" text still has the old value, which is then taken by EditGameDialog::apply.

Sedictious:
@LubomirR  I tested and you're right, I didn't catch this before. Simply renaming getEditString to something else (I tried getBackupString) fixes the problem. Shall I open a new PR or can someone handle this without an additional commit?

LubomirR:
I don't think this can be handled without an additional commit (we shouldn't rewrite history for master), so please open a new PR. Thanks!

Sedictious:
@LubomirR I commited the changes to my remote but the difference shows up as if the previous commit didn't take place. Is it an error on my part or just because of the commit history?

digitall:
@Sedictious : I took a look at your github scummvm fork and you have made that commit on a different branch i.e. https://github.com/Sedictious/scummvm/tree/edit_path_f than you made this pull request from i.e. https://github.com/Sedictious/scummvm/tree/dir_path ... in any case, neither of those would merge cleanly as @criezy committed this by cleanup so this is not the same commit and your master branch is outdated. I am assuming you are not that familar with Git and how to use rebase and branch in detail to cleanup / change this.

I would assume that the new commit you want to merge in would be https://github.com/Sedictious/scummvm/commit/530de6e0ebcfc69c367f8a2ca07a4d2707325b89 ... however that will not apply cleanly as @bgk has made a commit to clean this up already as ac4b2efd3fb7fd7a02fc2bc74de249045d353b2b ... so I think this may not be necessary as that the latest clean git master of ScummVM would not exhibit the issue?

If it still does, then you need to do an update of your master branch to be in sync with the latest scummvm/master, then branch a new branch from that, cherry-pick your commit onto it and fixup the merge conflicts and then open a new pull request ... thanks.

LubomirR:
I've just tested, the issue seems to be fixed by ac4b2ef, so no further action is needed.

PR #1304, opened 2018-08-18
COMMON: Cos/Sin Table switch internal structure so at() is faster

dafioram:
A new internal table has been added so that no if checks need to
be performed for the at() lookup.

The old table can still be accessed using getTable or atLegacy().

at() and atLegacy() return the same values, but at() is faster.

I am seeing that on average the at() is about 30% faster than the atLegacy() call.


sev-:
Thanks, merging. I also sent you an invite to the team.

PR #1306, opened 2018-08-19
GROOVIE: Disable loading saves during runtime

dafioram:
There is currently a crash that will happen when loading another save
during some puzzles. So we just disable loading at runtime.

Saving during runtime works just fine so we keep that.

See Trac#10664.

### DIFF
--- a/engines/groovie/groovie.cpp
+++ b/engines/groovie/groovie.cpp
@@ -350,8 +350,11 @@ Common::Platform GroovieEngine::getPlatform() const {
 bool GroovieEngine::hasFeature(EngineFeature f) const {
 	return
 		(f == kSupportsRTL) ||
-		(f == kSupportsSavingDuringRuntime) ||
-		(f == kSupportsLoadingDuringRuntime);
+		(f == kSupportsSavingDuringRuntime);
+		// Game will crash during loading at a few puzzles.
+		// So don't enable until thats been figured out.
+		// See Trac#10664.
+		//(f == kSupportsLoadingDuringRuntime);
 }
 
 void GroovieEngine::syncSoundSettings() {


bluegr:
That looks to be too generic. Does the game crash in all puzzles, or just the crypt puzzle?

dafioram:
When trying to load a saved game using the GMM while a puzzle is interactive will cause a crash in some puzzles

bluegr:
But you just said the same thing: "some puzzles". Which ones? In the ticket, you mention the crypt puzzle. Is there any other puzzle that has this issue?

dafioram:
I have updated the ticket to reflect the load behavior for various puzzles in the game.

dafioram:
Any objections to merging this? The loading can always be enabled later once thats figured out.

dafioram:
That sounds okay, we should probably talk about the gmm saving as well. Do you think #10663 is caused by the gmm saving I've added?

digitall:
You mean https://bugs.scummvm.org/ticket/10663 ... As you said there it is certainly possible. If the game state is not such that it allows loading without leaving some odd "leftover" state, then it is likely that in those situations the converse is true.

Really, the Groovie engine and savestate contents should be examined to work out what is not covered by the savestate and thus what might need reset at load or to be added to the save/load format.... but again for now, GMM saving might want the same solution to avoid invalid gamestate files.

scott-t:
At a hunch I would have said that loading games at any time should be safe (assuming you're loading games saved through the normal process, and when you 'load' them, the game picks up from as if it's just loaded straight out of the normal ouija board).

I could imagine an issue saving a game with the GMM though - the normal process of saving a game will return from the puzzle subscript, replace some of the variables that were pushed as a saved state (see o_loadscript and o_returnscript), then go through the save stages (opening ouija board, etc).  Variables 0x00 through 0xFF get cleared a lot (there's a dedicated opcode for it), but 0x100-0x106 are a some special ones that may not being tidied up properly, which combined with the state restore of 0x107 onwards that usually happens may cause some odd behaviour.

digitall:
Have merged #1376 which should fix this without needing to disable GMM load or save, so am going to close this.

@dafioram : Thanks for finding these bugs and proposing the initial fix... If you test and find any issues, please feel free to comment / update.

PR #1309, opened 2018-08-20
BUILD: Remove .dwo files on clean

eriktorbjorn:
I noticed that the ScummVM build process creates .dwo files in addition to the .o files when -gsplit-dwarf is used. This pull request is my attempt to remove them on "make clean". I'm rather rusty on Makefile syntax, but I _hope_ it's correct.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .*.swp
 .*.swo
 *.o
+*.dwo
 lib*.a
 .deps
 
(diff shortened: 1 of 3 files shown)


digitall:
Have tested and works fine for both "make" and "make clean" to remove the *.dwo littering codebase from the split-dwarf debug objects.

Reading the patch, this is pretty simple and the substitution of *.o to *.dwo looks fine and matches examples online, so am merging this as-is.

PR #1311, opened 2018-08-23
Fix build failure

h3xx:
Only happens when `--enable-plugins` is turned on.

It has to do with `$(PLUGIN-$(MODULE))` not substituting while inside of a make recipe, resulting in `dwp -e` executing with no argument, hence the FTBFS.

Version information:

- GNU make 4.2.1
- GNU dwp version 2.31.1-slack15
- scummvm e97b1e560d4f3a0eed758047e8c40ecc69c98231


digitall:
Replicated here and this patch fixes the issue. Will merge.
@h3xx : Thanks for the contribution.

PR #1314, opened 2018-08-25
WINTERMUTE: Flush ConfMan to disk on every change

lolbot-iichan:
There are 4 use-cases of storing persistent data by WME in ConfMan: registry read/write, remembering MostRecentSaveSlot, storing volume settings, and some "last_run" flag that seems to be never used both by engine and games I know of. However, only writing to ConfMan is implemented for all those cases, without flushing. In master branch of ScummVM the only place where ConfMan.flushToDisk() is called is BaseGame::autoSaveOnExit() that is actually never called (commented at platform_osystem.cpp). This means that all those settings are actually not saved to disk.
https://github.com/scummvm/scummvm/blob/3e34cec7ced19e3a0f91588653b8cf4868db18d6/engines/wintermute/platform_osystem.cpp#L122

Registry read/write is an API that is rarely used in most WME games I decompiled (not used at all by games Four, Hamlet, OpenQuest, Dreamcat Adventures, Rosemary, White Chamber, Trader of Stories; used only for page number at saves menu at Helga Deep in Trouble; used only for volume and subtitles settings in Tanya Grotter series), and effects of those values lost was minor for all those games, that's why this bug existed for years and noone ever noticed it. FoxTail game however use RegWriteNumber() API for dozen of settings which affects gameplay and have unwanted default values re-applied on every game save+quit+start+load loop, which makes this a serious bug.

So, we should either reintroduce autoSaveOnExit of some sort (which I consider risky, as it was removed for some reason I don't know of, and also it is not guaranteed to be called, as game process may crash or devices battery may run out before graceful shutdown of game, causing some data to be lost, which seems to me to be against RegWriteNumber semantics), or we could add flush to every writing to make sure nothing is lost (which causes writing to disk on every change of every parameter, which is a kind of overhead, demonstrated in attached changelist).

I wonder how it is handled in other engines and what should we actually do here. Need feedback.

### DIFF
--- a/engines/wintermute/base/base_game.cpp
+++ b/engines/wintermute/base/base_game.cpp
@@ -237,6 +237,7 @@ BaseGame::~BaseGame() {
 	LOG(0, "Shutting down...");
 
 	ConfMan.setBool("last_run", true);
+	ConfMan.flushToDisk();
 
 	cleanup();
 
@@ -1255,6 +1256,7 @@ bool BaseGame::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
 		int val = stack->pop()->getInt();
 		Common::String privKey = "priv_" + StringUtil::encodeSetting(key);
 		ConfMan.setInt(privKey, val);
+		ConfMan.flushToDisk();
 		stack->pushNULL();
 		return STATUS_OK;
 	}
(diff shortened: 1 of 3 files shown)


bluegr:
Nice catch!

You would normally need a single place where flushToDisk() is called, after the engine exits its main loop. The engine destructor might be another place, and I see you've put a call there already. The other calls seem superfluous, since the call you placed in the destructor will be called in any case.

lolbot-iichan:
@bluegr 
Thanks!

I removed superfluous calls, now only 2 flushToDisk() are left:
1. at ~BaseGame(), right after writing "last_run" to ConfMan
2. at SaveLoad::saveGame(), right after actual game save is written on disk and "most_recent_saveslot" is updated at ConfMan

I think the second one is needed as well, as player may expect all data to be persisted during the save operations so it is possible to turn on power completely right after save is done and nothing should be lost.

bluegr:
Great, thanks! I've squashed your commits in one, and merged. Closing

PR #1315, opened 2018-08-25
COMMON: Update RDFT and DCT cos/sin table constructor usage

dafioram:
When the constructor for the cos/sin table was changed from
number of bits to number of points all usages thoughout the
code should of been changed, but this was missed in RDFT and DCT.

Fixes Trac#10683.


sev-:
Thanks, merging.

PR #1317, opened 2018-08-25
LASTEXPRESS: multiple fixes in NPC logic

grechnik:
Checked the logic against the original game
(to be precise, DOS English version from GOG, although I think
AI logic has no significant differences with other versions).
Fixed a *lot* of errors with varying visibility for the user.

Also, save+exit+load sometimes resulted in memory corruption like
((EntityParametersSSII*)(new EntityParametersIIII))->param8 = 0;
load operation did not restore the correct type of NPC logic context,
the default one was used (which also has the smallest sizeof).
Should be fixed now. Save+load is still unusable because it locks
everybody waiting for kActionEndSound (the sound state is not restored),
but, at least, it should not corrupt the memory. Hopefully.


engines/lastexpress/debug.cpp
```diff
@@ -487,10 +487,7 @@ bool Debugger::cmdPlaySeq(int argc, const char **argv) {
 
 					// Handle right-click to interrupt sequence
 					Common::Event ev;
-					if (!_engine->getEventManager()->pollEvent(ev))
-						break;
-
-					if (ev.type == Common::EVENT_RBUTTONUP)
+					if (_engine->getEventManager()->pollEvent(ev) && ev.type == Common::EVENT_RBUTTONUP)
```
sev-: This changes the behviour. The equivalent one would be ``(!_engine || ev.type)``. Unless you have a reason to change it.

grechnik: This is intentional. The old behaviour is "while the event queue is not empty and does not start with a right-click event, remove the first event and play the sequence", which does not make much sense (in practice, it just draws first few frames of the sequence and exits). The new behaviour is "play the sequence until right-click event (or till the end of the sequence)".

engines/lastexpress/sound/entry.h
```diff
@@ -76,12 +76,15 @@ namespace LastExpress {
...
 class SubtitleEntry;
 
+// TODO: this union assumes little-endian machine
 union SoundStatusUnion {
 	uint32 status;
-	byte status1;
-	byte status2;
-	byte status3;
-	byte status4;
+	struct {
```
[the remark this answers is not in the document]
grechnik: Yep, this assumes little-endian host. I have not changed the code around this except `.statusX` -> `.b.statusX`. The code was totally broken and now it is broken on big-endian machines, I call that an improvement. There is a `// TODO` before this union to remind about the problem.

grechnik:
> Would you like to join the team and continue your work in-tree?

Yes, that would be nice. I won't be very active, though.

sev-:
Merging

PR #1318, opened 2018-08-26
MOHAWK: RIVEN: Return to main menu after the credits

dafioram:
Like the original, the game now returns to main menu upon the
credits finishing. The game cannot be resumed.

The states have been reset so combos are changed and puzzles are reset.

### DIFF
--- a/engines/mohawk/riven.cpp
+++ b/engines/mohawk/riven.cpp
@@ -94,6 +94,12 @@ MohawkEngine_Riven::MohawkEngine_Riven(OSystem *syst, const MohawkGameDescriptio
 	SearchMan.addSubDirectoryMatching(gameDataDir, "program");
 }
 
+void MohawkEngine_Riven::restart() {
+	startNewGame();
+	_menuTumbnail.reset();
+	goToMainMenu(false);
+}
+
 MohawkEngine_Riven::~MohawkEngine_Riven() {
 	delete _card;
 	delete _stack;
@@ -334,9 +340,14 @@ void MohawkEngine_Riven::processInput() {
 	}
 }
 
-void MohawkEngine_Riven::goToMainMenu() {
-	_menuSavedStack = _stack->getId();
-	_menuSavedCard = _card->getId();
+void MohawkEngine_Riven::goToMainMenu(bool allowResume) {
+	if (allowResume) {
+		_menuSavedStack = _stack->getId();
+		_menuSavedCard = _card->getId();
+	} else {
+		_menuSavedStack = -1;
+		_menuSavedCard = -1;
+	}
 
 	// If we are already in menu, do not call again
 	if (_menuSavedStack == kStackAspit && _menuSavedCard == 1) {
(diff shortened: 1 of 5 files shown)


dafioram:
The game seems to work fine, but if you try to watch one of the end games again from the main menu after finishing the game once, it will show the end game video flashing and then it will core dump.

dafioram:
After rebasing there is no video issues, but the end game videos can still segfault when trying to play another game ending after the first.

`Thread 1 "scummvm" received signal SIGSEGV, Segmentation fault.
0x000055555563ec4f in Common::HashMap<unsigned short, Mohawk::MohawkSurface*, Common::Hash<unsigned short>, Common::EqualTo<unsigned short> >::begin (
    this=0x555556483248) at ./common/hashmap.h:258
258				if (_storage[ctr] && _storage[ctr] != HASHMAP_DUMMY_NODE)

(gdb) bt
#0  0x000055555563ec4f in Common::HashMap<unsigned short, Mohawk::MohawkSurface*, Common::Hash<unsigned short>, Common::EqualTo<unsigned short> >::begin() (this=0x555556483248) at ./common/hashmap.h:258

#1  0x000055555563dc08 in Mohawk::GraphicsManager::clearCache() (this=0x555556483240) at engines/mohawk/graphics.cpp:79

#2  0x00005555555ff004 in Mohawk::MohawkEngine_Riven::changeToCard(unsigned short) (this=0x7ffff4ac7ca0 <main_arena+96>, dest=1)
    at engines/mohawk/riven.cpp:605

#3  0x000055555561e203 in Mohawk::RivenStack::runCredits(unsigned short, unsigned int, unsigned int) (this=0x5555565a9ab0, video=11, delay=5000, videoFrameCountOverride=0) at engines/mohawk/riven_stack.cpp:265
`
`

bgK:
One needs to be careful when calling `_vm->changeToStack`, because it destroys the currently active stack objet. That is to say the `this` pointer is no longer valid after calling it in this context. It's much safer to queue a stack change using the script system so it's executed on the next frame. See what's done in `ASpit::xaNewGame`. I'm not sure that this is your problem though.

dafioram:
This is the best version yet, no crashes, but I am seeing at least one issue.

The game doesn't return to the main menu after watching one of the ends, but instead goes to a blank screen, this is for the ending where Ghen kills the player in his office for refusing to enter the book.

Also the player is able to save at the main menu after restarting. That should be simple to fix.

dafioram:
Alright, I've fixed that game ending. It wasn't running the script to change to the main menu because the script manager was stopped in Ospit::xbookclick. I removed the blocking of the scripts from running.

I tried keeping that in and doing a clearing of the script manager que, but that lead to script running issues in doFrame(), maybe that stopAllScript() can be kept in by changing goToMainMenu menu change script to happen immediately rather than que it up, if that is desired, but that will need to be checked against the other game endings.

I am now able to play though all game endings in the same riven instance.

dafioram:
I loaded a saved game at the good ending, then played through the game again choosing the good ending again and it played fine.

 I did encounter Trac#10698, but I think thats not riven specific and most people won't be switching back and forth between windowed/fullscreen.

bgK:
Thanks! I've merged this manually, simplifying the code a bit.

PR #1319, opened 2018-08-27
WINTERMUTE: Wintermute subengine for FoxTail game

lolbot-iichan:
FoxTail is a point-and-click adventure game by Gingertips Game Studio, featuring extended and hacked WinterMute engine fork with closed sources. Most features are added by introducing new API methods and properties, but some are just hardcoded to be different from normal WME stuff.

I'd like to get some feedback on how I implemented those features while I look if I can do anything with known issues and re-test everything. Most arguable things are:
0. should new methods and properties be avaliable for other WME games or should we add if(BaseEngine::instance().getGameId() == "foxtail") guard everywhere in addition to subengine's #ifdef? Is if(BaseEngine::instance().getGameId() == "foxtail") the right way to detect FoxTail inside engine?
1. most SX-objects don't have "normal" constructors, only those that are using values from ScStack. Is it ok to put use ScStack for casting integers into makeSXDate, etc?
2. version properties - should they be the same as in actual games or return something else for debug purposes? Should we show ScummVM's version instead? What's the right way to get detected game version into the engine?
3. is it ok to supress an error if everything seems to work fine? I guess I should do some investigations on LoadItems runtime error that I suppress.
4. could I somehow avoid adding _minimizeSpacing to BaseFontBitmap to avoid messing with BaseFontBitmap::persist? I don't want to change sizeof(BaseFontBitmap).
5. I just made _savedTimestamp field of BasePersistenceManager public. Should I add getter function instead?

Known issues: 
1. Dinamic lighting is working, but there is a semi-transparent rectangle around actor. See https://bugs.scummvm.org/ticket/10686
2. Clock in grandma's room are not rendered correctly. See https://bugs.scummvm.org/ticket/10684
3. Progress bar of game save/load is rendered above transparent loading_saving.png, not below as it hacked in FoxTail. Will look into this.

### DIFF
--- a/engines/wintermute/ad/ad_entity.cpp
+++ b/engines/wintermute/ad/ad_entity.cpp
@@ -34,6 +34,7 @@
 #include "engines/wintermute/ad/ad_sentence.h"
 #include "engines/wintermute/base/base_active_rect.h"
 #include "engines/wintermute/base/base_dynamic_buffer.h"
+#include "engines/wintermute/base/base_engine.h"
 #include "engines/wintermute/base/base_file_manager.h"
 #include "engines/wintermute/base/base_game.h"
 #include "engines/wintermute/base/base_parser.h"
@@ -67,6 +68,10 @@ AdEntity::AdEntity(BaseGame *inGame) : AdTalkHolder(inGame) {
 	_walkToX = _walkToY = 0;
 	_walkToDir = DI_NONE;
 
+#ifdef ENABLE_FOXTAIL
+	_hintX = _hintY = -1;
+#endif
+
 	_theora = nullptr;
 }
 
@@ -97,6 +102,16 @@ TDirection AdEntity::getWalkToDir() const {
 const char *AdEntity::getItemName() const {
 	return _item;
 }
+
+#ifdef ENABLE_FOXTAIL
+int32 AdEntity::getHintX() const {
+	return _hintX;
+}
+
+int32 AdEntity::getHintY() const {
+	return _hintY;
+}
+#endif
 
 //////////////////////////////////////////////////////////////////////////
 bool AdEntity::loadFile(const char *filename) {
@@ -164,6 +179,10 @@ TOKEN_DEF(WALK_TO_X)
 TOKEN_DEF(WALK_TO_Y)
 TOKEN_DEF(WALK_TO_DIR)
 TOKEN_DEF(SAVE_STATE)
+#ifdef ENABLE_FOXTAIL
+TOKEN_DEF(HINT_X)
+TOKEN_DEF(HINT_Y)
+#endif
 TOKEN_DEF_END
 //////////////////////////////////////////////////////////////////////////
 bool AdEntity::loadBuffer(char *buffer, bool complete) {
@@ -210,6 +229,10 @@ bool AdEntity::loadBuffer(char *buffer, bool complete) {
 	TOKEN_TABLE(WALK_TO_Y)
 	TOKEN_TABLE(WALK_TO_DIR)
 	TOKEN_TABLE(SAVE_STATE)
+#ifdef ENABLE_FOXTAIL
+	TOKEN_TABLE(HINT_X)
+	TOKEN_TABLE(HINT_Y)
+#endif
 	TOKEN_TABLE_END
 
 	char *params;
@@ -487,6 +510,14 @@ bool AdEntity::loadBuffer(char *buffer, bool complete) {
 				i = DI_NONE;
 			}
 			_walkToDir = (TDirection)i;
+#ifdef ENABLE_FOXTAIL
+		case TOKEN_HINT_X:
+			parser.scanStr(params, "%d", &_hintX);
+			break;
+		case TOKEN_HINT_Y:
+			parser.scanStr(params, "%d", &_hintY);
+			break;
+#endif
 		}
 		break;
 
@@ -900,6 +931,24 @@ ScValue *AdEntity::scGetProperty(const Common::String &name) {
 		return _scValue;
 	}
 
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] HintX
+	//////////////////////////////////////////////////////////////////////////
+	else if (name == "HintX") {
+		_scValue->setInt(_hintX);
+		return _scValue;
+	}
+
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] HintY
+	//////////////////////////////////////////////////////////////////////////
+	else if (name == "HintY") {
+		_scValue->setInt(_hintY);
+		return _scValue;
+	}
+#endif
+
 	//////////////////////////////////////////////////////////////////////////
 	// WalkToDirection
 	//////////////////////////////////////////////////////////////////////////
@@ -950,6 +999,24 @@ bool AdEntity::scSetProperty(const char *name, ScValue *value) {
 		_walkToY = value->getInt();
 		return STATUS_OK;
 	}
+
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] HintX
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "HintX") == 0) {
+		_hintX = value->getInt();
+		return STATUS_OK;
+	}
+
+	//////////////////////////////////////////////////////////////////////////
+	// HintY
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "HintY") == 0) {
+		_hintY = value->getInt();
+		return STATUS_OK;
+	}
+#endif
 
 	//////////////////////////////////////////////////////////////////////////
 	// WalkToDirection
@@ -1011,6 +1078,11 @@ bool AdEntity::saveAsText(BaseDynamicBuffer *buffer, int indent) {
 	if (_walkToDir != DI_NONE) {
 		buffer->putTextIndent(indent + 2, "WALK_TO_DIR=%d\n", (int)_walkToDir);
 	}
+
+#ifdef ENABLE_FOXTAIL
+	buffer->putTextIndent(indent + 2, "HINT_X=%d\n", _hintX);
+	buffer->putTextIndent(indent + 2, "HINT_Y=%d\n", _hintY);
+#endif
 
 	for (uint32 i = 0; i < _scripts.size(); i++) {
 		buffer->putTextIndent(indent + 2, "SCRIPT=\"%s\"\n", _scripts[i]->_filename);
@@ -1107,6 +1179,13 @@ bool AdEntity::persist(BasePersistenceManager *persistMgr) {
 	persistMgr->transferSint32(TMEMBER_INT(_walkToDir));
 
 	persistMgr->transferPtr(TMEMBER_PTR(_theora));
+
+#ifdef ENABLE_FOXTAIL
+    if (BaseEngine::instance().isFoxTail(FOXTAIL_1_2_527, FOXTAIL_LATEST_VERSION)) { 
+	    persistMgr->transferSint32(TMEMBER(_hintX));
+	    persistMgr->transferSint32(TMEMBER(_hintY));
+	}
+#endif
 
 	return STATUS_OK;
 }
--- a/engines/wintermute/ad/ad_item.cpp
+++ b/engines/wintermute/ad/ad_item.cpp
@@ -535,6 +535,23 @@ bool AdItem::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack, 
 		return STATUS_OK;
 	}
 
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] RemoveNormalCursor
+	// Used while changing cursor type at some included script
+	// Return value is never used
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "RemoveNormalCursor") == 0) {
+		stack->correctParams(0);
+
+		delete _cursorNormal;
+		_cursorNormal = nullptr;
+
+		stack->pushNULL();
+		return STATUS_OK;
+	}
+#endif
+
 	//////////////////////////////////////////////////////////////////////////
 	// GetNormalCursor
 	//////////////////////////////////////////////////////////////////////////
@@ -583,6 +600,23 @@ bool AdItem::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack, 
 		}
 		return STATUS_OK;
 	}
+
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] RemoveHoverCursor
+	// Used while changing cursor type at some included script
+	// Return value is never used
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "RemoveHoverCursor") == 0) {
+		stack->correctParams(0);
+
+		delete _cursorHover;
+		_cursorHover = nullptr;
+
+		stack->pushNULL();
+		return STATUS_OK;
+	}
+#endif
 
 	//////////////////////////////////////////////////////////////////////////
 	// GetHoverCursor
--- a/engines/wintermute/ad/ad_response_box.cpp
+++ b/engines/wintermute/ad/ad_response_box.cpp
@@ -29,6 +29,7 @@
 #include "engines/wintermute/ad/ad_game.h"
 #include "engines/wintermute/ad/ad_response.h"
 #include "engines/wintermute/ad/ad_response_box.h"
+#include "engines/wintermute/base/base_engine.h"
 #include "engines/wintermute/base/base_dynamic_buffer.h"
 #include "engines/wintermute/base/base_file_manager.h"
 #include "engines/wintermute/base/base_parser.h"
@@ -190,6 +191,18 @@ bool AdResponseBox::createButtons() {
 					btn->setWidth(width);
 				}
 			}
+
+#ifdef ENABLE_FOXTAIL
+			if (BaseEngine::instance().isFoxTail()) {
+				btn->addScript("interface/scripts/dialogue_button.script");
+				btn->setWidth(120);
+				if (_fontHover == nullptr) {
+					btn->setFontHover(btn->getFont());
+					btn->setFontPress(btn->getFontHover());
+				}
+			}
+#endif
+
 			btn->setName("response");
 			btn->correctSize();
 
@@ -500,6 +513,22 @@ bool AdResponseBox::display() {
 	// prepare response buttons
 	bool scrollNeeded = false;
 	for (i = _scrollOffset; i < _respButtons.size(); i++) {
+
+#ifdef ENABLE_FOXTAIL
+		// FoxTail's "HORIZONTAL=TRUE" display boxes are actual 2x3 display boxes
+		// Tests show that this hack was removed in FOXTAIL_1_2_362
+		if (_horizontal && BaseEngine::instance().isFoxTail(FOXTAIL_OLDEST_VERSION, FOXTAIL_1_2_304)) {
+			if (i >= _scrollOffset + 6) {
+				scrollNeeded = true;
+				break;
+			}
+			_respButtons[i]->setVisible(true);
+			_respButtons[i]->_posX = 55 + 120 * (i / 3);
+			_respButtons[i]->_posY = 100 + 10 * (i % 3);
+			continue;
+		}
+#endif
+
 		if ((_horizontal && xxx + _respButtons[i]->getWidth() > rect.right)
 		        || (!_horizontal && yyy + _respButtons[i]->getHeight() > rect.bottom)) {
 
--- a/engines/wintermute/base/base_game.cpp
+++ b/engines/wintermute/base/base_game.cpp
@@ -1260,6 +1260,21 @@ bool BaseGame::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
 		return STATUS_OK;
 	}
 
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] RegistryFlush
+	// Return value is never used
+	// Used at SaveGameSettings() and Game.RegistryFlush()
+	// Called after a series of RegWriteNumber calls
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "RegistryFlush") == 0) {
+		stack->correctParams(0);
+		ConfMan.flushToDisk();
+		stack->pushNULL();
+		return STATUS_OK;
+	}
+#endif
+
 	//////////////////////////////////////////////////////////////////////////
 	// RegWriteNumber
 	//////////////////////////////////////////////////////////////////////////
--- a/engines/wintermute/base/base_game.cpp
+++ b/engines/wintermute/base/base_game.cpp
@@ -2329,6 +2329,84 @@ ScValue *BaseGame::scGetProperty(const Common::String &name) {
 		return _scValue;
 	}
 
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] SystemLanguage (RO)
+	// Returns Steam API language name string
+	//////////////////////////////////////////////////////////////////////////
+	else if (name == "SystemLanguage") {
+		switch (Common::parseLanguage(ConfMan.get("language"))) {
+		case Common::CZ_CZE:
+			_scValue->setString("czech");
+			break;
+		case Common::DA_DAN:
+			_scValue->setString("danish");
+			break;
+		case Common::DE_DEU:
+			_scValue->setString("german");
+			break;
+		case Common::ES_ESP:
+			_scValue->setString("spanish");
+			break;
+		case Common::FI_FIN:
+			_scValue->setString("finnish");
+			break;
+		case Common::FR_FRA:
+			_scValue->setString("french");
+			break;
+		case Common::GR_GRE:
+			_scValue->setString("greek");
+			break;
+		case Common::HU_HUN:
+			_scValue->setString("hungarian");
+			break;
+		case Common::IT_ITA:
+			_scValue->setString("italian");
+			break;
+		case Common::JA_JPN:
+			_scValue->setString("japanese");
+			break;
+		case Common::KO_KOR:
+			_scValue->setString("koreana");
+			break;
+		case Common::NB_NOR:
+			_scValue->setString("norwegian");
+			break;
+		case Common::NL_NLD:
+			_scValue->setString("dutch");
+			break;
+		case Common::PT_BRA:
+			_scValue->setString("brazilian");
+			break;
+		case Common::PT_POR:
+			_scValue->setString("portuguese");
+			break;
+		case Common::PL_POL:
+			_scValue->setString("polish");
+			break;
+		case Common::RU_RUS:
+			_scValue->setString("russian");
+			break;
+		case Common::SE_SWE:
+			_scValue->setString("swedish");
+			break;
+		case Common::UA_UKR:
+			_scValue->setString("ukrainian");
+			break;
+		case Common::ZH_CNA:
+			_scValue->setString("schinese");
+			break;
+		case Common::ZH_TWN:
+			_scValue->setString("tchinese");
+			break;
+		default:
+			_scValue->setString("english");
+			break;
+		}
+		return _scValue;
+	}
+#endif
+
 	//////////////////////////////////////////////////////////////////////////
 	// Platform (RO)
 	//////////////////////////////////////////////////////////////////////////
--- a/engines/wintermute/base/base_game.cpp
+++ b/engines/wintermute/base/base_game.cpp
@@ -1363,10 +1363,8 @@ bool BaseGame::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
 	else if (strcmp(name, "GetSaveSlotDescription") == 0) {
 		stack->correctParams(1);
 		int slot = stack->pop()->getInt();
-		char desc[512];
-		desc[0] = '\0';
-		SaveLoad::getSaveSlotDescription(slot, desc);
-		stack->pushString(desc);
+		Common::String desc = SaveLoad::getSaveSlotDescription(slot);
+		stack->pushString(desc.c_str());
 		return STATUS_OK;
 	}
 
--- a/engines/wintermute/base/base_game.cpp
+++ b/engines/wintermute/base/base_game.cpp
@@ -1368,6 +1368,49 @@ bool BaseGame::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
 		return STATUS_OK;
 	}
 
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] GetSaveSlotDescriptionTimestamp
+	// Return struct with "Description" and "Timestamp" fields in 1.2.362-
+	// Return array  with "Description" and "Timestamp" items  in 1.2.527+
+	// Timestamps should be comparable types
+	// Used to sort saved games by timestamps at save.script & load.script
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "GetSaveSlotDescriptionTimestamp") == 0) {
+		stack->correctParams(1);
+		int slot = stack->pop()->getInt();
+
+		TimeDate time;
+		SaveLoad::getSaveSlotTimestamp(slot, &time);
+		stack->pushInt(time.tm_sec);
+		stack->pushInt(time.tm_min);
+		stack->pushInt(time.tm_hour);
+		stack->pushInt(time.tm_mday);
+		stack->pushInt(time.tm_mon + 1);
+		stack->pushInt(time.tm_year + 1900);
+		stack->pushInt(6);
+		BaseScriptable *date = makeSXDate(_gameRef, stack);
+		stack->pushNative(date, false);
+
+		Common::String desc = SaveLoad::getSaveSlotDescription(slot);
+		stack->pushString(desc.c_str());
+
+		BaseScriptable *obj;
+		if (BaseEngine::instance().isFoxTail(FOXTAIL_1_2_527, FOXTAIL_LATEST_VERSION)) {
+			stack->pushInt(2);
+			obj = makeSXArray(_gameRef, stack);
+		} else {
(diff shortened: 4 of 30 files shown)


tobiatesan:
Neat, thanks! Will give this a detailed look in the next couple of days.

Are savegames still broken, right?

sev-:
At the number of places this violates our code formatting conventions (http://wiki.scummvm.org/index.php/Code_Formatting_Conventions). Could you please address those? I could mark those spots if you struggling with locating them.

As to how to properly detect the game, I recommend checking against _gameDescription.adDesc. gameId. Normally, there would be a recommendation to add another GID (gameid), but as I see, Wintermute engine does not use that functionality. This will assure that you're not relying on the game data but have a consistent game id to check against.

engines/wintermute/base/base_persistence_manager.h
```diff
@@ -45,6 +45,7 @@ class BaseGame;
 class BasePersistenceManager {
 public:
 	char *_savedDescription;
+	TimeDate _savedTimestamp;
```
somaen: As mentioned elsewhere, please just create a getter.

engines/wintermute/base/base_keyboard_state.cpp
```diff
@@ -432,6 +576,82 @@ uint32 BaseKeyboardState::keyCodeToVKey(Common::Event *event) {
 
 //////////////////////////////////////////////////////////////////////////
 Common::KeyCode BaseKeyboardState::vKeyToKeyCode(uint32 vkey) {
+#ifdef ENABLE_FOXTAIL
```
somaen: This might get cleaner if you create a separate function for the foxTail-specifics, and call that from here, that way this function avoids becoming this much longer.

engines/wintermute/base/base_keyboard_state.cpp
```diff
@@ -337,8 +348,141 @@ enum VKeyCodes {
...
+	ftVkF12        = 1073741893,
+
+	ftVkNumLock    = 1073741907,
+	ftVkScroll     = 1073741895
+};
+#endif
+
 //////////////////////////////////////////////////////////////////////////
 uint32 BaseKeyboardState::keyCodeToVKey(Common::Event *event) {
+#ifdef ENABLE_FOXTAIL
```
somaen: See comment below about separate functions. Can probably get away with a single set of #ifdef's around the implementations too.

engines/wintermute/base/font/base_font_bitmap.cpp
```diff
@@ -56,6 +57,9 @@ BaseFontBitmap::BaseFontBitmap(BaseGame *inGame) : BaseFont(inGame) {
 	_fontextFix = false;
 	_freezable = false;
 	_wholeCell = false;
+#ifdef ENABLE_FOXTAIL
```
somaen: In the interest of save game logic, I guess it might be preferable to unconditionally have this variable, but make it false for all non-foxtail games.

engines/wintermute/base/font/base_font_bitmap.cpp
```diff
@@ -509,6 +518,11 @@ bool BaseFontBitmap::persist(BasePersistenceManager *persistMgr) {
 		persistMgr->getBytes(_widths, sizeof(_widths));
 	}
 
+#ifdef ENABLE_FOXTAIL
+	if (!persistMgr->getIsSaving()) {
+		_minimizeSpacing = BaseEngine::instance().getGameId() == "foxtail";
```
somaen: Am I understanding this right? minimizeSpacing is always true in foxtail?

engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
```diff
@@ -274,6 +273,17 @@ uint32 BaseSurfaceOSystem::getPixelAt(Graphics::Surface *surface, int x, int y)
 	return 0;
 }
 
+//////////////////////////////////////////////////////////////////////////
+bool BaseSurfaceOSystem::getPixel(int x, int y, byte *r, byte *g, byte *b, byte *a)
+{
```
somaen: Wrong brace style, please use the same style you see in line 242.

engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
```diff
@@ -274,6 +273,17 @@ uint32 BaseSurfaceOSystem::getPixelAt(Graphics::Surface *surface, int x, int y)
...
+	uint32 pixel = getPixelAt(_surface, x, y);
+	*r = (pixel >> 24) & 255;
```
somaen: Is this endian-safe?

engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
lolbot-iichan: Fixed in b956156.

engines/wintermute/base/font/base_font_bitmap.cpp
lolbot-iichan: Yes. FoxTail has a globally enabled hack for making spaces between character and spaces between lines smaller than it is according to FONT settings. I added _minimizeSpacing to avoid checking if game is FoxTail on each BaseFontBitmap::getCharWidth(). 

Now when I think about it, maybe we should just make some globally avaliable bool at BaseEngine::instance() and remove _minimizeSpacing from BaseFontBitmap completely?

engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
lolbot-iichan: Probably not. Thanks, I'll look into it...

engines/wintermute/base/font/base_font_bitmap.cpp
lolbot-iichan: See comment below. Maybe _minimizeSpacing  should be removed completely.

engines/wintermute/base/base_persistence_manager.h
lolbot-iichan: Done in b6acee5.

engines/wintermute/base/base_keyboard_state.cpp
lolbot-iichan: Done in fd3846e. Also changed constant names according to http://wiki.scummvm.org/index.php/Code_Formatting_Conventions#Naming

engines/wintermute/base/base_keyboard_state.cpp
lolbot-iichan: Also done in fd3846e.

lolbot-iichan:
@tobiatesan 
Savegames seems to be fine. 
AdActor has no new persisted fields since I reimplemented SetSpeedWalkAnim using "frame->_delay = speedWalk". 
The only new persisted field left is _minimizeSpacing at BaseFontBitmap, which seems unneeded too.

@sev- 
I hope I fixed most code formatting violations with b956156. Please, let me know if I missed any more.
I currently use _gameDescription->adDesc.gameId for checking if current game is FoxTail.
I also use _gameDescription->adDesc.extra at aa97ca4 to check which exact FoxTail build it is.

@somaen 
1. https://github.com/scummvm/scummvm/pull/1319/files#diff-935ae44957169f7436a484684ab18be3R1385 here it is. I push char* to make SXString from Common::String::c_str(), then I push 7 integers to make SXDate from TimeDate, then I push zero to make SXObject. This code looks extremely ugly to me, however I seriously doubt SXObject constructor would ever be usefull outside of this code.
2. They are purely aesthetical, see https://imgur.com/a/UTVFWt7 . Original properties are "1.2.230" and "1291" for 1.2.230.1291, "1.2.362" and "2047" for 1.2.362.2047.
3. I'm not certain that it's noise, so I'll look deeper into what's wrong with LoadItems method and why it does not affect gameplay.
4. I made a second thought over BaseFontBitmap and now I think that _minimizeSpacing could be removed completely. See comments at review.
5. Getter for timestamp is implemented.

engines/wintermute/configure.engine
```diff
@@ -1,3 +1,4 @@
 # This file is included from the main "configure" script
 # add_engine [name] [desc] [build-by-default] [subengines] [base games] [deps]
-add_engine wintermute "Wintermute" yes "" "" "zlib 16bit highres"
+add_engine wintermute "Wintermute" yes "foxtail" "" "zlib 16bit highres"
+add_engine foxtail "FoxTail" yes
```
sev-: no endline

engines/wintermute/detection_tables.h
```diff
@@ -284,6 +287,32 @@ static const WMEGameDescription gameDescriptions[] = {
 	// Escape from the Mansion
 	WME_WINENTRY("escapemansion", "1.3",
 		WME_ENTRY1s("data.dcp",  "1e5d231b56c8a228cd15cb690f50253e", 29261972), Common::EN_ANY, ADGF_UNSTABLE, LATEST_VERSION),
+#ifdef ENABLE_FOXTAIL
+	// FoxTail 1.2.230.1291 (English)
+	WME_WINENTRY("foxtail", "1.2.230.1291",
+		WME_ENTRY1s("data.dcp", "651ae5b062073021edaca7e1de131eec", 59357572), Common::EN_ANY, ADGF_UNSTABLE | ADGF_DEMO, LATEST_VERSION),
```
sev-: why is it marked as ADGF_DEMO?

sev-:
Thanks, the code formatting is fine now.

engines/wintermute/configure.engine
lolbot-iichan: Done at 2c0ea02.

engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
lolbot-iichan: Fixed at b6197ae.

engines/wintermute/base/base_game.cpp
```diff
@@ -2973,6 +3140,44 @@ bool BaseGame::externalCall(ScScript *script, ScStack *stack, ScStack *thisStack
...
+				*it = '\0';
+				stack->pushString(begin);
+				((SXArray *)arr)->push(stack->pop());
+				begin = it + 1;
+			}
+		}
+
+		stack->pushNative(arr, false);
+
+		delete [] copy;
```
tobiatesan: We prefer `delete[]`.

engines/wintermute/base/font/base_font_bitmap.cpp
tobiatesan: > maybe we should just make some globally avaliable bool at BaseEngine::instance() and remove _minimizeSpacing from BaseFontBitmap completely?

I'd say it's a good idea.

Could this be the time to introduce some general way of marking _features_ (including "bugfeatures") required/expected by a WME game?

engines/wintermute/base/scriptables/script.cpp
```diff
@@ -618,7 +619,14 @@ bool ScScript::executeInstruction() {
 						_state = SCRIPT_WAITING_SCRIPT;
 						_waitScript->copyParameters(_stack);
 					}
-				} else {
+				}
+#ifdef ENABLE_FOXTAIL
+				else if (BaseEngine::instance().getGameId() == "foxtail" && !strcmp(methodName, "LoadItems")) {
+					// somehow FoxTail works just fine without LoadItems method
```
tobiatesan: How about logging a warning?

tobiatesan:
@somaen @sev-: tangentially, @lolbot-iichan has built a bunch of test games for the new features.
They currently reside in https://github.com/tobiatesan/wme_testsuite, but do you think that those could be moved somewhere under `scummvm/` (after cleaning it up a bit)?

sev-:
@tobiatesan that is a great idea, we could treat them as demos

bluegr:
Any news on this?

tobiatesan:
Sorry for the extended away, this is a terrible couple of months. 
I haven't heard from @lolbot-iichan... yet :)

bluegr:
And... 3 months have passed. Any news on this? After the reviews, are the changes in a state that can be merged?

engines/wintermute/base/scriptables/script.cpp
lolbot-iichan: Logged a warning in 39d4ceaafa256c1a311b040da1128894d9355d07

lolbot-iichan:
Sorry for being missing for such a long time... Anyway, 2nd chapter of FoxTail was released and here I am again, with a new bunch of changes...

2nd chapter's engine is a bit different not only from original WME Lite, but it's different from 1st chapter's FoxTail engine as well. So, I had to add engine versioning (see a39d89a9864785dc93058c796696a58011ba5c99). Commit c629f1f406ca544249d94f304953dcf5374b78e9 features example of version-depending addition to WME.

There are still some glitches (I don't know how to fix saveload progress bar with dirtyrects), there are still some drawing related bugs posted to bugs.scummvm.org that affects FoxTail), but this pull request allows to play FoxTail, and I'd like to continue work on subengine after reviewing and merging this.

tobiatesan:
Thank you for your work.

Have you considered disabling dirtyrects altogether while saving? It's hacky and I'm not saying it's a great idea, but...

lolbot-iichan:
> Have you considered disabling dirtyrects altogether while saving? It's hacky and I'm not saying it's a great idea, but...
Well, this MAY be a way, but before trying it I need to know, why do we even have 2 different render modes and what are future plans for those 2 modes?

tobiatesan:
Even hackier idea: manually mark the whole screen as dirty, if running FoxTail.

As for plans... there have always been lots of _plans_ for dirty rects, not so much actual work.
In 2013 I wrote some really hacky code to "optimize" dirty rects that ended up being more costly than doing a full `memcpy`, at this point I'm not sure if the idea is inherently broken or my code just sucked.
There were talks of using hardware acceleration under Residual, I don't know if anybody has ever done any work there since.

I'd _personally_ just use the above-mentioned hack, but I won't discourage any attempt at martyrdom :)

engines/wintermute/base/base_keyboard_state.cpp
```diff
@@ -334,11 +345,225 @@ enum VKeyCodes {
...
+	kFtVkF6         = 1073741887,
+	kFtVkF7         = 1073741888,
+	kFtVkF8         = 1073741889,
+	kFtVkF9         = 1073741890,
+	kFtVkF10        = 1073741891,
+	kFtVkF11        = 1073741892,
+	kFtVkF12        = 1073741893,
+
+	kFtVkNumLock    = 1073741907,
+	kFtVkScroll     = 1073741895
```
bluegr: These all look to be ORed with 1073741824 (01000000000000000000000000000000). Perhaps we can just remove this bit and handle them like the other keycodes?

engines/wintermute/ad/ad_response_box.cpp
```diff
@@ -190,6 +191,18 @@ bool AdResponseBox::createButtons() {
...
+			if (BaseEngine::instance().isFoxTail()) {
+				btn->addScript("interface/scripts/dialogue_button.script");
+				btn->setWidth(120);
+				if (_fontHover == nullptr) {
+					btn->setFontHover(btn->getFont());
+					btn->setFontPress(btn->getFontHover());
+				}
+			}
+#endif
+
```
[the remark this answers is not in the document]
lolbot-iichan: I didn't quite believe myself until I checked this:
```
>grep interface\\scripts\\dialogue_button.script foxtail.exe
Binary file foxtail.exe matches
```

So, yes, this exact script file path is hardcoded somewhere in their EXE-file, so I had to hardcode it too.

engines/wintermute/ad/ad_entity.cpp
```diff
@@ -1105,6 +1176,11 @@ bool AdEntity::persist(BasePersistenceManager *persistMgr) {
 
 	persistMgr->transferPtr(TMEMBER_PTR(_theora));
 
+#ifdef ENABLE_FOXTAIL
+	persistMgr->transferSint32(TMEMBER(_hintX));
+	persistMgr->transferSint32(TMEMBER(_hintY));
+#endif
+
```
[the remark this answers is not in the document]
lolbot-iichan: Should I remove #ifdef around _hintX & _hintY definition in header file then?

engines/wintermute/base/base_keyboard_state.cpp
lolbot-iichan: Well, removing 01000000000000000000000000000000 won't help much since those constants are different from `VKeyCodes` anyway.
Compare this (subsequent codes for some keys):
```
	kVkNumLock    = 144,
	kVkScroll     = 145
```
And this (not-subsequent codes for same keys):
```
	kFtVkNumLock    = 1073741907,
	kFtVkScroll     = 1073741895
```
This means that we would need to have 2 enums anyway, even if 01000000000000000000000000000000 part is removed.

engines/wintermute/ad/ad_entity.cpp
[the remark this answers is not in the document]
bluegr: You could do:
    #ifdef ENABLE_FOXTAIL
        persistMgr->transferSint32(TMEMBER(_hintX));
        persistMgr->transferSint32(TMEMBER(_hintY));
    #else
        persistMgr->transferSint32(0);
        persistMgr->transferSint32(0);
    #endif

bluegr:
Any news on this? The work done here is great, so it would be a shame to see it bitrot

engines/wintermute/ad/ad_response_box.cpp
```diff
@@ -495,6 +508,22 @@ bool AdResponseBox::display() {
...
+		// FoxTail's "HORIZONTAL=TRUE" display boxes are actual 2x3 display boxes
+		// Tests show that this hack was removed in FOXTAIL_1_2_362
+		if (_horizontal && BaseEngine::instance().isFoxTail(FOXTAIL_OLDEST_VERSION, FOXTAIL_1_2_304)) {
+			if (i >= _scrollOffset + 6) {
+				scrollNeeded = true;
+				break;
+			}
+			_respButtons[i]->setVisible(true);
+			_respButtons[i]->_posX = 55 + 120*(i/3);
+			_respButtons[i]->_posY = 100 + 10*(i%3);
```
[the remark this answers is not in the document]
lolbot-iichan: Fixed at cb6f6c558a3816a7145505ea98c2b976870a5adb

engines/wintermute/base/base_game.cpp
lolbot-iichan: Fixed at 858525c93ab97e8be47b567f123911ef9189c793

engines/wintermute/detection_tables.h
lolbot-iichan: Currently there are 2 chapters avaliable.
Technically, this still is a "Game in development", but in fact this game is bigger then some finished Wintermute games.
So I'm not sure if it should be ADGF_DEMO...

lolbot-iichan:
So, here I am again.
Changes since June:
1. Everything rebased & merged.
2. I removed `#ifdef` from detection part. Instead, I added a GUI error message in case FoxTail subengine is not compiled in and detected game is FoxTail (similar to GF_LOWSPEC_ASSETS check).
3. Kind of implemented @bluegr's hack with saveload at _hintX & _hintY.
4. I still have no idea how to make load progress bar work with dirty rects. I don't like @tobiatesan's idea of temporary changing mode (since those are separate modes that can be removed in future), and all my tries to force the drawing of the saveload overlay image failed, so I guess I'll leave it as is for now, so this pull request can be integrated.

engines/wintermute/ad/ad_entity.cpp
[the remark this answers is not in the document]
lolbot-iichan: Kind of fixed at acfee7222013254746ca84f6b59d938cbaf97d79.

`transferSint32(0)` seems to be the wrong syntax, and `transferSint32("whatever",NULL)` is a way to get SEGFAULT, so I used unused variable there.

lolbot-iichan:
It's kind of depressing that extention of some classes for FoxTail game breaks savegame compatibility for other games. And I'll have to break it again and again if new features would be added in new FoxTail releases. I wonder if there is any workaround for breaking savegames for FoxTail only.

engines/wintermute/wintermute.cpp
```diff
@@ -158,6 +162,17 @@ int WintermuteEngine::init() {
 		}
 	#endif
 
+	// check dependencies for games with FoxTail subengine
+	#if not defined(ENABLE_FOXTAIL)
+		if (BaseEngine::isFoxTailCheck(_gameDescription->targetExecutable)) {
+			GUI::MessageDialog dialog(_("This game requires FoxTail subengine, that is not compiled in."));
```
bluegr: “This game requires the Foxtail subengine...”

bluegr:
Any news on this?
Also, the Codacity finding isn't anything special, but it could be addressed if Common::String is used instead of plain C strings

bluegr:
Any update?

somaen:
> It's kind of depressing that extention of some classes for FoxTail game breaks savegame compatibility for other games. And I'll have to break it again and again if new features would be added in new FoxTail releases. I wonder if there is any workaround for breaking savegames for FoxTail only.

Well, we could always make some sort of "isFoxTail" property, and version foxtail saves separately from the other WME stuff. This makes slightly more sense as the subengine is disableable (IIRC?), and hence savegames would potentially be incompatible even for non-foxtail games depending on whether or not the foxtail sub-engine is enabled. (Although I guess we would share the actual version numbering to keep things sane, just guarding the foxtail stuff behind a set of isFoxTail()-checks where necessary, and otherwise defaulting them to noop-values if needed).

engines/wintermute/ad/ad_entity.cpp
```diff
@@ -1105,6 +1176,15 @@ bool AdEntity::persist(BasePersistenceManager *persistMgr) {
 
 	persistMgr->transferPtr(TMEMBER_PTR(_theora));
 
+#ifdef ENABLE_FOXTAIL
```
somaen: Hrm, might as well consider making the persistMgr Foxtail-aware, and just only persisting these if we are FoxTail.

engines/wintermute/ad/ad_item.cpp
```diff
@@ -581,6 +600,25 @@ bool AdItem::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack,
...
+#ifdef ENABLE_FOXTAIL
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] RemoveHoverCursor
+	// Used while changing cursor type at some included script
+	// Return value is never used
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "RemoveHoverCursor") == 0) {
+		stack->correctParams(0);
+
+		if (_cursorHover) {
```
somaen: Unneccesary check, delete of nullptr is a safe no-op, so you can reduce this to just "delete _cursorHover;"

engines/wintermute/base/saveload.cpp
```diff
@@ -159,31 +159,39 @@ void SaveLoad::afterLoadScript(void *script, void *data) {
 }
 
 Common::String SaveLoad::getSaveSlotFilename(int slot) {
+	Common::String filename;
 	BasePersistenceManager *pm = new BasePersistenceManager();
-	Common::String filename = pm->getFilenameForSlot(slot);
-	delete pm;
+	if (pm) {
```
somaen: When can this fail (silently)?

bluegr:
This is quite a big effort here, so it's a shame to see it bitrot. Any update on this?

lolbot-iichan:
> This is quite a big effort here, so it's a shame to see it bitrot. Any update on this?

I hope to get back to it before Christmas. I had some hell of an autumn with lots of deadlines before, but it's over and I have some time for hobby-projects again. Luckily, Wintermute had not much changes durin my absence and FoxTail's 3rd episode is still in development, so this PR is kind of still actual.

engines/wintermute/base/base_game.cpp
```diff
@@ -1929,6 +1986,95 @@ bool BaseGame::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
...
+
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] GetScreenMode
+	// Returns integer to be used as a pixelization mode multiplier
+	// Used to init and update controls at options.script and methods.script
+	// This implementation fakes window size as 2*320 x 2*180
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "GetScreenMode") == 0) {
+		stack->correctParams(0);
+		stack->pushInt(2);
```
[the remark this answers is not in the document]
lolbot-iichan: Those "screen modes" are basically duplicate of ScummVM's functionality of "2x", "3x" renderers. Game starts with "2x" mode by default (displayed as "640x360" button) and can be switched to "3x","4x","5x", etc, given enough screen space on actual display. 

изображение
изображение

We have discussed this before with @tobiatesan and @somaen, and decided to disable ingame controls of render modes, because ScummVM supports much more then simple "2x","3x" - there are different 2x filters like "2x", "HQ2x", "2xSAI", "AdvMAME2x", "TV2x". There is no way to pass list of those modes to game script, also game code does not expect "1x" mode to exist, so the best thing we could implement is switching between 2 modes - "2x" and "3x", which is much less than ScummVM is capable of.

It's safe and wise to force the game to always show "640x360" in-game button that does nothing (because game script think that hardware display size is 641x361) and let user control render modes with Alt+Ctrl+N combinations and other common ways shared for all games.

изображение

engines/wintermute/wintermute.cpp
Kawa-oneechan: also, "which is not"?

engines/wintermute/ad/ad_item.cpp
lolbot-iichan: Thanks! Fixed at 6060826512a0c9021cabc4cea0bc2d0879eef688

engines/wintermute/ad/ad_entity.cpp
lolbot-iichan: Fixed at 2be4dde404c39b2bdc37bb74d8095084f95bb7e4.

This was the only place shere PersistentManager related changes were made for FoxTail, so now  this pull request finally should keep savegame compatibility!

engines/wintermute/wintermute.cpp
lolbot-iichan: Thanks! 

Typos and bad english fixed at 8ab48532597a04ef749b2e16b8417f4158a08039.

lolbot-iichan:
Changes since Summer 2019:
1. Everything rebased, conflict resolved.
2. `isFoxTail(FOXTAIL_1_2_527, FOXTAIL_LATEST_VERSION)` check added for HINT_X & HINT_Y properties. Surprisingly, it seems to be the only line needed to make savegames of my branch and master branch compatible. 
3. Fixed minor issues and typos listed in comments.

engines/wintermute/base/saveload.cpp
lolbot-iichan: Hmmmm... No idea, actually.
I guess I was distracted by `if (!pm) { return false;}` in original code of `getSaveSlotDescription`.

Can we leave it as is?
Should I refactor it again & remove all `if (pm)` checks?
Should I refactor it again & add more `if (pm)` checks? =)

engines/wintermute/base/base_game.cpp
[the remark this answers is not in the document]
bluegr: Thanks for the explanation, I agree that we can force the game to always show a fixed value for the resolution, since we already have this functionality, to a much better degree, in ScummVM

engines/wintermute/base/base_game.cpp
```diff
@@ -1941,6 +1998,96 @@ bool BaseGame::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
...
+	//////////////////////////////////////////////////////////////////////////
+	// [FoxTail] ChangeWindowGrab
+	// Used at game.script on "Keypress" event on F11
+	// Readme of FoxTail says: "F11 - free the mouse pointer from the window"
+	// This implementation does nothing
+	// Return value is never used
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "ChangeWindowGrab") == 0) {
+		stack->correctParams(0);
+		stack->pushNULL();
```
bluegr: We have the mouse capture functionality in ScummVM, which is toggled using Control-M:

https://wiki.scummvm.org/index.php?title=User_Manual/Playing_a_game_with_ScummVM#Hotkeys

engines/wintermute/base/font/base_font_bitmap.cpp
```diff
@@ -346,6 +361,10 @@ bool BaseFontBitmap::loadBuffer(char *buffer) {
 	int i;
 	int r = 255, g = 255, b = 255;
 	bool custoTrans = false;
+#ifdef ENABLE_FOXTAIL
+	int ar = 255, ag = 255, ab = 255;
+	bool custoAlpha = false;
```
bluegr: customAlpha

engines/wintermute/base/saveload.cpp
bluegr: This particular check makes no sense to keep, so I'd suggest to remove it. Not sure about the other checks you're referring to

engines/wintermute/base/saveload.cpp
```diff
@@ -159,31 +159,39 @@ void SaveLoad::afterLoadScript(void *script, void *data) {
...
 	}
+	return description;
+}
 
-	if (!(pm->initLoad(filename))) {
+void SaveLoad::getSaveSlotTimestamp(int slot, TimeDate *time) {
+	memset(time, 0, sizeof(TimeDate));
+	Common::String filename = getSaveSlotFilename(slot);
+	BasePersistenceManager *pm = new BasePersistenceManager();
+	if (pm) {
```
bluegr: Same here, please remove this check

engines/wintermute/base/base_keyboard_state.cpp
```diff
@@ -403,8 +408,222 @@ enum VKeyCodes {
...
+	case kFtVkScroll:
+		return Common::KEYCODE_SCROLLOCK;
+	default:
+		warning("Unknown VKEY: %d", vkey);
+		return (Common::KeyCode)(vkey < KEYSTATES_ARRAY_SIZE ? vkey : 0);
+		break;
+	}
+}
+#endif
+
```
bluegr: Overall, there is a lot of duplication here, and this is prone to errors (e.g. adding a case in one function, but not the other). Also, the Foxtail enums are only used here, so IMHO we don't really need them.

You could use a simple list, which would contain the ScummVM keycode and the Foxtail keycode in each entry, e.g:

    struct FoxTailKeyCodeMapping {
        uint32 foxTailKeyCode;
        Common::KeyCode keycode;
    };

        const FoxTailKeyCodeMapping foxTailKeyCodeMapping[] = {
                { 1073741898, Common::KEYCODE_HOME },
                { 1073741882, Common::KEYCODE_F1 }
        };

And then you could do something like:

uint32 keyCodeToFoxTailKeyCode(Common::Event* event) {
	for (int i = 0; i < ARRAYSIZE(foxTailKeyCodeMapping); i++) {
		if (foxTailKeyCodeMapping[i].keycode == event->kbd.keycode)
			return foxTailKeyCodeMapping[i].foxTailKeyCode;
	}

	return event->kbd.ascii;
}

Common::KeyCode foxTailKeyCodeToKeyCode(uint32 vkey) {
	for (int i = 0; i < ARRAYSIZE(foxTailKeyCodeMapping); i++) {
		if (foxTailKeyCodeMapping[i].foxTailKeyCode == vkey)
			return foxTailKeyCodeMapping[i].keycode;
	}

	return Common::KEYCODE_INVALID;
}

engines/wintermute/base/base_game.cpp
lolbot-iichan: Technically, Ctrl+M support for toggling mouse capture is implemented at one backend only (SDL). Also, this implementation is directly inside `SdlEventSource::handleKeyDown()`, it's not something accessible from OSystem accesible to engines. Which means that is would require some refactoring to make `getWindow()->toggleMouseGrab();` callable from engines code.

Meanwhile, `Game.ChangeWindowGrab()` method of FoxTail's Wintermute is used only once, in F11 handler. This hotkey is not mentioned in FoxTail GUI and can only be discovered by reading readme_en.txt file.

Do we really need re-implement F11 hotkey, while we have Ctrl+M on SDL already and while we have nothing on non-SDL anyway? I'd skip it, at least in initial PR of FoxTail subengine.

engines/wintermute/base/font/base_font_bitmap.cpp
lolbot-iichan: What about `custoTrans`? This names is used in 3 files: base/base_frame.cpp, base/base_sub_frame.cpp, base/font/base_font_bitmap.cpp. 

Should I change both names in all files? Changing `custoAlpha` feels inconsistent.

engines/wintermute/base/saveload.cpp
lolbot-iichan: Removed at 48ea8b6.

engines/wintermute/base/saveload.cpp
lolbot-iichan: Removed at 48ea8b6.

engines/wintermute/base/font/base_font_bitmap.cpp
lolbot-iichan: Renamed `custoAlpha` and `custoTrans` at 6b4e06b13ca2d697f8de80f3fab9b493590256e3.

engines/wintermute/base/base_keyboard_state.cpp
lolbot-iichan: Very good idea, thanks for pointing this!

While implementing this suggestion and running various tests I discovered, that difference in key mappings between FoxTail and usual games is not due to FoxTail's engine differences from WME Lite, but were made in WME Lite itself! WME 1.x is using "Virtual-Key Codes" mapping
and WME Lite and FoxTail are using "SDL_Keycode" mapping.

So, here is total refactoring of keycode handling.

Why nobody noticed this earlier? Because there are only 4 currently known series that are based on WME Lite forks (J.U.L.I.A. series, Reversion series, FoxTail and Securanote) and because most WME games, both 1.x-based and Lite-based are using Keyboard class only for handling Enter and Escape keys, which are the same in both mappings (Backspace, Space and Tab are also the same).

engines/wintermute/base/base_game.cpp
bluegr: I do agree that mapping this is outside the scope of this pull request

bluegr:
Excellent! Thanks for fixing all the issues, and many thanks for your work on this.

This can now be merged. Great work

lolbot-iichan:
Thanks for reviewing, suggestions and helping this happen! I'm really happy to have Foxtail subengine at last integrated to master and I had extremely good time today playing FoxTail on Nindendo Switch with nightly build of ScummVM

PR #1320, opened 2018-08-27
MOHAWK: RIVEN: Let credits roll longer

dafioram:
Fixes Trac#10675.

Previously, the credits ended as soon as the last row was shown
of the final credits image.

Now some more black rows (empty rows) are shown and finally a few
seconds of black. I set it to 8 seconds of wait beyond where
the credits where previously stopping.

In order to do this updateCredits was enhanced to work past the end
of the last credits image (and just keep adding empty rows).

The original game shows a black screen for a longer period than this.

The first and last credit image indices were also assigned to #defines.


dafioram:
I think this change is a behavior improvement over the current code, but perhaps there are obvious code mistakes that I have made.

I feel good about it since I have tested it with multiple ends.

dafioram:
Further testing and clean up can be performed in tree.

PR #1323, opened 2018-09-01
CRYO: EDEN: Move graphics into a separate class

dafioram:
Eden game object now contains a graphics object with which to
delegate graphics operations and store states of the graphics.

Before all the graphics and eden code where together.

Much of the video playing is done in the graphics class
so I have moved a lot of the video state into there.

Some graphics related variables were moved out of eden and into
graphics, but many are still in eden.

Since they are still coupled there are lots of getters and setters.
For example both eden_graphics and eden share a handle to the same
video object.

I have made a few more things public than desirable.

I changed graphics to eden_graphics since it is specialized to eden
and not just cryo.

### DIFF
--- a/engines/cryo/eden.cpp
+++ b/engines/cryo/eden.cpp
@@ -42,6 +42,7 @@
 #include "cryo/cryolib.h"
 #include "cryo/eden.h"
 #include "cryo/sound.h"
+#include "cryo/eden_graphics.h"
 
 namespace Cryo {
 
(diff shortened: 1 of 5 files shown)


sev-:
Why did you close it? The changes are fine, and they're worth merging.

dafioram:
It's very possible that bugs have been introduced in this partial refactoring.

bluegr:
Almost a year later...
From a casual review of the code, the changes look fine. Plus, there hasn't been any new activity on the engine since the pull request was opened. What kind of bugs are you referring to?

dafioram:
Possibly run time. I tested it after the refactor, but I didn't play a significant amount of the game.

bluegr:
The engine is not yet complete, but this change is a step towards the right direction, as it provides a better separation of concerns.

I'm merging this, and if any regressions occur, they can be fixed in-tree

PR #1324, opened 2018-09-01
WIN32: Rewrite Inno setup installer script.

lotharsm:
This is a rewrite of the Inno setup installer script.

-  Information like AppVersion and Copyright is now directly fetched from the scummvm.exe. Now we don't have to touch this file in case the ScummVM version increases.

~~- Moved the documentation files in a seperate subfolder and create links to them in the Start Menu to avoid cluttering the Start Menu on Windows 8 and higher.~~

- Added language entries to the installer for all languages supported by translations.scummvm.org.

- Add option to create a ScummVM shortcut to the desktop.

~~- Remove shortcuts to the "Save Games (old default)" folder since current versions of ScummVM won't even install on Windows versions lower than Windows XP.~~

I am a bit unsure what to do with the scummvm.iss.in file since all this seems to do is to contain an additional placeholder for the version numbers. Since this script fetches it directly from the executable, the scummvm.iss.in file shouldn't be needed anymore?


lotharsm:
I wonder if we should also move the default installation directory from {pf} to {userappdata} so we don't need elevated rights anymore.

dists/win32/ScummVM.iss
```diff
@@ -157,95 +163,94 @@ begin
...
+;ScummVM executable and libraries
 Source: SDL2.dll; DestDir: {app}; Flags: replacesameversion
 ;Source: SDL2_net.dll; DestDir: {app}; Flags: replacesameversion
-;Source: WinSparkle.dll; DestDir: {app}; Flags: replacesameversion
+;Source: WinSparkle.dll; DestDir: {app}; Flags: replacesameversion
+
 ;Migration script for saved games in Windows NT4 onwards
 Source: migration.bat; DestDir: {app}; Flags: ignoreversion; MinVersion: 0, 1
-Source: migration.txt; DestDir: {app}; Flags: ignoreversion; MinVersion: 0, 1
+Source: migration.txt; DestDir: {app}; Flags: ignoreversion; MinVersion: 0, 1
```
bluegr: Please readd the newline here

dists/win32/ScummVM.iss
```diff
@@ -157,95 +163,94 @@ begin
...
+Name: galician; MessagesFile: compiler:Languages\Galician.isl
+Name: greek; MessagesFile: compiler:Languages\Greek.isl
+Name: polish; MessagesFile: compiler:Languages\Polish.isl
+Name: ukrainian; MessagesFile: compiler:Languages\Ukrainian.isl
+
+[Tasks]
+Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}" 
+
+[Icons]
+;Optional desktop icon
```
bluegr: Please add a space after the ; for consistency with the existing comments (and the comments in our source code)

dists/win32/ScummVM.iss
```diff
@@ -157,95 +163,94 @@ begin
...
+Name: "{userdesktop}\ScummVM"; Filename: "{app}\scummvm.exe"; Tasks: desktopicon
+
+;Start menu icons
```
bluegr: Same here

dists/win32/ScummVM.iss
```diff
@@ -157,95 +163,94 @@ begin
...
+Name: {group}\Documents; Filename: {app}\Documents; WorkingDir: {app}\Documents; Comment: Documents; IconIndex: 0; Languages: not german;
+Name: {group}\Dokumente; Filename: {app}\Documents; WorkingDir: {app}\Documents; Comment: Dokumente; IconIndex: 0; Languages: german;
 
 [Run]
-Filename: {app}\ScummVM.exe; Flags: nowait skipifdoesntexist postinstall skipifsilent
-;Creates a symbolic link for standard save games area, under Windows Vista/7
-Filename: {cmd}; Parameters: "/c ""mklink /d ""%userprofile%\Saved Games\ScummVM"" ""%appdata%\ScummVM\Saved games"" "" "; MinVersion: 0, 6.1
+Filename: {app}\ScummVM.exe; Flags: nowait skipifdoesntexist postinstall skipifsilent
+
+;Creates a symbolic link for standard save games area, under Windows Vista and higher
```
bluegr: And here too (there are other places below, please adapt accordingly)

bluegr:
Overall, excellent work!
The change to the default installation directory is a valid point, but how will the existing installations be affected? Will the previous uninstaller remove the old files, so that the new installer places them in the new folder?

lotharsm:
Sadly not. Updating an existing folder without uninstalling the old version first will lead to a "duplicate" installation - the old one in Program Files, the new one in AppData. Since the shortcuts are updated by the new installer, this is a minor issue IMHO.

Kirben:
The documentation files are on the Start Menu for easy access, we shouldn't be forcing users to use Windows Explorer for any access.

The migration tools are required for anyone who used old ScummVM version under Windows XP/Vista/7 too. The saved games location changed from '\Program Files\ScummVM' to userappdata in the distant past.

{userappdata} is only meant for application data, applications should never be installed in that area.

Otherwise the other changes seem fine.

lotharsm:
Thanks for your comments!

>The documentation files are on the Start Menu for easy access, we shouldn't be forcing users to use Windows Explorer for any access. 
Even though this looks not really great on Windows 8 and higher?

>The migration tools are required for anyone who used old ScummVM version under Windows XP/Vista/7 too. The saved games location changed from '\Program Files\ScummVM' to userappdata in the distant past.
Okay, but how relevant do you think is an upgrade path from such ancient versions?

>{userappdata} is only meant for application data, applications should never be installed in that area.
Okay, sounds plausible. What about {userpf} which doesn't need elevated rights too?

lotharsm:
The problem for me is that Windows 10 doesn't support sub-folders on the Start Menu anymore, so that's not an option unfortunately. Maybe it's just my personal preference, but having all the document shortcuts sitting directly in the Program group is not very nice...

>It does seem unlikely for people to upgrade from an ScummVM version released over 6 years ago, but at least keep a link to current saved games location, to make it easier for back ups.

Okay, I'll do this then.

>{userappdata} and {userpf} use similar paths, so the later is no better. ScummVM been installed to {pf] means it can be used by all users, and elevated rights exist so people can choose to block installers.

Okay, I'll leave this at {pf}.

lotharsm:
Kirben, I reverted some of my changes according to your comment. Is everything fine now in your opinion or do you have any more suggestions? Thanks for your help with this!

dists/win32/ScummVM.iss
```diff
@@ -1,251 +1,287 @@
...
+EnableDirDoesntExistWarning=false
+Compression=lzma2/ultra
+OutputDir=C:\{#AppName}
+OutputBaseFilename=scummvm-win32
+DisableStartupPrompt=true
+AppendDefaultDirName=false
+SolidCompression=true
+DirExistsWarning=no
+SetupIconFile=graphics\scummvm-install.ico
UninstallDisplayIcon={app}\scummvm.exe
```
Kirben: Shouldn't there be a new line between these two?

Kirben:
Seems fine to merge, other than that small change required.

The win32dist target in ports.mk will need 'scummvm-install.bmp' added too, but I can do that separately.

dists/win32/ScummVM.iss
lotharsm: Thanks, seems that Inno Studio screwed up the encoding - again. Will solve this ASAP obviously.

lotharsm:
Kirben, do you know if the file ScummVM.iss.in is still required?

Kirben:
ScummVM.iss.in file isn't used by Inno Setup but the *.in files seems to be used by one of the build systems.

dists/win32/ScummVM.iss
lotharsm: After investigation, I can say that the file itself is correct, this seems to be an rendering/displaying error in Github.

criezy:
> Now we don't have to touch this file in case the ScummVM version increases.

You are not supposed to touch directly the `ScummVM.iss` when increasing the ScummVM version. Instead it would automatically be updated alongside all the other files that contain the version by running `update-version.pl`.

The  `ScummVM.iss.in` file (as well as other files with the `.in` suffix) is used by `update-version.pl` to generate the `ScummVM.iss` file by substituting the `@VERSION@` with the actual version (it also substitutes `@VER_MAJOR@`, `@VER_MINOR@`, `@VER_PATCH@` and `@VER_EXTRA@`).

If your updated `ScummVM.iss` no longer contains the version, then you should indeed remove the `ScummVM.iss.in` AND update `devtools/update-version.pl` so that it no longer tries to process this file. Our wiki (notably http://wiki.scummvm.org/index.php/HOWTO-Release) will also need to be updated to reflect the change.

lotharsm:
>if your updated ScummVM.iss no longer contains the version,

Correct, the ScummVM.iss script doesn't contain any version information now and just pulls the necessary information right from the executable file.

>then you should indeed remove the ScummVM.iss.in AND update devtools/update-version.pl so that it no longer tries to process this file. Our wiki (notably http://wiki.scummvm.org/index.php/HOWTO-Release) will also need to be updated to reflect the change.

Thanks, I'll do this then.

lotharsm:
Do you consider this ready to be merged now or is there something left to do?

Kirben:
Yes, merged as requested, thank you for the changes.

PR #1325, opened 2018-09-02
MOHAWK: RIVEN: Don't allow GMM saving at the main menu

dafioram:
Currently, the game can be saved as soon as the game has started
at the main menu.

This change prevents that and other times that the menu is opened.
The game can still be saved via the main menu once a game has
been started (not using the gmm).

GMM saving still works where it used to outside of the main menu.


dafioram:
I don't think I can use the isGameStarted since it is always false at the main menu?

If you want the ability to gmm save while at the main menu we can keep that behavior.

PR #1326, opened 2018-09-05
IOS: Hide keyboard when rotating to landscape

jlegeny:
On iOS12 at least, the rotation did not force the keyboard to hide. This
fix will force it to be retracted in landscape and restored when
rotating back to the portrait position.


digitall:
This PR has been open for more than a month, so should have been enough time for review.

Since several IOS users have reported testing this without issues, and it has been incorporated in two of the most recent IOS specific pull requests, I think it is worth merging this fix and then seeing what impact it has in terms of buildbot and other IOS users testing with the nightly builds.

Merging...

PR #1327, opened 2018-09-08
Atan2 table

dafioram:
1. Add a float lookup table for atan2, similar to star trek's atan2 lookup, except that one is for a fixed size table and uses ints. This lookup is faster than atan2().
2. Use the table in titanic star control. I tested it as well.

Making a atan2 table based only on ints may also be useful to add.

### DIFF
--- a/common/atan2table.cpp
+++ b/common/atan2table.cpp
@@ -1,0 +1,91 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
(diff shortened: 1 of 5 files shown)


foobar13372:
Does this improve anything? How much faster is "faster"? LUT's are great on limited hardware that is needing the calculation very often - this isn't the case here. For now I only see increased code complexity that doesn't add anything.

But maybe I'm just not seeing the intention here, that's why I'm asking.

lskovlun:
One problem with this is that in cases where LUTs are actually useful they can easily contain little idiosyncrasies on a per-game basis (round-off errors, edge cases and the like). So you end up having to have one for each game anyway.

dafioram:
@foobar13372 it is 65% faster than the float atan2 call. While titanic may not speed up that much since the atan2 is not used much something like bladerunner does a lot more atan2 so it would benefit more.

@lskovlun I'm not sure I understand, how can an engine do an atan2 function correctly, but not be replaced by a more general atan2 call? I'm thinking of star trek here. We can always select a lower number of points/bits to model less precision to stay true to an engine.

PR #1328, opened 2018-09-09
Fix launch on recent iOS device (iOS 11.4.1)

jonnyijapan:
I just followed the instructions here: http://wiki.scummvm.org/index.php/Compiling_ScummVM/iPhone

But I would get this error at the last step when Xcode was about to launch the app on my iPhone.

screen shot 2018-09-10 at 0 33 23

Until I fixed it this way. There are probably a lot more that could be dealt with in the Info.plist for recent iOS versions, but anyways.

### DIFF
--- a/dists/ios7/Info.plist
+++ b/dists/ios7/Info.plist
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleExecutable</key>
-	<string>ScummVM</string>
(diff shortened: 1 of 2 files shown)


tsoliman:
I had the same issue. Confirmed for an iPad 2018 that this fixes the install/launch problem.

There were so many other issues compared to v1.8.1 on iOS 9 on an iPad mini 1 (e.g. iOS keyboard handling)

Looking forward to more fixes. 👍

PR #1329, opened 2018-09-11
SCI: Script patch for LB2 crate room bug #10701

sluicebox:
Fixes a lockup in the original game

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -3711,6 +3711,56 @@ static const uint16 laurabow2PatchRememberWiredEastDoor[] = {
 	PATCH_END
 };
 
+// The crate room (room 460) in act 5 locks up the game if you enter from the
+//  elevator (room 660), swing the hanging crate, and then attempt to leave
+//  back through the elevator door.
+//
+// The state of the wall crate that blocks the elevator door is tracked by
+//  setting local0 to 1 when you push it out of the way, but Sierra forgot
+//  to reinitialize local0 when you re-enter via the elevator door, causing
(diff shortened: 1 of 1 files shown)


dafioram:
Nice work.

I've tested that this resolves the bug.

Can you update your comments so it includes the acts that this occurs in (Act 5?). Can you brake up your two paragraphs into more so that it is easier to read.  Like after the first sentence, Can you add which rooms this concerns in the comments?

How do you know that this also happens in the original? Viewing the scripts/disassembly or actually playing the game (dosbox)?

sluicebox:
I've updated the comments. Yes, I confirmed the bug in the original in dosbox.

dafioram:
Can you combine your two commits into one (Using rebase interactive or amend)? You will have to force push, but that is fine.

digitall:
@dafioram : This looks good now. OK to merge?

dafioram:
I think it is fine. it might be good to let some time pass in case @bluegr wants a chance to look at this.

digitall:
@dafioram : OK Thanks. Better to do an explicit request for review from him though, rather than just hope he sees this. Will do that. If he is happy / indifferent, we can then merge.

bluegr:
Nice work on the patch!

I really liked the way you used the eq operator here - basically, you add its result (0 or 1).

I reviewed your patch, and tested your changes. It all works fine. Well done, merging

digitall:
@sluicebox : Does this patch fix this bug as well, or is this a different script issue?
https://bugs.scummvm.org/ticket/10702

sluicebox:
That one is different. I haven't looked closely at it but it's in a different room and unrelated.

PR #1332, opened 2018-09-13
Auto-hide/show keyboard for non-iPad only (iPhone) + allow show/hide of keyboard using 4-finger swipes

jonnyijapan:
An extension of this PR: https://github.com/scummvm/scummvm/pull/1326

For iPhone, the auto hide in landscape, auto show in portrait is pretty okay!

But I don't think there is a great need to autoshow/hide the keyboard for iPad - the iPad software keyboard has a hide button. So I fixed that.

Also if the user for whatever reason wants to show/hide the keyboard at will, he/she may do so using four-finger swipes up/down. Hopefully this doesn't clash with anything. I don't think there were any "four finger" gestures in the old scumm games...

Trying this out now myself for a while.


digitall:
@jonnyijapan : Thanks for the code contribution. Have now merged PR #1326 so could you rebase your PR to remove this and the merge commit please? Also, could you modify the commit messages of the other commits to comply with http://wiki.scummvm.org/index.php/Commit_Guidelines#Commit_message_formatting ... "IOS: Added" etc.

jonnyijapan:
@digitall  Ok will fix.

jonnyijapan:
Closing here, new PR at #1358

PR #1333, opened 2018-09-13
LASTEXPRESS: added detection entry for gog release

mgerhardy:
This adds a detection entry for the GOG release of LastExpress.

Questions:
- What about `AD_LISTEND` - should it be used?
- The `CD*` files are in a subdir named `data/` - how to deal with that. Using `data/CD1.HPF` didn't work. I had to copy the files into the same dir as `HD.HPF` to be able to detect the game.

### DIFF
--- a/engines/lastexpress/detection.cpp
+++ b/engines/lastexpress/detection.cpp
@@ -192,6 +192,24 @@ static const ADGameDescription gameDescriptions[] = {
 		GUIO1(GUIO_NOASPECT)
 	},
 
+	// The Last Express (GOG)
+	//   expressw.exe ???
+	//   express.exe  2010-12-14 13:49:04
+	{
+		"lastexpress",
+		"",
+		{
+			{"HD.HPF",  0, "ab940d5815008b176502f759ae753fb7", 30715904},   // 2000-03-01 17:03:58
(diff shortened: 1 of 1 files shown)


digitall:
@mgerhardy : Thanks for this patch. It looks reasonable as-is... at least comparing to the other detection entries.

To answer your two questions; Yes, AD_LISTEND should be used. It appears to have been omitted from the LASTEXPRESS engine detection entries. Since this maps to {0, NULL, 0, NULL} then assuming the detection structures are initialized to 0 which is also NULL, then you would probably get away with this normally... especially as the detection file list is a fixed size array of 14 ... so no pointer or bad access. Just glitches maybe in the detections.

However, I don't think that should block adding this, as that fix should be done to all entries in another commit. I will look at doing that after merging this.

Similarly, the data/CD* files issue should be fixed by adding a search path to the resource loading code found in engines/lastexpress/resource.{cpp, h} as per SearchMan or similar, but that is an issue for further work on the LASTEXPRESS engine. I will look at adding a FIXME to the detection entry to cope, but for now, you / users will have to move / link files instead.

digitall:
Have fixed the missing AD_LISTEND markers in commit f0f02c4a3f1be13058c2334bae694312a315a030

PR #1338, opened 2018-09-14
PRINCE: fixed warnings

mgerhardy:


digitall:
@mgerhardy : Why was this closed? I assume because the proposed patch broke the build in Travis CI?

bonki:
Unfortunately it's not as simple as that and this is not the correct fix (see the corresponding comments I added a while ago).

digitall:
@bonki: Sorry, I didn't follow up my initial comment. I tried compling this myself and it failed badly. I am aware of the issue which produces the warnings and that it is tricky to solve since .c_str() values are not meant to be saved and they are freed... so either you do scumm_strdup() and have memory leakage ... or crashes :/

PR #1340, opened 2018-09-19
SCI: Script patch for LB2 desk lamp, bug #10706

sluicebox:
Fixes a broken message in floppy version

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -3761,6 +3761,33 @@ static const uint16 laurabow2PatchFixCrateRoomEastDoorLockup[] = {
 	PATCH_END
 };
 
+// LB2 Floppy 1.0 attempts to show a non-existent message when using the
+//  carbon paper on the desk lamp in room 550.
+//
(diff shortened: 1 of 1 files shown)


dafioram:
Works well. Good fix.

m-kiewitz:
Very good, thanks for also finding this original game bug.

But please add the following 3 lines to your comments.
// Applies to at least: English floppy
// Responsible method: deskLamp:<noname300>(39), which is really doVerb
// Fixes bug: #10706

I own at least 1 localized floppy version. Will check that one, if it has the same issue.
If that's the case, I can verify if the patch also applies to that one.

m-kiewitz:
// Applies to at least: English floppy
should mention 1.000 too of course. I missed that.

sluicebox:
I've updated the comment.

I have quite a few of the localized floppy versions, possibly all, and you should find that all of them are based on version 1.1 which fixes the bug. Extremely interested if you find otherwise!

bluegr:
Very good work! Thanks for your patch. I've tried it and it works great :)
The floppy version of LB2 did not have a selector dictionary (i.e. dictionary 997), which is why most object methods don't have names.

Merging, thanks!

PR #1342, opened 2018-09-22
SCI: Script patch for LB2 armor inset, bug #10709

sluicebox:
Fixes missing messages in floppy version


dafioram:
Nice work.

Why does the signature get matched as soon as the game is loaded from your save (room 430)? It doesn't happen if 430 is entered via another room, only on start up.

sluicebox:
Ha, I don't know the answer but it's some form of "I messed up". 

That save already has rm440 in memory somehow ("vo rm440" works), which is weird because that's exactly what I was trying to avoid, so... uh... "Oops!"

I just uploaded a second uncontaminated save to the bugtracker.

dafioram:
That looks good.

Were you able to test this change with any localizations and/or floppy 1.0+ (you said that the signature would match with later floppy versions)?

sluicebox:
Yes, I tested with 1.1 en/gr/es (identical script resources) and 1.1 fr (like 1.1 but with a handful of extra localization tweaks and bugfixes)

The script patch, though applied to later versions, doesn't interfere with them because patched 1.1 becomes:

if (insert) insert:handleEvent
else if (joystick event while behind curtain) ...
else if ((event type == move) || inset) super:handleEvent

That way when there is an inset, Sierra's 1.1 bugfix handles it and skips my patch.

dafioram:
Excellent,

Let's wait a week incase anyone else wants to comment.

PR #1343, opened 2018-09-23
SCUMM: MONKEY2: Skip extra code on FMTOWNS, fixes bug #2233

BenCastricum:
The different behavior in FM-TOWNS is caused by a few extra lines of code the is specifically added to move the map piece directly to the tree house. This patch bypassed the extra code causing the FM-TOWNS version behave like the other versions. All other required resources are still there.

### DIFF
--- a/engines/scumm/script_v5.cpp
+++ b/engines/scumm/script_v5.cpp
@@ -420,6 +420,16 @@ void ScummEngine_v5::o5_actorFromPos() {
 void ScummEngine_v5::o5_actorOps() {
 	static const byte convertTable[20] =
 		{ 1, 0, 0, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 20 };
+	// Fix for bug #2233 "MI2 FM-TOWNS: Elaine's mappiece directly flies to treehouse"
+	// There's extra code inserted in script 45 from room 45 that caused that behaviour,
+	// the code below just skips the extra script code.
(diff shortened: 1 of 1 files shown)


digitall:
Link to bug artifact for reference:
https://bugs.scummvm.org/ticket/2233

@BenCastricum : Thanks for this patch. As this is self-contained, it should be fine to merge, though would be good to get a review from SCUMM engine dev. 

One question: Does this change in the scripts look intentional? i.e. Was this done to avoid some script lockup or is this unclear? The bug history does not seem to be clear on that. Apart from that, did you test playing through from this point to make sure this is completable? Thanks again... and I think this should be merged in the absence of any further clarity ... It can always be disabled / removed if it turns out to cause issues in future.

BenCastricum:
I added a diff of the 2 versions of the scripts in the bug report. But if you ask me, the code was added there deliberately and in a very clean way and with clear purpose (skip a part of the game). However that could also mean that they added it just for testing purposes, but forgot to remove it later on.

I did test the patch and checked if it didn't cause any other problems further in the game, but the things I tested all worked as normal.

digitall:
Thanks. Odd, as you say, maybe leftover testing code and you have tested this so I would be happy to merge it. If there is any obscure issue, it will only show up from longer term user testing anyway..

digitall:
Will wait a week in case anyone wants to comment or modify this, but then will merge as-is.

digitall:
@bluegr I had the same thoughts, hence why I asked @BenCastricum to clarify. Given he has tested and this seems fine, I think just merging and seeing if any user feedback indicates any issue over longer term will be the only way... but as he said, this could just be a testing left over.

PR #1344, opened 2018-09-29
SCI: Script patch for LB2 act 4 init, bug #10716

sluicebox:
Fixes a lockup in the original game

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -3761,6 +3761,55 @@ static const uint16 laurabow2PatchFixCrateRoomEastDoorLockup[] = {
 	PATCH_END
 };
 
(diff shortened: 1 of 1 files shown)


sluicebox:
Player performance and those messages are tracked by globals 15 and 131 through 135 and aren't related to global 111 so all of that remains the same. (But that does happen in the same method)

111 only tracks what Yvette's up to so that scenes are initialized correctly, not if you've necessarily seen what she was up to, and so if you miss some things in act 3 then it needs to get fast forwarded to 11 so that act 4 plays out correctly. The game already does this fast forwarding in rm510:init when you end act 3 upstairs, they just forgot to do that when it ends in the basement. Adding the fast forwarding to actBreak:init to cover all cases is what Sierra did in the later versions.

digitall:
@bluegr OK to merge?

bluegr:
Yes, merging

PR #1345, opened 2018-09-30
Add Emulator category to scummvm.desktop

rrigby:
Additional category listed in https://standards.freedesktop.org/menu-spec/latest/apas02.html

### DIFF
--- a/dists/scummvm.desktop
+++ b/dists/scummvm.desktop
@@ -11,5 +11,5 @@ Exec=scummvm
 Icon=scummvm
 Terminal=false
 Type=Application
-Categories=Game;AdventureGame;
+Categories=Game;AdventureGame;Emulator;
 StartupNotify=false


bluegr:
ScummVM is not an emulator. It doesn't emulate a platform, but it's a rewrite of original executable game files. As such, I believe that adding it to an emulation category is wrong

digitall:
Quote from https://en.wikipedia.org/wiki/ScummVM#Features
"While ScummVM appears to function equivalently as a game emulator, the ScummVM team does not consider it as such."

This was based on a 2017 interview with @sev- i.e. https://www.pcgamer.com/uk/how-scummvm-is-keeping-adventure-games-alive-one-old-game-at-a-time/

"ScummVM is not an emulator," confirms Eugene Sandulenko, current Project Lead. "It has been confused a number of times with one, and actually it does contain several, though mostly for sound cards. When it comes to the games though, we don't emulate the originals—we rewrite them. "

@rrigby Looking through the list of categories you linked, we are already declaring Game and AdventureGame. RolePlaying could be added since we now support some RPGs ... but that is also a bit marginal.

Overall, I tend to concur with @bluegr's viewpoint.

rrigby:
@bluegr @digitall I understand. It's unfortunate there's not a GameEngine or Interpreter category, that is why I thought Emulator might fit. Because it's slightly misleading to open the application expecting a game then be prompted to actually provide the game to run on it. I think within the limited category choices Emulator is still a worthwhile addition to make the distinction that this is not a game unto itself, but it is ultimately your call.

dreammaster:
In my opinion, I agree with @bluegr and @digitall . There's little point in trying to pretend it's part of a category it's clearly not.. just because it shares in common the fact that to users that they can select games to play from it.

digitall:
@rrigby Thanks for the suggestion, but you hit a slight "hot button" topic for the project. We really should add an item to the FAQ and Wiki about "Is ScummVM an Emulator?" to avoid this and similar questions in future as it does come up with regularity on the forums for new users.

PR #1346, opened 2018-10-01
SCI: Fix LB2 Yvette/Tut premature murder message

sluicebox:
Fixes wrong message in floppy versions, bug #10723


bluegr:
Looks good, thanks for your work!

PR #1347, opened 2018-10-01
SCI: Fix LB2CD Yvette/Tut murderless message

sluicebox:
Fixes wrong message in cd versions, bug #10724


bluegr:
Looks good. Thanks for your work!

PR #1348, opened 2018-10-02
Change "da_DK" to "da"

scootergrisen:


digitall:
This looks like a pretty simple change, but the impact on the translations site and toolchain will need to be checked.

@criezy: OK to merge, or does this need some update / change to translations site code?

criezy:
Renaming the file should be fine for Weblate, but we need to be carefull when committing to avoid conflicts (make sure there is no pending commit on the weblate server, lock the website before committing). I will take care of it this weekend when I have some time to be around in case something goes wrong.

Also I am not sure changing the `Language` inside the po file will stick. Most po editing tools I have tried automatically set or update the Language field when editing the file, and weblate is one of them (which was the reason why I needed to introduce the `X-Language-name` field when transitioning to weblate instead of abusing the `Language` one to store the user-friendly language as was done before). We can try it though. Maybe weblate is intelligent enough to detect that `da` is a valid language code and will not revert it to `da_DK`...

And the commit message does not follow our [commit guidelines](http://wiki.scummvm.org/index.php/Commit_Guidelines). Either @scootergrisen can update it, or I will make the changes directly without using this PR (they are simple enough).

criezy:
And I forgot to mention that I don't think this will have any impact on the toolchain. We no longer use the `Language` information, and the toolchain uses a `*.po` wildcard to detect available translations.

This may have an impact to properly detect that the system language is available and can be used by default as the file name is compared to the `OSystem::getSystemLanguage()` to do so, but as far as I remember `TranslationManager::setLanguage(const String &)` properly handle the case where the country code is not present.

scootergrisen:
I'm not so good at git so before i figure out how to change the commit message its properly much easier for one of you to make the changes.

criezy:
File renamed in commit 5390c25. I also changed the Language in the file itself, but as indicated before I am not sure this change will stick.

scootergrisen:
Thanks :smile_cat:

PR #1352, opened 2018-10-08
KYRA: Fix Urbish Mines Lobster Monster Crash in LOL.

digitall:
This occurs because of an out of buffer access when trying to play the monster sound effect.

This is from bug Trac #10665.

While this code change is restricted to Kyra's LOL code and thus only that game, and the fix does work to prevent the out of range access, it is not clear that this is the correct solution and will not stop other sound effects from playing. Also, it removes the sound effect from this monster, rather than correcting it to the correct effect... This could do with amendment by someone familar with the KYRA / LOL engine to see if there is a better solution, but if not, then merging this at least prevents LOL from crashing.

### DIFF
--- a/engines/kyra/sound_lol.cpp
+++ b/engines/kyra/sound_lol.cpp
@@ -161,7 +161,7 @@ void LoLEngine::snd_playSoundEffect(int track, int volume) {
 		return;
 
 	_lastSfxTrack = track;
-	if (track == -1 || track >= _ingameSoundIndexSize)
+	if (track == -1 || track >= (_ingameSoundIndexSize - 500) / 2)
 		return;
 
 	volume &= 0xFF;


digitall:
Since it has been more than 2 weeks with no update on this, I am going to merge as-is. This is still better than bad memory accesses causing segmentation fault crashes... and if we get an active Kyra LOL developer, they can take a look at this in future.

athrxx:
Thanks for fixing the invalid memory access. I am not sure right now whether your approach affects sfx playback in a negative way.

After being out of the project for so long it took me some effort to get ScummVM to actually build with MSVC (due to the various library dependencies), but now it seems to work again.

I will try to find a more accurate solution for this issue here if necessary.

digitall:
@athrxx : No problem. Yes, not sure myself but there is something hinky about the sound index table as that check is meant to prevent out of bounds accesses, but the table is not the size indicated i.e. hence the extra 500 bytes needs to be removed and seems like the index size may be in bytes, rather than the 2 bytes (uint16) of the table entries...

Glad you managed to get ScummVM and required libraries built and are still around :)

Yes, happy for more "correct" solution if this is not the root cause. There is more detail of the issue with valgrind traces, savegame to replicate at https://bugs.scummvm.org/ticket/10665

PR #1353, opened 2018-10-09
SCI: Script patch for LB1 input/collision lockups

sluicebox:
Fixes 20+ lockups in the original game, bug #10733


digitall:
Ugh. Mea Culpa. This will need to be manually rebased and committed since it appears the "Resolve Conflicts" ends up with a tangled mess.

bluegr:
I can't reproduce the bug in room 43... can you be a bit more specific on how to reproduce this?

sluicebox:
Sure! I should have included a screenshot from the start. The goal is make this happen:

cb

The key here is you're entering the command while Laura has bumped into an obstacle but is still displayed as walking (leg up). The easiest way to deliberately do that is to slow down the game, walk up to the closet until you bump into it and stop, and then push the arrow key again, which will switch Laura to a walking frame, and then quickly push a letter key to bring up the input box before she goes back to standing.

bluegr:
OK, I managed to trigger the bug, and reviewed the changes. Quite an obscure bug, kudos for finding it out and fixing this! The changes seem harmless enough, and with a quick check there don't seem to be any obvious regressions from this change.

Merging

sluicebox:
It resolves the first two, the third is something else.

I'm really excited about this one, I'd been hunting it for an embarrassingly long time but it always seemed so random and impossible to reproduce. It always happens at the worst times! ...Several programs have been made over the years to flush it out.

PR #1354, opened 2018-10-10
SCI: Script patch for LB1 Lillian/bed, bug #10742

sluicebox:
Fixes wrong and missing messages in DOS version


bluegr:
Nice work! Merging

PR #1356, opened 2018-10-17
ZVISION: Experimental sound support for the high-resolution ZGI movies (DVD version)

eriktorbjorn:
This has been sitting on my hard drive for quite some time now...

Back before clone2727 left the ScummVM project, he and I were trying to get the sound working in the high-resolution ZGI movies. When he left, there was still major problems with the audio/video sync and the project fell by the wayside. When GOG released the DVD version, this became a lot more relevant so I made an effort to update the code and received permission from clone2727 to release it.

There are some caveats:

- It uses liba52 for sound support. I _hope_ this is portable enough for our purposes.
- I've only actually tested the three opening movies (i.e. up to and including the "Propaganda on Parade" movie.)
- The movies use higher resolution than the game itself, but the window remains the same size so movies may be downscaled. This is far from ideal, and it wasn't the case when work began on this.
- It only uses the video timestamps for sync, not the audio ones. This is hopefully good enough. I had a hell of a time getting it to work at all, and ended up having to buffer packets so that they could be handled out of order. Otherwise, it would get out of sync no matter what I did.

There may be other things to consider, but the truth is that I don't have the time nor the knowledge to get it any better than it is now. Hopefully someone else will be able to work out the last kinks. I hope I rebased the code properly. Git is an intelligence test that I keep failing...


eriktorbjorn:
One thing I forgot to mention is that there seems to be some sort of interlacing going on in some parts of the "Propaganda on Parade" intro video. It doesn't really have anything to do with this pull request, but since I don't see it if I play the video in a media player it's something that should be looked into, I guess.

I had hoped Wikipedia's articles on "Telecine" or "Three-two pull down" would give me ideas, but nope. No clue what to do about it.

bluegr:
Nice work from everyone involved!
liba52 looks to be portable (they even say so on their website)

Out of curiosity, why can't we change the screen resolution when a video is played, and we resort to downscaling?

angstsmurf:
Wouldn't the best solution be to always run in the highest resolution needed and upscale the non-video parts?

eriktorbjorn:
> Out of curiosity, why can't we change the screen resolution when a video is played, and we resort to downscaling?

As I said (though perhaps not clearly enough), when work started on this the backend would change the window size when the videos were playing. But at some point, probably related to that you can now resize the game window at will, even with the SDL renderer and not just the OpenGL renderer, this was changed. There's this comment in backends/graphics/sdl/sdl-graphics.cpp:

```
// We only update the actual window when flags change (which usually means
// fullscreen mode is entered/exited), when updates are forced so that we
// do not reset the window size whenever a game makes a call to change the
// size or pixel format of the internal game surface (since a user may have
// resized the game window), or when the launcher is visible (since a user
// may change the scaler, which should reset the window size)
if (!_window->getSDLWindow() || _lastFlags != flags || _overlayVisible || _allowWindowSizeReset) {
```

So it probably has something to do with _allowWindowSizeReset, but I'm not familiar with the inner workings on the backend.

eriktorbjorn:
> Out of curiosity, why can't we change the screen resolution when a video is played, and we resort to downscaling?

As I said (though perhaps not clearly enough), when work started on this the backend would change the window size when the videos were playing. But at some point, probably related to that you can now resize the game window at will, even with the SDL renderer and not just the OpenGL renderer, this was changed. There's this comment in backends/graphics/sdl/sdl-graphics.cpp:

```
// We only update the actual window when flags change (which usually means
// fullscreen mode is entered/exited), when updates are forced so that we
// do not reset the window size whenever a game makes a call to change the
// size or pixel format of the internal game surface (since a user may have
// resized the game window), or when the launcher is visible (since a user
// may change the scaler, which should reset the window size)
if (!_window->getSDLWindow() || _lastFlags != flags || _overlayVisible || _allowWindowSizeReset) {
```

So it probably has something to do with _allowWindowSizeReset, but I'm not familiar with the inner workings on the backend.

eriktorbjorn:
> Out of curiosity, why can't we change the screen resolution when a video is played, and we resort to downscaling?

As I said (though perhaps not clearly enough), when work started on this the backend would change the window size when the videos were playing. But at some point, probably related to that you can now resize the game window at will, even with the SDL renderer and not just the OpenGL renderer, this was changed. There's this comment in backends/graphics/sdl/sdl-graphics.cpp:

```
// We only update the actual window when flags change (which usually means
// fullscreen mode is entered/exited), when updates are forced so that we
// do not reset the window size whenever a game makes a call to change the
// size or pixel format of the internal game surface (since a user may have
// resized the game window), or when the launcher is visible (since a user
// may change the scaler, which should reset the window size)
if (!_window->getSDLWindow() || _lastFlags != flags || _overlayVisible || _allowWindowSizeReset) {
```

So it probably has something to do with _allowWindowSizeReset, but I'm not familiar with the inner workings on the backend.

criezy:
It should indeed run at the highest resolution and upscaled some parts rather than downscale others. I am guessing the zvision engine does not call `initGraphicsModes` when it starts with the list of resolutions it uses. Doing so should allow the backend to choose the best window size.

eriktorbjorn:
> Wouldn't the best solution be to always run in the highest resolution
needed and upscale the non-video parts?

Maybe, but I thought the game graphics looked ugly when I resized the window to 800x600 pixels. And you'd still have the problem where subtitles would look smaller during cutscenes, wouldn't you? So personally I would prefer the window to change size as needed during gameplay like it used to, at least as long as the window hasn't been manually resized. But there's probably no one-size-fits-all solution here...

criezy:
The engines that can change resolution during game play should now call `initGraphicsModes` when starting to indicate all the resolutions that it will use. It is then up to the backend to decide what to do with this information. The SDL backends sets the window size to the highest of the provided resolutions and keeps it at this size all the while the game is running. You can see a discussion in PR #1007 that added the feature. They is indeed no magical behaviour that fits all the cases and all the tastes and we had to settle for what we though was best.

Also at the time the change was made this indeed meant that the parts in lower resolution would be upscaled to fit the bigger window. Since then a stretch mode has been added however, which means that while the window will remain at the higher resolution we can instruct the SDL backend to not scale the display and instead use black borders for the parts at a lower resolution (by select the `Center` stretch mode in the options).

engines/engine.cpp
```diff
@@ -210,6 +210,9 @@ void initCommonGFX() {
 
 		if (gameDomain->contains("filtering"))
 			g_system->setFeatureState(OSystem::kFeatureFilteringMode, ConfMan.getBool("filtering"));
+
+		if (gameDomain->contains("stretch_mode"))
+			g_system->setStretchMode(ConfMan.get("stretch_mode").c_str());
```
digitall: It would be good if this could be merged seperately from the rest of this pull request, since it is a good fix anyway and I think this may be the fix for a problem reported on the forum a few days back:
http://forums.scummvm.org/viewtopic.php?t=14732

criezy: Oops! This is an oversight from myself when I added the feature. But it is indeed unrelated to this PR. I will commit that change this evening when I am back home (unless somebody does it before me).

criezy:
Commit 9fb8e87 is actually not the correct way to handle the game having multiple resolutions, although it probably works with the SDL backend.

The idea is to do something like this when starting the engine (before the first call to `initGraphics`):
```
Graphics::ModeList modes;
modes.push_back(Graphics::Mode(WINDOW_WIDTH, WINDOW_HEIGHT));
if (getGameId() == GID_GRANDINQUISITOR && (getFeatures() & GF_DVD))
	modes.push_back(Graphics::Mode(HIRES_WINDOW_WIDTH, HIRES_WINDOW_HEIGHT));
initGraphicsModes(modes);
```

engines/engine.cpp
digitall: @criezy : Happy for you to commit it later. Though @eriktorbjorn will have to rebase the PR at that point to remove this commit...

criezy:
Good work on this feature!
I finally installed my copy of ZGI from GoG and it was nice to see those high resolution videos with sound :-)

Before merging this you will need to drop the last commit (because I have already pushed it to trunk) and amend the previous one as per my previous comment (to use `initGraphicsModes()` before calling `initScreen()` in `ZVision::initialize()`). Actually I would suggest to also check if we are using the high-res videos or not, so the code would be:
```
Graphics::ModeList modes;
modes.push_back(Graphics::Mode(WINDOW_WIDTH, WINDOW_HEIGHT));
if (getGameId() == GID_GRANDINQUISITOR && (getFeatures() & GF_DVD) && (!ConfMan.hasKey("mpegmovies") || ConfMan.getBool("mpegmovies")))
        modes.push_back(Graphics::Mode(HIRES_WINDOW_WIDTH, HIRES_WINDOW_HEIGHT));
initGraphicsModes(modes);
```

I have made the changes locally to test this PR, and I can force push if you want.

eriktorbjorn:
> I have made the changes locally to test this PR, and I can force push if you want.

If you can do that, that'd be great because I'm short on time right now. There's also a typo ("procude" instead of "produce") in the commit message for commit b326ff6, if that can be fixed.

By the way, if the hi-res graphics mode is only specified if "mpegmovies" is set, what happens if you enable it after the game has begun?

criezy:
You can enable the high-res movies while the game is running? I missed that. If you do that then the window would remain at its small size (unless resized by the user of playing in fullscreen) and the high-res videos would be downscaled. Then indeed it might be better to not check if they are enabled or not when starting the engine and always add the high-res size to the used graphics modes.

bluegr:
> > You can enable the high-res movies while the game is running? I missed that.
> 
> I think so. I can't check at this very moment, but can't you enable/disable them from the game's own options dialog?

I've just checked and no, you can't toggle the ho-res movies while the game is running. You can toggle if movie upscaling will occur via line skipping, or via doubling pixels.

eriktorbjorn:
> I've just checked and no, you can't toggle the hi-res movies while the game is running. You can
> toggle if movie upscaling will occur via line skipping, or via doubling pixels.

My copy of the game has a switch labelled "MPEG" in the lower right corner of the options dialog that switches between the hi-res and lo-res movies.

bluegr:
Ah it has been awhile. I remember now: this option needs to be set in the script manager, otherwise it is disabled and not shown in the preferences dialog. I specifically added code to hide it when libmpeg2 has not been included in the build, and it wasn't included in my local build, which is why I didn't see the option :)

The code in question is in zvision.cpp:222:

    #ifndef USE_MPEG2
    	// libmpeg2 not loaded, disable the MPEG2 movies option
    	_scriptManager->setStateValue(StateKey_MPEGMovies, 2);
    #endif

eriktorbjorn:
> The code in question is in zvision.cpp:222:
> 
> ```
> #ifndef USE_MPEG2
> 	// libmpeg2 not loaded, disable the MPEG2 movies option
> 	_scriptManager->setStateValue(StateKey_MPEGMovies, 2);
> #endif
> ```

Ah, so I guess this should be "if !defined(USE_MPEG2) || !defined(USE_A52)" now? (I'm not sure if there's any such condition for having the option in ScummVM's engine options though.)

eriktorbjorn:
And perhaps the hi-resolution mode should only be set if ScummVM is compiled with the necessary libraries to play the hi-resolution movies.

eriktorbjorn:
> @eriktorbjorn: yes, and yes to both of your questions :P

Hmm... Well, unfortunately I'm still short on time and probably will be fore the rest of the week. I'm uncertain how the aforementioned force push criezy made affects things if I commit to my branch, so I won't do anything at this moment. I see there are a couple of other places where it only checked for USE_MPEG that should probably include USE_A52 as well.

I don't like having such a minor thing hold this up, so uf anyone else can fix it I'd be grateful. Again. :-)

criezy:
I have pushed a commit to check on `USE_A52` everywhere where it was checking `USE_MPEG` and also added that change when initializing the graphics mode.

This seems to work well. The only place where the option is still visible when either libmpeg2 or liba52 is not used is in the engine options. This one is more problematic as the flag is set in the config file when adding the game to ScummVM.

criezy:
I forgot the engine options in the config are updated when starting the engine. So this is not as bad as I though and I have added a commit to remove the mpeg engine option from the detection table when libmpeg2 or liba52 is not used.

This is not perfect as we now get the following behaviour:
1. Add game with version of ScummVM that does not support playing the mpeg movies
The option is not visible in the game edit dialog.
2. Now upgrade to a ScummVM version that supports playing the mpeg movies
The option is still not visible in the game edit dialog.
3. Start the game, then return to launcher.
The option is now visible in the game edit dialog.

But I think this is better than always having the option visible.

PR #1357, opened 2018-10-18
iOS: iOS 12, iPhone X Support

yoshisuga:
- Specify using libc++ when creating the xcode project in iOS 12 
- Support the iPhone X/XR/XS/XS Max screens and adjust the view frame according to the safe area to account for the notch

This PR includes @jlegeny's keyboard rotation fix as well.


digitall:
@yoshisuga : Thanks for the code contribution. Have now merged PR #1326 so could you rebase your PR to remove this and the merge commit please? Also, could you modify the commit messages of the other commits to comply with http://wiki.scummvm.org/index.php/Commit_Guidelines#Commit_message_formatting ... "IOS: Added" etc.

yoshisuga:
Hi! I rebased, squashed my commits and formatted the commit message accordingly.

janisozaur:
Images in this PR are not optimised, their sizes can be reduced significantly.

backends/platform/ios7/ios7_video.mm
```diff
@@ -792,6 +793,24 @@ - (void)initSurface {
 
 	[self setViewTransformation];
 	[self updateMouseCursorScaling];
+    [self adjustViewFrameForSafeArea];
+}
+
+-(void)adjustViewFrameForSafeArea {
+    if ( @available(iOS 11, *) ) {
```
criezy: `@available` is a pretty new addition to the clang compiler, and I suspect its use here will break for example the cross-compilation on our buildbot. Maybe using `if ([[UIApplication sharedApplication] keyWindow] respondsToSelector:@selector(safeAreaInsets)]` would work here instead?

digitall:
@yoshisuga : It would be good if you amended the commit messages to match our standard i.e. IOS: ... rather than iOS: .... and also keep the first line length down to around 80 characters so it does not wrap. See previous link I posted. Otherwise this is looking good... However, I suggest landing PR #1358 first as it is simpler and then doing a minor fixup / rebase of this to rebase on that.

criezy:
OK. Time so see if builedbot likes it or not.
I have squashed and slightly reworded the commits, and pushed to master.

criezy:
The build is fixed now, but I had to disable altogether the safe area code when compiling with a SDK older than iOS 11, even if running on a device that has iOS 11 or above. So to get this feature ScummVM needs to be compiled with SDK 11 or above.

Unfortunately I did not find any way, other than a horrible hack, to get this feature compiled even when using older SDKs so that it is there when running on iOS versions that supports it.

ticky:
Sorry to bump an old thread but is there rationale for not building with a newer SDK at this point? according to Apple’s stats, [93% of iPhones and 84% of iPads are on iOS 12 or newer](https://developer.apple.com/support/app-store/). Would a similar PR be more acceptable now?

Mataniko:
@ticky No reason not to, except there is no dedicated iOS maintainer to handle the updates to the new SDK. If someone wants to take a stab at this, the team will be happy to help them with any questions they have.

criezy:
We still have some users on older iOS versions though, so I am not sure that dropping support for older versions would be a good idea, unless we really have to.

PR #1358, opened 2018-10-20
Auto-hide/show keyboard for non-iPad only (iPhone) + allow show/hide of keyboard using 4-finger swipes

jonnyijapan:
I don't think there is a great need to autoshow/hide the keyboard for iPad - the iPad software keyboard has a hide button. So I fixed that.

Also if the user for whatever reason wants to show/hide the keyboard at will, he/she may do so using four-finger swipes up/down. Hopefully this doesn't clash with anything. I don't think there were any "four finger" gestures in the old scumm games... I've been playing through a couple of old scumm games this way and I think it works well. This means you can use keyboard whenever you want, regardless if the device is in landscape or portrait.

Original PR #1332

### DIFF
--- a/backends/platform/ios7/ios7_video.mm
+++ b/backends/platform/ios7/ios7_video.mm
@@ -887,15 +887,12 @@ uint getSizeNextPOT(uint size) {
 - (void)deviceOrientationChanged:(UIDeviceOrientation)orientation {
 	[self addEvent:InternalEvent(kInputOrientationChanged, orientation, 0)];
 
-	if (!iOS7_isBigDevice()) {
-		// iPad software keyboard has a close button so don't bother controlling iPads.
-		BOOL isLandscape = (self.bounds.size.width > self.bounds.size.height);
-		if (isLandscape) {
-			[_keyboardView hideKeyboard];
-		} else {
-			[_keyboardView showKeyboard];
-		}
-	}
+  BOOL isLandscape = (self.bounds.size.width > self.bounds.size.height);
+  if (isLandscape) {
+    [_keyboardView hideKeyboard];
+  } else {
+    [_keyboardView showKeyboard];
+  }
 }
 
 - (UITouch *)secondTouchOtherTouchThan:(UITouch *)touch in:(NSSet *)set {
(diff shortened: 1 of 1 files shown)


jonnyijapan:
@digitall  Here is the rebased branch with fixed commit messages.

digitall:
Thanks for updating this... Please standby.

digitall:
@criezy : Not a MacOS / iOS user myself; Does this look OK to you? It looks fine as-is in terms of changes so just merge and see what the buildbot makes of it?

criezy:
Now that I have a working iOS development environment again, I am planning to take a look at this and the other iOS PR this weekend.

digitall:
@criezy: Thanks. Good luck on all of these. If you can take a look at the outstanding bugs as well, since I think these PRs may close out a few of them i.e.
https://bugs.scummvm.org/query?status=new&component=Port%3A+iOS&group=component&col=id&col=summary&col=status&col=type&col=priority&col=time&col=changetime&col=game&col=reporter&col=keywords&report=9&order=priority

digitall:
@criezy: Do you have any issue with adding the four finger swipe support as per the last two commits as-is? If this is not an issue, will merge by cherry-pick.

I assume the issue is just with the first commit in this PR and whether the show/hide should be automatic or not. Maybe this might be better as a GUI / config option since it appears to be a user preference?

jonnyijapan:
@criezy well since it's possible to hide the keyboard on iPad (apart from iPhone where there is no native hide button on the keyboard)

But then again I honestly haven't used the iPad for playing a lot and sure I can also agree that having it consistent with iPhone might also make a good choice. So either way is fine with me 👍🏻

criezy:
I would indeed prefer to keep the consistency and auto show/hide the keyboard when rotating the device on both iPhone and iPad. So I would drop the first commit.

The other two commits are fine but I would recommend to squash them since the handlers added in the third commit are referenced in the second one.

jonnyijapan:
@criezy 
Finally got some time to fix this. I'm gonna close this PR. I opened a new one with a fixed branch, just one commit. https://github.com/scummvm/scummvm/pull/1487 Please review

PR #1360, opened 2018-10-22
SCI: Disable LB2 speed test

sluicebox:
Fixes bug #10761


bluegr:
Very nice work! It would be great if we disabled the speed test in other games as well, as with today's systems it's no longer relevant, and an incorrect benchmark could cause speed related issues.

Merging

PR #1362, opened 2018-10-23
SCI32: Fix GK1 missing points for Mosely/veve

sluicebox:
Fixes missing points in english pc floppy version, bug #10763

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -1431,6 +1431,62 @@ static const uint16 gk1CazanouxPathfindingPatch[] = {
 	PATCH_END
 };
 
+// GK1 english pc floppy has a missing-points bug on day 5 in room 240.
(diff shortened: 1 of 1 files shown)


bluegr:
Again, great work, as usual! :)

Same question as in the other pull request: if this bug has already been fixed by Sierra, is there any reason to fix it again ourselves?

m-kiewitz:
I'm not sure about that. I think if it's just some minor bug, which can easily be fixed by us as well, then I don't see a reason why we shouldn't hot-patch it.


For serious bugs, which Sierra fixed and that can't easily be hot-patched by us, we should add an information box right when starting the game for the case that the Sierra patch was not applied. I think for some game(s) we already do this. For example Conquests of the Longbow.

bluegr:
@m-kiewitz: Indeed. I agree on this.

@sluicebox: Great work again! Merging

PR #1363, opened 2018-10-24
SCI32: Fix GK1 honfour keycard lockup

sluicebox:
Fixes lockup in original game, bug #10767

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -1431,6 +1431,34 @@ static const uint16 gk1CazanouxPathfindingPatch[] = {
 	PATCH_END
 };
 
+// GK1 english pc floppy locks up on day 10 in the honfour (room 800) when
+//  using the keycard on an unlocked door's keypad. This is due to mistakenly
+//  calling handsOff instead of handsOn. Sierra fixed this in floppy patch 1.0a
+//  and all other versions.
+//
+// We fix this by changing handsOff to handsOn and passing 0 as the caller
+//  to gkMessager:say since the script disposes itself.
+//
+// Applies to: English PC Floppy only
+// Responsible method: sUnlockDoor:changeState(2)
+// Fixes bug #10767
+static const uint16 gk1HonfourUnlockDoorSignature[] = {
+	0x7c,                           // pushSelf
+	0x81, 0x5b,                     // lag 5b
+	0x4a, SIG_MAGICDWORD,           // send e [ gkMessager:say ... self ]
+	SIG_UINT16(0x000e),
+	0x38, SIG_UINT16(0x0216),       // push 0216 [ handsOff ]
+	SIG_END
+};
+
+static const uint16 gk1HonfourUnlockDoorPatch[] = {
+	0x76,                           // push0
+	0x81, 0x5b,                     // lag 5b
+	0x4a, PATCH_UINT16(0x000e),     // send e [ gkMessager:say ... 0 ]
+	0x38, PATCH_UINT16(0x0217),     // push 0217 [ handsOn ]
+	PATCH_END
+};
+
 //          script, description,                                      signature                         patch
 static const SciScriptPatcherEntry gk1Signatures[] = {
 	{  true,    51, "fix interrogation bug",                       1, gk1InterrogationBugSignature,     gk1InterrogationBugPatch },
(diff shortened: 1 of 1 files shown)


m-kiewitz:
Hi,


I just can't login from work. The patch looks fine, documentation also looks fine, but PATCH_ADDTOOFFSET(+x) should be used instead of copying over all the byte code into the patch data.


Kind Regards

  M. Kiewitz

digitall:
@m-kiewitz : Now that is a feature I haven't seen used before! :)

Thanks for the review. @sluicebox: OK to rework using PATCH_ADDTOOFFSET(+x) ?

m-kiewitz:
PATCH_ADDTOOFFSET(+x) simply adds a number to the patch offset, which lets you basically skip bytes.
The feature can also be useful when an offset/id is used within the script code, which isn't the same for all versions. There is also a feature, that let's you copy over bytes from the script code and allows one to copy them to a different offset. And there is also a feature, that gets a uint16 from the original script code and adds a specified number to it. That let's one to adjust offsets for example.

m-kiewitz:
@sluicebox btw. interested in officially joining our team?

bluegr:
There are four things I wanted to point out here:
- Great work, as usual!
- If this bug has already been fixed by an official Sierra patch, is there any reason to spend time fixing it again?
- PATCH_ADDTOOFFSET helps in a lot of ways, as @m-kiewitz mentioned
- Same question from me, @sluicebox would you like to officially join our team? :)

engines/sci/engine/script_patches.cpp
```diff
@@ -1431,6 +1431,34 @@ static const uint16 gk1CazanouxPathfindingPatch[] = {
...
+//  to gkMessager:say since the script disposes itself.
+//
+// Applies to: English PC Floppy only
+// Responsible method: sUnlockDoor:changeState(2)
+// Fixes bug #10767
+static const uint16 gk1HonfourUnlockDoorSignature[] = {
+	0x7c,                           // pushSelf
+	0x81, 0x5b,                     // lag 5b
+	0x4a, SIG_MAGICDWORD,           // send e [ gkMessager:say ... self ]
+	SIG_UINT16(0x000e),
```
bluegr: Nitpicking: the above two lines should be merged in one, since they form the same command, like so:

```suggestion
	0x4a, SIG_MAGICDWORD, SIG_UINT16(0x000e), // send e [ gkMessager:say ... self ]
```

sluicebox:
Wow, a lot going on on this one!

I would be honored to join!! Keeping Sierra games running is the only legitimate use of computers anymore.

We're on the same page regarding bugs Sierra fixed. I'm mentioning the floppy patches to avoid confusion in case some power developer is testing with a floppy version that they've applied an official patch to. In my experience it's unlikely that a normal user has gone and applied external Sierra patches, assuming they even knew that was possible. It's more realistic to just say "fixed in CD version" in which case I consider all fixable takes-you-out-of-the-game bugs in floppy versions fair game, fixed in CD or not, like lock-ups, missing points (so frustrating), and error messages. That said, I can read a room, and I'll see if I can rustle up an unpatched one next =)

I like PATCH_ADDTOOFFSET but I'm interested to see how this one could be rewritten use that. I don't think any bytes remain in their original position plus some of the re-ordered instructions are optimized (like pushi 0005 => pushi 05). I'm only looking at existing patches for examples though so I'm sure I'm missing some of the script patcher's power. (btw thanks for making such a cool script patcher!)

m-kiewitz:
@sluicebox 
Then I guess welcome on board?
I guess you will need to contact @sev- for IRC credentials (please get yourself a cloak as well, sev can get you into contact, that way your IP address won't be visible to other people) plus ability to commit for yourself and so on.

Maybe I'm a bit confused, but it seems this script patch changes the first byte/opcode and the last opcode (byte + uint16) and everything else stays the same? So one could do a PATCH_ADDTOOFFSET(+5) to skip all of those unmodified bytes.

Fully agreed on that external patching. Most users do not know and my opinion is that they shouldn't need to know. The game should simply run properly with as many game bugs fixed as possible and no user actions required.

Fun fact about that patcher code - it's based on my x86 VM patcher, which was capable of fixing quite a few run-time errors in old DOS programs on-the-fly as well, that happened to occur when the computer was too fast for the run-time library. For example "Run Time Error 200 (Divide by Zero)". Really useful especially for cases where the .EXE is compressed and can't be uncompressed and thus file patching is not possible.

The code was re-used and adjusted accordingly after we originally put in extra C++ code for every single fix of script bugs, which got really messy pretty fast. This here also means that people can extract the patches and then apply them manually to their games / simply dump the patched script, so that the bug is fixed even when using the original interpreter. Plus one patch can cover 5 game versions all at the same time, which a file patch can't unless the script is exactly the same for all versions (major issue with some fan patches, some fan patches even break certain other versions of games).

For script patching sometimes there aren't enough bytes available to add additional script code even when using all sorts of funky script optimization tricks.
I thought about ways to add additional script code space. Maybe one could add a few bytes at the end of the script, but it gets a bit complicated, because script sizes are not the same quite often for different game versions, so jumps would have to get adjusted accordingly and of course I would prefer some kind of generic solution, that doesn't require manual adjustments. The signatures + patch data right now are also not fit for that job. There would also be a way adding another segment, but that's also problematic because of the way SCI handles segments.

Script patches are always meant to be fully optional. What this means is that the game should work fine even after applying a patch, running the game, saving, then removing the patch and restoring the saved game. This is why adding an additional code segment wouldn't work out. We would break saved games.

sluicebox:
I see how this patch looks that way but it's more complicated. It inserts 13 new bytes to award the points, which is a tall order that looked hopeless, but I got lucky in how the script is structured:

```
(if (isFlag 27)
	(GKEgo getPoints: -999 2) ; the new line
	(setFlag 26)
	(GkMessager say: 17 16 56 0 self) ; "Really? So that's what Hartridge used..."
else
	(GkMessager say: 17 16 55 0 self) ; "Really? So this is the whole pattern..."
)
```

Since both code paths end in a similar call with only one different parameter I moved the initial common pushes for setting up the say call in front of the flag test to buy a lot of bytes. In the process I optimized one of those pushes to save another byte. Then I optimized the bnt to save another byte and changed the destination to the push 55 after the patch and complete the say call. After the new instructions award the points I push 56 for that code path's cond and also jump into the middle of that second say (after the push 55) and let it complete the call.

It's clearer if you can see the instructions after the patch, but I know it's still complicated.

sluicebox:
That's great that your patching technique goes that far back. Patching Sierra instructions is a lot more fun for me than x86 was.

PR #1364, opened 2018-10-25
Fix Dreamcast crt0 link order

tsowell:
This fixes a regression that has been preventing the Dreamcast builds from booting for the past couple of months.  Only SCUMMVM.BIN (not scummvm.elf) was affected.  The problem and fix are described in more detail in the commit messages, but please let me know if you have any questions.

### DIFF
--- a/configure
+++ b/configure
@@ -5037,7 +5037,7 @@ else
 int main(void) { setlocale(LC_ALL, ""); return 0; }
 EOF
 		_detectlang=no
-		cc_check $LDFLAGS $CXXFLAGS && _detectlang=yes
+		cc_check && _detectlang=yes
 	fi
 
 	define_in_config_h_if_yes $_detectlang 'USE_DETECTLANG'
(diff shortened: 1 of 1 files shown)


digitall:
@tsowell : Looks good to me. As you indicate in the first commit, the CFLAGS and LDFLAGS are not necessary in translation cc_check as they are already here: https://github.com/scummvm/scummvm/blob/174b0d18643a6fc14135157ec7c0e4e2deaeeb20/configure#L285

The second commit looks fine as well. Will merge shortly...

digitall:
@tsowell : Thanks. Ah cool that sounds very good. Well, the latest libronin is v0.6 (I think our buildbot is still on v0.5). See http://dcevolution.sourceforge.net/index.php?id=libronin

The source code is in a git repo: https://bitbucket.org/ihi/libronin/src , so you can clone / fork, branch and add your change commits.

You can either create a bitbucket account to do a git clone, branch, work and pull request there... or if you don't want to create an account, you can clone from there and push to a repo on your github account ... and then drop the libronin devs a note by e-mail or similar to pull from a specific branch.

As for the ScummVM side work / changes, feel free to add branches to your scummvm clone and push public ... even if not yet ready for PR :)

PR #1365, opened 2018-10-28
SCI32: Fix GK1 ego speed bugs

sluicebox:
Fixes ego speed bugs in the original game, bug #10780

I made two versions of the police station patch because there are two significantly different versions of that room's script. Please take a look and see if they can be rewritten as one that applies to both. The big obstacle is that in the second version the compiler switched from a 2 byte jmp instruction to a 3 byte one right in the middle of the section being patched.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -125,6 +125,8 @@ static const char *const selectorNameTable[] = {
 	"test",         // Torin
 	"get",          // Torin, GK1
 	"has",          // GK1
+	"newRoom",      // GK1
+	"normalize",    // GK1
 	"set",          // Torin
 	"clear",        // Torin
 	"masterVolume", // SCI2 master volume reset
@@ -198,6 +200,8 @@ enum ScriptPatcherSelectors {
 	SELECTOR_test,
 	SELECTOR_get,
 	SELECTOR_has,
+	SELECTOR_newRoom,
+	SELECTOR_normalize,
 	SELECTOR_set,
 	SELECTOR_clear,
 	SELECTOR_masterVolume,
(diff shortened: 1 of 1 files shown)


engines/sci/engine/script_patches.cpp
```diff
@@ -1515,6 +1515,134 @@ static const uint16 gk1Day5MoselyVevePointsPatch[] = {
...
+
+static const uint16 gk1PoliceEgoSpeedFixV1Patch[] = {
+	0x31, 0x1b,                     // bnt 1b [ state 5 ]
+	SIG_ADDTOOFFSET(+19),
+	0x81, 0x00,                     // lag 0
+	0x4a, PATCH_UINT16(0x0006),     // send 6 [ GKEgo:setPri: -1 ]
+	0x3a,                           // toss
+	0x48,                           // ret
+	0x33, 0x00,                     // jmp 0 [ waste 2 bytes ]
+	0x38, PATCH_UINT16(0x0300),     // pushi 0300 [ normalize ]
```
m-kiewitz: Please use PATCH_SELECTOR16() for patches like this.
Otherwise the value is hardcoded and if there is a version that doesn't use the same values, it won't work.

engines/sci/engine/script_patches.cpp
```diff
@@ -1515,6 +1515,134 @@ static const uint16 gk1Day5MoselyVevePointsPatch[] = {
...
+//  Sierra's 1.0b patch and the second is in cd and localized floppies. The
+//  script was restructured and the compiler used different sized instructions.
+//
+// Applies to: All PC Floppy and CD versions. TODO: Test Mac, should apply
+// Responsible method: sGabeEnters:changeState
+// Fixes bug #10780
+static const uint16 gk1PoliceEgoSpeedFixV1Signature[] = {
+	0x31, 0x1f,                     // bnt 1f [ state 5 ]
+	SIG_ADDTOOFFSET(+19),
+	0x38, SIG_UINT16(0x0143),       // pushi  [ ignoreActors ]
```
m-kiewitz: There is also SIG_SELECTOR16(), please use that too.

If you use a selector for patching that is not yet available, please add it to the 2 tables at the start of this .cpp file.

engines/sci/engine/script_patches.cpp
```diff
@@ -1515,6 +1515,134 @@ static const uint16 gk1Day5MoselyVevePointsPatch[] = {
...
+	SIG_ADDTOOFFSET(+60),
+	SIG_MAGICDWORD,
+	0x32, SIG_UINT16(0x0012),       // jmp 12 [ end of method ]
+	0x3c,                           // dup
+	0x35, 0x01,                     // ldi 1
+	0x1a,                           // eq?
+	0x31, 0x0c,                     // bnt c [ end of method ]
+	0x38, SIG_UINT16(0x0211),       // pushi 0211
+	0x78,                           // push1
+	0x38, SIG_UINT16(0x00c8),       // pushi 00c8
```
m-kiewitz: 0x00c8 is probably newRoom, please use SIG_SELECTOR16 as well.

engines/sci/engine/script_patches.cpp
```diff
@@ -1515,6 +1515,134 @@ static const uint16 gk1Day5MoselyVevePointsPatch[] = {
...
+	0x31, 0x3d,                     // bnt 3d [ state 1 ]
+	PATCH_ADDTOOFFSET(+60),
+	0x48,                           // ret
+	0x38, PATCH_UINT16(0x0300),     // pushi 0300 [ normalize ]
+	0x76,                           // push0
+	0x81, 0x00,                     // lag 0
+	0x4a, PATCH_UINT16(0x0004),     // send 4 [ GKEgo:normalize ]
+	0x38, PATCH_UINT16(0x0211),     // pushi 0211
+	0x78,                           // push1
+	0x38, PATCH_UINT16(0x00c8),     // pushi 00c8
```
m-kiewitz: same here, PATCH_SELECTOR16(newRoom) etc.

sluicebox:
Thanks, forgot I'm not roughing it in Dagger of Amon Ra anymore =)

bluegr:
Nice catch! And great work, as always! :)

Thanks for making the changes requested by @m-kiewitz. I believe this is can be merged now.

PR #1366, opened 2018-10-28
SCI32: Remove GK1 French/Spanish CD SysLogger hotkey

sluicebox:
Removes a problematic active debug hotkey, bug #10781


bluegr:
Nice catch! Merging

PR #1367, opened 2018-10-28
SCI32: Fix GK1 floppy Magentia missing message

sluicebox:
Fixes a missing message error in floppy version. bug #10782


bluegr:
Nice work, merging as-is

PR #1370, opened 2018-10-31
Add support for A Gaspocket Adventure: Aliens Ate My Cookies Demo

DevanWolf:
A demo for a canceled HE game has left buried then was released in 2016 but ScummVM has never supported the game files, but we're trying to add support for the demo so it can be added normally (but without this patch you can't add the game unless you rename "gaspocketdemo" to something like "pjgames").


digitall:
@DevanWolf : Thanks for the code patches. Your commits seem to touch more than just adding support for that demo. Can you link to a copy of the demo for testing purposes please? Also, the commit messages are not compliant with https://wiki.scummvm.org/index.php/Commit_Guidelines#Commit_message_formatting, specifically they could do with "SCUMM:" or "SCUMM: HE:" prefix. I can still look at merging this manually and fixing up the commit messages, but if you want to rebase the PR branch to fix you can...

criezy:
Where the changes related to Putt-Putt Goes to the Moon included in this pull request on purpose?

Also I am not familiar with the SCUMM engine detection code, but my understanding is that the `engines/scumm/scumm-md5.h` file is generated automatically from `devtools/scumm-md5.txt` but it looks like you edited it manually instead (or forgot to include the changes to `devtools/scumm-md5.txt`).

DevanWolf:
> Where the changes related to Putt-Putt Goes to the Moon included in this pull request on purpose?
> 
> Also I am not familiar with the SCUMM engine detection code, but my understanding is that the `engines/scumm/scumm-md5.h` file is generated automatically from `devtools/scumm-md5.txt` but it looks like you edited it manually instead (or forgot to include the changes to `devtools/scumm-md5.txt`).

It was to fix the flag that Putt-Putt Goes to the Moon will use GM since it uses MT-32 instead of GM even though they are GM (as per [here](http://bugs.scummvm.org/ticket/6730))

Kirben:
Direct support for A Gaspocket Adventure: Aliens Ate My Cookies Demo was not added in the past, since it was a leaked demo.

The setupcd program included with the original DOS version of  Putt-Putt Goes to the Moon doesn't offer any choice for General MIDI or Roland MT32, so it can't be shown what the game was meant for either.

dreammaster:
If that's the case, then I agree that it's unfortunate, but IMHO we shouldn't support it. Though some form of detection entry should at least be added, with a ADGF_PIRATED flag, and a note on it's origins. That way if the question of support comes up again, it'll be easier to know that it's intentionally unsupported.

DevanWolf:
According to http://forums.scummvm.org/viewtopic.php?t=14008 lists the following problems:

-There's notable audio stuttering during the second half of the intro.bik video.
-On most screens the cursor is given wrong colors.
-The debugging features such as pressing G to enter the room selection menu work even when WhosABigCheater is not set to "I".

Also, pressing Ctrl in the Jessie's bedroom room that will show up: "scumm time:0 burn time:0" still works, but fails to display when running in the original executable. It likely needs some investigation.

DevanWolf:
Bump

dreammaster:
Since no-one else has chimed in, I'm going reject and close this. For several reasons, firstly being that the commits don't meet our guidelines, the ambiguity about changing autogenerated files, and finally because it's a leaked demo, so effectively piracy.

Mataniko:
Sorry I'm late on this but I do want to chime in.

There are two reasons to rejecting this PR:
1) It should be atomic, and only deal with Gaspocket without any other changes. It also needs to be cleaned up to meet project guidelines. (Commit messages are one example)

2) We don't condone piracy, and as dreammaster said, this is a gray area.  

However, I'm of the opinion that we are first and foremost a preservation effort for these old games. A user could legitimately have a copy of this demo, for example by working at HE at the time. I don't think that supporting the demo encourages it's distribution in the same manner that supporting any other demo/game encourages illegitimate distribution.

DevanWolf:
WHAT!? But that was a good demo! We need this to run properly in ScummVM! Without this mod it will not run!

PR #1372, opened 2018-10-31
SCI32: Fix GK1 day 1 Grace phone speech timing

sluicebox:
Fixes timing that cuts off speech. bug #10787


bluegr:
Nice work! Merging

PR #1374, opened 2018-11-02
SCI32: Fix GK1 day 5 snake attack problems

sluicebox:
Fixes audio, timing, and graphics problems in the original. bug #10793


bluegr:
Nice work! Merging

PR #1376, opened 2018-11-03
GROOVIE: Fix GMM loads, block saves in a puzzle

scott-t:
As per discussion in #1306. This patch blocks GMM saves within a puzzle (we can't guarantee valid save state) and handles properly loading by properly jumping back into the base game script.


digitall:
@scott-t : This looks good to me and since you are a GROOVIE expert, merging :)

PR #1377, opened 2018-11-03
GROOVIE: Fix Spaces in Savegame Names

digitall:
This is a patch originally submitted as part of Trac bug 6276 with minor fixup:
https://bugs.scummvm.org/ticket/6276

### DIFF
--- a/engines/groovie/saveload.cpp
+++ b/engines/groovie/saveload.cpp
@@ -128,6 +128,7 @@ Common::InSaveFile *SaveLoad::openForLoading(const Common::String &target, int s
 				case 0:
 					break;
 				case 16: // @
+				case 254: // . (generated when pressing space)
 					c = ' ';
 					break;
 				case 244: // $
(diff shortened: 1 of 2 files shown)


dafioram:
I looked into this at one point, but stopped since I couldn't reconcile the gmm and ouiji saved game name.

If I remember correctly either the gmm saved name needed to be more restrictive like the in game does (a limited character set) or the in game be made to be more general.

The latter seems like a lot of work to make spaces and a few special characters possible.

digitall:
Well, this is mainly just to fix spaces... which are possible in the original dialogs using the keyboard input. The on-screen Oujia Board keyboard has a restricted character set with no spacebar visible ... but the keyboard input on that screen does accept spacebar :/ ... so this just fixes that. Fixing the GMM savenames when using Unicode names is out of scope! ;-)

scott-t:
The game itself with it's custom font can (probably) only handle 0-9 A-Z from memory, so it's more than just restricting the larger Unicode set.

Assuming it does as it says on the tin, the changes look like they are fine.

digitall:
@scott-t : OK, thanks for checking them over. Merging...

PR #1382, opened 2018-11-06
SCI32: Fix QFG4 crest bookshelf HAND interaction

Vhati:
Prevents an incorrect message in the original game, bug [#10756](https://bugs.scummvm.org/ticket/10756)

One can enter the patched room (663) via a debugger teleport, the bat-infested stairway (632), or the EXIT bookshelf (661). Clicking HAND on the crest will open the bookshelf's secret passage, and hero will automatically walk to the exit.

Originally, the player could linger and interact with the bookshelf in its open state, for which the HAND message would be inaccurate.


digitall:
@Vhati: Thanks for the QFG4 patch. Please standby. It might take a while before a SCI engine developer has time to review this.

m-kiewitz:
> I _think_ qg4CrestBookshelfMotionPatch/qg4CrestBookshelfMotionFloppyPatch can be combined, but it involves a macro I haven't used yet, so I may be wrong. The patches differ by one byte whose value is +1 in floppy, and their signatures also differ in one byte that's +1 in floppy, so I _think_ PATCH_GETORIGINALBYTEADJUST can be used.

Correct. That's what it was meant for.

Vhati:
@sluicebox:
> I think qg4CrestBookshelfMotionPatch/qg4CrestBookshelfMotionFloppyPatch can be combined [...] PATCH_GETORIGINALBYTEADJUST

Would it be prudent to blindly offset the class number though? I erred on the side of specificity (known value, known replacement) because it'd break any edition that didn't fit the assumed class table order. The downside of the adjust macro is forfeiting certainty (a whitelist of 1) that the original value is something you expected and would want to adjust exactly that way.

Had the class definitions for MoveTo (script 64992) and PolyPath (script 64945) been written in tandem in the same script, that would've been reassuring. I guess in this case, their nearness in purpose and script number - and the limited pool of editions - ought to be reassuring enough. I'll combine the motion patches.

	English Floppy  English CD
	--------------  ----------
	0x08 Event      0x08 Event
	                0x09 Cue
	0x09 Array      0x0a Array
	...             ...
	0x1b Motion     0x1c Motion
	0x1c MoveTo     0x1d MoveTo
	0x21 Polygon    0x22 Polygon
	0x22 PolyPath   0x23 PolyPath
	...             ...
	0x33 DText      0x35 DText
	                0x36 DButtoon
	                0x37 DEdit
	0x36 DIcon      0x38 DIcon
	0x38 Plane      0x39 Plane  (Even w/ same order increment from prev entry can differ)
	                0x3a DSelector
	                0x3b SelectorDText
	                0x3c FileSelector
	                0x3d Dialog
	                0x3e Controls
	0x3f IconI      0x40 IconI
	...             ...
	0x4e Sounds     0x4f Sounds
	0x4f Cue                    (Floppy's cue is way down here)
	0x50 Game       0x50 Game

> qg4CrestBookshelfSignature and qg4CrestBookshelfFloppySignature only differ in the offset for the Polygon class in the first instruction.

I don't know how to find sLeaveSecretly at runtime, or how to predict it at patch time. That's why signatures are deciding the location.

The sigs are extended as far back as Polygon just to *have* a distinguishing byte. A wildcard would defeat the point of the specificity. If an inappropriate location is written for the edition being played, it will crash the game.

Compared to MoveTo vs PolyPath, tying a random edition's Polygon class number to the code layout of that room is far more reckless. I should check the German floppy...

Vhati:
/Catching up with master and combining motion patches.

	git fetch upstream
	git rebase upstream/master
	[edit away the conflict from subsequent commits in master]
	git add "engines/sci/engine/script_patches.cpp"
	git rebase --continue
	[write the unified motion patch, test, commit]
	git rebase -i HEAD~2  # squash my bookshelf commits
	git push origin +qfg4_bookshelf

sluicebox:
An anthology CD showed up in the mail so now I can follow along!

In my opinion this is too significant of a change to the game's behavior. You can normally click the crest to open the secret passage, and then click it again to close it. This patch forces entering the passage when opening it, meaning the passage can never be closed by the user. The script sCloseSecretDoor can no longer be run. There's a strong expectation of original behavior from fans and purists wouldn't be okay with removing a part of the game, even a seemingly minor one.

If I'm understanding this right, there's a slightly inaccurate message box when clicking Hand on the open bookshelf, so this patch solves that by banning being in the room with the open bookshelf.

Doesn't this also happen in room 662? Looks like it also has a bookshelf secret passage, and its hand message contains "If there's a secret passage here, the trigger is somewhere else in the room".

Unless there's a better message they intended to show, I don't know what can be done about this one.

Vhati:
@sluicebox:
> This patch forces entering the passage when opening it

There are 3 bookshelf passages. Candle (662), Crest (663), and EXIT (661).

The other two automatically exit the moment they are triggered. The crest-driven one is an outlier.

sluicebox:
Yikes, so 662 has a sCloseSecretDoor script too and code to run it if you try to close the door but that can never happen since they force you into the passage.

It looks like they first wrote code for both rooms to let you toggle the passage, then changed it to automatically walk in, but forgot to do that in both rooms and didn't delete the dead code. I like when we can reconstruct events from the crime scene.

In that case I'm okay with the behavior change. Could you please add a comment that this makes this room consistent with the others?

The signatures/patches look like they could be trimmed a bit. toss/ret at the end of the sigs/patches can be tossed since they don't change anything and don't disambiguate.

Since the first 14 bytes of the bookshelf signatures don't get changed by the patch, only those that disambiguate the position are needed The sigs could be reduced to:

```
SIG_MAGICDWORD,
0x4a, SIG_UINT16(0x003e),           // send 3e
0x36,                               // push
SIG_ADDTOOFFSET(+5),
0x38, SIG_SELECTOR16(handsOn),      // pushi handsOn (begin clobbering)
...
0x54, SIG_UINT16(0x0004),           // self 04
SIG_END
```

The motion signature can be reduced by half since that's the only MoveTo 29 151 in the script:

```
0x51, SIG_ADDTOOFFSET(+1),          // class MoveTo
0x36,                               // push
SIG_MAGICDWORD,
0x39, 0x1d,                         // pushi x = 29d
0x38, SIG_UINT16(0x0097),           // pushi y = 151d
SIG_END
```

Could you put "CD" in the variable and description names of the CD-only versions?

Sorry I forgot about github review tools until *after* I got done typing!

Vhati:
@sluicebox:
> I'm okay with the behavior change.

\*grin\*
.
.
> comment that this makes this room consistent
>
> The [bookshelf] sigs could be reduced
>
> motion signature can be reduced by half
>
> Could you put "CD" in the variable and description names of the CD-only versions?

Done.  Thanks for streamlining.

Having dropped Polygon, I needed something to distinguish CD from floppy sigs to choose their respective lofsa value. I changed handsOn from SELECTOR16 to UINT16.

m-kiewitz:
> Having dropped Polygon, I needed something to distinguish CD from floppy sigs to choose their respective lofsa value. I changed handsOn from SELECTOR16 to UINT16.

We could do this in a different way.
Patches can be enabled or disabled by code.
You could (or I could) add a tiny bit of code, that enables the correct patch depending on if it's a CD version running or a floppy version).

To have a patch disabled by default, simply change the first value of the SciScriptPatcherEntry to "false".

That way we are still using the SELECTOR code.

Vhati:
@m-kiewitz:
> Patches can be enabled or disabled by code.
> [...]
> Please use the enablePatch() method.
> That code also already detects CD versions, so it should be pretty easy for you to add that code for QfG4 in case you have some C++/C knowledge)

Ooh. Excellent!
I'll see what I can do.

Vhati:
Added dynamic toggling.

Edited the patch table's description, changing its disambiguating edition into a prefix,
in line with a "CD: fix a thing (1/n)" format.

engines/sci/engine/script_patches.cpp
```diff
@@ -9557,6 +9666,13 @@ void ScriptPatcher::processScript(uint16 scriptNr, SciSpan<byte> scriptData) {
 					enablePatch(signatureTable, "CD: audio + text support");
 				}
 				break;
+			case GID_QFG4:
+				// Chooses between similar signatures that patch with a different lofsa address
+				if (g_sci->isCD()) {
+					enablePatch(signatureTable, "CD: fix crest bookshelf");
+				} else {
+					enablePatch(signatureTable, "Floppy: fix crest bookshelf");
+				}
```
m-kiewitz: Please add a break here.

Vhati: Oh right, switch case. Fixed.

PR #1383, opened 2018-11-06
SCI32: Fix QFG4 autosave deleting manual saves

Vhati:
Prevents the autosave script from enforcing a 20 savegame limit, bug [#10758](https://bugs.scummvm.org/ticket/10758)

To trigger it reliably for testing, I enable autosave inside the adventurers' guild and walk outside.

    // When autosave is enabled, Glory::save() (script 0) deletes savegame files in
    // a loop: while disk space is insufficient for a new save, or while there are
    // 20+ saves.
    [...]
    // No good can come of this loop. So we skip it entirely. If the disk truly is
    // out of space, a message box will complain, and the player can delete saves
    // voluntarily.

I haven't caused that circumstance to confirm. SCI Companion says it would create a Print object (apparently an in-game dialog), with the message "Save Error".

    // Note: Glory::save() contains another space freeing loop, but it might be
    // unreachable.

That other loop only checks free space and is enclosed within "(if (>= temp10 0)" shortly after initializing temp10 to -1.
.
.
I was a little concerned about GetSaveFiles() (up to 99 results) getting dumped into an IntArray declared with capacity 21. That would've had implications for the other slot cap removal patches, because their script 64990 had a similar array. No problem there. it seems to dynamically resize.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -7140,8 +7140,48 @@ static const uint16 qg4InnPathfindingPatch[] = {
 	PATCH_END
 };
 
+// When autosave is enabled, Glory::save() (script 0) deletes savegame files in
+// a loop: while disk space is insufficient for a new save, or while there are
+// 20+ saves. Since ScummVM handles slots differently and allows far more
+// slots, this deletes all but the most recent 19 manual saves, merely by
+// walking from room to room!
+//
+// Ironically, kGetSaveFiles() (kfile.cpp) and the debugger's 'list_files'
+// command rely on listSavegames() (file.cpp), which specifically omits the
+// autosave slot, so the script will only ever delete manual saves. And the
+// space check doesn't take into account the reduced demand when overwriting an
+// existing autosave.
+//
+// No good can come of this loop. So we skip it entirely. If the disk truly is
+// out of space, a message box will complain, and the player can delete saves
+// voluntarily.
+//
+// Note: Glory::save() contains another space freeing loop, but it might be
+// unreachable.
+//
+// Applies to at least: English CD, English floppy
+// Responsible method: Glory::save()
+// Fixes bug: #10758
+static const uint16 qg4AutosaveSignature[] = {
+	SIG_MAGICDWORD,
+	0x43, 0x3f, SIG_UINT16(0x0002),     // callk CheckFreeSpace[3f], 02
+	0x18,                               // not
+	0x2f, 0x05,                         // bt 05 [skip other OR condition]
+	0x8d, 0x09,                         // lst temp[9] (savegame file count)
+	0x35, 0x14,                         // ldi 20d
+	0x20,                               // ge?
+	0x30, PATCH_UINT16(0x0038),         // bnt [end the loop]
+	SIG_END
+};
+
+static const uint16 qg4AutosavePatch[] = {
+	0x32, PATCH_UINT16(0x0044),         // jmp [end the loop]
+	PATCH_END
+};
+
 //          script, description,                                     signature                      patch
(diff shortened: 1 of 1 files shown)


digitall:
@Vhati: Thanks for this. We will either need you to rebase and squash those two secondary commits ... or we can do it by "Squash and Merge" which I think takes the commit message of the first commit... but better if you can do this. If you want to do this, you do it on your branch qfg4_autosave on your fork and the Pull Request should track it :) See http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

digitall:
@Vhati: Thanks for sorting that. Looking good now. As normal, please standby as may take a while for a SCI engine developer to get time to review.

engines/sci/engine/script_patches.cpp
```diff
@@ -7140,8 +7140,51 @@ static const uint16 qg4InnPathfindingPatch[] = {
...
+	0x30, PATCH_UINT16(0x0038),         // bnt [end the loop]
+	SIG_END
+};
+
+static const uint16 qg4AutosavePatch[] = {
+	0x34, PATCH_UINT16(0x0000),         // ldi 0 (waste 12 bytes...)
+	0x34, PATCH_UINT16(0x0000),         // ldi 0
+	0x34, PATCH_UINT16(0x0000),         // ldi 0
+	0x34, PATCH_UINT16(0x0000),         // ldi 0
+	0x32, PATCH_UINT16(0x0038),         // jmp [end the loop]
```
m-kiewitz: Good work,
but please remove the "ldi 0" patches and move the "jmp" upwards plus adjust the offset of the jump.

engines/sci/engine/script_patches.cpp
```diff
@@ -7140,8 +7140,51 @@ static const uint16 qg4InnPathfindingPatch[] = {
...
+//
+// No good can come of this loop. So we skip it entirely. If the disk truly is
+// out of space, a message box will complain, and the player can delete saves
+// voluntarily.
+//
+// Note: Glory::save() contains another space freeing loop, but it might be
+// unreachable.
+//
+// Applies to at least: English CD, English floppy
+// Fixes bug: #10758
```
m-kiewitz: Please also add a comment line saying:
"// Responsible method: Glory::save()"
inbetween "Applies to" and "Fixes bug"

Vhati:
/Blame method. Remove erasure bytes. Increase jmp offset by 12d to compensate. Test. Squash commits.

bluegr:
Great work! Thanks for applying the requested changes.

Merging

sluicebox:
I'm delighted someone is taking QFG4 bugs on, I've never played it but it's notoriously buggy, especially the floppy version. Great job, and welcome!

This patch needs to be slightly adjusted as I'm 99% sure it leaks stack. It also uses a PATCH_ macro in the signature, a crime I'm frequently guilty of. =)

The patch replaces callk with jmp. Prior to the callk instruction there must be instructions that push the parameter count, even if that count is zero, and then push each parameter, assuming they exist. This patch needs to move the jmp farther forward to overwrite that first parameter count push. As it stands now, every time this code executes the stack grows.

m-kiewitz:
Good catches, @sluicebox 
I should have looked closer :/

Vhati:
@sluicebox:
> I'm delighted someone is taking QFG4 bugs on

I'm racing a YouTuber who's working through the series. :P

> I'm 99% sure it leaks stack.

My hello world patch continues to provide teachable moments. :P

> Prior to the callk instruction there must be instructions that push the parameter count, even if that count is zero

Good to know for when I start writing optimized subroutines from scratch.

> It also uses a PATCH_ macro in the signature

*facepalm*

PR #1384, opened 2018-11-08
SCI32: Fix QFG4 great hall entry from barrel room

Vhati:
Fixes a doorMat region that prematurely kicked hero back to the previous room, bug [#10731](https://bugs.scummvm.org/ticket/10731)

	// The castle's great hall has a doorMat region that intermittently sends hero
	// back to the room they just left (the barrel room) the instant they arrive.
	//
	// Entry from room 623 starts hero at (0, 157), the edge of the doorMat. We
	// shrink the region by 2 pixels. Then sEnterTheRoom moves hero safely across.
	// The region is a rectangle. Point 0 is top-left. Point 3 is bottom-left.

Before and after pictures of the polygon are attached to the bug report.


sluicebox:
Do we know the reason why this bug only happens "intermittently"?

Also, the original reporter's save game was using NRS patches so let's make sure that's not involved.

sluicebox:
Great, the NRS patches make a bunch of timing changes, so I wanted to rule that out as a factor in this. Which now we can!

sEnterTheRoom:changeState(1) sometimes sets the room script to sAllRituals. Rgn:setScript disposes of any current script. Since this patch changes sEnterTheRoom to being the room script, wouldn't that cause sEnterTheRoom to be disposed in state 1 and prevent it from reaching later states?

sluicebox:
I took a look and the factor other than timing is hero's loop on the prior screen. doorMat sends you to the barrel room if you're standing on it and facing left. (Loops 1, 5 and 7.) When you're in the barrel room, the way you open the door affects which direction you're facing when the rooms change. At high speeds, doorMat:doit runs before the loop from the prior screen has been updated by walking right, so if you were facing left in the prior room then doorMat boots you.

It looks like when you do anything with the door you start facing left. If you open it normally then you walk through it facing north, which is why that works at all speeds, but if you pick the lock then the door opens and the room changes without moving you, so you're still facing left. I don't know how the other methods of opening this door work.

You can test opening the door normally in the save game by setting global 125 to 1 (or any value other than 2. is this hero type?)

Instead of changing ownership of the script that handles all entrances to 633, which seems risky, can something be done about how hero's loop ends up when exiting the barrel room? Or how it's initialized when entering from the barrel room?

Vhati:
@sluicebox:
> sEnterTheRoom:changeState(1) sometimes sets the room script [...] wouldn't that cause sEnterTheRoom to be disposed

Ehh, yes. \*shifty eyes\* However!

Those scripts are for cutscene evesdropping, from the east, on villain NPCs, who kill hero on sight. You're only allowed to truely enter the room and move freely when they aren't present. The doormat and cutscenes were already effectively mutually exclusive.

Notice that sAllRituals begins with does handsOff() and ends with newRoom().
Same with sListenToKatrina.

> doorMat sends you to the barrel room if you're standing on it and facing left.

Ha!

> setting global 125 to 1 (or any value other than 2. is this hero type?

My notes have a lot of globals sussed out, but not that one yet.

> Ooooh, what if doorMat's polygon was just changed by one pixel to not start at the very edge of the screen?

If one pixel is really all that's causing the fuss. I can't argue with the elegance.

Vhati:
@sluicebox:
> wouldn't that cause sEnterTheRoom to be disposed in state 1 and prevent it from reaching later states?

/Reads again.
Hm. The entry script's later states are door closing...
I've slept too many times since I played through normally. I need to play through that bit again to be sure what happens.

Vhati:
@sluicebox:
> setting global 125 to 1 (or any value other than 2. is this hero type?

Looking over script 140, the class selection screen with 3 doors, I think so.
~~~~
(switch global125
	(0 (fighterSign setCel: 1))
	(1 (wizardSign setCel: 1))
	(2 (robberSign setCel: 1))
)
~~~~

Vhati:
@sluicebox:
> I have english 1.0 floppies
My condolences. XD

> It seems to me the coordinate patch is the best bet
I guess so.

> the side effect of clipping states 2 and 3 
Incidentally, I set a breakpoint on sEnterTheRoom:changeState(), and the cutscenes didn't let it get past state 1 either. I got caught up in other investigations before digging further.

Vhati:
Okay, this patch is a polygon nudge now. PR description updated.

Unrelated: The barrel room has an extra door noise when entering from the great hall. Only on floppy. If I learn more about that, I'll give it a separate ticket.

bluegr:
Many thanks for your feedback everyone! The revised patch is much less intrusive, and less risky than the initial approach. Merging

PR #1387, opened 2018-11-11
SCI32: Fix QFG4 crash when gliding across swamp

Vhati:
Fixes typos that lead to Grooper::doit() arithmetic errors, bug [#10777](https://bugs.scummvm.org/ticket/10777)


bluegr:
Nice catch! Merging

PR #1388, opened 2018-11-11
SCI32: Fix QFG4 autosave stack leak

Vhati:
This leak was introduced in commit [#a37e388](https://github.com/scummvm/scummvm/commit/a37e38830496f4a414bd671428aedbe8831652ff), as discovered after request #1383 had merged.

Moreover, the autosave patch is confirmed to work with the German floppy edition.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -7267,10 +7267,17 @@ static const uint16 qg4InnPathfindingPatch[] = {
 // Note: Glory::save() contains another space freeing loop, but it might be
 // unreachable.
 //
-// Applies to at least: English CD, English floppy
+// Applies to at least: English CD, English floppy, German floppy
 // Responsible method: Glory::save()
 // Fixes bug: #10758
 static const uint16 qg4AutosaveSignature[] = {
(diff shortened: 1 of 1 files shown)


sluicebox:
I just saw that you included the relevant script in the ticket, thank you. Guess I'll finally have to pickup an anthology cd and some floppies and follow along =)

I think this can be a little simpler. You're trying to skip a while loop so you're patching in a jmp and calculating the target offset but you're overwriting the second condition in an AND. The first condition (temp9 > 0) occurs  before that, so there must already be a bt or bnt right before your patch whose target is exactly where you're trying to land. If you just change that instruction's opcode from bt (or bnt) to jmp then your patch will be only one byte with no offset calculations.

Vhati:
@sluicebox:
> there must already be a bt or bnt right before your patch whose target is exactly where you're trying to land

Bah. Where's the fun in that. =P

Yes there is a branch to the same spot immediately beforehand. I saw it as I was stepping through to check my work. Prior to this project that was my level of assembly patching experience: forcing comparisons/jumps. The simplified VM here makes learning so much easier.

I'll shift the signature up a line.

bluegr:
Nice catch! Merging

PR #1389, opened 2018-11-13
SCI32: Fix QFG4 consecutive moonrises

Vhati:
Fixes the time panel never switching from night to day, bug [#10775](https://bugs.scummvm.org/ticket/10775)

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -7356,11 +7356,45 @@ static const uint16 qg4SetLooperPatch2[] = {
 	PATCH_END
 };
 
+// The panel showing time of day gets stuck displaying consecutive moonrises.
+// Despite time actually advancing, the sun will never rise again because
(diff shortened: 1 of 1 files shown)


engines/sci/engine/script_patches.cpp
```diff
@@ -7356,11 +7356,42 @@ static const uint16 qg4SetLooperPatch2[] = {
 	PATCH_END
 };
 
+// The panel showing time of day gets stuck displaying consecutive moonrises.
+// Despite time actually advancing, the sun will never rise again because
+// local5 is set to 1 for night UI and never resets until hero moves to
+// another room.
+//
```
bluegr: Can you please add the explanation you wrote in the bug - you wrote that temp 0 is not used in this function, thus we can safely ignore and reuse the code block that manipulates it in order to fix this bug.

Without this explanation, it’s not clear why we are removing and reusing this code block altogether

bluegr:
Thanks! Merging

PR #1393, opened 2018-11-14
GLK: New Glk engine

dreammaster:
It's been a long time coming, but this engine aims to close the major remaining gap of story-based adventures that ScummVM doesn't yet support.. that of interactive fiction (aka "text adventure"). Some of them, like Infocom, rivaled Lucasarts and Sierra for making compelling stories. Some of the later Infocom ones, in particular, are hybrid text and graphics, making them close to point and click adventures. Unlike my earlier Gargoyle engine of nearly a decade ago, which was a specific implementation of just the Frotz interpreter (for Infocom games), this new implementation uses the GLK portability layer, that was developed some years back, to ease porting the various IF interpreters to different systems. As such, it will make it easier to add support for the different interpreters as sub-engines within this engine.

This new version also has several further advantages over my previous attempt. The GLK layer provides support for Unicode, which allows more flexible display of foreign languages.. coupled with ScummVM's support for U32String and passing them to the font drawing routines. ScummVM also supports copying and pasting to/from the clipboard, which wasn't available back when I did my original engine.

At the moment, I've only implemented the Scott Adams interpreter as a sub-engine as a proof of concept, and an in-progress Frotz sub-engine. It should be fairly easy to likewise add other interpreters in afterwards. I've kept all the method names and types available as in the original GLK, so the bulk of work for new sub-engines will simply be setting them up in their own sub-namespace, massaging them as necessary to have a proper class structure, and any miscellaneous ScummVM things like the detection code, and supporting loading from the launcher.

The engine comes with it's own data file, which is the set of freely available GLK TTF fonts packaged up in a renamed zip file called fonts.dat - due to a prior suggestion to keep it generic so as to be more cleanly reused by other engines. The Scott Adams games are freely available from his website at http://www.msadams.com/. The scott sub-engine detects the games using md5 matches on any filename with either a .dat or .saga extension. So you can rename the games as you please for better naming clarity.

Currently, I've implemented the bulk of the GLK compatibility layer, except for the sound handling and some of the graphic methods. These are waiting on implementing a sub-engine which makes use of them, like Frotz, so I can test the results of what I implement.


devtools/credits.pl
```diff
@@ -602,6 +602,13 @@ sub add_paragraph {
 			begin_section("Fullpipe");
 				add_person("Eugene Sandulenko", "sev", "");
 			end_section();
+
+			begin_section("Gargoyle");
```
digitall: Leftover... Change to ScummGLK?

dists/scummvm.rc
```diff
@@ -32,6 +32,9 @@ cryo.dat               FILE    "dists/engine-data/cryo.dat"
 #if ENABLE_DRASCULA   == STATIC_PLUGIN
 drascula.dat           FILE    "dists/engine-data/drascula.dat"
 #endif
+#if ENABLE_GARGOYLE   == STATIC_PLUGIN
```
digitall: Leftover... Change to ENABLE_GLK ?

dreammaster:
@bgK Good call on deriving from MetaEngine. I was able to clean up the detection code somewhat, and do away with the master game list that duplicated all the Ids from both games.

dreammaster:
With the changeover to MetaEngine, I've also added in code to the detector so that if the filename is blank (which happens when launching directly from the command line), it will scan in the folder and launch the first filename with a matching game Id. So the command line will work now, you just won't have the option to choose which specific file if there are multiple versions of the game in the same folder.

bgK:
@dreammaster with your changes to the detection code this looks ok to be merged IMO :+1: 

This is a quick memory leak report, if it is of interest: https://gist.github.com/bgK/db7eb24f3d48e0aa7a3f04d1e6039012.

dreammaster:
> This is a quick memory leak report, if it is of interest:  https://gist.github.com/bgK/db7eb24f3d48e0aa7a3f04d1e6039012.

Thanks. It'd slipped my mind to check for leaks

digitall:
dreammaster: I know not ready for rebase and merge of this PR yet ... but this is currently broken i.e.
"This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. "

Not quite sure why, but you may want to manually rebase on top of the latest master and deal with fixing up.

dreammaster:
> Not quite sure why, but you may want to manually rebase on top of the latest master and deal with fixing up.

No problem. Done.

dreammaster:
NOTE: I'm temporarily marking the engine as not to be merged yet (even after the review period). I read/recalled in passing that the later Infocom game Beyond Zork uses a special character graphics font, and I want to chance to review the details. Maybe add the necessary data as an extra font in the fonts.dat file doing a rebase, and avoid the waste a fresh commit would require if the engine was already in master.

eriktorbjorn:
As far as i know, only Beyond Zork uses the special font. It was used to draw an on-screen map, progress bars (for strength, dexterity, etc.) and to render words you didn't understand (usually spells) as runes if your intelligence wasn't high enough to understand them.

The font is described in https://www.inform-fiction.org/zmachine/standards/z1point1/sect16.html

It might be possible - perhaps preferable? - to remap the rune symbols to the appropriate Unicode characters, assuming the bundled fonts have them. The other symbols... probably not so much.

The later games (Zork Zero, Journey, Shogun and Arthur) used bitmap graphics for illustrations and a couple of puzzles. I know that Frotz has some support for it - the SDL version of frotz seems to handle it fairly well, though some colors were wrong last time I tried it, and I think you have to download converted graphics files for the pictures - but I don't know how well Glk can support the screen model...

dreammaster:
I'm wondering if it'll be worthwhile future proofing the fonts.dat file, and add in any extra needed fonts for ZVision and/or Wintermute now whilst I can still rebase the history..

PR #1395, opened 2018-11-15
SCI32: Fix QFG4 talking to absent innkeeper

Vhati:
Fixes inn dialogue after Igor's rescue, bug [#10753](https://bugs.scummvm.org/ticket/10753)

    // Visiting the inn after rescuing Igor sets a plot flag. Such flags are tested
    // on subsequent visits to decide the dialogue options when clicking MOUTH on
    // hero. That particular check neglects time of day, allowing hero to talk to
    // an empty room after midnight... and get responses from the absent innkeeper.
    //
    // The inn's init() boils down all the checks into values for local2,
    // representing discrete situations. Then there's a switch block in sInitShit()
    // that acts on those values, making arrival announcements and setting new
    // flags.
    //
    // "So Dmitri says the gypsy didn't really kill Igor after all." sets flag 132.
    // "I must thank you for saving our Tanya." sets flag 134. If you have the
    // former, not the latter, and other situations didn't apply, local2 = 11 and
    // this bug occurs.
    //
    // A day in QFG4 is broken up into 3-hour spans: 6,7,0,1,2,3,4,5. Where 6 is
    // midnight. The sun rises in 0 and sets in 4. Current span is global[123]. The
    // innkeeper is not around from midnight to morning.
    //
    // To make room, we optimize an earlier block's time check:
    // "t <= 3 || t is one of [4 5]" becomes "t <= 5". The bugged block gets a
    // similar pre-midnight check.


sluicebox:
Talking to an empty room, I love it.

I haven't quite made sense of this patch, which isn't a criticism since I don't know the game, but a few thoughts...

The two PATCH_GETORIGINALBYTEADJUSTs are unnecessary and make the patch harder to read. The purpose of that macro is to make a patch work against different versions of scripts that have different offsets. Your signature has hard-coded offsets so these macros will always produce the same results, so there's no point to the macros, just put the new offsets you want. Otherwise when someone reads the code they have to do that math themselves.

Shorter version: you have hard coded offsets in your signature so put hard coded offsets in your patch.

I browsed the wiki for known versions of this game and what's on collection CDs, it looks like there's only three: 1.0 floppy, 1.1 floppy (localized aka the german version on compilation cd), and the cd version. From your comments it sounds like you've got all three so you can operate with a high degree of confidence that you have all scripts out there. It looks like you're being admirably cautious about the compatibility of your patches, but since you also appear to have all known versions, you can operate accordingly. I believe that the smallest signature/patch you can come up with that applies to all three will be all that's ever needed.

I've only been doing this for a little while so I'm on the same page. Excited that I'll be able to play this game someday and not worry about bugs! =)

Vhati:
@sluicebox:
> I haven't quite made sense of this patch

Not sure if you mean the game's tangled logic, as detailed in the bug report, or the asm modification.

In QFG4, players can click MOUTH on other characters to **ask things**, and on hero to **tell things**. For telling, an obvious character in the room is assumed as the listener - or the narrator says no one's around. The innkeeper has a daily schedule. He sleeps off-screen at midnight until morning. Apparent'y his sprite is determined independently of the click-on-hero dialogue options. Or he ventriloquises in his sleep. Because there are times he is not visible, yet he responds when you "Tell about yourself".

The assembly is rather simple. The sig covers two adjacent blocks within a large cond (an if-else structure that uses a series of conditions to decide a value for local[2], like a big ternary operator)

Each block in a cond has... Jumps to the end when a block's conditions are fully satisfied and it decides the value to be stored in local[2] (it junps to sal). And branches that short-circuit over to the next block when any condition fails.

The bug is in the second block here, which needed its conditions prepended with a check to time of day. The first block had a gratuitous call. By shortening the first block, I created a gap.

> you have hard coded offsets in your signature so put hard coded offsets in your patch.

Yeah. I committed, slept, and thought the same when I woke up this morning. Will do.

At the time, I'd envisioned the branch as landing at the next block, and in using the freed bytes to prepend 7 bytes to that next block, I scooted up the destination by 7 bytes - saving myself the math of counting the actual patch bytes *between* the branch and the next block's new prepended start.

The jump lands at the end, like every block's jump does, to set local[2]. By shortening, the jump appeared earlier, and I lazily adjusted the difference without counting. Having slept, that 17 *is* counter intuitive to me now, unrelated to the aforementioned 7 bytes.

No excuse for that. I even wrote a python scraper to to my byte counting for me!

Vhati:
> Apparent'y his sprite is determined independently of the click-on-hero dialogue options.

Correction: the inkeeper sprite WAS dependent on local[2]==11... partially.

The sprite had a white list of situations, including 11. Then it was doing an additional pre-midnight check independently. Prior to this patch, that meant his sprite appeared on fewer occasions than there were times that hero could talk to him. Their criteria were out of sync.

I posted on the bug report with a source excerpt.

Vhati:
Hm. The local2 blocks, being base criteria, should be at least as restrictive as the sprite. I may need to amend or revisit this patch, expanding to fix the next block (local[2]==12, after saving Tanya).

Dunno what situation 12 entails currently, but there's no time check on it.

Vhati:
Grr. 12 needs fixing too.

It unlocks "Tell about adventures". After midnight, the absent innkeeper replies, "You have done much to help Mordavia. We have much to thank you for."

Vhati:
/Expanded patch to fix both block 11 and block 12.
Should be viable now.

digitall:
@Vhati: Happy for review now?

Vhati:
@digitall:
> Happy for review now?

Sure.

As these feature branches fall behind master and into conflict, I've been rebasing & force pushing to catch them up. If I force push last minute substantive changes, I add a comment to the PR. Am I doing that right?

digitall:
@Vhati: Yes, I think so. In the sense that I don't think there is a better solution.

bluegr:
Nice work! Merging

PR #1396, opened 2018-11-15
ENGINES: Add optional extra configuration entries when creating new targets

dreammaster:
As per digitall's request, this separate pull request contains the single commit for the change to the engines  to support optional extra entries to be set to be written into the generated section for a game entry.

Background: Since most interactive fiction games are only a single, it'll likely frequently be the case that multiple games will be in the same folder, even potentially multiple versions of the same game. Because of this, when the user selects a detected game entry supported by the Glk engine, it uses this new functionality to additionally store the filename the match was for. This way, when the game starts, it will ensure that the specific file is used.

One of the earlier concerns with this, that running games directly from the command line wouldn't work, has already been resolved. The engine has fallback code that will scan the specified folder for any file that detects as being of the given game Id, and play it. Also, this won't affect adding any non-Glk games, since if the _extraConfigEntries is empty, the createTargetForGame works as it always has.

### DIFF
--- a/engines/game.h
+++ b/engines/game.h
@@ -148,6 +148,21 @@ struct DetectedGame {
 	 */
 	GameSupportLevel gameSupportLevel;
 
+	/**
+	 * A list of extra keys to write to the configuration file
+	 */
+	Common::StringMap _extraConfigEntries;
+
+	/**
+	 * Allows adding of extra entries to be saved as part of the detection entry
+	 * in the configuration file.
+	 * @remarks		Any entry added using this should not be relied on being present
+	 *				in the configuration file, since starting games directly from the
+	 *				command line bypasses the game detection code
+	 */
+	void addExtraEntry(const Common::String &key, const Common::String &value) {
+		_extraConfigEntries[key] = value;
+	}
 private:
 	/**
 	 * Update the description string by appending (EXTRA/PLATFORM/LANG) to it.
--- a/base/plugins.cpp
+++ b/base/plugins.cpp
@@ -580,6 +580,11 @@ Common::String EngineManager::createTargetForGame(const DetectedGame &game) {
 	addStringToConf("extra", game.extra, domain);
 	addStringToConf("guioptions", game.getGUIOptions(), domain);
 
+	// Add any extra configuration keys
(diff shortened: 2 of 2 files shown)


digitall:
@dreammaster : Thanks for doing this. One question, I assume that game engines adding extra keys to the config file can only do so under their domain i.e. their game entries?

Just that was wondering about malicious or more likely just namespace collisions causing config file corruption or just weird bugs... Is that possible?

dreammaster:
Correct. The new code is in the method that can creates the section for newly added games, so anything added on the detection entry just gets written along with them. So there's no possibility of anything added interfering with other parts of the file

engines/game.h
```diff
@@ -148,6 +148,17 @@ struct DetectedGame {
 	 */
 	GameSupportLevel gameSupportLevel;
 
+	/**
+	 * A list of extra keys to write to the configuration file
```
bgK: It would be nice to document that the properties set in this variable are not guaranteed to be present in all cases when the game is launched. e.g. when launching from the command line without a target. MetaEngines may need to have fallback code for when they are not here.

engines/game.h
```diff
@@ -148,6 +148,17 @@ struct DetectedGame {
...
+	 */
+	Common::StringMap _extraConfigEntries;
+
+	/**
+	 * Convenience operator for adding entries to the _extraConfigEntries
+	 */
+	Common::String &operator[](const Common::String &key) {
```
bgK: I'm not so sure it is a good idea to have this operator. This class defines what a game is when it was found by a detector, it does not behave like a bag of properties.

dreammaster: Hmmm.. you're right, and we wouldn't want anyone to accidentally use it thinking it can be used to set other core/standard properties. I'll change it to an explicit method, and add in the comments as you suggested above visa ve. command line.

digitall:
@dreammaster : Since there appear to be no objections to this and it appears to be pretty reasonable and self-contained, are you happy for me to merge?

bluegr:
This commit has been already added as part of the glk engine merge. Thus, I’m closing this PR as it’s outdated now

PR #1398, opened 2018-11-16
manpage: update cmdline options and hotkeys

kparal:
Add missing cmdline options and hotkeys to the manpage. Add long options
where they were missing. Update existing options description and values.
Improve formatting for important words, arguments, etc. Sort the options
in the same way they are sorted in README, so that future synchronization
is easier. Add info into several places saying that README is more
comprehensive than this manpage.

----

Hello, it took me a while to realize that your manpage contained much less information than the readme file. I decided to fix that and update it. This is my first time editing a manpage, I wasn't familiar with the syntax, but I verified in renders fine for me (Fedora 29).

### DIFF
--- a/dists/scummvm.6
+++ b/dists/scummvm.6
@@ -7,117 +7,152 @@
 .Nd graphic adventure game interpreter
 .Sh SYNOPSIS
 .Nm scummvm
-.Op Ar options
-.Op Ar game
+.Op Ar OPTIONS
+.Op Ar GAME
 .Sh DESCRIPTION
 .Nm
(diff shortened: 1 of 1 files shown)


digitall:
@kparal : Looks good. Will give it a bit of time for review and comments, but should be fine to merge. Thanks.

lotharsm:
Looks good to me too. Thanks for your work!

digitall:
Since there appear to be no objections and this looks fine, merging as-is.

kparal:
Thanks. And also thanks for working on a great and very useful project.

PR #1400, opened 2018-11-17
CREATE_PROJECT: Remove obsolete references to msvc8

SupSuper:
Just something that's been bugging me for a while. :)


digitall:
Ah, I see dists/msvc8 does not exist... Well, looks pretty self-explanatory. Merging as-is.

PR #1402, opened 2018-11-17
GUI: Add an option to set the GUI language to the game language

bgK:
If this option is enabled, when starting a game the GUI language is changed to the game language. That way, if a game uses the ScummVM save and load dialogs, they are in the same language as the game.

This is one of the changes we did for the 25th anniversary release of the Myst games.

The option looks like this in the GUI:
screenshot from 2018-11-17 15-09-03


digitall:
@bgK : Looks reasonable. I assume this changes the setting in the transient domain? One question, does the GUI language get restored to default when returning to launcher? Just want to avoid the launcher being in the game language unexpectedly... (I have some games in German for testing, but not sure I know enough to flail with the Launcher :)

bgK:
> I assume this changes the setting in the transient domain?

The GUI persists the user choice in the configuration file. The transient domain is not affected.

> One question, does the GUI language get restored to default when returning to launcher?

Yes.

bluegr:
Looks good to me as well. Nice work! :)

digitall:
Have reuqested a quick review from @sev- but this should be fine to merge as-is. If no further comments or objections are raised, I will merge this later this week.

bgK:
> If no further comments or objections are raised, I will merge this later this week.

Thank you, but I very much prefer merging my own pull requests. I usually use that opportunity to do a last testing pass.

PR #1404, opened 2018-11-18
POSIX: Set the default screenshots path to the XDG Pictures directory

bgK:
The screenshots are written to `$XDG_PICTURES_DIR/ScummVM Screenshots` as defined in the XDG user directories specification (https://www.freedesktop.org/wiki/Software/xdg-user-dirs/).

With this change the screenshots are saved to a user writable path by default on Linux when ScummVM is installed through a package manager.

### DIFF
--- a/README
+++ b/README
@@ -2896,6 +2896,8 @@ The default location, when no screenshot path is defined in the config file,
 depends on the OS:
   Windows: In Users\username\My Pictures\ScummVM Screenshots.
   macOS X: On the Desktop.
+  Other unices: In the XDG Pictures user directory,
+                 e.g. ~/Pictures/ScummVM Screenshots.
   Any other OS: In the current directory.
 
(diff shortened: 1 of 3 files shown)


lotharsm:
Sound like a good change to me.

Does that mean that currently saving screenshots is broken on Linux when scummvm is installed through a package manager?

bgK:
> Does that mean that currently saving screenshots is broken on Linux when scummvm is installed through a package manager?

By default, the screenshot path was the working directory. When running ScummVM from the .desktop entry then the working directory is likely the home directory which would have worked. When launching from a shell, it depends whether the shell's current directory is writable.

criezy:
I think it indeed makes sense to change the default screenshots directory to the XDG_PICTURES_DIR on Linux/Unix.

PR #1405, opened 2018-11-18
WIN32: Try to attach to the parent console early in the initialization

bgK:
Don't create a new console if our parent process has one or has already initialized standard input / output for us.

This fixes the command line output not being visible. It also prevents us from opening of a new console window when starting ScummVM from a console.

### DIFF
--- a/backends/platform/sdl/win32/win32.cpp
+++ b/backends/platform/sdl/win32/win32.cpp
@@ -34,6 +34,7 @@
 #define _WIN32_IE 0x500
 #endif
 #include <shlobj.h>
+#include <io.h>
 
 #include "common/scummsys.h"
 #include "common/config-manager.h"
@@ -52,7 +53,37 @@
 
 #define DEFAULT_CONFIG_FILE "scummvm.ini"
 
+#if !defined(ENABLE_CONSOLE_WINDOW)
+static bool hasStandardOutput() {
+	if (_fileno(stdout) >= 0 || _fileno(stderr) >= 0) {
+		// _fileno was broken for SUBSYSTEM:WINDOWS from VS2010 to VS2012/2013.
+		// http://crbug.com/358267. Confirm that the underlying HANDLE is valid.
+		intptr_t stdoutHandle = _get_osfhandle(_fileno(stdout));
+		intptr_t stderrHandle = _get_osfhandle(_fileno(stderr));
+		if (stdoutHandle >= 0 || stderrHandle >= 0)
+			return true;
+	}
+
+	return false;
+}
+#endif
+
 void OSystem_Win32::init() {
+#if !defined(ENABLE_CONSOLE_WINDOW)
+	// Initialize standard output. This is done early so the output
+	// from the command line parsing is visible in the console window.
+	//
+	// If we already have console output, don't do anything.
+	// This is the case when called from a terminal emulator like mintty.
+	//
+	// Otherwise, if the parent process has a console, attach to it.
+	if (!hasStandardOutput() && AttachConsole(ATTACH_PARENT_PROCESS)) {
+		freopen("CONIN$",  "r", stdin);
+		freopen("CONOUT$", "w", stdout);
+		freopen("CONOUT$", "w", stderr);
+	}
+#endif
+
 	// Initialize File System Factory
 	_fsFactory = new WindowsFilesystemFactory();
 
@@ -69,20 +100,17 @@ void OSystem_Win32::init() {
 }
 
 void OSystem_Win32::initBackend() {
+#if defined(ENABLE_CONSOLE_WINDOW)
 	// Console window is enabled by default on Windows
 	ConfMan.registerDefault("console", true);
 
 	// Enable or disable the window console window
 	if (ConfMan.getBool("console")) {
-		if (AllocConsole()) {
-			freopen("CONIN$","r",stdin);
-			freopen("CONOUT$","w",stdout);
-			freopen("CONOUT$","w",stderr);
-		}
 		SetConsoleTitle("ScummVM Status Window");
 	} else {
 		FreeConsole();
 	}
+#endif
 
 	// Create the savefile manager
 	if (_savefileManager == 0)
--- a/backends/platform/sdl/win32/win32.mk
+++ b/backends/platform/sdl/win32/win32.mk
@@ -55,6 +55,9 @@ ifdef USE_SPARKLE
 	cp $(WIN32SPARKLEPATH)/WinSparkle.dll $(WIN32PATH)
 	sed -e '/WinSparkle\.dll/ s/^;//' -i $(WIN32PATH)/ScummVM.iss
 endif
+ifdef ENABLE_CONSOLE_WINDOW
+	sed -e '/--no-console/ s/^;//g' -i $(WIN32PATH)/ScummVM.iss
+endif
 	unix2dos $(WIN32PATH)/*.txt
 	unix2dos $(WIN32PATH)/doc/*.txt
 	unix2dos $(WIN32PATH)/doc/cz/*.txt
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -81,7 +81,7 @@ static const char HELP_STRING[] =
 	"  --auto-detect            Display a list of games from current or specified directory\n"
 	"                           and start the first one. Use --path=PATH to specify a directory.\n"
 	"  --recursive              In combination with --add or --detect recurse down all subdirectories\n"
-#if defined(WIN32) && !defined(_WIN32_WCE) && !defined(__SYMBIAN32__)
+#if defined(WIN32) && defined(ENABLE_CONSOLE_WINDOW) && !defined(_WIN32_WCE) && !defined(__SYMBIAN32__)
 	"  --console                Enable the console window (default:enabled)\n"
 #endif
 	"\n"
@@ -665,7 +665,7 @@ Common::String parseCommandLine(Common::StringMap &settings, int argc, const cha
 			END_OPTION
 #endif
 
-#if defined(WIN32) && !defined(_WIN32_WCE) && !defined(__SYMBIAN32__)
+#if defined(WIN32) && defined(ENABLE_CONSOLE_WINDOW) && !defined(_WIN32_WCE) && !defined(__SYMBIAN32__)
 			// Optional console window on Windows (default: enabled)
 			DO_LONG_OPTION_BOOL("console")
 			END_OPTION
--- a/configure
+++ b/configure
@@ -170,6 +170,7 @@ _optimizations=auto
 _use_cxx11=no
 _verbose_build=no
 _text_console=no
+_console_window=yes
 _mt32emu=yes
 _build_scalers=yes
 _build_hq_scalers=yes
@@ -1009,6 +1010,7 @@ Optional Features:
   --disable-eventrecorder  disable event recording functionality
   --enable-updates         build support for updates
   --enable-text-console    use text console instead of graphical console
+  --disable-console-window disable the console window (Windows only)
   --enable-verbose-build   enable regular echoing of commands during build
                            process
   --disable-bink           don't build with Bink video support
@@ -1229,6 +1231,8 @@ for ac_option in $@; do
 	--disable-eventrecorder)  _eventrec=no   ;;
 	--enable-text-console)    _text_console=yes ;;
 	--disable-text-console)   _text_console=no ;;
+	--enable-console-window)  _console_window=yes ;;
+	--disable-console-window) _console_window=no  ;;
 	--with-fluidsynth-prefix=*)
 		arg=`echo $ac_option | cut -d '=' -f 2`
 		FLUIDSYNTH_CFLAGS="-I$arg/include"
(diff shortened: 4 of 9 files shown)


digitall:
@bgK: Kudos for looking at this! Hopefully this will fix the various bugs regarding this.

Vhati:
Does this mean you're still building a GUI executable, not a console one?
.
.
From the "Use the Windows subsystem for scummvm.exe" commit...

    <SubSystem>" << ((setup.devTools || setup.tests) ? "Console" : "Windows")

In my bug report, the [stackoverflow](https://stackoverflow.com/questions/54536/win32-gui-app-that-writes-usage-text-to-stdout-when-invoked-as-app-exe-help) link pointed out...

> **GUI executable:** cmd.exe will assume that your program does no console I/O so won't wait for it to terminate before continuing, which in interactive mode (ie not a batch) means displaying the next ("C:\>") prompt and reading from the keyboard. So even if you use AttachConsole your output will be mixed with cmd's output, and the situation gets worse if you try to do input. This is basically a non-starter.

> **Console executable:** Contrary to belief, there is nothing to stop a console executable from displaying a GUI, but there are two problems.
>
> The first is that if you run it from the command line with no arguments (so you want the GUI), cmd will still wait for it to terminate before continuing, so that particular console will be unusable for the duration. This can be overcome by launching a second process of the same executable [...], passing the DETACHED_PROCESS flag to CreateProcess() and immediately exiting. The new process can then detect that it has no console and display the GUI.

Vhati:
The StackOverflow author went to extra lengths to hide their console window.

For ScummVM, it ought to be sufficient to always build with the 'console' subsystem, and - if it makes sense for other platforms? - make allocation of a new status window conditional (a condition that'd always be false on this OS).

Vhati:
I see, the executable accepts a "--no-console" arg. So the app needs to launch with an OS-provided console, parse args, and then optionally sever its ties to the parent window (i.e. CreateProcess above).

bgK:
> Does this mean you're still building a GUI executable, not a console one?

Yes. There does not seem to be a solution that handles our use case perfectly. When targeting the Console subsystem, I believe there will always be a console windows flashing for a brief amount of time until the process closes it either using FreeConsole or by starting another detached process and quitting. With the Windows subsystem, when starting the process from an existing console, cmd.exe reclaims the input immediately and the user can type commands at the same time as the process writes its output.

Since the most important use case on Windows is starting ScummVM as a GUI program, I'd say targeting the Windows subsystem so there is no flashing console window is a better option even if that means having kind of broken console output. Please share your thoughts.

Vhati:
@bgK:
> I'd say targeting the Windows subsystem so there is no flashing console window is a better option even if that means having kind of broken console output.

I'm not personally bothered by a flashing window. A live cmd prompt accepting keystrokes on the status window for the duration of a game feels wronger, but it'd only occur when the game was launched from an interactive prompt.

Admittedly, that would probably only happen when someone is getting acquainted (--help --list-targets, etc) or redirecting output to troubleshoot or something.

I happen to be launching with batch scripts, not interactive.

Vhati:
If the status window (or parent console attachment) is still enabled by default, the flashing would only occur for users who opt to suppress it.

bgK:
> I'm not personally bothered by a flashing window. A live cmd prompt accepting keystrokes on the status window for the duration of a game feels wronger, but it'd only occur when the game was launched from an interactive prompt.

With ScummVM being used for commercial re-releases of games, having a console window showing does not look very professional in my opinion.

Perhaps it is time to change ScummVM not to open a console window by default on Windows? We now have the unknown game dialog for users to notice the game their are trying to add is not recognized. Errors have been displaying in the GUI for a long time. The other platforms don't open a console window.

Kirben:
The changes make no sense, the problems you describe sound like they are related to your build system.

ScummVM will already use the current console window, if started via a Command Prompt, and there are currently no problems with the console output.

Vhati:
@Kirben:
> they are related to your build system [...] there are currently no problems with the console output

https://bugs.scummvm.org/ticket/10809

> Last release binary that could print was 1.8.1 (May 22 2016 15:57:02).

Kirben:
The problems never occur in my builds, and buildbot been broken, and not fixed again isn't surprising. It looks like '-mwindows' is still based used at link time, maybe from SDL2 config file?

Vhati:
@Kirben:
> maybe from SDL2 config file?

I recently set up an MSYS environment. Minimal cruft.

Editing out the two mentions of "-mwindows" in "MinGW/bin/sdl-config" indeed fixed it for me.

bgK:
Thank you for your feedback. I'll leave this open for a few more days to gather more opinions. If the consensus is for ScummVM to be a Console application then I'll submit a patch to `configure` to make sure the correct linker flags are being used. Asking people to edit the SDL configuration on their toolchain seems like a bad idea to me as it can affects the behavior of other programs they build.

Kirben:
Well SDL shouldn't including "-mwindows" in the first place, as SDL is used for a variety of applications, many of which offer console options.
If you mean adding "-mconsole" via the configure script, I think it is even more confusing to be passing along both options at linker time.

Vhati:
bgK:
> a patch to configure to make sure the appropriate linker flags are being used.
> [...]
> forcefully removing the -mwindows flag if it is there and adding the -mconsole flag

A reasonable solution. When I saw the old tickets had identified a dependency that started demanding a problematic and unnecessary build requirement by default, I assumed this had already been done. Or was about to this time.

Kirben:
Agreed, removing the -mwindows flag and adding the -mconsole flag instead, sounds like the best option.

sluicebox:
Wow, I never noticed that my VS builds were Console subsystem, I assumed they were Windows like the daily builds. This explains some things! I'm glad that inconsistency is getting fixed.

It sounds like we're also discussing which subsystem all the Windows builds should use? I vote Windows.

Just to reiterate the eternal subsystem dilemma...
Console: Window appears before your code runs so hiding it will flash
Windows: You're in control but on the hook for maintaining win32 code with 98% console compatibility

The second one is the only way to have a truly optional console window so lets use it everywhere. Since it sounds like we're also discussing hiding said window...

Console subsystem would be an obstacle to the dream of completely hiding the console window by default. This should be the goal for future releases and probably even daily builds. I'm surprised to see that the 2.0 release shows it by default. The target audience for a "noconsole" shortcut doesn't know what that means. (And it doesn't help that there's not a space between those words)

At this point any information that a normal user needs that only appears in the console is a bug. If it's so important, why don't Mac users see it?

lotharsm:
> Perhaps it is time to change ScummVM not to open a console window by default on Windows? We now have the unknown game dialog for users to notice the game their are trying to add is not recognized. Errors have been displaying in the GUI for a long time. The other platforms don't open a console window.

I highly agree with that. In my opinion, the console window should be hidden by default and only be available through a seperate shortcut for debugging purposes.

bgK:
Let me try to do some tallying:
* Kirben: keep ScummVM as a Console application
* Vhati: Console as well (but would not mind Windows with the console window disabled by default?)
* sluicebox: make it a Windows application
* rootfather: the console should be hidden by default
* bgK: make it a Windows application with the console hidden by default

Is that right ?

Vhati:
@bgK:
> Vhati: Console as well (but would not mind Windows with the console window disabled by default?)

Yeah.
I have some small vague concern about how command line use would be affected. Not enough to fight over it. I'll forget my qualms eventually... so long as I don't one day find myself routinely using an interactive prompt and discover the concern was valid. :P

Kirben:
Once again this request was only opened due to buildbot mis-configuration, which I'm tired of dealing with. It makes no sense at all, to switch an application that offers console options, to been a windows application.

The console window is enabled by default due to the numerous messages still passed along, just scan the source code for 'warning'. This is essential for any games that are still in an incomplete state (feature or game play wise), and no one has bothered to redirect the remaining messages for stable games either.

An argument could be made for a hidden console in the release version only, due to the addition of the unknown game dialog window, and all incomplete games been disabled.

digitall:
I do agree with @Kirben that we should not remove or hide the console by default until we have a solution where all debug, warning and error calls are sent to GUI messages or to a "soft-console" as per the debug console which can be cut and paste from... and probably a/the log file. ... and that is a much larger scope to sort out.

digitall:
Anyway, I will update the wiki to note the need to modify sdl-config / pkg-config sdl output.

bgK:
> Anyway, I will update the wiki to note the need to modify sdl-config / pkg-config sdl output.

Please don't. This issue should be resolved in the ScummVM build system. Encouraging users to monkey patch their toolchain cannot be lead to anything good.

digitall:
@bgK: Sorry already done before I saw your comment.

I know what you mean but I am not "encouraging them". If they are compiling their own MinGW builds, they are already power users or nascent developers and thus the ability to use a text editor is not beyond them :)

Overall, the issue already exists, we already need to hack the sdl-config and pkg-config files to remove the -mwindows line ... and it keeps getting missed at least for buildbot toolchain updates and we get bug reports. Documentation is a good first line of defense on that.

If we get a fix into configure, then we can change the wiki notes into a "If you have issues with console / command-line output, then check this...", but the notes on buildbot admin page should be retained so we avoid the continual reoccurrence of this on toolchain updates.

criezy:
I agree that the console should probably be visible by default in development builds, but I think it would make sense to hide it by default in release builds.

I won't comment on the choice between `-mwindows` and `-mconsole` since I don't know what the implications are. But if we do want to use `-mconsole` I agree that we should add a workaround for it in `configure` rather than tell users to edit their `sdl-config` (especially if they are also using SDL with other applications that may rely on the `-mwindows` being provided by `sdl-config`).

bgK:
I have changed the code in this pull request to:
* Add a configure / create-project option `--disable-console-window`
* Remove `-mwindows` from the output of the `sdl-config` script if present.
* Specify the subsystem to use when building scummvm.exe.

The new `console-window` option works as follow:

When the console window option is enabled (default), the main executable is
built as a Console subsystem program so that Windows opens a console
on startup. The program is blocking when started by cmd.exe.
The --no-console runtime command line argument allows to hide the
console window as soon as possible, but it remains visible for a
small amount of time.

When the console window is disabled, the main executable is built as
a Windows subsystem program. Windows does not open a console window.
ScummVM attaches to the parent console output stream when possible,
but the output is mixed with the output of subsequent commands.
It is not possible to create a console window using command line.

Hopefully this can satisfy all the use cases. It is possible to build release-style
executables that don't show a console window (even for a short period of time
on startup). Or to build executables that optionally show a console window
and have a perfectly working command line. Thoughts ?

Vhati:
@bgK:
> When the console window is disabled, the main executable is built as
a Windows subsystem program. [...] It is not possible to create a console window using command line.

Not even from a "scummvm.exe --console" start menu shortcut?

bgK:
@Vhati 
> Not even from a "scummvm.exe --console" start menu shortcut?

My thought is that you really don't care about a console window in that case. If you want to see the output, it is in the log file (or you can open ScummVM from a console / terminal emulator, or you can set USE_WINDBG to see it in the Visual Studio output window). Also it keeps the code simple.

Vhati:
OT Question: Is there a way to send text from the in-game debugger to a log? Like a debugflag or something?

bgK:
> Is there a way to send text from the in-game debugger to a log?

I don't think there is a way with the graphical debug console currently. However if you spend a lot of time in ScummVM's debugger you might be interested in the `--enable-text-console` configure flag. It allows to use the console ScummVM was launched from as the debugger, with the convenience of you favorite terminal emulator and the usual keyboard shortcuts of the readline library.

Vhati:
@bgK:
> if you spend a lot of time in ScummVM's debugger you might be interested in the --enable-text-console configure flag. It allows to use the console ScummVM was launched from as the debugger, with the convenience of your favorite terminal emulator

Excellent!

\*pause\*

\*looks up\*
.
.
Vhati:
> I'll forget my qualms eventually... so long as I don't one day find myself routinely using an interactive prompt and discover the concern was valid. :P

Kirben:
The console window is already optional via a command line or config file option, switching to a compile time option is going backwards, especially for people are already using the option. Switch the default for the console config option, would be easier if release builds changed in the future.

The only difference seems to be no very brief flash of a console window, is a brief flash really that bad? most DOS games from GOG have similar flash, due to DOSBox logo been shown before switching to full screen.

sluicebox:
My vote for was "make it a Windows application with the console hidden by default".

But I'm a little confused by this talk of making ScummvM a Windows subsystem application, as if it isn't already. The 1.9 and 2.0 release builds are Windows subsystem. The mingw-w32-master-latest buildbot builds are Windows subsystem. The VS project has been Console, an inconsistency that this PR fixes.

Vhati:
@sluicebox:
> I'm a little confused by this talk [...] The 1.9 and 2.0 release builds [...] The mingw-w32-master-latest buildbot builds are Windows subsystem.

By accident.

ScummVM never explicitly chose a subsystem, so a console app was the implicit default in gcc. Some time ago, a script that ships with SDL started reporting a spurious build requirement that all projects linking against it must use the windows subsystem.

Previously, someone had edited SDL's script to STOP doing that - returning to the implicit default of console. And whenever SDL was updated, if that script wasn't re-edited, this happened all over again.

Kirben:
Using "scummvm.exe --console" no longer works according to your previous message, when using the new configure option. Offering a different build where config file options no longer work as expected isn't a good idea.

henke37:
#1639 affected this code. You should probably take a quick peek at it.

henke37:
#2405 should solve this neatly.

bluegr:
Thanks for the work done here!

The needed implementation has now been done in other commits and pull requests over the years. So, the feature is now present in daily builds, therefore this is now a good time to close this pull request, as I don't believe that there's anything else that can be used from the original submission.

Thanks again everyone for the work done and the discussion made on this feature. If there's anything else regarding the windows console that hasn't been implemented on master, feel free to chime in.

Closing this

PR #1406, opened 2018-11-20
SCI: Fix KQ6CD wallflower dance bug

sluicebox:
Fixes script bug in the original in Text mode. bug #10811


engines/sci/engine/script_patches.cpp
```diff
@@ -112,6 +112,7 @@ static const char *const selectorNameTable[] = {
 	"startAudio",   // King's Quest 6 CD / Laura Bow 2 CD for audio+text support
 	"modNum",       // King's Quest 6 CD / Laura Bow 2 CD for audio+text support
 	"has",          // King's Quest 6, GK1
+	"modeless",     // King's Quest 6
```
bluegr: Please change the description to King’s Quest 6 CD, like above

sluicebox: Good catch, comment updated.

bluegr:
Thanks for your work! Merging

PR #1407, opened 2018-11-20
SCI32: Fix QFG4 stairway pathfinding

Vhati:
Fixes walkable area when entering flipped stairways, bug [#10757](https://bugs.scummvm.org/ticket/10757)

    // When entering flipped stairways from the upper door, hero is initially
    // placed outside the walkable area. As a result, hero will float around
    // inappropriately in stairways leading to Tanya's room (620) and to the iron
    // safe (624).
    //
    // The polygon's first and final points are the top of the stairs. It's quite
    // narrow up there, and the final segment doesn't trace the wall very well. We
    // move the final point down and over to round out the path. Point 0 takes 19's
    // original place. Point 1 takes 0's original place.

Before and after pictures of the polygon are attached to the bug report.


Vhati:
I noticed I'd plotted hero in a different spot in the after picture. That mistake was cruft from tinkering with the python script I'd written to generate the image. The real spot turned out to be uncomfortably close to the edge, so I adjusted the patch, pushed, and reuploaded the after picture.

bluegr:
Thanks for your work! Works great, the approach taken to solve this is good. Merging

PR #1408, opened 2018-11-20
SCI32: Fix QFG4 setScale error when levitating

Vhati:
Fixes typos that lead to a "y value less than vanishingY" message, bug [#10726](https://bugs.scummvm.org/ticket/10726)

    // There are typos where hero is passed to Prop::setScale(), a method which
    // expects integers. We change it to Prop::setScaler().

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -141,8 +141,8 @@ static const char *const selectorNameTable[] = {
 	"fore",         // KQ7
 	"back",         // KQ7
 	"font",         // KQ7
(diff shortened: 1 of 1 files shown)


bluegr:
Have you tried this? Cause I had a similar change in s local branch, only to find that the objects in question did not have a setScaler method in QFG4

bluegr:
Right. Understood. Happy you fixed the issue correctly, seems my local patch was problematic.

Please chill out a bit :)

bluegr:
Just reviewed this. Nice work! Merging

PR #1409, opened 2018-11-20
3DS: Update port to current codebase

Nitrus:
Enables the Nintendo 3DS port to be compiled with recent implementation of devkitPro and devkitARM toolchain.


digitall:
Since all these patches look reasonable, the changes are limited to the 3DS port and there is no buildbot worker to be updated (currently), I am going to merge these as-is.

@Nitrus : Thanks for the code contribution.

PR #1415, opened 2018-11-23
BASE: Don't set fullscreen on Return to Launcher

sluicebox:
Return to Launcher currently resets kFeatureFullscreenMode. This undoes
switching to/from fullscreen through a hotkey. This happens if the
"fullscreen" config setting exists, which it doesn't on a fresh install,
which means that Return to Launcher's out-of-the-box behavior changes 
once a user visits Options and clicks OK, no matter what they enter and
even if they don't change anything.

This makes Return to Launcher not change kFeatureFullscreenMode, which
seems preferable, and makes behavior before and after saving Options
consistent. Issue #10810

### DIFF
--- a/base/main.cpp
+++ b/base/main.cpp
@@ -282,7 +282,7 @@ static Common::Error runGame(const Plugin *plugin, OSystem &system, const Common
 	return result;
 }
 
-static void setupGraphics(OSystem &system) {
+static void setupGraphics(OSystem &system, bool isInitialSetup) {
 
 	system.beginGFXTransaction();
 		// Set the user specified graphics mode (if any).
@@ -292,7 +292,7 @@ static void setupGraphics(OSystem &system) {
 
 		if (ConfMan.hasKey("aspect_ratio"))
 			system.setFeatureState(OSystem::kFeatureAspectRatioCorrection, ConfMan.getBool("aspect_ratio"));
-		if (ConfMan.hasKey("fullscreen"))
+		if (isInitialSetup && ConfMan.hasKey("fullscreen"))
 			system.setFeatureState(OSystem::kFeatureFullscreenMode, ConfMan.getBool("fullscreen"));
 		if (ConfMan.hasKey("filtering"))
 			system.setFeatureState(OSystem::kFeatureFilteringMode, ConfMan.getBool("filtering"));
@@ -459,7 +459,7 @@ extern "C" int scummvm_main(int argc, const char * const argv[]) {
 	if (settings.contains("disable-display")) {
 		ConfMan.setInt("disable-display", 1, Common::ConfigManager::kTransientDomain);
 	}
-	setupGraphics(system);
+	setupGraphics(system, true);
 
 	// Init the different managers that are used by the engines.
 	// Do it here to prevent fragmentation later
@@ -601,7 +601,7 @@ extern "C" int scummvm_main(int argc, const char * const argv[]) {
 		}
 
 		// reset the graphics to default
-		setupGraphics(system);
+		setupGraphics(system, false);
 		if (0 == ConfMan.getActiveDomain()) {
 			launcherDialog();
 		}


bluegr:
Shouldn't the same logic apply to the other parts that are initialized in setupGraphics(), i.e. aspect ratio, filtering and stretch mode?

criezy:
I have not tried it, but doesn't it break the behaviour when overriding the graphics settings for a game to set a different fullscreen mode than what we have in the launcher, when we return from this game?

For example if the launcher is using windowed mode but we edit the game settings to use fullscreen, when we start the game it switches to fullscreen, and when we return to the launcher we would expect to switch back to windowed. And I am wondering if the change here might not break this behaviour.

sluicebox:
@bluegr Maybe, I'm less familiar with the consequences of that, but any of those settings that can be toggled with a hotkey are currently being undone. If we can come up with improved fullscreen behavior it sounds right that it should be applied to those too.

@criezy That's a good example because right now, on a fresh install, that's what happens if you set a game's options and never visit the regular Options screen. (That could be considered a separate bug, that the initial global config state is different than once saved with defaults in GUI.) Yes, this change would do what you described once global options have been saved. I personally think that scenario is less useful than this but I know that's not the point.

Another approach could be to remember when a graphics setting is changed via hotkey and not revert that setting on Return to Launch.

The sequences I'd like handled are:

1. Alt+enter to toggle modes
2. Start game that has no graphics overrides
3. Return to Launcher doesn't change modes

And:

1. Start game
2. Alt+enter to toggle modes
3. Return to Launcher doesn't change modes

Do those sequences sound reasonable?

Right now there's a conflict between the fullscreen hotkey and the global fullscreen checkbox. The checkbox is "I want fullscreen" vs "I want a window" but what I want is "Just don't change it on me", which is the out of the box behavior that gets lost once any global changes are saved.

sluicebox:
The more I think on it, a better solution would involve tracking when hotkeys are used. If it weren't for them, I think everything is fine, so we can just close this.

I am still curious though if the sequences I'm proposing sound reasonable to others. If so I'll take another swing at this another time.

angstsmurf:
Just FYI, I made a tangentially related bug report about the conflict between the alt + enter hotkey and fullscreen settings some time ago: https://bugs.scummvm.org/ticket/9968

macca8:
As an end user, I’d like to offer some feedback based on observed behaviour and not the underlying code.

The current behaviour is fine and totally predictable, and already includes the desired behaviour as the factory default. 

The missing entry in the Config File on a fresh install appears intentional since this file behaves as a record of user preferences. Global options available in the Options Menu aren't added until the user clicks Apply or OK on an Options tab, and any removals revert to their initial (factory) value in the Options Menu. An option can retain its initial value and still be classed as a user preference.

The Alt-Enter hotkey overrides the default setting for the current environment (Launcher OR game), but doesn’t change the underlying default behaviour… creating a temporary preference.

This works well until it encounters a user preference upon switching to a different environment. In that case, the temporary preference is purged, and the user preference takes priority (though it may not be obvious when the user preference matches the temporary preference).

This actually makes sense, since the assumption is that the hotkey won’t be needed again because the user has applied preferred settings to the relevant 'Fullscreen mode' option and further screen mode changes won’t be necessary. The hotkey is still available for temporarily switching modes if needed.

This obviously interferes with the factory default behaviour, which is dependent on the user using the hotkey to change screen modes.

Removing the fullscreen entry from the Config File is all that’s needed to restore the desired behaviour. It’s easy enough for a user to do this manually, but a control to do this would make life easier.

I’ve noticed that whenever the Config File is updated, all Options with a checkbox style are added to the Config File, and any Options that have a value of 'default' are excluded.

Since we’re only targeting the 'Fullscreen mode' option, perhaps that option could be converted from a checkbox style to a dropdown menu style (with settings of default, Fullscreen & Windowed).

The Fullscreen & Windowed settings would be equivalent to the current enabled/disabled settings (and could retain the same 'fullscreen=true/false' entry in the Config File when present), and the 'default' setting would prevent the Fullscreen mode from being accidentally converted to a user preference, as is the case now if the Apply or OK buttons are clicked on any Options tab.

If I’ve completely underestimated the complexity of implementing this, I apologize. Hopefully I’ve at least provided some food for thought.

criezy:
I am going to close this now as I don't think the change is a good one.

As @macca8 wrote, the main issue here is that some settings, such as the fullscreen / windowed mode, are always saved when changing an options, and cannot be unset from the GUI (you need to edit the config file manually). So indeed this means there are in effect only two states, on (fullscreen) and off (windowed), with a third (unset) that is a bit hidden and that we might want to make more accessible. I am not sure what the best way to approach this is though. Using a dropdown menu for this option seems a bit strange, but with a checkbox I don't see how to handle an `option not set` state.

PR #1416, opened 2018-11-24
SCI32: Fix QFG4 conditional void calls

Vhati:
Fixes arithmetic hacks in combat, bugs [#10138](https://bugs.scummvm.org/ticket/10138), [#10419](https://bugs.scummvm.org/ticket/10419), [#10710](https://bugs.scummvm.org/ticket/10710), [#10814](https://bugs.scummvm.org/ticket/10814)

Supersedes commits 4dc9f0e and 01f3e6c

	// In QFG4, the kernel func SetNowSeen() returns void. In other words, it
	// doesn't modify the accumulator to store a result.
	//
	// The function updates boundary box properties of a given View object, prior
	// to collision tests. IF (collision detection is warranted, and IF those tests
	// are true), THEN respond to the collision.
	//
	// SetNowSeen() was inserted in the middle of the IF block's list of conditions.
	// That way it can be short-circuited, and it runs before the tests.
	//
	// Problem: void functions make no promise about their truth value. After the
	// call, acc will be *whatever* happened to already be in there. This is bad.
	//
	// "(| (SetNowSeen horror) $0001)"
	//
	// Original Solution: whatever it is, ensure it is non-zero. Do a bitwise OR
	// against a literal 1. Doesn't matter what leaked into acc. The OR wrapper
	// fakes a true return value so the IF block won't break.
	//
	// Someone exploited an unsafe compiler with a clever hack. It even worked when
	// an object snuck into the accumulator. ScummVM enforces safe arithmetic. Now
	// the code cleverly crashes - unless a number accidentally leaks into acc.
	//
	// We remove the OR wrapper and manually set the accumulator to true when the
	// call returns. Same effect. The IF block goes on evaluating conditions.

Discovery process is detailed on the ticket for [#10814](https://bugs.scummvm.org/ticket/10814).
The hack was used exclusively in combat code.
Savegames for each monster will be attached to the ticket.


Vhati:
Revised. No unwrapping with ldi 1. I left the wrapper in place and used push1 to feed OR a number instead, as I'd initially planned but mistakenly doubted last night.

bluegr:
Nice work! There is a variant of kSetNowSeen, which returns a boolean, which was used in newer SCI versions. Seems like the intended behavior in these scripts is to handle a boolean. Thus, the cleanest solution would be to use this variant of kSetNowSeen in QFG4, instead of patching all of the script locations where it is used.

m-kiewitz:
> 
> 
> Nice work! There is a variant of kSetNowSeen, which returns a boolean, which was used in newer SCI versions. Seems like the intended behavior in these scripts is to handle a boolean.

Those were my thoughts as well, but needs to be verified by looking into the interpreter.

Vhati:
@bluegr:
> There is a variant of kSetNowSeen, which returns a boolean, which was used in newer SCI versions. Seems like the intended behavior in these scripts is to handle a boolean.

My speculation is that was still void at the time, and needing to hack it factored into their decision to make it boolean later.

> the cleanest solution would be to use this variant of kSetNowSeen in QFG4

Probably.

Where not used as a conditional, the return value would be irrelevant. Grepping ALL occurrences of SetNowSeen... they never use it in a fashion that would store a result from it.

Where used as a conditional, the OR wrappers would never let it abandon the block. Relevant only insofar as it'd be a number.

So the end effect of booleanizing the func would be the same as this patch. False returns won't matter, but they'll still be numbers.

Vhati:
m-kiewitz:
> looking into the interpreter

I don't know how to do that myself. Presumably with a debugger build of DOSBox. No experience debugging 16bit exes, sierra or otherwise. I wouldn't know how to find the breakpoint. immediately after this kernel call.

I assume someone here could handle that?

m-kiewitz:
> I don't know how to do that myself. Presumably with a debugger build of DOSBox. No experience debugging 16bit exes, sierra or otherwise. I wouldn't know how to find the breakpoint. immediately after this kernel call.
> 
> I assume someone here could handle that easily enough?

SCI2 and SCI3 are 32-bit internally, I tried to get into contact with someone on the team, who should probably have IDA-files. Maybe @bluegr also has those. I don't know. It definitely has to be checked to make sure, and we should probably even look into the interpreters of other SCI2.1 games just to make sure.

I primarily worked/work on SCI16 (which is up to SCI1.1), I think I only have one for Gabriel Knight 1, which is why I can't look into it.

You can download IDA for free (there is a free version available, it's a disassembler), and look into the interpreter that way, but I'm afraid the free one is I think for 16-bit DOS ones, and not for SCI32 ones.

Vhati:
@m-kiewitz:
> SCI2 and SCI3 are 32-bit internally,

Oops, I misinterpreted this to mean 16bit then.
~~~~
file "sierra.exe"
MS-DOS executable ,\b LE for MS-DOS, DOS4GW DOS extender (embedded)
~~~~
64bit windows refused to run it. CheatEngine and OllyDbg choked opening it. It's not PE32, so I figured 16bit.

A trivial google says [DOS4GW](https://en.wikipedia.org/wiki/DOS/4G) is a "32bit DOS extender".

Vhati:
@m-kiewitz:
> I'm afraid the free one is I think for 16-bit DOS ones, and not for SCI32 ones.

Maybe this would be helpful to you?
.
.
[Thread](https://www.vogons.org/viewtopic.php?t=24744): Vogons - How can I place breakpoints on 32-bit code of DOS4GW bound executable file?

> DOS4GW bound EXE file consists of two parts: a 16-bit DOS stub that executes dos4gw.exe, and 32-bit LE(Linear Executable) that is loaded by the dos4gw.exe loader. Of course the LE part contains the application code

[Article](https://blog.torh.net/2015/10/30/disassemble-dos4gw/): Disassemble DOS/4GW

> since the game was using a DOS extender, also known as DOS/4GW, loading it directly into Ida Pro Free wouldn't help.
>
> The Solution
> I found the open source DOS extender [DOS/32](http://dos32a.narechk.net/index_en.html). This has a utility known as [SUNSYS Bind](http://dos32a.narechk.net/manual/html/util/1.htm) Utility which can be used to “Unbind and discard existing Extender/Stub from LE/LX/LC executable” as it says in the documentation.
> [...]
> This will produce **an unbinded file which can be loaded directly into Ida Pro Free**. In my case I ended up with a file ending in .LE, which stands for "Linear Executable".
>
> Hopefully this can help others who want to peak into old DOS games.

lskovlun:
> > Nice work! There is a variant of kSetNowSeen, which returns a boolean, which was used in newer SCI versions. Seems like the intended behavior in these scripts is to handle a boolean.
> 
> Those were my thoughts as well, but needs to be verified by looking into the interpreter.

No accumulator writes that I can see. So it still returns void at this point.

m-kiewitz:
> No accumulator writes that I can see. So it still returns void at this point.

Hi Lars,
Long time no see.
Very weird that it doesn't write anything?! So it may be even a special interpreter version?

lskovlun:
> > No accumulator writes that I can see. So it still returns void at this point.
> 
> Hi Lars,
> Long time no see.
> Very weird that it doesn't write anything?! So it may be even a special interpreter version?

QfG4 is simply very early SCI2. But as @bluegr states above, this was indeed added at a later time. 
A little investigation shows that it is not in SQ6 (interp dated June 19, 1995), but it is in the Torin demo (interp dated August 25, 1995). For full game releases it is in Phantasmagoria (interp dated September 19, 1995) and later. The question is, can we just add this feature for all of SCI32, or do we need to do it conditionally.

Vhati:
@lskovlun:
> No accumulator writes that I can see. So it still returns void at this point.

Glad to hear I wasn't just imagining this. ^_^

Vhati:
@m-kiewitz:
> Ah bugger. Now I understand. It doesn't SET anything, so it keeps the old ACC.

Exactly.
~~~~
// Problem: void functions make no promise about their truth value. After the
// call, acc will be *whatever* happened to already be in there. This is bad.
//
// "(| (SetNowSeen horror) $0001)"
//
// Original Solution: make whatever it is non-zero. Do a bitwise OR against 1.
// The OR wrapper fakes a true return value
~~~~


> You may of course see it as valid code, but for us it's buggy script code in a sense b/c not doing anything in those cases may result in serious issues down the line, which simply worked in the original interpreter.
>
>I would still keep the script patching then. Documentation needs to be updated though.

I don't understand your comment.

~~~~
// Someone exploited an unsafe compiler with a clever hack.[...] Now
// the code cleverly crashes - unless a number accidentally leaks into acc.
~~~~

Oh. You thought I was complaining about ScummVM's safety? No no no.
I meant whoever wrote the wrappers was being TOO clever, dangerous.

lskovlun:
> It should actually work on Quest for Glory 4 CD, because that one SHOULD get detected as SCI2.1 (early).

You missed the equals sign. The old (acc-preserving) behavior goes up to and including SCI 2.1 early (plus two explicitly named games). So the detection should be OK, it's just that the code improperly uses this SCI2.1 feature in older code. Perhaps the change was debated at the time, but didn't make it into the final game. Or conversely, it was taken out of the production interpreters for a while for some reason.

Vhati:
Revised: "clever" -> "too-clever".

Vhati:
Is everyone just too polite to tell me it's ridiculous that someone would hack a void into true?

I don't see a legitimate reason the author would add an OR wrapper if they expected a useful boolean return value from SetNowSeen().

@lskovlun:
> Perhaps the change was debated at the time, but didn't make it into the final game. Or conversely, it was taken out of the production interpreters for a while for some reason.

You were probably suggesting the wrapper was there all along and only coincidentally kept the script functional when the API switched to void, but...

A theory that the author initially wrote the func naked as a condition with boolean in mind, in a time of API debate, then wrapped at the last minute when the API settled on void is another interesting one.

Vhati:
As a sanity check, I edited my local source:
* SetNowSeen() returning boolean for QFG4
* patching away the push-ldi-OR with a jmp, so SetNowSeen()'s return value in acc (genuine this time) would be evaluated as a condition - potentially aborting the collision tests if the View weren't found to do collision tests against.

Combat scripts seemed fine without the wrapper when I compiled and tested.
Consistent with never having needed a wrapper under a boolean API.

Of course, it's always possible my tests missed something.

m-kiewitz:
> You missed the equals sign. The old (acc-preserving) behavior goes up to and including SCI 2.1 early (plus two explicitly named games). So the detection should be OK,

You missed my edit. I actually didn't miss it anyway. I thought the version was detected incorrectly, but a few seconds later I understood what was actually going on.
It keeps ACC from previous opcodes, which is why it holds an object and which is why we detect a problematic arithmetic operation because of that.

PR #1418, opened 2018-11-25
IOS: Restore and fix one-touch swipes to emulate arrow keys

MWumpusZ:
See "Controls" on http://wiki.scummvm.org/index.php/IPhone
Fixes https://bugs.scummvm.org/ticket/5914
    
The first commit restores the functionality that was already implemented once upon a time, but only works in portrait orientation.  The existing code attempts to compensate for the device orientation; I assume that this was necessary once upon a time, but no longer.  Removing that is in a separate commit, in case it has some context where it is required that I am not aware of.

### DIFF
--- a/backends/platform/ios7/ios7_osys_events.cpp
+++ b/backends/platform/ios7/ios7_osys_events.cpp
@@ -443,61 +443,22 @@ void  OSystem_iOS7::handleEvent_keyPressed(Common::Event &event, int keyPressed)
 bool OSystem_iOS7::handleEvent_swipe(Common::Event &event, int direction, int touches) {
 	if (touches == 1) {
 		Common::KeyCode keycode = Common::KEYCODE_INVALID;
-		switch (_screenOrientation) {
-		case kScreenOrientationPortrait:
-			switch ((UIViewSwipeDirection)direction) {
-				case kUIViewSwipeUp:
-					keycode = Common::KEYCODE_UP;
-					break;
-				case kUIViewSwipeDown:
-					keycode = Common::KEYCODE_DOWN;
-					break;
-				case kUIViewSwipeLeft:
-					keycode = Common::KEYCODE_LEFT;
-					break;
-				case kUIViewSwipeRight:
-					keycode = Common::KEYCODE_RIGHT;
-					break;
-				default:
-					return false;
-			}
-			break;
-		case kScreenOrientationLandscape:
-			switch ((UIViewSwipeDirection)direction) {
-				case kUIViewSwipeUp:
-					keycode = Common::KEYCODE_LEFT;
-					break;
-				case kUIViewSwipeDown:
-					keycode = Common::KEYCODE_RIGHT;
-					break;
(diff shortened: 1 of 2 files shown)


digitall:
@MWumpusZ : You will need to use git rebase -i master and drop the two merge commits on your branch and it should be fine.

@criezy: Can you review / test this? It looks reasonable, but am not an IOS developer / user.

criezy:
At first glance the change looks OK. I will try to find some time tomorrow evening to compile and test this.

dwatteau:
Hi,

I use ScummVM quite often on an iPhone 8, and I'd like to know what this commit is supposed to fix?

What I see with these changes is that most SCUMM games are now a pain to play, because now I need to use two fingers to move accurately the cursor on the screen, instead of one. Single-finger movements just "stop" after a short distance and are thus unusable, when I test this branch.

Is that the expected behavior? Playing the games just became more painful and complicated, IMO.

MWumpusZ:
@dwatteau This shouldn't break anything that worked before.

The iPhone port documentation (http://wiki.scummvm.org/index.php/IPhone) claims that the arrow keys can be simulated on iOS by single-finger swiping in the four main directions (up, left, right, down).  That hasn't worked for some time, and the patch should fix that.  Single-finger swipes were previously not being handled at all.

Moving by (single finger) tapping should not be affected at all.

Having said that, I tested with SCI and AGI games; I gather you are having issues with SCUMM games?  Can you tell me a specific game and action which works differently before and after?

MWumpusZ:
@dwatteau s issue *can* be worked around by touching "long enough" that the swipe recognizer won't trigger before beginning to move, but that doesn't feel like an acceptable workaround.

Purely technically, it's possible to recognize both actions, but so far I'm not entirely sure how to do that "right" from a usability perspective.  I don't think we want the swipe recognizer to trigger at all if the user is sliding their finger to move the mouse around.
I'll see if it's possible to get the swipe recognizer to trigger when the touch is released within the "swipe time" - at the moment it is triggering after a swipe motion, regardless of whether the touch is released or not.  That should be *better* for usability, but we'll have to see how the distinction feels in use.

MWumpusZ:
@dwatteau My initial intention was just to fix the age old bug report, since it seemed easy enough.  It turns out it's not as simple an issue as it looks :)

@criezy  My recent commit demonstrates that both gestures can be recognised from the same touches, but the result is not usable in my opinion.  I'm going to have a go at making the swipe recognizer only trigger when the touch is released, to see how that feels.  That'll probably take a bit though, free time is at a premium most days at the moment.  Can *I* mark this PR as "needs work" or otherwise indicate it shouldn't yet be merged?

criezy:
I an closing this now as a different change was merged two weeks ago that uses the three-fingers swipe to simulate the arrow keys (and changed the keyboard show/hide to use the pinch in/out gesture instead).

You are of course welcome to work on further improvements on the gestures, and open a new pull request for it.

PR #1420, opened 2018-11-26
DISTS: Fix typo

jwilk:
Slect → Select


digitall:
@jwilk : Thanks for your contribution. Merging as-is.

PR #1424, opened 2018-11-27
SCI32: Fix GK1 chair missing message, bug #10820

sluicebox:


digitall:
@sluicebox : This will need an interactive rebase to remove the new merge commit and rebase on top of the latest git master ... This is unavoidable due to the number of pending pull requests on the same part of the same file i.e. the SCI engine patch tables.

bluegr:
Nice work! It would be preferable to find a better signature method for this, but this will have to do. Merging

sluicebox:
Tell me about it! I'm so ashamed that that that's the largest offset in a script patch yet, but that method is adrift in a sea of copy and paste...

Thanks for doing all the merging, sorry those were all in the same place

PR #1425, opened 2018-11-27
SCI: Fix QFG1VGA antwerp wandering, bug #9564

sluicebox:
Fixes bugs in the original that lockup and crash the interpreter


bluegr:
Nice work! Merging

PR #1426, opened 2018-11-30
SCI32: Fix GK1 knitting basket obstacle bug #10707

sluicebox:
Fixes an obstacle not always being added to a room

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -1931,9 +1931,10 @@ static const uint16 gk1SysLoggerHotKeyPatch[] = {
 	PATCH_END
 };
 
-// After interrogating Gran in room 380, clicking on objects while seated
-//  causes Gabriel to briefly flicker into standing and other frames.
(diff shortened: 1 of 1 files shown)


engines/sci/engine/script_patches.cpp
```diff
@@ -1988,9 +1999,9 @@ static const uint16 gk1GranChairFlickerPatch[] = {
 	0x38, PATCH_UINT16(0x00af),         // pushi 00af
 	0x39, 0x75,                         // pushi 75
 	0x81, 0x00,                         // lag 0
+	0xa3, 0x00,                         // sal local0 [ local0 just needs to be non-zero ]
```
bluegr: Why does it need to be non-zero? Please add a comment here

sluicebox: Good call, I should have also explained the setCel/cel seelctor change so I updated the comments for both in the signature and patch.

bluegr:
Thanks for your work! Merging

PR #1427, opened 2018-11-30
SCI: Fix LB2 armor hall door pathfinding bug #9952

sluicebox:
Works around pathfinding algorithm difference that breaks the game


bluegr:
Same comment as with PR #1428:
Please add a comment that this script patch is a workaround. We can merge it for now, and remove it, together with other similar script patches, once we have a pathfinding algorithm that is 100% faithful to what the original SSCI did

bluegr:
**Update:** I'll merge this together with the other pathfinding related pull requests, and mark these and any other pathfinding related script patches we already have as workarounds

PR #1428, opened 2018-11-30
SCI: Fix SQ5 drive bay pathfinding, bug #7155

sluicebox:
Works around a pathfinding algorithm edge case

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -8907,9 +8907,35 @@ static const uint16 sq5PatchToolboxFix[] = {
 	PATCH_END
 };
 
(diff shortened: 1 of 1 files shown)


m-kiewitz:
I don't really like patching this out, because the actual problem is the different pathfinding algo (as you have already written as well).

That code needs to be rewritten. More see bug report.

bluegr:
Please add a comment that this script patch is a workaround. We can merge it for now, and remove it, together with other similar script patches, once we have a pathfinding algorithm that is 100% faithful to what the original SSCI did

bluegr:
**Update:** I'll merge this together with the other pathfinding related pull requests, and mark these and any other pathfinding related script patches we already have as workarounds

PR #1429, opened 2018-11-30
SCI32: Fix QFG4 peering at absent bats

Vhati:
Fixes castle doors when all bats have been killed, bug [#10789](https://bugs.scummvm.org/ticket/10789)

One patch for the lower door.
````
// In the room (644) connected to the lower door of the bat-infested stairway,
// peering through the keyhole *always* reports bats.
//
// As you kill the bats, global plot flags are set. Normally flags 331-334
// would be checked via proc0_4(). They're in a bitmask, among other flags, so
// we can check all simultaneously (0000000000011110).
//
// global[520] & 30 == 30
//
// Room 644 has an IF-ELSE deciding between largely redundant calls to
// gloryMessager::say(). We make room by combining them.
````

Three for the upper door.
````
// In the crest bookshelf room (663) connected to the upper door of the
// bat-infested stairway, peering through the keyhole *always* reports bats.
//
// As you kill the bats, global plot flags are set. Normally flags 331-334
// would be checked via proc0_4(). They're in a bitmask, among other flags, so
// we can check all simultaneously (0000000000011110).
//
// global[520] & 30 == 30
//
// Patch 1: There was no space for this in the responsible method. Instead, we
//  rewrite a different method that became obsolete after the crest bookshelf
//  patch: sCloseSecretDoor::changeState().
//
// Patch 2: We modify sPeepingTom to call our rewritten sCloseSecretDoor. This
//  has two variants, toggled to match the detected edition with enablePatch()
//  below. Aside from the patched lofsa value, they are identical.
//
// Requires patch: qfg4CrestBookshelf (CD or Floppy)
````

To test...
* Create a new character.
  * Rogues can't open castle doors unless they have lockpicks (send hero get 24).
  * Other classes are unhindered.
* Optional: To enter 663 via the secret passage, Spoof being in room 661 before teleporting there.
  * vv g 11 661
* Teleport: "room 644" or "room 663".
* Click HAND on the right door. Peer through the keyhole.
* Go east. Kill bats (click on them as they swoop).
* Return, and peer again.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -162,6 +162,7 @@ static const char *const selectorNameTable[] = {
 	"state",        // RAMA
 	"getSubscriberObj", // RAMA
 	"approachVerbs", // QFG4
+	"changeState",  // QFG4
 	"cue",          // QFG4
 	"heading",      // QFG4
 	"moveSpeed",    // QFG4
@@ -256,6 +257,7 @@ enum ScriptPatcherSelectors {
 	SELECTOR_state,
 	SELECTOR_getSubscriberObj,
 	SELECTOR_approachVerbs,
+	SELECTOR_changeState,
 	SELECTOR_cue,
 	SELECTOR_heading,
 	SELECTOR_moveSpeed,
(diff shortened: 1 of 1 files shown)


Vhati:
@Vhati:
> I don't know how to comment to a specific line

I believe you do that by looking at the code, and a blue plus appears beside the line numbers. Click that.

> Shouldn't [644] come after room 633 in the next line?

Yep. Thanks for noticing.

Vhati:
Added fixes for the upper room.

Vhati:
Switched to UINT16 for a couple large pushi values (>127).
I just learned 8bit pushi ops can wind up negative.

They'd worked here, but I prefer to avoid that ambiguity.

> op 0x39: pushi B data (2 bytes)
>
> Sign extension for 0x39 is performed where required.

bluegr:
I finally found some time to review this. Well done, rewriting a now unused function to save bytes is quite clever :)

Merging

PR #1430, opened 2018-12-01
WIN32: Fix missing export errors in Windows 98

SupSuper:
Fixes bug [#10613](https://bugs.scummvm.org/ticket/10613) by replacing API calls not available in Windows 98 with older equivalents.

You still need to compile with SDL 1.2 yourself, but otherwise it works: https://i.imgur.com/MEhve8U.png

### DIFF
--- a/backends/platform/sdl/win32/win32.cpp
+++ b/backends/platform/sdl/win32/win32.cpp
@@ -166,10 +166,8 @@ Common::String OSystem_Win32::getSystemLanguage() const {
 	char langName[9];
 	char ctryName[9];
 
-	const LCID languageIdentifier = GetUserDefaultUILanguage();
-
-	if (GetLocaleInfo(languageIdentifier, LOCALE_SISO639LANGNAME, langName, sizeof(langName)) != 0 &&
-		GetLocaleInfo(languageIdentifier, LOCALE_SISO3166CTRYNAME, ctryName, sizeof(ctryName)) != 0) {
+	if (GetLocaleInfo(LOCALE_USER_DEFAULT, LOCALE_SISO639LANGNAME, langName, sizeof(langName)) != 0 &&
+		GetLocaleInfo(LOCALE_USER_DEFAULT, LOCALE_SISO3166CTRYNAME, ctryName, sizeof(ctryName)) != 0) {
 		Common::String localeName = langName;
 		localeName += "_";
 		localeName += ctryName;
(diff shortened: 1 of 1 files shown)


backends/platform/sdl/win32/win32.cpp
```diff
@@ -192,7 +190,7 @@ Common::String OSystem_Win32::getScreenshotsPath() {
 	char picturesPath[MAXPATHLEN];
 
 	// Use the My Pictures folder.
-	if (SHGetFolderPath(NULL, CSIDL_MYPICTURES, NULL, SHGFP_TYPE_CURRENT, picturesPath) != S_OK) {
+	if (!SHGetSpecialFolderPath(NULL, picturesPath, CSIDL_MYPICTURES, FALSE)) {
```
bluegr: This function is marked as unsupported in MSDN

backends/platform/sdl/win32/win32.cpp
```diff
@@ -166,10 +166,8 @@ Common::String OSystem_Win32::getSystemLanguage() const {
 	char langName[9];
 	char ctryName[9];
 
-	const LCID languageIdentifier = GetUserDefaultUILanguage();
```
bluegr: Why did you just completely rip this call off? If this is not supported in Windows 98, it should be enclosed in a Windows version check, instead of being removed like this

SupSuper: Since we're only doing this call to pass the result to `GetLocaleInfo`, we can avoid it entirely by passing `LOCALE_USER_DEFAULT` directly to `GetLocaleInfo`, which works across all Windows versions.
Plus `GetUserDefaultUILanguage` returns a **`LANGID`** while `GetLocaleInfo` takes a **`LCID`,** so the original code looks suspect.

backends/platform/sdl/win32/win32.cpp
SupSuper: Well I doubt they support anything 9x era, but from my testing it still works as recently as Windows 10. If you think it's unreliable I can dynamically link to the function instead (a version check isn't enough since Win32 calls are linked at compile-time).

backends/platform/sdl/win32/win32.cpp
bluegr: You're quite right. We can just use LOCALE_USER_DEFAULT, instead of calling GetUserDefaultUILanguage()

backends/platform/sdl/win32/win32.cpp
bluegr: Indeed, it still works, and Microsoft is known to try and keep backwards compatibility with wrapper functions.

bluegr: And you are correct about the compile-time checks. How would you link dynamically to it?

SupSuper: You have to use `GetProcAddress` to get a function pointer instead, you can see an example in the Win32TaskbarManager: https://github.com/scummvm/scummvm/blob/master/backends/taskbar/win32/win32-taskbar.cpp#L381

bluegr: I believe that this is a good way to tackle this: check if SHGetFolderPath() exists and use it, otherwise fall back to SHGetSpecialFolderPath() for Windows 9x

sluicebox: Just wanted to chime in and say that detecting and calling the new function with GetProcAddress() and linking to and falling back on the old one is the right way to handle this. Also, fanatics who care about keeping things compiling on Win98 are very much my people. =)

SupSuper: @bluegr Replaced with the appropriate check.

digitall:
Since this should be good and @bluegr has now approved, merging this so that buildbot can check it...
@SupSuper : Thanks for looking at this.

PR #1431, opened 2018-12-03
SCI32: Fix QFG4 copy protection crash, bug #10773

sluicebox:


bluegr:
Nice work, thanks! Merging

engines/sci/engine/script_patches.cpp
```diff
@@ -8217,6 +8217,46 @@ static const uint16 qfg4ConditionalVoidPatch[] = {
 	PATCH_END
 };
 
+// The copy protection in floppy versions has a script bug which uses disposed
+//  objects and crashes our interpreter. This appears to work in Sierra's
+//  interpreter although they fixed the script bug in the CD version.
```
m-kiewitz: This part of the comment should get changed.

The Sierra interpreter used straight offsets directly to the data and allowed further access, which sometimes caused severe corruption.
We use a different approach. When references to objects or other data are free'd, they will cause an error when accessing that data reference afterwards (which is of course way better, that way we can detect script bugs).

sluicebox: What would you prefer? I don't see what's inaccurate about that high level description. Sierra fixed this the script bug by adding logic to not dispose these objects so that they wouldn't use disposed objects, so I don't know how this is anything other than a use-after-free class of bug that appears to work in the original.

Is your concern that the comment implies that it's a bug to re-use any object after disposal? I get that that's okay, but in this case that's specifically the bug: disposing sets View:plane to 0 so using that View it leads to kAddItemToScreen erroring. If "uses" were changed to "displays" would that be better?

m-kiewitz: > What would you prefer? I don't see what's inaccurate about that high level description.

Well, I personally don't like it when something says "crashes". For me "crashing" is an actual crash as in the code fully breaking, and not an error detection message. I don't think our interpreter actually crashes, but detects an error and gives out a detailed error message (correct me, if I'm wrong here, I haven't tested this one out personally. If it actually crashes that should get fixed, the interpreter should never really crash, no exception should ever occur).
And it should also mention that it works in the original interpreter because Sierra allowed accessing already disposed objects to some degree and we don't.

>Is your concern that the comment implies that it's a bug to re-use any object after disposal?
No, that's exactly right.
It is of course a script bug, when its trying to access an already disposed object. We already had several of those.

My concern is that it sounds like our interpreter would actually crash (as in literally crash, not end the game and showing an error message) and that it somehow works in the original interpreter (which implies that there is something wrong in ours, and not the other way).

At some point someone may read this comment and go "wait, is there some bug inside this interpreter? Why did it crash in this one while it worked fine in Sierras? Let's look into this and waste time).

We check and detect tons of things that Sierra did not check or care for. That's why some games had some random bugs happening at some random point in the game, which don't (shouldn't) happen in our interpreter at least after all of these script bugs are fixed/worked around.

sluicebox: Gotcha, makes sense, and I agree. I went back and forth on that word. Updated description in #1433

martiniturbide: Dear Martin, is there a way to contact you at your email?  I'm trying at the "@klinikum-konstanz.de" domain, but I'm not sure if you still have your email there. Mine is martiniturbide@gmail.com if you can write me back. Regards

PR #1432, opened 2018-12-03
SCI: fix full view remap effect

Kawa-oneechan:
The new stuff added to view.cpp for SCI11+ never actually triggers because I'd neglected to have `GfxAnimate::drawCels()` pass the scaleSignal that the remap effect requires, thus defaulting to zero.

This has been tested and correctly lets the work-in-progress full version of The Dating Pool draw characters standing in shadow.

### DIFF
--- a/engines/sci/graphics/view.cpp
+++ b/engines/sci/graphics/view.cpp
@@ -832,8 +832,8 @@ void GfxView::draw(const Common::Rect &rect, const Common::Rect &clipRect, const
 						// SCI16 remapping (QFG4 demo)
 						if (g_sci->_gfxRemap16 && g_sci->_gfxRemap16->isRemapped(outputColor))
 							outputColor = g_sci->_gfxRemap16->remapColor(outputColor, _screen->getVisual(x2, y2));
-						// SCI16+ remapping (Catdate)
-						if ((scaleSignal & 0x200) && g_sci->_gfxRemap16 && !g_sci->_gfxRemap16->isRemapped(outputColor))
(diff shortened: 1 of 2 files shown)


engines/sci/graphics/view.cpp
```diff
@@ -832,8 +832,8 @@ void GfxView::draw(const Common::Rect &rect, const Common::Rect &clipRect, const
 						// SCI16 remapping (QFG4 demo)
 						if (g_sci->_gfxRemap16 && g_sci->_gfxRemap16->isRemapped(outputColor))
 							outputColor = g_sci->_gfxRemap16->remapColor(outputColor, _screen->getVisual(x2, y2));
-						// SCI16+ remapping (Catdate)
-						if ((scaleSignal & 0x200) && g_sci->_gfxRemap16 && !g_sci->_gfxRemap16->isRemapped(outputColor))
+						// SCI11+ remapping (Catdate)
```
Kawa-oneechan: I guess bluegr _and_ I both missed these two comments.

engines/sci/graphics/view.cpp
```diff
@@ -832,8 +832,8 @@ void GfxView::draw(const Common::Rect &rect, const Common::Rect &clipRect, const
...
+						if ((scaleSignal & 0x200) && g_sci->_gfxRemap16)
```
Kawa-oneechan: Anything under 253 is of course never set to be remapped.

bluegr:
Thanks, merging

PR #1433, opened 2018-12-04
SCI32: Update QFG4 script patch description

sluicebox:
Updated patch comments to better explain error, as requested in #1431


bluegr:
Thanks! This looks good, merging it, and if any other adjustments to the comments are needed, they can be made in-tree.

PR #1436, opened 2018-12-05
SCI32: Fix QFG4 obscured ropes when rescuing Igor

Vhati:
Fixes a missing graphics update after priority is changed, bug [#10751](https://bugs.scummvm.org/ticket/10751)

````
// In the graveyard rescuing Igor, ropes are briefly obscured by crypt pillars
// in the background Pic. The Pic assigns a priority to the pillars for depth.
// Ropes are initialized without priority. Then there's a setPri() call.
//
// The floppy edition's Actor::doit() readily calls UpdateScreenItem(). Thus it
// promptly responds to the new priority, bringing the ropes to the front.
//
// The CD edition changed Actor to require a bit flag on the "signal" property
// before it would call UpdateScreenItem(). So the CD edition graphics don't
// update until much later, when the ropes begin an animation.
//
// We patch the heap for script 500 (the graveyard) to give rope1 and rope2
// that "signal" bit as soon as they're created.
````


bluegr:
Thanks for your work! Merging

PR #1437, opened 2018-12-05
SCI: Fix LONGBOW Amiga pub error, bug #9688

sluicebox:
Fixes a script bug which causes a kernel signature mismatch

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -3479,10 +3479,47 @@ static const uint16 longbowPatchBerryBushFix[] = {
 	PATCH_END
 };
 
+// The Amiga version of room 530 adds a broken fDrunk:onMe method which prevents
+//  messages when clicking on the drunk on the floor of the pub and causes a
+//  signature mismatch on every click in the room. fDrunk:onMe passes an Event
+//  object as an integer x coordinate and an uninitialized parameter as a y
+//  coordinate to kOnControl. This is a signature mismatch and would cause onMe
+//  to return false on every click and prevent hit testing from dispatching
+//  events to fDrunk. It's unclear why Sierra added this method to this one
+//  Feature in the room. Even if it worked, Feature:onMe already does this.
+//
+// We fix this by replacing fDrunk:onMe's contents with a call to super:onMe
+//  which calls kOnControl correctly and does proper hit testing, making its
+//  behavior consistent with the DOS version, which doesn't override onMe.
+//
+// Applies to: English Amiga Floppy
+// Responsible method: fDrunk:onMe
+// Fixes bug #9688
+static const uint16 longbowSignatureAmigaPubFix[] = {
+	SIG_MAGICDWORD,
+	0x67, 0x20,                     // pTos onMeCheck
+	0x39, 0x03,                     // pushi 03
+	0x39, 0x04,                     // pushi 04
+	0x8f, 0x01,                     // lsp 01
+	0x8f, 0x02,                     // lsp 02
+	0x43, 0x4e, 0x06,               // callk OnControl 6
+	SIG_END
+};
+
+static const uint16 longbowPatchAmigaPubFix[] = {
+	0x38, PATCH_UINT16(0x00c4),     // pushi 00c4 [ onMe, hard-coded for amiga ]
+	0x76,                           // push0
+	0x59, 0x01,                     // &rest 1
+	0x57, 0x2c, 0x04,               // super Feature 4 [ super: onMe &rest ]
+	0x48,                           // ret
+	PATCH_END
+};
+
 //          script, description,                                      signature                     patch
 static const SciScriptPatcherEntry longbowSignatures[] = {
 	{  true,   210, "hand code crash",                             5, longbowSignatureShowHandCode, longbowPatchShowHandCode },
 	{  true,   225, "arithmetic berry bush fix",                   1, longbowSignatureBerryBushFix, longbowPatchBerryBushFix },
+	{  true,   530, "amiga pub fix",                               1, longbowSignatureAmigaPubFix,  longbowPatchAmigaPubFix },
 	SCI_SIGNATUREENTRY_TERMINATOR
 };
 


m-kiewitz:
> Fixes a script bug which causes a kernel signature mismatch

Did this not work at all in the original interpreter?
I mean the code itself. What exactly happened in the original interpreter?

sluicebox:
It's a weird one because there doesn't seem to be any point to this script that they added, even if it did work.

They overrode onMe in this one inconsequential Feature in the room and replicated a small bit of the work that Feature:onMe already does, but did it wrong, and passed an object for the x parameter and a non-existent parameter 2 for the y parameter. The interpreter either ignored this or treated them as integers (I don't know which the amiga interpreter is doing under the hood) but either way the result is that fDrunk:onMe didn't return the correct answer, but from my tests nothing actually depends on this return value.

onMe is supposed to accept either an x and y parameter or a single object with x and y properties. fDrunk:onMe assumes that only the first one will happen, but the second one happens on every click.

sluicebox:
This is fDrunk:onMe:
```
(method (onMe param1 param2)
	(return (& onMeCheck (OnControl 4 param1 param2)))
)
```

m-kiewitz:
Well, you may also try to simply add a workaround entry for this one, making it skip the call.
If that works out, and makes it work properly as well, then I guess that may be better than a full script patch.
If it was broken in the original interpreter because of this script bug, there would be a reason to of course patch it out properly.

sluicebox:
I think the advantage of the script patch in this case is that it restores onMe to returning the correct value as it did in DOS before they overrode it with a broken one. Feature:onMe handles both integers and object parameters and does the right thing. 

I'm 99% sure that just skipping the kernel call would be fine too, but if I understand that option right then onMe would still be returning the wrong answer (which seems like it doesn't matter), but it seems like making it work is less likely to have side effects.

sluicebox:
Which workaround table/type should I try for this? I've only done uninit reads.

Vhati:
@sluicebox:
> I've only done uninit reads.

OT question: should I add workarounds for every "uninit read" warning I see, or are those workarounds only meant to deal with noticeable bugs?

m-kiewitz:
> Which workaround table/type should I try for this? I've only done uninit reads.

kOnMe currently does not have one.
I would have to add it.
Let me look into this during the upcoming weekend.

> OT question: should I add workarounds for every "uninit read" warning I see, or are those workarounds only meant to deal with noticeable bugs?

Please only add workaround entries when our interpreter error()s out. I think for some uninitialized reads it doesn't error() out. And there are ways to combine entries in various situations.

Vhati:
@m-kiewitz:
> only add workaround entries when our interpreter error()s out. I think for some uninitialized reads it doesn't error() out.

I see these in my status window, no noticeable ill effects. Do they qualify?
Or do you mean when the debugger drops down and ends the game?
````
WARNING: Uninitialized read for parameter 1 from method hero::changeGait (room 770, script 28, localCall ffffffff)!

WARNING: Uninitialized read for parameter 3 from method gloryMessager::nextMsg (room 790, script 64924, localCall ffffffff)!
````

**Comment by sluicebox** [comment]
Okay, now I see what's going on, this bug does have an effect on the game. It breaks messages in Amiga and the script patch is necessary to fix those. I'll update the comments.

I was confused because I only saw one drunk man at the pub at the table, and his messages work, but there's also a second one next to him who is difficult to see because he's just some dark feet on the floor sticking out from behind a wall. I never noticed him before! They have similar coordinates and are named tDrunk and fDrunk. fDrunk is the one on the floor that's causing problems. Too many drunks.

The fDrunk:onMe they added breaks hit testing since they're calling kOnControl with an object for an x and an uninitialized param for y. fDrunk:onMe would then always return false and so you wouldn't get any messages when clicking on the feet. I suppose if you were really unlucky it might return true when clicking somewhere else and then you'd get the wrong message. The script patch avoids the illegal kernel call and fixes the messages by restoring working hit testing.

**Comment by sluicebox** [comment]
Easy to see how a tester would miss that they'd broken the messages on this guy...

scummvm00020

File: engines/sci/engine/script_patches.cpp [hunk 1]
```diff
@@ -3479,10 +3479,46 @@ static const uint16 longbowPatchBerryBushFix[] = {
...
+// Applies to: English Amiga Floppy
+// Responsible method: fDrunk:onMe
+// Fixes bug #9688
+static const uint16 longbowSignatureAmigaPubFix[] = {
+	SIG_MAGICDWORD,
+	0x67, 0x20,                     // pTos onMeCheck
+	0x39, 0x03,                     // pushi 03
+	0x39, 0x04,                     // pushi 04
+	0x8f, 0x01,                     // lsp 01
+	0x8f, 0x02,                     // lsp 02
```
bluegr: This looks to be quite generic code. Could it also erroneously match with scripts of the PC version?

m-kiewitz:
> I see these in my status window, no noticeable ill effects. Do they qualify?
> Or do you mean when the debugger drops down and ends the game?

I mean the latter.
We added the warnings later. Those are uninitialized parameters.
This may still cause problems, but happens quite often and until now I think we had no situation where those caused any issues. I may be wrong on that.

The ones where the debugger actives and an actual error gets thrown, those are definitely problematic ones that caused tons of issues. At first we did not even detect those and used zero all the time. The original interpreter though reused memory. So in a few cases, the scripts really need some special number to be inside the uninitialized memory like 1 or 5 and when that's not in there the scripts go crazy.

That's why we added those workarounds later so that we can specify the number.

The warning for subroutine parameters is simply there to notice those too. They do not need fixing, unless they cause some script bug.

Vhati:
@m-kiewitz:
Understood. Thanks for clarifying.

m-kiewitz:
> 
> 
> @m-kiewitz:
> Understood. Thanks for clarifying.

Oh I see that I didn't specify the error() ones.
That's uninitialized variables, especially local variables.
Local variable block is defined, but some are not set. Sierra didn't zero out the local variable block, but kept whatever was in that memory area. We do not do that. We set zeroes, well in this case a special value, so that we can detect uninitialized reads.

The other ones are as I said parameters for subroutines.

engines/sci/engine/script_patches.cpp
sluicebox: I've tested against PC and the sequence is there. Those instructions are indeed generic but they come after pTos onMeCheck. I can include the next instruction, callk OnControl 6, to seal the deal. =)

bluegr:
Thanks for your work! Merging

PR #1438, opened 2018-12-06
SCI32: Fix QFG4 extraneous door sound in castle

Vhati:
Fixes a second door sound when entering certain rooms, bug [#10827](https://bugs.scummvm.org/ticket/10827)

````
// Rooms 622 and 623 play an extra door sound when entering. They both
// delegate to script 645. It schedules sEnter, which indeed has an extra
// sound. The CD edition removed the line. We remove it, too.
//
// Applies to at least: English floppy, German floppy
````

To test, teleport to either room.


bluegr:
Thanks for your work! Merging

PR #1439, opened 2018-12-07
SCI32: Fix GK1 mummy animation, bug #10828

sluicebox:
Fixes an animation script bug that occurs in the original


bluegr:
Nice work! Merging

PR #1440, opened 2018-12-07
SCI32: Fix QFG4 inventory leaks across restarts

Vhati:
Fixes item properties that weren't getting reset, bug [#10768](https://bugs.scummvm.org/ticket/10768)

````
// Some inventory item properties leak across restarts. Most conspicuously, the
// torch icon appears pre-lit after a restart, if it had been lit before.
//
// script 16 - thePiepan (item #28): loop, cel, value
// script 35 - theBroom (item #39): cel, value
// script 35 - theTorch (item #44): cel
//
// Glory::restart() tries to revert a bunch of globals to their original state.
// Each value was individually loaded into acc and assigned (ldi+sag, ldi+sag,
// ldi+sag, etc).
//
// One range of globals could be distilled to multiples of 45. We optimize
// those with a loop. Another range was arbitrary. We stack up those values all
// at once, then loop over the stack to pop(), assign, and do the next global.
// We use the freed bytes to reset the 3 items' properties with a subroutine.
````

To test...

* Create a character.
* Get the affected items and the means to modify them.
```
# These first three are available in the starting room.
theThrowdagger: send hero get 5
theFlint:       send hero get 21
theTorch:       send hero get 44

theHair:        send hero 38
theBroom:       send hero 39

theBerries:     send hero get 30
thePiePan:      send hero get 28
```
* Flint + torch = Torch icon will become lit.
* Hair + broom = Broom's icon will have hair woven through it. Weight increases.
* Berries + pan = Pan's icon will appear filled with ingredients. Weight increases.

* Restart. New character. Get the items again.
* Without this patch, the items' previous icon and weight will persist.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -156,7 +156,9 @@ static const char *const selectorNameTable[] = {
 	"plane",        // RAMA
 	"state",        // RAMA
 	"getSubscriberObj", // RAMA
+	"loop",         // QFG4
 	"setLooper",    // QFG4
+	"value",        // QFG4
 #endif
 	NULL
 };
@@ -237,7 +239,9 @@ enum ScriptPatcherSelectors {
 	SELECTOR_plane,
 	SELECTOR_state,
 	SELECTOR_getSubscriberObj,
-	SELECTOR_setLooper
+	SELECTOR_loop,
+	SELECTOR_setLooper,
+	SELECTOR_value
 #endif
 };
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: engines/sci/engine/script_patches.cpp [hunk 1]
```diff
@@ -8579,9 +8583,142 @@ static const uint16 qfg4DoubleDoorSoundPatch[] = {
...
+	0x3c,                               // dup       (global[378] = 8)
+	0x39, 0x05,                         // pushi 5d  (global[379] = 5)
+	0x39, 0x0a,                         // pushi 10d (global[380] = 10)
+	0x39, 0x0f,                         // pushi 15d (global[381] = 15)
+	0x39, 0x14,                         // pushi 20d (global[382] = 20)
+	0x39, 0x06,                         // pushi 6d  (global[383] = 6)
+	0x39, 0x08,                         // pushi 8d  (global[384] = 8)
+	0x39, 0x07,                         // pushi 7d  (global[385] = 7)
+	0x39, 0x0a,                         // pushi 10d (global[386] = 10)
+	0x3c,                               // dup       (global[387] = 10)
```
bluegr: Why do you need to initialize all these parameters here?

Vhati: The original asm pushed values and set specific globals, which happened to all be adjacent.

push&set, push&set, push&set, push&set...

I made room by optimizing.

push push push push... loop through the stack N times, each time setting the next global. The loop is shorter than all those individual assignment ops.

bluegr: Right. Please add that description to the comment, as it provides a better understanding of your changes

bluegr:
Thanks for your work! Merging

PR #1441, opened 2018-12-08
SCI32: Fix QFG4 town gate after a staff dream

Vhati:
Fixes hero getting stuck due to inactive nightfall detection, bug [#10830](https://bugs.scummvm.org/ticket/10830)

````
// Waking after a dream by the staff in town prevents the room from creating a
// doorMat at nightfall, if hero rests repeatedly. The town gate closes at
// night. Without the doorMat, hero isn't prompted to climb over the gate.
// Instead, hero casually walks south and gets stuck in the next room behind
// the closed gate.
//
// Since hero wakes in the morning, sAfterTheDream disposes any existing
// doorMat. It neglects to reset local[2], which toggles rm270::doit()'s
// constant checks for nightfall to replace the doorMat.
//
// We cache an object lookup and use the spare bytes to reset local[2].
//
// Note: There was never any sunrise detection. If hero rests repeatedly until
// morning, the doorMat will linger to needlessly prompt about climbing the
// then-open gate. Harmless. The prompt sets global[423] (1=climb, 2=levitate).
// The gate room only honors that global at night, so hero will simply walk
// through. Heroes unable to climb/levitate would be denied until they re-enter
// the room.
````

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -8579,6 +8579,77 @@ static const uint16 qfg4DoubleDoorSoundPatch[] = {
 	PATCH_END
 };
 
+// Waking after a dream by the staff in town prevents the room from creating a
+// doorMat at nightfall, if hero rests repeatedly. The town gate closes at
+// night. Without the doorMat, hero isn't prompted to climb over the gate.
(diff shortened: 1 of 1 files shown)


bluegr:
Thanks for your work! Merging

PR #1443, opened 2018-12-08
SCI32: Fix QFG4 iron safe room doors

Vhati:
Fixes mixed up oiled flags and east door sending hero west, bug [#10829](https://bugs.scummvm.org/ticket/10829)

````
// In the castle's iron safe room (643), the righthand door may send hero west
// instead of east - if it was oiled before it was opened (not picked).
//
// The room uses local[2] to remember which door it last decided was nearest.
// The proximity check when opening the right door doesn't reliably set
// local[2]. The assignment was buried inside an IF block testing the oiled
// flag to decide whether the door should squeak. So if the door's been oiled,
// local[2] is not set. If hero had entered the safe from from the west,
// rm643::init() would set local[2] to the left door, and sOpenTheDoor would
// remember LEFT as it decided where to send hero to next.
//
// We move the local[2] assignment out of the IF block, to always run.


// In the castle's iron safe room (643), plot flags are mixed up. When hero
// oils either door, the other door's flag is set. Adjacent rooms oil their
// respective doors properly from the outside. We switch the flags inside.
````


Vhati:
Sorry about rolling back your merge. Thought of an edit I needed squash.

git reset --hard 437701e
git rebase upstream/master
Fixed inconsistencies in the summary: "local2" to "local[2]".
git add "engines/sci/engine/script_patches.cpp"
git rebase --continue
git push origin +qfg4_safe_doors

bluegr:
Thanks for your work, merging

PR #1446, opened 2018-12-15
SHERLOCK: Speed up Blackwood cutscene in The Case of the Serrated Sca…

eriktorbjorn:
When capturing Blackwood in The Case of the Serrated Scalpel, the whole scene seems to be running in slow motion. This simply doubles the speed for it. It may still be a tad slow, but at least it fits the music much better I think.

I'm attaching a savegame (gzipped to get around GitHub's filter on unsupported file types) near the cutscene. If I didn't mess up the files, all you need to do to trigger the cutscene is to use the hammer on the door.

[scalpel.004.gz](https://github.com/scummvm/scummvm/files/2682556/scalpel.004.gz)

### DIFF
--- a/engines/sherlock/scalpel/scalpel.cpp
+++ b/engines/sherlock/scalpel/scalpel.cpp
@@ -1035,8 +1035,8 @@ void ScalpelEngine::startScene() {
 			// Blackwood's capture
 			_res->addToCache("final2.vda", "epilogue.lib");
 			_res->addToCache("final2.vdx", "epilogue.lib");
-			_animation->play("final1", false, 1, 3, true, 4);
-			_animation->play("final2", false, 1, 0, false, 4);
+			_animation->play("final1", false, 1, 3, true, 2);
+			_animation->play("final2", false, 1, 0, false, 2);
 			break;
 
 		case RESCUE_ANNA:


dreammaster:
Looks good to me. I'll merge it for now. If you do notice other cutscenes running slow, it can be revisited whether to speed them up likewise, or try for a global fix. I kind of like it best how this pull request does, since there's less chance of any speedups causing problems for Rose Tattoo as a side-effect.

eriktorbjorn:
> If you do notice other cutscenes running slow, it can be revisited whether to speed them up likewise, or try for a global fix. I kind of like it best how this pull request does, since there's less chance of any speedups causing problems for Rose Tattoo as a side-effect.

The cutscene where Moorehead gets killed also seems very slow. Again, changing the speed from 4 to 2 (this time for "SUBWAY1", "SUBWAY2" and "SUBWAY3") seems to make it run at a much more appropriate speed compared to the music.

In the attached savegame, simply pick up the typewriter to trigger the scene.

[scalpel.010.gz](https://github.com/scummvm/scummvm/files/2683767/scalpel.010.gz)

eriktorbjorn:
And here's a savegame for Lord Brumwell's suicide. Exit through the door to the left to trigger. Again, changing "suicid" from speed 4 to speed 2 makes it run for about as long as the music so that's probably the right speed for it.
[scalpel.014.gz](https://github.com/scummvm/scummvm/files/2683807/scalpel.014.gz)

eriktorbjorn:
The final cutscene. (Use the iron bar on the door to trigger it.) Changing "finalr1", "finalr2", "finale1", "finale2", "finale3" and "finale4" to speed seems appropriate here as well. Actually, it would probably match the music better if it was a little bit faster, but I don't have the time to fiddle with that now.

[scalpel.018.gz](https://github.com/scummvm/scummvm/files/2683835/scalpel.018.gz)

PR #1448, opened 2018-12-16
GLK: FROTZ: Proof of concept Unicode runes instead of bitmaps

eriktorbjorn:
This is probably premature, but just as a proof of concept I've added some code to remap the Beyond Zork runic characters (which use a bitmap font) to instead use Unicode runic characters when printed in the lower window. (Are they ever printed in the upper window? I honestly don't remember.) This should look a lot nicer. The same thing could probably also be done with the up/down arrow characters displayed at the beginning of the game.

To get the Unicode runic characters, I've added the Runic Noto font to fonts.dat and changed its version number to 1.1.

I'm attaching a gzipped savegame that I had hoped would demonstrate the scene in the game where Prince Foo and the black rider are speaking in runes if your intelligence is too low to understand them. When I played it on the Mac back in the day, this looked very strange because only lower-character letters have runes. The upper case characters came out as map drawing characters, etc.

Unfortunately, by the time the version I tested this with (Release 57) was released, Infocom had apparently realized this and changed the game to render it as normal text instead.

Still the savegame does demonstrate some other shortcomings of the ScummVM frotz implementation (too much spacing between lines in the upper window, no way to press up/down arrows to scroll the inventory, a "[MORE]" prompt in the upper window, etc.) So it stays.

[beyondzork.004.gz](https://github.com/scummvm/scummvm/files/2683489/beyondzork.004.gz)


dreammaster:
Overall, this looks good, and I was able to test it a bit by changing check against GRAPHICS_FONT to 1 in screen_char. However, the one thing I didn't like was the creation of a new core Glk style which will likely end up being just for Frotz, I've locally made a further change to your code to map runic onto the existing style_User2, and load the runic font in the FrotzScreen class.. this also makes it more consistent with the existing code that loads the character graphics font.

I'm also aware of the problem with spacing between lines in the upper window in Beyond Zork, but not the scrolling problem. Thanks for letting me know.

PR #1450, opened 2018-12-17
SCI32: Simplify GK1 ego speed script patch

sluicebox:
Simplifies a script patch into one version instead of two larger ones.

When I first wrote this I believed the fix (calling normalize) had to
take place in state 5 of the script, but state 4 is okay too. State 4
is simpler to patch and avoids the problem bytes that changed between
versions and required two patches.


bluegr:
Nice work! Merging

PR #1456, opened 2018-12-19
MSVC: Update project generation

SupSuper:
For the few of us actually building ScummVM with Visual Studio. 😉 

- Clean up `create_project` warnings. They were caused by the implicit casts when using `tolower`/`toupper` so I think they're fine silenced.
- Improve compile times by replacing the [/Gm flag](https://docs.microsoft.com/en-us/cpp/build/reference/gm-enable-minimal-rebuild) with the [/MP flag](https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes) which is better suited for large projects (shaved a whole minute off a clean build).
- Fix natvis known issues and add new classes.
- Fix `missing translations.dat` error when not using build events.
- Update build event scripts (old DLLs, old installer, etc).
- Update `ideprojects` target.

### DIFF
--- a/devtools/create_project/scripts/postbuild.cmd
+++ b/devtools/create_project/scripts/postbuild.cmd
@@ -4,27 +4,27 @@ REM ---------------------------------------------------------------
 REM -- Post-Build Script
 REM ---------------------------------------------------------------
 REM
-REM Copy engine data, themes, translation and required dlls to the
+REM Copy engine data and required dlls to the
 REM output folder and optionally create an installer
 REM
 REM Expected parameters
 REM    Root folder
 REM    Output folder
-REM    Architecture
+REM    SDL version
 REM    Libs folder
 REM    Installer ("1" to build, "0" to skip)
 
 if "%~1"=="" goto error_root
 if "%~2"=="" goto error_output
-if "%~3"=="" goto error_arch
+if "%~3"=="" goto error_sdl
 if "%~4"=="" goto error_libs
 if "%~5"=="" goto error_installer
 
 echo Copying data files
 echo.
 
-xcopy /F /Y "%~4/lib/%~3/SDL.dll"                          "%~2" 1>NUL 2>&1
-xcopy /F /Y "%~4/lib/%~3/freetype6.dll"                    "%~2" 1>NUL 2>&1
+xcopy /F /Y "%~4/%~3.dll"                                  "%~2" 1>NUL 2>&1
+xcopy /F /Y "%~4/WinSparkle.dll"                           "%~2" 1>NUL 2>&1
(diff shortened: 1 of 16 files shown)


sluicebox:
I tested this with VS2015 and it significantly speeds up compiles, even when few engines are enabled. The translations.dat warning is finally gone too as that is now being correctly embedded as a resource. Great job!

digitall:
This looks good and is limited to MSVC Win32 port generation code so should be fine to merge and deal with any issues in tree. Will leave it a few more days for further comment / review.

devtools/create_project/scripts/postbuild.cmd
```diff
@@ -4,37 +4,36 @@ REM ---------------------------------------------------------------
...
 echo Copying data files
 echo.
 
-xcopy /F /Y "%~4/lib/%~3/SDL.dll"                          "%~2" 1>NUL 2>&1
-xcopy /F /Y "%~4/lib/%~3/freetype6.dll"                    "%~2" 1>NUL 2>&1
+xcopy /F /Y "%~4/%~3.dll"                                  "%~2" 1>NUL 2>&1
+xcopy /F /Y "%~4/WinSparkle.dll"                           "%~2" 1>NUL 2>&1
 xcopy /F /Y "%~1/backends/vkeybd/packs/vkeybd_default.zip" "%~2" 1>NUL 2>&1
 xcopy /F /Y "%~1/backends/vkeybd/packs/vkeybd_small.zip"   "%~2" 1>NUL 2>&1
-xcopy /F /Y "%~1/gui/themes/translations.dat"              "%~2" 1>NUL 2>&1
```
bluegr: What is the reason for removing freetype6.dll and translations.dat here?

bluegr:
Very nice work! It's really good that you updated all this functionality. The new build flags and the generated natvis files will help a lot with debugging.

I've tested this, and it improves both the build time and the natvis-assisted debugging. Overall, it's a step in the right direction, as it addresses several issues with the MSVC builds, without introducing obvious regressions or issues, and if we spot any, we can fix them in tree. This has been open for 3 days, which include a weekend, so I don't suppose there will be any further comments on this.

I've talked with @digitall, and he's also OK with merging this.

Merging

devtools/create_project/scripts/postbuild.cmd
SupSuper: The translations.dat is built into the executable now as part of dists/scummvm.rc, and there's no longer a freetype6.dll in scummvm_libs (seems to be statically-linked now).
The vkeybd_* files might also be obsolete (if necessary they should be added to the .rc) but I couldn't find any references to them so left them alone.

PR #1457, opened 2018-12-20
SCI: Fix LONGBOW Day 7 Tuck/Net scene, bug #10847

sluicebox:
Fixes a bug in the original that plays a scene when it shouldn't


bluegr:
Nice catch, good work! Merging

PR #1458, opened 2018-12-23
SCI32: Fix QFG4 rations icon removal after eating

Vhati:
Fixes an icon lingering in the verb bar after scheduled meals, bug [#10772](https://bugs.scummvm.org/ticket/10772)

````
// Rations are not properly decremented by daily scheduled meal consumption.
// Rations are consumed periodically as time advances. If rations are the
// active inventory item when the last of them is eaten, that icon will persist
// in the verb bar.
//
// We make room by caching values in temp variables and by consolidating
// gloryMessager::say() args in a subroutine. Then we add code for cleaning up
// the verb bar when rations are exhausted (test active item, advance cursor,
// hide item icon) - similar to the localproc in script 16, called by
// interactive items to remove themselves.
````
Pseudocode for the patch is [here](https://bugs.scummvm.org/ticket/10772#comment:6).

Sorry about the length. That's as short as I could manage.
Clearing the icon is not a trivial operation, and the freed bytes were distributed across several blocks, which had to be written in their condensed forms.

&nbsp;

To test...
* Acquire or drop rations until only one remains.
    * "send hero get 4"
    * "send hero drop 4 1"
* Optional: adjust stamina and health. The status bars may take a sec to notice.
    * vv g 265 0
    * vv g 264 0
* Select the rations, so they appear in the verb bar.
* Either rest or idle until evening when rations are automatically eaten...
* Or eat immediately with the debugger: "send hero eatMeal"

Clicking rations on hero will NOT run eatMeal(), but will instead increment a global that tracks preemptively eaten rations. When eatMeal() eventually runs, it decrements that global in lieu of consuming another ration.

Repeating calls to eatMeal() without rations causes several messages with increasing urgency, stamina loss, health loss, and ultimately death.

The death screen starts out blank for a bit when triggered via the debugger. The screen is fine when idly standing around and keeping stamina/health at 0 - fighting regeneration with the debugger - until eatMeal() runs naturally.

### DIFF
--- a/po/be_BY.po
+++ b/po/be_BY.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: ScummVM 1.8.0git\n"
 "Report-Msgid-Bugs-To: scummvm-devel@lists.scummvm.org\n"
-"POT-Creation-Date: 2018-12-18 23:42+0100\n"
+"POT-Creation-Date: 2018-12-19 07:31+0100\n"
 "PO-Revision-Date: 2018-08-06 21:35+0000\n"
 "Last-Translator: Fiodar <fedor_qd@mail.ru>\n"
 "Language-Team: Belarusian <https://translations.scummvm.org/projects/scummvm/"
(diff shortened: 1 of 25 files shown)


Vhati:
Travis has a strange complaint.
````
engines/sci/engine/script_patches.cpp:263:2: error: redefinition of enumerator
      'SELECTOR_owner'
        SELECTOR_owner,
        ^
engines/sci/engine/script_patches.cpp:216:2: note: previous definition is here
        SELECTOR_owner
        ^
    C++      engines/sci/engine/selector.o
1 error generated.
make: *** [engines/sci/engine/script_patches.o] Error 1
````

That selector only appears once in the file ([line 259](https://github.com/Vhati/scummvm/blob/4bfccab7c73ccf69dfba2d21be06d1b9db5a333a/engines/sci/engine/script_patches.cpp#L259)), and it builds successfully for me.

Vhati:
Ah, I'm having trouble fetching the latest commits.
9fec79a had added owner for Longbow.
My branch didn't see that.

PR #1459, opened 2018-12-23
SCI32: Fix QFG4 rations icon removal after eating

Vhati:
Fixes an icon lingering in the verb bar after scheduled meals, bug [#10772](https://bugs.scummvm.org/ticket/10772)

````
// Rations are not properly decremented by daily scheduled meal consumption.
// Rations are consumed periodically as time advances. If rations are the
// active inventory item when the last of them is eaten, that icon will persist
// in the verb bar.
//
// We make room by consolidating common gloryMessager::say() args in a
// subroutine and cache values with temp variables. We add code to clean up
// the verb bar when rations are exhausted (test if rations were the active
// item, advance the cursor, hide the bar's invItem icon) - similar to the
// localproc in script 16, called by combinable items to remove themselves.
````
Pseudocode for the patch is [here](https://bugs.scummvm.org/ticket/10772#comment:6).

Sorry about the length. That's as short as I could manage.
Clearing the icon is not a trivial operation, and the freed bytes were distributed across several blocks, which had to be written in their condensed forms.

&nbsp;

To test...
* Acquire or drop rations until only one remains.
    * "send hero get 4"
    * "send hero drop 4 1"
* Optional: adjust stamina and health. The status bars may take a sec to notice.
    * vv g 265 0
    * vv g 264 0
* Select the rations, so they appear in the verb bar.
* Either
    * Rest or idle until evening when rations are automatically eaten...
    * Or eat immediately with the debugger: "send hero eatMeal"

&nbsp;

Clicking rations on hero will NOT run eatMeal(), but will instead increment a global that tracks preemptively eaten rations. When eatMeal() eventually runs, it decrements that global in lieu of consuming another ration.

Repeating calls to eatMeal() without rations causes several messages with increasing urgency, stamina loss, health loss, and ultimately death.

The death screen starts out blank for a few seconds when triggered via the debugger. However, the screen is fine when idly standing around and keeping stamina/health at 0 - fighting HP regeneration with the debugger - until eatMeal() runs naturally.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -124,7 +124,7 @@ static const char *const selectorNameTable[] = {
 	"ignoreActors", // Laura Bow 1 Colonel's Bequest
 	"setVol",       // Laura Bow 2 CD
 	"at",           // Longbow, QFG4
-	"owner",        // Longbow
+	"owner",        // Longbow, QFG4
 	"delete",       // EcoQuest 1
 	"size",         // EcoQuest 1
 	"signal",       // EcoQuest 1, GK1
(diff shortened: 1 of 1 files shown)


digitall:
@bluegr: This PR seems to have gone through a fair number of force-pushes to correct, but if you have time / energy, it should be ready for review as has been stable for over 2 weeks...

Vhati:
Hi, sorry 'bout my absence.
My bug hunting spree lost momentum there. Focus drifted.

Most of the pushes were just catching up with master.
This PR's been functionally unchanged for months.

I totally understand why it'd be a low priority to review though: subtle bug, complex patch.

bluegr:
I finally found some time to review this. The changes are good, and easy to understand. Thanks for working on this :)

Merging

PR #1461, opened 2018-12-24
SCI32: Fix QFG4 statue item description

Vhati:
Fixes an inaccurate message when examining the statue, bug [#10770](https://bugs.scummvm.org/ticket/10770)

````
// Examining the statue inventory item from the monastery - bug #10770
// The description says "squid-like monster", yet the icon is
// clearly an insect. It turned Chief into "an enormous beetle". We
// change the phrase to "monstrous insect".
//
// Note: The German string contains accented characters.
//  0x84 "a with diaeresis"
//  0x94 "o with diaeresis"
//
// Values were pulled from SCI Companion's raw message data. They
// are escaped that way here, as encoded bytes.
````
This PR intercepts a message lookup and returns a hardcoded string, matching the language of the detected edition.

English
* Before: "Carefully wrapped in a shopping bag is the grotesque sculpture of a horrible squid-like monster."
* After: "Carefully wrapped in a shopping bag is the grotesque sculpture of a horrible, monstrous insect."

German
* Before: "Die groteske Skulptur eines schrecklichen, tintenfischartigen Ungeheuers ist sorgf&auml;ltig in die Einkaufstasche eingewickelt."
* After: "Die groteske Skulptur eines schrecklichen, monstr\x94sen insekts ist sorgf\x84ltig in die Einkaufstasche eingewickelt."

&nbsp;

To test...
* Restore or create a new character.
* Acqure the statue.
    * send hero get 43
* Bring up the inventory window and click EYE on the statue.

### DIFF
--- a/engines/sci/engine/message.cpp
+++ b/engines/sci/engine/message.cpp
@@ -249,6 +249,33 @@ bool MessageState::getRecord(CursorStack &stack, bool recurse, MessageRecord &re
 			t.cond = 13;
 		}
 
+		if (g_sci->getGameId() == GID_QFG4 && stack.getModule() == 16 &&
+			t.noun == 49 && t.verb == 1 && t.cond == 0 && t.seq == 2) {
+			// Examining the statue inventory item from the monastery - bug #10770
+			// The description says "squid-like monster", yet the icon is
+			// clearly an insect. It turned Chief into "an enormous beetle". We
+			// change the phrase to "monstrous insect".
+			//
+			// Note: The German string contains accented characters.
+			//  0x84 "a with diaeresis"
+			//  0x94 "o with diaeresis"
+			//
+			// Values were pulled from SCI Companion's raw message data. They
+			// are escaped that way here, as encoded bytes.
+			record.tuple = t;
+			record.refTuple = MessageTuple();
+			record.talker = 99;
+			if (g_sci->getSciLanguage() == K_LANG_GERMAN) {
+				record.string = "Die groteske Skulptur eines schrecklichen, monstr\x94sen insekts ist sorgf\x84ltig in die Einkaufstasche eingewickelt.";
+				record.length = 112;
+			} else {
+				record.string = "Carefully wrapped in a shopping bag is the grotesque sculpture of a horrible, monstrous insect.";
+				record.length = 95;
+			}
+			delete reader;
+			return true;
+		}
+
 		if (g_sci->getGameId() == GID_QFG4 && stack.getModule() == 579 &&
 			t.noun == 0 && t.verb == 0 && t.cond == 0 && t.seq == 1) {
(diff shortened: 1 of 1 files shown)


lskovlun:
This is the first time we have a patch that fixes what looks like a "writing bug" rather than a "coding bug" (broadly speaking). I feel we should have a discussion on whether we want to include this sort of fixes. Some of the other engines do that, but then again, some of the other engines have hardcoded texts too, so there is a "necessity" (for lack of a better word) factor there.

Also, if we choose to merge this, the German does need fixing. I'm not a native speaker myself, though, so I'll refrain from proposing a change here. @m-kiewitz ?

m-kiewitz:
"After: "Die groteske Skulptur eines schrecklichen, ungeheuerlich insekt ist sorgf\x84ltig in die Einkaufstasche eingewickelt."

Please change to:
After: "Die groteske Skulptur eines schrecklichen, monströsen insekts ist sorgf\x84ltig in die Einkaufstasche eingewickelt."

Umlaut needs to be replaced of course, but no idea which char is correct.

m-kiewitz:
And I would prefer a proper patch table.
@lskovlun
I need your help on that plane issue, we got 1 serious bug still happening in PQ:SWAT.

Vhati:
@m-kiewitz:
> Please change to: monströsen insekts

Done. Thanks!

> Umlaut needs to be replaced of course, but no idea which char is correct.

Can you elaborate?
"\x84" displays correctly in-game, so whatever is downstream understands the mysterious encoding.

Byte 0x84 cannot exist in ascii or unicode text files, such as message.cpp. It needs to be abstracted, if only as a hex literal of the encoded byte. A text file encoded in the Windows-1252 flavor of ANSI renders it as "„" .

Abstract codepoints are a unicode feature. While it's good practice to parse arbitrary bytes and decode to unicode strings in memory, that's either not happening, or this function is part of the early parsing phase?

Do you mean "\\nn" filtered through MessageState::stringHex() and hexDigitToWrongInt()?

Are there named constants to concatenate?

m-kiewitz:
"monströsen" contains an Umlaut, which probably needs to be replaced/escaped.

\x84 is for the umlaut "ä". I don't know which char is the correct one for "ö".
ScummVM itself does not care in that regard. It's about Sierra's internal fonts.

Vhati:
@m-kiewitz:
I realized 40 minutes late. Reloaded and saw your response. =)
Character replaced.

Vhati:
Added values for the rest of MessageRecord's properties and rearranged the if-else to only contain language-dependent lines, with shared code before & after.

bluegr:
Thanks for your work!

If we have other cases where messages need to be adapted, then a patch table will be needed. For a single case like this one, it won’t be necessary.

Merging

PR #1462, opened 2018-12-24
SCI32: Fix QFG4 Rusalka flowers dialogue

Vhati:
Fixes mixed up text/audio when giving her flowers, bug [#10849](https://bugs.scummvm.org/ticket/10849)

````
// The CD edition mangled the Rusalka flowers dialogue.
// In the floppy edition, there are 3 lines, the first from
// the narrator, then two from Rusalka. The CD edition omits
// narration and only has the 3rd text, with the 2nd audio! The
// 3rd audio is orphaned but available.
//
// We only restore Rusalka's lines, providing the correct text
// for seq:1 to match the audio. We respond to seq:2 requests
// with Rusalka's last text. The orphaned audio (seq:3) has its
// tuple adjusted to seq:2 in resource_audio.cpp.
````

To test...
* Restore or create a character.
* Acquire flowers.
    * send hero get 40
* Teleport to the lake.
    * room 520
* Click the flowers on Rusalka.

### DIFF
--- a/engines/sci/engine/message.cpp
+++ b/engines/sci/engine/message.cpp
@@ -255,6 +255,42 @@ bool MessageState::getRecord(CursorStack &stack, bool recurse, MessageRecord &re
 			t.verb = 1;
 		}
 
+		if (g_sci->getGameId() == GID_QFG4 && g_sci->isCD() && stack.getModule() == 520 &&
+			t.noun == 2 && t.verb == 59 && t.cond == 0) {
+			// The CD edition mangled the Rusalka flowers dialogue. - bug #10849
+			// In the floppy edition, there are 3 lines, the first from
+			// the narrator, then two from Rusalka. The CD edition omits
+			// narration and only has the 3rd text, with the 2nd audio! The
+			// 3rd audio is orphaned but available.
+			//
+			// We only restore Rusalka's lines, providing the correct text
+			// for seq:1 to match the audio. We respond to seq:2 requests
+			// with Rusalka's last text. The orphaned audio (seq:3) has its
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: engines/sci/engine/message.cpp [hunk 1]
```diff
@@ -255,6 +255,37 @@ bool MessageState::getRecord(CursorStack &stack, bool recurse, MessageRecord &re
...
+				record.talker = 28;
+				record.string = "Thank you for the beautiful flowers.  No one has been so nice to me since I can remember.";
+				record.length = 89;
+				delete reader;
+				return true;
+			} else if (t.seq == 2) {
+				record.tuple = t;
+				record.refTuple = MessageTuple();
+				record.talker = 28;
+				record.string = "I guess I shouldn't try to drown you now, since you've been so kind.";
```
bluegr: This message exists in module 520, with the following conditions:
noun: 2, verb: 59, cond: 0, seq: 1, talker: 28

It would be preferable to just change seq to 1 in this case, instead of hardcoding the text

Vhati: @blugr:
> This message exists in module 520.

I tried that at first. It causes an endless loop.
If seq:2 becomes 1, another attempt is made after that to look up seq:2, and so on...

Vhati:
I suspect relying on recursion, pushing refs onto the cursorStack, and using the stack size to end a loop would be problematic, if it would even work. Seems like messages are being looked up individually as it is.

I could be mistaken. Haven't wrapped my head around the recursive behavior.

bluegr:
Thanks for your work! Merging

PR #1466, opened 2018-12-27
SCI32: Fix GK1 Wolfgang phone obstacle bug, #10853

sluicebox:
Fixes bug in original which places ego outside the room's obstacle boundary


bluegr:
Nice work! Merging

PR #1467, opened 2018-12-27
SCI32: Plane id fixes 

lskovlun:
This pull request provides a proper fix for bug #10773, #10778 and possibly other bugs that don't have tickets.

SCI32 creates a number of internal planes for various purposes. In most interpreters, these planes have ID numbers starting at 20000. These planes do not have a corresponding script object, and are denoted by (scummvm) in the output of the plane_list command. One such plane (and the only one relevant to GK1 and QfG4) is called initPlane in the scummvm source code. Most memory handles are smallish, so starting at 20000 makes conflicts with script-created planes much less likely (but not impossible). In the early interpreters of GK1 and QfG4 floppy, the numbering actually starts at zero. This means that operations on screen items with plane == 0 will operate on the initPlane implicitly. As more features were implemented that used this kind of plane, conflicts became likely, which probably explains why Sierra changed it. This also explains why they had to fix this bug even though it was asymptomatic in SSCI 2.0: As soon as they rebuilt the game with SCI2.1 for the CD release, the bug would have become symptomatic.

### DIFF
--- a/engines/sci/console.cpp
+++ b/engines/sci/console.cpp
@@ -488,6 +488,9 @@ bool Console::cmdGetVersion(int argc, const char **argv) {
 	debugPrintf("Resource map version: %s\n", g_sci->getResMan()->getMapVersionDesc());
 	debugPrintf("Contains selector vocabulary (vocab.997): %s\n", hasVocab997 ? "yes" : "no");
 	debugPrintf("Has CantBeHere selector: %s\n", g_sci->getKernel()->_selectorCache.cantBeHere != -1 ? "yes" : "no");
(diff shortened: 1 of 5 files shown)


sluicebox:
Thanks for figuring this out, I'm happy to see those script patches go. I tested the two QFG4 issues and this fixes both.

lskovlun:
btw, it is possible that PQ4 floppy is in fact affected by this. I can't check (only have the CD version), so I've left it out for now.

bluegr:
Thanks for your work @lskovlun! A much better solution than what we got. I've tested this out, works great.

Merging

PR #1470, opened 2018-12-30
SCI: Fix QFG1VGA Mac Antwerp controls, bug #10856

sluicebox:
Fixes a script bug in the mac version that always kills the player


bluegr:
Nice work! Merging

PR #1472, opened 2019-01-03
SCI32: Stub out kRestartGame

lskovlun:
Restarting was only supported in QfG4, GK1 and PQ4. Those three used
makeshift script code. The kernel call was stubbed out in SSCI as
well. Fixes bug #10681.


lskovlun:
I'm thinking maybe we should rename kRestartGame to kRestartGame16, just to clarify what's going on. We already have several other kernel functions using that convention.

m-kiewitz:
Fine by me.

bluegr:
Agree with the change, and also agree to rename kRestartGame to kRestartGame16. Can you please add that change in this pull request, for completeness?

m-kiewitz:
@lskovlun
btw. I won an auction today for Police Quest SWAT Japanese!
A few other Japanese SCI32 releases are left, but very hard to track down.

PR #1473, opened 2019-01-04
SCI32: Fix QFG4 forest pathfinding

Vhati:
Adds workarounds for odd detours during entry

[#10857](https://bugs.scummvm.org/ticket/10857) - Script patch
````
// When entering forest room 557 from the east (563), hero is supposed to move
// only a short distance into the room. ScummVM's pathfinding sends hero off
// course, to the middle of the room and back.
//
// There's an unwalkable stream in the SE corner, and hero's coords were within
// its polygon. We lower the top two points to keep hero on the outside.
````
Side note: Hand decompiled code for the forest::init() is attached there, in case in case anyone is curious about random encounters and pre-combat setup. It's shared across all the forest rooms.

&nbsp;

[#10858](https://bugs.scummvm.org/ticket/10858) - AStar() workaround

ScummVM's pathfinding has a penalty that discourages travel to vertices on a screen edge. 
QFG1VGA already had a boolean to exempt one room.
I reformatted it in a way that can add additional rooms.
````
// WORKAROUND: This check is needed in SCI1.1 games, such as LB2. Until our
// algorithm matches better what SSCI is doing, we exempt certain rooms where
// the check fails.
bool penaltyWorkaround =
	// QFG1VGA room 81 - Hero gets stuck when walking to the SE corner (bug #6140).
	(g_sci->getGameId() == GID_QFG1VGA && g_sci->getEngineState()->currentRoomNumber() == 81) ||
#ifdef ENABLE_SCI32
	// QFG4 room 563 - Hero zig-zags into the room (bug #10858).
	// Entering from the south (564) off-screen behind an obstacle, hero
	// fails to turn at a point on the screen edge, passes the poly's corner,
	// then approaches the destination from deeper in the room.
	(g_sci->getGameId() == GID_QFG4 && g_sci->getEngineState()->currentRoomNumber() == 563) ||
#endif
	false;

if (s->pointOnScreenBorder(vertex->v) && !penaltyWorkaround)
	new_dist += 10000;
````

&nbsp;

Both tickets have images of the polygons and hero's start/end coords.

### DIFF
--- a/engines/sci/engine/kpathing.cpp
+++ b/engines/sci/engine/kpathing.cpp
@@ -1407,15 +1407,22 @@ static void AStar(PathfindingState *s) {
 			// This difference might lead to problems, but none are
 			// known at the time of writing.
 
-			// WORKAROUND: This check fails in QFG1VGA, room 81 (bug report #3568452).
-			// However, it is needed in other SCI1.1 games, such as LB2. Therefore, we
-			// add this workaround for that scene in QFG1VGA, until our algorithm matches
-			// better what SSCI is doing. With this workaround, QFG1VGA no longer freezes
-			// in that scene.
-			bool qfg1VgaWorkaround = (g_sci->getGameId() == GID_QFG1VGA &&
-									  g_sci->getEngineState()->currentRoomNumber() == 81);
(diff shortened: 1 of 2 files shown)


bluegr:
Nice work, thanks!

Merging

PR #1475, opened 2019-01-09
SCI32: Fix QFG4 Trigger and Open spells

Vhati:
Fixes repeat casting and combination with Summon Staff, bug [#10860](https://bugs.scummvm.org/ticket/10860)

````
// The Trigger spell stalls and never reaches handsOn when preceded by a
// successful Summon Staff. An IF block calls hero::setCycle(Beg, self), which
// cues self on completion. Its condition tests hero's "view" property and
// executes if the staff is absent. There are no means to advance when the
// staff is present.
//
// Open (script 13) is largely identical w/o this bug. We match its behavior.
//
// Due to differences between editions, this is addressed with two patches. The
// first inserts a "seconds" property assignment before the IF, where it'll
// always cue. We make room by condensing the IF conditions. There are two
// "view" comparisons. Instead of sending for it twice, we recycle the view
// with pprev. The second patch removes setCycle's cue by nulling its last arg.
````

````
// The Open and Trigger spells init a green Prop for their effect. They don't
// dispose it the first time, and the effect is absent on further castings.
//
// The author specifically nerfed dispose() on its first call by testing if a
// variable has been set before allowing super::dispose(). We erase the
// branch to ensure that the effect is always disposed.
````
To test...
* Set flag 450 for Summon Staff to work.
  * vv g 528 8192
* Learn the "Summon Staff" spell.
  * send hero learn 32 200
* Teleport to the castle gate.
  * room 600
&nbsp;
* Cast Open on the gate more than once (1st row, 1st icon).
* Cast Trigger more than once (1st row, 3rd icon).
* Cast Summon Staff (3rd row, 3rd icon). Then cast Trigger.

The green effect will be visible with all castings, and the game won't get stuck.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -9502,6 +9502,108 @@ static const uint16 qfg4Forest557PathfindingPatch[] = {
 	PATCH_END
 };
 
+// The Trigger spell stalls and never reaches handsOn when preceded by a
+// successful Summon Staff. An IF block calls hero::setCycle(Beg, self), which
+// cues self on completion. Its condition tests hero's "view" property and
+// executes if the staff is absent. There are no means to advance when the
+// staff is present.
+//
+// Open (script 13) is largely identical w/o this bug. We match its behavior.
+//
+// Due to differences between editions, this is addressed with two patches. The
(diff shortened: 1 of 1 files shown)


bluegr:
Nice work, thanks! Merging

PR #1477, opened 2019-01-12
SCI32: Fix QFG4 guild secret passage

Vhati:
Fixes the walkable polygon not changing when it's revealed, bug [#9894](https://bugs.scummvm.org/ticket/9894)

````
// In the thieves' guild (room 430), the tunnel is not immediately walkable
// when it is revealed (by moving a barrel and solving a puzzle). Hero must
// re-enter the room to update the polygon.
//
// Curing Chief *will* immediately replace the polygon. However, most players
// will lack the item necessary on the first visit. Meeting Chief is how they
// learn about the item. If they go get it, they'll re-enter the room.
//
// The room's init has a cond block to check plot flags and declare one of 3
// polygons. The 3rd condition also inits secritExit, an invisible Feature
// that sends hero out of town when walked upon.
````

````
// Patch 1: Other patches ensure the passage will be walkable the moment it is
//  revealed. Chief is standing inside it. We skip the original code that would
//  set up the passage as he gets cured. It is redundant now. If hero can reach
//  him, the passage is already revealed. We won't let secritExit init twice.
````

````
// Patch 2: We free bytes in rm340::init() by condensing Feature inits with a
//  loop. Stack up their addresses. Pop & send repeatedly. Then we declare a
//  subroutine that disposes any existing obstacles, jumps into the cond block
//  to declare the 3rd poly, jumps back, passes it to addObstacles(), and inits
//  secritExit.
//
//  When the cond block's 3rd condition runs, we immediately call our
//  subroutine to do everything and end the cond, leaving the original polygon
//  declaration intact below the jump.
````

````
// Patch 3: The passage starts opening at sBarrelMove state 8. We need more
//  room than case 8 can offer, so we arrange for *multiple* cases to run
//  during state 8 - by omitting the final jump that would short-circuit.
//
//  Cases 1-5 have derelict code, once intended to move the barrel back and
//  forth, now only left. This is because barrel::doVerb(4) schedules
//  sBarrelMove in the absence of flag 254 and sets register=1 if the barrel is
//  in the left position already. Case 0 uses the same criteria in deciding to
//  skip to state 6. Thus cases 1-5 never see register==1. The barrel never
//  moves back, and bytes predicated on register==1 are available.
//
//  We reduce case 2 to only the necessary ops and splice in a new case that
//  runs during state 8 as a prologue to the original case 8. Our prologue
//  calls the subroutine to add the 3rd polygon. This patch has two variants,
//  toggled to match the detected edition with enablePatch() below. Aside from
//  the call offset, they are identical.
````
&nbsp;
To test...

First arrival
* Create a rogue character.
* room 340
* The room will init with the 1st polygon.
* Walk along the east wall to confirm it's not walkable yet.
* Click HAND on the poster, and "Take Guild Card".
* Use the card on the doors to open them.
* This immediately switches to the 2nd polygon.
* Walk through those doors to the safe.
* Walk along the east wall to confirm it's not walkable yet.
* Climb the ladder to leave.

Returning with access to the safe
* Set flag 243 (opened doors), if you hadn't just done that manually.
    * vv g 515 4096
* room 340
* The room will init with the 2nd polygon.
* Walk along the east wall to confirm it's not walkable yet.
* Walk to the safe.
* Climb the ladder to leave.

Returning to reveal the passage
* Set flags 243 and 241 (opened desk).
  * vv g 515 20480
* room 340
* Click HAND on the barrel below the poster.
  * Row 1: Blue Blue Yellow
  * Row 2: Green Green Green
  * Row 3: Red Green Red
* This immediately switches to the 3rd polygon.
* Walk into the passage heading east.
* Without the patches, this would not be possible.

Returning with the passage revealed
* Set flags 243, 241, and 254 (revealed passage), if you hadn't just done that manually.
    * vv g 515 20482
* room 340
* The room will init with the 3rd polygon.
* Walk into the passage heading east.

&nbsp;

An image is attached to the ticket.
* It shows the 3rd polygon, allowing access to the safe and passage.
* The pink rectangle is secritExit, the place hero needs to reach to go outside.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -159,9 +159,11 @@ static const char *const selectorNameTable[] = {
 	"plane",        // RAMA
 	"state",        // RAMA
 	"getSubscriberObj", // RAMA
+	"approachVerbs", // QFG4
 	"cue",          // QFG4
 	"heading",      // QFG4
 	"moveSpeed",    // QFG4
+	"obstacles",    // QFG4
 	"sayMessage",   // QFG4
 	"setLooper",    // QFG4
 	"setSpeed",     // QFG4
@@ -249,9 +251,11 @@ enum ScriptPatcherSelectors {
 	SELECTOR_plane,
 	SELECTOR_state,
 	SELECTOR_getSubscriberObj,
+	SELECTOR_approachVerbs,
 	SELECTOR_cue,
 	SELECTOR_heading,
 	SELECTOR_moveSpeed,
+	SELECTOR_obstacles,
 	SELECTOR_sayMessage,
 	SELECTOR_setLooper,
 	SELECTOR_setSpeed,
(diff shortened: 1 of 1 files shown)


Vhati:
I can streamline patch 1 a bit if there's an acceptable way to pop an object off the stack into the accumulator. IIRC the only opcodes that do that are arithmetic.

I think ScummVM technically permits adding 0 to an object to get the effect of a pop().

This is about doing repeated sends each time with a different acc.

A pop() would let me stack up object values, pop each to acc and send in a plain loop... without the overhead of a subroutine.
&nbsp;

So... would popping with "object + 0" be bad practice?

Vhati:
Hrm... I could move an object from the stack, to a temp var, to acc...

I'll experiment and see how that looks.

Vhati:
Much better. : )

Vhati:
Um, is it gonna be a problem that I'm not disposing this polygon when it's not used as an obstacle?

I can't explicitly fit that in rm340::dispose().
Will ScummVM take care of that with automatic garbage collection when the room is gone?

m-kiewitz:
> Um, is it gonna be a problem that I'm not disposing this polygon when it's not used as an obstacle?
> 
> I can't explicitly fit that in rm340::dispose().
> Will ScummVM take care of that with automatic garbage collection when the room is gone?

This is not good in any case. Script patches shouldn't basically add bugs (in this case memory leaks).
Some people extract ScummVM script patches and use them in the original interpreter, too.
Would it be possible to dispose + add the correct polygon on demand, when it's needed?

Vhati:
@m-kiewitz:
> Script patches shouldn't basically add bugs (in this case memory leaks).

Agreed. So close.

> Would it be possible to dispose + add the correct polygon on demand, when it's needed?

Rather than having rm340::init() store a live object in case it *might* be used... set up a subroutine to populate local2. One call nearby within the room's init. And a really loooong call from the reveal Script.

That'll require adding couple modest-size patches (CD and floppy byte counts differ). There's room for splicing in another prologue case to do that. If it's declared just before being added as an obstacle, the room will dispose it properly.

Thanks for the prodding!

Vhati:
Ready. I like how this turned out.

The subroutine does everything on demand, including secritExit::init(). The reveal script (sBarrelMove) is only patched in order to splice in single new case, just for the long distance call.

bluegr:
Overall, really nice work, and your script patch is basically a rewrite of the logic for this room action.

You have done a lot of testing on this, and at first glance, it all seems to be working correctly.

Merging

PR #1478, opened 2019-01-13
KYRA: (EOB) - on touchscreens, swap left and right licks in active item slots

singalen:
Using in-hand item, e.g. sword (now a single tap) is a much more frequent action than removing them (now a 2-finger right-click emulation).

Otherwise, whenever I fight, I end up taking swords from my characters’ hands instead of hitting enemies with them.

I find that I'm not the only one to have [problems playing EoB on touchscreens](https://forums.scummvm.org/viewtopic.php?f=17&t=14042&p=86826#p86826).

Probably this needs a documentation update too?
Criticism is welcome.

### DIFF
--- a/engines/kyra/gui_eob.cpp
+++ b/engines/kyra/gui_eob.cpp
@@ -32,6 +32,7 @@
 #include "backends/keymapper/keymapper.h"
 #include "common/system.h"
 #include "common/savefile.h"
+#include "common/config-manager.h"
 #include "graphics/scaler.h"
 
 namespace Kyra {
@@ -873,10 +874,21 @@ int EoBCoreEngine::clickedWeaponSlot(Button *button) {
 	static const uint8 sY[] = { 27, 27, 79, 79, 131, 131 };
 	int slot = sY[button->arg] > _mouseY ? 0 : 1;
 
-	if ((_gui->_flagsMouseLeft & 0x7F) == 1)
+	bool useItem = (_gui->_flagsMouseLeft & 0x7F) != 1;
+	// Do not swap buttons if there's an item held by mouse. Then we probably 
+	// want to put it in hand, not to attack with that hand.
+	bool itemInMouse = _itemInHand != 0;
+	if (??? && !itemInMouse) {
+		// Otherwise, whenever I fight, I end up taking swords 
+		// from my characters’ hands instead of hitting enemies with them.
+		useItem = !useItem;
+	}
+
+	if (useItem) {
 		gui_processWeaponSlotClickLeft(button->arg, slot);
-	else
+	} else {
 		gui_processWeaponSlotClickRight(button->arg, slot);
+	}
 
 	return 1;
 }


singalen:
Got it, thanks! Redoing.

singalen:
After looking for a way to figure out if a RMB click is a touchscreen event or a touchpad or a mouse event, I see no way to find this out in runtime.

In SDL, there's [SDL_MouseButtonEvent.which](https://wiki.libsdl.org/SDL_MouseButtonEvent) which lets one determine the source device. 
In Common::Event, I see no such field. Maybe it makes sense to add one? Another alternative would be to add something like `SDL_GetMouseState()`.

We have to determine the difference in runtime, because on certain computers there could be both: touchscreen on a PC, or a USB mouse on Android device. 

The best way to keep an actual mouse behavior I see is to actually #ifdef it out for Android/iOS/whatever.

tsoliman:
> Point is, this is NOT a backend issue. It's a specific game whose use of RMB in particular GUI control damages its gameplay with ScummVM's emulation of RMB. No other game affected.

I am not sure I understand how it isn't a backend issue when we're talking about backend-specific ifdefs inside engine code. (Specifically: whether the current backend has a touchscreen)

I am not saying it isn't game-specific. It is a game-specific backend issue. To me it is the same as adding WSAD support to EOB/LOL/etc. My understanding is that it is supposed to be handled in the EventMapper layer, which is game-aware to some extent.

Putting backend-specific `#ifdef`s in an engine's code smells to me of concern-leakage.

singalen:
@tsoliman so what is your suggestion?

athrxx:
I understand that using the key mapper might be difficult, since this feature request concerns only one particular GUI action (clicking the weapon slots). 

However, changing the default engine behaviour for all tablet users is not something I wish to do.
 
How would something like that work for you?
https://github.com/athrxx/scummvm/tree/eob-buttonswap

You can turn on this option in the launcher if you want, but it will not be the default.
Note that you have to run the engine once for the new game option to "register", before you'll see it in the launcher.

singalen:
That's fantastic, thanks! Somewhy I thought that it's difficult to add arbitrary options to ScummVM.

I was considering making a PR where `Common::Event` has a `source` field, describing if it's a mouse, touchpad or a touchscreen, like in SDL. This would make such a touchscreen enhancements much easier in the future. But this option is a good compromise. I'll clone the branch and rebase my change on top of it.

On the all platforms point: as a tablet player, I see the change as positive for all touchscreen platforms. It's quite a painful and error-prone process - trying to do a right-click gesture each time you need to swing a sword. RMB emulation is a cross-platform function in ScummVM (I think), so the engine improvement should also be cross-platform.

athrxx:
Since there hasn't been any more activity here (and the actual PR is broken)  I have now made a commit which adds the button swap option to the engine.

Can this PR be closed?

tsoliman:
Fixed by e8ee079b86f683f5faa4cb54f5f358ff7a684081

Thanks everyone

PR #1479, opened 2019-01-15
SCI32: Fix QFG4 castle gate after geas

Vhati:
Fixes hero being able to walk through the closed gate, bug [#10871](https://bugs.scummvm.org/ticket/10871)

````
// After hero is geas'd in the dungeon (room 670) and teleported to the gate
// (600), hero can walk through the closed gate and exit north to the castle
// entrance. Two IF blocks with inconsistent conditions decide whether the
// gate is open and whether to use a polygon that extends beyond the gate.
// When re-entering from the forest (552), the gate is only open and passable
// if hero is qualified.
//
// The room has distinct situations for merely teleporting from the dungeon
// (local[0] = 11) and for entering from the forest while geas'd and carrying
// all the ritual scrolls (local[0] = 10). The latter sets a vital plot flag.
// Adding those checks and the flag to the former, plus opening the gate,
// would be non-trivial.
//
// We edit the polygon's IF condition to remove the dungeon check, making the
// closed gate impassable so hero will have to return from the forest.
````
&nbsp;
To test...

* Create a new character.
* Spoof the current room number as if you were in the dungeon.
  * vv g 11 670
* Set flag 110 (geas)
  * vv g 506 2
* Acquire the ritual scrolls.
  * send hero get 52
  * send hero get 53
  * send hero get 54
  * send hero get 55
  * send hero get 56
* Teleport to the gate.
  * room 600
* The gate will be solid. Without the patch, the gate would be closed yet passable (regardless of whether hero had the scrolls already). Because of the missing flag, Ad and Kat would not be waiting at the castle entrance.

* Go south and return.
* The gate will be open and passable. Ad and Kat will be waiting.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -9604,6 +9604,46 @@ static const uint16 qfg4EffectDisposalPatch[] = {
 	PATCH_END
 };
 
+// After hero is geas'd in the dungeon (room 670) and teleported to the gate
+// (600), hero can walk through the closed gate and exit north to the castle
+// entrance. Two IF blocks with inconsistent conditions decide whether the
+// gate is open and whether to use a polygon that extends beyond the gate.
+// When re-entering from the forest (552), the gate is only open and passable
+// if hero is qualified.
+//
+// The room has distinct situations for merely teleporting from the dungeon
+// (local[0] = 11) and for entering from the forest while geas'd and carrying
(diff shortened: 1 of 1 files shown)


bluegr:
Nice work! Merging

bluegr:
Could you please fix the merge conflicts?

bluegr:
Nice work, thanks! Merging

PR #1482, opened 2019-01-18
ZVISION: Boost volume for MPEG cutscenes

eriktorbjorn:
The high-resolution videos play back at much lower volume than the original ones. This adds hard-coded values for how much to amplify each cutscene. It's all done by ear, and it does introduce some
clipping, but I think it should be acceptable. 

Of course, it could also be a problem with the audio decoder, so this may be the wrong approach entirely.

### DIFF
--- a/engines/zvision/video/video.cpp
+++ b/engines/zvision/video/video.cpp
@@ -50,96 +50,7 @@ Video::VideoDecoder *ZVision::loadAnimation(const Common::String &fileName) {
 		animation = new ZorkAVIDecoder();
 #if defined(USE_MPEG2) && defined(USE_A52)
 	else if (tmpFileName.hasSuffix(".vob")) {
-		// For some reason, we get much lower volume in the hi-res
-		// videos than in the low-res ones. So we artificially boost
-		// the volume here. This is an approximation, but I've tried
-		// to match the old volumes reasonably well.
-		//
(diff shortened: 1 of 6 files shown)


eriktorbjorn:
I forgot to mention that this is in response to my own bug report #10862, "AUDIO: AC3 volume seems too low".

eriktorbjorn:
> What are your thoughts on variable change from decibel to gain_dB?

I don't have any strong opinions for or against. At least not at the moment.

engines/zvision/video/video.cpp
```diff
@@ -49,8 +49,99 @@ Video::VideoDecoder *ZVision::loadAnimation(const Common::String &fileName) {
...
+		} else if (tmpFileName == "g000d981.vob") {
+			// The Enterprise logo. Has no low-res version. Its
+			// volume is louder than the other logo animations.
+			amplification = 6.2;
+		} else if (tmpFileName.hasPrefix("g000d")) {
+			// The Dolby Digital and Activision logos. They have no
+			// low-res versions, but I've used the low-resolution
+			// Activision logo (slightly different) as reference.
+			amplification = 8.5;
+		}
```
bluegr: IMHO, this code block that decides the amplification for each video should be moved into a separate function

bluegr:
Well done! I believe this is good for now, since no other engine is using the AC3 decoder

PR #1483, opened 2019-01-19
SCI: Cleanup for all script patches

Vhati:
* Corrected opcode comments.
* Corrected SIG macros in patches and vice versa.
* Added some ticket numbers to descriptions.
* Named & bracketed indices for [ global | local | temp | param ] vars.
  * Indices themselves are left in each author's markup.
* Some description revision and word wrapping.
* Lowered uppercase hex bytes. Padded single-digit hex bytes.
  * Space-padded decimal bytes so their lack of 0x prefix stands out among the regular hex bytes.
* Removed stray whitespace characters.
* Parenthesizd any hex values where selector macros are present, since they've become reminders.
* Four digit uint16 macros. An afterthought. Not exhaustively enforced.
* Removed a workaround I'd commented away and forgotten about.

These should be innocuous changes.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -3443,42 +3443,41 @@ static const SciScriptPatcherEntry lighthouseSignatures[] = {
 #endif
 
 // ===========================================================================
-// Script 210 in the German version of Longbow handles the case where Robin
-// hands out the scroll to Marion and then types his name using the hand code.
-// The German version script contains a typo (probably a copy/paste error),
-// and the function that is used to show each letter is called twice. The
-// second time that the function is called, the second parameter passed to
-// the function is undefined, thus kStrCat() that is called inside the function
-// reads a random pointer and crashes. We patch all of the 5 function calls
-// (one for each letter typed from "R", "O", "B", "I", "N") so that they are
-// the same as the English version.
+// When Robin hands out the scroll to Marion and then types his name using the
+// hand code, the German version's script contains a typo (likely a copy/paste
+// error), and the local procedure that shows each letter is called twice. The
+// The procedure expects a letter arg and returns no value, so the first call
+// takes its letter and feeds an undefined value to the second call. Thus the
+// kStrCat() within the procedure reads a random pointer and crashes.
+// 
+// We patch all of the 5 doubled local calls (one for each letter typed from
+// "R", "O", "B", "I", "N") to be the same as the English version.
 // Applies to at least: German floppy
-// Responsible method: unknown
+// Responsible method: giveScroll::changeState(19,21,23,25,27) in script 210
 // Fixes bug: #5264
 static const uint16 longbowSignatureShowHandCode[] = {
-	0x78,                            // push1
-	0x78,                            // push1
-	0x72, SIG_ADDTOOFFSET(+2),       // lofsa (letter, that was typed)
+	0x78,                            // push1 (1 call arg)
+                                     //
+	0x78,                            // push1 (1 call arg)
+	0x72, SIG_ADDTOOFFSET(+2),       // lofsa (letter that was typed)
 	0x36,                            // push
-	0x40, SIG_ADDTOOFFSET(+2),       // call
-	0x02,                            // perform the call above with 2 parameters
-	0x36,                            // push
-	0x40, SIG_ADDTOOFFSET(+2),       // call
-	SIG_MAGICDWORD,
-	0x02,                            // perform the call above with 2 parameters
-	0x38, SIG_SELECTOR16(setMotion), // pushi "setMotion" (0x11c in Longbow German)
-	0x39, SIG_SELECTOR8(x),          // pushi "x" (0x04 in Longbow German)
+	0x40, SIG_ADDTOOFFSET(+2), 0x02, // call [localproc], 02
+                                     //
+	0x36,                            // push (the result is an arg for the next call)
+	0x40, SIG_ADDTOOFFSET(+2), SIG_MAGICDWORD, 0x02, // call [localproc], 02
+                                     //
+	0x38, SIG_SELECTOR16(setMotion), // pushi setMotion (0x11c in Longbow German)
+	0x39, SIG_SELECTOR8(x),          // pushi x (0x04 in Longbow German)
 	0x51, 0x1e,                      // class MoveTo
 	SIG_END
 };
 
 static const uint16 longbowPatchShowHandCode[] = {
 	0x39, 0x01,                      // pushi 1 (combine the two push1's in one, like in the English version)
-	PATCH_ADDTOOFFSET(+3),           // leave the lofsa call untouched
-	// The following will remove the duplicate call
-	0x32, PATCH_UINT16(0x0002),      // jmp 02 - skip 2 bytes (the remainder of the first call)
-	0x48,                            // ret (dummy, should never be reached)
-	0x48,                            // ret (dummy, should never be reached)
+	PATCH_ADDTOOFFSET(+3),           // leave the lofsa untouched
+	// The following will remove the first push & call
+	0x32, PATCH_UINT16(0x0002),      // jmp 02 [to the second push & call]
+	0x35, 0x00,                      // ldi 0 (waste 2 bytes)
 	PATCH_END
 };
 
@@ -3530,7 +3529,7 @@ static const uint16 longbowSignatureBerryBushFix[] = {
 	// jump location for the "bt" instructions
 	0x30, SIG_UINT16(0x0011),        // bnt [skip over follow up code, to offset 0c35]
 	// 55 bytes until here
-	0x85, 00,                        // lat temp[0]
+	0x85, 0x00,                      // lat temp[0]
 	SIG_MAGICDWORD,
 	0x9a, SIG_UINT16(0x0110),        // lsli local[110h] -> 110h points normally to 110h / 2Bh
 	// 5 bytes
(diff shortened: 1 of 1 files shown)


Vhati:
This PR was a fun exercise in [editing past commits](https://stackoverflow.com/a/1186549).

* Find the hash for a commit to modify.
  * git log
* git rebase -i "1234abc^"
  * Choose 'edit' for that commit.
  * Save and quit from the editor.
* Modify the file.
* git add "..."
* git commit --amend
  * Save and quit from the editor.
* Return to the present.
  * git rebase --continue

Vhati:
@m-kiewitz:
> Please move the QfG3 fix into a separate pull request, because that one is supposed to actually change something. The others are not.

Will do.

Vhati:
The QfG3 fix has been removed and given its own PR.

Vhati:
That's a bit irksome, though understandable.
I went back to remove some commas from QFG1.
Downstream commits got new hashes as a side effect of the rebase.
The PR thread split my pretty list.

Oh well. They're still together in GitHub's "Files Changed" tab.
"Changes from all commits" drop down menu.

Vhati:
Since GitHub's been quiet, I added a little more.

Fixed ldi comments in GK1, PoliceQuest, QFG2, QFG4
Added ldi comments in KQ6
Padded single-digit hex bytes in SCI32, LB1

Revised pushi comments (mostly unquoting/unparenthesizing selectors) in EcoQuest, GK1, KQ6, LB1, Longbow, PoliceQuest, SpaceQuest, Hoyle

engines/sci/engine/script_patches.cpp
```diff
@@ -3279,9 +3279,9 @@ static const uint16 kq7SubtitleFixSignature3[] = {
 	0x63, 0x28,                 // pToa initialized
 	0x18,                       // not
 	0x31, 0x07,                 // bnt [skip init code]
-	0x38, SIG_SELECTOR16(init), // pushi init ($8e for 2.00, $93 for 1.51)
```
m-kiewitz: I don't really like removing comments like these, that show which values were used in which version of the game.

engines/sci/engine/script_patches.cpp
```diff
@@ -3322,32 +3321,36 @@ static const uint16 kq7BenchmarkSignature[] = {
...
 	0x78,                      // push1
 	0x38, SIG_UINT16(0xfdd4),  // pushi 64980
 	SIG_END
 };
 
 static const uint16 kq7BenchmarkPatch[] = {
-	0x34, PATCH_UINT16(10000), // ldi 10000
-	0x48,                      // ret
+	0x34, PATCH_UINT16(0x2710), // ldi 10000
+	0x48,                       // ret
```
m-kiewitz: And I don't really see a point in change this either. In this case it's meant to be a specific decimal value.
Changing it to hex in the patch opcode doesn't make it easier readable, quite the opposite actually.

engines/sci/engine/script_patches.cpp
```diff
@@ -2367,17 +2369,17 @@ static const uint16 kq5PatchWitchCageInit[] = {
...
+// but the signals are never set and thus the game hangs. We disable that code,
+// so that the battle works again. This also happened in the original
+// interpreter. We must not change similar code, that happens before.
+//
 // Applies to at least: French PC floppy, German PC floppy, Spanish PC floppy
 // Responsible method: stingScript::changeState, dragonScript::changeState, snakeScript::changeState
 static const uint16 kq5SignatureMultilingualEndingGlitch[] = {
 	SIG_MAGICDWORD,
-	0x89, 0x57,                      // lsg global[57h]
+	0x89, 0x57,                      // lsg global[87]
```
m-kiewitz: I also don't really see a reason to change lines like these.
The original was consistent. Hexadecimal for both. And the ScummVM SCI-debugger accepts hex as well as decimal values for viewing globals.
Now it's mixed up, which makes it easier that errors (actual value mismatches) go unnoticed.

engines/sci/engine/script_patches.cpp
```diff
@@ -5596,30 +5596,30 @@ static const SciScriptPatcherEntry laurabow2Signatures[] = {
...
 	0x43, 0x62, 0x04,                // callk StrAt
 	0x36,                            // push
 	0x35, 0x20,                      // ldi 20
 	0x04,                            // sub
-	0xa1, SIG_ADDTOOFFSET(+1),       // sag global[57h] -> FM-Towns [9Dh]
+	0xa1, SIG_ADDTOOFFSET(+1),       // sag global[87] -> 57h, FM-Towns = 9Dh
 	// 35 bytes
 	0x39, 0x03,                      // pushi 03
-	0x89, SIG_ADDTOOFFSET(+1),       // lsg global[1Dh] -> FM-Towns [1Eh]
+	0x89, SIG_ADDTOOFFSET(+1),       // lsg global[29] -> 1Dh, FM-Towns = 1Eh
```
m-kiewitz: This right here gets now really confusing, while in retrospect it was already quite confusing already and should have been worded differently.

like for example:
// sag global[57h], for FM-Towns: sag global[90h]

engines/sci/engine/script_patches.cpp
```diff
@@ -5647,38 +5647,38 @@ static const uint16 mothergoose256PatchReplay[] = {
...
 	// 8 bytes
 	0x39, 0x06,                      // pushi 06
 	0x39, 0x0c,                      // pushi 0Ch
 	0x76,                            // push0
 	0x76,                            // push0
-	0x38, PATCH_UINT16(200),         // push 200
-	0x38, PATCH_UINT16(320),         // push 320
+	0x38, PATCH_UINT16(0x00c8),      // pushi 200
+	0x38, PATCH_UINT16(0x0140),      // pushi 320
 	0x78,                            // push1
```
m-kiewitz: Same as before, I personally do not see a reason to change the decimal values here.
Those are meant to be 320x200 (screen size), that's why I used decimal notation instead of hexadecimal.

engines/sci/engine/script_patches.cpp
```diff
@@ -6502,7 +6501,7 @@ static const uint16 pq3PatchGiveLocketPoints[] = {
 	0x7a,                                // push2
 	0x38, PATCH_UINT16(0x00ff),          // pushi 0x00ff - using last flag slot, seems to be unused
 	0x39, 0x0a,                          // pushi 10d - 10 points
-	0x45, 0x06, 0x04,                    // callb export000_6, 4
+	0x45, 0x06, 0x04,                    // callb 4 proc0_6
```
m-kiewitz: I used this opcode as it appears in SCI Viewer.
I don't see a reason to change those at all.
In fact it gets more confusing for me, because I'm still using SCI Viewer almost all the time when looking at script code.

engines/sci/engine/script_patches.cpp
```diff
@@ -7146,22 +7148,22 @@ static const uint16 qfg1vgaSignatureWhiteStagDagger[] = {
 	0x1a,                               // eq?
 	0x31, 0x16,                         // bnt [next parameter check]
 	0x76,                               // push0
-	0x45, 0x02, 0x00,                   // callb export 2 from script 0, 0
```
m-kiewitz: Maybe this notation for CALLBs may even be better, because it makes it obvious what the call actually does.
"call export 2 from script 0". Anyone should understand that.

engines/sci/engine/script_patches.cpp
```diff
@@ -7146,22 +7148,22 @@ static const uint16 qfg1vgaSignatureWhiteStagDagger[] = {
...
 	0x39, 0x03,                         // pushi 03
 	0x39, 0x51,                         // pushi 51h
 	0x76,                               // push0
 	0x76,                               // push0
 	0x7c,                               // pushSelf
-	0x81, 0x5b,                         // lag global[5Bh] -> qg1Messager
+	0x81, 0x5b,                         // lag global[91] -> qg1Messager
 	0x4a, 0x0e,                         // send 0Eh -> qg1Messager::say(3, 51h, 0, 0, stagHurt)
-	0x33, 0x12,                         // jmp -> [ret]
+	0x33, 0x12,                         // jmp [end the switch]
```
m-kiewitz: should be "end of routine" / "end of method". The original comment was more obvious where the jump is supposed to lead to.

Vhati:
@m-kiewitz:
Thank you for the feedback. I'll diff and remove those types of counterproductive edits.

Vhati:
@m-kiewitz, re jmp to "ret" / "end the switch" / "end of method":
I was thinking of the toss that switch cases all short-circuit to, which is often followed by ret in changeState() methods.

Speaking of... I've seen a number of doVerb() methods in QFG4, with explicit early "return 1" lines within switches (ldi 1 & ret) that do NOT toss.

Does ret reset the stack on its own?
Is it a script bug to ret without tossing?
Would that be a bug that needs correcting?

Vhati:
* For bracketed vars, indices themselves are left in each author's markup.
* Reverted decimal uint16 macros.
* Space-padded the rare decimal bytes so their lack of 0x prefix stands out among the regular hex bytes.
* More ticket numbers.
* Standardized callb/calle comments to "//opcode [export X of script Y], framesize"
* Added commas to callk and super comments where framesize is mentioned.

m-kiewitz:
Thank you very much.
Had almost no time today b/c work, will look into it as soon as time allows.

Vhati:
longbowSignatureShowHandCode looks wrong.
````
0x78,                            // push1
0x78,                            // push1
0x72, SIG_ADDTOOFFSET(+2),       // lofsa (letter, that was typed)
0x36,                            // push
0x40, SIG_ADDTOOFFSET(+2),       // call
0x02,                            // perform the call above with 2 parameters
0x36,                            // push
0x40, SIG_ADDTOOFFSET(+2),       // call
SIG_MAGICDWORD,
0x02,                            // perform the call above with 2 parameters
````
I'm reading that as...
````
0x78,                            // push1 (1 call arg)
                                 //
0x78,                            // push1 (1 call arg)
0x72, SIG_ADDTOOFFSET(+2),       // lofsa (letter, that was typed)
0x36,                            // push
0x40, SIG_ADDTOOFFSET(+2), 0x02, // call with *framesize* 2
0x36,                            // push (stack up the result as an arg)
                                 //
0x40, SIG_ADDTOOFFSET(+2), 0x02, // call with *framesize* 2
````
That fits with the [specs](https://wiki.scummvm.org/index.php/SCI/Specifications/SCI_virtual_machine/The_Sierra_PMachine).
"op 0x40: call W relpos, B framesize (4 bytes)"

Sidenote: The wiki doesn't mention that SCI32 uses uint16 for framesize on call / callb / calle / callk.
For example, in QFG4, 0x40 takes two uint16 args (5 bytes).

m-kiewitz:
> longbowSignatureShowHandCode looks wrong.

I have not added that script patch, and I also don't own that specific version, so I can't check.
I'm not sure, maybe @bluegr added it.
I would be very surprised if it doesn't work at all.

Vhati:
@m-kiewitz:
> I would be very surprised if it doesn't work at all.

I'm referring to the comments, not the bytes. It probably works.
* The comments say "perform the call above with 2 parameters" apparently about the last byte of a call which would be framesize=2... which is double the arg count, not the arg count itself.
* There's only one actual arg present in the sig.
* Calls have arg counts pushed in advance, and there are push1's, not push2's.

Vhati:
Longbow patch amended.

Vhati:
@sluicebox:
> I meant "icons". Both. I am deeply uncomfortable with the scope of these mass comment edits.

Okay, the 's' is restored, wrapping the word to the next line.

Reviewing the entire file at once can be intimidating. In GitHub's diff, you can navigate among commits with [n]ext and [p]rev keyboard shortcuts... or the "changes from all commits" dropdown menu in the upper left. Each commit lets you review sections of the file at a time.

Regex searches were used only to find inconsistencies. No risky automated replacement was performed.

Is there anything else you're specifically uncomfortable with?

bluegr:
First of all, nice work, once again! :)

I’ve reviewed this multiple times, as the changes are quite a lot, and the commits kept changing after the reviews. In its current state, I don’t see anything major that needs to be changed.

This has been reviewed by three people already, and I don’t believe that any further major review findings are to be found.

Since the changes here will be needed for any future script patch pull requests, I’m merging this. If there are any minor leftover issues, these can be fixed in-tree.

PR #1484, opened 2019-01-19
SCI: Fix QFG3 combat speed throttling patch

Vhati:
Fixes a bad byte.

Original qfg3PatchCombatSpeedThrottling1
```
0x80, 0xd2,                         // lsg global[D2h]
^^^^
0x14,                               // or
0x31, 0x06,                         // bnt [skip code] - saves 4 bytes
```
The op byte does not match the instruction's expected length or opcode comment.
* 0x80 would be: lag uint16.
* 0x89 is: lsg byte.

&nbsp;

Confirmed with the debugger's disasm...

Original (0x80)
````
0027:102f: 83 01          lal   01
0027:1031: 1c             ne?
0027:1032: 80 d2 14       lag   14d2            ; 5330
0027:1035: 31 06          bnt   06  [103d]
````
Corrected (0x89)
````
0027:102f: 83 01          lal   01
0027:1031: 1c             ne?
0027:1032: 89 d2          lsg   d2              ; 210
0027:1034: 14             or
0027:1035: 31 06          bnt   06  [103d]
````

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -7980,28 +7980,41 @@ static const uint16 qfg3PatchMissingPoints2[] = {
 // Responsible method: combatControls::dispatchEvent (script 550) + WarriorObj in heap
 // Fixes bug: #6247
 static const uint16 qfg3SignatureCombatSpeedThrottling1[] = {
-	0x31, 0x0d,                         // bnt [skip code]
+	0x3f, 0x03,                         // link 3d (these temp vars are never used)
+	0x76,                               // push0
+	0x43, 0x42, 0x00,                   // callk GetTime, 0d
+	0xa1, 0x58,                         // sag global[88]
+	0x36,                               // push
+	0x83, 0x01,                         // lal local[1]
+	SIG_ADDTOOFFSET(+3),                // ...
 	SIG_MAGICDWORD,
-	0x89, 0xd2,                         // lsg global[D2h]
+	0x89, 0xd2,                         // lsg global[210]
 	0x35, 0x00,                         // ldi 0
 	0x1e,                               // gt?
-	0x31, 0x06,                         // bnt [skip code]
-	0xe1, 0xd2,                         // -ag global[D2h] (jump skips over this)
-	0x81, 0x58,                         // lag global[58h]
-	0xa3, 0x01,                         // sal local[01]
+	SIG_ADDTOOFFSET(+6),                // ...
+	0xa3, 0x01,                         // sal local[1]
 	SIG_END
 };
 
 static const uint16 qfg3PatchCombatSpeedThrottling1[] = {
-	0x80, 0xd2,                         // lsg global[D2h]
-	0x14,                               // or
-	0x31, 0x06,                         // bnt [skip code] (saves 4 bytes)
-	0xe1, 0xd2,                         // -ag global[D2h]
-	0x81, 0x58,                         // lag global[58h]
-	0xa3, 0x01,                         // sal local[01] (jump skips over this)
-	// our code
-	0x76,                               // push0
-	0x43, 0x2c, 0x00,                   // callk GameIsRestarting (add this to trigger our speed throttler)
+	0x76,                               // push0  (no link, freed +2 bytes)
+	0x43, 0x42, 0x00,                   // callk GetTime, 0d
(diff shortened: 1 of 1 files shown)


bluegr:
Nice catch! Odd that this went unnoticed. Have you tried playing the game with this change? Is the patch applied properly? Are there any regressions?

Vhati:
@bluegr:
> Is the patch applied properly?

It does get applied, which is to say the signature matches, causing the patch bytes to be written.
The OP quotes were copied from the debugger.

> Nice catch! Odd that this went unnoticed.
> [...]
> Have you tried playing the game with this change?

I only got into combat to check the disassembly. Didn't actually play.

I was operating on the assumption that if it accidentally 'worked' when it ignored local[1] and checked global[5330], it ought to work better with the intended global[210].

I'll do some playtesting with and without the original patch for a baseline, then see if the corrected patch has the same effect...

Vhati:
I can't verify that it's actually working. Didn't notice an in-game effect for comparison.

Not sure what this patch is doing to the surrounding code as it squeezes in a kernel call.

&nbsp;

Here's the unpatched code from SCI Companion.

script 550 - combatControls::dispatchEvent()
````
(method (dispatchEvent pEvent &tmp [temp0 3])
	(if
		(and
			(!= (= global88 (GetTime)) local1)
			(> global210 0)
		)
		(-- global210)
		(= local1 global88)
	)
	(super dispatchEvent: pEvent &rest)
	(g8_Sounds eachElementDo: #check)
	(g7_timers eachElementDo: #doit)
	(Animate (g5_cast elements?) 1)
	(if global37
		(= global37 0)
		(g5_cast eachElementDo: #motionCue)
	)
)
````
Whenever time passes since the last check, and global210 is > 0, decrement and remember the time with local1.

&nbsp;

Here's what the corrected patch does...
````
(method (dispatchEvent pEvent &tmp [temp0 3])
	(if
		# The logical 'and' becomes a bitwise OR.
		#  Previously requiring both conditions to be non-zero, this now runs if
		#  either one of them is, which could make global210 decrement faster.
		# The gt? comparison is gone, so if global210 goes negative, the
		#  condition will always be true.
		(|
			(!= (= global88 (GetTime)) local1)
			global210
		)
		(-- global210)
		(= local1 global88)
	)
	(GameIsRestarting)

	(super dispatchEvent: pEvent &rest)
	(g8_Sounds eachElementDo: #check)
	(g7_timers eachElementDo: #doit)
	(Animate (g5_cast elements?) 1)
	(if global37
		(= global37 0)
		(g5_cast eachElementDo: #motionCue)
	)
)
````
local1 used to make sure time had passed since the last check.
The bitwise OR undermines that. Dunno if that's desired.

I couldn't find where global210 was initialized or what it's used for. Maybe the whole IF is derelict?
Indeed, the debugger reported global210's value was 0 before my new character entered combat. It was negative (for a byte) during combat (0000:ffdc, 0000:ffc0, 0000:ff73). If I set global210 to 1, it becomes 0000:ff81.

global210 decrements endlessly (whenever time passes or global210 is non-zero... effectively always).

&nbsp;

The original patch did this.
````
(method (dispatchEvent pEvent &tmp [temp0 3])
	(!= (= global88 (GetTime)) local1)
	(if global5330
		(-- global210)
		(= local1 global88)
	)
	(GameIsRestarting)
// ...
````

global5330 is not a valid index.
Stepping through in the debugger reports the original 0x80 lag op does not set the acc register at all.

That means the IF is doing a non-zero test on whatever was previously in acc, in this case the first condition with local1.

Again, global210 decrements endlessly (whenever time passes). It's not even trying to stop at 0.

Vhati:
> I couldn't find where global210 was initialized or what it's used for.

Apparently it's a countdown to expire the Reversal spell.

script 0 - Glory::doit()
````
(if (and (not (proc0_6 7)) (proc0_6 8))
	(switch (-- global210)
		(0
			(proc0_5 8)
			# "Your Reversal spell has worn off."
			(g91_qg3Messager say: 0 0 10 0 0 0)
		)
		(10
			# "Your Reversal spell is wearing thin."
			(g91_qg3Messager say: 0 0 11 0 0 0)
		)
	)
)
````

bluegr:
Yes, the kernel call matters here for our speed throttler. The changes look good, nice work!

Merging

PR #1487, opened 2019-01-30
IOS: Add support for three finger swipes up/down for showing/hiding so…

jonnyijapan:
Also if the user for whatever reason wants to show/hide the keyboard at will, he/she may do so using <s>four</s>three-finger swipes up/down. Hopefully this doesn't clash with anything. I don't think there were any "<s>four</s>three finger" gestures in the old scumm games... I've been playing through a couple of old scumm games this way and I think it works well. This means you can use keyboard whenever you want, regardless if the device is in landscape or portrait.

Original PR #1358

### DIFF
--- a/backends/platform/ios7/ios7_video.mm
+++ b/backends/platform/ios7/ios7_video.mm
@@ -349,17 +349,18 @@ uint getSizeNextPOT(uint size) {
 }
 
 - (void)setupGestureRecognizers {
-	UISwipeGestureRecognizer *swipeUpFourFingers = [[UISwipeGestureRecognizer alloc] initWithTarget:self action:@selector(fourFingersSwipeUp:)];
-	swipeUpFourFingers.direction = UISwipeGestureRecognizerDirectionUp;
-	swipeUpFourFingers.numberOfTouchesRequired = 4;
-	swipeUpFourFingers.delaysTouchesBegan = NO;
-	swipeUpFourFingers.delaysTouchesEnded = NO;
+	const NSUInteger KEYBOARDSWIPETOUCHCOUNT = 3;
+	UISwipeGestureRecognizer *swipeUpKeyboard = [[UISwipeGestureRecognizer alloc] initWithTarget:self action:@selector(keyboardSwipeUp:)];
+	swipeUpKeyboard.direction = UISwipeGestureRecognizerDirectionUp;
+	swipeUpKeyboard.numberOfTouchesRequired = KEYBOARDSWIPETOUCHCOUNT;
+	swipeUpKeyboard.delaysTouchesBegan = NO;
+	swipeUpKeyboard.delaysTouchesEnded = NO;
 	
-	UISwipeGestureRecognizer *swipeDownFourFingers = [[UISwipeGestureRecognizer alloc] initWithTarget:self action:@selector(fourFingersSwipeDown:)];
-	swipeDownFourFingers.direction = UISwipeGestureRecognizerDirectionDown;
-	swipeDownFourFingers.numberOfTouchesRequired = 4;
-	swipeDownFourFingers.delaysTouchesBegan = NO;
-	swipeDownFourFingers.delaysTouchesEnded = NO;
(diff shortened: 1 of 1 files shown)


criezy:
Thank you. At a first glance this seems fine to me, but I would like to wait a few days until I have some time to test this before merging.

criezy:
I have now had the opportunity to test this, and by default for me it does nothing. This is because 4 and 5 fingers swipes are used by iOS for multitasking (switch between running applications). However if I turn off multitasking Gesture (n Settings > General > Multitasking) then the 4 fingers swipe up and down work as expected to show/hide the virtual keyboard.

Unless we can find a better gesture, I am tempted to merge this as is anyway since not all iOS devices support multitasking, since we can turn off multitasking gesture easily on those that do, and since this commit doesn't change anything when multitasking gestures are enabled, and works as expected when they are not.

I will wait one more week before merging to see if anybody has comments about this.

jonnyijapan:
What device are you using? I'm using iPhone 6s plus and four fingers doesn't really do anything in the default case. If it does on newer devices there might be a way to disable that by application but I need to confirm that.

criezy:
That's an iPad Air (from 2013) with iOS 12.1.
I think those 4 or 5 fingers gestures to switch between apps were introduced in iOS 12, and they may be specific to the iPads.

Note: Only the swipe up, left, and right actually do something when multitasking gestures are enabled. And despite not doing anything, the 4 fingers swipe down still do not work to hide the keyboard. I need to disable multitasking gestures for both the up and down swipe to work to show/hide the keyboard.

I don't know if it is possible to disable the multitasking gesture for a given app only, but even if it were, I am not sure it would be a good idea. The user can disables those globally, and if he doesn't he would probably expect those to work in all the apps.

And since rotating the iPad still works to show/hide the keyboard, for me it is not a big issue if the gesture doesn't work.

jonnyijapan:
Tried on my iPad mini 2 (which is getting pretty old, likely 2012-13 or so) and the four finger swipe for keyboard show/hide work, I don't think it supports the four/five finger swipes to switch apps. I can't find that setting either.

I suggest you try on an iPhone, which this PR is meant for really. The keyboard really gets in the way of most UI in portrait mode and it's impossible to hide the keyboard on an iPhone in another manner. And the screens get too small for anyone to see in portrait mode.

But having said that to get around the problem we could just change this to three fingers instead for the show/hide keyboard in ScummVM. The reason I chose four was pretty random, it didn't collide with anything at the time, and then iOS 12 came  😅 
I'm gonna try changing it to three and see if collides with anything with some testing.

jonnyijapan:
Three fingers seems ok so far, tried on my iPad mini 2 and iPhone 6s Plus.

criezy:
Unfortunately I don't have an iPhone to try the changes on. But it works well on iPad now. As far as I can see we were indeed not using the 3 fingers swipe for anything yet, and it doesn't conflict with iOS gestures either.

This is good to merge now in my opinion, although we will probably want to squash the two commits when we do it. I will wait until the weekend as I indicated previously (and that way I will also have time to fix the build in the unlikely event it breaks on buildbot).

Thank you for your work on this!

criezy:
@jonnyijapan If you have the opportunity to try my changes from PR #1502, let me know what you think.

And as a note so that I hopefully don't forget about it, to combine the two PRs we will need to change in this one
```
- (void)keyboardSwipeUp:(UISwipeGestureRecognizer *)recognizer {
	[_keyboardView showKeyboard];
}
	
- (void)keyboardSwipeDown:(UISwipeGestureRecognizer *)recognizer {
	[_keyboardView hideKeyboard];
}
```
to
```
- (void)keyboardSwipeUp:(UISwipeGestureRecognizer *)recognizer {
	[self showKeyboard];
}
	
- (void)keyboardSwipeDown:(UISwipeGestureRecognizer *)recognizer {
	[self hideKeyboard];
}
```
so that it tracks properly if the keyboard is shown or not.

criezy:
This has now been merged manually with a small change as indicated in my previous comment.

PR #1488, opened 2019-01-31
SCI: Fix ECO1CD giving Superfluous the oily shell

sluicebox:
Fixes a script bug in the CD version, bug #10881


bluegr:
Nice work! Merging

PR #1493, opened 2019-02-03
ENGINE: Show unknown games if there is no recognized one

jcrona:
Prevents some recognized games to trigger the unknown game dialog as well.

For instance, Myst Masterpiece Edition is detected as an unknown variant of AGOS or TONY because of its data1.cab file, even if it is also properly detected as Myst:
```
The game in '/home/jc/ScummVM/Games/Myst_ME' seems to be an unknown game
variant.

Please report the following data to the ScummVM team at
https://bugs.scummvm.org/ along with the name of the game you tried to add and
its version, language, etc.:

Matched game IDs for the AGOS engine: feeble-win, dimp-win, jumble-win,
puzzle-win, swampy-win
Matched game IDs for the Tony Tough and the Night of Roasted Moths engine:
tony-win

  {"data1.cab", 0, "79006d2c70761dbec747e95953e636d2", 451220},
```
This means that the user first sees the unknown game dialog asking to report the game to the ScummVM team before seeing the proper add game dialog for Myst.

This has been tested with the following version of Myst ME:
```
        // Myst Masterpiece Edition
        // English Windows
        // From clone2727
        {
                {
                        "myst",
                        "Masterpiece Edition",
                        AD_ENTRY1("myst.dat", "c4cae9f143b5947262e6cb2397e1617e"),
                        Common::EN_ANY,
                        Common::kPlatformWindows,
                        ADGF_NO_FLAGS,
                        GUI_OPTIONS_MYST_ME
                },
                GType_MYST,
                GF_ME,
                0,
        },
```

### DIFF
--- a/gui/launcher.cpp
+++ b/gui/launcher.cpp
@@ -553,15 +553,15 @@ bool LauncherDialog::doGameDetection(const Common::String &path) {
 	// could be contained in the specified directory.
 	DetectionResults detectionResults = EngineMan.detectGames(files);
 
-	if (detectionResults.foundUnknownGames()) {
+	Common::Array<DetectedGame> candidates = detectionResults.listRecognizedGames();
+
+	if (candidates.size() == 0 && detectionResults.foundUnknownGames()) {
 		Common::String report = detectionResults.generateUnknownGameReport(false, 80);
 		g_system->logMessage(LogMessageType::kInfo, report.c_str());
 
 		UnknownGameDialog dialog(detectionResults);
 		dialog.runModal();
 	}
-
-	Common::Array<DetectedGame> candidates = detectionResults.listRecognizedGames();
 
 	int idx;
 	if (candidates.empty()) {


Mataniko:
I think the idea here is to encourage users to report unknown variants. Even if we correctly guess the game, you can't assume that a variant will work properly which might cause more frustration.

There should be a better balance here where we tell the user we recognize the game, but still prompt them to report the new variant.

criezy:
I think the idea of this PR is that if an exact match is detected, in this case the Myst ME, then we don't want to report incorrect unknown game variants detection on the same files from other engines.

Another test case for this is the Soltys (cge) and Sfinx (cge2) games (both freely available on our web site) that use the same files (`vol.cat` and `vol.dat`). This for examples results in the Soltys files being detected as an exact match from the cge engine, and being reported as an unknown game variant from the cge2 engine.

This is related to the work done by @bgK in PR #1187. In that PR he indicates that his change opens the door to some features such as 
> Ignoring unknown game detection results from one engine when there is an exact match from another (removing false positives when two games from different engines use files with the same name for detection)

That being said I am not convince the change in this PR is good. It might silence too many game variant detection reports. Some of things to be careful about are:
- Could a game be listed as a detected game by an engine while still being at the same time an unknown variant? Some engine may accept to run unknown variants. So even if the list of candidates is not empty we may still want to report the unknown variant.
- Case where we have two different games in the directory (which might mostly be an issue with the mass detection that does recursive search?), one of which being an unknown variant.

The comment from bgK I quoted above does give some pointers about unknown entries we might want to silence: when there is a candidate from a different engine on the same files (same path and same file name).

jcrona:
I agree that my PR might silence too many reports.

The real issue to me is that this dialog feels like an error, while the "real" game has actually been properly detected.
One way to deal with it might be to say just that in the dialog, explaining that those N games have been detect (in this case Myst ME), and that some unknown variants might have been detected too.

However, I do not like the fact that every time I am going to add Myst ME, I will get an unknown variant detection report...
And in this case, since the report is triggered by a file (data1.cab) that actually belongs to Myst ME, it means that there is no way someone can put in the same folder Myst ME and an AGOS or TONY game. So this particular report will always be wrong.

One solution might be to implement an exclusive logic in the engines, so that when there is an exact match for one engine, all candidates from excluded engines are removed since necessarily wrong.

Mataniko:
We don't know that Myst is the right game here, there are other games that detect data1.cab. It just so happens that we also detect your game.

The right thing to do is to just add your hash to the engine, not modify the behavior across the entire list of engines.

jcrona:
Hi Mataniko,

Actually, the Mohawk engine is not using data1.cab but myst.dat to detect Myst ME. And on that file a positive match already occurred. So we know at least that Myst is actually present in the folder.

bluegr:
Revisiting this, the change in the detection scheme is correct: if we get an exact match from one engine, then we don’t want to list it as an unknown variant from another engine.

Regarding BgK’s points:

- It’s not possible to have an exact match in one engine, and an unknown variant in another
- If we have two games in the same folder, then they are not properly placed anyway. Again, the only issue here is that the unknown game dialog wouldn’t appear in such a scenario, which is an edge case anyway, and is not normal

Overall, I got a positive feedback on this change, which is, IMHO, towards the right direction

bgK:
I'm sorry for the late feedback, but the change proposed here is not correct. It's possible for a game to be both runnable and unknown (`candidates` is not empty and `detectionResults.foundUnknownGames()` is true), especially for fallback detected entries. In which case the unknown game dialog needs to be shown.

So either @jcrona wants to implement a more precise rule as mentioned by @criezy or this should be closed.

bluegr:
Hm, true, the fallback detector could lead us into a situation like the one that @bgK describes. Perhaps we could check if the list of candidates we have contains entries found via fallback detection

bgK:
> Perhaps we could check if the list of candidates we have contains entries found via fallback detection

Fallback detection is a feature of the Advanced Detector only. The abstraction the launcher uses has no idea it exists nor should it. The whole point of the `hasUnknownFiles` flag is for a detector to signal the caller that the user should be warned and incited to report the unknown version.

My suggestion is the following:
* Run the detectors as is done currently
* If there are multiple results show the list of detected games. For games with unknown files, add a "unknown variant" label suffix to the game description.
* If the user selects a known game, add it.
* If the user selects a game with unknown files, show the unknown game dialog.
* In the unknown game dialog, if the game can be added, display a button "Add anyway". Another button "Back" is always available and goes back to the detected games list.

bluegr:
This has now been superseded by #1616. Closing

PR #1497, opened 2019-02-10
SKY: Decrease cursor redraw delay in menu

lotharsm:
Currently, there's a very noticeable mouse cursor lag in the save/load/settings dialog in Beneath a Steel Sky. I noticed that for whatever reason the redrawing delay is set to 50ms which causes this behavior. By reducing the redrawing delay to 20ms I achieved a very smooth cursor behavior which now matches the behavior of the original interpreter (tested with the latest DosBox release).

Even though this change seems to be quite trivial, I'm submitting this patch as a pull request since I am not familiar with the engine code at all. I couldn't spot any unwanted side effects.

### DIFF
--- a/engines/sky/control.cpp
+++ b/engines/sky/control.cpp
@@ -488,7 +488,7 @@ void Control::doControlPanel() {
 		_text->drawToScreen(WITH_MASK);
 		_system->updateScreen();
 		_mouseClicked = false;
-		delay(50);
+		delay(20);
 		if (!_controlPanel)
 			return;
 		if (_keyPressed.keycode == Common::KEYCODE_ESCAPE) { // escape pressed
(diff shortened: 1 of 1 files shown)


bluegr:
Since the delay is the same in all places that it’s used, it should be placed in a define at the top of control.cpp

lotharsm:
@bluegr, there's a delay of 150ms that's used in `void Control::animClick`.

bluegr:
@lotharsm: the 150ms delay is applied when a mouse click event occurs. The other delays of 20ms are applied during control panel animations. Thus, these two delays can be defined on top, as follows:

    #define ANIM_DELAY 20
    #define CLICK_DELAY 150

lotharsm:
@bluegr, I added the #defines as you suggested.

May I ask why you voted for a #define here instead of using `const int`? Just curious about it.

bluegr:
Thanks! Well, there's no real reason, really. In this case, a const int can be used just as well, or you could just use an enum

lotharsm:
Thanks for your feedback! Ready to go now? :)

bluegr:
Your patch is correct, as the delay is applied in all the graphics update loops in the control panel, so the 50ms delay is a bit high. The lag isn't THAT noticeable, IMHO, but it's there, and indeed this patch reduces it.

Merging

PR #1499, opened 2019-02-10
DINGUX: Fix dinguxsdl graphics backend

gameblabla:
The dingux backend port will not compile anymore (bool setGraphicsMode(const char *name) no longer exists it seems) so removing it allows GCC to compile it.
There's also an issue with ScummVM's cursor being drawn in-game, this affects games like Gabriel.
Mainline SDL code fixed this but the Dingux port was not brought up to par so fixing that of course.

(On a side note, i did notice that some other backends similar to this one (GPH, MotoSDL etc...) can suffer from the same issues)

### DIFF
--- a/backends/graphics/dinguxsdl/dinguxsdl-graphics.cpp
+++ b/backends/graphics/dinguxsdl/dinguxsdl-graphics.cpp
@@ -136,7 +136,7 @@ void DINGUXSdlGraphicsManager::initSize(uint w, uint h, const Graphics::PixelFor
 }
 
 void DINGUXSdlGraphicsManager::drawMouse() {
-	if (!_cursorVisible || !_mouseSurface) {
+	if (!_cursorVisible || !_mouseSurface || !_mouseCurState.w || !_mouseCurState.h) {
 		_mouseBackup.x = _mouseBackup.y = _mouseBackup.w = _mouseBackup.h = 0;
 		return;
 	}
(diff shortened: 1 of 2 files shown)


digitall:
Not sure what is going on there, but bool setGraphicsMode(const char *name) is defined by the OSystem layer and should be implemented by the backend. Default implementation and definition are here:
https://github.com/scummvm/scummvm/blob/master/common/system.cpp#L121
https://github.com/scummvm/scummvm/blob/master/common/system.h#L566

The rest of the code changes I am sure are fine and limited to Dingux so if you take a look at fixing the Dingux OSystem API for the function then I will merge this to tree...

gameblabla:
Ah, i guess that makes sense but i don't see any platforms in backends/platform actually implementing it except for DS/PSP and in those cases, they are stubs. I'm guessing it's linking against that function in system.cpp instead. Not sure why that specific line of code is needed in the header when it's basically not used by other platforms in their own headers.
In the future, i should probably just use the SDL code and do it from scratch, (that half-line GFX code looks kind of nasty) but right now it does not even compile out of the box and it works fine for me as it is. So what do you think i should ?

digitall:
Hmm. I think you are right i.e. if it is not overridden and reimplemented by the backend, then the default implementation will apply and that should be fine for most backends as long as the version that takes id works. Will merge code as-is then. Thanks for your contribution.

PR #1502, opened 2019-02-12
iOS: Implement support for kFeatureVirtualKeyboard

criezy:
This allows showing and hiding the keyboard when starting or ending text editing.

In this implementation I have decided to hide the keyboard when ending text editing only in landscape mode. This is because the keyboard is automatically shown when switching to portrait mode, and not just when starting editing. This way the keyboard remains always shown in portrait mode, whether editing text or not.

I have tested this with by giving/removing focus to the search field in the launcher, and with the save game dialog.

The idea for this came from reviewing PR #1487 and I think complement that PR nicely.

### DIFF
--- a/backends/platform/ios7/ios7_osys_main.cpp
+++ b/backends/platform/ios7/ios7_osys_main.cpp
@@ -171,6 +171,7 @@ bool OSystem_iOS7::hasFeature(Feature f) {
 	switch (f) {
 	case kFeatureCursorPalette:
 	case kFeatureFilteringMode:
+	case kFeatureVirtualKeyboard:
 		return true;
 
 	default:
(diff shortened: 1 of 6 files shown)


jonnyijapan:
@criezy 

Looks good. It's cool, I didn't know ScummVm had this feature, or that it had stuff concerning Feature:s working like this. I tried it and it works, and I also tried merging this with my three-finger swipe branch and they work well together.

One thing, if we go on and merge both of these branches, maybe it would be better to have my branch use `[self showKeyboard];` and `[self hideKeyboard];` that you introduced so that everything is streamlined and the `_keyboardVisible` flag is updated properly when the keyboard is manually shown/hidden.

criezy:
This has now been merged manually.

PR #1521, opened 2019-03-03
SCI/COMMON: Fix sync36 patch files, matchString

sluicebox:
Fixes a regression from three years ago that broke SCI's ability to load sync36 patch files. This broke GK1 lip syncs and possibly other games. Bug #10788.

06641f29a7bdcda280b0291f215193f055c83969 added a '#' wildcard to Common::matchString that matches against any digit. SCI sync36 patch files start with the '#' character and SCI uses the pattern '#???????.???' to locate them. The '#' wildcard broke this pattern as it doesn't match against the '#' character.

This fix adds an escape to matchString and updates SCI to use it.


bluegr:
Nice work! Merging

PR #1526, opened 2019-03-03
GUI: Allow GUI background colors other than black in Classic Theme

lotharsm:
I am currently working on a "Classic Mode" theme that uses another background color than the default block.

For "unknown" reasons (at least they are unknown to me), the current code throws an assertion failure when the background color in the theme is not set to black. I almost assume that this is a leftover from previous times, but since I'm not absolutely sure if this will break something else, I'm submitting this as a PR.

### DIFF
--- a/graphics/VectorRendererSpec.cpp
+++ b/graphics/VectorRendererSpec.cpp
@@ -2422,10 +2422,10 @@ drawBevelSquareAlgClip(int x, int y, int w, int h, int bevel, PixelType top_colo
 			++ptr_y;
 		}
 	} else {
-					while (i-- ) {
-						blendFillClip(ptr_left, ptr_left + w, ptr_x, ptr_y, _bgColor, 200);
-						ptr_left += pitch;
-					}
+		while (i-- ) {
+			blendFillClip(ptr_left, ptr_left + w, ptr_x, ptr_y, _bgColor, 200);
+			ptr_left += pitch;
+		}
 	}
 
 	x = MAX(x - bevel, 0);
--- a/graphics/VectorRendererSpec.cpp
+++ b/graphics/VectorRendererSpec.cpp
@@ -2353,7 +2353,6 @@ drawBevelSquareAlg(int x, int y, int w, int h, int bevel, PixelType top_color, P
 	ptr_left = (PixelType *)_activeSurface->getBasePtr(x, y);
 	i = h;
 	if (fill) {
-		assert((_bgColor & ~_alphaMask) == 0); // only support black
 		while (i--) {
 			darkenFill(ptr_left, ptr_left + w);
 			ptr_left += pitch;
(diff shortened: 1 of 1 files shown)


digitall:
Hmm... That line with the assert was added by commit 905f9591ab664cdef498d689d413ea2d40ebb442

digitall:
The second instance is a cut-and-paste with modification from that first instance done in commit ee00156d5450b01508894013dc446ef3d928d781

digitall:
@lotharsm : If you want some context, check https://logs.scummvm.org/log.php?log=scummvm.log.20Feb2012&format=html from around 20.32 onwards. There was no clear ScummBot notification for that commit as it was either bundled as a single push...

digitall:
Overall, it is not clear to me why that was limited to black background color unless it was just "only for Classic theme" sanity check. Maybe asked @wjp if he can remember, but probably fine to remove that IMHO.

criezy:
A quick look at commit 905f959 makes me think that the idea was to replace alpha blending with the background colour when drawing the bevel with a faster alternative that assumes that the background colour is black. So if the background colour is not black, the call to `darkenFill` will not give the expected result. And I assume the assert was added to highlight this.

I assume there was a reason to try to make the classic theme as fast as possible. If we want the option to use a non-black background colour while keeping the optimisation when the background colour is black, maybe something like below would work:
```
// Optimize the case where the background color is black
if ((_bgColor & ~_alphaMask) == 0) {
	while (i--) {
		darkenFill(ptr_left, ptr_left + w);
		ptr_left += pitch;
	}
} else {
	while (i--) {
		blendFill(ptr_left, ptr_left + w, _bgColor, 200);
		ptr_left += pitch;
	}
}

```

bluegr:
Agree with @criezy’s analysis. The asserts were most probably added to optimize for black backgrounds

bluegr:
@lotharsm: both functions perform filling, one also does clipping, while the other doesn’t. Thus, both asserts need to be adjusted with the code that criezy wrote.

lotharsm:
I adjusted the code so it now matches @criezy's suggestions. I hope I haven't skrewed up the syntax of blendFillClip(), but at a first check, everything compiles and runs as expected.

graphics/VectorRendererSpec.cpp
```diff
@@ -2409,13 +2414,18 @@ drawBevelSquareAlgClip(int x, int y, int w, int h, int bevel, PixelType top_colo
...
 		while (i--) {
 			darkenFillClip(ptr_left, ptr_left + w, ptr_x, ptr_y);
 			ptr_left += pitch;
 			++ptr_y;
 		}
+	} else {
+					while (i-- ) {
+						blendFillClip(ptr_left, ptr_left + w, ptr_x, ptr_y, _bgColor, 200);
+						ptr_left += pitch;
+					}
```
bluegr: Spacing is wrong here

lotharsm: Thanks, solved in 29295ac

bluegr:

PR #1527, opened 2019-03-05
CRYOMNI3D: New engine

lephilousophe:
This is still a work in progress but I think it's mature enough to be seen by other eyes than mine.

This is a new engine which implements for now the game Versailles 1685 (a "fake 3D" adventure game from Cryo).
For now, only the first chapter is completable. I still have to implement the other levels.
Major game mechanisms are implemented (3D view, dialogs, general level logic, inventory, toolbar).
Side mechanisms are there too (loading and saving, music, documentation area).
What miss is the levels logic with objects handling and state variables.
I still need to handle all ScummVM interactions (pausing and loading/saving from outside the game) maybe with the help from someone else.

The code for this engine may could be used later for Egypt and China at least (games I own). Although there is code in common (like the Omni3D stuff) there is also a lot of code specific for each game. I am not still sure about what will be really common.


bluegr:
Very nice indeed! The Cryo games will be a great addition!

Having said that, from an earlier discussion, I was under the impression that the code which handles character drawing is pure 3D (at least for Atlantis), so there should be a part of the engine that isn’t pseudo, but real 3D... is this correct?

criezy:
Nice!
I guess I will now have to find some time to undust my Mac version of the game and try to see how different it is from the DOS and Windows versions.

engines/cryomni3d/versailles/engine.cpp
```diff
@@ -0,0 +1,1443 @@
...
+	// functor is deleted in ZoneFixedImage
+	functor = nullptr;
+
+	if (_nextPlaceId == -1u) {
+		_forcePaletteUpdate = true;
+	}
+}
+
+unsigned int CryOmni3DEngine_Versailles::getFakeTransition(unsigned int actionId) const {
+	for (const FakeTransitionActionPlace *ft = kFakeTransitions; ft->actionId != nullptr; ft++) {
```
bgK: `ft->actionId != nullptr` results in a compilation failure when building with GCC's `-std=c++11`. `ft->actionId` is not a pointer.

lephilousophe:
Thanks @bgK for your review. I just pushed two commits which fix the points you mentioned.
@bluegr: In Versailles, characters are pre-rendered and it's a short video running in loop. Not sure for Egypt and China as they have "OmniSync" thing which synchronize speech with lips.
@criezy: Maybe the files are the same, the 2-CD set I got has common data for DOS and Windows.

BTW, I forgot to mention that only french is implemented for now because all the strings are stored in the program and not in a data file. These datas are in a dedicated C++ file for now but I think we will have to generate some ScummVM specific data file to avoid having it in the code.

I forgot to remove some debug define which skipped startup videos... That's now fixed.

engines/cryomni3d/detection_tables.h
```diff
@@ -0,0 +1,114 @@
...
+			Common::FR_FRA,
+			Common::kPlatformDOS,
+			ADGF_UNSTABLE,
+			GUIO1(GUIO_NOASPECT)
+		},
+		GType_VERSAILLES,
+		0,
+		0,
+	},
+
```
criezy: To detect my French Macintosh version (the executable is named "Versailles"):
```
	// Versailles 1685
	// French Macintosh
	// From criezy
	{
		{
			"versailles",
			"",
			AD_ENTRY1s("Versailles", "f81935517b1bbb58acf70f25efa5c7f3", 375868),
			Common::FR_FRA,
			Common::kPlatformMacintosh,
			ADGF_UNSTABLE,
			GUIO1(GUIO_NOASPECT)
		},
		GType_VERSAILLES,
		0,
		0,
	},
```

engines/cryomni3d/detection_tables.h
```diff
@@ -0,0 +1,114 @@
...
+		GType_VERSAILLES,
+		0,
+		0
+	},
+};
+
+static const ADFileBasedFallback fileBased[] = {
+	{ &fallbackDescs[0].desc,  { "VERSAILL.EXE", 0 } },
+	{ &fallbackDescs[0].desc,  { "VERSAILL.PGM", 0 } },
+	{ 0, { 0 } }
```
criezy: I also added the Macintosh executable here:
```
	{ &fallbackDescs[0].desc,  { "Versailles", 0 } },
```

engines/cryomni3d/dialogs_manager.cpp
```diff
@@ -0,0 +1,558 @@
...
+	for (; lineP != nullptr && *lineP != ':'; lineP = nextLine(lineP)) { }
+
+	/* Return the label name without it's ':' */
+	return nextChar(lineP);
+}
+
+Common::String DialogsManager::findVideo(const char *data) const {
+	data = previousMatch(data, ".FLC");
+	if (data == nullptr) {
+		return nullptr;
```
criezy: I get an error here: no conversion from nullptr to Common::String
So I changed this to:
```
		return Common::String();
```

engines/cryomni3d/fixed_image.h
```diff
@@ -0,0 +1,128 @@
...
+#ifndef CRYOMNI3D_FIXED_IMAGE_H
+#define CRYOMNI3D_FIXED_IMAGE_H
+
+#include "common/func.h"
+
+#include "engines/cryomni3d/cryomni3d.h"
+#include "engines/cryomni3d/objects.h"
+
+namespace Graphics {
+class Surface;
```
criezy: `Surface` is a struct, not a class.

engines/cryomni3d/mouse_boxes.h
```diff
@@ -0,0 +1,73 @@
...
+
+#ifndef CRYOMNI3D_MOUSE_BOXES_H
+#define CRYOMNI3D_MOUSE_BOXES_H
+
+#include "common/array.h"
+#include "common/str.h"
+
+namespace Common {
+class Point;
+class Rect;
```
criezy: `Point` and `Rect` are structs, not classes.

engines/cryomni3d/sprites.h
```diff
@@ -0,0 +1,102 @@
...
+#include "common/str.h"
+
+#include "graphics/cursor.h"
+
+namespace Common {
+class ReadStream;
+}
+
+namespace Graphics {
+class Surface;
```
criezy: Surface is a struct, not a class.

engines/cryomni3d/versailles/engine.cpp
```diff
@@ -0,0 +1,1443 @@
...
+	fonts.push_back("font02.CRF");
+	fonts.push_back("font03.CRF");
+	fonts.push_back("font04.CRF");
+	fonts.push_back("font05.CRF");
+	fonts.push_back("font06.CRF");
+	fonts.push_back("font07.CRF");
+	fonts.push_back("font08.CRF");
+	fonts.push_back("font09.CRF");
+	fonts.push_back("font10.CRF");
+	fonts.push_back("font11.CRF");
```
criezy: For the macintosh version I had to change this as the fonts are named differently. I don't know if I am supposed to load all the fonts there, but this is what I did.

```
	Common::Array<Common::String> fonts;
	if (getPlatform() == Common::kPlatformMacintosh) {
		fonts.push_back("MPW12.CRF");
		fonts.push_back("MPW9.CRF");
		fonts.push_back("arial11.CRF");
		fonts.push_back("fruitL10.CRF");
		fonts.push_back("fruitL11.CRF");
		fonts.push_back("fruitL12.CRF");
		fonts.push_back("fruitL13.CRF");
		fonts.push_back("fruitL14.CRF");
		fonts.push_back("fruitL16.CRF");
		fonts.push_back("fruitL18.CRF");
		fonts.push_back("fruitL9.CRF");
		fonts.push_back("garamB18.CRF");
		fonts.push_back("garamB20.CRF");
		fonts.push_back("garamB22.CRF");
		fonts.push_back("geneva10.CRF");
		fonts.push_back("geneva12.CRF");
		fonts.push_back("geneva13.CRF");
		fonts.push_back("geneva14.CRF");
		fonts.push_back("geneva15.CRF");
		fonts.push_back("geneva9.CRF");
		fonts.push_back("helvet11.CRF");
		fonts.push_back("helvet12.CRF");
		fonts.push_back("helvet13.CRF");
		fonts.push_back("helvet14.CRF");
		fonts.push_back("helvet16.CRF");
		fonts.push_back("helvet24.CRF");
		fonts.push_back("helvet9.CRF");
	} else {
		fonts.push_back("font01.CRF");
		fonts.push_back("font02.CRF");
		fonts.push_back("font03.CRF");
		fonts.push_back("font04.CRF");
		fonts.push_back("font05.CRF");
		fonts.push_back("font06.CRF");
		fonts.push_back("font07.CRF");
		fonts.push_back("font08.CRF");
		fonts.push_back("font09.CRF");
		fonts.push_back("font10.CRF");
		fonts.push_back("font11.CRF");
	}
```

engines/cryomni3d/versailles/engine.cpp
```diff
@@ -0,0 +1,1443 @@
...
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/spr8col", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/spr8col/bmpok", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/wam", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/objets", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/gto", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/dial/flc_dial", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/dial/voix", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/textes", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/music", 1);
+	SearchMan.addSubDirectoryMatching(gameDataDir, "datas_v/sound", 1);
```
criezy: In the Macintosh version I have, on the CDs some of the subdirectories are actually in a `DATAV_HD` directory while others are indeed in `Datas_V`. I just decided to move everything to `Datas_V` when copying to the disk, but if we wanted to support the original structure here are those that are in `DATAV_HD` originally:
- fonts
- GTO
- menu
- spr8col
- textes
- wam

And for information in `DATAV_HD` I also have the following directories that are not listed here (and so far it seems to work without the need to add them to the search paths):
- low_ram
- SAVEGAME

engines/cryomni3d/versailles/engine.h
```diff
@@ -0,0 +1,453 @@
...
+#include "cryomni3d/sprites.h"
+#include "cryomni3d/wam_parser.h"
+
+#include "cryomni3d/versailles/documentation.h"
+#include "cryomni3d/versailles/toolbar.h"
+#include "cryomni3d/versailles/dialogs_manager.h"
+
+namespace Graphics {
+class ManagedSurface;
+class Surface;
```
criezy: `Surface` is a struct, not a class.

engines/cryomni3d/cryomni3d.h
```diff
@@ -0,0 +1,164 @@
...
+#include "engines/engine.h"
+
+#include "font_manager.h"
+#include "objects.h"
+#include "sprites.h"
+
+class OSystem;
+
+namespace Common {
+class Point;
```
criezy: `Point` is a struct, not a class.

engines/cryomni3d/versailles/engine.cpp
```diff
@@ -0,0 +1,1443 @@
...
+			// TODO: cursorUseZones
+			moving = false;
+		} else {
+			if (!exit) {
+				// TODO: countdown
+				g_system->updateScreen();
+			}
+		}
+		if (!exit && !leftButtonPressed) {
+			g_system->delayMillis(50);
```
bgK: Is there any reason for using such a long delay? It keeps the framerate low; the game could run very smoothly on modern systems.

lephilousophe: I put that at the beginning of the development.
The original game doesn't have this sleep time and it makes all movements way too fast.
In the same way, the original game uses CPU at 100% which this delay prevents.

Maybe it can be decreased but that will mean to adjust Omni3D movements at least.

What do you mean when you say that the framerate is low? Do you mean that the fps figure is low or that movements are slow?

engines/cryomni3d/dialogs_manager.cpp
lephilousophe: I don't even know how GCC let that pass without any error or warning.
I will try to address your comments this evening.

engines/cryomni3d/versailles/engine.cpp
lephilousophe: This one may be a little harder than that.
In PC version the fonts are kind of selected by their order. For example, if we switch two push_back, the fonts will be switched on display.
So with your patch, only the first 11 fonts will be used. I am not sure what Cryo did with all these fonts in Mac version.
I am not even sure all 11 fonts are used in PC version.

engines/cryomni3d/versailles/engine.cpp
lephilousophe: Maybe I should remove the datas_v component from the search path. In fact, at the root there is the program and datas_v. This is not really useful, it justs mimics the Windows installation and CDs.
Maybe having every directory directly at root is good enough.

DATAV_HD is the contents of what will be installed on hard drive. There is the same kind of distinction on PC CD: it's still called DATAS_V but placed elsewhere on the disk.
SAVEGAME is used to store saves and it's not need in ScummVM as I use the facilities provided.
I don't know what low_ram contains maybe it's some smaller file to use when there isn't enough memory?

criezy: The datas_v level could indeed be removed.

On the mac version there is actually no installer. The installation instructions just tell the user to copy a folder from the first CD to the hard drive, and that folder contains both the executable and the DATAV_HD directory. So on the hard drive we end up with a directory that is still named DATAV_HD.

I will check the content of the low_ram directory this evening. I was indeed thinking along the same line as you, but that folder only contains a few txt files from what I remember, so that seemed strange.

engines/cryomni3d/versailles/engine.cpp
criezy: I suspected if would be more complicated that that ;-)
The changed allowed me to at least start the game, but the text display didn't look quite right. I guess I will have to do some digging in the executable...

engines/cryomni3d/versailles/engine.cpp
bgK: I mean the frames per second figure is low. On my system, the game runs at 15 fps. Mouse movement is not very smooth, and neither is the Omni3D movement. Perhaps the delay could be reduced and the Omni3D movement speed computed relative to the game speed to have framerate independence.

lephilousophe: I see what you mean and I agree. This just needs some thinking and checks.
I am not sure if there are other parts of the code which depends on this wait.
I will try to see that next week-end.

lephilousophe:
@criezy: I just applied your fixes and your proposal for MacOS support.
I removed DATAS_V folder level.
I would like to correct what I said earlier: savegame directory is useful for tour mode support.
It contains the bootstrap visit used to switch to this mode.
If the folder exists the tour mode is enabled, else the entry menu is not shown.

criezy:
I just started to look at the executable (and test ghidra at the same time) and the very first thing it does in the main is call a init_low_ram() function which itself check the existence of 4 files to set some global variables to either 0 or 1. It appears to be a rudimentary configuration system to enable or disable some features.

So I looked at what I have in my `low_ram` folder and it confirms that suspicion. There are 4 text files with the following content:

0dec.txt:
```
En cas de problème de fonctionnement sur des machines très limitées en mémoire,
déplacez ce fichier dans le répertoire de l'application Versailles.
```

0depl.txt:
```
En cas de problème de fonctionnement sur des machines très limitées en mémoire,
déplacez ce fichier dans le répertoire de l'application Versailles.

Cela supprimera les animations de déplacement
```

0mus.txt:
```
En cas de problème de fonctionnement sur des machines très limitées en mémoire,
déplacez ce fichier dans le répertoire de l'application Versailles.

Cela empechera de jouer la musique.
```

0par.txt:
```
En cas de problème de fonctionnement sur des machines très limitées en mémoire,
déplacez ce fichier dans le répertoire de l'application Versailles.

Cela supprimera les paroles des dialogues.
```

I think we should be fine in term of memory usage for ScummVM, so we probably don't have to worry too much about this :-P

criezy:
I have started to look at the code that loads the font in the mac executable. So far I have found out that it loads all of the fonts except one. But I have yet to check the places where they are used, so I don't yet know which ones are really used.

The loading is done in that order (the one commented out is the one that exists but is not loaded):
```
		fonts.push_back("garamB18.CRF");
//		fonts.push_back("garamB20.CRF");
		fonts.push_back("garamB22.CRF");
		fonts.push_back("geneva15.CRF");
		fonts.push_back("geneva14.CRF");
		fonts.push_back("geneva13.CRF");
		fonts.push_back("geneva12.CRF");
		fonts.push_back("geneva10.CRF");
		fonts.push_back("geneva9.CRF");
		fonts.push_back("helvet24.CRF");
		fonts.push_back("helvet16.CRF");
		fonts.push_back("helvet14.CRF");
		fonts.push_back("helvet13.CRF");
		fonts.push_back("helvet12.CRF");
		fonts.push_back("helvet11.CRF");
		fonts.push_back("helvet9.CRF");
		fonts.push_back("fruitL9.CRF");
		fonts.push_back("fruitL10.CRF");
		fonts.push_back("fruitL11.CRF");
		fonts.push_back("fruitL12.CRF");
		fonts.push_back("fruitL13.CRF");
		fonts.push_back("fruitL14.CRF");
		fonts.push_back("fruitL16.CRF");
		fonts.push_back("fruitL18.CRF");
		fonts.push_back("arial11.CRF");
		fonts.push_back("MPW12.CRF");
		fonts.push_back("MPW9.CRF");
```

And with that order I actually get something that looks very good (which doesn't mean that it is correct, but I will need more time to check that).

lephilousophe:
Sorry for the force-push, I forgot to tag a commit message.

@criezy I managed to find which fonts were used and which ones weren't on MacOS.
@bgK I didn't understood where you were annoyed by the delay and thought it was in events polling function.
That's way simpler where you reacted. I removed the delay in warp handling and divided the deltas by 5 (it should have been 6 but it's already quite slow now in normal mode, maybe I should use a 4).
I also lowered the delays in fading functions to let updateScreen be called more often.

tsoliman:
It's ok to force push unless it is the main repo.

bluegr:
You mentioned Egypt and China... how about the 3 Atlantis games? They’re available for purchase on gog.com

lephilousophe:
@tsoliman: Thanks, it's good to know.
@bluegr: I mentioned Egypt and China because I played them when I was younger. I never played Atlantis though I own CDs of Atlantis I and II. I admit I would like to implement games I already know before the others (if nobody wants to implement them before) because I think it will spoil my discovery of the games.

engines/cryomni3d/versailles/logic.cpp
```diff
@@ -0,0 +1,1036 @@
...
+					_nextPlaceId = _currentPlaceId;
+				}
+				_inventory.removeByNameId(100);
+				// Revealed paper
+				collectObject(98, fimg);
+				_gameVariables[GameVariables::kGotRevealedPaper] = 1;
+				setGameTime(3, 1);
+			} else if (objID == 96) {
+				// Pamphlet about arts
+				playInGameVideo("PAP_BRUL");
```
criezy: In the mac version this is actually called `PAP-BRUL` (with a dash and not an underscore), so it currently fails with 'WARNING: Failed to open movie file PAP_BRUL/PAP_BRUL.hns!` when getting to this point.
Is this a difference between the Mac and Windows/DOS version, or a typo here (and in the other places where this video is played)?

lephilousophe: It seems they mixed dashes and underscores during development. In PC version all paths are converted from dashes to underscore and there is no file with dash on the CDs.
I will have to do some substitution at various places...

lephilousophe:
I just finished implementation and basic tests of level 2.
There are a bunch of fixes for various things I didn't notice before.
There are also some features I didn't implement before but which are needed for level 2.
Times may be a little mixed up because I tried to sort my commits to avoid build failure between two of them.
@criezy: I haven't found time to fix underscores yet. I will try to fix it this evening or tomorrow.

lephilousophe:
@criezy I just dropped a patch to fix your problem with dashes.
Before, I translated file names with dashes to underscore directly in constants.
Now, I keep the dashes in file names and then, on PC versions, I translate them to underscores. That's like in original executable.

lephilousophe:
I am still working on it. I have almost finished to implement level 3.
It will need tests and debug after that and I will be able to push the changes.
I hope to push it during next week.

i30817:
Salammbo: battle for carthage is probably the last game using this engine (or a closely related successor).

bluegr:
There are about 15 games using this engine, although later versions may be using true 3D graphics for displaying the character closeups. Here's the list:

https://www.mobygames.com/game-group/graphics-engine-omni3d

bluegr:
Just wanted to mention that there has been work on the Cryo Interactive games in the past by @Templier, which can be found in the following public branches:

https://github.com/Templier/scummvm/tree/cryo/engines/cryo
https://github.com/Templier/scummvm/tree/ring/engines/ring

In the first engine (cryo), work was done on the game China: The Forbidden City

In the second engine (ring), work was done on the following games:

- Pilgrim: Faith as a Weapon
- The Ring: The Legend of the Nibelungen
- Faust: Seven Games of the Soul
- TimeScape: Journey to Pompeii
- Pilgrim 2: The Legend of the Prophet and the Assassin
- Pilgrim 3: The Secrets of Alamut
- Jerusalem: The Three Roads to the Holy Land

It seems like parts of @Templier's cryo engine and this one could be merged in the future?

lephilousophe:
I already saw the cryo branch of @Templier.
I planned to get from it what I could when I start to work on Egypt and China.
The ring engine seems to be quite distinct from the other games so I am not sure it's a good idea to try a merge.

criezy:
I have both the Ring and Faust games, and while they were published by Cryo, they were actually both developed by Arxel Tribe. I have only very briefly looked at those, and it was a while ago, but from what I remember they indeed use a different engine from the Cryo games (I think Arxel Tribe developed their own graphic engine called CINview?).

lephilousophe:
I finally finished to implement Versailles 1685 game.
I think the engine is ready to merge it just misses some bugfix on events I will put in a separate pull request.

engines/cryomni3d/font_manager.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+#include "common/debug.h"
+#include "common/file.h"
+#include "graphics/managed_surface.h"
+
+#include "engines/cryomni3d/font_manager.h"
+
+namespace CryOmni3D {
+
+FontManager::FontManager() : _currentFont(nullptr), _transparentBackground(false),
```
sev-: You could actually implement it as another Common::Font subclass, and use the rendering and word wrapping facilities from there. Could save you from writing tons of code.

engines/cryomni3d/image/codecs/hlz.cpp
```diff
@@ -0,0 +1,140 @@
...
+	decodeFrameInPlace(stream, -1, dst);
+
+	return _surface;
+}
+
+Graphics::PixelFormat HLZDecoder::getPixelFormat() const {
+	return Graphics::PixelFormat::createFormatCLUT8();
+}
+
+static inline bool getReg(Common::SeekableReadStream &stream, uint32 *size, uint32 *reg,
```
sev-: Why static?

engines/cryomni3d/image/hlz.h
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+/**
+ * @file
+ * Image decoder used in engines:
```
sev-: It is better to remove this leftover.

engines/cryomni3d/module.mk
```diff
@@ -0,0 +1,38 @@
+MODULE := engines/cryomni3d
+
+MODULE_OBJS = \
+	cryomni3d.o \
```
sev-: I recommend sorting these alphabetically.

engines/cryomni3d/sprites.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+void Sprites::loadSprites(Common::ReadStream &spr_fl) {
+	byte magic[4];
+
+	while (true) {
+		if (spr_fl.read(magic, sizeof(magic)) == 0) {
+			break;
+		}
+		if (memcmp(magic, "SPRI", sizeof(magic))) {
```
sev-: I am not sure how well it will work, but the common way would be:

```
  if (spr_fl.readUint32BE() != MKTAG('S','P','R','I'))
      error("Invalid sprite magic");
```

engines/cryomni3d/versailles/dialogs_manager.cpp
```diff
@@ -0,0 +1,367 @@
...
+		// Empty wave file
+		delete audioDecoder;
+
+		unsigned int duration = 100 * text.size();
+		if (duration < 1000) {
+			duration = 1000;
+		}
+
+		bool skipWait = false;
+		unsigned int end = g_system->getMillis() + duration;
```
sev-: If the duration is longer than 0.33 seconds here, the cursor lag could be noticeable. To avoid that, put `updateScreen()` and `delayMillis(10)` here (you've removed the delay from the `pollEvents()` earlier ;) )

engines/cryomni3d/versailles/dialogs_manager.cpp
```diff
@@ -0,0 +1,367 @@
...
+						_engine->setupPalette(palette, 0, 256);
+					}
+
+					// Only refresh the moving part of the animation
+					const Graphics::Surface subFrame = frame->getSubArea(drawRect);
+					g_system->copyRectToScreen(subFrame.getPixels(), subFrame.pitch, drawRect.left, drawRect.top,
+					                           subFrame.w, subFrame.h);
+				}
+			}
+			g_system->updateScreen();
```
sev-: Another place where `delayMillis(10)` should be inserted here instead of `pollEvents()`. Otherwise, your input reaction will feel laggy.

engines/cryomni3d/versailles/documentation.cpp
```diff
@@ -0,0 +1,2070 @@
...
+		const char *bmp;
+		Common::Point imgPos;
+		Common::Rect linesPos;
+		const Common::String *title;
+		Graphics::Surface highlightedImg;
+
+		Category(const char *record_, const char *bmp_, const Common::Point &imgPos_,
+		         const Common::Rect &linesPos_, const Common::String *title_) :
+			record(record_), bmp(bmp_), imgPos(imgPos_), linesPos(linesPos_), title(title_) { }
+	} categories[8] = {
```
sev-: make it const?

engines/cryomni3d/versailles/documentation.cpp
```diff
@@ -0,0 +1,2070 @@
...
+				// Go back to current record
+				break;
+			} else if (nextRecord != "planG") {
+				_currentRecord = nextRecord;
+				break;
+			} else {
+				// We can't go up to previous case, so let's do a round
+				action = 7;
+				break;
+			}
```
sev-: I recommend moving `break;` statement out of the ifs to the `case 8:` level, so it does not look like a fall through statement.

engines/cryomni3d/versailles/documentation.cpp
```diff
@@ -0,0 +1,2070 @@
...
+		Graphics::Surface highlightedImg;
+
+		Area(const Common::Point &areaPos_, const char *bmp_, unsigned int messageId_,
+		     const char *record_ = nullptr) :
+			areaPos(areaPos_.x, areaPos_.y, areaPos_.x, areaPos_.y), record(record_), bmp(bmp_),
+			messageId(messageId_), message(nullptr) { }
+		Area(const Common::Rect &areaPos_, unsigned int messageId_, const char *record_ = nullptr) :
+			areaPos(areaPos_), record(record_), bmp(nullptr), messageId(messageId_), message(nullptr) { }
+	} areas[] = {
+		Area(Common::Point(174, 181), "APL.bmp", 74),
```
sev-: Why do you want to create all these objects on heap instead of making them const and moving outside of the method?

engines/cryomni3d/versailles/engine.cpp
```diff
@@ -0,0 +1,1594 @@
...
+				gameStep();
+
+				// We shouldn't return from gameStep without an abort command
+				assert(_abortCommand != AbortNoAbort);
+				switch (_abortCommand) {
+				default:
+					error("Invalid abortCommand: %d", _abortCommand);
+					// Shouldn't return
+					return Common::kUnknownError;
+				case AbortFinished:
```
sev-: please consider renaming them to kAbort... in accordance with our code naming conventions.

engines/cryomni3d/versailles/engine.h
```diff
@@ -0,0 +1,632 @@
...
+	AbortLoadGame = 2,
+	AbortNewGame = 3,
+	AbortNextLevel = 5,
+	AbortFinished = 6,
+	AbortGameOver = 7
+};
+
+struct GameVariables {
+	enum Var {
+		kCollectScore = 0,             // 0
```
sev-: I recommend renaming these to kVar... to vaoid confusion with other enums.

engines/cryomni3d/versailles/dialogs_manager.cpp
bgK: It might be useful to be aware that `OSystem::updateScreen` implicitly waits for the next VSync on systems with VSync enabled (most desktop platforms these days). Assuming a 60Hz display the wait can be up to 16.6 ms if a frame boundary was just passed. Maybe you want to account for that, maybe not.

sev-: Where do you see that in code?

bgK: If it's implicit, it's not in our code :)
This page explains nicely how wait for VSync works: https://www.khronos.org/opengl/wiki/Swap_Interval

sev-: Oh. If OpenGL backend uses it, then it could have a negative side effects. We recommend to call updateScreen() as often as an engine wants, and then backend should just update the physical screen only when it needs to. Basically, some engines could call it as much as 200 times per second (`delayMillis(5)`).

engines/cryomni3d/versailles/logic.cpp
```diff
@@ -0,0 +1,4650 @@
...
+void CryOmni3DEngine_Versailles::drawSafeDigits(Graphics::ManagedSurface &surface,
+        const Graphics::Surface(&bmpDigits)[10], const unsigned char (&safeDigits)[kSafeDigitsCount]) {
+	for (unsigned int i = 0; i < ARRAYSIZE(safeDigits); i++) {
+		const Graphics::Surface &digit = bmpDigits[safeDigits[i]];
+		Common::Point dst(kSafeDigitsX[i % 4], kSafeDigitsY[i / 4]);
+		surface.transBlitFrom(digit, dst);
+	}
+}
+
+const char *CryOmni3DEngine_Versailles::kSafeDates[] = { "1643", "1668", "1674" };
```
sev-: This is named incorrectly, it is not a constant but an array

engines/cryomni3d/versailles/logic.cpp
```diff
@@ -0,0 +1,4650 @@
...
+			fimg->_exit = true;
+			break;
+		}
+		if (fimg->_zoneUse && fimg->_currentZone == 0 && currentGameTime() >= 2) {
+			// Collected quill
+			collectObject(128, fimg);
+			_gameVariables[GameVariables::kCollectQuill] = 1;
+			// Try to solve epigraph
+			ZonFixedImage::CallbackFunctor *functor =
+			    new Common::Functor1Mem<ZonFixedImage *, void, CryOmni3DEngine_Versai
```
sev-: Please use `byte` here

sev-: consider using `FROM_BE_32(_countdownValue)`

engines/cryomni3d/versailles/menus.cpp
```diff
@@ -0,0 +1,1068 @@
...
+#endif
+				else if (*it == 26 && !_isPlaying) {
+					*it = -26;
+				} else if (*it == 29 && !_isPlaying) {
+					*it = -29;
+				} else if (*it == -42 && canVisit()) {
+					*it = 42;
+				} else if (*it == 48) {
+					unsigned int omni3D_speed = ConfMan.getInt("omni3d_speed");
+					switch (omni3D_speed) {
```
sev-: I recommend adding default case here as well.

engines/cryomni3d/versailles/music.cpp
```diff
@@ -0,0 +1,280 @@
...
+#include "common/error.h"
+#include "common/file.h"
+
+#include "cryomni3d/versailles/engine.h"
+
+namespace CryOmni3D {
+namespace Versailles {
+
+const char *CryOmni3DEngine_Versailles::kMusicFiles[8][8] = {
+	{ "1amb", }, // Level 1
```
sev-: I think a bunch of warnings will be generated here about unitialised members.

engines/cryomni3d/video/hnm_decoder.cpp
```diff
@@ -0,0 +1,400 @@
...
+	//uint32 filesize = stream->readUint32LE();
+	stream->skip(4);
+	uint32 frameCount = stream->readUint32LE();
+	//uint32 tabOffset = stream->readUint32LE();
+	stream->skip(4);
+	uint16 soundBits = stream->readUint16LE();
+	uint16 soundChannels = stream->readUint16LE();
+	uint32 frameSize = stream->readUint32LE();
+
+	char unknownStr[16];
```
sev-: please use `byte` here. `char` is not portable

engines/cryomni3d/wam_parser.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include "engines/cryomni3d/omni3d.h"
+
+namespace CryOmni3D {
+
+void WAMParser::loadStream(Common::ReadStream &stream) {
+	char str[16];
+
+	_places.clear();
+
+	stream.readByte();
```
sev-: Consider putting `(void)` in front to designate clearly that you want to ignore the result

engines/cryomni3d/versailles/dialogs_manager.cpp
bgK: > We recommend to call updateScreen() as often as an engine wants, and then backend should just update the physical screen only when it needs to. Basically, some engines could call it as much as 200 times per second (delayMillis(5)).

This is not good. `updateScreen` needs to be called exactly once per engine frame. Even if nothing was drawn (to allow the backend to display its own information, to resync with vsync, possibly to allow programs that hook in, like Steam overlay, to update). Calling it more often will cause unexpected delays if wait for vsync is enabled.

Ideally the sleep time should be variable to account for the time where the CPU was stalled waiting for vsync.

 A simple game loop might be:
```cpp
const uint32 targetFrameDuration = 1000 / 60;

while (!shouldQuit()) {
    uint32 frameStart = _system->getMillis();

    doFrame();
    _system->updateScreen();

    uint32 loopElapsed = _system->getMillis() - loopStart;
    if (loopElapsed < targetFrameDuration)
        _system->delayMillis(targetFrameDuration - loopElapsed);
}
```

engines/cryomni3d/dialogs_manager.cpp
```diff
@@ -0,0 +1,558 @@
...
+		if (!strncmp(ifLine, "AND IF ", 7)) {
+			ifLine += 7;
+		} else {
+			finishedConditions = true;
+		}
+	}
+
+	/* We are in the (implicit) THEN part of the IF
+	 * ifLine points to the instruction */
+	return ifLine;
```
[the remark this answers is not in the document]
lephilousophe: I wanted to implement the same algorithms there was in the game to be sure it would behave the same. But I agree that could be done in lex/yacc.
I didn't know this was allowed in scummvm.

engines/cryomni3d/font_manager.cpp
lephilousophe: Same as above, I wanted to have the exact same behaviour. Maybe it's not what scummvm aims.

engines/cryomni3d/image/codecs/hlz.cpp
lephilousophe: It's a function not part of the the class much like a macro. I put it static to not populate global symbols and to avoid name clashes (it's local to current cpp file).

engines/cryomni3d/versailles/dialogs_manager.cpp
sev-: What you say is exactly the opposite of the original design of the method.

Engines should not know about hardware specifics. Moreover, as you explained yourself, only a few backends have a notion of vsync, or otherwise enforced frame rate.

Thus, the code loop you put there is, in fact, belongs to the internals of updateScreen(). In there, if your backend cares about the framerate, you need to skip unneeded calls if you're being called too often.

Also, most of the engines supported by ScummVM did not care about the framerate, and would require serious refactoring, especially when the main loop of the engine is in scripts. And I understand that you're coming from the 3d engines of ResidualVM where it was modernised, thus, all of this may sound lame to you, but this is the specifics of the 80s and 90s :).

engines/cryomni3d/versailles/documentation.cpp
lephilousophe: I wanted to avoid calling constructor at startup of scummvm. It's an array of objects with constructors that will be called at init time.

engines/cryomni3d/versailles/engine.h
lephilousophe: I put them in GameVariables structure to avoid clash. These enum members can only be accessed using GameVariables::kStuff.

engines/cryomni3d/versailles/logic.cpp
```diff
@@ -0,0 +1,4650 @@
...
+#include "common/system.h"
+
+#include "cryomni3d/fixed_image.h"
+
+#include "cryomni3d/versailles/engine.h"
+
+namespace CryOmni3D {
+namespace Versailles {
+
+static const char *imagesObjects[] = {
```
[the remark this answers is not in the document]
lephilousophe: As above, static here is because variable is not a member of class so I make it local to cpp file.
This array is static const of basic type and ends up being created at compilation.

engines/cryomni3d/versailles/logic.cpp
```diff
@@ -0,0 +1,4650 @@
...
+	surface.transBlitFrom(_sprites.getSurface(4), Common::Point(spriteX, spriteY),
+	                      _sprites.getKeyColor(4));
+	/*
+	_fontManager.displayInt(markers[id].x - 10, markers[id].y, id + 14);
+	}
+	*/
+}
+
+// This array contains images for all paintings it must be kept in sync with _paintingsTitles
+static const char *imagesPaintings[] = {
```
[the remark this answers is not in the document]
lephilousophe: Same as above :)

engines/cryomni3d/versailles/music.cpp
lephilousophe: It's a static const variable of basic type so it get's zero initialized as allowed by C++ standard.
No warning there with GCC nor Clang in Travis CI.

lephilousophe:
Thank you for your review.
I already reviewed parts in place. I will fix parts where I didn't comment during this week.
For my comments please confirm if it doesn't follow coding standard of the project.

The only uncertainty I get is for all the pollEvents and delayMillis thing.
As there is some discussion about what to do between you and @bgK. I am not sure about what to do to be clean, portable and having it still working.
Original game engine doesn't sleep or anything like that so it tends to use some CPU.
Maybe music handling lets it yield but I am not sure about it.

About cleanup of detection.cpp, it's indeed full of leftovers but what annoys me is that I didn't finish to integrate engine with scummvm for everything about saves.

engines/cryomni3d/dialogs_manager.cpp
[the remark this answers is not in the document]
sev-: Heh: https://github.com/scummvm/scummvm/blob/master/engines/director/lingo/lingo-gr.y

engines/cryomni3d/versailles/logic.cpp
lephilousophe: I am not sure to understand what you mean here

engines/cryomni3d/versailles/logic.cpp
lephilousophe: It seems there are many engines using kConstant names for arrays. Should I really change it?

engines/cryomni3d/versailles/documentation.cpp
lephilousophe: I can't make it const because there is a Surface in each object which gets written below when loading the image file

engines/cryomni3d/dialogs_manager.cpp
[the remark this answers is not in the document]
lephilousophe: Should I move to lex/yacc or can I let it like it is?
The main advantage of the current situation is that it works and is compliant with the original game.

engines/cryomni3d/versailles/logic.cpp
lephilousophe: I assume this review is for when I mangle _countdownValue and the game variable.
It's not exactly a swap between bytes but more a conversion from an int to a char array with a colon in the middle so using FROM_BE_32 would not do what is intended.

lephilousophe:
I finished the most simple editions. I now need to tackle the delayMillis/pollEvents/updateScreen problem.
With all the discussion I am not quite sure to understand what I need to do to avoid using to much CPU but keep UI responsive.
I will try to check what other engines do.

engines/cryomni3d/versailles/logic.cpp
sev-: Any example? In the end, since the usage of this array is pretty local, it may stay as is.

engines/cryomni3d/versailles/logic.cpp
sev-: Neither do I. I could refer to `unsigned char` but this comment is obviously on the wrong line. I couldn't find it.

sev-: Okay.

engines/cryomni3d/dialogs_manager.cpp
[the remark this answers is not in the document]
sev-: You can let it as is, especially since it has been debugged. It was just an open question from me.

sev-:
In order to keep the UI responsive, you need to call in your loop the OSystem in the following order:

```
while(1) {
   _engine->pollEvents();
   // do your game logic
   g_system->updateScreen();
   g_system->delayMillis(10);
}
```

Currently, you keep the delay inside of your `pollEvent` call, and that is a misconception: that call should not have this side effect. As a result of this, your screen updates may look laggy, as you have the delay _before_ you blit your updates to the screen.

This is why I recommend removing the `delayMillis()` call out of the pollEvents(), and adding it to every place where you have a loop, right after the updateScreen() call.

In case you do not call `pollEvents()` and do just the screen update and delay, your cursor will freeze on the screen.

engines/cryomni3d/versailles/logic.cpp
lephilousophe: Using : `grep -R ' k[A-Z].*\[\]' ./engines/` I found that tony, pegasus, sci, gob, draci, dreawweb, supernova, gnap, neverhood, bladerunner, illusions and cryo have some. bbvs has a lot of them.

lephilousophe:
@sev- I think I addressed the changes you requested.
I noticed some more bugs in the process and fixed them.
Is there anything left I let slip?

sev-:
Very good! All is great! I am merging it now, please watch the buildbot, and I hope you will join the team and continue working on the engine.

neuromancer:
@lephilousophe is the [demo](https://archive.org/details/playhns_003) supported?

lephilousophe:
@neuromancer It's not the game. It's only a video about it.
That's not supported by ScummVM. Video decoding is there and playing could be done but I am not sure it's useful.

PR #1529, opened 2019-03-09
SCI: Fix SQ4 zero gravity blast lockup, bug #10912

sluicebox:


bluegr:
If you set the no collide flag of the bullet, how will shooting ego still work correctly?

sluicebox:
Great question, ego getting shot is independent of colliding with the bullet or laser line. The shooting script determines whether or ego is going to get hit based on a number of factors before it draws anything. If ego is going to get shot then it sets a flag, then draws the laser line and runs the blast animation, and then tests that flag afterwards to see if it should proceed with ego getting hit.

bluegr:
Thanks for the explanation. Merging

PR #1539, opened 2019-03-18
SCI: Apply SQ4 timing workaround to non-VGA versions

sluicebox:
Fixes bug #6193 by applying the SQ4 floppy kPaletteAnimation timing
workaround from 99de89c974fc24bf58b034842750e522d7d441d4 to versions
with less than 256 colors. This includes DOS EGA, PC-98, and Amiga.

This also patches out the intro script's workaround that uses different
timing in EGA mode, as this is incompatible with ours and makes the
intro screen last over 3 minutes instead of 5 seconds.

### DIFF
--- a/engines/sci/engine/kgraphics.cpp
+++ b/engines/sci/engine/kgraphics.cpp
@@ -646,18 +646,17 @@ reg_t kPaletteAnimate(EngineState *s, int argc, reg_t *argv) {
 	bool paletteChanged = false;
 
 	// Palette animation in non-VGA SCI1 games has been removed
-	if (g_sci->_gfxPalette16->getTotalColorCount() < 256)
-		return s->r_acc;
-
-	for (argNr = 0; argNr < argc; argNr += 3) {
-		uint16 fromColor = argv[argNr].toUint16();
-		uint16 toColor = argv[argNr + 1].toUint16();
-		int16 speed = argv[argNr + 2].toSint16();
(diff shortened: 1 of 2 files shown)


engines/sci/engine/kgraphics.cpp
```diff
@@ -646,18 +646,17 @@ reg_t kPaletteAnimate(EngineState *s, int argc, reg_t *argv) {
...
-	if (g_sci->_gfxPalette16->getTotalColorCount() < 256)
-		return s->r_acc;
-
-	for (argNr = 0; argNr < argc; argNr += 3) {
-		uint16 fromColor = argv[argNr].toUint16();
-		uint16 toColor = argv[argNr + 1].toUint16();
-		int16 speed = argv[argNr + 2].toSint16();
-		if (g_sci->_gfxPalette16->kernelAnimate(fromColor, toColor, speed))
-			paletteChanged = true;
+	if (!(g_sci->_gfxPalette16->getTotalColorCount() < 256)) {
```
bluegr: This is hard to read now. It should be changed to:
if (g_sci->_gfxPalette16->getTotalColorCount() == 256)

sluicebox: Good call, I wasn't sure if there was some edge case where it could ever be greater.

bluegr:
Nice work! Merging

PR #1540, opened 2019-03-19
IOS7: Add Remapping of Arrow Keys missing from On Screen Keyboard

digitall:
These are remapped to ']' for Up, '\' for Down, '_' for Left and '|' for
Right.

This code is slightly modified from a forum post by donkthemagicllama.

### DIFF
--- a/backends/platform/ios7/ios7_osys_events.cpp
+++ b/backends/platform/ios7/ios7_osys_events.cpp
@@ -430,6 +430,25 @@ void  OSystem_iOS7::handleEvent_keyPressed(Common::Event &event, int keyPressed)
 		keyPressed = Common::KEYCODE_RETURN;
 		ascii = Common::ASCII_RETURN;
 		break;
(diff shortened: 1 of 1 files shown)


digitall:
@criezy: Since I am not an IOS developer / user, I don't really have a clear opinion on the details of this. I was just transferring a cleaned up version of the patch from this forum post:
https://forums.scummvm.org/viewtopic.php?f=15&t=14453

Overall, I see your point on "_" in savegame naming, though I don't think the symbols are used apart from that. Can you suggest an alternative based on the IOS keyboard layout?

"implement an input accessory view to add a row of keys at the top of the standard iOS keyboard (with the function keys, arrow keys, escape key, maybe also one to open the GMM)" sounds like the right and less hacky long term solution... but no one has provided any code for that yet, so probably should merge this with minor modifications and a FIXME note about that so it is replaced when a IOS developer has time / energy

criezy:
This is now superseded by PR #1706.
Sorry that it took me so long to finally take a look at this.

PR #1255, opened 2018-07-17
WIP: Scaler plugins rebased

sandsmark:
doesn't even build yet, but it's a start

edit: as @sev- said below, this is a rebase of PR #271 on top of master.

PR #1541, opened 2019-03-19
WIN32: Add scummvm.exe.manifest file and disable HiDPI scaling (on hold)

lotharsm:
This PR adds an Windows Application Manifest file that Windows XP+ uses for setting various options during the execution of the program.

This manifest file disables the HiDPI scaling on Windows 7+, to fix the blurry output on HiDPI displays. This is currently most noticable when using the GUI and adds a slight blur to the games.

By declaring ScummVM as "DPI aware", we permit Windows do perform it's own scaling. Disabling scaling by the OS is already the default on Linux, maybe @criezy can comment on the macOS side of things.

A test build is available through https://cloud.serra.network/index.php/s/jsJCjs7X6dnsBwa/download.

In this screenshot, the upper screenshot shows the current behavior on a HiDPI display (look at the GUI scaling), where the lower screenshot shows the behavior with this manifest file on the exact same machine.
scummvm-hidpi-scaling

The screenshots give you a vague representation, on a real HiDPI display the difference is quite obvious.

### DIFF
--- a/dists/scummvm.rc
+++ b/dists/scummvm.rc
@@ -15,6 +15,7 @@
 IDI_ICON               ICON    DISCARDABLE     "icons/scummvm.ico"
 IDI_COUNT              ICON    DISCARDABLE     "icons/count.ico"
 
+CREATEPROCESS_MANIFEST_RESOURCE_ID RT_MANIFEST "dists/win32/scummvm.exe.manifest"
 ID_GDF_XML             DATA    "dists/win32/scummvm.gdf.xml"
 
 scummclassic.zip       FILE    "gui/themes/scummclassic.zip"
@@ -106,7 +107,7 @@ BEGIN
             VALUE "FileDescription", "ScummVM: A free implementation of various adventure game engines\0"
             VALUE "FileVersion", SCUMMVM_VERSION "\0"
             VALUE "InternalName", "scummvm\0"
-            VALUE "LegalCopyright", "Copyright © 2001-2019 The ScummVM Team\0"
+            VALUE "LegalCopyright", "Copyright ï¿½ 2001-2019 The ScummVM Team\0"
             VALUE "LegalTrademarks", "'SCUMM', and all SCUMM games are a TM of LucasArts. Simon The Sorcerer is a TM of AdventureSoft. Beneath a Steel Sky and Broken Sword are a TM of Revolution. Flight of the Amazon Queen is a TM of John Passfield and Steve Stamatiadis. \0"
             VALUE "OriginalFilename", "scummvm.exe\0"
             VALUE "ProductName", "ScummVM\0"
--- a/dists/scummvm.rc
+++ b/dists/scummvm.rc
@@ -107,7 +107,7 @@ BEGIN
             VALUE "FileDescription", "ScummVM: A free implementation of various adventure game engines\0"
             VALUE "FileVersion", SCUMMVM_VERSION "\0"
             VALUE "InternalName", "scummvm\0"
-            VALUE "LegalCopyright", "Copyright ï¿½ 2001-2019 The ScummVM Team\0"
+            VALUE "LegalCopyright", "Copyright © 2001-2019 The ScummVM Team\0"
             VALUE "LegalTrademarks", "'SCUMM', and all SCUMM games are a TM of LucasArts. Simon The Sorcerer is a TM of AdventureSoft. Beneath a Steel Sky and Broken Sword are a TM of Revolution. Flight of the Amazon Queen is a TM of John Passfield and Steve Stamatiadis. \0"
             VALUE "OriginalFilename", "scummvm.exe\0"
             VALUE "ProductName", "ScummVM\0"
--- a/devtools/update-version.pl
+++ b/devtools/update-version.pl
@@ -49,6 +49,7 @@ my @subs_files = qw(
 	dists/gph/README-GPH
 	dists/gph/scummvm.ini
 	dists/riscos/!Boot,feb
+	dists/win32/scummvm.exe.manifest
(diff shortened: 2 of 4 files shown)


dafioram:
Does this fix Trac#10587?

lotharsm:
I haven't tested this yet since I don't have the build mentioned in this ticket anymore.

But since I could "fix" this by disabling hidpi scaling for the ScummVM executable through the compatibility options, it is very likely that #10587 will be fixed too.

lotharsm:
@dafioram, regarding Trac#10587: I ran a few tests and the current master branch still showed this crash once. Seems slightly better with SDL 2.0.9 now.

However, during those tests, disabling the HiDPI scaling via this PR seems to fix the issues for me since I couldn't replicate the crash here.

dafioram:
Sounds like a win.

lotharsm:
At least I couldn't spot any unwanted side effects during my tests. If the application is run on Windows XP which doesn't understand the DPI awareness setting in the manifest file, it will simply ignore this manifest directive.

dists/scummvm.rc
```diff
@@ -106,7 +107,7 @@ BEGIN
             VALUE "FileDescription", "ScummVM: A free implementation of various adventure game engines\0"
             VALUE "FileVersion", SCUMMVM_VERSION "\0"
             VALUE "InternalName", "scummvm\0"
-            VALUE "LegalCopyright", "Copyright � 2001-2019 The ScummVM Team\0"
+            VALUE "LegalCopyright", "Copyright � 2001-2019 The ScummVM Team\0"
```
criezy: Why is this line changed? Was the encoding of the file changed maybe?
The `?` character here is a copyright character, and the encoding of the file is I think ISO 8859-1 originally.

lotharsm: This is most likely cause by Visual Studio Code which for whatever reason decided to treat this file as UTF-8. I'll revert it to ISO 8859-1.

lotharsm: Fixed by 6b0dd0c.

lotharsm:
After merging this PR, we could tackle the issues with the GUI being too small on 4K displays using SDL_GetDisplayDPI by providing a "highres" theme and do the GUI scaling there (e.g. starting with a 640x400 display and providing larger fonts right in the theme file), just like we already do for "lowres"...

bluegr:
I'm using Windows 10, with a UHD resolution (3200x1800). By disabling HiDPI scaling, ScummVM looks miniscule in this resolution, and is practically unreadable. I've even tried to use a 3x filter (Control-Alt-+), but the results are still bad.

It's possible to disable HiDPI scaling from Windows, by right-clicking scummvm.exe and going to Properties->Compatibility->Change High DPI settings->set High DPI scaling override to "Application"

Thus, I believe that it's wrong to just disable HiDPI scaling, without the application actually supporting such high resolutions. And since it's possible for users to override this behavior, IMHO it's wrong to disable it like this.

We can disable it in the future, once we can support arbitrary scalers (e.g. an 8x scaler). Have a look at pull requests #1255 and #271. Until then, my opinion is that this feature should remain enabled as it is right now.

lotharsm:
> I'm using Windows 10, with a UHD resolution (3200x1800). By disabling HiDPI scaling, ScummVM looks miniscule in this resolution, and is practically unreadable. I've even tried to use a 3x filter (Control-Alt-+), but the results are still bad.

Thanks for your feedback! Are you referring to the GUI here or also to the games themselves? Regarding the GUI, I am currently working on proper DPI detection via SDL_GetDisplayDPI(), creating a windows with 800x600 and then using the theme file to upscale the GUI:
hidpi-demo

bluegr:
@lotharsm: Yes, I'm talking about the GUI. I usually like to play games in fullscreen, but that's me :)

sluicebox:
Adding the manifest this way breaks Visual Studio builds atm. The VS project is set to automatically generate a manifest and add it to the resource script. This a creates duplicate resource, linker errors. The VS project settings need to be changed to disable manifest generation if we're explicitly adding one.

manifest

I'm also concerned about this making the default window size impractically small on high dpi. The machine I'm at now is Win7 2560x1440 with text size at "Medium - 125%" in the Windows parlance, I didn't see any difference, maybe I need to have it higher, but I think we need to understand the ramifications of this more. I already feel like the initial size isn't that useful to me at this resolution.

lotharsm:
@sluicebox, thanks for your comment regarding VS! I wasn't aware that VS is not smart enough to ask the user what to do at this point... ;)

I must confess that I'm a bit lost regarding the proper GUI size for now...

sluicebox:
Making things look nice at all dpi settings is a chore and you're a hero for taking it on. Every time I've done it I felt like I was starting over from scratch since it's all so app-specific.

Although it looks like I'm not affected by this yet, I'm leaning towards what @bluegr is saying and holding off until there's scaling to compensate. Even that first screenshot of Windows doing the scaling looks fine to me, but maybe that's a limitation of screenshots, or maybe I just have bad taste. =)

lotharsm:
> Making things look nice at all dpi settings is a chore and you're a hero for taking it on. Every time I've done it I felt like I was starting over from scratch since it's all so app-specific.

Thanks :) Yep, everything is app-specific in this regard. However, I already figured out where I could hook up the DPI detection in main.cpp (or is it possible to change the window resolution at a later point?) for the SDL ports and then, based on the new window size, provide a proper theming through the theme files.

sev-:
@lotharsm Do I understand correctly, that it is OK to close this PR, and wait until the runtime detection is implemented?

lotharsm:
@sev- Correct - this PR is more or less obsolete.

PR #1542, opened 2019-03-19
BASE: Enable aspect ratio correction by default

lotharsm:
Currently, the aspect ratio correction is disabled by default.

I hereby propose to enable it by default instead, since I suspect that the vast majority of the users use this function anyway. However, feel free to discard this PR if you think we shouldn't change this.

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -210,7 +210,7 @@ void registerDefaults() {
 	// Graphics
 	ConfMan.registerDefault("fullscreen", false);
 	ConfMan.registerDefault("filtering", false);
-	ConfMan.registerDefault("aspect_ratio", false);
+	ConfMan.registerDefault("aspect_ratio", true);
 	ConfMan.registerDefault("gfx_mode", "normal");
 	ConfMan.registerDefault("render_mode", "default");
 	ConfMan.registerDefault("desired_screen_aspect_ratio", "auto");


lotharsm:
@criezy, so do you think it would be better to put this "on hold" for at least until you worked on PR #1132?

criezy:
Maybe?

I guess the question here, is whether having the correct aspect ratio correction for all games would make me change my mind on whether it is a good idea or not to enable the correction by default. And I am still not sure that is the case. But I am not sure that is not the case either (I am really on the fence here).

Also seeing how long ago I opened that PR, I would not hold my breath on it being finished anytime soon ;-)

bonki:
I'm with @criezy on this, no strong opinion but a slight tendency against changing the default at this point.

bluegr:
Agree with the others. I have no strong opinion about this feature, since in some non-standard game resolutions it doesn't work as intended. So, enabling it makes sense only for some games at the moment (e.g. some FM-TOWNS games running at 320x240)

lotharsm:
Thanks a lot for your feedback! Closing this PR then for now.

AndywinXp:
Hey folks. Is there a way this can be re-taken into consideration? I'll quote myself from discord:

 

> Devs, I got told (what I think to be) a very valid point of discussion: we currently don't enforce aspect ratio correction by default. Why is that? Is there an historical reason why it is this way?
>
> Of course you all know that 320x200 games are not widescreen but are instead meant to be stretched on a 4:3 screen surface. In order to preserve the correct artistic vision we offer aspect ratio correction. But... Why isn't this a default thing? Are there known cases of games which would break if this setting was on by default?

Currently, for games which should not have aspect ratio correction whatsoever we have the `GUIO_NOASPECT` flag...

SirBogman:
I strongly feel that this default should be changed.

I don't understand why anyone would ever want to turn it off.
I think that many users aren't going to realize that this is a problem. They aren't going to change the setting. Things will be slightly wrong in most games and they probably won't notice. But it might make them feel like something is wrong. But they will not know what is wrong.

What's the point of having it be "pixel perfect" if the pixels are the wrong shape? Monitors these days are much larger than 320x200 so we are going to be scaling anyway.

Having the default be false means that the majority of users will never realize that the settings is wrong and won't change it. And overall it makes the experiences that people are having lower quality than they would have been.

SirBogman:
I have purchased games on GoG that were distributed with ScummVM, realized the aspect ratio was wrong, and then had to waste time looking into how to fix it. This should not be the default experience that most people will receive. They will likely not even realize what is wrong, but may have an uneasy feeling about it.

AndywinXp:
@SirBogman I strongly agree with you 🙂 I'll restore this PR asap.

SirBogman:
This is such a great project and it would be great to make it even better by default.

i30817:
When I used dosbox, I always added the pixel perfect patch (later superseded by the shader version of the same) for crisp aspect ratio adapting to big monitors. I wonder if something like that is possible instead of a naive thing that interacts with the default scaling...

PR #1544, opened 2019-03-20
SCI: PC-98 sound driver

athrxx:
This is a sound driver for the PC-98 versions of Quest for Glory, Police Quest II, Space Quest IV, Kings Quest V and Castle of Dr. Brain.

I commit this in the form of a pull request, since I've not just added the driver file, but also had to make a  change to the SCI audio code (commit 5c49a2a).

To be more specific, I had to add an implementation of SCI0 sound driver opcode 3 which initializes the sound resource in the driver. PQ II checks two different playId flags and assigns different channel types accordingly. This can't be done through the existing music engine code (especially not by abusing midi control change 0x4B which isn't even implemented in the original PQ II PC-98 driver).
The only one of the existing driver which uses some sort of init method is the MT-32/GM driver. I've replaced that method and updated that driver to the changes. 

I could also update the Adlib and CMS drivers, but I didn't want to do that in this PR. These drivers have channel mapping disabled via #ifdef. I don't know how well this works, but that's basically how they avoid the init track method with its channel assignments.

The PC-98 driver has been tested with the five games mentioned above and seems to work well. Dr. Brain was only tested quickly (no really pleasant music to listen to), but it has the exact same driver version (exact same byte size, exact same file time stamp) as SQ IV.

The PC-98 games all contain MT-32 resources, too. So there is probably not much demand for this driver. But people have even written PC speaker drivers, so why not have this, too...

### DIFF
--- a/engines/sci/module.mk
+++ b/engines/sci/module.mk
@@ -77,6 +77,7 @@ MODULE_OBJS := \
 	sound/drivers/fmtowns.o \
 	sound/drivers/midi.o \
 	sound/drivers/pcjr.o \
+	sound/drivers/pc9801.o \
 	video/seq_decoder.o
 
 
--- a/engines/sci/resource.h
+++ b/engines/sci/resource.h
@@ -654,6 +654,7 @@ public:
 	struct Track {
 		byte type;
 		byte channelCount;
+		SciSpan<const byte> header;
 		Channel *channels;
 		int16 digitalChannelNr;
 		uint16 digitalSampleRate;
--- a/engines/sci/resource_audio.cpp
+++ b/engines/sci/resource_audio.cpp
@@ -766,6 +766,8 @@ SoundResource::SoundResource(uint32 resourceNr, ResourceManager *resMan, SciVers
 		// Digital sample data included? -> Add an additional channel
 		if (resource->getUint8At(0) == 2)
 			_tracks->channelCount++;
+		// header information that can be passed to the SCI0 sound driver
+		_tracks->header = resource->subspan(0, _soundVersion == SCI_VERSION_0_EARLY ? 0x11 : 0x21);
 		_tracks->channels = new Channel[_tracks->channelCount];
 		channel = &_tracks->channels[0];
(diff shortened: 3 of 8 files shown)


bluegr:
Overall, very nice work! Well done :)

However, we will need to test the behavior of the new code for early SCI0 games (i.e. KQ4 early, LSL2 early and XMAS card 1988, especially the first two), since this changes the channel initialization for these games

athrxx:
Thanks!
Yes, testing the SCI0 early games would be good thing. Hopefully someone has any of these.
I don't think that my changes make any difference there, since Adlib and CMS have their #ifdefs and Midi doesn't use voice mapping. And I did keep the necessary reverb reset also for SCI0 early...

Well, this is the theory. Testing is of course still necessary. I tested only SCI0 late.

If I get my hands on a SCI0 early MT32 driver file I'll also have a look inside.

athrxx:
I found Xmas Card 1988. What a weird piece of software :-)
I can confirm that it works with Adlib and MT-32.
Our CMS driver doesn't even support SCI0.
While that software isn't really suitable for testing the more interesting part is that it does have all the driver files. I am going to take a look at the init track methods. The MT-32 driver is only 2 KB though, so I doubt that there is any complex magic in there.

athrxx:
I have now added the SCI0_EARLY related code to the midi driver's init method.

Actually the SCI0_EARLY init method has more flexibility than SCI0_LATE. It allows different channels to serve as rhythm and control channels. At least in Xmas Card 1988 there is no practical relevance though, since all of the 5 or 6 music tracks that this software has use channel 9 as rhythm channel and channel 15 as control channel.

Any tracks actually using this functionality might cause issues on GM devices, since the sysex message will not work for these.

Please keep in mind that this is only remotely related to the actual PR (PC-98 driver). I can always remove this code and just have the reverb reset as it is now... (Xmas Card 1988 doesn't even have the reverb reset. Also ,there isn't any patch or reverb data in that driver file).

bluegr:
Pretty solid work! Well done, the overall design is well thought and implemented

Merging

PR #1545, opened 2019-03-20
PINK: Add detection entry for Danish "pokus" version

lotharsm:
Detection information was provided by sauravisus in
https://bugs.scummvm.org/ticket/10919.

@whiterandrek, can you please take a look at this? Since the detection table for the PINK engine doesn't do any "fancy" stuff, the change itself is quite trivial and *should* be ready to go as is.


whiterandrek:
Looks good, thank you!

PR #1555, opened 2019-03-24
SCI: Fix SQ4CD biker timepod message, bug #10922

sluicebox:


sluicebox:
I remember getting confused by this bug in the 90s!

raziel-:
line 11696 and line 11695 needs to be switched as they break the room number ordering
A room 614 fix comes before a room 613 fix

sluicebox:
That's an intentional trade-off, the order is just organizational. There are other tables that don't make any attempt. 

In this case that prior line is for a patch applied to a lot of rooms with different patch counts in each one, so it's easier to make sense of that one when you can see all 5 of those entries grouped together. (At least it was for me when I wrote' em the other day)

bluegr:
I agree with @sluicebox that the ordering makes more sense the way it is now

Other than that, the patch looks good! Merging

raziel-:
Got it

PR #1556, opened 2019-03-25
SCI: Fix Mac icon bar crash on restart (KQ6, FPFP)

sluicebox:
Add support for reinitializing the mac icon bar when restarting.
Restarting runs the game's init script which calls KPlatform again.
Prior to this, restarting these games would fail an assertion.


bluegr:
Nice catch! Merging

PR #1559, opened 2019-03-27
SCI: Synchronize QFG4 control panel audio settings

sluicebox:
Fixes Audio button in control panel getting out of sync with
the message mode set by ScummVM.


bluegr:
Nice catch! Merging

PR #1560, opened 2019-03-27
SCI: Fix kAddMenu language separator parsing

sluicebox:
Fixes bug #10926 that prevents SQ3 German Amiga from loading


bluegr:
Nice! Merging

PR #1562, opened 2019-03-29
SCI: Remove MidiDriver_AmigaMac platform inference

sluicebox:
Tell MidiDriver_AmigaMac the game's platform instead of it
attempting to infer this by which patch resources exist.

Fixes bug #10925 where SQ3 German Amiga is treated as Mac because
it happens to contain a patch 7 resource.


sluicebox:
Thanks!

bluegr:
Thanks for the changes. Merging

PR #1563, opened 2019-03-31
GRAPHICS: Add support for PICT opcode $90 BitsRect

sluicebox:
Add limited support for unpacked PICT bits which FPFP Mac requires.

SCI games use packed PICT bits unless an image is less than 8 bytes
per row in which case they can't be packed, like FPFP's shovel icon.

Fixes bug #7059 which prevents the game from being completed.

### DIFF
--- a/image/pict.cpp
+++ b/image/pict.cpp
@@ -75,6 +75,7 @@ void PICTDecoder::setupOpcodesCommon() {
 
 void PICTDecoder::setupOpcodesNormal() {
 	setupOpcodesCommon();
+	OPCODE(0x0090, on_bitsRect, "BitsRect");
 	OPCODE(0x0098, on_packBitsRect, "PackBitsRect");
 	OPCODE(0x009A, on_directBitsRect, "DirectBitsRect");
 	OPCODE(0x8200, on_compressedQuickTime, "CompressedQuickTime");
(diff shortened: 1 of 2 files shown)


image/pict.cpp
```diff
@@ -344,13 +350,26 @@ void PICTDecoder::unpackBitsRect(Common::SeekableReadStream &stream, bool withPa
...
-			error("Unpacked DirectBitsRect data (padded)");
+			// only support 1 bpp for now as there is currently only one known
+			//  SCI pict that requires any unpacked support.
+			if (bytesPerPixel == 1 && packBitsData.pixMap.pixelSize == 8) {
+				for (uint16 b = 0; b < packBitsData.pixMap.rowBytes; b++) {
+					byte pixel = stream.readByte();
+					if (b < width) {
+						buffer[i * width + b] = pixel;
+					}
+				}
```
bluegr: You don’t need the for loop here, since this is just a direct copy of data.

You could just read all the bytes like this:
stream.read(buffer[i * width], packBitsData.pixMap.rowBytes);

Not sure why the width check is needed in this case, but you could use MIN() for that, if you want to keep it

sluicebox: Got rid of the loop, that's cleaner, but here's what's going on...

"width" and packBitsData.pixMap.rowBytes aren't always the same.

In the case of the shovel, there's a zero byte at the end of each row. This is padding that may or may not appear (alignment? 24-bit? see existing vague comment)

Even without the padding, these values might be different. The purpose of this opcode family is to copy bits and clip them to a rectangle. "width" has that clipping factored in, so they could be different for that reason too.

In the shovel's case, we need to consume 6 bytes from the stream but only apply 5 to buffer on each row.

This should cover all cases.

sluicebox:
shovel

Debugging uncompressed images are fun =)

bluegr:
Ah, right. Thanks for the explanation :)

Looks good now, merging

PR #1564, opened 2019-03-31
SCI: Fix Mac icon bar event handling

sluicebox:
This fixes two edge cases that allow mouse presses to fall through the icon bar in Mac versions of KQ6 and FPFP. Depending on the cursor this results in unexpected messages or walking to unexpected rooms when clicking an icon.

The problem is that the mac icon bar peeks at the first mouse press event in the queue whenever kGetEvent is called and handle only that message. After this kGetEvent continues to handle whatever is in the queue. This has two problems:

1. If there are two mouse press events in the queue when kGetEvent is called then the mac icon bar will handle the first and kGetEvent will handle the second, even though mac icon bar should have handled both. This is easily reproducible by clicking two mouse buttons at the same time. (Just mash 'em.)

2. If a mouse press event appears in the queue after mac icon bar peeks but before kGetEvent gets the next event then kGetEvent will handle it. This is harder to deliberately reproduce but program speed seems to be a factor. Just try playing either game for a few minutes. On my lower machine it's frequent and when I attach a debugger it's very frequent.

The fix is to remove queue peeking and allow the mac icon bar to handle every before kGetEvent does.

### DIFF
--- a/engines/sci/engine/kevent.cpp
+++ b/engines/sci/engine/kevent.cpp
@@ -49,13 +49,6 @@ reg_t kGetEvent(EngineState *s, int argc, reg_t *argv) {
 	SegManager *segMan = s->_segMan;
 	Common::Point mousePos;
 
-	// For Mac games with an icon bar, handle possible icon bar events first
-	if (g_sci->hasMacIconBar()) {
-		reg_t iconObj = g_sci->_gfxMacIconBar->handleEvents();
-		if (!iconObj.isNull())
-			invokeSelector(s, iconObj, SELECTOR(select), argc, argv, 0, NULL);
-	}
-
 	// If there's a simkey pending, and the game wants a keyboard event, use the
 	// simkey instead of a normal event
 	// TODO: This does not really work as expected for keyup events, since the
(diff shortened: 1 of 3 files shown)


sluicebox:
Well, I'm stumped, I don't know why gcc doesn't like this.

kevent.cpp can no longer see SciEventType, even though it's defined in event.h which it includes.

Help, I need an adult?

digitall:
@sluicebox : I am not a C++ language expert, but the error from travis CI's GCC log is:
engines/sci/engine/kevent.cpp:85:20: error: ‘SciEventType’ is not a class or namespace

https://en.cppreference.com/w/cpp/language/enum shows that these type of scoped enumerations are only valid on class or struct enum declarations. ... Maybe just try dropping the scope as I don't think it is needed to access the enum value so change to drop the scope. Will make a line comment

engines/sci/engine/kevent.cpp
```diff
@@ -78,6 +71,21 @@ reg_t kGetEvent(EngineState *s, int argc, reg_t *argv) {
...
+		reg_t iconObj = NULL_REG;
+		if (g_sci->_gfxMacIconBar->handleEvents(curEvent, iconObj)) {
+			if (!iconObj.isNull()) {
+				invokeSelector(s, iconObj, SELECTOR(select), argc, argv, 0, NULL);
+			}
+
+			// The mouse press event was handled by the mac icon bar so change
+			// its type to none so that generic event processing can continue
+			// without the mouse press being handled twice
+			curEvent.type = SciEventType::kSciEventNone;
```
digitall: Suggestion as per my other comment, change this to:
curEvent.type = kSciEventNone;

digitall: @sluicebox : Yep, have tested locally with gcc-8.2.0 on Linux x86_64. This branch fails with the same error, but making this change fixes compilation.

sluicebox: @digitall Thanks! I completely misread that error message in the exact same way a dozen times, never even looked at the right line of code once =)

bluegr:
Nice! Looks good now, and compiles properly. Thanks for your work!

Merging

PR #1566, opened 2019-03-31
COMMON: Move the Lua code from SWORD25 to COMMON

nipunG314:
This PR moves the Lua code from the SWORD25 engine to COMMON so that Lua can be integrated into other engines.


bluegr:
Thanks for this.

Is there any use for LUA scripts outside of the sword25 engine?

A lot of code has been removed from the LUA interpreter, since it is quite specific to the needs of the sword25 engine. If we need a more generic LUA interpreter, we should adapt the full original LUA code, not this engine-specific interpreter

nipunG314:
I need the LUA scripts to write the engine for Hyperspace Delivery Boy, so they will probably be used there. If the current LUA setup is not sufficient for this, then I suppose it would make sense to adapt the full original code.

bluegr:
Well, the original sources *have* been adapted, but some unused (in sword25) non-portable LUA functionality has been removed, or adapted. An example is the LUA I/O code. Check the history:

https://github.com/scummvm/scummvm/commits/master?after=62fc4c94bfd5b66fbfa444d686c5e86a84bea562+34&path%5B%5D=engines&path%5B%5D=sword25&path%5B%5D=util&path%5B%5D=lua

If you don’t need this functionality, then the interpreter can be used as-is, otherwise, it’ll need to be adapted for this engine. It would be best if you checked the game’s scripts for these cases.

sev-:
Thanks.

Now, one more thing is missing. Since now the code lives in the common/, it would be great if you could turn it into an engine feature, similar to png. See sword25/configure.engine

bluegr:
Any news on this?

bluegr:
Any news on this? Now there are conflicts because of changes in PR #1769

nipunG314:
@bluegr Sorry, got seriously sidetracked on this. I have rebased the breaking changes. Will work on the Engine Feature idea next

configure
```diff
@@ -3789,6 +3794,26 @@ case $_host in
 		;;
 esac
 
+#
+# Enable Lua only for engines that support it
```
bluegr: “Enable LUA only for engines that require it”

Kawa-oneechan: "Lua" is actually correct. It's not an acronym.

bluegr: You are quite correct, my mistake. It's "Lua"

configure
```diff
@@ -1030,6 +1032,7 @@ Optional Features:
   --disable-nuked-opl      don't build Nuked OPL driver
   --disable-16bit          don't enable 16bit color support
   --disable-highres        don't enable support for high resolution engines >320x240
+  --disable-lua            don't enable support Lua
```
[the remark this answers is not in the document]
bluegr: Please disregard this one. Your comment is correct, the proper naming is "Lua"

nipunG314:
create_msvc wasn't detecting the Lua engine feature. I have re-worked the feature as a default option, and added a Feature Flag for it.

configure
```diff
@@ -5397,8 +5405,8 @@ if test "$_build_scalers" = yes ; then
 	fi
 fi
 
-if test "$_mt32emu" = yes ; then
-	echo_n ", MT-32 emulator"
```
bluegr: I believe that these have been removed by mistake

common/module.mk
```diff
@@ -19,6 +19,39 @@ MODULE_OBJS := \
...
+	lua/lstrlib.o \
+	lua/ltable.o \
+	lua/ltablib.o \
+	lua/ltm.o \
+	lua/lua_persist.o \
+	lua/lua_persistence_util.o \
+	lua/lua_unpersist.o \
+	lua/lvm.o \
+	lua/lzio.o \
+	lua/scummvm_file.o \
```
bluegr: These should not be compiled in all the time in common code, which is why the feature flag has been requested. They should be enclosed in a ENABLE_LUA ifdef. i.e:

    ifdef ENABLE_LUA
    MODULE_OBJS += \
	xxx.o
    endif

Check common/module.mk for an example

configure
nipunG314: Re-added

common/module.mk
nipunG314: I have wrapped it now

nipunG314:
Any other changes required?

sev-:
Great, merging!

PR #1569, opened 2019-04-02
SCI: Fix SQ4CD Vohaul pocketpal text+speech mode

sluicebox:
Fixes a scene which is incompatible with our text+speech mode, bug #10241


bluegr:
Nice work! Merging

PR #1574, opened 2019-04-03
SCI: Attempt to fix keyboard controls in contained polygons - DO NOT MERGE

Kawa-oneechan:
Just to get people's attention, pretty wild issue with keyboard control.

A lot of rooms in SCI11 games have walkable areas defined with contained access polygons, covering the floor. Others have barred access polygons on the walls. Some games prefer one over the other, sometimes exclusively. The issue is that the former doesn't work right, returning incorrect points when trying to find out which is the closest in a given direction. Some games of course don't allow keyboard control at all, making the walk cursor itself move like the others.

For example, the lobby in Larry 6 has two barred polygons, and keyboard control is allowed. You can press the left or right arrow keys to navigate to either hallway, or go upstairs. These three rooms use contained polygons, and suddenly keyboard controls don't respond as they should, favoring mostly the left edge of the polygon no matter which direction you try to go.

Compare and contrast that with Pharkas, which favors barred polygons for at least the half dozen screens I tested.

I've narrowed it down to `Point::sqrDist` returning 0xFFFFFF for all points considered, because 30,000 is far further away than its cutoff point 4096. Removing this check yields values that are actually different. Different enough that my little test game worked perfectly fine, and Larry 6 was noticeably improved.

It feels like an ugly hack and I am _not_ a mathematically-inclined person, so I don't feel this is _quite_ the right solution. Hence the "do not merge".

### DIFF
--- a/engines/sci/engine/kpathing.cpp
+++ b/engines/sci/engine/kpathing.cpp
@@ -833,7 +833,10 @@ int PathfindingState::findNearPoint(const Common::Point &p, Polygon *polygon, Co
 		new_point.x = p1.x + u * (p2.x - p1.x);
 		new_point.y = p1.y + u * (p2.y - p1.y);
 
-		new_dist = p.sqrDist(new_point.toPoint());
+		//new_dist = p.sqrDist(new_point.toPoint());
+		int diffx = ABS(new_point.x - p.x);
+		int diffy = ABS(new_point.y - p.y);
+		new_dist = uint(diffx * diffx + diffy * diffy);
 
 		if (new_dist < dist) {
 			near_p = new_point;


m-kiewitz:
The current implementation is not based on the original SCI interpreter, but is quite different and that was done even intentionally back then (it was written for the FreeSCI project) because of patents, which were still valid back then.

The patents have expired until now and I'm currently working on reproducing Sierra's code, which should then make it work exactly like in the original interpreter and that should solve all sorts of issues.

Please no hacks.

Kawa-oneechan:
"Please no hacks"? In this case _especially_, I agree wholeheartedly!

waltervn:
This might be a regression from 88ec6f40

OmerMor:
It seems I already investigated what appears to be the exact same problem 3 years ago:
[#7140](https://bugs.scummvm.org/ticket/7140)

Thunderforge:
This PR has been open for over a year now. Should we merge it or close it? I don't see the value in keeping it open as "DO NOT MERGE" if no work is going to continue on it.

PR #1576, opened 2019-04-06
SCI: Fix LONGBOW v1.0 game breaking bugs

sluicebox:
This PR fixes the three game breaking bugs in Longbow version 1.0 that make Sierra's patch practically mandatory. This allows us to remove the warning at startup recommending Sierra's patch.

Endless Green Man riddles & Forest sweep instant-death: 

These two are symptoms of the same script bug that one patch fixes. A global variable is re-used from another scene without being initializing and its random value can break things.

Ordering a drink & other pub actions:

A text resource is missing 8 entries in the 5.25 floppy version. I added a workaround to detect the broken resource, repair the offsets, and return "** MISSING MESSAGE **" for the rest. This isn't perfect but it's an improvement from incorrect messages and a game-ending error. The missing entries are really just a one-time sequence of four texts that only occur when the player does a multi-room sequence of actions I consider unusual. I assume just typing the 8 missing strings myself is not okay. =) (but tell me if it is!)

### DIFF
--- a/engines/sci/engine/kernel.cpp
+++ b/engines/sci/engine/kernel.cpp
@@ -905,6 +905,22 @@ Common::String Kernel::lookupText(reg_t address, int index) {
 	int textlen = textres->size();
 	const char *seeker = (const char *)textres->getUnsafeDataAt(0);
 
+	if (g_sci->getGameId() == GID_LONGBOW && address.getOffset() == 1535 && textlen == 2662) {
+		// WORKAROUND: Longbow 1.0's text resource 1535 is missing 8 texts for
+		//  the pub. It appears that only the 5.25 floppy release was affected.
+		//  This was fixed by Sierra's 1.0 patch.
+		if (index >= 41) {
+			// texts 41+ exist but with incorrect offsets
+			index -= 8;
+		} else if (index >= 33) {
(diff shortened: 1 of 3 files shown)


bluegr:
Very nice work on this, well done!

I put that warning there, because Sierra already released a patch for these game-breaking bugs. Not relying on the patch is great.

Merging

angstsmurf:
Sorry, I'm just curious: What would the problem be with adding the missing messages? They are included in Sierra's official patch, right?

sluicebox:
@angstsmurf Yes, the patch includes the full resource with all the room's messages. I didn't add them because I don't know the project's stance on including game content without permission, even if it's a small amount. In this case it's just 8 dialogue strings, but still. With that many I'd want to find the right place in the codebase to put them, rather than dump a conversation in the middle of kernel.cpp, and I didn't want to do all that unless I knew it was okay.

@bluegr If that *is* okay though, I'd be happy to find the right place and add them. I don't think this will be the last Sierra bug where "just type the text they meant" gets the best results.

These were real fun ones to work on, I'm a big Longbow fan. How great is it that the save game in the bug tracker from 9 years ago still works? (It was a big help!)

PR #1536, opened 2019-03-12
SCI: Update Sound:vol in processUpdateCues

sluicebox:
Sound:vol is updated when updating cues starting in SCI_VERSION_1_MIDDLE.

Fixes bug #10244. SQ4 localized floppy versions depend on this when getting
in the orange ship, they fade music and wait for the volume to reach zero.

Confirmed against asm that this is not in SQ4 floppy (early) but is in
LSL1VGA (middle) and SQ4 localized floppies and SQ4CD (late).

PR #1578, opened 2019-04-08
SCI: Add parser support for LSL3, SQ3 German Amiga

sluicebox:
I recently got a hold of the dual German/English Amiga versions of LSL3 and SQ3. They're unusual in that they were released in 1991 and are SCI_VERSION_1_MIDDLE, making them now the latest known EGA parser games, and breaking some existing version checks.

This updates the parser/vocab version checks so that vocab gets loaded.

I've done some spot-checking of the Amiga interpreter for version 1 middle code to make sure that these versions weren't being misidentified, and they indeed have newer code. For example, they contain the Sound:vol update in UpdateCues (https://github.com/scummvm/scummvm/pull/1536), which wasn't in the first version of SQ4.

### DIFF
--- a/engines/sci/engine/kparse.cpp
+++ b/engines/sci/engine/kparse.cpp
@@ -169,7 +169,7 @@ reg_t kSetSynonyms(EngineState *s, int argc, reg_t *argv) {
 	Vocabulary *voc = g_sci->getVocabulary();
 
 	// Only SCI0-SCI1 EGA games had a parser. In newer versions, this is a stub
-	if (getSciVersion() > SCI_VERSION_1_EGA_ONLY)
+	if (!g_sci->hasParser())
 		return s->r_acc;
 
 	voc->clearSynonyms();
--- a/engines/sci/parser/vocabulary.cpp
+++ b/engines/sci/parser/vocabulary.cpp
@@ -60,7 +60,7 @@ Vocabulary::Vocabulary(ResourceManager *resMan, bool foreign) : _resMan(resMan),
 		_resourceIdBranches += 10;
 	}
 
-	if (getSciVersion() <= SCI_VERSION_1_EGA_ONLY && loadParserWords()) {
+	if (g_sci->hasParser() && loadParserWords()) {
 		loadSuffixes();
 		if (loadBranches())
 			// Now build a GNF grammar out of this
--- a/engines/sci/sci.cpp
+++ b/engines/sci/sci.cpp
@@ -311,11 +311,7 @@ Common::Error SciEngine::run() {
 	_kernel->init();
 
 	_features = new GameFeatures(segMan, _kernel);
-	// Only SCI0, SCI01 and SCI1 EGA games used a parser
-	_vocabulary = (getSciVersion() <= SCI_VERSION_1_EGA_ONLY) ? new Vocabulary(_resMan, false) : NULL;
-	// Also, XMAS1990 apparently had a parser too. Refer to https://forums.scummvm.org/viewtopic.php?t=9135
-	if (getGameId() == GID_CHRISTMAS1990)
-		_vocabulary = new Vocabulary(_resMan, false);
+	_vocabulary = hasParser() ? new Vocabulary(_resMan, false) : NULL;
 
 	_gamestate = new EngineState(segMan);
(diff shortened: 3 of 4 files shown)


m-kiewitz:
Are those really SCI1 MIDDLE?
Or do they get detected incorrectly?

sluicebox:
That was my question too. Like I said, I found code in the interpreter I know to appear in version 1 middle that's not present in version 1 early (SQ4 floppy), so it looks like they used the latest code for these 1991 releases.

As far as the various resource and compression formats, they all match LSL1VGA DOS aside from EGA vs VGA views. I believe the individual feature detection to be correct, having checked most against the resource files, and that the game runs. (Sounds aren't playing quite right, but one thing at a time.) SCI Companion gets the same results (I know its code is based on ScummVM) and successfully reads all the resources.

```
Detected features:
------------------
Sound type: Late SCI1
Graphics functions type: Late SCI0
Lofs type: Middle SCI1
Move count type: increment
SetCursor type: Early SCI0
PseudoMouse ability: yes
View type: EGA
kAnimate fastCast enabled: no
Uses palette merging: yes
Uses 16 bit color matching: yes
Resource volume version: SCI0 / Early SCI1
Resource map version: Middle SCI1
Contains selector vocabulary (vocab.997): yes
Has CantBeHere selector: yes
Game version (VERSION file): 1.000
```

bluegr:
Well done!

It seems we could include the xmas1990 demo in these checks (right now, there’s only one check for it). Could you test that one as well?

sluicebox:
Good idea, I'll check out that demo and see if that should be factored into the check.

My sneaky goal in this PR is raise this update to the SCI version timeline with you experts and make sure it sounds correct. I wouldn't be surprised if playtesting turns up some more version checks that need to be amended, so pipe up if this sounds like  a misdiagnosis of the situation. So far though, I've been able to run around the first SQ3 area without problems.

I'm so late to the party, I have to dig up the exotic versions like these to have fun with =)

sluicebox:
Well, I looked into it, and not to sidetrack the Amiga issue but...

If I were involved back when this was brought up I would have argued against the Xmas 1990 parser code being added. In fact, I'm arguing for that now.

The Xmas 1990 demo does not have a parser. You can verify this by typing and no prompt appearing. No Prompt == No Parser. (I believe computer science textbooks call this the NP vs NP problem)

This code was added to support someone's hand-altered copy (they added extraneous files!) that enables dead code: an input box with a prompt, a single generic response, and zero code that responds to it. I mean... I guess that makes me feel better about my life choices? If you want to do that to a noninteractive demo that you've patched, more power to you, but you can also go ahead and patch your own ScummVM source while you're at it. Or just run "strings" and move on with your life. Because right now typing an integer into the prompt crashes ScummVM with an assertion. Unacceptable, rip it out.

I don't care that a Sierra developer typed two strings that their compiler failed to exclude. This is dead uninteresting code that ScummVM has no business going out of its way to enable in a hacked version. Even by Sierra fan minutia, of which I proudly trade, this doesn't cut it. Not only should we not infect hasParser() with this, as it seeks to clarify the situation, we should remove it. (I know of so much more interesting dead code in real games we wouldn't dream of activating)

I just spent a bit of time learning all this. That code has a cognitive cost I don't want anyone else to pay. 

#opinions! #merryChristmas! #mexicanBeerWasOnSale?

OmerMor:
I'd argue that the xmas 1990 demo does not have a parser, but the interpreter included with that demo does indeed has parser support.
IIRC, this is the only VGA SCI interpreter with parser support, so it enables us to build fan games in VGA with a parser.
Here's an initial game template for that: https://github.com/OmerMor/SCI01-VGA-Template
This has nothing to do with ScummVM BTW.

bluegr:
Understandable, an exception for that demo shouldn’t have been added in the first place. Could you please remove that xmas1990 check? I added the check, but now that I re-read the relevant forum post, and your answer and I’ve replayed it, it was a bad decision

sluicebox:
I'll axe it. The underlying reasoning totally makes sense, and @OmerMor, I'm a huge fan of your work and am thrilled that you've joined the thread. When you solved that Colonel's Bequest easter egg it was an inspiration. Also I was insanely jealous. Ping me on IRC some time, homey =)

bluegr:
Thanks for your work :) Merging

PR #1579, opened 2019-04-08
Sword25: image trace on quick cursor changes fix

Fantohin:
Fix: https://bugs.scummvm.org/ticket/7061
Additional check in RenderObjectQueue::exists() that 
checks if a RenderObject should be 
redrawn. 
After the change RenderObjectManager::render() 
was tested to update 0 objects in the first active scene,
when player does nothing including not moving cursor 
& 2 objects when player moves cursor.
It seems glitch happens when  cursor image
changes from image X to image Y to image X in a short 
time frame leaving image of state Y not re-rendered.

### DIFF
--- a/engines/sword25/gfx/renderobjectmanager.cpp
+++ b/engines/sword25/gfx/renderobjectmanager.cpp
@@ -52,7 +52,8 @@ void RenderObjectQueue::add(RenderObject *renderObject) {
 bool RenderObjectQueue::exists(const RenderObjectQueueItem &renderObjectQueueItem) {
 	for (RenderObjectQueue::iterator it = begin(); it != end(); ++it)
 		if ((*it)._renderObject == renderObjectQueueItem._renderObject &&
-			(*it)._version == renderObjectQueueItem._version)
+			(*it)._version == renderObjectQueueItem._version && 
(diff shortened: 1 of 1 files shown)


Fantohin:
-------------------------------------------------------------------------------------------------------
Tested on:
ScummVM 2.1.0git5089-g59c0383c6f (Apr 7 2019 14:55:14)
Features compiled in: Vorbis FLAC MP3 ALSA SEQ TiMidity RGB zLib MPEG2 FluidSynth Theora AAC A/52 FreeType2 JPEG PNG cloud (servers, local)

DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=18.0.3
DISTRIB_CODENAME=Illyria
DISTRIB_DESCRIPTION="Manjaro Linux"

bluegr:
Thanks! The dirty rects code was added when the engine was merged into ScummVM, so it’s good that you found the bug there, and tested the bugfix :)

Please, fix spacing: the rest of the checks are split, one per line, so you should do the same with the new check, so that this pull request can be merged

Fantohin:
Done. Happy to be of any help.

bluegr:
Thanks for your work! Merging

PR #1580, opened 2019-04-08
COMMON: fixed reading ini files with section containing more than one…

whiterandrek:
Current implementation can't read section name that contains more than one word(
e.g
[Part 0]
[Part 2 Chapter 1]

> [The beginning each of section is denoted by a line of the following form: "[section type]" where section type is a series of words](http://www.nongnu.org/chmspec/latest/INI.html)


bluegr:
Nice catch! Thanks for your work.

Merging

PR #1581, opened 2019-04-09
WINTERMUTE: Fixed normalizeAngle's output range from 0-360 to 0-359.

kyranet:
```
WINTERMUTE: Fixed normalizeAngle's output range from 0-360 to 0-359.

When normalizing an angle, we expect the number to be between 0 and
359 (since 360 is 0), this changes the util so 360 is transformed to 0.

Technically this is change makes the range [0, 360) instead of [0, 360], floating
numbers (such as 359.8) would not be returned as negative thanks to the
following `while` loop.
```

### DIFF
--- a/engines/wintermute/utils/utils.cpp
+++ b/engines/wintermute/utils/utils.cpp
@@ -44,7 +44,7 @@ void BaseUtils::swap(int *a, int *b) {
 
 //////////////////////////////////////////////////////////////////////////
 float BaseUtils::normalizeAngle(float angle) {
-	while (angle > 360) {
+	while (angle > 359) {
 		angle -= 360;
 	}
 	while (angle < 0) {


bluegr:
Nice catch! An off-by-one error.

Thanks for spotting this

Could you please change the commit description, so that it matches our style and guidelines? Since this is a change for the Wintermute engine, the commit message should start with WINTERMUTE:

engines/wintermute/utils/utils.cpp
```diff
@@ -44,7 +44,7 @@ void BaseUtils::swap(int *a, int *b) {
 
 //////////////////////////////////////////////////////////////////////////
 float BaseUtils::normalizeAngle(float angle) {
-	while (angle > 360) {
+	while (angle > 359) {
```
criezy: That would work if the angle was a `int`, but since this is a `float` this is incorrect. For example now we would get -0.2 instead of 359.8.

kyranet: The following loop covers it:
https://github.com/scummvm/scummvm/blob/8469b8749c9e710992d22bdfa07e6afcff9cfd4d/engines/wintermute/utils/utils.cpp#L50-L52

kyranet:
I changed the PR's title @bluegr, want me to also amend the commit's message and force-push?

bluegr:
@kyranet: yes please

engines/wintermute/utils/utils.cpp
criezy: Right. That will teach me to look beyond the displayed diff! :-P
I was expecting to see `>= 360` and didn’t look further.

criezy:
So now my question (to a Wintermute engine developer I guess) would be: since the original engine has the same bug (see https://bitbucket.org/MnemonicWME/wmelite/src/fe84bd13b78b06ba7ecd9b4739eb622d95ebf92b/src/utils.cpp?at=default&fileviewer=file-view-default#utils.cpp-89), is it better to keep the same behaviour as the original engine? Is there for example a risk that some games are relying on the "buggy" behaviour and would break with this change?

tobiatesan:
@criezy I would always err on the side of duplicating the original behaviour.
That's what we've historically done, IIRC.
Things get more interesting if mnemonic has fixed it at some point.
Either way - be it replicating the original or not - I would however always output a warning/info.

bluegr:
I concur with @tobiatesan. This is obviously a bug in the original, which hasn't been fixed, but we need to check if there's any test case for it.

Thus, throwing a warning, and fixing the resulting angle like we do here sounds like the best course of action. If we do find any cases where this behavior occurs, we can check what the expected result is.

tobiatesan:
Any news on this one?

tobiatesan:
By the way, I guess this is tangentially related to https://github.com/scummvm/scummvm/pull/1319#discussion_r214525094 - i.e. the general problem of handling bugfeatures.

bluegr:
This has been merged in tree, together with a commit that adds a warning for such cases. Thanks for your work! Merging

PR #1583, opened 2019-04-11
 COMMON: Use a prefix table to speed up the Huffman decoder

bgK:
A high resolution mod for The Longest Journey is in the works for ResidualVM. It appeared recently that the Bink decoder was not fast enough to decode one frame of video within the frame budget (16.6 ms) when using high resolution videos. This PR implements some optimizations to speed up decoding:

* Fix BitStream's peekBits not to seek the underlying stream. Seeking can be slow when the stream is a file.
* Change BitStream's multi-bit operations to work on multiple bits at once rather than iterating over single-bit operations.
* Use a prefix lookup table to speed up decoding of short Huffman codes.

With these changes decoding one frame of a 1920x576 Bink video goes from 20 ms to 10 ms on my computer. This is still quite a long time when compared to the frame budget, but is good enough for now.

I'd like to thank DrMcCoy again for providing patches of a similar optimization he worked on for xoreos.


bluegr:
Very nice work! Many thanks bgK and DrMcCoy :)

Two questions:
- Have these changes been tested with the ScummVM games that use this decoder? If I’m not mistaken, these include Myst/Riven (via QuickTime) and some HE games
- Why has the test_get_after_set_symbols() test been removed? Is it obsolete, after these changes?

bgK:
Thanks for your review.

I have tested this with:
- The Bink decoder (In The Longest Journey and Myst III)
- The WMA decoder (Myst III for XBox)
- The SVQ1 decoder (Myst ME)
- The Indeo 4/5 decoder (Starship Titanic)

`test_get_after_set_symbols` was removed because its goal was to test the `Huffman::setSymbol` that was unused and that I have removed (The implementation becomes non-trivial with the optimization).

bluegr:
Excellent! Nice work guys :)

Since all the codecs that are using bitstreams and huffman codes have been tested, and since this code is also used in two other projects, it’s fine to merge.

Merging

PR #1584, opened 2019-04-15
SCI: Fix diskdump patch file headers

sluicebox:
This fixes a bug in the 'diskdump' debugging command that creates an additional patch file header when dumping resources that were loaded from a patch file instead of a volume. This would corrupt the dumped patch file.

The code would always create and write a new header, even if it was about to write an existing one that was read in from a patch file. Now it's either one or the other.

### DIFF
--- a/engines/sci/resource.cpp
+++ b/engines/sci/resource.cpp
@@ -224,10 +224,14 @@ void Resource::unalloc() {
 }
 
 void Resource::writeToStream(Common::WriteStream *stream) const {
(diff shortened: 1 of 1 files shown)


bluegr:
Nice catch! I missed that bit when implementing the diskdump command

Merging

sluicebox:
You're my hero, I love diskdump! I've been able to make some fun things thanks to that dump-all mode.

PR #1586, opened 2019-04-16
MSVC: Add VS2019 and cleanup versions

SupSuper:
Now that Visual Studio 2019 is out, I decided to take the chance to clean up the Visual Studio project generation, since it's led to inconsistencies between versions and kept the folders growing every new release.

Instead of having a msvc folder for every version, I combined all the create_project scripts and left the version differences up to the create_project generation. Visual Studio is backwards-compatible so I doubt anyone needs multiple parallel versions of project files.

The supported versions are the same (VS2008 and later, with the addition of 2019), I only cleaned up the folder structure and code. I also added version auto-detection when none is specified, to make the new developer experience smoother.

Let me know what you think, and I'd appreciate testing on old Visual Studio versions to make sure I didn't break anything.


devtools/create_project/codeblocks/create_project.cbp
```diff
@@ -36,6 +36,8 @@
 			<Add option="-Wall" />
 			<Add option="-fexceptions" />
 		</Compiler>
+		<Unit filename="..\cmake.cpp" />
+		<Unit filename="..\cmake.h" />
```
Mataniko: How is CMake being used here? Why did codeblocks need to be updated?

devtools/create_project/create_project.cpp
```diff
@@ -1082,6 +1095,17 @@ const Tool s_tools[] = {
...
+
+const MSVCVersion s_msvc[] = {
+//    Ver    Name                     Solution                     Project    Toolset    LLVM
+	{  9,    "Visual Studio 2008",    "10.00",          "2008",     "4.0",     "v90",    "LLVM-vs2008" },
+	{ 10,    "Visual Studio 2010",    "11.00",          "2010",     "4.0",    "v100",    "LLVM-vs2010" },
+	{ 11,    "Visual Studio 2012",    "11.00",          "2012",     "4.0",    "v110",    "LLVM-vs2012" },
+	{ 12,    "Visual Studio 2013",    "12.00",          "2013",    "12.0",    "v120",    "LLVM-vs2013" },
+	{ 14,    "Visual Studio 2015",    "12.00",            "14",    "14.0",    "v140",    "LLVM-vs2014" },
+	{ 15,    "Visual Studio 2017",    "12.00",            "15",    "15.0",    "v141",    "llvm"        },
+	{ 16,    "Visual Studio 2019",    "12.00",    "Version 16",    "16.0",    "v142",    "llvm"        }
```
Mataniko: Just double checking, they moved from xx to Version xx?

devtools/create_project/create_project.cpp
```diff
@@ -1147,6 +1171,63 @@ ToolList getAllTools() {
...
+		if (s_msvc[i].version == version)
+			return &s_msvc[i];
+	}
+
+	return NULL;
+}
+
+int getInstalledMSVC() {
+	int latest = 0;
+#if defined(_WIN32) || defined(WIN32)
```
Mataniko: Is this required?

devtools/create_project/msvc10/create_project.vcxproj
```diff
@@ -59,10 +59,8 @@
     </Link>
     <PostBuildEvent>
       <Command>@echo off
-xcopy /Y "$(TargetPath)" "$(SolutionDir)\..\..\..\dists\msvc10\"
-xcopy /Y "$(TargetPath)" "$(SolutionDir)\..\..\..\dists\msvc9\"
-xcopy /Y "$(TargetPath)" "$(SolutionDir)\..\..\..\dists\codeblocks\"
-xcopy /Y "$(TargetPath)" "$(SolutionDir)\..\..\..\dists\iphone\"</Command>
```
Mataniko: Why are we copying codeblocks and iphone to begin with?

devtools/create_project/msvc10/create_project.vcxproj.filters
```diff
@@ -53,6 +56,9 @@
     <ClCompile Include="..\xcode.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="..\cmake.cpp">
```
Mataniko: Same as above, where is cmake being used?

sluicebox:
This is some heroic work. Great job!

I tested 2019 and 2015 with no problems. In 2010 I could generate the project and it looks right. It doesn't compile for me, but that was the case before this PR, and almost certainly because I don't have 2010 prebuilt libraries setup.

The 2008 create_project project file changes are missing closing quotes on VCPostBuildEventTool CommandLine and the project can't load. Adding those fixes that, but that post build command still isn't working. It's copying create_project.exe to codeblocks but not msvc. I suspect the multi-line xml-embedded batch file acrobatics needs a tweak. As with 2010, I can't test a full build in 2008, but the generated project loads and looks right.

If I understand this, when building create_project with anything other than 2008, devtools\create_project\msvc10 should be used. That took me a few moments to reason out. Since it's now just that directory and msvc9, presumably because that one is too different, how about the "msvc10" directory just be renamed "msvc"? The 2010 reference will be more and more of a forgotten historical artifact as time passes and this would be it clear that 2008 is the exception.

devtools/create_project/msvc9/create_project.vcproj
```diff
@@ -82,6 +82,9 @@
 			/>
 			<Tool
 				Name="VCPostBuildEventTool"
+				CommandLine="@echo off
+xcopy /Y &quot;$(TargetPath)&quot; &quot;$(SolutionDir)\..\..\..\dists\msvc\&quot;
+xcopy /Y &quot;$(TargetPath)&quot; &quot;$(SolutionDir)\..\..\..\dists\codeblocks\&quot;
```
sluicebox: Missing terminating quote, and VS2008 will run this all as one line. This monstrosity (generated by VS2008) works:

`CommandLine="@echo off&#x0D;&#x0A;xcopy /Y &quot;$(TargetPath)&quot; &quot;$(SolutionDir)\..\..\..\dists\msvc\&quot;&#x0D;&#x0A;xcopy /Y &quot;$(TargetPath)&quot; &quot;$(SolutionDir)\..\..\..\dists\codeblocks\&quot;"`

devtools/create_project/msvc10/create_project.vcxproj.filters
SupSuper: create_project has a CMakeProvider generator (since 2016 according to Git). Even if it's not used, it needs to be included for the project to compile.

devtools/create_project/msvc10/create_project.vcxproj
SupSuper: Good question, it was already like that before I got here. :) Theoretically, one could compile create_project in MSVC and later decide to generate a CodeBlocks project with it, but it seems unlikely.
Maybe it would make more sense for the MSVC project to only copy to the MSVC target?

devtools/create_project/create_project.cpp
SupSuper: create_project can be compiled on non-Windows platforms (eg. to generate Xcode projects).

devtools/create_project/codeblocks/create_project.cbp
SupSuper: I simply made it consistent with the MSVC project files, but maybe it would be better in a separate PR.

devtools/create_project/create_project.cpp
SupSuper: Yup. For reference, here's the whole thing:
```
Microsoft Visual Studio Solution File, Format Version 12.00
# Visual Studio Version 16
```
I don't think this information is actually *necessary*, but why fix what isn't broken.

bluegr:
I too have checked this out. Works great, nice work!

I don't have anything else to add, apart from the comments made by @Mataniko and @sluicebox above. IMHO, we shouldn't perform major changes to the tool, just to support ancient VS versions, like 2008.

PR #1590, opened 2019-04-19
SCI: Update QFG1VGA speed test patch for Mac

sluicebox:
Updates the speed test patch to work with the newer and significantly
different Mac version


bluegr:
Nice work, thanks! Merging

PR #1591, opened 2019-04-21
IMAGE: Avoid pixel format conversions in the JPEG and PNG decoders

bgK:
In ResidualVM:
* Myst III decodes large amounts of JPEG images (each node is made of at least 6 640x640 images).
* The WIP HD mod for The Longest Journey has its backgrounds stored as 1080p PNG images.

The issue I've been having with the current JPEG and PNG decoders is that they perform color conversion to a somewhat arbitrary pixel format.
* That color conversion is rather slow when dealing with large images.
* The output pixel format being arbitrary, the calling code is required to perform a second pass of color conversion. Ideally images should be decoded directly to the screen format so they can be blitted to the screen directly.
* The output pixel format has a different byte order on big endian and little endian systems, making it hard for the caller to do byte level pixel manipulation (fast) and forcing it to go through Graphics::PixelFormat (slow).

With those constraints, it's hard to keep image load times low.

In this PR, I propose the following changes:
* Add a new `setOutputPixelFormat` method to the JPEG decoder. This method allows to take advantage of the very optimized color conversion routines in `libjpeg-turbo` to decode the JPEG images directly to the caller selected pixel format.
* Change the PNG decoder to always produce RGBA byte order output surfaces. `libpng` is not as flexible as `libjpeg-turbo` when it comes to selecting the output pixel format; and the color conversion routines in `libpng` are not very fast. The caller is responsible for converting to a suitable pixel format. But at least the decoder is no longer performing slow color conversion and the output pixels can be easily transformed without going through `Graphics::PixelFormat::colorToARGB`.

### DIFF
--- a/engines/glk/picture.cpp
+++ b/engines/glk/picture.cpp
@@ -125,6 +125,7 @@ Picture *Pictures::load(uint32 id) {
 		palette = png.getPalette();
 		palCount = png.getPaletteColorCount();
 	} else if (f.open(Common::String::format("pic%u.jpg", id))) {
+		jpg.setOutputPixelFormat(g_system->getScreenFormat());
 		jpg.loadStream(f);
 		img = jpg.getSurface();
 	} else if (f.open(Common::String::format("pic%u.raw", id))) {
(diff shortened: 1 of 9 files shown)


bluegr:
Great work! +1 from me

If I’m not mistaken, the libjpeg-turbo extensions are handled with the JCS_EXTENSIONS and JCS_ALPHA_EXTENSIONS checks. Thus, this code should also work with the regular libjpeg library, right?

Since 6 days have already passed, I propose to merge this tomorrow, if nobody objects

bgK:
Thanks for taking a look. The `libjpeg-turbo` checks are indeed handled with those defines.

Merging now.

PR #1592, opened 2019-04-21
ENGINES: Allow overriding the splash screen

lotharsm:
This patch allows overwriting the splash screen that appears when a game is started directly without using the Launcher with a custom .png file.

The original idea and patch was proposed by Karl Johnson from Cyan, Inc. during the development of their 25th Anniversary Collection of the Myst series.

This change provides publishers that bundle their game with ScummVM a less "intrusive" way of showing their game is powered by ScummVM.

I know that we currently propose not to alter the splash screen at all. However, if a, well, "less" friendly publisher wants, he still can remove the splash screen completely. This new method provide a good solution in my opinion. I added a comment to the source code that first recommends the user not to change the splash screen and only recommend the new methods afterwards.

To protect our brand, a custom splash screen will include a hardcoded "powered by ScummVM $version" at the bottom of the screen. Additionally, it's recommended that publishers include our logo into their splash screen.

Example:
splashdemo

Currently, the new feature is disabled by default. You have to run the configure script with --enable-splash-override in order to enable it.

I know that this change might be controversal and I fully understand if the team decides to dismiss this PR. However, I hope you can follow the points I mentioned above.

### DIFF
--- a/configure
+++ b/configure
@@ -182,6 +182,7 @@ _no_pragma_pack=no
 _bink=yes
 _cloud=auto
 _pandoc=no
+_splashoverride=no
 # Default vkeybd/keymapper/eventrec options
 _vkeybd=no
 _keymapper=no
@@ -1019,6 +1020,8 @@ Optional Features:
   --enable-text-console    use text console instead of graphical console
   --enable-verbose-build   enable regular echoing of commands during build
                            process
+  --enable-splash-override allow overriding the default splash screen with
+                           a custom image (requires PNG support)
   --disable-bink           don't build with Bink video support
   --opengl-mode=MODE       OpenGL (ES) mode to use for OpenGL output [auto]
                            available modes: auto for autodetection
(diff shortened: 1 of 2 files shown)


Kroc:
If a set of splashes were provided with ScummVM, then a function to create a desktop shortcut from any game would be able to make use these.

lotharsm:
@kroc Sounds like a good idea to tackle next. However, I think that at the current point in time it’s a little bit out of scope regarding this PR. :)

bluegr:
I don’t really like this idea. There’s a reason for the splash screen, to properly attribute the work that has been done in ScummVM.
Allowing a custom splash screen defeats the purpose...

Kroc:
This is one of those "it's going to happen anyway" problems. If ScummVM provides its own mechanism then it's 99% more likely that companies will use that instead of ripping out the splash manually.

lotharsm:
Thanks @blugr and @Kroc for your comments! What @Kroc said is exactly my point of view - *if* a publisher now decideds that our splash screen should be removed, he *will* simply rip out the code. After this change, we at least have the chance that our copyright notice remains intact.

Please also note that splash screen overriding is disabled by default, even in the release builds and that the comment in the code recommends leaving the splash screen intact without altering it at all.

sev-:
And this is precisely what I want: to see if somebody wants to blatantly abuse us. Then we will know their bad intentions and real attitude towards us.

We allowed Cyan to do it since they were very supportive and collaborative from the very beginning.

And yes, I _want_ the splash screen to be intrusive, this is the whole purpose of it. Historically a number of vendors repeatedly "forgot" to credit ScummVM Team for their hard work and follow GPL.

PR #1593, opened 2019-04-21
COMMON: Fix compiler warning in usrt.cpp

jepael:
Compiling emitted this warning:

`warning: comparison between signed and unsigned integer expressions`

Fixed by type casting the char at string index to match the internal type.


jepael:
How do I get rid of the extra commit where I updated my fork?

PR #1594, opened 2019-04-21
COMMON: Fix compiler warning

jepael:
Compiling emits the following warning:
`comparison between signed and unsigned integer expressions`

Cast the string character type to native type for comparison to fix it.


criezy:
Thank you for your contribution.
It makes sense and this is a simple change, so I don't foresee any issue with it. Merging as is.

PR #1596, opened 2019-04-22
CREATE_PROJECT: Add Text console debugger feature

sluicebox:
This adds a feature flag to create_project so that --enable-text-console sets USE_TEXT_CONSOLE_FOR_DEBUGGER.

The configure script has this, and create_project always sets it for xcode, and now it can be set for others like Visual Studio.


Mataniko:
Makes sense. Merging as is.

PR #1597, opened 2019-04-23
SCI: Fix QFG1VGA inventory weight warn, bug #10942

sluicebox:
Fix a script bug that breaks the game when picking mushrooms or other
actions if they trigger an inventory weight warning


bluegr:
Nice work, thanks!

Merging

PR #1598, opened 2019-04-23
SCI: Fix QFG1VGA Mac great hall, bug #10945

sluicebox:
Fixes an uninitialized temp read in Mac version. This requires a small script patch as opposed to a workaround because the value being read is dynamic.


bluegr:
Great, thanks!

PR #1599, opened 2019-04-26
GRAPHICS: Fix ManagedSurface dirty rects when drawing lines

SupSuper:
Fixes an invalidRect assert when drawing lines which don't go from top-left to bottom-right.


bluegr:
Nice and clean way to manage this. Thanks for your work!

PR #1600, opened 2019-04-27
SCI: Fix QFG1VGA Mac giant fight, bug #10948

sluicebox:
Another uninitialized temp read introduced in the Mac version. Script patch because the value is dynamic.


bluegr:
Thanks! Merging

PR #1602, opened 2019-04-28
GUI: Use a truetype font for the debug console

bgK:
The current debug console font is IMO a bit too small; it is also hardcoded.

This PR changes the following:
* The debug console font is now customizable via the GUI theme; making it possible to use TTF fonts.
* Change the Modern and Remastered themes to use the Source Code Pro font for the debug console. The size of the font depends on the height of the window (3 steps: < 400px, < 800px and above).

I'm not a lawyer but believe it's ok for use to embed that font according to the SIL Open Font License. Otherwise, I'm open to suggestions for monospaced fonts that look good in ScummVM.

This is how it looks:
Screenshot from 2019-04-28 08-57-09


bluegr:
Looks good! Well done

I’m not a lawyer either, but clause 2 of the license is pretty straightforward:

    Original or Modified Versions of the Font Software may be bundled,
    redistributed and/or sold with any software, provided that each copy
    contains the above copyright notice and this license.

sluicebox:
Gorgeous. I'm switching back from USE_TEXT_CONSOLE_FOR_DEBUGGER!

bluegr:
We should leave this open for a week, in case there are any other opinions on it. 

Unless someone disagrees, I propose to merge this next weekend, on Sunday the 5th of May

PR #1604, opened 2019-04-28
SCI: Change sendMidiCommand non-midi error to warn

sluicebox:
This changes an error() to warning() when SciMusic::cmdSendMidi is called against a non-midi sound as this occurs in FPFP.

FPFP sets its gameMusic2 to an audio sample in at least one scene, and other scenes call gameMusic2:mute which calls kDoSound SendMidi. This occurs in all FPFP versions. Fixes bug #10952

This was promoted from warning() to error() in: https://github.com/scummvm/scummvm/commit/db70d66e4ae872ebdeb2a88ef0870708a30c7667

### DIFF
--- a/engines/sci/sound/music.cpp
+++ b/engines/sci/sound/music.cpp
@@ -800,8 +800,15 @@ void SciMusic::sendMidiCommand(uint32 cmd) {
 
 void SciMusic::sendMidiCommand(MusicEntry *pSnd, uint32 cmd) {
 	Common::StackLock lock(_mutex);
-	if (!pSnd->pMidiParser)
-		error("tried to cmdSendMidi on non midi slot (%04x:%04x)", PRINT_REG(pSnd->soundObj));
+	if (!pSnd->pMidiParser) {
+		// FPFP calls kDoSound SendMidi to mute and unmute its gameMusic2 sound
+		//  object but some scenes set this to an audio sample. In Act 2, room
+		//  660 sets this to audio of restaurant customers talking. Walking up
+		//  the hotel stairs from room 500 to 235 calls gameMusic2:mute and
+		//  triggers this if gameMusic2 hasn't changed. Bug #10952
+		warning("tried to cmdSendMidi on non midi slot (%04x:%04x)", PRINT_REG(pSnd->soundObj));
+		return;
+	}
 
 	pSnd->pMidiParser->mainThreadBegin();
 	pSnd->pMidiParser->sendFromScriptToDriver(cmd);


bluegr:
I’m a bit undecided on this. Sending MIDI commands to a sound slot where no MIDI is playing is obviously wrong. So, the options we got here are:

- Ignore the buggy script behavior, like you did here
- Patch the problematic scripts
- Stop sound slots with digital samples, when MIDI commands are sent to them

I’m more inclined towards the second option: patch the offending scripts, since this erroneous behavior only occurs in Pharkas

m-kiewitz:
When exactly is it called?
When using the original interpreter, what happens when the user has no sound card capable of playing samples? Is the Midi call possibly for these situations?

sluicebox:
I don't have a strong opinion on which way to go, but this doesn't crash the DOS interpreter, at least, not with audio settings set to soundblaster. I guess the question is what are we trying to accomplish with this scenario being an error vs warning, now that the error has revealed that this is a thing that happens, and a warning still logs it.

@m-kiewitz Walking up the hotel stairs after entering the restaurant in act 2 triggers it, there might be more. Reproduction steps and a save game are at https://bugs.scummvm.org/ticket/10952

To go into more detail: gameMusic2 is a global Sound object. Sound:mute, which can mute or unmute, is implemented as a call to kDoSound SendMidi. Sound:mute was implemented differently in earlier games, but ones around this time and later (SQ5, GK1) use SendMidi. Room 600 (and possibly others) play a sample instead of midi with gameMusic2 and stop playing when you leave. Room 235 (when entering and exiting) and room 400 (when displaying and removing the inBunting inset) call gameMusic2:mute. gameMusic2 may or may not be playing at that time. In this case, it's not playing, but the resource it's still associated with is a sample.

A script patch means either patching two rooms or patching room 600 and doing a thorough audit of all other scripts that set gameMusic2 to see if that number is for an audio sample, and then patching them. Either way it seems like a lot and I'm not sure what the patch would look like. Sound:mute might be patchable but at that point I'd want that applied to all games with this Sound:mute. I can imagine another game muting a global sound still associated with a non-playing sample.

I was on the fence too so it's not a coincidence the PR I submitted to get the ball rolling was the easy one. =)

sluicebox:
I keep saying room 600 but I meant 660, updating the comment...

m-kiewitz:
In some Sierra games there were Midi tracks implemented as sound effects, which were used when the computer did not have a sound card capable of playing samples.

Did Freddy Pharkas also do this?
If yes, then the mute call may have been supposed to mute such a Midi sound effect and thus wouldn't even be a bug.

What does the original interpreter do in that situation, when there is no soundblaster card set up?

sluicebox:
The restaurant crowd sound number is 2662, there's an audio and sound resource. The audio is the sample of people talking, the sound is a silent midi. Is that midi just a placeholder so that something gets played even when there's no hardware that can play samples?

I don't know (much) about what the original interpreter is doing under the hood in these scenarios at this point, just that it's not crashing.

In this scenario, the sound already stopped playing several screens ago, but you're right, there may be another sample/midi combo later in the game they're expecting to be playing in gameMusic2. A lot happens in room 235 and I'm not to the end yet. (Got derailed by a much more complicated crash to untangle...)

sluicebox:
Ah, finished the game, first time since the 90s when I bought it during Sierra's 5.25 floppy clearance sale.

I'm leaning more towards the warning now that I've seen some stuff.

I can now trip this in several completely different ways. Lots of scripts play an audio sample through gameMusic2, not just the restaurant, so it's common throughout the game for it to point to a non-midi sound that played in some earlier room.

The second action that crashes is looking at the hotel balcony from the gazebo in room 400, which like the stairs, can be done at almost any point in the game.

I don't know what a patch would look like for this but I suspect it would be disproportionately complex to the goal of avoiding a line that voluntarily ends the game. I'm sure it can be done but it's not clear to me what the buggy script is.

The original Mac interpreter also doesn't crash on this. (and it *loves* to crash)

I think the error has done its job and as a warning it would still be helpful. Similar engine warnings have helped me greatly in fixing tangentially related bugs just by being in the logs once things really go off the rails.

bluegr:
Fair enough. Thanks for your work and for playtesting this change :)

Merging

PR #1607, opened 2019-04-30
CONFIGURE: Fix incorrect help text

raziel-:
On using the path where libcurl is installed (assuming that the old help text is referring to the actual library, be it static or shared) this will result in libcurl not being found by configure.
Instead, the path should point to where libcurl-config is installed, so it may get picked up and processed properly
(See also --with-freetype-prefix and --with-sdl-prefix)
e.g.
On my SDK installaliton libcurl.a and libcurl.so are placed in /sdk/local/newlib/lib, while libcurl-config is in /sdk/local/newlib/bin.
Using the latter makes configure properly picking up and working with libcurl. The other don´t.

```

two together.

- Bullet points are okay, too

- Use a hanging indent
```

### DIFF
--- a/configure
+++ b/configure
@@ -1110,7 +1110,7 @@ Optional Libraries:
   --with-sdlnet-prefix=DIR prefix where SDL_Net is installed (optional)
   --disable-sdlnet         disable SDL_Net networking library [autodetect]
 
-  --with-libcurl-prefix=DIR    prefix where libcurl is installed (optional)
+  --with-libcurl-prefix=DIR    prefix where libcurl-config is installed (optional)
   --disable-libcurl        disable libcurl networking library [autodetect]
 
 Some influential environment variables:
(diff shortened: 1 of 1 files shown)


lotharsm:
One thing I missed on the last PR:

Please alter the commit message according to our guidelines:

CONFIGURE: Fix incorrect help text

configure
```diff
@@ -4876,7 +4876,7 @@ echo "$_libunity"
 #
 find_freetype() {
         # Wrapper function which tries to find freetype
-        # either by calling freetype-config or by using
+        # either by callimg freetype-config or by using
```
criezy: Why that change? Wasn’t `calling` correct?

raziel-: Grr, of course it was correct.
I thought i was adding this change to my other patch while it was being merged and ended up creating a new PR but with the old data.
Sorry

criezy:
Usually `prefix` in this context means one level up, so in your case `/sdk/local/newlib`. Have you tried to use that? It should work.

So I think the old text is actually correct, as it is not asking for the lib directory, but for this `prefix.` And I would tend to reject your change.

raziel-:
Heh, i'm so noob that it hurts.

Yes, that works well too.

Sorry for the noise :-/

PR #1612, opened 2019-05-01
agos: mute fix for adlib accolade

Fantohin:
Work in progress on https://bugs.scummvm.org/ticket/10310
This implements _masterVolume and setVolume in MidiDriver_Accolade_AdLib and uses a cast to call setVolume() in agos/midi.cpp in pause() if AdLib & Accolade are used.
The other drivers mostly seem to not implement channel allocation so I decided to follow TO_DO comment in the MidiDriver_Accolade_AdLib code.
Not in the PR: 
Implementing the same _masterVolume and setVolume in MidiDriver_Simon1_AdLib  and calling that in pause() seems to fix the same defect for Simon 1.
For mt32 using _driver to send control message for volume for all channels seems to fix this for mt32.
Making PR in hopes of getting some input on how to proceed.

### DIFF
--- a/engines/agos/midi.cpp
+++ b/engines/agos/midi.cpp
@@ -28,6 +28,8 @@
 #include "agos/agos.h"
 #include "agos/midi.h"
 
+#include "agos/drivers/accolade/adlib.cpp"
+
 #include "agos/drivers/accolade/mididriver.h"
 #include "agos/drivers/simon1/adlib.h"
 // Miles Audio for Simon 2
@@ -89,7 +91,7 @@ int MidiPlayer::open(int gameType, bool isDemo) {
 	assert(!_driver);
 
 	Common::String accoladeDriverFilename;
-	MusicType musicType = MT_INVALID;
+	musicType = MT_INVALID;
 
 	switch (gameType) {
 	case GType_ELVIRA1:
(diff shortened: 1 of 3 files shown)


digitall:
@Fantohin : Thanks for your investigation and code patch for that bug. I will update the artifact to cross link to this. It may take some time before a developer with knowledge of the AGOS engine has time to review this so please standby. One minor point in the meantime, if you refer to https://wiki.scummvm.org/index.php/Commit_Guidelines , if you could rebase this to fix the commit message to be "AGOS: Mute fix for ADLIB Accolade" that would be optimal. Thanks again for your contribution.

digitall:
Since over a month has passed, I think sufficient time has passed for review / objections. This code looks fine and is limited to AGOS engine, so I am going to merge so this improvement to AGOS can be tested with nightly builds and any patches done by the AGOS team in tree.

@Fantohin : Thanks for your code contribution.

engines/agos/midi.cpp
```diff
@@ -28,6 +28,8 @@
 #include "agos/agos.h"
 #include "agos/midi.h"
 
+#include "agos/drivers/accolade/adlib.cpp"
```
criezy: This is is likely the cause of several builds now failing on buildbot due to multiple definitions of several symbols. All the symbols defined in `engines/agos/drivers/accolade/adlib.o` are also now also defined in `engines/agos/midi.o`.

digitall: Damn. Mea Culpa. I should have spotted that. @Fantohin : Not the best idea as @criezy indicated. I can see why you did this though as the code structure is missing headers and does not make your changes easy.

Should be fixed by commit 3a0ed17e66e01d8b5175904ed821244be8b278b8 which I have pushed, but will see what buildbot indicates.

PR #1613, opened 2019-05-02
SCI: Disable FPFP Mac script that uses missing view

sluicebox:
FPFP Mac is missing view 844 of the cook leaving town. This is the unusual view that 67908559167844a1a207d1544dfcdbd8f8ca1260 fixed for PC versions. In Mac it's simply missing and the scene would crash in the original.

This a minor optional scene so I patched it out for Mac so that it doesn't crash. Fixes bug #10954

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -1274,11 +1274,37 @@ static const uint16 freddypharkasPatchMacEasterEgg[] = {
 	PATCH_END
 };
 
+// FPFP Mac is missing view 844 of Hop Singh leaving town, breaking the scene.
+//  This occurs when going to the desert (room 200) after the restaurant closes
+//  but before act 3 ends. This would also crash the original so we just disable
+//  this minor optional scene.
+//
+// Applies to: Mac Floppy
+// Responsible method: rm200:init
+// Fixes bug #10954
+static const uint16 freddypharkasSignatureMacHopSingh[] = {
+	0x89, 0x77,                     // lsg 77
+	0x35, 0x13,                     // ldi 13
+	0x1a,                           // eq? [ did restaurant just close? ]
+	0x31, 0x46,                     // bnt 46 [ skip hop singh scene ]
+	SIG_ADDTOOFFSET(+0x41),
+	SIG_MAGICDWORD,
+	0x72, 0x01, 0xd0,               // lofsa hopSingh [ hard-coded big endian for mac ]
+	0x4a, 0x20,                     // send 20 [ hopSingh init: ... setScript: sLeaveTown ]
+	SIG_END
+};
+
+static const uint16 freddypharkasPatchMacHopSingh[] = {
+	0x33, 0x4b,                     // jmp 4b [ always skip hop singh scene ]
+	PATCH_END
+};
+
 //          script, description,                                      signature                            patch
 static const SciScriptPatcherEntry freddypharkasSignatures[] = {
 	{  true,     0, "CD: score early disposal",                    1, freddypharkasSignatureScoreDisposal, freddypharkasPatchScoreDisposal },
 	{  true,    15, "Mac: broken inventory",                       1, freddypharkasSignatureMacInventory,  freddypharkasPatchMacInventory },
 	{  true,   110, "intro scaling workaround",                    2, freddypharkasSignatureIntroScaling,  freddypharkasPatchIntroScaling },
+	{  true,   200, "Mac: skip broken hop singh scene",            1, freddypharkasSignatureMacHopSingh,   freddypharkasPatchMacHopSingh },
 	{  true,   235, "CD: canister pickup hang",                    3, freddypharkasSignatureCanisterHang,  freddypharkasPatchCanisterHang },
 	{  true,   270, "Mac: easter egg hang",                        1, freddypharkasSignatureMacEasterEgg,  freddypharkasPatchMacEasterEgg },
 	{  true,   320, "ladder event issue",                          2, freddypharkasSignatureLadderEvent,   freddypharkasPatchLadderEvent },


m-kiewitz:
Hmmmm, I would prefer it if this patch was optional and maybe we checked if the view as a stand alone resource exists. That way people would be able to fix the issue by dumping the view from their PC release (if that actually works that is, but I don't see a reason why it wouldn't).

What does the original interpreter does in this case? Can you please check?
Does it crash as well?

sluicebox:
The original Mac interpreter crashes.

I like restoring but I think if you saw this "scene" you'd see I'm being generous with that term and it's so minor it's not worth it. It's just 5 seconds of a character walking off screen that occurs at the edge of the game during a brief event window. I'd post a link but I couldn't find a single youtube playthrough that has it, and I checked as many as I could find to avoid playing all the way through on DOS for the other bug. Everyone stays in town solving the puzzles at hand. I think that's why this one is so problematic, it's just out there and unimportant and got forgotten.

Disk-dumping views doesn't work across platforms, there are native endian values in there, so that crashes.

m-kiewitz:
Well, one could fix the view data, hell I could even try to add detection code, that figures this out all by itself.
I implemented the ability to dump 256-color views into 16-color games too just for fun.

Anyway, I don't like removing content from the game even when it's just really small content.
It's simple to add this patch but with a check included that simply looks if the view is available or not.
Patch it out when the view is not available. Don't do anything when it is available.
I don't want to block any fan from fixing this version of the game.
And one single fan would be enough. That person could upload the view data somewhere (we ourselves can't, because it would still be copyrighted content, if we could I would add the view myself).

sluicebox:
How about we start with this to stop the Mac crashing and then you can add the view check?

I don't know how to add view-dependent script patches, I wouldn't be able to test it since I'm not converting the view to big endian, and I just don't think anyone would take advantage of it in this case even if they knew they could.

For context, even in the PC version this has never not crashed in the history of ScummVM and no one said peep until I blundered into it and fixed it the other day. It's 5 seconds of walking animation that even youtube playthrough folks don't encounter. And that's their job! That's great if we use the resource if it's there on Mac but I don't want us to lose sight that this notably inconsequential.

Yes, I meant just mac-to-other cross-platform dumps are incompatible.

bluegr:
Well, we can always perform the check in code, and enable the patch if the view is missing. We already have a lot of script patches, which are enabled conditionally via enablePatch().

m-kiewitz:
> How about we start with this to stop the Mac crashing and then you can add the view check?
I can add the code right afterwards, of course.
It's trivial though.
Just set the script patch to not be enabled (the very first value in the table is set to true - which means enabled, which means the patch is automatically applied)
The code is inside ScriptPatcher::processScript.
			// Do additional game-specific initialization
			switch (gameId) {
			case GID_KQ5:
				if (g_sci->_features->useAltWinGMSound()) {
					// See the explanation in the kq5SignatureWinGMSignals comment
					enablePatch(signatureTable, "Win: GM Music signal checks");
				}
Simply add a case for Freddy Pharkas, check the resource and then call "enablePatch" on your patch.
Like this:
if (g_sci->getResMan()->testResource(ResourceId(kResourceTypeView, XXXX))
That's it.
I can do it too, but I'm sure that you are able to put it together.

And for testing, you can simply use the DOS view, or any other view. When it error()s out like before (I hope it did, if it actually crashes, then that crash should be fixed) when the file exists and works when it's removed, then that's good enough.

> I don't know how to add view-dependent script patches, I wouldn't be able to test it since I'm not converting the view to big endian
You wouldn't even have to do this, add a view patch -> it should  crash again.
Remove the view patch -> it shouldn't crash.

>and I just don't think anyone would take advantage of it in this case even if they knew they could.
A fan even combined Space Quest 4 Floppy and CD. And we pissed a few people off, because they thought we wanted to blacklist that whole so called "patch", although in fact we black listed it only, because it was a download with literally every already combined, which we have to consider a pirated copy.
That thing was changed to a tool, which rips the resources out of the floppy version (of the user), which we then did not blacklist of course, because that's how it should have been done from the start.

> For context, even in the PC version this has never not crashed in the history of ScummVM and no one said peep until I blundered into it and fixed it the other day. 
If you removed the sequence on PC as well, then I simply missed that and the very same should apply.

sluicebox:
Yes, what you're saying makes total sense, and I apologize because I've gotten us off track. I humbly beg everyone's forgiveness in recapping the confusing timeline...

1. Sierra ships an unusual view 844 in DOS Floppy / CD that has loops that reference loops that reference loops that reference loops. It runs fine in DOS. No SCI resource viewer handles this view correctly. The view has crashed ScummVM until I fixed it the other day in 67908559167844a1a207d1544dfcdbd8f8ca1260

2. Sierra ships Mac Floppy without the view at all. "Oops." It crashed in Mac in 1993. It crashes now in ScummVM. This script patch is is the only way for a player who shows up with just the data files to play without exploding. It doesn't remove content, it adds the ability to walk to this room at the end of act 3.

My objection to making the patch conditional on the view's presence is that I am certain no one will ever take advantage of that effort and I can't test that it would work because I'd have to manufacture a big endian view, which even if I wanted to do, I'm certain would make me the only person to ever do so.

This is going to sound snarky but it's not, I just want the context understood: the Mac version doesn't load in 2.0, didn't for a year, and only one dude noticed, and even then I'm the one who shepherded the fix in. The Mac version has never been completable in ScummVM until a few weeks ago when I fixed b195ff5e190388f8598a7db4822d57f73f4b6b65, which was reported by the same guy 3 years ago with no activity. You can count the people in this world who care about this version by their usernames in the bug tracker, and it looks like half of them have given up.

It's just hard for me to justify writing code to support an external resource file that doesn't exist for a few inconsequential seconds of walking animation that no one encounters in a version that demonstrably no one plays. I hope I've demonstrated that I'm a lunatic for making this stuff work who should be committed but even I can't cosign this minutia. I'm not against the feature, I just don't believe in it, and I think someone who does should write and explain it. We all have limits! =)

bluegr:
Personally, I don’t mind about a missing edge case animation. Having the game crash is worse

bluegr:
The additional check is only necessary for cases where the user has modified the game resources.

Thus, this can be merged, as it works for normal cases, and any additional checks can be added afterwards in-tree.

PR #1614, opened 2019-05-02
AmigaOS4: Activate syntax highlighting in GitHub

polluks:
Only cosmetic...


digitall:
@polluks : Thanks for your code contribution. This PR is a bit tangled. There are two spurious merge commits which can be removed by a rebase and the two "real" commits do not comply with our commit guidelines i.e. https://wiki.scummvm.org/index.php/Commit_Guidelines

However, I think the changes are fairly clear here i.e. first changing the return value to 20 for the RM2AG.rexx script and secondly, changing the .rx files to .rexx line endings so that editors / Github do syntax highlighting? Will check with our AmigaOS4 porter and then will apply these changes manually. Please standby..

@raziel- : Any problem with making these changes?

raziel-:
I wouldn't have any objections if there wouldn't be the fundamental problem of the (now) missing text (ascii) based readme (on AmigaOS4), which went missing since the change to use a markdown version of it. (which isn't supported on AmigaOS4, at least not with basic tools respectively only by loading it into a web browser, which defies the readme being "simple" to open and use)Actually I'm pondering about completely dropping the rexx script (and conversion to AmigaGuide).Its either that or a more or less complete rewrite of the script.There was talks about the possibility of keeping an ascii based readme for (some) platforms, but I'm not sure how practicable that would be on ScummVM's side?
null

polluks:
@raziel- This is the solution
https://github.com/scummvm/scummvm/blob/9ebb91411aaf57c0ccc5667b2f0743c5454fd1da/Makefile.common#L239

criezy:
> There was talks about the possibility of keeping an ascii based readme

Technically the markdown file is a ascii file (except for the dot on the very first line that is UTF-8). It you read it as a text file the formatting instructions make it slightly less easy to read than the old plain text readme, but it remains perfectly readable.

Also a few weeks ago I also stumbled on a powershell script to convert a markdown file to a amiga guide. The logic could possibly be extracted (and simplified since we don't have any image to handle) to write something that would work on AmigaOS.

Mataniko:
We can use pandoc to output a plaintext for Amiga, I wouldn't be surprised it it supports whatever specialized for at you need. We already generate a .txt file for some other distributions.
null

criezy:
Yes using pandoc would be ideal, although I don't think it can output to amiga guide format, but it could possibly output to something else (such as plain text or html) that could then easily be converted to amiga guide. 

The reason I didn't mention pandoc is that my understanding was that there is no AmigaOS port of pandoc, so compiling on AmigaOS (as opposed to cross-compiling on another system) would not be able to generate this file.

raziel-:
Thank you all for the replies.I was put off by the auto generated links in the first line of markdown readme.I thought it might be cluttered like this in the whole file.I can work with simply ignoring those then, I guess.I'd rather not rely on a third party program because while it may generate a working AG file, it could also produce bugs inside it which I'd have to fix manually afterwards.Guess I have to simply rewrite the script.So...no objections for the change from me
null

bluegr:
Thanks for your feedback @raziel-. Squashing and merging

raziel-:
Not sure if it's still ok to post on a closed item, but i can see now why it's impossible (or rather a PITA) to rewrite my script to use the markdown version of readme.

i have to use the RAW version of it and in there lots of lines get rolled over to the next line (why, i don't know, maybe they are too long?) and because of that i would need to "fix" them in my script which is too much work..

e.g. (note the "notes")
- [3.17) Simon the Sorcerer games
        notes](#317-simon-the-sorcerer-games-notes)
      - [3.18) Starship Titanic game
        notes](#318-starship-titanic-game-notes)
      - [3.19) The Curse of Monkey Island
        notes](#319-the-curse-of-monkey-island-notes)
      - [3.20) The Feeble Files notes](#320-the-feeble-files-notes)
      - [3.21) The Legend of Kyrandia
        notes](#321-the-legend-of-kyrandia-notes)

If this could be fixed in the readme itself, i'd be grateful and start rewriting my script

Thanks

PR #1617, opened 2019-05-04
COMMON: Add SeekableWriteStream class

dreammaster:
This introduces a new SeekableWriteStream class to play a similar
role as the SeekableReadStream class, but for write streams.
It add methods for seek and size, and the following classes have
been changed to support it:
1) Memory stream classes
2) DumpFile class
3) StdIOStream class

I tried to have minimal impact on the existing system. Please note
the following:
1) DumpFile still only has a reference to a WriteStream internally. I
didn't want to change it to SeekableWriteStream, since that would
require also converting he file system node methods to return it as
well. And I'm not sure if there are any backends that have their own
custom implementations of files other than StdIOStream. If there are,
they'd eventually need to be updated to use SeekableWriteStream
as well, and we can eventually change all the various methods to
return SeekableWriteStream instead of WriteStream. For now,
DumpFile's seek and size methods try to dynamically cast the handle
to a SeekableWriteStream, and pass on the method if it supports it.

2) SeekableMemoryWriteStream's seek method is changed from
returning an int32 position to the standard bool success/fail. I tried
compiling with the method temporarily set to void without error,
which means there's nothing depending on the old result anyway.

### DIFF
--- a/common/file.cpp
+++ b/common/file.cpp
@@ -219,4 +219,14 @@ bool DumpFile::flush() {
 
 int32 DumpFile::pos() const { return _handle->pos(); }
 
+bool DumpFile::seek(int32 offset, int whence) {
+	SeekableWriteStream *ws = dynamic_cast<SeekableWriteStream *>(_handle);
+	return ws ? ws->seek(offset, whence) : -1;
+}
+
+int32 DumpFile::size() const {
+	SeekableWriteStream *ws = dynamic_cast<SeekableWriteStream *>(_handle);
+	return ws ? ws->size() : -1;
+}
+
 } // End of namespace Common
--- a/common/file.cpp
+++ b/common/file.cpp
@@ -221,7 +221,7 @@ int32 DumpFile::pos() const { return _handle->pos(); }
 
 bool DumpFile::seek(int32 offset, int whence) {
 	SeekableWriteStream *ws = dynamic_cast<SeekableWriteStream *>(_handle);
-	return ws ? ws->seek(offset, whence) : -1;
+	return ws ? ws->seek(offset, whence) : false;
 }
 
 int32 DumpFile::size() const {
--- a/backends/fs/stdiostream.h
+++ b/backends/fs/stdiostream.h
@@ -28,7 +28,7 @@
 #include "common/stream.h"
 #include "common/str.h"
 
-class StdioStream : public Common::SeekableReadStream, public Common::WriteStream, public Common::NonCopyable {
+class StdioStream : public Common::SeekableReadStream, public Common::SeekableWriteStream, public Common::NonCopyable {
 protected:
 	/** File handle to the actual file. */
 	void *_handle;
@@ -43,17 +43,17 @@ public:
 	StdioStream(void *handle);
 	virtual ~StdioStream();
 
-	virtual bool err() const;
-	virtual void clearErr();
-	virtual bool eos() const;
+	virtual bool err() const override;
+	virtual void clearErr() override;
+	virtual bool eos() const override;
 
-	virtual uint32 write(const void *dataPtr, uint32 dataSize);
-	virtual bool flush();
+	virtual uint32 write(const void *dataPtr, uint32 dataSize) override;
+	virtual bool flush() override;
 
-	virtual int32 pos() const;
-	virtual int32 size() const;
-	virtual bool seek(int32 offs, int whence = SEEK_SET);
-	virtual uint32 read(void *dataPtr, uint32 dataSize);
+	virtual int32 pos() const override;
+	virtual int32 size() const override;
+	virtual bool seek(int32 offs, int whence = SEEK_SET) override;
+	virtual uint32 read(void *dataPtr, uint32 dataSize) override;
 };
 
 #endif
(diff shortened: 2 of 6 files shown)


dreammaster:
Not sure if the Travis compilation failure for the devtools create_titanic is legitimate or not. If it is, I'm buggered if I can understand what the actual problem is. Worse, I'm having trouble recreating a Visual Studio project for create_titanic to try compiling it on my portable laptop. So I'm at a loss if it's a problem or not.

bgK:
> Not sure if the Travis compilation failure for the devtools create_titanic is legitimate or not. If it is, I'm buggered if I can understand what the actual problem is. Worse, I'm having trouble recreating a Visual Studio project for create_titanic to try compiling it on my portable laptop. So I'm at a loss if it's a problem or not.

This is a real problem. `MemoryWriteStreamDynamic::seek` is defined in `stream.cpp`. However that file is not linked in `create_titanic`, so the virtual function table for `MemoryWriteStreamDynamic` cannot be created.

backends/saves/savefile.cpp
```diff
@@ -56,6 +56,16 @@ int32 OutSaveFile::pos() const {
 	return _wrapped->pos();
 }
 
+bool OutSaveFile::seek(int32 offset, int whence) {
+	SeekableWriteStream *ws = dynamic_cast<SeekableWriteStream *>(_wrapped);
+	return ws ? ws->seek(offset, whence) : -1;
```
bgK: `-1` should be `false` here.

common/file.cpp
```diff
@@ -219,4 +219,14 @@ bool DumpFile::flush() {
 
 int32 DumpFile::pos() const { return _handle->pos(); }
 
+bool DumpFile::seek(int32 offset, int whence) {
+	SeekableWriteStream *ws = dynamic_cast<SeekableWriteStream *>(_handle);
+	return ws ? ws->seek(offset, whence) : -1;
```
bgK: `-1` should be `false` here.

dreammaster: Thanks. I've committed a fix for the problems

bluegr:
Overall, it’s good that you’ve added this functionality, which completes the file stream API nicely.

However, files saved via DumpFile and OutSaveFile are usually created sequentially, so adding seeking in these cases would add functionality that isn’t used in this kind of code

dreammaster:
Well, there are some use cases. The Glulxe sub-engine, for example, writes out data and uses seeking to go back and write out chunk sizes after the fact. This whole work on seekable write streams grew out of that. The sub-engine has a lot of code that treats writes to memory the same as writes to files, and relied on being able to seek. I was also lucky in that the engine creates savegames without compression, to allow for compatibility with existing Quetzal save formats, so the seeking this pull request adds in will work. So without the ability to treat both the save file and memory streams as a seekable write streams, I wouldn't easily be able to simplify a lot of the saving code.

bluegr:
Interesting. Thanks for the heads-up :)

bluegr:
Since this modifies common functionality, I propose to merge this on Sunday, 12th May - one week in total, in case there are any further comments on it

backends/saves/savefile.cpp
```diff
@@ -56,6 +56,16 @@ int32 OutSaveFile::pos() const {
 	return _wrapped->pos();
 }
 
+bool OutSaveFile::seek(int32 offset, int whence) {
+	SeekableWriteStream *ws = dynamic_cast<SeekableWriteStream *>(_wrapped);
+	return ws ? ws->seek(offset, whence) : false;
+}
+
+int32 OutSaveFile::size() const {
+	SeekableWriteStream *ws = dynamic_cast<SeekableWriteStream *>(_wrapped);
```
fingolfin: There are (or at least, used to be) many savefile implementions which do not provide seeking. And as you point out, compressed saves don't support seeking.

So then engines expecting seekable savefiles won't work. But then the engine has no way to find out about this unless they try to seek here, it seems? Well, I guess they could try to seek and check for an error; but perhaps it would make more sense to add a specific method to check whether seeking is supported? 

But even then, you'd exclude many backends. Perhaps a better solution for your original problem would be to first create the savestate inside a memory buffer (possibly managed by a `SeekableMemoryWriteStream`), and then once that is done, just dump the whole thing into a normal `OutSaveFile`?

criezy: I agree with @fingolfin comment that using first a seekable memory stream and then dumping it to a file in the engine might be a better approach than trying to make the file stream seekable.

dreammaster: fingolfin: I wasn't aware that that would be the case (other backends not using StdIOStream). But in retrospect I guess it makes sense, since particularly for consoles, save file creation/retrieval could need special custom code. I guess having SeekableWriteStream for the Magnetic sub-engine is a bust.

The question remains whether I should revert OutSaveFile from SeekableWriteStream to WriteStream. I'm leaning towards yes, since it will avoid confusion in the future until and unless all supported save file implementations ever support it.

criezy: @dreammaster: I would also lean toward reverting that change.

dreammaster: And done. I've dropped the OutSaveFile commit.

PR #1618, opened 2019-05-04
I18N: Correct Portuguese (Portugual) metadata

SupSuper:
Fixes the Portuguese language name displayed by ScummVM and the plural rule (not sure if ScummVM uses plurals but can't hurt).


lotharsm:
Thanks, merged!

PR #1622, opened 2019-05-10
CREDITS: mark Vicent Marti as retired

fingolfin:
He was last active in 2009 and has long moved on to work on other things.

Thank you for merging my PR from yesterday so quickly. I just realized that Vicent's entry should also be updated.

Perhaps also further people should be given credit for the GUI? Here's the top 10 of people who modified the GUI C++ source code (to filter out i18n changes; of course one may also want to credit people who worked on the themes
```
$ git shortlog -sn gui/*.cpp gui/*.h gui/*/*.cpp gui/*/*.h  | head -n 10
   580	Max Horn
   297	Eugene Sandulenko
   290	Johannes Schickel
   118	Vicent Marti
   114	Alexander Tkachev
   105	Thierry Crozat
    98	Torbjörn Andersson
    41	Filippos Karapetis
    39	Bastien Bouclet
    29	Willem Jan Palenstijn
```

### DIFF
--- a/devtools/credits.pl
+++ b/devtools/credits.pl
@@ -1039,7 +1039,7 @@ begin_credits("Credits");
 
 			begin_section("GUI");
 				add_person("Max Horn", "Fingolfin", "(retired)");
-				add_person("Vicent Marti", "tanoku", "");
+				add_person("Vicent Marti", "tanoku", "(retired)");
 				add_person("Eugene Sandulenko", "sev", "");
 				add_person("Johannes Schickel", "LordHoto", "(retired)");
 			end_section();
--- a/AUTHORS
+++ b/AUTHORS
@@ -419,7 +419,7 @@ ScummVM Team
 
     GUI:
        Max Horn                     - (retired)
-       Vicent Marti
+       Vicent Marti                 - (retired)
        Eugene Sandulenko
        Johannes Schickel            - (retired)
 
--- a/gui/credits.h
+++ b/gui/credits.h
@@ -512,6 +512,7 @@ static const char *credits[] = {
 "C0""Max Horn",
 "C2""(retired)",
 "C0""Vicent Marti",
+"C2""(retired)",
 "C0""Eugene Sandulenko",
 "C0""Johannes Schickel",
 "C2""(retired)",


devtools/credits.pl
```diff
@@ -1039,7 +1039,7 @@ sub add_paragraph {
 
 			begin_section("GUI");
 				add_person("Max Horn", "Fingolfin", "(retired)");
-				add_person("Vicent Marti", "tanoku", "");
+				add_person("Vicent Marti", "tanoku", "(retired)");
```
fingolfin: One could of course also point out that it was (initially) a GSoc task, but I don't remember in which year etc., so I'll leave that to somebody else, if it is desired.

sev-: Yes, that's a good point. He was never part of the team full time. And we do not mark one time contributors, which we have a few, as retired.

Also, could you please check your e-mail, I sent you a request regarding Google Analytics.

bluegr:
Hi @fingolfin! Great to see you around :)

Thanks for updating the status, and you are right that more people should be credited for the GUI code.

Since you’ve already updated the status of the maintainers, could you also add the other people you found in the credits for the GUI code? This could be done as part of this PR, as you’ve raised the issue here.

Thanks again!

fingolfin:
I don't think it makes sense for me to do that: I don't know who should get credit and who not, that's something the ScummVM team needs to decide. I also don't know who is active and who not; I just happen to know for Vicent.
So, I have no plans to update this PR; if it not deemed quite right for your needs, that's of course fine, just close it then.

bluegr:
My suggestion was to discuss about the contributors in this PR, since they are updated anyway.

This can be a discussion done via the -devel list, and the results can be discussed in another pull request. In the meantime, this change is fine to be merged.

sev-:
@fingolfin Am I being ignored by you completely? Looks like this.

bluegr:
After a discussion with @sev-, it does not make much sense to mark GSoC people as retired, as they are usually contributors for a specific period, so they do not actually retire :)

Vincent, along with all GSoC’ers should be marked as “GSoC task”. I’ll revert this commit, and mark the rest of the GSoC people as such

PR #1626, opened 2019-05-12
GUI: Add dark mouse cursor based on the "Adwaita" theme from GNOME.org

lotharsm:
To be honest, I never was a huge fan of the mouse cursor we included starting with the ScummVM Modern Theme. To find a better replacement for the ScummVM Remastered theme, I decided to use a mouse cursor based on the Adwaita icon set from the GNOME project (dual licensed as LGPL + CC-BY-SA).
adwaita
Please let me hear your thoughts.


Mataniko:
I don't know enough of the history, but can't we just keep the OS cursor and leave it in the users hands?

bgK:
> I don't know enough of the history, but can't we just keep the OS cursor and leave it in the users hands?

Some of the platforms ScummVM supports don't have an OS cursor. I know for sure this is the case for the PlayStation 3.

lotharsm:
> Some of the platforms ScummVM supports don't have an OS cursor. I know for sure this is the case for the PlayStation 3.

Since I'm not familiar with the relevant parts of the GUI code - would it be possible to at least disable using our "custom" cursor for Win/Mac/Linux?

bluegr:
This dark mouse cursor looks off with the orange theme. The current orange and black one is a much better match.

criezy:
I am with @blugr here. I actually prefer the current orange cursor to your black cursor (or the macOS cursor - the two are very similar).

lotharsm:
Thank you for your feedback :)

PR #1628, opened 2019-05-14
SUPERNOVA: Grammar fixes in English newspaper images

JenniBee:
I've started to play through the first episode of Mission: Supernova, to attempt to polish the English translation a bit.

I did some grammar fixes on the English newspaper images. The first mostly had commas added where needed, and the phrase "one of the planet in the Arsano system" was changed to "one of the planets in the Arsano system".

The second newspaper also mostly had commas added where needed, the phrase "and getting permission" was changed to "and received permission", and "commited" was changed to "committed".

I've changed both the Photoshop and GIMP images to reflect the changes.


criezy:
Thank you, those are all good changes. But there are a few issues:
- The gimp image has all the text duplicated in. The original image has one background layer and four text layers for each block of text. But rather than edit the text layers, it looks like you have added a new image layer on top of them with the corrected text? That will need to be fixed.
- The size in the ppm files is not recognized by our create_supernova tool. But I actually think this is an issue with our tool and not with your images, and I will update the tool (the short story is that our tool expect a space between the width and height value, but your images have a end of line character instead, which is perfectly valid according to the [format description](http://netpbm.sourceforge.net/doc/pbm.html), and the issue is with our tool being too strict). So I will change that in our tool. Out of curiosity how did you create those? Did you export them using the gimp export to pbm?

If you want I can redo the gimp images properly, restarting from the original ones and editing the existing text layers to add your changes. Or do you want to try to do it yourself? The idea is in gimp to use the text tool, click on the existing text, and if asked if you want to edit the current text say yes (and then edit the text).

criezy:
I have now redone the gimp images properly, incorporating your changes. Hopefully I didn't forget any.
I also took this opportunity to change the font and make the image more readable. This was pushed in commit 96c429d.

PR #1629, opened 2019-05-18
WINTERMUTE: Fix fill rect area on endSaveLoad()

lolbot-iichan:
Wintermute fills screen area with black color when save/load process is finished.
There was a misprint in ScummVM's code which caused only left part of the screen go black.


digitall:
I checked against the Common::Rect constructors here:
https://github.com/scummvm/scummvm/blob/master/common/rect.h#L101

I initially thought that this was wrong with the parameters being top, left, bottom, right ... but they are _NOT_ ... they are x1, y1 i.e. top left point and x2, y2 i.e. bottom right point.

Thus @lolbot-iichan is perfectly correct that these parameters are reversed. I think this is pretty self evident and will merge as-is.

PR #1632, opened 2019-05-21
SCI: Fix FPFP CD 'Dummy Msg' message texts

sluicebox:
This fixes 17 message texts in the CD version of FPFP, bug #10964.

This is a new one, the audio is right but the text is in all the wrong places, with strings like "Dummy Msg" in most of them. I'm assuming this is because this game didn't formally offer a text mode in the CD version. The installer didn't even have a No Soundcard option. What's weird is that these three sequences are the only ones like this, everything else seems fine, so I don't expect this to need future updates.

Let me know what you think. I'm not wild about touching processString() for this, but it's const char* on down, and now future workarounds can also trim trailing text by setting MessageRecord::length.

### DIFF
--- a/engines/sci/engine/message.cpp
+++ b/engines/sci/engine/message.cpp
@@ -351,6 +351,70 @@ bool MessageState::getRecord(CursorStack &stack, bool recurse, MessageRecord &re
 			return true;
 		}
 
+		// FPFP CD has several message sequences where audio and text were left
+		//  out of sync. This is probably because this version didn't formally
+		//  support text mode. Most of these texts just say "Dummy Msg". All the
+		//  real text is there but it's either concatenated or in a different
+		//  tuple. We extract and return the correct text. Fixes #10964
+		if (g_sci->getGameId() == GID_FREDDYPHARKAS && g_sci->isCD() &&
+			g_sci->getLanguage() == Common::EN_ANY && !g_sci->isDemo()) {
+			byte textSeq = 0;
+			uint32 substringIndex = 0;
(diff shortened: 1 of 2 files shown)


bluegr:
Nice work overall!

Some comments:
- It would be good to create a workaround mechanism for these cases, and place them inside workarounds.cpp. At the moment, we got workarounds for uninitialized local variables and invalid arithmetic operations
- FPFP did not originally have a text+speech option, which is why these cases exist
- Is it not possible to have text (without speech) in this version?
- KQ7 suffers from similar issues: it does contain subtitles, but they were left unfinished in the final version, and were not enabled. Perhaps similar subtitle workarounds could be used for this game as well in the future

sluicebox:
Yes, text without speech wasn't allowed in the original, which is what makes this one unusual. There's no UI for it in the game and they even removed the usual "No Audio/Voice Card" option from the installer.

I like the idea of generalizing all the text workarounds, and I've been hoping to take that on eventually. The advantage to procrastinating is that as these pile up it makes it clearer what all needs to be done, but I think we're close. Lord help me, I've jumped into QFG4 with both feet, and that will probably unearth some more.

The tricky thing is that so many of these seem to be one-offs and we'd need audio and sync36 workarounds too. It all looks doable, but it might take a while. I see a half dozen script patches that could be removed with that in place, which is exciting.

Should we merge this first in anticipation of a workaround refactor that removes all the game-specific stuff from getRecord()?

bluegr:
Agree. This can be merged, and the code can be refactored in-tree. Merging

PR #1633, opened 2019-05-21
COMMON: Fix button state desynchronization when warping mouse

lephilousophe:
When warping mouse, if there is a mouse release event in the event queue, it gets eaten by the purgeMouseEvents function and _buttonState gets desynchronized.
This pull request prevents this by updating _buttonState even when purging events.


sev-:
Good catch, thank you!

PR #1636, opened 2019-05-23
SCI: Fix QFG4 random revenant kopeks, bug #10966

sluicebox:
Another win for a well placed warning() in the engine! =)


bluegr:
Nice work, once again! :) Merging

PR #1637, opened 2019-05-24
CONFIGURE: Remove the 'u' flag from 'ar' in mingw32

lotharsm:
This patch silences the warnings about the (outdated?) 'u'
flag that appears at least when building and linking ScummVM
statically on mingw32:

"`u' modifier ignored since `D' is the default (see `U')"

The cause for this appears to be a bug in recent versions of
libtool that most likely will also appear when building on
Linux. However, since I have not tested this yet, this patch
only disables it for mingw32.

The only downside from this patch is that building a non-clean
build will take (slightly?) longer since we are no longer
just updating (that's what the 'u' is for) changed libraries
during the linking stage, but linking all .a files again.

### DIFF
--- a/configure
+++ b/configure
@@ -3228,7 +3228,7 @@ if test -n "$_host"; then
 		*mingw32*)
 			_sdlconfig=$_host-sdl-config
 			_windres=$_host-windres
-			_ar="$_host-ar cru"
+			_ar="$_host-ar cr"
 			_ranlib=$_host-ranlib
 			;;
 		mips-sgi*)


lotharsm:
In case there are no objections, I propose to merge this next sunday (June 2nd).

bluegr:
After reading through the discussion in this thread:
https://github.com/rsyslog/rsyslog/issues/1179
and this:
https://bugzilla.redhat.com/show_bug.cgi?id=1155273
as well as the linked threads on these, removing the u flag is a good way to go. Quoting:

    The warning message is correct.
    Deterministic archives do not store file 
    timestamps, so the 'u' option to only 
    replace newer files cannot work.  If you 
    are able to modify the build machinery for 
    your project you can fix the problem by 
    either using "ar cr ..." or "ar crUu ..." 
    depending upon whether you want to 
    maintain the deterministic nature of the 
    libraries or not.  (I recommend the 
    former).

Thus, I agree on the rationale of the patch, and that it should be merged eventually

lotharsm:
Thanks for your feedback @bluegr! Merging!

PR #1638, opened 2019-05-25
STARTREK: Add detection for reported English floppy variant

lotharsm:
Reported by ccorbacho in trac/#10921


bluegr:
Thanks, merging

PR #1641, opened 2019-05-26
BLADERUNNER: Add flag to disable sewer bridge breaking

dafioram:
This will make that part less frustrating. The rat can still get you.

### DIFF
--- a/engines/bladerunner/script/ai/free_slot_a.cpp
+++ b/engines/bladerunner/script/ai/free_slot_a.cpp
@@ -541,6 +541,10 @@ void AIScriptFreeSlotA::FledCombat() {
 }
 
 void AIScriptFreeSlotA::checkIfOnBridge() {
+	if (_vm->_cutContent && Query_Difficulty_Level() == kGameDifficultyEasy) {
+		// Make the bridge indestructible on easy mode for the enhanced version
+		return;
+	}
 	float x, y, z;
 	Actor_Query_XYZ(kActorFreeSlotA, &x, &y, &z);
 	// bug? this should probably check if McCoy is close enough because bridge will break long after rat died and player tries to walk through
--- a/engines/bladerunner/script/scene/ug15.cpp
+++ b/engines/bladerunner/script/scene/ug15.cpp
@@ -40,7 +40,8 @@ void SceneScriptUG15::InitializeScene() {
 	} else {
 		Setup_Scene_Information(-238.0f, 48.07f,  222.0f, 180);
 		if (Game_Flag_Query(kFlagUG15RatShot)
-		 && Random_Query(1, 10) == 10
+		 && (Random_Query(1, 10) == 10)
+		 && !(_vm->_cutContent && Query_Difficulty_Level() == kGameDifficultyEasy)
 		) {
 			Game_Flag_Reset(kFlagUG15RatShot);
 		}


peterkohaut:
To be honest, I don't like this approach as it is making the encounter  too easy (i've tried that long time ago).
Btw, I have planned to make it easier by not breaking the bridge if rat is shot on the bridge but mccoy is not standing on it and walking on it afterwards.

peterkohaut:
Or make default for easy difficulty only

engines/bladerunner/bladerunner.h
```diff
@@ -39,6 +39,7 @@
 #define BLADERUNNER_DEBUG_CONSOLE 0
 #define BLADERUNNER_ORIGINAL_SETTINGS 0
 #define BLADERUNNER_ORIGINAL_BUGS 0
+#define BLADERUNNER_INDESTRUCTIBLE_BRIDGE 1 // If enabled, the Act 4 bridge in the sewers won't collapse if McCoy and the rat are on it or if McCoy runs
```
peterkohaut: remove this, its not needed

engines/bladerunner/script/ai/free_slot_a.cpp
```diff
@@ -544,7 +544,7 @@ void AIScriptFreeSlotA::checkIfOnBridge() {
 	float x, y, z;
 	Actor_Query_XYZ(kActorFreeSlotA, &x, &y, &z);
 	// bug? this should probably check if McCoy is close enough because bridge will break long after rat died and player tries to walk through
-	if (-150.0 <= x && x < -30.0f) {
+	if (-150.0 <= x && x < -30.0f && !BLADERUNNER_INDESTRUCTIBLE_BRIDGE) {
```
peterkohaut: replace with "easy" settings  _vm->_settings->getDifficulty() == kGameDifficultyEasy and do it only when BLADERUNNER_ORIGINAL_BUGS  is set to 0

dafioram:
Does that look good?

engines/bladerunner/script/ai/free_slot_a.cpp
```diff
@@ -543,8 +544,14 @@ void AIScriptFreeSlotA::FledCombat() {
 void AIScriptFreeSlotA::checkIfOnBridge() {
 	float x, y, z;
 	Actor_Query_XYZ(kActorFreeSlotA, &x, &y, &z);
+	
+	bool isEasyDifficulty = _vm->_settings->getDifficulty() == kGameDifficultyEasy;
```
peterkohaut: sorry for nitpicking, but can you reorganize it a bit? that lines above are not used in this case  and can you use similar style of #if as the rest of the code, like:

... checifonbridge() {
#if !BLADERUNNER_ORIGINAL_BUGS
	if ( _vm->_settings->getDifficulty() == kGameDifficultyEasy) { // Make the bridge indestructable on easy mode, unless original behavior is desired
		return;
	}
#endif
	...oldcode
}


and i think you have a typo in indestructible"

engines/bladerunner/script/ai/free_slot_a.cpp
```diff
@@ -543,8 +544,14 @@ void AIScriptFreeSlotA::FledCombat() {
...
+	if (isEasyDifficulty && BLADERUNNER_ORIGINAL_BUGS == 0) { // Make the bridge indestructable on easy mode, unless original behavior is desired
+		return;
+	}
+
 	// bug? this should probably check if McCoy is close enough because bridge will break long after rat died and player tries to walk through
-	if (-150.0 <= x && x < -30.0f) {
+	if (-150.0 <= x && x < -30.0f) {
```
bluegr: Please fix this stray whitespace

engines/bladerunner/script/ai/free_slot_a.cpp
dafioram: Technically this is more of a QoL enhancement then a bug fix over the original, how do you feel about changing the name of the flag to be BLADERUNNER_ORIGINAL_BEHAVIOR.

peterkohaut: good point, then can you use _vm->_cutContent? we have a special version which has plenty of this kind of stuff (when adding new game to scummvm it is "Blade Runner with restored content")

dafioram:
Updated.

antoniou79:
Just chiming in to say that I'm ok with this being an improvement contained in the Easy Mode case. 

Just for the record, the idea for this "puzzle" is that the bridge is too rickety and will collapse if McCoy shoots at it / shoots the rat while it's on it. The suggested solution is to have McCoy trigger the mouse and then move out of the bridge and downwards, so the rat will cross the bridge and then McCoy can shoot it safely. However, this is one of the things that's maybe fun to design (maybe) but does not work well with the player since it's a sudden death with no warning, and this was and still it a notorious section of the original game (the other one being the Moraji encounter).

dafioram:
@antoniou79 would lengthening the count down timer improve the Moraji encounter?

Also if I kill the rat on the bridge and then leave left and go right again it doesn't show the dead rat, but he will spawn in the same place again so it just resets. 

So I don't know if there is a way to beat the rat bridge part other than the way you said (before this commit)

antoniou79:
> @antoniou79 would lengthening the count down timer improve the Moraji encounter?
I think the worse part of that encounter is the blast radius after you exit the Dermo Design Lab.  The timer is less of an issue for me, since the player understands that the situation is critical, and a bomb is involved.
I agree, though, that disabling the timer or lengthening it would further improve that sequence. 
Please keep such changes in the "enhanced" version and preferably only for easy mode (unless the weird blast radius is also some bug).
> 
> Also if I kill the rat on the bridge and then leave left and go right again it doesn't show the dead rat, but he will spawn in the same place again so it just resets.
The rat being gone is normal (in a way intended based on the rat AI logic), but the encounter on the bridge should not reset.  I tried a few times, and what you describe happened only once. so I'm guessing it's a bugged condition we should look into.

dafioram:
The rat not being dead resets with probability 10% every time you enter the room from the left side. Its possible that it was greater than 10% with how the comparison was being done so I've added a second commit to make that clear.

I don't know if this now deviates from the original or not.

@antoniou79 It looks like the rat possibly respawning was intentional. Would a desired enhancement/bug fix be that the rat never respawns once UG15 goal is set to kGoalFreeSlotAUG15WalkOut? (I.e., Once the rat appears for the first time McCoy could kill the rat or not and run left and when he comes back the rat would not appear). 

Currently, the rat will always reappear if McCoy doesn't kill it. Maybe we could overload the rat being dead field to also mean that the rat has walked out. So that running away would be a solution.

engines/bladerunner/script/ai/free_slot_a.cpp
```diff
@@ -543,6 +544,11 @@ void AIScriptFreeSlotA::FledCombat() {
 void AIScriptFreeSlotA::checkIfOnBridge() {
 	float x, y, z;
 	Actor_Query_XYZ(kActorFreeSlotA, &x, &y, &z);
+
+	if (_vm->_cutContent && _vm->_settings->getDifficulty() == kGameDifficultyEasy) {
+		// Make the bridge indestructible on easy mode for the enhanced version
+		return;
+	}
```
bluegr: Shouldn’t this check be made before the call to Actor_Query_XYZ()?

antoniou79:
> The rat not being dead resets with probability 10% every time you enter the room from the left side. Its possible that it was greater than 10% with how the comparison was being done so I've added a second commit to make that clear.

You're right. I should have payed more attention to the initialize script of the scene. So it is intentional (albeit in theory it will rarely ever happen). Personally, I'm in favor of keeping the intended stuff at least for the normal or hard mode. So, I'd suggest, as an improvement that makes sense, that in Easy mode and (_cutContent) this "Random_Query(1, 10) == 10" check could be skipped, so that the fight won't repeat itself if the player killed the rat once.

> @antoniou79 It looks like the rat possibly respawning was intentional. Would a desired enhancement/bug fix be that the rat never respawns once UG15 goal is set to kGoalFreeSlotAUG15WalkOut? (I.e., Once the rat appears for the first time McCoy could kill the rat or not and run left and when he comes back the rat would not appear).
> Currently, the rat will always reappear if McCoy doesn't kill it. Maybe we could overload the rat being dead field to also mean that the rat has walked out. So that running away would be a solution.

I'm not a fan of this solution. If the player has the sense to run away from the rat, they should be able to shoot it too, and since the bridge won't break (Easy mode, enhanced version due to this fix) then they should be ok. Leaving the scene to make the rat go away is a counter-intuitive solution for me.

engines/bladerunner/script/scene/ug15.cpp
```diff
@@ -40,7 +40,7 @@ void SceneScriptUG15::InitializeScene() {
 	} else {
 		Setup_Scene_Information(-238.0f, 48.07f,  222.0f, 180);
 		if (Game_Flag_Query(kFlagUG15RatShot)
-		 && Random_Query(1, 10) == 10
+		 && (Random_Query(1, 10) == 10)
```
peterkohaut: this changes nothing except of improved readability, commit message is a bit misleading

dafioram:
Alright, I updated the first commit so that the check for the uncut easy version is done before the actors position is checked for the bridge breaking and for the second commit I added the rat not respawning if playing the uncut on easy (otherwise it happens with %10 probability).

How does this look for everyone?

peterkohaut:
looks good to me

antoniou79:
It's ok for me, too.

engines/bladerunner/script/scene/ug15.cpp
```diff
@@ -40,7 +41,8 @@ void SceneScriptUG15::InitializeScene() {
...
+		 && !(_vm->_cutContent && _vm->_settings->getDifficulty() == kGameDifficultyEasy)
```
antoniou79: Actually, could you replace your _vm->_settings->getDifficulty() with Query_Difficulty_Level() in the file?
And then you can remove the "#include "bladerunner/settings.h"

dafioram:
I've switched out the _vm->_settings->getDifficulty() with Query_Difficulty_Level()  in both places. How does that look?

PR #1643, opened 2019-05-29
ENGINES: KINGDOM - New Engine (DO NOT MERGE)

Strangerke:
This pull request contains an engine for "Kingdom: The Far Reaches"
Only the Windows version is targetted so far.

The engine is missing a video player for the MVE format, which is slightly annoying as the game is a FMV. Nevertheless, the game should be completable.

The purpose of this pull request is to polish the existing code based on your reviews but also to expose publicly this old code so that developers may have a look at the MVE player. The format is known, as TMM guessed the missing opcodes and implemented them in ffmpeg.

Thanks in advance for your comments.


engines/kingdom/kingdom.cpp
```diff
@@ -0,0 +1,1628 @@
...
+			checkMainScreen();
+			if (_aTimerFlag) {
+				_aTimerFlag = false;
+				retval = 0x2F1;
+			} else if (_bTimerFlag) {
+				_bTimerFlag = false;
+				retval = 0x2F2;
+			} else if (_skylarTimerFlag) {
+				_skylarTimerFlag = false;
+				retval = 0x2F5;
```
bluegr: Please add an enum for these values of retval

engines/kingdom/kingdom.cpp
```diff
@@ -0,0 +1,1628 @@
...
+
+	CursorMan.showMouse(true);
+	_cursorDrawn = true;
+}
+
+void KingdomGame::cursorType() {
+	_mouseValue = 0;
+	if (_logic->_currMap != 1 && _logic->_statPlay >= 30) {
+		int var2 = _logic->_statPlay == 901 ? 16 : 0;
+		int var6 = _logic->_statPlay == 901 ? 35 : 16;
```
bluegr: Please add an enum for these values (901 is used a lot in other places, too)

engines/kingdom/kingdom.cpp
```diff
@@ -0,0 +1,1628 @@
...
+		int tmpVal = checkMouseMapAS();
+		if (tmpVal == -1) {
+			cursorTypeExit();
+			return;
+		} else {
+			_mouseValue = tmpVal;
+		}
+	}
+
+	switch(_mouseValue) {
```
bluegr: Please add an enum for all the cases here

engines/kingdom/kingdom.cpp
```diff
@@ -0,0 +1,1628 @@
...
+	if (_mouseValue >= 0x400)
+		_cursorDef = _cursorTable[_mouseValue - 0x400];
+	else 
+		_cursorDef = (_mouseValue != 0) ? 0x68 : 0x67;
+}
+
+int KingdomGame::checkMouseMapAS() {
+	if (isDemo()) {
+		for (int i = 0; i < 16; i++) {
+			if (_cursorX >= _mouseMapASDemo[_logic->_currMap][i]._minX && _cursorX < _mouseMapASDemo[_logic->_currMap][i]._maxX
```
bluegr: You could set a variable with the value of _mouseMapASDemo[_logic->_currMap][i], so that you can reuse it, and make this code more readable

engines/kingdom/logic.cpp
```diff
@@ -0,0 +1,605 @@
...
+	_health = 12;
+	_healthOld = 1;
+	_lastObs = false;
+	enAll();
+	_pouch = true;
+
+	initOpcodes();
+}
+
+void Logic::initOpcodes() {
```
bluegr: Is it possible to name these a bit better? GPLXXX isn't helpful

engines/kingdom/logic.cpp
```diff
@@ -0,0 +1,605 @@
...
+	_vm->_iconsClosed = false;
+	_vm->_treeLeftSta = _vm->_oldTLS;
+	_eye = _oldEye;
+	_help = _oldHelp;
+	_pouch = _oldPouch;
+}
+
+void Logic::gameHelp() {
+	if (!_vm->_gameMode) {
+		if (_vm->_userInput == 0x43C) {
```
bluegr: Again, please add an enum of these hardcoded values

engines/kingdom/constants.cpp
```diff
@@ -0,0 +1,5696 @@
...
+ *
+ */
+
+#include "kingdom/kingdom.h"
+
+namespace Kingdom {
+
+const char *_rezNames[] = {
+	"",
+	"KMAP001.lbm",
```
bluegr: These are just sequentially numbered, it makes sense to just compute them on the fly, IMHO

engines/kingdom/constants.cpp
```diff
@@ -0,0 +1,5696 @@
...
+	"KLF204.lbm",
+	"KLF205.lbm",
+	"KLF206.lbm",
+	"KLF207.lbm",
+	"KLF208.lbm",
+	"KLF209.lbm",
+};
+
+const char *_movieNames[] = {
+	"",
```
bluegr: Same here, these are just sequentially numbered, it makes sense to calculate them at run time

bluegr:
Overall, nice work! Hopefully, the MVE decoder can be implemented, as the rest of the engine seems to have been implemented.
I understand that without the movie player, the game can't be tested, right? So the implementation hasn't been tested at all?

engines/kingdom/kingdom.cpp
```diff
@@ -0,0 +1,1628 @@
...
+	const byte *data = (const byte *)surface->getPixels();
+	::Graphics::Surface *screen = g_system->lockScreen();
+	for (uint curX = 0; curX < 320; curX++) {
+		for (uint curY = 0; curY < 200; curY++) {
+			const byte *src = data + (curY * 320) + curX;
+			byte *dst = (byte *)screen->getBasePtr(curX, curY);
+			if (*src != 0xFF)
+				*dst = *src;
+		}
+	}
```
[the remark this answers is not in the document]
Strangerke: I think it's because of the transparency check (0xFF)

engines/kingdom/GPLogic4.cpp
```diff
@@ -0,0 +1,376 @@
...
+	if (_health == 0) {
+		_vm->playMovie(194);
+		_vm->_noIFScreen = true;
+		_vm->fShowPic(107);
+		_currMap = 3;
+		dsAll();
+		_pouch = false;
+		_vm->playSound(1);
+		memset(_inventory, -1, 19);
+		_statPlay = 994;
```
bluegr: This block handles character death, and is repeated throughout the game logic. It would be better to put it into its own function, e.g. playerDied().

Also, we infer that movie 194 and pic 107 are the "Player death" ones, and that screen 994 is the starting one, so it's easy to map these to an enum as well.

engines/kingdom/kingdom.cpp
Strangerke: It's now using a Common::Rect().contains(Common::Point)

engines/kingdom/constants.cpp
Strangerke: The LBM are not all sequential, only the first hundred or so. It looks to me easier to keep the current initialization

engines/kingdom/constants.cpp
Strangerke: Agreed. I'll check the MAC version just in case, and I'll switch to this

Strangerke: Arf, not the same in the mac version...

engines/kingdom/GPLogic4.cpp
Strangerke: I won't add the enum right now because I'm not able to play the video and the picture isn't self-explaining. But the code has been moved.

engines/kingdom/logic.cpp
Strangerke: Not really for the moment. The original uses 4 super large functions to handle all the hardcoded logic. I split it by statplay, but without the videos I can't really tell what it's doing...

engines/kingdom/logic.cpp
Strangerke: the value of userInput is defined in the hotspots. Some values (like 0x190) are used in A LOT of scenes with different meanings, I don't know exactly how to handle that properly

sev-:
?

PR #1644, opened 2019-05-30
BLADERUNNER: Make the Moraji encounter easier on easy

dafioram:
This extending the bomb duration by 10secs while Moraji
is chained.

It extends the bomb from blowing up by 1.1 secs after Moraji
is freed.

Finally, the blast radius is decreased so that the player may survive closer.

All of this only happens on the uncut version when playing on easy.

I picked 1.1 sec for the post-freed time because if it is longer
Moraji will continue to run in place when he reaches his destination
outside Dermo Design (If its the path where he falls after the bomb
blows).

### DIFF
--- a/engines/bladerunner/script/ai/moraji.cpp
+++ b/engines/bladerunner/script/ai/moraji.cpp
@@ -47,7 +47,10 @@ bool AIScriptMoraji::Update() {
 	 && !Game_Flag_Query(kFlagDR05BombActivated)
 	) {
 		AI_Countdown_Timer_Reset(kActorMoraji, kActorTimerAIScriptCustomTask2);
-		AI_Countdown_Timer_Start(kActorMoraji, kActorTimerAIScriptCustomTask2, 30);
+		int bombTime = 30; // Original value
+		if (_vm->_cutContent && Query_Difficulty_Level() == kGameDifficultyEasy)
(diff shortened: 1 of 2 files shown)


engines/bladerunner/script/ai/moraji.cpp
```diff
@@ -81,6 +84,8 @@ void AIScriptMoraji::TimerExpired(int timer) {
 void AIScriptMoraji::CompletedMovementTrack() {
 	if (Actor_Query_Goal_Number(kActorMoraji) == kGoalMorajiRunOut) {
 		AI_Countdown_Timer_Reset(kActorMoraji, kActorTimerAIScriptCustomTask2);
+		if (_vm->_cutContent && Query_Difficulty_Level() == kGameDifficultyEasy)
+			Delay(1100); // Allow more time for the player to get away once Moraji is freed before the bomb goes off
```
antoniou79: To be honest, I don't like this delay here. This introduces extra delay after Moraji has reached his "98" waypoint. Since in general Moraji's behavior is complicated in that scene and needs some thorough debugging (eg he never reaches his "97" waypoint), even though currently "it just works", I would prefer it if we avoided this altogether.

dafioram:
I've roomed the 1.1 sec delay hows does this look?

PR #1645, opened 2019-06-02
WINTERMUTE: Show warning on running 2.5D games

lolbot-iichan:
ScummVM's Wintermute is based on WME Lite engine, that does not support 3d entities.

https://wiki.scummvm.org/index.php/Wintermute#Detection currently states that "no mechanism for giving the user feedback about 3D-games not being supported is currently added".

This pull request is providing such a mechanism.

New GameFeature is added: GF_3D_ASSETS
This feature is set to 2 games added on
0d8834b5616aa8fc936c750ed92ef4423610d641 as both of those games use
ACTOR3DX entities.
As this feature is not supported, an additional warning is shown on game start.

"Zbang! The Game" has a non-interactive intro comics sequence that is playable with ScummVM, so I decided to show a skippable warning instead of completely disabling running such games.


bluegr:
Ideally, this warning should be shown to the user automatically, by parsing the game scripts, when a game requiring a non-supported feature is started.

Adding these flags manually is an option, but it's prone to errors. Even if we add this as it is now, an automated mechanism should be added in the future

lolbot-iichan:
> 
> 
> Ideally, this warning should be shown to the user automatically, by parsing the game scripts, when a game requiring a non-supported feature is started.
> 
> Adding these flags manually is an option, but it's prone to errors. Even if we add this as it is now, an automated mechanism should be added in the future

Done: https://github.com/scummvm/scummvm/pull/1645/files

bluegr:
Great work! This is the way to go. A simple yet effective check :)

Two minor things left:
- All messages should be enclosed in _s(“...”) functions, so that they can be translated using our translation mechanism. You should include common/translation.h, and add the file that contains strings to be translated in the POTFILES file. This is already done for Wintermute for the detection.cpp file, so check that as an example. You should include the message to be shown, and the text of the dialog options
- In the message you show, you mention that 3D character support isn’t implemented yet. A more correct message would be that 3D character support is out of ScummVM’s scope, so that is clear to the player that this feature won’t be implemented at all within ScummVM

lolbot-iichan:
> 
> 
> Great work! This is the way to go. A simple yet effective check :)
> 
> Two minor things left:
> 
>     * All messages should be enclosed in _s(“...”) functions, so that they can be translated using our translation mechanism. You should include common/translation.h, and add the file that contains strings to be translated in the POTFILES file. This is already done for Wintermute for the detection.cpp file, so check that as an example. You should include the message to be shown, and the text of the dialog options
> 
>     * In the message you show, you mention that 3D character support isn’t implemented yet. A more correct message would be that 3D character support is out of ScummVM’s scope, so that is clear to the player that this feature won’t be implemented at all within ScummVM

Thanks, done at 506ea835d15670522523f57c6714cb0c23423994
Also added _s() to another error: "This game requires PNG, JPEG and Vorbis support."

criezy:
Actually here you should have been using `_()` and not `_s()`. You can see the difference between the two on https://wiki.scummvm.org/index.php?title=Supporting_GUI_Translation

lolbot-iichan:
Thanks for pointing so quickly!
Fixed both moments.

criezy:
And thank you for fixing those so quickly!
This seems fine to me now.

bluegr:
Thanks! It looks good now. Merging

tobiatesan:
Nice, thank you!

PR #1646, opened 2019-06-03
WINTERMUTE: Support Tanya Grotter dilogy

lolbot-iichan:
Add a pair of russian games made with WME by Creative Dream Studio at 2008.

More info on added games: 
(in English, announcement at WME forum)
http://forum.dead-code.org/index.php?topic=2720

(info & walkthrough, in Russian)
http://questzone.ru/enzi/game/1221
http://questzone.ru/enzi/game/1225

Both games are now completable with ScummVM!

Only a few hacks were required:
1. one path was absolute in released game, added prefix for workaround code
2. fixed IsKeyDown to support a cheat code (Ctrl+Alt+A, works with debug_mode=true only)
3. silenced a warning that was flooding console with debug_log=true

### DIFF
--- a/engines/wintermute/base/base_keyboard_state.cpp
+++ b/engines/wintermute/base/base_keyboard_state.cpp
@@ -81,7 +81,7 @@ bool BaseKeyboardState::scCallMethod(ScScript *script, ScStack *stack, ScStack *
 	if (strcmp(name, "IsKeyDown") == 0) {
 		stack->correctParams(1);
 		ScValue *val = stack->pop();
-		int vKey;
+		uint32 vKeyCode;
 
 		if (val->_type == VAL_STRING && strlen(val->getString()) > 0) {
 			const char *str = val->getString();
@@ -89,12 +89,12 @@ bool BaseKeyboardState::scCallMethod(ScScript *script, ScStack *stack, ScStack *
 			if (temp >= 'A' && temp <= 'Z') {
 				temp += ('a' - 'A');
 			}
-			vKey = (int)temp;
+			vKeyCode = (int)temp;
 		} else {
-			vKey = val->getInt();
+			vKeyCode = vKeyToKeyCode(val->getInt());
 		}
 
-		bool isDown = _keyStates[vKeyToKeyCode(vKey)];
+		bool isDown = _keyStates[vKeyCode];
 
 		stack->pushBool(isDown);
 		return STATUS_OK;
--- a/engines/wintermute/base/base_keyboard_state.cpp
+++ b/engines/wintermute/base/base_keyboard_state.cpp
@@ -318,6 +318,50 @@ enum VKeyCodes {
 	kVkInsert     = 45,
 	kVkDelete     = 46,
 
+	kVkA          = 65, //printable
+	kVkB          = 66, //printable
+	kVkC          = 67, //printable
+	kVkD          = 68, //printable
+	kVkE          = 69, //printable
+	kVkF          = 70, //printable
+	kVkG          = 71, //printable
+	kVkH          = 72, //printable
+	kVkI          = 73, //printable
+	kVkJ          = 74, //printable
+	kVkK          = 75, //printable
+	kVkL          = 76, //printable
+	kVkM          = 77, //printable
+	kVkN          = 78, //printable
+	kVkO          = 79, //printable
+	kVkP          = 80, //printable
+	kVkQ          = 81, //printable
+	kVkR          = 82, //printable
+	kVkS          = 83, //printable
+	kVkT          = 84, //printable
+	kVkU          = 85, //printable
+	kVkV          = 86, //printable
+	kVkW          = 87, //printable
+	kVkX          = 88, //printable
+	kVkY          = 89, //printable
+	kVkZ          = 90, //printable
+
+	kVkNumpad0    = 96, //printable
+	kVkNumpad1    = 97, //printable
+	kVkNumpad2    = 98, //printable
+	kVkNumpad3    = 99, //printable
+	kVkNumpad4    = 100, //printable
+	kVkNumpad5    = 101, //printable
+	kVkNumpad6    = 102, //printable
+	kVkNumpad7    = 103, //printable
+	kVkNumpad8    = 104, //printable
+	kVkNumpad9    = 105, //printable
+	kVkMultiply   = 106, //printable
+	kVkAdd        = 107, //printable
+	kVkSeparator  = 108, //printable
+	kVkSubtract   = 109, //printable
+	kVkDecimal    = 110, //printable
+	kVkDivide     = 111, //printable
+
 	kVkF1         = 112,
 	kVkF2         = 113,
 	kVkF3         = 114,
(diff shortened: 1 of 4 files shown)


bluegr:
Are you sure about the changes to IsKeyDown()? Have you tried any other game with these changes in place? Does the keyboard work as expected?

lolbot-iichan:
> Are you sure about the changes to IsKeyDown()? Have you tried any other game with these changes in place? Does the keyboard work as expected?
IsKeyDown() is rarely used in actual games, and is mainly used with VK_SPACE to display something while space is pressed. So, I created a test game to test it: https://github.com/tobiatesan/wme_testsuite/pull/3

This function supports 2 types of syntax: IsKeyDown(string) and IsKeyDown(int).
Expected behaviour from original WME: both on pressing A or Shift+A, we should get:
* IsKeyDown("a") == true
* IsKeyDown("A") == true
* IsKeyDown(65) == true
* IsKeyDown(97) == false

So, this function is case-insensitive for IsKeyDown(string) and use a single keycode for lower and upper case for IsKeyDown(int). 97 happen to be a keycode for keypad 1, so handled a keypad too.

I have never seen a WME game checking ascii key with IsKeyDown(int), but fixed this anyway.

engines/wintermute/base/file/base_disk_file.cpp
```diff
@@ -122,6 +122,7 @@ Common::SeekableReadStream *openDiskFile(const Common::String &filename) {
 		const char* const knownPrefixes[] = { // Known absolute paths
 				"c:/windows/fonts/", // East Side Story refers to "c:\windows\fonts\framd.ttf"
 				"c:/carol6/svn/data/", // Carol Reed 6: Black Circle refers to "c:\carol6\svn\data\sprites\system\help.png"
+				"d:/engine/\0xD2\0xC32/tg_ie_080128_1005/data/", //Tanya Grotter and the Disappearing Floor
```
bluegr: Space at the beginning. Also, which file is referred to in that path?

bluegr:
Thanks! The changes look straightforward. Please, add more information about the hardcoded path.

engines/wintermute/base/scriptables/script_engine.cpp
```diff
@@ -433,7 +433,6 @@ int ScEngine::getNumScripts(int *running, int *waiting, int *persistent) {
 			numPersistent++;
 			break;
 		default:
-			warning("ScEngine::GetNumScripts - unhandled enum");
```
tobiatesan: Why?

tobiatesan:
Nice, thank you!

I'm not sure about "silencing a warning" just because it's flooding console, especially with verbose logging activated... Can you provide some more reasoning about that?

engines/wintermute/base/base_keyboard_state.cpp
```diff
@@ -81,20 +81,20 @@ bool BaseKeyboardState::scCallMethod(ScScript *script, ScStack *stack, ScStack *
 	if (strcmp(name, "IsKeyDown") == 0) {
 		stack->correctParams(1);
 		ScValue *val = stack->pop();
-		int vKey;
```
tobiatesan: Looks straightforward to me.

engines/wintermute/base/scriptables/script_engine.cpp
tobiatesan: Why? I'm not sure that "there's a lot of these messages with verbose debugging on" is a good reason.
Perhaps we need one more warning level?

bluegr: If that message is there for debug reasons only, [debugN()](https://github.com/scummvm/scummvm/blob/d2f5023ee9ec5534a9f7a4754c4f1e805c0fb605/common/debug.h#L33) could be used instead

lolbot-iichan: getNumScripts() is actually used only once, inside BaseGame::displayDebugInfo(), when displaying number of running/waiting/persistent threads:
		scrTotal = _scEngine->getNumScripts(&scrRunning, &scrWaiting, &scrPersistent);
		sprintf(str, "Running scripts: %d (r:%d w:%d p:%d)", scrTotal, scrRunning, scrWaiting, scrPersistent);
In ScummVM:
https://github.com/scummvm/scummvm/blob/d4679a4cc4318c0a581f9a823730d6adcfa0b74e/engines/wintermute/base/base_game.cpp#L3777
In WME1:
https://github.com/retrowork/Wintermute-Engine/blob/663d154b93de87f8cd13c86f34345e38a1e727d3/src/engine_core/wme_base/BGame.cpp#L5243

So, entire purpose of this switch block is to calculate some numbers to be shown on debug screen, it is not used for any game logic stuff.

Now, let's check original WME1 implementation: https://github.com/retrowork/Wintermute-Engine/blob/master/src/engine_core/wme_base/ScEngine.cpp#L528

As you can see, there was no "default:" block at all. Unhandled enum elements was the original intended behaviour. 
"default:" block and warning() were added in 4c61cf042b09a64593a8beb6e1e9d72a883a7cda for silincing a compilation warning.

I guess we should either remove this warning, because it does not provide any useful information, we already know that there are some states not listed in sprintf's line, or we can extend this debug information line printing code to include counters for all the states and handle all the enum states.

somaen: How does the explanation above map to the "tons of warnings" mentioned in the commit message?

If this is purely for debug, then at a minimum a comment, but possibly a debugN-message would be better?

somaen:
The comment that explains the intent behind IsKeyDown would actually be good to adapt into an actual code comment in the relevant block, to reduce future confusion.

engines/wintermute/base/file/base_disk_file.cpp
lolbot-iichan: Referred path "d:\engine\ТГ2\tg_ie_080128_1005\data\interface\pixel\pixel.png" contains non-ASCII characters (cyrillic letters "Т" and "Г").

What's the right way to address this path in comment?

I'd personally rather not use non-ASCII letters in the code, even inside the comments.
However, "\0xD2\0xC" looks confusingly similar to windows-style folder.
Maybe we could use some other type of escaping?

engines/wintermute/base/scriptables/script_engine.cpp
lolbot-iichan: > How does the explanation above map to the "tons of warnings" mentioned in the commit message?

Well, method getNumScripts() is called for every frame, producing a warning for every script in those states: SCRIPT_FINISHED, SCRIPT_ERROR, SCRIPT_WAITING_SCRIPT, SCRIPT_THREAD_FINISHED. For subject games there are some scripts with such states, so warnings are constantly printed to Windows console (every frame?), which noticeably slows the game down because of "tons of warnings" (have not checked, but my guess is that SCRIPT_WAITING_SCRIPT is the reason).

> If this is purely for debug, then at a minimum a comment 

I agree - this topic appeared to be way too hot for silently removing warning. =)
I will add a comment once we decide what to do.

>possibly a debugN-message would be better?

I could replace warning() with debugN() or leave warning() as is, but I fail to see a scenario, where debugN() would be useful here. Would we gain any useful information from log lines like "unhandled enum" in this context?

A. If we consider some/all states are dangerous/suspicious/should be traced for some reason - it would be better to add debugN logs on entering/leaving those states (which has nothing to do with getNumScripts() function).

B. If we need a full log of which exact script was in which exact state on which exact game tick - then a more detailed state logger should be implemented (and I'm not sure that getNumScripts() is the right place for this per-tick logger).

C. If we need to see the amount of scripts that are in those states - getNumScripts() should be extended so that more information is added to sprintf() and displayed on screen (that will make ScummVM's debug mode slightly different from original WME code, but why not).

D. If we just need to check that this `switch()` code is imperfect by knowing that there are some states that are unhandled - well, we already know it without reading any logs. We could just leave a comment inside the "default:" block.

engines/wintermute/base/file/base_disk_file.cpp
bluegr: It’s not worth adding non-English characters just for comment for a file name that’s supposed to be in English. Just add the file name in English, noting the Cyrillic characters, as you did here

engines/wintermute/base/scriptables/script_engine.cpp
bluegr: I would go with D. If the original did not handle such cases, then there is no reason to report them. So, the best scenario would be to add the warning as a debugN() statement, in case it’s needed for debugging.

In any case, a comment should be added, specifying that the default case is unhandled in the original.

tobiatesan: I second `debugN()` + comment.

lolbot-iichan: Done at 49a105b091bc29890ef5bb7557cfcb84a6a19174

engines/wintermute/base/file/base_disk_file.cpp
lolbot-iichan: Done at f2dc35ad335f2f4d2630d411058ff4b76c7f1db3

lolbot-iichan:
> 
> 
> The comment that explains the intent behind IsKeyDown would actually be good to adapt into an actual code comment in the relevant block, to reduce future confusion.

Done at fa03dea0ed653aa9ebc7b454da9c8e526433840d.
Also added a warning() for non-ASCII IsKeyDown(string) calls.

engines/wintermute/base/scriptables/script_engine.cpp
```diff
@@ -433,6 +433,15 @@ int ScEngine::getNumScripts(int *running, int *waiting, int *persistent) {
 			numPersistent++;
 			break;
 		default:
+			// Those states were not handled in original WME as well:
+			// * SCRIPT_FINISHED,
+			// * SCRIPT_ERROR,
+			// * SCRIPT_WAITING_SCRIPT,
+			// * SCRIPT_THREAD_FINISHED		
+			debugN("ScEngine::GetNumScripts - unhandled enum: %d\n", _scripts[i]->_state);
```
tobiatesan: Nice.

bluegr:
The issues identified have been addressed now. @somaen?

bluegr:
The problems mentioned have been addressed now, and it looks that there aren’t any further issues with this pull request.

Thanks for your work! Merging

PR #1647, opened 2019-06-03
WIN32: Fix leaked handles in CreateProcess

SupSuper:
Make sure to close handles created by `CreateProcess` as per https://docs.microsoft.com/en-us/windows/desktop/ProcThread/creating-processes

(closing them doesn't affect the running process)


digitall:
This looks pretty self explanatory, but would good to get a check from WIN32 porter and/or someone familar with this. @lotharsm ?

bluegr:
This change is correct, and is done according to the MSDN documentation. Thanks! Merging

PR #1648, opened 2019-06-04
WINTERMUTE: Add detection for the Kite v1.4

lolbot-iichan:
Some info for ScummVM wiki:

Game resolution: 1105x700
WME version is avaliable here: https://anatolii.itch.io/the-kite
Moddb link goes to steam that seems to serve Unity version that is not supported...

Game can be marked as completable.


bluegr:
Thanks! I’ve updated the game’s page in our wiki

Merging

PR #1649, opened 2019-06-04
 SCI: Create message workaround system 

sluicebox:
Adds a new workaround system for known broken messages and their
corresponding audio and sync resources. This replaces all special
cases in c++ and several script patches with data structures and
generic handling.

Common message bugs:
- Wrong tuple requested by game script
- Wrong tuple in message resource
- Wrong message text that exists in another record
- Missing message text
- Audio or sync resource with different tuple than message

Some notes:
- I've tested each workaround (whew!)
- The first commit is for a message script patch of mine that should have been an uninitialized parameter workaround from the start. Oops!
- The existing c++ workarounds in getRecord() wouldn't have worked on CD versions since they didn't remap audio tuples, but they happened to all be for floppy versions. Audio remapping is now handled automatically.
- I have a few obscure missing-message bugs that I couldn't justify the c++ or script patches for but I'll happily add them once this is in place
- If there's anything to be done about the KQ7 unfinished messages, this scheme should be flexible enough to use

### DIFF
--- a/engines/sci/engine/message.cpp
+++ b/engines/sci/engine/message.cpp
@@ -25,6 +25,7 @@
 #include "sci/engine/kernel.h"
 #include "sci/engine/seg_manager.h"
 #include "sci/engine/state.h"
+#include "sci/engine/workarounds.h"
 #include "sci/util.h"
 
 namespace Sci {
@@ -183,10 +184,17 @@ public:
 #endif
 
 bool MessageState::getRecord(CursorStack &stack, bool recurse, MessageRecord &record) {
-	Resource *res = g_sci->getResMan()->findResource(ResourceId(kResourceTypeMessage, stack.getModule()), false);
+	// find a workaround for the requested message and use the prescribed module
+	int module = stack.getModule();
+	MessageTuple &tuple = stack.top();
+	SciMessageWorkaroundSolution workaround = findMessageWorkaround(module, tuple.noun, tuple.verb, tuple.cond, tuple.seq);
+	if (workaround.type != MSG_WORKAROUND_NONE) {
+		module = workaround.module;
+	}
+	Resource *res = g_sci->getResMan()->findResource(ResourceId(kResourceTypeMessage, module), false);
 
 	if (!res) {
-		warning("Failed to open message resource %d", stack.getModule());
+		warning("Failed to open message resource %d", module);
 		return false;
 	}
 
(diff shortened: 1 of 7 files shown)


bluegr:
Great Scott! This is wonderful! Well done!

I’ll need to check this out more thoroughly, but, at a first glance, it’s very well made and ties in to the existing workaround logic beautifully. I’ll check it out more thoroughly this week

Kawa-oneechan:
It's a work of art if you ask me.

engines/sci/engine/workarounds.cpp
```diff
@@ -1134,4 +1136,203 @@ SciWorkaroundSolution trackOriginAndFindWorkaround(int index, const SciWorkaroun
...
+//
+// The one kind of broken message that we don't handle yet is when the audio is
+//  completely missing from the game. QFG4 has most of those. In speech-only
+//  mode this means no audio or text and in dual mode the text disappears
+//  quickly as most games use audio length for the message delay. That logic is
+//  in each game's Messager/Narrator/Talker scripts.
+static const SciMessageWorkaroundEntry messageWorkarounds[] = {
+	// game              media             language       room   mod    n    v    c   s    workaround-type          mod    n    v    c   s tlk  idx  len  text
+	// FPFP CD has several message sequences where audio and text were left out of sync - bug #10964
+	//  Some of the texts just say "Dummy Msg" and the real values are concatinated in the first record.
```
bluegr: Typo: "concatenated"

bluegr:
Thanks for your work! The approach is great, and follows the other workaround systems we got in place.
It can be merged.

Please address the merge conflicts that have shown up. I tried to fix them via github's UI, but wasn't successful :/

engines/sci/engine/message.cpp
```diff
@@ -183,10 +184,17 @@ class MessageReaderV4_MacSCI32 : public MessageReader {
 #endif
 
 bool MessageState::getRecord(CursorStack &stack, bool recurse, MessageRecord &record) {
-	Resource *res = g_sci->getResMan()->findResource(ResourceId(kResourceTypeMessage, stack.getModule()), false);
+	// find a workaround for the requested message and use the prescribed module
+	int module = stack.getModule();
+	MessageTuple &tuple = stack.top();
```
OmerMor: Why don't you keep using the MessageTuple wherever you're passing the (noun, verb, cond, seq) tuple?
It would make for clearer code IMO.

sluicebox: If you're asking why findMessageWorkaround() takes noun/verb/cond/seq parameters as opposed to just a MessageTuple, that would indeed be clearer, and that's how I first wrote it, but then workarounds.cpp has to include message.h and I didn't want to introduce that since I'm now including workarounds.h from message.cpp. It would compile, but I didn't like it, so I punted and went with the individual parameters.

Thanks for checking it out! You have absolutely found the line I grumbled about the most.

sluicebox:
@bluegr Thanks for reading my comments! I'm going to pay my git penance now and get this merged. I held back on a PR that I knew would conflict with this, then I went and blew it with a different one. =)

engines/sci/engine/message.cpp
OmerMor: You can always move MessageTuple to another file, but it's not a big deal. It's only a readability (and maintainability) issue. I only noticed this because the noun/verb/cond/seq parameters are all over the codebase which is a sign of high cohesion. A future PR perhaps.

PR #1650, opened 2019-06-04
SCI: Fix backend events treated as keyboard events

sluicebox:
This fixes several backend events being misinterpreted as SCI keyboard events.

The most noticeable is EVENT_SCREEN_CHANGED which gets translated as kSciEventKeyDown. For example, toggling fullscreen during the SQ4 intro will cause the game to prompt if you want to skip the intro. (This isn't due to the ALT+ENTER keys as the backend code claims those events.)

EventManager::getScummVMEvent() is a large function that translates an incoming Common::Event to an SciEvent. First it handles a few misc event types, then mouse events, and finally it does a lot of keyboard event mapping, but it doesn't verify that the event type is keyboard and instead assumes it must be. Non-keyboard events that the function doesn't first explicitly handle bounce around the mapping code and come out as effectively random SCI keydown and keyup events that reach the game.

Other examples are EVENT_WHEELUP, EVENT_WHEELDOWN, and the new EVENT_DROP_FILE. My guess is that all these events appeared after this SCI event code was written. Now keyboard event types are explicitly verified.

### DIFF
--- a/engines/sci/event.cpp
+++ b/engines/sci/event.cpp
@@ -253,6 +253,11 @@ SciEvent EventManager::getScummVMEvent() {
 		}
 	}
 
+	// Handle keyboard events for the rest of the function
+	if (ev.type != Common::EVENT_KEYDOWN && ev.type != Common::EVENT_KEYUP) {
+		return noEvent;
+	}
+
 	// Check for Control-Shift-D (debug console)
 	if (ev.type == Common::EVENT_KEYDOWN && ev.kbd.hasFlags(Common::KBD_CTRL | Common::KBD_SHIFT) && ev.kbd.keycode == Common::KEYCODE_d) {
 		// Open debug console


digitall:
Good catch. Change looks good, self-contained and self-explanatory. @bluegr is more expert on SCI engine internals, so should review, but should be fine to merge.

waltervn:
Sorry for going slightly off-topic here, but did you try volume up/down? I have problems with that causing keypresses pretty much everywhere, even in the GUI.

sluicebox:
I don't have that symptom and I doubt this will help, but I *think* that used to happen to me, so now that this is under my belt I'll keep an eye out.

This one just prevents events that aren't packaged as keypresses from getting treated as such by SCI, and that sounds like the volume events were already packaged as keypresses by the backend or OS. On the other hand, the unexpected events bounced around so much in that SCI method that there's no telling each's outcome, so it's possible this helps, at least in SCI.

bluegr:
Good catch! There’s also a similar check in lines 269-273, with a SCI version check, which is now redundant. Could you please remove it?

sluicebox:
That second check does something though, it prevents keyup events on earlier versions. I could change that second check from "ev.type != Common::EVENT_KEYDOWN" to "ev.type == Common::EVENT_KEYUP" if that's clearer.

bluegr:
The new check does:

	if (ev.type != Common::EVENT_KEYDOWN && ev.type != Common::EVENT_KEYUP) {
		return noEvent;
	}

and then the check below that does:

	if (ev.type != Common::EVENT_KEYDOWN && getSciVersion() < SCI_VERSION_1_1) {
		return noEvent;
	}

The second check will never evaluate to true now, so it's redundant and should be removed

bluegr:
Sorry, my bad. Yes, you are quite right. The second check is fine, but at least it should be moved together with the new check. I tried to merge these two in one statement, but it became harder to read.

criezy:
Keeping the two checks separate is fine I think, if it makes the code easier to read.
And something like `ev.type != Common::EVENT_KEYDOWN && (ev.type != Common::EVENT_KEYUP || getSciVersion() < SCI_VERSION_1_1)` would in my opinion make it harder to read.

I don't think the second check needs to be changed, but maybe the comment could be clarified along the line of
>  only sent keydown events (and no keyup events) to the interpreter

sluicebox:
I think the current order of checks makes sense:
- is this a keyboard event?
- is this the debugger hotkey?
- okay now all the SCI stuff...

Other than that I don't have an opinion about the other comments or code.

This thread is making me feel great about how long it took me to figure any of this out =)

bluegr:
Thanks! Merging

PR #1653, opened 2019-06-07
XEEN: Check bounds of stat values array, bug #10971

kurisuke:


dafioram:
Is 23 or 24 referenced somewhere else in the code?

bluegr:
@dafioram: The STAT_VALUES array is initialized with a size of 24 in resources.h:188. The patch guards against accessing OOB indices.

Not sure if these are handled correctly by ignoring them, or if the offending script(s) should be patched instead. @dreammaster is the engine maintainer, and he can shed some more light onto this

kurisuke:
24 is the size of the arrays STAT_VALUES, STAT_BONUSES and RATING_TEXT defined in engines/xeen/resources.h

Perhaps it's a good idea to refer to it directly, e.g.:
`const int idxMax = sizeof(Res.STAT_VALUES) / sizeof(int) - 1;`
?

dafioram:
It might look cleaner to loop over the array elements (using array size) and just exit the loop if the stats check is true (if stats array value > value break). 

If you wanted to keep both checks in the "for" 2nd argument you could move the index check against size() _before_ the stats check so it wouldn't need a minus 1 (idx < size() && stats array <= stats value).

Another way would be to keep things the same and just clip the index value to max of size()-1.

kurisuke:
The three arrays are used to set bonus modifiers, as well as a description string, for a certain range of values. See:
https://xeen.fandom.com/wiki/Statistics#Statistic_Cutoff_Levels

The array values of STAT_VALUES are as follows:
`$1 = {3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 25, 30, 35, 40, 50, 75, 100, 125, 150, 175, 200, 225, 250, 0}`
So we cannot use the last element check before doing the loop.

IMHO your proposed option 1 (loop until arraysize - 1, break when array value > stats value) would still be the most readable variant. I have updated the pull request accordingly.

dafioram:
Thats interesting that the last element is a zero. Does it make sense that an input stats value > 250 should be set to the last element (0)? Perhaps that is done to reject a value that is too large?

Also can you squash your two commits into one?

kurisuke:
The values from the STAT_VALUES array are never used directly. They are used for specifying ranges of values, by giving the upper limit. After the loop, the `idx` variable refers to the appropriate "bucket" or range. Then the `idx` is used to address the STAT_BONUSES and RATING_TEXT arrays to get the bonus / text values for the appropriate range.

See the values of those arrays below (taken from gdb) and compare to the table in the linked wiki page above:

`STAT_BONUSES = {-5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 20}`

`RATING_TEXT = {0x55555a71611a "Nonexistant", 0x55555a716126 "Very Poor", 0x55555a716130 "Poor", 0x55555a716135 "Very Low", 0x55555a71613e "Low", 0x55555a716142 "Averarage", 
  0x55555a71614c "Good", 0x55555a716151 "Very Good", 0x55555a71615b "High", 0x55555a716160 "Very High", 0x55555a71616a "Great", 0x55555a716170 "Super", 
  0x55555a716176 "Amazing", 0x55555a71617e "Incredible", 0x55555a716189 "Gigantic", 0x55555a716192 "Fantastic", 0x55555a71619c "Astoundig", 
  0x55555a7161a6 "Astonishing", 0x55555a7161b2 "Monumental", 0x55555a7161bd "Tremendous", 0x55555a7161c8 "Collosal", 0x55555a7161d1 "Awesome", 
  0x55555a7161d9 "AweInspiring", 0x55555a7161e6 "aUltimate"}`

(side note: I think the typos are in the original game data?)

The last range is for values > 250 (`idx == 23`), so there is no upper range and there is no more check necessary here. As for why the game designers added a "0" here, I think it's just for symmetry / padding. But still there are valid values defined at `STAT_BONUES[23]` and `RATING_TEXT[23]`.

dreammaster:
Double-checked the original, and it looks like I got the original value wrong for the last entry. It was supposed to be "65535", which I previously thought was just an invalid ending value, but was intentional to provide an upper bound. Since your commit is cleaner than trying to release a new dat file just for the one value, or hacking the resource loading code, I'll accept your solution as is.

PR #1654, opened 2019-06-07
SCI: Update SQ4 Sequel Police patch for arcade

sluicebox:
This updates the script patch for the SQ4 floppy crash when clicking inventory items on sequel police in the mall:

- Applies the patch to the arcade where the bug also exists (and in more versions)
- Fixes additional instance of the bug/crash when clicking the ATM card on police in arcade
- Fixes English PC floppy verb handling in arcade so that police can be clicked as intended
- Updates patch to not leak a stack element

This bug has a complex history but the short version is that now no versions crash and all versions display messages.

The leaked stack element didn't cause a problem in practice because the ScummVM interpreter masked it by thinking there were temp variables on the stack (I don't think there are) and tossing the extra element along with the rest when returning from the method.


bluegr:
Well done! Quite clean and elegant way of patching this. Merging

PR #1655, opened 2019-06-09
XEEN: fix result type for readUint32LE()

kurisuke:
uint32 parameters for script commands (e.g. give experience to character) are intermediate stored in a uint16 (`result`), so the 2 high bytes are cut off. This breaks late game behaviour when XP > 65535 is given. The commit corrects the type of the intermediate value.


bluegr:
Nice catch! Merging

PR #1656, opened 2019-06-09
XEEN: Add missing armor class modifiers

kurisuke:
Armor class modifiers for a material (e,g, Wooden, Leather, ..., Diamond, Obsidian) are taken from a global array `METAL_LAC` read from the resource file `xeen.ccs`.

However, the array seems to be incomplete, as only 9 values are specified, even though the code in items.cpp can refer to the full range of 22 available materials:

```
        if (item._material >= 37 && item._material <= 58) {
                strength += Res.METAL_LAC[item._material - 37];
        }
```
See for the full list of modifiers: https://xeen.fandom.com/wiki/Item_Enchantments#Materials

This commit extends `METAL_LAC` to the full length of 22 available materials, so that the correct modifiers for higher-quality materials are used.

However, the extension of the array leads to some problems:
- The resource file `xeen.ccs` needs to be regenerated due to the changed constants.
- Due to the new array sizes, the data offsets in `xeen.ccs` updated code are changed, so it is not backwards compatible. Users would need to replace the `xeen.ccs` file when updating.

I am unsure how to solve this without breaking too much, and would like to ask @dreammaster as engine maintainer for his input.


dreammaster:
Ah, you're right. The array in question was indeed too short - there are places in executable that have references to part way through the list, so I didn't realize that those needed to be included as well.

PR #1658, opened 2019-06-09
GRAPHICS: Fix ManagedSurface::copyFrom memory and rectangles handling

lephilousophe:
When calling ``ManagedSurface::copyFrom()``, ``_disposeAfterUse`` should be set to
``DisposeAfterUse::YES`` because inner surface frees up old pixels array and creates a new one.

If you don't do this, creating an empty ``ManagedSurface`` through default constructor and then using ``copyFrom()`` on it will lead to memory leak when the ``ManagedSurface`` will be destroyed because it won't call ``Surface::free()`` on it.

In addition I fix a bug on ``copyFrom()`` which should use ``addDirtyRect()`` like ``create()`` does instead of ``clearDirtyRects()``.

I don't think this will break any engine but, in doubt, the engines using ``ManagedSurface::copyFrom()`` seems to be: director, cryomni3d, xeen and access.

### DIFF
--- a/graphics/managed_surface.cpp
+++ b/graphics/managed_surface.cpp
@@ -124,6 +124,17 @@ void ManagedSurface::free() {
 	_offsetFromOwner = Common::Point(0, 0);
 }
 
+void ManagedSurface::copyFrom(const ManagedSurface &surf) {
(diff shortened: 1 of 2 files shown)


dreammaster:
Looks good to me

PR #1659, opened 2019-06-10
SCI: Fix SQ4 Big & Tall missing room message

sluicebox:
Celebrating the new message workaround system with a message bug that can't be worked around.

Now you can click Look in room 381 (CD version)


bluegr:
Haha :)
I see, in this case the actual call is missing. Right.

Thanks again for your work! Merging

PR #1662, opened 2019-06-14
SCI: Fix FPFP reappearing items, bug #10975

sluicebox:
Fixes two similar bugs where inventory items reappear after being
used due to incorrect or missing flag checks


bluegr:
Great work! Merging

PR #1665, opened 2019-06-15
WINTERMUTE: Do not optimize out alpha blits on rotation

lolbot-iichan:
This fixes [defect #10684](https://bugs.scummvm.org/ticket/10684): 1-pixel width sprite becomes a big rectangle
if rotation is applied.

Case was originally reproduced with FoxTail game for WME, then reduced
to testcase
https://github.com/tobiatesan/wme_testsuite/tree/master/time_test/packages

### DIFF
--- a/engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
+++ b/engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
@@ -433,7 +433,8 @@ bool BaseSurfaceOSystem::drawSprite(int x, int y, Rect32 *rect, Rect32 *newRect,
 	// But no checking is in place for that yet.
 
 	// Optimize by not doing alpha-blits if we lack alpha
-	if (_alphaType == Graphics::ALPHA_OPAQUE && !transform._alphaDisable) {
+	// If angle is not 0, then transparent regions are added near the corners
+	if (_alphaType == Graphics::ALPHA_OPAQUE && transform._angle == 0) {
 		transform._alphaDisable = true;
 	}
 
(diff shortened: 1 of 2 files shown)


lolbot-iichan:
I wonder if we should use ALPHA_BINARY or ALPHA_FULL here.

engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
```diff
@@ -433,7 +433,8 @@ bool BaseSurfaceOSystem::drawSprite(int x, int y, Rect32 *rect, Rect32 *newRect,
 	// But no checking is in place for that yet.
 
 	// Optimize by not doing alpha-blits if we lack alpha
-	if (_alphaType == Graphics::ALPHA_OPAQUE && !transform._alphaDisable) {
```
bluegr: Why has the check for transform._alphaDisable been removed?

lolbot-iichan: It was a `if(b!=true){b=true;}` pattern, which has equal effect to just `{b=true;}`, but was harder to read in line of several checks and I was not sure if it is performance-critical here to use `if(b!=true){b=true;}` for a single boolean.

bluegr:
I believe that ALPHA_FULL would be more appropriate in this case, since it does a normal alpha blitting. ALPHA_BINARY is the optimized “transparency or not” blitting.

Since in this case we want to disable optimized blits, ALPHA_FULL should be the right option here

lolbot-iichan:
> Since in this case we want to disable optimized blits, ALPHA_FULL should be the right option here
Ok, fixed!

bluegr:
Thanks! Squashing and merging

tobiatesan:
Thanks. Your work here is first class.

PR #1666, opened 2019-06-15
MSVC: Use determinstic UUIDs for project files

SupSuper:
Do you dread everytime you run create_project, watching as MSVC has to reload the whole solution from scratch, probably messing up your workspace state and losing your open files?

Well, dread no more! This updates create_project so that the project UUIDs are deterministic (instead of changing everytime) between regenerations, so that MSVC can more effectively cache and reload the workspace.

This is done by replacing them with MD5 hashes of the project names, which are effectively unique across the solution, but consistent everytime. I used the Win32 APIs to do it since this is only applicable to MSVC so I didn't think it's worth pulling in a whole md5.cpp for it.

### DIFF
--- a/devtools/create_project/create_project.cpp
+++ b/devtools/create_project/create_project.cpp
@@ -1496,7 +1496,7 @@ void ProjectProvider::createProject(BuildSetup &setup) {
 	}
 
 	// We also need to add the UUID of the main project file.
-	const std::string svmUUID = _uuidMap[setup.projectName] = createUUID();
+	const std::string svmUUID = _uuidMap[setup.projectName] = createUUID(setup.projectName);
 
 	createWorkspace(setup);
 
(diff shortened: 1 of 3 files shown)


devtools/create_project/create_project.cpp
```diff
@@ -1595,16 +1595,19 @@ ProjectProvider::UUIDMap ProjectProvider::createToolsUUIDMap() const {
...
 }
 
+const int kUUIDLen = 16;
+
 std::string ProjectProvider::createUUID() const {
 #ifdef USE_WIN32_API
 	UUID uuid;
-	if (UuidCreate(&uuid) != RPC_S_OK)
+	RPC_STATUS status = UuidCreateSequential(&uuid);
+	if (status != RPC_S_OK && status != RPC_S_UUID_LOCAL_ONLY)
```
SupSuper: `UuidCreateSequential` is faster since we don't need cryptographic security

henke37:
If you want to be picky you should set the appropriate version bits and obey the UUID format. Which requires hash based UUIDs to be rooted in another UUID.

SupSuper:
@henke37 Better?

henke37:
Yeah, that should obey the rules.

bluegr:
Nice, that really helps when recreating the MSVC project files! Merging

PR #1667, opened 2019-06-16
SCI: Fix SQ4 Monolith Burger door message, bug #10976

sluicebox:
This message bug requires a script patch because the wrong message it asks for is a valid message which it correctly asks for on a different line.


bluegr:
Nice! Thanks for your work

PR #1669, opened 2019-06-17
SCI: Fix SQ4 Sequel police electronics store bugs

sluicebox:
"How's QFG4 going?"
"Let's just say even SQ4 mall bugs are more fun..."

This fixes lockup, animation, and flag bugs when hiding in the electronics store, bug #10977.

Whew! This one has been sitting in the queue for a while, but the hole kept getting deeper. It looks like this entire sequence was never tested on CD, since it locks up immediately and has message typos and animation and flag bugs after that. I never knew you could do this stuff as it takes an inefficient escape route, which I guess is how testers missed it.

These fixes make a dozen lines of audio available for the first time, including a "new" death message in front of monolith burger.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -125,6 +125,7 @@ static const char *const selectorNameTable[] = {
 	"looper",       // Space Quest 4
 	"nMsgType",     // Space Quest 4
 	"doVerb",       // Space Quest 4
+	"setRegions",   // Space Quest 4
 	"loop",         // Laura Bow 1 Colonel's Bequest, QFG4
 	"setLoop",      // Laura Bow 1 Colonel's Bequest, QFG4
 	"ignoreActors", // Laura Bow 1 Colonel's Bequest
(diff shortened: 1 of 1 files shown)


engines/sci/engine/script_patches.cpp
```diff
@@ -12304,6 +12306,171 @@ static const uint16 sq4CdPatchBikerTimepodMessage[] = {
...
+//  clears flag 23 and doesn't test 22 while room 390 doesn't clear 23 and does
+//  test 22. The result is that room 370 doesn't see that you've escaped the
+//  Skate-O-Rama and assumes the police are still pursuing from room 400.
+//
+// We fix this by only running sp1Squeeze in room 370 when coming from room 400.
+//  This is equivalent to adding the missing flag 22 check but can be done in
+//  the available bytes. To make room we use the clear-flag method's return
+//  value, which is the previous flag value, and only test if the previous room
+//  number is even, which only room 400 is. This patch is split into two parts
+//  to surround a single bnt instruction whose size changed between versions.
```
bluegr: Sorry, I'm confused - can't you create a single patch in this case?

sluicebox: The bnt instruction that's in the middle has a 16-bit operand in some versions and an 8-bit in the other (even though it's the same value) and there's not a mechanism for "skip 2 or 3 bytes".

I updated the comment to be clearer that it's the operand size in the instruction.

bluegr: That makes sense. Thanks!

bluegr:
Well done! Quite an edge case, and it looks like very few people actually stumbled upon these bugs

PR #1670, opened 2019-06-17
SCI32: Fix QFG4CD Domovoi inn events, bug #10978

sluicebox:
This fixes a frequently encountered bug that subtly leaves the game in an unwinnable state. It's a regression in the CD version.

This patch does not change the game's behavior unless it's in the state where you're stuck, in which case the floppy behavior is restored and the necessary event still occurs. It took a while to work through all the surrounding events and prove that, hence the big comment.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -11377,6 +11377,51 @@ static const uint16 qfg4RunesPuzzlePatch2[] = {
 	PATCH_END
 };
 
+// The Domovoi in room 320 has a complex bug in the CD version. If you don't
+//  talk to him before Bella wakes you up then you can't get the doll and the
+//  game can't be completed. The event logic was changed in the floppy patch and
+//  again in the CD version, which introduced the bug. Working backwards...
+//
+// To get the doll from the inn's cabinet:
+// - It must be late at night
+// - Inn event 15 has occurred or is occurring
+//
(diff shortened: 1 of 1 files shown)


engines/sci/engine/script_patches.cpp
```diff
@@ -11377,6 +11377,51 @@ static const uint16 qfg4RunesPuzzlePatch2[] = {
...
+//  only effect is to suppress a mandatory event, and so it is safe to remove.
+//
+// Applies to: English CD
+// Responsible methods: rm320:init, heroTeller:respond
+// Fixes bug: #10978
+static const uint16 qfg4DomovoiInnSignature[] = {
+	SIG_MAGICDWORD,
+	0x78,                               // push1
+	0x38, SIG_UINT16(0x0088),           // pushi 0088
+	0x45, 0x04, SIG_UINT16(0x0002),     // callb proc0_4 [ is flag 136 set? ]
```
bluegr: So when is flag 136 set?

sluicebox: I made a few edits, is that clearer? Thanks for reading these comments so thoroughly, this one has been sitting around for a while because I knew writing it up would take some time and patience.

bluegr: Clear, thanks! :)

bluegr:
Well done! This is one of the most obscure and well known bugs in QFG4. Great work, once again! :)

PR #1677, opened 2019-06-17
SLUDGE: don't return failure in playMovie [DON'T MERGE]

dos1:
While playMovie is not implemented yet, it seems a better idea to just skip playing a movie instead of crashing the whole game.

### DIFF
--- a/engines/sludge/movie.cpp
+++ b/engines/sludge/movie.cpp
@@ -953,7 +953,7 @@ int playMovie(int fileNumber) {
 
 	setPixelCoords(false);
 #endif
-	return 1;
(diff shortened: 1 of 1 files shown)


dos1:
Actually nvm that, judging from the error message it's likely a bug somewhere in error handling:
```
Calling: CALLIT

Call of non-function!
```
Judging to OpenSLUDGE docs, game should be able to handle playMovie failure itself.

bluegr:
So, this should be closed then?

dos1:
Heh, yeah, probably :D Haven't looked closed at that issue yet.

PR #1678, opened 2019-06-17
SLUDGE: fix crash in FloorManager::setFloorNull

dos1:
numPolygons wasn't getting reset to 0, so the next invocation was trying to delete non-existent polygons.


bluegr:
Indeed, this looks to be an oversight. Nice catch!

PR #1679, opened 2019-06-17
SLUDGE: fix data file handling with custom encoding setting

dos1:
The check was inverted - it failed when everything was alright and succeeded on error...


bluegr:
Nice catch! Merging

PR #1680, opened 2019-06-17
SLUDGE: fix background position not taking camera into account

dos1:
When a game sets camera position with aimCamera, the background should move together with all the objects above it.


bluegr:
Since you are a contributor of the original sludge engine, it’s great that you found this bug!

Thanks again for your work! Merging

dos1:
I'm basically running my old SLUDGE game (https://tremendouscorp.com/) and fixing stuff that comes up :D

PR #1685, opened 2019-06-18
WINTERMUTE: Fix Split() method of ext_string

lolbot-iichan:
It's hard to believe, but this fixes [bug #10432](https://bugs.scummvm.org/ticket/10432) WME Carol Reed
Mysteries hint system not working.

Carol Reed hint system happen to heavily use line.Split(";") results,
which were wrong by 1 byte (delimeter was appended to result while it
shouldn't be).

I started with decompiling Carol Reed source code, reproducing issue
with a stand-alone test project (which worked with WME and showed [null]
with ScummVM). then minimized it to a minimal testcase:

```
var line = new String("New Goal;Visit Christina at the Art
Museum;1;0;S;;");
var ar = line.Split(";");
if((ar[0] == "New Goal")) { var g = ar[1]; }
```

### DIFF
--- a/engines/wintermute/base/scriptables/script_ext_string.cpp
+++ b/engines/wintermute/base/scriptables/script_ext_string.cpp
@@ -303,7 +303,7 @@ bool SXString::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
 			uint32 ch = (i == str.size()) ? '\0' : str[i];
 			if (ch =='\0' || delims.contains(ch)) {
 				if (i != start) {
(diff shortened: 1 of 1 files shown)


bluegr:
The change is correct, as the code in the commit is the same as the original:
https://bitbucket.org/MnemonicWME/wmelite/src/fe84bd13b78b06ba7ecd9b4739eb622d95ebf92b/src/SXString.cpp#lines-297

This code has been introduced in commit d136e20d6e1851b1261db83787fea9d5fb42cd89

Thanks for your work! Merging

PR #1688, opened 2019-06-19
WINTERMUTE: Add workaround for sprite glitch at Rosemary game

lolbot-iichan:
This kind of fixes https://bugs.scummvm.org/ticket/6572 WME: Rosemary -
Sprite flaw on going upwards

Some Rosemary sprites have non-fully transparent background pixels,
sprites of walking up has LOTS of them. Testing Rosemary walking
character in a test WME game demonstrates some glitches with original
WME as well. However, character sprite is downscaled most of time, and
with original WME it isn't that seen when sprite is downscaled.

This pull request provides a workaround for this case instead of
changing downscaling algorithms:
* if game ID is "rosemary"
* if loaded sprite is from "actors" folder
* if bytesPerPixel is 4
* if some pixel's alpha is between 1 and 15
* set this pixel's alpha to 0

### DIFF
--- a/engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
+++ b/engines/wintermute/base/gfx/osystem/base_surface_osystem.cpp
@@ -26,6 +26,7 @@
  * Copyright (c) 2011 Jan Nedoma
  */
 
+#include "engines/wintermute/base/base_engine.h"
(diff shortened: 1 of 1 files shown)


bluegr:
Thanks for your work! That glitch is fixed now. Merging

PR #1690, opened 2019-06-21
BLADERUNNER: Skip videos with Esc key only

tsoliman:
Previously any keyup event skipped the intro videos (including Shift)

### DIFF
--- a/engines/bladerunner/bladerunner.cpp
+++ b/engines/bladerunner/bladerunner.cpp
@@ -1164,7 +1164,7 @@ void BladeRunnerEngine::handleKeyUp(Common::Event &event) {
 		return;
 	}
 
-	if (_vqaIsPlaying && event.kbd.keycode == Common::KEYCODE_ESCAPE) {
+	if (_vqaIsPlaying && (event.kbd.keycode == Common::KEYCODE_ESCAPE || (event.kbd.keycode == Common::KEYCODE_RETURN && event.kbd.flags == 0))) {
 		_vqaStopIsRequested = true;
 		_vqaIsPlaying = false;
 
--- a/engines/bladerunner/bladerunner.cpp
+++ b/engines/bladerunner/bladerunner.cpp
@@ -1164,7 +1164,7 @@ void BladeRunnerEngine::handleKeyUp(Common::Event &event) {
 		return;
 	}
 
-	if (_vqaIsPlaying) {
+	if (_vqaIsPlaying && event.kbd.keycode == Common::KEYCODE_ESCAPE) {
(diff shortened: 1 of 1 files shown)


dafioram:
Thats how the original does it.

engines/bladerunner/bladerunner.cpp
```diff
@@ -1164,7 +1164,7 @@ void BladeRunnerEngine::handleKeyUp(Common::Event &event) {
 		return;
 	}
 
-	if (_vqaIsPlaying) {
+	if (_vqaIsPlaying && event.kbd.keycode == Common::KEYCODE_ESCAPE) {
```
peterkohaut: please, can you add "KEYCODE_RETURN" as well as that is used for different skips

bluegr:
I agree that it's annoying to skip videos by pressing any key. Pressing just escape or enter is more straightforward and in line with the way most games handle this

tsoliman:
Not at home right now, but it'll be interesting to see how alt-enter is handled (toggle fullscreen) once I allow enter/return to skip. 

Before this change, I start scummvm, start blade runner, go full screen (alt-enter), accidentally skip.

engines/bladerunner/bladerunner.cpp
tsoliman: I added that and I also added a check to make sure Alt isn't pressed.

tsoliman:
I'm not happy with how the Return skip is handled on keyup.

If you start the game, then go fullscreen, then the movie will skip or not skip depending on the release order. If you release Alt and then Return, you'll skip (bad). If you release Return and then Alt, then you won't skip (good).

This is an edge case so maybe it's not a big deal. 🤷

(I am not sure I understand the consequences of switching the handling to keydown, so I won't do so)

sluicebox:
Hmm... alt+enter keydown and keyups shouldn't be hitting the engine in the first place. See SdlGraphicsManager::notifyEvent, which claims both events.

I recently fixed the SCI engine responding to alt+enter, and it turned out that it was the EVENT_SCREEN_CHANGED event that gets generated by toggling fullscreen that was being misinterpreted by the engine as a keyboard event. The backend was correctly claiming the alt+enter keydown/keyup events, and I couldn't find a way to get that to fail, so I'd be surprised if they are reaching the Bladerunner code. (Unless this engine has a different input interface, I don't know)

ALT+S (screenshot) is implemented in the same way. Since that doesn't involve toggling fullscreen, it's easier to debug.

tsoliman:
I did some debugging .. keydown-return never reaches the engine (because fullscreening consumes it?) but keyup-return does.

PR #1694, opened 2019-06-21
SDL: Include cursorman.h if OpenGL is disabled

lotharsm:
Currently, the SDL backend only includes cursorman.h if we use OpenGL.
However, the engines that use CursorMan as well as main.ccp and
launcher.cpp include cursorman.h without having OpenGL enabled.

Since cursorman.h doesn't seem to include anything specific to OpenGL
and to improve consistency, this commit disables the check for OpenGL
before including graphics/cursorman.h in the SDL backend.


criezy:
in this file CursorMan is used only if we use OpenGL (since the code that uses CursorMan is in a `#ifdef USE_OPENGL` block itself). I don't see the point of including a file that is not needed when we don't use OpenGL. Usually we try to go the other way and minimize file inclusion to speed up compilations.

PR #1696, opened 2019-06-21
BLADERUNNER: Skip movies on keydown not keyup

tsoliman:
This doesn't suffer from key-repeating.

### DIFF
--- a/engines/bladerunner/bladerunner.cpp
+++ b/engines/bladerunner/bladerunner.cpp
@@ -1218,7 +1218,7 @@ void BladeRunnerEngine::handleKeyDown(Common::Event &event) {
 		return;
 	}
 
-	if (_vqaIsPlaying && (event.kbd.keycode == Common::KEYCODE_ESCAPE || event.kbd.keycode == Common::KEYCODE_RETURN)) {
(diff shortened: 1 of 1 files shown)


bluegr:
This is strange. Why do you want to do this on keydown? Keyup is the expected place for this handling

engines/bladerunner/bladerunner.cpp
```diff
@@ -1225,6 +1218,13 @@ void BladeRunnerEngine::handleKeyDown(Common::Event &event) {
 		return;
 	}
 
+	if (_vqaIsPlaying && !event.kbdRepeat && (event.kbd.keycode == Common::KEYCODE_ESCAPE || event.kbd.keycode == Common::KEYCODE_RETURN)) {
```
bluegr: It would be more preferable to leave this code in the keydown event handling, and add the kbdRepeat check there

peterkohaut: sorry, i haven seen this comment :(

tsoliman: @bluegr I'm very confused. You're commenting in code that's in handlekeydown and saying "leave it in keydown"?

tsoliman:
> This is strange. Why do you want to do this on keydown? Keyup is the expected place for this handling

Why? Keydown is the more appropriate and natural place for handing anything that's actuated by keys. Take a look at the full screen toggle handling or the screenshot handling.

peterkohaut:
that was the original idea - to have it in keydown, but due to sdl bug it was not possible

PR #1698, opened 2019-06-22
COMMON & WINTERMUTE: Use non-1252 for 125X games

lolbot-iichan:
TLDR: this fixes some known Russian and Czech games, but breaks savegame compatibility.

This all started with me just adding 2 small free games to detection tables, but then I noticed that Russian versions of those games shows complete rubbish instead of text. Being a Russian, I know very well that those exact kind of bugs are common when locale is set to 1252 instead of 1251. So, I began collecting data...

I have bought/downloaded ALL games, listed in Wintermute detection table, that have non-1252 languages. I have unpacked those games, looked inside font files and string files and then tested them on ScummVM-dayly, native WinterMute and my ScummVM branch. And here is the results:

1. Some games don't use True Type fonts: 5ma (CZ_CZE, PL_POL), deadcity (CZ_CZE, RU_RUS), grotter1 (RUS), grotter2 (RUS), thekite (RUS), tib (RUS). Those are NOT affected, because no charset conversion is needed.

2. Some games use True Type fonts with CHARSET=1 and UTF-8 resources: dirtysplit (CZ_CZE), reversion1 (ZH_CNA, RU_RUS, LV_LAT), reversion2 (ZH_CNA, RU_RUS), twc (multilang). Those are NOT affected, because UTF8 case is supported inside ScummVM's Wintermute.

3. Some games use True Type fonts with CHARSET=1 and 125X resources: bookofgron (RU_RUS), carolreed4 (RU_RUS), kulivocko (CZ_CZE), reversion1 (PL_POL). Those games WERE affected: kulivocko and reversion1 showed wrong characters for 1250-specific chars if custom TTF file was missing (so, developer provided a workaround in font); bookofgron & carolreed4 showed rubbish instead of Russian text regardless font files. Now this is fixed.

4. Several games  use True Type fonts with CHARSET=10, which means "cyrillic": driller (RU_RUS), oknytt (RU_RUS), onehelluvaday (RU_RUS). Those games WERE affected: they showed rubbish instead of Russian text. Now this is fixed.

There are no known games with other CHARSET values, however I added not just 1250 and 1251, but also 1253, 1255 and 1257 just in case, because those languages are listed in Common::Language and have simple 1-byte encodings. Adding JOHAB / Shift-JIS / GB2312 / etc seems pointless at this moment as there is no known games with those.


ccawley2011:
Wouldn't it be better to use `Common::convertToU32String()` for codepage conversion?

lolbot-iichan:
> Wouldn't it be better to use `Common::convertToU32String()` for codepage conversion?

That's funny, I didn't notice `typedef Common::U32String WideString` and thought they are not compatible. 

Sure, I will rework this.

lolbot-iichan:
Implemented Common::convertUtf32ToUtf8(), code is based on Wintermute::ConvertUTF32toUTF8()

Implemented Common::convertFromU32String(), using convertUtf32ToUtf8 for UTF8 or lookup through
conversion table for single-byte encodings

engines/wintermute/detection_tables.h
```diff
@@ -376,7 +378,15 @@ static const WMEGameDescription gameDescriptions[] = {
 	WME_WINENTRY("dreaming", "",
 		WME_ENTRY1s("data.dcp", "4af26d97ea063fc1277ce30ae431de90", 8804073), Common::EN_ANY, ADGF_UNSTABLE, LATEST_VERSION),
 
-	// Dreamcat
+	// The Driller Incident (English)
```
bluegr: Please fix spacing here (that was me, I tried to fix the merge conflicts, and I missed the spacing)

common/ustr.cpp
```diff
@@ -461,6 +461,57 @@ U32String convertUtf8ToUtf32(const String &str) {
 	return u32str;
 }
 
+// This is a quick and dirty converter.
```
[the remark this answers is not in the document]
lolbot-iichan: 1. Original version is more effective. This version features `buffer = (char)(...) + buffer;` pattern, which causes unnecessary memory copying and reallocations, while original version works with raw bytes.
2. Original version had an option for performing strict/nonstrict conversion for 0xD800...0xDFFF interval.
3. Original version returned a result code of conversion. This version just insert '�' on characters that does not fit in 4 bytes and does not inform caller on any errors.

I guess, I can add those as "//TODO" since I already listed it?

engines/wintermute/detection_tables.h
lolbot-iichan: Fixed at 363be76bef8cf852823ed141fc83b05fc76af893

common/ustr.cpp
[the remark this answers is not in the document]
bluegr: Yes, please add these points as a comment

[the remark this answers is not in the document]
lolbot-iichan: Added at 15df085197b6705cd02e4c5a9902d6934fdd3ac9.
Also reviewed convertUtf8ToUtf32() and wrote TODOs as well.

common/ustr.cpp
```diff
@@ -428,7 +428,13 @@ void U32String::initWithCStr(const char *str, uint32 len) {
 	_str[len] = 0;
 }
 
-// This is a quick and dirty converter.
+// //TODO: This is a quick and dirty converter. Refactoring needed:
+// 1. This version is unsafe! There are no checks for end of buffer
+//    near i++ operations.
+// 2. Original version have an option for performing strict / nonstrict
```
bluegr: Nitpicking: “Original version has an option...”

common/ustr.cpp
```diff
@@ -461,6 +467,64 @@ U32String convertUtf8ToUtf32(const String &str) {
 	return u32str;
 }
 
+// //TODO: This is a quick and dirty converter. Refactoring needed:
+// 1. Original version is more effective.
+//    This version features buffer = (char)(...) + buffer; pattern that causes
+//    unnecessary copying and reallocations, original code works with raw bytes
+// 2. Original version have an option for performing strict / nonstrict
```
bluegr: Nitpicking: “Original version has an option...”

common/ustr.cpp
lolbot-iichan: Thanks for pointing those. I'm not that good with English. Fixed at 63111560216a00afe31a7cee71062f22eda89a43

common/ustr.cpp
lolbot-iichan: Fixed at 6311156

bluegr:
Merging, as mentioned. Thanks for your work!

PR #1700, opened 2019-06-23
SCI32: Fix QFG4 Wraith fan-patch, bug #10711

sluicebox:
This excludes NRS patches 53.HEP/SCR from QFG4CD since they introduce a lockup bug with wraiths. (Occurs when game-time global 88 is >= $8000.) The GOG release includes every NRS patch except this one.

The only consequence to excluding is that it can break existing save games, but since this game is tagged as unstable, and this only affects save games using this patch that also have the wraith script currently loaded, I think that's acceptable.

I think we should support the rest of the NRS patches in this game since the GOG release includes them, making them effectively a real version, and since this game is so buggy that fans are likely to have already applied them.

By "support" I just mean:
	- Don't enable the "fan-made patch" warning
	- Do our best to make our script patches compatible

I've looked at each patch file and the changes are so minor that I don't think there will be more problems, but if there are we can exclude them too. This patch was just missing an Abs() call that the others have. 

Making script patches compatible just means minding differences in offsets and operand sizes, which needs to be done anyway for multi-version patches, and this only applies when patching of the 11 NRS scripts. I already did that for the Ad Avis room and it's hard to imagine anything being harder than that. (I HOPE)

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -11708,6 +11708,37 @@ static const uint16 qfg4FighterSpearPatch[] = {
 	PATCH_END
 };
 
+// The NRS fan-patch for wraiths has a bug which locks up the game. This occurs
+//  when a wraith initializes while game time is greater than $7fff. The patch
+//  throttles wraith:doit to execute no more than once per game tick, which it
+//  does by storing the previous game time in a new local variable whose initial
+//  value is zero. This technique is used in several patches but this one is
+//  missing a call to Abs that the others have. Once game time reaches $8000 or
+//  greater, the signed less-than test will always pass when the local variable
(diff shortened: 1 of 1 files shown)


sluicebox:
Alternative: we could patch the script, it looks like just a one-byte patch to change a signed less-than to an unsigned-less-than would do it.

bluegr:
> Alternative: we could patch the script, it looks like just a one-byte patch to change a signed less-than to an unsigned-less-than would do it.

Fixing the patch is better than excluding it altogether, so please go with that option

sluicebox:
Great, I'm happy to do it the fun way. =)

This immediately fixes the locked up saved game in the bug tracker, upon loading the wraith will animate and the game will proceed.

bluegr:
Great! Thanks :)

PR #1704, opened 2019-06-23
BLADERUNNER: Update POTFILES

lotharsm:
The strings "Sitcom mode" and "Shorty mode" are included in
detection.cpp, so we add this file to the POTFILES.


engines/bladerunner/POTFILES
```diff
@@ -1 +1,2 @@
 engines/bladerunner/bladerunner.cpp
+engines/bladerunner/detection.cpp
```
criezy: You are missing a end of line on this last line.
Our build system does not like that and would concatenate the line with the first line of the next POTFILES to be processed, resulting in an error.

criezy:
Thank you. This indeed needs to be added. So just fix the missing EOL on the last line and this can be merged.

lotharsm:
Thanks, fixed :)

PR #1705, opened 2019-06-23
WIN32: Use HINSTANCE for OSystem_Win32::openUrl

lotharsm:
This implementation now matches previous ShellExecute() calls, so we are now consistent in that regard. It also silences a warning in Mingw complaining about a type mismatch.


bluegr:
Nice :) this matches the other ShellExecute() calls, as mentioned. Merging

PR #1706, opened 2019-06-23
Improve iOS inputs

criezy:
I am submitting this as a Pull Request to see if others have feedback before I merge this. I am in particular looking for feedback from iPhone users since I only have an iPad, and was only able to test iPhone using the simulator. I am planning to leave it open for one week and then merge.

The first commit adds an input accessory view to the keyboard, which allows to add one row of buttons. <del>Due to widths limitation, less buttons are added on iPhone than iPad.</del> _A scrollview is used so that we can have more buttons than fits the screen._
* iPad _and iPhone_
  * Main Menu button
  * Escape key
  * _Tab key_
  * Return key
  * F1 to F12 keys
  * Arrow keys
* <del>iPhone</del>
  * <del>Return key</del>
  * <del>F1 to F10 keys</del>

Here is a screenshot for iPad:
ipad_keyboard

The second commit removes the mapping of some keys to the return and F1 to F10 keys since those are now available in the keyboard. In particular this removes the mapping of the `dash` character to `F1` and fixes [bug #10314: LSL7: Unable to open Xqwsts’ locker on iOS](https://bugs.scummvm.org/ticket/10314).

The third commit change the gesture to show and hide the keyboard from a three finger swipe up/down to a pinch in/out. This allows to map the three finger swipes gesture to arrow keys instead. This fixes [bug #5914: swipe to simulate arrow keys doesn't work](https://bugs.scummvm.org/ticket/5914). One finger swipes are not available as this conflicts with moving the cursor in touchpad mode.

### DIFF
--- a/backends/platform/ios7/ios7_common.h
+++ b/backends/platform/ios7/ios7_common.h
@@ -40,7 +40,8 @@ enum InputEvent {
 	kInputApplicationSuspended,
 	kInputApplicationResumed,
 	kInputSwipe,
-	kInputTap
+	kInputTap,
+	kInputMainMenu
 };
 
 enum ScreenOrientation {
--- a/backends/platform/ios7/ios7_keyboard.h
+++ b/backends/platform/ios7/ios7_keyboard.h
@@ -35,6 +35,7 @@
 - (UITextView *)inputView;
 - (void)setInputDelegate:(id)delegate;
 - (void)handleKeyPress:(unichar)c;
+- (void)handleMainMenuKey;
 
 - (void)showKeyboard;
 - (void)hideKeyboard;
--- a/backends/platform/ios7/ios7_keyboard.h
+++ b/backends/platform/ios7/ios7_keyboard.h
@@ -26,9 +26,11 @@
 #include <UIKit/UIKit.h>
 #include <UIKit/UITextView.h>
 
+@class TextInputHandler;
+
 @interface SoftKeyboard : UIView<UITextViewDelegate> {
 	id inputDelegate;
-	UITextView *inputView;
+	TextInputHandler *inputView;
 }
 
 - (id)initWithFrame:(CGRect)frame;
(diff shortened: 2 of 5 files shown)


bluegr:
Nice! I like the fact that you’ve removed the hardcoded special button assignments.

Pinch works fine for the main menu. However, why would you want to map three finger swipes to arrow keys? Won’t two finger swipes work for these?

criezy:
I have now added a scrollview so that we can have all the buttons on iPhone as well. And I have also added a button for the Tab key as somebody mentioned on the forum that it is needed to display the inventory in KQ1.

> However, why would you want to map three finger swipes to arrow keys? Won’t two finger swipes work for these?

Those are already used for other actions:
* two fingers swipe up: enable/disable mouse click and drag mode
* two fingers swipe down: show the main menu
* two fingers swipe right: enable/disable touchpad mode

criezy:
OK. It's time to see if buildbot likes this.

Purduecoz:
Any way to turn off the input accessory view if a physical keyboard is attached?  I am finding the input accessory view blocks a few rows of the bottom of the game output when my iPad is in landscape mode

criezy:
Good question. Since I don't have any hardware keyboard for iPad I didn't think of that. So currently there is no way to do it. We will need to add some code to detect if an external keyboard is in use. I will take a look (hopefully that is something I should be able to test with the iOS simulator).

Purduecoz:
Not sure if you have any old generic Bluetooth keyboard - if so, you’ll be able to pair it with the iPad. Really appreciate you taking a look at this. Pre-input accessory view everything was actually working great with a bluetooth keyboard... but I went with the type cover because the Scummvm’s iOS software keyboard functionality wasn’t cutting it for AGI/SCI. Might not have had to do that had your commit been in a few months sooner! <grin>, but the type cover really does provide a great iPad experience with ScummVM.

________________________________

criezy:
I forgot I had a bluetooth keyboard for my computer  :-P
I have now pushed a change that should hide the input accessory view when a physical keyboard is connected. It only works with iOS 9 or above though. There is no public API to detect if a keyboard is connected, so I had to use one of the hack floating around (the only one that worked for me). There would be ways to make it work with older iOS versions, but I decided one hack was enough for today.

Purduecoz:
Thank you!  It's almost perfect... except the return button on the physical keyboard doesn't do anything, at least in the AGI and SCI interpreters.  The parser line just sits there blinking at me.  I'm thinking this may be a side effect of commit e5709ed?

criezy:
Right. I should have been a bit more careful when I removed all those mapping. Mapping `LF` to the return key is indeed needed. Sorry for this. I have now added that mapping back.

Purduecoz:
Works brilliantly.  Thank you for your contributions to the project!

PR #1707, opened 2019-06-24
WIN32: Make use of the window handle when calling ShellExecute.

henke37:
It's generally good practice to provide the window handle when you have one.


SupSuper:
Looks good.

bluegr:
Thanks! Merging

PR #1710, opened 2019-06-25
SCI32: Fix Scaler assert on 1 pixel wide Rects

sluicebox:
This fixes a bug in the SCI32 cel scaling code that fails an assertion in debug builds in the swamp on QFG4 floppy,  bug #10765.

Floppy versions are crashing when drawing cel 0 from any loop in view 535 of hands rising from the swamp.

The cel that's crashing is a 1x1 transparent pixel. Each loop starts with a cel like that.

The reason this has only been happening in floppy is GameFeature::hasEmptyScaleDrawHack():

```
inline bool hasEmptyScaleDrawHack() const {
	// Yes: KQ7 (all), PQ4CD, QFG4CD, SQ6, Phant1
	// No: All SCI2, all SCI3, GK2, LSL6hires, PQ:SWAT, Torin
	// Unknown: Hoyle5, MGDX, Shivers
	const SciGameId &gid = g_sci->getGameId();
	return getSciVersion() > SCI_VERSION_2 &&
		getSciVersion() < SCI_VERSION_2_1_LATE &&
		gid != GID_LSL6HIRES &&
		gid != GID_GK2 &&
		gid != GID_PQSWAT &&
		gid != GID_TORIN;
}

```
This returns true for QFG4 CD and false for floppy. If true then scaling is skipped if the target rectangle's left and right or top and bottom are equal, but it's that condition that the scaling code asserts on, and so hasEmptyScaleDrawHack has masked the bug from the CD version.

From the SCALER_Scale constructor:
```

	_minX(targetRect.left),
	_maxX(targetRect.right - 1),
	
	...
	assert(_minX <= _maxX);
```

Subtracting one makes the assertion fail when left == right, which is the case for 1 pixel wide BR-exclusive Rects like ScummVM's. Subtracting one would be correct for BR-inclusive Rects like Sierra's.

_minX and _maxX are only included in debug builds and only used for this and two other asserts.

### DIFF
--- a/engines/sci/graphics/celobj32.cpp
+++ b/engines/sci/graphics/celobj32.cpp
@@ -168,7 +168,7 @@ struct SCALER_Scale {
 	_row(nullptr),
 #ifndef NDEBUG
 	_minX(targetRect.left),
-	_maxX(targetRect.right - 1),
+	_maxX(targetRect.right),
 #endif
 	// The maximum width of the scaled object may not be as wide as the source
 	// data it requires if downscaling, so just always make the reader


bluegr:
These assertions are there for a reason: avoid bad memory access. Changing the check like this would let the problematic case slip through. Perhaps it would be better to add an exception for 1x1 sprites only

sluicebox:
I got fooled, It's not that a 1x1 (or 1xAnything) cel is a direct problem, but rather that the 1x1 cel's target rectangle has been calculated as 0x0 and the scaler code doesn't expect 0xAnything. (Why does SCALER_Scale only validate left/right and not top/bottom?)

I don't understand why there's a SCI feature check for whether or not to proceed with scaling to an empty target rectangle:

```
void CelObj::scaleDrawNoMD(Buffer &target, const Ratio &scaleX, const Ratio &scaleY, const Common::Rect &targetRect, const Common::Point &scaledPosition) const {
	// In SSCI the checks are > because their rects are BR-inclusive; our checks
	// are >= because our rects are BR-exclusive
	if (g_sci->_features->hasEmptyScaleDrawHack() &&
		(targetRect.left >= targetRect.right ||
		 targetRect.top >= targetRect.bottom)) {
		return;
	}
```

I can understand that SSCI added a return-early check at some point because they realized scaling to an empty target was an expensive no-op, but I don't know why we would reproduce both code paths unless there's a tangible effect to scaling to an empty rectangle. (But what would that be? And if it existed, you haven't seen it, because that path asserts)

Is there a reason to not just delete hasEmptyScaleDrawHack() and always skip scaling to empty rectangles?

```
if ((targetRect.left >= targetRect.right || targetRect.top >= targetRect.bottom)) {
	return;
}
```

bluegr:
I agree, it just doesn’t make any sense to have this check only for specific SCI versions. It would be better to remove hasEmptyScaleDrawHack altogether, and perform this sanity check for empty rectangles in all SCI32 games

bluegr:
Applied the scaler sanity checks to all SCI32 versions in 2fd416b0ac9750916e388e95429700275f98113a. Closing this

PR #1712, opened 2019-06-26
WINTERMUTE: Increase window title viewport height by 1 for "5 Lethal Demons" game

lolbot-iichan:
For some reason getFontHeight() is off-by-one comparing to height set in
TITLE_RECT, which made text being bigger then title rect and drawing was
skipped.

Which means, either game author was wrong by 1 and original Wintermute was more tolerant to off-by-one errors, or something is wrong with getFontHeight(). As there are no reports of bugs related to wrong font height, I believe the issue is with this one game, not the graphics\fonts\ttf.cpp.

So, let's provide a quick workaround for this game.

This fixes https://bugs.scummvm.org/ticket/6501

### DIFF
--- a/engines/wintermute/ui/ui_window.cpp
+++ b/engines/wintermute/ui/ui_window.cpp
@@ -27,6 +27,7 @@
  */
 
 #include "engines/wintermute/base/base_game.h"
+#include "engines/wintermute/base/base_engine.h"
 #include "engines/wintermute/base/base_parser.h"
 #include "engines/wintermute/base/base_active_rect.h"
 #include "engines/wintermute/base/base_dynamic_buffer.h"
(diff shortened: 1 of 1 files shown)


bluegr:
This is odd... is this the only game where this is happening? Perhaps there's an off-by-one error, when calculating the font height or the title height?

lolbot-iichan:
> This is odd... is this the only game where this is happening? Perhaps there's an off-by-one error, when calculating the font height or the title height?

I'll reduce it to a minimal test case in a separate dcp package and provide information on how font height and title height are calcucated.

lolbot-iichan:
Reduced to a minimal testcase:
1. Download https://github.com/lolbot-iichan/wme_testsuite/tree/master/wintitle_test/packages
2. Download http://web.archive.org/web/20070705114336if_/http://dead-code.org/5ld/5ld_setup.exe
3. Put "Artistik CE.ttf" from "Five lethal demons" to my demo
Run "game.exe" - header "Quit?" is visible.
Run ScummVM - header "Quit?" is hidden.

Debugging shows, that it was hidden because of this code:
```
while (maxHeight > 0 && lines.size() * _lineHeight > maxHeight) {
    lines.pop_back();
}
```

Value `maxHeight` is calculated avaliable height. It is calculated from `TITLE_RECT {25,65,270,100}`, which is `35` (`100-65`).
Value `_lineHeight` is `_font->getFontHeight()`, which for "Artistik CE.ttf" is `36`.

I don't have a debug build of WME to check original `LineHeight` value, but they use this API to get it:
```
SIZE Size;
GetTextExtentPoint32(hDC, "Ay", 2, &Size);
int LineHeight = Size.cy;
```

bluegr:
Thanks for debugging this! So, this is a mistake in the original game, which went undetected in the original WME, and it was caught by a sanity check. Thus, your original workaround for this case is correct.

Thanks for your work! Merging

PR #1714, opened 2019-06-26
SCI32: Fix QFG4 Ad Avis capture lockup, bug #11001

sluicebox:
Fixes a lockup in a mandatory scene in the first floppy version


bluegr:
Nice! Thanks

PR #1715, opened 2019-06-26
SCI32: Make QFG4 tentacle patch GOG compatible

sluicebox:
This makes the tentacle script patch compatible with the NRS scripts that ship with the GOG release. The instructions are just too different between versions for a single signature to match, and I don't see an alternate patch approach. Without this, only one half of the patch is applied and the tentacles in this scene get stuck . 

One of the other tentacle script patches still doesn't match but that's good because the NRS version fixes that bug.

I've tested all the existing script patches for compatibility with the fan patches (whew!) and this is the only one that needs adjustment. Everything else matches and works great, which is a relief.


bluegr:
Nice! Thanks :) Merging

PR #1717, opened 2019-06-27
SCI32: Fix QFG4 character selection, bug #11002

sluicebox:
For those of you wondering if the game with a bug in its very last click could also manage to break the very first...

=)


bluegr:
Nice! I remember that I had issues with the character selection screen when clicking too quickly on a character.

Thanks for your work! Merging

PR #1721, opened 2019-06-30
SCI: Fix SQ4 English Amiga dress logic, bug #11004

sluicebox:


bluegr:
Nice catch! Thanks for your work.
Merging

PR #1722, opened 2019-06-30
MSVC: Fix build with recent versions of libcurl

lephilousophe:
The missing CURL_STATICLIB prevents linking with static libcurl.lib
(which is what is used for 32-bit builds)
The missing libraries raise external symbol errors with recent versions
of libcurl due to curl/curl@7921659.
More explanations can be found in curl/curl#2474.

### DIFF
--- a/devtools/create_project/msbuild.cpp
+++ b/devtools/create_project/msbuild.cpp
@@ -366,6 +366,7 @@ void MSBuildProvider::outputGlobalPropFile(const BuildSetup &setup, std::ofstrea
 	              "\t\t\t<MultiProcessorCompilation>true</MultiProcessorCompilation>\n"
 	              "\t\t</ClCompile>\n"
 	              "\t\t<Link>\n"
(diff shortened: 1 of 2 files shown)


SupSuper:
These should be conditional, as it's possible to build project with curl disabled. See for example: https://github.com/scummvm/scummvm/blob/6bc875fd03ddd06fcf7dfb08e8ec46215addc567/devtools/create_project/create_project.cpp#L411-L415

Extra libraries should be added via `setup`. I don't know why defines need to be "hacked" into x64 though.

lephilousophe:
Well, I just added a hack over the hack.
It seems that, to filter out some stuff for x64, MSVC  generator rebuild a defines list.
At this stage we don't have ``curlEnabled`` anymore to add define conditionnally.
It's the whole hack section which should be removed.

bluegr:
I agree with @SupSuper that these libraries should not be enforced, and should be conditionally enabled

SupSuper:
@lephilousophe I made a new PR to fix the hack and solve your issues, let me know if it works for you: https://github.com/scummvm/scummvm/pull/1725

lephilousophe:
Yours is totally better.
I will check if everything compile well but I can already close this PR.

PR #1723, opened 2019-06-30
JANITORIAL: Avoid C4121 warnings in MSVC by reordering fields

lephilousophe:
C4121 is about 'symbol': alignment of a member was sensitive to packing


bluegr:
Thanks, merging

PR #1724, opened 2019-07-01
WINTERMUTE: Fix slider bug at 1 1/2 Ritter and Oknytt

lolbot-iichan:
There is a bug introduced in WME Lite.

Testcase:
1. Download https://github.com/lolbot-iichan/wme_testsuite/tree/master/slider_test/packages
2. Download https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/wmelite/wmelite_10_win.zip
3a. Run game.exe -> Slider moves with the mouse while it is pressed and
released when mouse is released
3b. Run wmelite.exe -> Slider is never released
3c. Run ScummVM -> Slider is never released

Related bugs:
https://bugs.scummvm.org/ticket/6567
https://bugs.scummvm.org/ticket/9861

Reason: slider is a button object that changes it's X until "LeftRelease" event is revieved and breaks the endless loop.
Sample code: https://github.com/lolbot-iichan/wme_testsuite/blob/master/slider_test/data/interface/system/speechvolume.script
However, WME Lite does not send "LeftRelease" events to any objects, if Game object can handle such event, even is game is frozen and UI is shown.

Original code: https://github.com/lolbot-iichan/Wintermute-Engine/blob/master/src/engine_core/wme_ad/AdGame.cpp#L2218
Changed in WME Lite: https://github.com/lolbot-iichan/wmelite/blob/master/src/AdGame.cpp#L2120

This behaviour was introduced in SVN period of wmelite, I believe it was
made on purpose for handling some iOS scenarios at commit #37 mentioned
on page https://code.google.com/archive/p/wmelite/source/default/commits

My proposal is to mark iOS apps as WME_LITE and fill exact WME_1_X_X/WME_LITE versions for all known
games. WME_1_X_X can be seen by right-clicking on main executable and seems to be non-empty for almost all existing WME games.

### DIFF
--- a/engines/wintermute/ad/ad_game.cpp
+++ b/engines/wintermute/ad/ad_game.cpp
@@ -38,6 +38,7 @@
 #include "engines/wintermute/ad/ad_scene.h"
 #include "engines/wintermute/ad/ad_scene_state.h"
 #include "engines/wintermute/ad/ad_sentence.h"
+#include "engines/wintermute/base/base_engine.h"
 #include "engines/wintermute/base/base_file_manager.h"
 #include "engines/wintermute/base/font/base_font.h"
 #include "engines/wintermute/base/base_object.h"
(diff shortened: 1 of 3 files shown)


bluegr:
Well, it’s not a “pirate” translation, just a fan made translation, applied as a patch to the original. Can you please modify the description of the first commit, and remove the “pirate” bit, please?

lolbot-iichan:
> 
> 
> Well, it’s not a “pirate” translation, just a fan made translation, applied as a patch to the original. Can you please modify the description of the first commit, and remove the “pirate” bit, please?

Sorry, just a slang word, I didn't realized this sounds like some kind of blame in English.
Fixed at baa2044280ee826c562e8f992c8bfcfff8eba4b8

bluegr:
Thanks! Great work! Merging

PR #1725, opened 2019-07-01
CREATE_PROJECT: Fix missing MSVC x64 defines

SupSuper:
This addresses the issues brought up in PR #1722 in a more robust way;

To disable nasm from x64, the whole define list was rewritten from scratch using `getFeatureDefines` and `getEngineDefines`. This ended up removing [important defines](https://github.com/scummvm/scummvm/blob/master/devtools/create_project/create_project.cpp#L391) added later in the setup phase, which presumably someone noticed, as they were [poorly re-added](https://github.com/scummvm/scummvm/blob/master/devtools/create_project/msvc.cpp#L158).
Instead, I opted to just find and remove the nasm define for x64. Better one little hack than a whole pile of them. :)

This also adds the new libcurl Win32 dependencies and cleans up the rest so they're only generated for the appropriate targets. More detailed explanations in the commit messages.

### DIFF
--- a/devtools/create_project/create_project.cpp
+++ b/devtools/create_project/create_project.cpp
@@ -366,6 +366,7 @@ int main(int argc, char *argv[]) {
 	StringList featureDefines = getFeatureDefines(setup.features);
 	setup.defines.splice(setup.defines.begin(), featureDefines);
 
+	bool backendWin32 = false;
 	if (projectType == kProjectXcode) {
 		setup.defines.push_back("POSIX");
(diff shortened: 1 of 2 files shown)


lephilousophe:
I failed to compile in Release mode but I suspect more my VM than the code.
I got a titanic.lib which is over 850MB and got permission denied when linking it in the last stage.
Anyway, I think it will fix the bugs I got. It had in Debug.

bluegr:
Well done with the pull request @SupSuper, and thanks for testing it @lephilousophe. This is good to be merged now

PR #1728, opened 2019-07-01
SCI: Disable SQ4 timepod code for missing room

sluicebox:
This disables a timepod code that crashes the CD version because it's for a room that was removed from the game.

Normally this wouldn't be necessary, since no one will accidentally enter the *eighteen* digits, but the NRS version of SQ4 recreates this room and so the code appears on easter egg sites. People aren't reading the fine print about fan patches and think it's a regular easter egg code. Recently a video appeared on youtube of people attempting it, crashing in ScummVM as they would have in the original, and loudly blaming ScummVM.

Now that I see how easy it is to selectively enable patches by resource, which I was way wrong about, I'm doing that here. A fan version of this room's script already exists, and I love it, so if you have that as patch files then it will still work.

https://spacequest.fandom.com/wiki/Room_Removed_For_Legal_Reasons

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -13618,6 +13618,31 @@ static const uint16 sq4CDPatchSewerLockup[] = {
 	PATCH_END
 };
 
+// SQ4CD had an easter egg room of things removed from Sierra games for legal
+//  reasons, but the room itself was removed from the game. Instead the room's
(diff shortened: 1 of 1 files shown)


bluegr:
Nice! Thanks for working on this quite obscure easter egg. Merging

m-kiewitz:
> a video appeared on youtube of people attempting it, crashing in ScummVM as they would have in the original, and loudly blaming ScummVM.

Link please

PR #1735, opened 2019-07-06
BACKENDS: Fix crash when save file is owned by another user

orgads:
fopen() fails, but the null handle was passed on to OutSaveFile.


bluegr:
Nice catch, thanks!
Merging

PR #1736, opened 2019-07-07
GLK: Fix off-by-one memory read

lephilousophe:
I think this part of code is erroneous as it shouldn't check the previous item on the first iteration of the loop.
Moreover, this code seems to raise an internal compiler error of OpenPandora build. I hope my rewrite will fix it as well.


dreammaster:
Good catch on that. Fingers crossed it helps openpandora :)

PR #1737, opened 2019-07-08
SCI: Implement early GetLongest() behavior, bug #10000

sluicebox:
This implements an off-by-one bug and comparison change in text length calculations that was in all SCI interpreters until late 1990. This fixes bug #10000 (mwahahaha)

On the notebook screen at the end of LB1, the last "E" in the word "INCOMPLETE" is wrapped onto the next line in ScummVM. It's a script bug, but thanks to the interpreter bug it accidentally works in the original.

GetLongest() returns the number of characters that can be displayed in a given width. "INCOMPLETE" is 52 pixels but the script specifies a width of 50. GetLongest() should return 9 and set the string pointer to "E" but in SSCI when a string exceeds the width before a space character is reached, the exceeding character was originally applied to the result. Normally this isn't a scenario that a script should be in, you'd be in trouble if the first word in your string was longer than your maximum width, so this bug might only affect buggy scripts.

Sierra eventually fixed the bug and at the same time changed the width comparison from greater than or equals to greater than. I've added both behaviors for games that contained them. To test this I patched the LB1 script to pass kDisplay a yellow background instead of transparent to see what SSCI was really drawing, then patched the width to various values to see the wrapping behavior and confirm that SSCI and ScummVM now do the same thing.

Versions with the original behavior:
* All SCI0
* Some SCI01: KQ1 and QFG1 Japanese but not multilingual SQ3 and LSL3
* QFG2
* Only the first SCI1 early game: KQ5 English DOS floppy versions
	
Every other SCI1 game I've looked at has the fix. By the interpreter version numbers, the fix in the SCI1 timeline starts with LSL1 Demo.

To locate this in any DOS SCI16 interpreter, search for these unique instructions in the text width function:

```
8a d8    mov bl, al
32 ff    xor bh, bh
```
	
The function that calls it the most, 4 or 5 times, is GetLongest(). Versions with the bug increment the character counter and string pointer prior to testing the text width and returning.

### DIFF
--- a/engines/sci/graphics/text16.cpp
+++ b/engines/sci/graphics/text16.cpp
@@ -25,6 +25,7 @@
 #include "graphics/primitives.h"
 
 #include "sci/sci.h"
+#include "sci/engine/features.h"
 #include "sci/engine/state.h"
 #include "sci/graphics/cache.h"
 #include "sci/graphics/coordadjuster.h"
@@ -50,6 +51,7 @@ void GfxText16::init() {
 	_codeFontsCount = 0;
 	_codeColors = NULL;
 	_codeColorsCount = 0;
+	_useEarlyGetLongestTextCalculations = g_sci->_features->useEarlyGetLongestTextCalculations();
 }
 
 GuiResourceId GfxText16::GetFontId() {
@@ -250,6 +252,14 @@ int16 GfxText16::GetLongest(const char *&textPtr, int16 maxWidth, GuiResourceId 
 		if (tempWidth > maxWidth)
 			break;
 
+		// the previous greater than test was originally a greater than or equals when
+		//  no space character had been reached yet
+		if (_useEarlyGetLongestTextCalculations) {
+			if (lastSpaceCharCount == 0 && tempWidth == maxWidth) {
+				break;
+			}
+		}
+
 		// still fits, remember width
 		curWidth = tempWidth;
 
@@ -273,7 +283,7 @@ int16 GfxText16::GetLongest(const char *&textPtr, int16 maxWidth, GuiResourceId 
 	} else {
 		// Break without spaces found, we split the very first word - may also be Kanji/Japanese
 		if (curChar > 0xFF) {
-			// current charracter is Japanese
+			// current character is Japanese
 
 			// PC-9801 SCI actually added the last character, which shouldn't fit anymore, still onto the
 			//  screen in case maxWidth wasn't fully reached with the last character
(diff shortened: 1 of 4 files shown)


tsoliman:
Is the the rationale to implement the interpreter bug (as opposed to fixing the script bugs) due to how infeasible it would be to patch all the script bugs? Were any of the script bugs already patched and will now be broken?

HTo be clear: I am just curious. I am not opposed to this at all.)

sluicebox:
Yes, the rationale is to accurately implement the interpreter behavior so that scripts run as they did before without individual modification. I don't know how many other scripts depended on this behavior (possibly few, possibly none) and good news, I don't need to.

I'm as sure as I can be that the off-by-one behavior is a bug and comfortable calling it that but who knows, maybe it was intentional at the time. It doesn't really matter, that's what shipped.

There are no existing script patches for this issue as this issue wasn't known until I wrote what you read! =)

engines/sci/graphics/text16.cpp
```diff
@@ -340,6 +357,46 @@ int16 GfxText16::GetLongest(const char *&textPtr, int16 maxWidth, GuiResourceId
 	return resultCharCount;
 }
 
+// GetLongest() originally contained an off-by-one character bug which was
+//  fixed in late 1990. As part of this change, a width test was changed
+//  from greater than or equals to greater than. This method returns which
+//  behavior to use based on the current version and game.
+bool GfxText16::useEarlyGetLongestCalculations() {
```
bluegr: It would be better to move this inside features.cpp, like the rest of the version-specific feature checks

engines/sci/graphics/text16.cpp
```diff
@@ -340,6 +357,46 @@ int16 GfxText16::GetLongest(const char *&textPtr, int16 maxWidth, GuiResourceId
...
+	// - LB1 PC 1.000.046
+	// - ICEMAN PC 1.033
+	// - SQ3 English PC 1.018
+	// - PQ2 Japanese 1.000.052
+	case SCI_VERSION_0_EARLY:
+	case SCI_VERSION_0_LATE:
+		return true;
+
+	// SCI01: confirmed KQ1 and QFG1 Japanese,
+	// fixed in SQ3 and LSL3 multilingual PC
```
bluegr: Perhaps it’s possible to check for changes to core game scripts, instead of hardcoding game IDs? Other checks in features.cpp have been implemented like this

sluicebox:
Oh yes, GameFeatures is the right place for this, I've moved it and expanded the feature description.

I would like to identify by script and took a look but I don't see how. This doesn't appear to be an api change that's reflected in scripts, it just seems to be an isolated fix for an interpreter edge case that may not even affect any other screens. (But I hope it does!)

Another wrinkle is that the Japanese interpreters have their own versions of GetLongest() to handle double byte, where the old behavior was implemented but differently, and that seems to be why it reappears in QFG1 English/Japanese even though it's a newer interpreter than the SQ3/LSL3 multilingual versions.

The complicated one is SCI01 because the bug is in the first and last interpreters, but there's only five documented versions and I already inspected them, so it ends up not being that complicated https://wiki.scummvm.org/index.php/Sierra_Game_Versions#SCI_Games

Any suggestions?

bluegr:
Great work! Many thanks :) Quite an obscure change, indeed. Glad you managed to clear this mess.

Changes to the SCI interpreters like this one were quite common, which is why we put them all in one place.

This will clear the issues we had with obscure minor differences to text rendering.
Merging

OmerMor:
This is great! :-)
Anecdotally, here is Sierra's original comment for this bug fix:
https://github.com/OmerMor/SCI16/blob/master/INTERP/SCI.DOC#L1094-L1096

PR #1739, opened 2019-07-09
SCI: Remove GAMEOPTION_ORIGINAL_SAVELOAD from Mac versions

sluicebox:
This removes the option for using original save/load screens from SCI16 and GK1 Mac versions because they didn't have them and so it currently crashes with signature mismatches and other problems.

Instead the Mac interpreter would just launch the operating system's open/save file dialog. The game scripts don't contain the original UI.

restore


bluegr:
I remember that SCI0 - SCI1 Mac games had the usual save/load dialogs... but it's been awhile

sluicebox:
Interesting, do you remember which ones? I confirmed SQ3 and that's pretty early.

If you have any Mac versions lying around then can confirm this in Game:save and Game:restore, they just make a kernel call instead of running dialogs first.

bluegr:
My mistake then. I don’t have my Mac versions handy on this PC, which is why I said what I remembered from my other PC.

Thanks for your work! Merging

PR #1740, opened 2019-07-09
SCI: allow fanmade games to use resource.000

Kawa-oneechan:
SCI11 doesn't have split volumes after all. Fixes The Dating Pool.


bluegr:
Thanks for this. Merging

PR #1746, opened 2019-07-12
SCI: Fix ICEMAN destroyer timer, bug #11017

sluicebox:


bluegr:
Nice work! Tricky bug indeed. Merging

engines/sci/engine/script_patches.cpp
```diff
@@ -2694,6 +2696,128 @@ static const SciScriptPatcherEntry gk2Signatures[] = {
...
+
+static const uint16 icemanDestroyerTimer1Patch[] = {
+	0x30, PATCH_UINT16(0x001f),         // bnt 001f [ state 8 ]
+	PATCH_ADDTOOFFSET(+0x1f),
+	0x3c,                               // dup
+	0x35, 0x08,                         // ldi 08
+	0x1a,                               // eq?
+	0x31, 0x04,                         // bnt 04 [ state 9 ]
+	0x35, 0x78,                         // ldi 78
+	0x65, 0x10,                         // aTop cycles [ cycles = 120 ]
```
m-kiewitz: I would prefer it if that would be changed to seconds.

sluicebox:
We could do that, I went this route to avoid making an editorial change to the timer during the part that worked. Are you concerned that this is too short, or that it should be in wall time? Most (all?) of the global submarine timers are in game cycles, so tying it to game speed is consistent with the rest.

I got the 120 cycle value from the maximum remaining after the bridge scene at usable speeds, so that avoided having to make up a seconds value, and already gives you a few more cycles than you often have. The only action you're supposed to do during this short timer is press up to walk and automatically sit in the chair, and that's game speed dependent, so making up a seconds value means coming up with a long enough one to reasonably work at slowest speed, though that means you can wander farther (another room?) at fast speeds.

Do you want to try some values? The "seconds" offset is 0x12, if you find a value that makes things better at all speeds, I'm all for it. I really only care about the premature death on the bridge. I witnessed two people, people who knew what they were doing, get stuck on it for fifteen minutes and take turns slamming enter to dismiss the text boxes, trying different speeds, etc, and so I took up vengeance driven development.

m-kiewitz:
I replaced quite a few .cycles in Gabriel Knight 1 with .seconds.
I think most of the time I divided it with a number, maybe 10 - I can't remember anymore.
Anyway, the problem with cycles is that it's dependent on the machine speed, we should somewhat already keep it around the same amount using the speed throttler, but it's not good at all. Seconds however should work the same everywhere.

sluicebox:
Ah, yes, so on these older ones it ends up working differently than newer ones like GK1, and I think this is the one situation where a cycle timer is appropriate.

In the EGA parser games like this, the user-controlled game speed determines how fast cycles decrements, since Game:play throttles its doit calls with kWait gGameSpeed. In this case, the timed action is walking from the start point to a chair, and that takes the same number of cycles at all game speeds but not  the same number of seconds.

At some point the user-controlled Speed setting changed from controlling the speed of the entire game to just ego's speed, along with whatever individual things needed to be explicitly synchronized to that. A cycle based timer for those newer games is indeed big trouble. (Maybe unrelated, but I fixed some conspicuous timing bugs in GK1, like cut-off speech during the intro, and my theory is that no contemporary PCs just couldn't run that game at full speed, so even when they were programming wall clock timings they were unintentionally factoring in a lag that would go away in a year or two.)

We can change this to seconds, but there's not a conversion between cycles and seconds since the user controls how much long a cycle takes. If we make this a timeout in seconds then we have to invent one, and make sure it's long enough to not timeout before ego can walk to the chair at the slowest speed. It's not like that's hard but it is an editorial change, and unlike the newer games it's not necessary because cycles works at all speeds. Are you getting different results?

PR #1747, opened 2019-07-12
GRAPHICS: Fix ManagedSurface transBlitFrom bugs

SupSuper:
Should be fairly self-explanatory, just some copy-paste errors in ManagedSurface. PRing in case something was relying on them (it only seems to be called by macgui).


bluegr:
Looks good, +1 from me

sev-:
Good catch!

PR #1748, opened 2019-07-13
WINTERMUTE: Huge update of detection tables

lolbot-iichan:
I downloaded/bought most of WME games so that we can use exact engine versions in detection table instead of LATEST_VERSION (which is equal to WME_LITE behaviour, that might differ from original WME 1.x).

I guess this pull request might also be useful for updating ScummVM wiki, since some wiki information on Wintermute games is outdated.

This also includes checksums from https://bugs.scummvm.org/ticket/11009 and https://bugs.scummvm.org/ticket/11011


bluegr:
Very nice work! Well done

Since you have cleaned up a lot of entries in the detection tables, could you also mark the games that are 2.5D, so that they can be easily distinguished from the rest?
I am aware of the following 2.5D entries:
- Basis Octavus
- Pizza Morgana
- Zbang! The game

Not sure if there are any more 2.5D games in there

PR #1749, opened 2019-07-15
SCI: Implement full kWait behavior

sluicebox:
This implements kWait's full behavior as in SSCI. This fixes PQ3 bug #11020 where ScummVM adds a half-second delay before every message when a character's portrait is displayed.

kWait is currently implemented as a simple sleep for the specified ticks, but SSCI factors in the last time kWait was called, and doesn't necessarily sleep at all. The specified ticks is the maximum to sleep for. If that many ticks haven't yet elapsed since kWait was last called then that difference is the amount to sleep for.

PQ3 calls kWait 30 before displaying a message after a portrait, and so in the original this would only pause if kWait had been called within the last half a second, which usually wasn't the case. (Script 0, exported function 16)

kWait's work is done in EngineState:wait() which was also being used by internal operations. Those shouldn't influence any of kWait's behavior so I've made a dedicated sleep() method for them to use. wait() is now solely for kWait. This has been a problem before: https://github.com/scummvm/scummvm/commit/2f9967524f610f277c9c5109d1339e3e87957804

I've tested this on a lot of games. Most just use this in their core loop and classes but I've also tested the few that directly call kWait with a significant value.

Could someone with Hoyle5 please test the script patch that resurrects kWait in SCI32 and uses it in a spin loop? I don't have that game, but I've tested other kWait spin loops, and there shouldn't be a problem.

### DIFF
--- a/engines/sci/engine/kgraphics.cpp
+++ b/engines/sci/engine/kgraphics.cpp
@@ -394,9 +394,9 @@ reg_t kTextSize(EngineState *s, int argc, reg_t *argv) {
 }
 
 reg_t kWait(EngineState *s, int argc, reg_t *argv) {
-	int sleep_time = argv[0].toUint16();
+	uint16 ticks = argv[0].toUint16();
 
-	const int delta = s->wait(sleep_time);
+	const uint16 delta = s->wait(ticks);
 
 	if (g_sci->_guestAdditions->kWaitHook()) {
 		return NULL_REG;
--- a/engines/sci/engine/state.cpp
+++ b/engines/sci/engine/state.cpp
@@ -134,14 +134,26 @@ void EngineState::speedThrottler(uint32 neededSleep) {
 	}
 }
 
-int EngineState::wait(int16 ticks) {
+uint16 EngineState::wait(uint16 ticks) {
 	uint32 time = g_system->getMillis();
-	const int tickDelta = ((long)time - (long)lastWaitTime) * 60 / 1000;
-	lastWaitTime = time;
 
+	uint32 ms = ticks * 1000 / 60;
+	uint32 duration = time - lastWaitTime;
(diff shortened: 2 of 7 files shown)


m-kiewitz:
Will verify that.

bluegr:
I will also check this tonight

bluegr:
I just tested this with PQ3 and other SCI1.1 games, as well as the Hoyle 5 games, and they all work great! :)
Well done, nice work!
+1 from me. Awaiting input from @m-kiewitz

lskovlun:
We should also have some explicit testing with SCI0.

sluicebox:
I did a full play through of Colonel's Bequest with this change, that's the one I'm overly familiar with and figured if there's any game I might notice a subtle timing change in, it would be that one. Like other SCI0 games I looked at, kWait throttles the game loop to game speed in Game:play and Game:replay, so I'd hope if there were a problem there it would be immediately noticeable. I also tested the rooms that explicitly call kWait: walking over the floorboards in the chapel room 58 and exiting room 74. It all looked and felt the same.

I don't know much about the speed throttler but I ran a few scenes that are mentioned in the comments and that there are other throttle workarounds for:
- KQ6 Intro
- SQ4 Sequel Police
- SQ4 Intro

Run side-by-side prior to the change, they appeared the same.

I should have mentioned that I verified kWait in the disasm of a lot of interpreters along the SCI16 timeline and they all did this.

m-kiewitz:
I have verified that even an early SCI0 interpreter has this code.

bluegr:
Great. So this works fine. If there are no objections by Sunday, we can merge this

bluegr:
So, this has been tested, this code is contained in the original, and no obvious ill side effects have been found.

Thanks for your work! Merging

PR #1753, opened 2019-07-18
BLADERUNNER: Delete subtitles temporary variable

dafioram:
It was previously only being deleted when the subtitles file was found.

### DIFF
--- a/engines/bladerunner/subtitles.cpp
+++ b/engines/bladerunner/subtitles.cpp
@@ -122,14 +122,14 @@ Subtitles::~Subtitles() {
 //
 void Subtitles::init(void) {
 	// Loading subtitles versioning info if available
-	TextResource *versionTxtResource = new TextResource(_vm);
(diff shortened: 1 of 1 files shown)


engines/bladerunner/subtitles.cpp
```diff
@@ -147,12 +147,13 @@ void Subtitles::init(void) {
 		       _subtitlesInfo.languageMode.c_str(),
 		       _subtitlesInfo.credits.c_str());
 
-		delete versionTxtResource;
-		versionTxtResource = nullptr;
 	} else {
 		debug("Subtitles version info: N/A");
 	}
 
+	delete versionTxtResource;
```
peterkohaut: oh I missed that, thanks!
versionTxtResource can be local variable directly (not on heap), so no "new"  and "delete" are needed

dafioram:
Updated.

PR #1754, opened 2019-07-19
CLOUD: Make Storages connect via cloud.scummvm.org

Tkachov:
@Mataniko suggested more secure and convinient way of connecting cloud storages and implemented it on [cloud.scummvm.org](http://cloud.scummvm.org/), so I've updated my code to use it.

Along with that, I've updated GUI of Cloud tab to be more user-friendly and moved local webserver into a separate Wi-Fi Sharing tab. There are also a couple minor Widgets changes, minor "common/util" addition and some changes in Cloud-related stuff in this PR.

Cloud tab with connected Box storage

Hope this time Clouds will finally make it to release build (=


Mataniko:
Looks great, if there are no objections in the next couple of days I'll merge.

common/util.cpp
```diff
@@ -163,4 +163,43 @@ bool isPunct(int c) {
 	return ispunct((byte)c);
 }
 
+
+#pragma mark -
+
+
+Common::String getHumanReadableBytes(uint64 bytes, Common::String &unitsOut) {
+	Common::String result = Common::String::format("%lu", bytes);
```
criezy: Using `%lu` here might be an issue since on some systems a long int is 32 bits.
That being said it seems to me that this line, as well as lines 174 to 178, could just be removed.

criezy:
I had a very quick look. It looks good overall, but I would like to get a bit more time to check the GUI changes (I am planning to do that this evening).

common/util.cpp
Tkachov: 1) Should I use `%llu` instead?
2) If we remove these lines (guess not completely, but leaving `if (bytes >= 1024) bytes /= 1024;`), the smallest units this function would return would be megabytes. It's a little bit inconsistent and probably not really convinient for files less than 1 MB.

criezy: Right. Indeed those lines cannot be fully removed.
Line 171 can be removed though, as well as line 176 (the two lines on which you format the result with a `%lu`) since result is always set again on line 201 from the float value.

By the way, is there a reason why you don't set the float before the first `if` case (and divide it in the if case) to also have one decimal after the dot in the kb case in your string? Here it's a bit inconsistent with kb always having no decimal and Mb, Gb, and Tb having one.

Tkachov: Hmm, I didn't mean to return floating point on bytes and KBs; but I wanted to not repeat the same line three times (in each of those `if`) and put it afterwards.

So, the idea was that bytes (0..1023) could be printed as is; KBs are small enough, so there is no need of decimal part, but after that (in MBs, GBs, TBs) decimal part gets important. May be it's better to use decimal for KBs for consistency though.

criezy: You are likely to still want to not use decimal point in the bytes case though, so the question on `%lu` remains.

One possibility would indeed be to use `%llu` and cast the `bytes` value to an `unsigned long long int` explicitly to avoid a warning in case `uint64` is defined as an `unsigned long int`. However `%llu` was added in the C99 standard, and I am wondering if there might still be compilers out there that does not support it (I think we might still be using some pretty old compilers for some ports).

Here since you are only going to use it if `bytes` is less than 1024 (or 1024 * 1024 if you use it for Kb as well), you know it fits in a 32 bits unsigned integer. So you could also downcast to `unsigned long int` and keep using `%lu`).

backends/cloud/basestorage.cpp
```diff
@@ -0,0 +1,214 @@
...
+		warning("BaseStorage: no refresh token available to get new access token.");
+		if (callback) (*callback)(BoolResponse(nullptr, false));
+		return;
+	}
+
+	Networking::JsonCallback innerCallback = new Common::CallbackBridge<BaseStorage, BoolResponse, Networking::JsonResponse>(this, &BaseStorage::tokenRefreshed, callback);
+	if (errorCallback == nullptr)
+		errorCallback = getErrorPrintingCallback();
+
+	Common::String url = Common::String::format("https://cloud.scummvm.org/%s/refresh?code=%s", cloudProvider().c_str(), _refreshToken.c_str());
```
bgK: I suggest passing the refresh token as a POST parameter (or a request header) so it is not stored in the web server logs.

gui/options.cpp
```diff
@@ -2156,81 +2208,158 @@ void GlobalOptionsDialog::handleCommand(CommandSender *sender, uint32 cmd, uint3
...
+			}
+
+			if (cancel) {
+				return;
+			}
+		}
+
+		if (_storageWizardConnectionStatusHint)
+			_storageWizardConnectionStatusHint->setLabel(_("Connecting..."));
+		CloudMan.connectStorage(
```
bgK: In my opinion after entering the code and after the account information request completes, the user should be asked to confirm the account is the one he expects before ScummVM starts uploading files to the remote storage.
This is to avoid upload of user files to attacker controlled storage account in case the server is compromised, and to avoid phishing attempts (plz enter this code in ScummVM).

Mataniko: Right now we store the auth token for about 10 minutes, allowing users to log in to multiple devices before it expires.

Alternatively, we can change the server implementation to delete the token after the first use in addition to the 10 minutes lifetime.

I've been leaning towards the latter to prevent someone spamming attempts to try and grab a token, but the attack vector is small enough as it is and so low value since we're pretty well scoped.

backends/cloud/basestorage.cpp
Mataniko: POST wouldn't be appropriate here, but if we want to move it to a header that's fine.

bgK: I'd be glad to be enlightened as to why POST is not appropriate :)

gui/options.cpp
bgK: I agree with you regarding the token retrieval endpoint. However I'm not sure how this relates to my point. Perhaps I was not clear enough?

backends/cloud/basestorage.cpp
Mataniko: @Tkachov I'm going to move it from a query param to a header named `X-ScummVM-Refresh-Token`

gui/options.cpp
Mataniko: Thinking about it some more, once a token is retrieved, it should be deleted from the server.

backends/cloud/savessyncrequest.cpp
```diff
@@ -113,6 +115,13 @@ void SavesSyncRequest::directoryListedCallback(Storage::ListDirectoryResponse re
 				_filesToUpload.push_back(file.name());
 			else
 				_filesToDownload.push_back(file);
+
+			if (_localFilesTimestamps[name] == DefaultSaveFileManager::INVALID_TIMESTAMP)
+				debug(9, "- uploading file %s, because it is has invalid timestamp", name.c_str());
+			else if (_localFilesTimestamps[name] > file.timestamp())
+				debug(9, "- uploading file %s, because it is %d seconds newer than remote\n\tlocal = %d; \tremote = %d", name.c_str(), _localFilesTimestamps[name] - file.timestamp(), _localFilesTimestamps[name], file.timestamp());
+			else
+				debug(9, "- downloading file %s, because it is %d seconds older than remote\n\tlocal = %d; \tremote = %", name.c_str(), file.timestamp() - _localFilesTimestamps[name], _localFilesTimestamps[name], file.timestamp());
```
criezy: I am guessing there is a `d` missing after the `remote = %` here.

gui/options.cpp
```diff
@@ -104,14 +104,17 @@ enum {
...
 	kRunServerCmd = 'rnsv',
 	kCloudTabContainerReflowCmd = 'ctcr',
 	kServerPortClearCmd = 'spcl',
 	kChooseRootDirCmd = 'chrp',
-	kRootPathClearCmd = 'clrp'
+	kRootPathClearCmd = 'clrp',
+	kConnectStorageCmd = 'Cnnt',
+	kOpenUrlStorageCmd = 'OpUr',
+	kPasteCodeStorageCmd = 'PsCd',
+	kDisconnectStorageCmd = 'DcSt',
```
criezy: Please remove the comma after the last enum value. This is a C++11 extension and we not all ports support this yet.

gui/options.cpp
```diff
@@ -1731,64 +1747,27 @@ void GlobalOptionsDialog::build() {
...
-	if (g_system->getOverlayWidth() > 320)
-		_rootPathButton = new ButtonWidget(container, "GlobalOptions_Cloud_Container.RootPathButton", _("/root/ Path:"), _("Specifies which directory the Files Manager can access"), kChooseRootDirCmd);
-	else
-		_rootPathButton = new ButtonWidget(container, "GlobalOptions_Cloud_Container.RootPathButton", _c("/root/ Path:", "lowres"), _("Specifies which directory the Files Manager can access"), kChooseRootDirCmd);
-	_rootPath = new StaticTextWidget(container, "GlobalOptions_Cloud_Container.RootPath", "/foo/bar", _("Specifies which directory the Files Manager can access"));
-
-	_rootPathClearButton = addClearButton(container, "GlobalOptions_Cloud_Container.RootPathClearButton", kRootPathClearCmd);
-
+	addCloudControls(container, "GlobalOptions_Cloud_Container.", context);
+#endif USE_LIBCURL
```
criezy: Should be `#endif // USE_LIBCURL`

criezy:
I am getting an error when it tries to synchronise the save files: `Dropbox returned error: path/disallowed_name/..!`

After debugging a bit it turns out it is trying to upload the `.DS_Store` file that resides in my save directory (this is a hidden sytem file on macOS used to store attributes on the folder).

I think the cloud sync should be ignoring all the files starting with a `.` (at least on macOS and Linux).

backends/cloud/basestorage.cpp
Tkachov: OK. I'm in for POST too though (not semantically appropriate for RESTful, but who really cares?)

gui/options.cpp
Tkachov: While this is less convinient, I guess that's the price of being more secure. I will do that.

BTW, can anybody share what they are seeing as "username" for OneDrive? I get some gibberish, which is probably how cyrillic is sent by MS. Might also be unicode, which isn't supported by our JSON implementation.

backends/cloud/savessyncrequest.cpp
Tkachov: Oof.

gui/options.cpp
Tkachov: Will do.

gui/options.cpp
Tkachov: OK, will fix.

Tkachov:
Yeah, I developed it on Windows, so haven't encounter such stuff. I think it's OK to ignore files starting with `.` on Windows as well.

Should we ignore such files from the remote storage though? For example, if some other provider allows uploading files starting with `.`, and some user manually puts it in his ScummVM folder. I think it's "Yes" for games download, but "maybe?" for saves sync.

criezy:
I would say yes for saves sync as well. I cannot see any use case where a user might want to put a file starting with a `.` in his ScummVM savegames directory and want it to be synchronized.

gui/options.cpp
Tkachov: Yeah, they send "\uXXXX\uXXXX..." in JSON, which we're parsing as "\u00XX\u00XX..." because AFAIK there is no unicode support in ScummVM. I'm not sure we're OK with that because gibberish instead of username ("display" name, actually) would be confusing.

common/util.cpp
Tkachov: I've commited updated version, that does downcast to `unsigned long int`, so I guess we can resolve this one?

Tkachov:
Now files with '.' in name are ignored in both saves sync and game files download.

common/util.cpp
criezy: Yes, I am marking this as resolved.

gui/options.cpp
criezy: We have `Common::U32String` which maybe could help here?

Tkachov: Well... it looks like I've been able to parse the string correctly in json.cpp using `U32String`, and then I use `convertFromU32String()`. I've tried all available encodings, though it probably should be UTF-8, and neither give acceptable results. In UTF-8 and Windows-1251 (default cyrillic) I see something resembeling what is needed (similar to what happens in other applications which handle cyrillic/utf-8 combination poorly). `convertUtf32ToUtf8()` doesn't seem to work either.

What I get is "\x04<correct cyrillic char>\x04<correct cyrillic char>..." in scummvm.ini and just a weird combination of cyrillic chars in GUI. I don't know what will happen if there would be Chinese or Japanise.

criezy: Yes, the GUI itself doesn't support unicode, which means that you would need to convert to whichever encoding it is using to display it. If translations is enabled, this would be `Common::TranslationManager::getCurrentCharset()` and I guess this could be implemented in `U32String` using the `Common::TranslationManager::getCharsetMapping()` and do a reverse mapping. However this might well fail, for example in your case if you are not using Russian in the ScummVM GUI, it will probably be using an encoding that does not support cyrillic characters. The only set of characters that we are guaranteed to have are the ASCII characters, so we would need to have a fallback that transliterate to ASCII.

@vyzigold is flying through his GSoC tasks and almost finished work on text-to-speech. The next task will be to work on an encoding conversion mechanism, using iconv when possible, in ScummVM. This is something that would be useful for the text to speech (for example to convert from CP850, used by some game engines, to whatever the TTS system needs, usually UTF-8). And it might help here as well. Also @bgK recently mentioned there is something like that in ResidualVM. However while iconv has some support for transliteration, I am not sure it can transliterate cyrillic to ASCII so we might still have issues when the GUI encoding does not support some of the characters used in the string.

Tkachov: If I shift unicode cyrillic chars 864 positions down, I get needed chars in GUI. The same constant is stated in Russian article about ISO 8859-5 in Wikipedia: https://ru.wikipedia.org/wiki/ISO_8859-5 (it's not in English version of the article).

Basically, it's enough to use the following code:
```
if (string[i] >= 0x410 && string[i] < 0x450) { // cyrillic
	charsetString += (char)(string[i] - 0x410 + 176);
}
```

When I'm using English, `TransMan.getCurrentCharset()` says it's `ASCII`. When I switch to Russian, it's `iso-8859-5`. But, `TransMan.getCharsetMapping()` returns a very strange table:
```
1f => Я
80000020 => а
```
while in Unicode (and, as it follows from my experiment and that Wikipedia article, in actual ISO 8859-5):
```
42f => Я
430 => а
```
So, there should be continuous interval of uppercase and then lowercase alphabet characters, but TransMan's mapping table maps them in two different intervals and they don't seem to be correct.

I guess we can either leave everything as it is for this PR so someone fixes it later, or I can write a dumb converter for now ("if using ISO 8859-5, shift 864 positions down"), so basically fix it for cyrillic only.

criezy: The mapping table in the `TranslationManager` seems correct to me and it gives me a mapping from the GUI encoding to unicode. For example for the `ISO 8859-5` it gives me:
```
207 -> 1071 (0x42f)
208 -> 1072 (0x430)
```
And that is indeed a shift by 864.
As I noted to go from unicode to ISO-8859-5 we need to of a reverse mapping (searching for unicode values in the array to get the index (and the index is equal to the character in the current GUI encoding).

However, it is important to note that the `TranslationManager`character mapping tables does not contain the raw unicode values but also has a flag indicating if the character is required (it were designed for loading TTF fonts and check if a font is usable or not for this language). So you need to use a mask of `0x7FFFFFFF` to get the actual unicode value: `TransMan.getCharsetMapping()[207] & 0x7FFFFFFF => 1071`

That being said, in my opinion it would be fine to wait for the work on handling encodings to be done to implement that part cleanly.

Mataniko: @bgK what's the concern around server compromise/security in the current implementation? 

If the server is compromised, we have a bigger problem than revocable cloud keys, as an attacker can modify the source for our daily builds.

The phishing vector is so small, and the reward is non-existent (ScummVM saves), as we are properly scoped on all cloud providers to our own files (with the exception of Box, which doesn't have such a scope)

If someone was inclined, they could just get the cleartext refresh token from scummvm.ini

Any hardening should be done server side, but the average user should be presented with a simple workflow to get this going.

Mataniko: We can put warnings in both client and server to never trust a key not gotten directly from the website and shorten the TTL of the shortcode to help mitigate.

bgK: > If the server is compromised, we have a bigger problem than revocable cloud keys, as an attacker can modify the source for our daily builds.

Indeed, modifying downloadable binaries has a much higher value. Compromising the key server could be moderately interesting as it allows to target existing installations.

> The phishing vector is so small, and the reward is non-existent (ScummVM saves).

Well, the cloud sync system also copies files from the remote storage to the user's device. So it would be possible to distribute malware that way (Possibly a specially crafted save file for some engine that achieves local code execution).

> We can put warnings in both client and server to never trust a key not gotten directly from the website.

I'd be curious to know how you would do that.

Tkachov: Mataniko means adding text messages on cloud.scummvm.org (like "Don't tell this code to anyone") and in ScummVM GUI (like "Only connect storage if you've got code from cloud.scummvm.org").

I agree that cloud sync technically could be used to deliver payload, but I don't think it should be possible to execute it. Plus I hope it's not really possible to compromise the server in the first place.

So, I'm kinda fine with how it is, but I'm also more or less fine with adding confirmation step. Mataniko says it just complicates the feature for end user while it's already secure enough. Just let me know what you guys decide (=

Tkachov: I've added a warning label and a button to manually enable newly connected storage. Until then it does not allow saves sync or files downloading.

This is how storage connection wizard and newly connected storage look like:


If we decide not to make users go through this additional step, we can always revert this commit.

Tkachov: So @bgK @criezy is there anything else to do?

criezy: > Storage is yet disabled.

That sounds strange to me. I would suggest "Storage is still disabled." or "Storage is not yet enabled." (the latter sounds better to me than the former).

Or: "You are almost there. Verify that the username is correct and enable the storage".

Tkachov: OK, now it's "4. Storage is not yet enabled. Verify that username is correct and enable it:"

Mataniko:
Seems like we're ready to merge. @Tkachov Can you rebase to resolve the conflict?

Tkachov:
Yeah, sure.

gui/options.cpp
bgK: Sounds good to me. Thanks a lot @Tkachov.

PR #1755, opened 2019-07-19
WINTERMUTE: Update Carol Reed series engine versions

lolbot-iichan:
This sets exact WME 1.x versions for remeaining Carol Reed series games.
* Carol Reed 4 EN is using WME 1.7.2, like Carol Reed 4 EN Demo, unlike Carol Reed 4 RU.
* Carol Reed 12+ are all using WME 1.9.3

Thanks to Albadross for collecting those versions:
https://bugs.scummvm.org/ticket/10863


bluegr:
Thanks! Merging

PR #1756, opened 2019-07-19
COMPOSER: Make unload library error a warning

angstsmurf:
This fixes Gregory and the Hot Air Balloon. See bug #11021.

This is just a quick and lazy fix. I have not looked into the underlying reason 
why the game tries to unload a library which isn't loaded in the first place,
but making it a warning does not seem to have any adverse effects.

### DIFF
--- a/engines/composer/composer.cpp
+++ b/engines/composer/composer.cpp
@@ -535,7 +535,8 @@ void ComposerEngine::unloadLibrary(uint id) {
 		return;
 	}
 
-	error("tried to unload library %d, which isn't loaded", id);
+	warning("tried to unload library %d, which isn't loaded", id);
+	return;
 }
 
 bool ComposerEngine::hasResource(uint32 tag, uint16 id) {


angstsmurf:
I tried to bisect this, but it turns out not to be a regression, as it already occurs in ScummVM 1.5.0, where the Composer engine was introduced. To trigger it, just play until Gregory writes his todo-list (after crashing with the balloon), click on the todo-list and then spam Escape.

bluegr:
So was this never found, or perhaps there’s another issue here?

angstsmurf:
My guess is that it was just never noticed. 

Bug [#11021](https://bugs.scummvm.org/ticket/11021) is really about (at least) two different bugs, one of which is a duplicate of [#10861 ](https://bugs.scummvm.org/ticket/10861) and was fixed by e45dd70  ("archive does not contain 'ANIM' 2715."). Confusion about this is probably why the bug reporter claims that the game worked in ScummVM 1.9 but not in the latest dailies.

bluegr:
Makes sense :) Thanks for working on this.
Merging

PR #1757, opened 2019-07-19
COMPOSER: Don't crash when skipping invalid animations

angstsmurf:
Don't try to run a NULL animation. This fixes a segmentation
fault in Gregory and the Hot Air Balloon that sometimes occurs
when making sundaes.


bluegr:
Quite a straightforward change. Thanks! Merging

PR #1760, opened 2019-07-21
WINTERMUTE: Fix language selection regression bug for Reversion

lolbot-iichan:
Ooops. Latest refactoring of those if/else blocks was wrong, as a result
Reversion 1&2 ignored all the xlanguage packages. Fixed.

### DIFF
--- a/engines/wintermute/base/base_file_manager.cpp
+++ b/engines/wintermute/base/base_file_manager.cpp
@@ -228,38 +228,60 @@ bool BaseFileManager::registerPackages() {
 			// TODO: Select based on the gameDesc.
 			if (_language != Common::UNK_LANG) {
 				// English
-				if (_language != Common::EN_ANY && (fileName == "english.dcp" || fileName == "xlanguage_en.dcp" || fileName == "english_language_pack.dcp")) {
-					continue;
+				if (fileName == "english.dcp" || fileName == "xlanguage_en.dcp" || fileName == "english_language_pack.dcp") {
+					if (_language != Common::EN_ANY) {
+						continue;
+					}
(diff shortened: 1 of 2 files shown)


bluegr:
I’ve been running all day this weekend and I’m a bit tired, so... what’s the bug here?

lolbot-iichan:
> I’ve been running all day this weekend and I’m a bit tired, so... sorry if I’m missing the obvious, but what’s the bug here?

Look at simplified code:
```
				if (_language != Common::EN_ANY && fileName == "xlanguage_en.dcp") {
					continue;
				} else if (_language != Common::RU_RUS && fileName == "xlanguage_ru.dcp") {
					continue;
				...
				} else if (fileName.hasPrefix("xlanguage_")) {
					warning("Unknown language package: %s", fileName.c_str());
					continue;
				}
```
Original idea was:
1. Ignore "xlanguage_en.dcp" if lang is not English
2. Ignore "xlanguage_ru.dcp" if lang is not Russian
3. ...
N. Ignore **OTHER** "xlanguage_*.dcp" packages (this code is here just in case Reversion would get a new translation)

However, this code was wrong and resulted in:
1. Ignore "xlanguage_en.dcp" if lang is not English
2. Ignore "xlanguage_ru.dcp" if lang is not Russian
3. ...
N. Ignore **ALL** "xlanguage_*.dcp" packages (including "xlanguage_en.dcp", "xlanguage_ru.dcp")

bluegr:
Ah right. Missed the default case. Thanks :)
The other commit is not related, but it’s quite a minor change.

Merging

PR #1761, opened 2019-07-21
SCI: Fix kDrawPic palette validation

sluicebox:
This fixes kDrawPic's palette validation, fixing Camelot bug #11024.

Camelot calls kDrawPic 112 2 1 4 in the aqueduct to paint a black screen, but four is an invalid palette number.

GfxPicture::drawVectorData validates _EGApaletteNo by resetting it to zero if it's greater than three, the max legal value, but it does this after _EGApaletteNo has already been used to read from the array EGApalettes and set the pointer *EGApalette, and it doesn't reset the pointer, so it continues to read out of bounds values.

Visual Studio debug builds expose this, causing the pic to be drawn pink instead of black.


bluegr:
Nice catch! Merging

PR #1762, opened 2019-07-21
COMMON: Support text wrapping with even width line segments

antoniou79:
The new arguments are optional. The lines segments will be close to the same width.

The algorithm is similar to the one we had for Blade Runner but not exactly the same, since that one would wrap a line at a white space after the theoretical split point (quotient of full line text width divided by target lines number)

### DIFF
--- a/graphics/font.h
+++ b/graphics/font.h
@@ -175,8 +175,8 @@ public:
 	 * @param initWidth the starting width of the first line, for partially filled lines (optional)
 	 * @return the maximal width of any of the lines added to lines
 	 */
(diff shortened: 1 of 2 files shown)


bluegr:
Looks good, well done! +1 from me

graphics/font.h
```diff
@@ -175,8 +175,8 @@ class Font {
 	 * @param initWidth the starting width of the first line, for partially filled lines (optional)
 	 * @return the maximal width of any of the lines added to lines
```
sev-: Could you please describe the new parameters here?

sev-:
Looks great, just lacks a small piece of documentation.

sev-:
Great! merging

PR #1764, opened 2019-07-21
WINTERMUTE: Add temporary debug logs for inspecting bug 6567

lolbot-iichan:
For some reason, on Amiga backend, while volume slider is pressed in 1
1/2 Ritter, mouse starts to move to the right slowly. I need additional
logs to see if mouse is moved by system or by game script. Details:
https://bugs.scummvm.org/ticket/6567

Those logs are active only when "debug_cursor" is set to true at
ConfMan, so normal usecases are not affected. However, I plan to revert
this PR once bug is closed.

### DIFF
--- a/engines/wintermute/ad/ad_game.cpp
+++ b/engines/wintermute/ad/ad_game.cpp
@@ -2134,6 +2134,10 @@ bool AdGame::validMouse() {
 
 //////////////////////////////////////////////////////////////////////////
 bool AdGame::onMouseLeftDown() {
+	if (_debugCursorPos) {
+		LOG(0, "CURSOR: onMouseLeftDown");
+	}
+
 	if (!validMouse()) {
 		return STATUS_OK;
 	}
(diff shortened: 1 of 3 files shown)


bluegr:
It would be better to either make a special build with this debug code, or clean it up a bit and add it permanently.

I believe that the original reporter (Raziel) could also apply this change as a patch and create a local debug build with this change

lolbot-iichan:
I don't have enough skills to cross-build an Amiga port under Windows, and I don't think that I may have access to BuildBot to request such build. I also don't like the idea of adding those logs permanently - there are no other known bugs that would require such logs.

@raziel-, what do you think about building a local debug build from this PR (and about the whole story with being involved in testing lots of sliders >_<)? Is it OK for you?

raziel-:
Local debug is fine, being the debug monkey is also fine :-)

I can do it Friday earliest though, just point me to either the debug tree or patch.

Thank you very much and please keep on fixing :-)

lolbot-iichan:
> I can do it Friday earliest though, just point me to either the debug tree or patch.

This bug waited more then five full years to be inspected, it can wait some more days for sure. =)

bluegr:
The patch is now used locally by @raziel-, and work is continued within the bug itself. Thus, this PR is no longer needed to be open

PR #1765, opened 2019-07-22
WINTERMUTE: Add detection for Italian demo of On the Tracks of Dinosaurs

lolbot-iichan:
NOTE: This is a 2.5D game that is out of ScummVM scope

Developer's Download page states that there are EN/RU/IT versions:
http://reptilesquest.000webhostapp.com/index.php?page=5
However, rq_demo_beta25_multilang.exe allows to select only EN/RU.
italian.dcp can be extracted from rq_demo_beta25_multilang.exe using any
tool for Inno Setup installer.


bluegr:
Interesting. Thanks for spotting that

PR #1777, opened 2019-07-27
SCI: Add detection entry for debug version of Codename: Iceman

lotharsm:
This special debug build was salvaged by Clint Basinger ("LGR") from an IBM PCjr that was previously owned by Ken Williams from Sierra On-Line.

The story of this particular PC is pretty wild. Clint got it from a PC warehouse/thrift store and the most interesting stuff at the beginning was the pretty rare hard drive expansion. Later it was revealed that the PC was previously owned by Sierra On-Line. Later Ken Williams confirmed that the PC was previously owned by himself.

Clint managed to copy the files from the (almost dead) hard drive that only occasionally even started and uploaded them to archive.org.

Full video: https://www.youtube.com/watch?v=Z-VBITW94zI

Screenshots taken with ScummVM:
scummvm00023
scummvm00024
scummvm00025
scummvm00026
scummvm00027

### DIFF
--- a/engines/sci/detection_tables.h
+++ b/engines/sci/detection_tables.h
@@ -323,7 +323,7 @@ static const struct ADGameDescription SciGameDescriptions[] = {
 		AD_LISTEND},
 		Common::EN_ANY, Common::kPlatformDOS, 0, GUIO_STD16_UNDITHER	},
 
-	// Codename: Icmeman - English DOS Debug build, Version 1.009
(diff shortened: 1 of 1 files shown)


lotharsm:
> 
> 
> Iceman *

D'oh! Thanks, good catch!

bluegr:
Quite an interesting discovery :) Discovering lost little gems like this one is great.

Thanks for finding this! Merging

PR #1779, opened 2019-07-29
QUEEN: skip talk on right click

BLooperZ:
DOS version of the game allows skip talking on right mouse click,
this commit adds support for this functionality.

Also, when talking the game ignores click on the close button (of the window)
until dialogue is ended. so it applies skip talk on this event as well.

old behaviour:
right click while talking -> wait for talking to end -> do action as if there was right click now
(if meanwhile mouse was moved, it clicks somewhere else).

new behaviour (matches DOS executable):
right click while talking -> skip dialogue -> do action because there was right click now.

Thanks

EDIT:
add skip talk to left click as well to match DOS executable


bluegr:
Very good, nice work! Thanks for working on this. Merging

PR #1782, opened 2019-08-01
DISTS: Fix OpenGL renderer for builds with MSYS2

antoniou79:
DISTS: Fix OpenGL renderer for MSYS2/Mingw(64 or 32)

The issue pertains to MSYS2 adding a default manifest file (default-manifest.o) to the executable

The bug is for PC systems with GPU drivers that were not properly supported for Windows 10 systems, like Intel HD Graphics series 1rst and 2nd generations. In those systems, launching a game in ScummVM (built with MSYS2/Mingw) with the OpenGL renderer would cauase the game screen to be a white blank image, and various warnings would be output to the console eg.
"WARNING: GL ERROR: GL_INVALID_ENUM on glTexSubImage2D(0x0DE1, 0, 0, area.top, src.w, area.height(), _glFormat, _glType, src.gere.cpp:167)!"

This was due to MSYS2/Mingw builds trying to load the (poorly supported) GPU driver, while "advertising" support for Windows 10 in their embedded default Manifest file. Hence, the GPU driver dll (eg ig4icd64.dll for Intel HD graphic series) would be unloaded, causing the bug.

More information is available in the following links:
 * https://github.com/pal1000/save-legacy-intel-graphics

 * https://github.com/LWJGL/lwjgl/issues/119

 * https://github.com/msys2/MSYS2-packages/issues/454

 * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69880

### DIFF
--- a/devtools/create_project/msbuild.cpp
+++ b/devtools/create_project/msbuild.cpp
@@ -398,6 +398,7 @@ void MSBuildProvider::createBuildProp(const BuildSetup &setup, bool isRelease, b
 	              "\t<PropertyGroup>\n"
 	              "\t\t<_PropertySheetDisplayName>" << setup.projectDescription << "_" << configuration << outputBitness << "</_PropertySheetDisplayName>\n"
 	              "\t\t<LinkIncremental>" << (isRelease ? "false" : "true") << "</LinkIncremental>\n"
+	              "\t\t<GenerateManifest>false</GenerateManifest>\n"
 	              "\t</PropertyGroup>\n"
 	              "\t<ItemDefinitionGroup>\n"
 	              "\t\t<ClCompile>\n";
--- a/devtools/create_project/visualstudio.cpp
+++ b/devtools/create_project/visualstudio.cpp
@@ -289,6 +289,7 @@ void VisualStudioProvider::createBuildProp(const BuildSetup &setup, bool isRelea
 		              "\t<Tool\n"
 		              "\t\tName=\"VCLinkerTool\"\n"
 		              "\t\tLinkIncremental=\"1\"\n"
+		              "\t\tGenerateManifest=\"false\"\n"
 		              "\t\tIgnoreDefaultLibraryNames=\"\"\n"
 		              "\t\tSetChecksum=\"true\"\n";
 	} else {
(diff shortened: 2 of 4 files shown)


lotharsm:
Can you verify that the change doesn't break compilation with Visual Studio? IIRC @SupSuper had a problem compiling ScummVM with Visual Studio when I simply added a manifest file the last time.

antoniou79:
> Can you verify that the change doesn't break compilation with Visual Studio? IIRC @SupSuper had a problem compiling ScummVM with Visual Studio when I simply added a manifest file the last time.

Not soon. I don't use Visual Studio. Could someone else try and verify this?

sluicebox:
It was me! https://github.com/scummvm/scummvm/pull/1541#issuecomment-475919057

The same applies here, create_project needs to be updated to set the visual studio project setting to not generate a manifest, which it does by default, otherwise the linker will bomb because one already exists. We didn't get that far on the other PR, but I can take a look later today to see what to change and test that.

The resource script should use the ID CREATEPROCESS_MANIFEST_RESOURCE_ID like in the other PR. (it's in WinUser.h)

That bare-bones manifest file is good and exactly what Visual Studio defaults to generating. Having this in place will make it easy to play around and test other manifest settings that come up in things like the DPI Wars, etc.

sluicebox:
We can also get rid of DISCARDABLE in the resource script, that hasn't had any affect since Win16.
https://docs.microsoft.com/en-us/windows/win32/menurc/common-resource-attributes

antoniou79:
> We can also get rid of DISCARDABLE in the resource script, that hasn't had any affect since Win16.
> https://docs.microsoft.com/en-us/windows/win32/menurc/common-resource-attributes

Should I remove all the DISCARDABLE instances there?

sluicebox:
Let's leave those, I'll zap them in a later commit with an explanation and keep this one free of distractions.

The ID change looks good, I'll get create_project figured out later and tested against a few visual studios.

When this is done and the commits are squashed, I think it should be renamed to something like WIN32: Add Application Manifest and then mention in the commit message that it has the side effect of preventing the obscure bug in the msys2 build.

sluicebox:
Here's the changes to create_project needed for visual studio compatibility:

https://gist.github.com/sluicebox/3e406fdac61abc5cdf95aab97f6c7897

I'm sure there are countless better ways contribute these three lines than a diff paste, sorry!

SupSuper:
Looks good! Every modern Windows application should have a manifest (MSVC adds one by default) as it makes them Vista-aware and gets rid of backwards compatibility hacks.

Tested it on MSVC and old Windows and there don't seem to be any breaking changes.

bluegr:
Great work! Creating a default (and custom, later on) manifest is a good way of controlling the behavior of the final executable.

Since this has been tested on different compilers and several versions of Windows without any adverse effects, it is good to be merged.

Merging

PR #1785, opened 2019-08-02
GUI: Minor fixes (scrollbars, tooltips, tabs, EditText)

Tkachov:
I've been making a custom theme based on Modern Remastered these couple of days (I call it Midnight theme). While I was at it, I locally fixed some things in GUI. Well, some changes are not really fixes, but edits that I thought look better.

These **are not** changes for my theme only, but sometimes they are better illustrated with my theme (scrollbars, for example). Below are pics comparing what was before the fixes and after these in both my theme and Modern Remastered (click to view in full size).

### 15ee60f

- list widget scrollbar is not cropped on the right now:

01_cropped_scrollbar_in_list_widget fw

.

- scrollbar buttons now take as much space as they should, not overlapping with scrollbar handle:

02_scrollbar_buttons_size fw

(illustrated in Midnight theme only, because Modern Remastered does not have any background or borders in scrollbar buttons, so you cannot see the overlapping)

### 95b8c99

- scrollbar repositioned in ScrollContainerWidget, and widget itself is now drawn so its bottom border goes lower than clipped contents:

03_scroll_container fw

(unfortunately, I couldn't resize scrollbar to have the same height as container  for my Midnight theme, because then in Modern Remastered it overlap tabs not 1px, but 2px. As a result, we can see these "double borders" in Midnight)

(we can see that "tab" drawstep behaves strangely, and draws 2 more pixels where ScrollContainerWidget's area already begins; using "square" drawstep in Midnight theme solves that and looks more tidy)

### be1e8ae

- added 2px padding in tooltip:

04_tooltip fw

.

- shifted text in EditTextWidget 1px up, so it's more vertical aligned:

05_edittext fw

### 7dda053

- fixed TabWidget's < and > buttons to use values from the theme:

06_tab_buttons fw

.
 
And a long pic for lowres:

07_lowres fw

.

So, 15ee60f and 7dda053 are actual fixes (cropped scrollbar, TabWidget typos) and 95b8c99 with be1e8ae are just my edits that look prettier (vertical aligned text in EditText, 2px margin in tooltip, repositioned scrollbar).

P.S. Midnight theme is available [on my site](http://tkachov.ru/scummvm/themes/midnight/)

### DIFF
--- a/gui/ThemeEngine.cpp
+++ b/gui/ThemeEngine.cpp
@@ -1030,7 +1030,7 @@ void ThemeEngine::drawScrollbar(const Common::Rect &r, int sliderY, int sliderHe
 	drawDD(kDDScrollbarBase, r);
 
 	Common::Rect r2 = r;
-	const int buttonExtra = (r.width() * 120) / 100;
+	const int buttonExtra = r.width() + 1; // scrollbar.cpp's UP_DOWN_BOX_HEIGHT
 
 	r2.bottom = r2.top + buttonExtra;
 	drawDD(scrollState == kScrollbarStateUp ? kDDScrollbarButtonHover : kDDScrollbarButtonIdle, r2,
--- a/gui/Tooltip.cpp
+++ b/gui/Tooltip.cpp
@@ -49,8 +49,8 @@ void Tooltip::setup(Dialog *parent, Widget *widget, int x, int y) {
 	const Graphics::Font *tooltipFont = g_gui.theme()->getFont(ThemeEngine::kFontStyleTooltip);
 
 	_wrappedLines.clear();
-	_w = tooltipFont->wordWrapText(widget->getTooltip(), _maxWidth - 4, _wrappedLines);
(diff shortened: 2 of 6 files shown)


Mataniko:
Looks great, thank you for doing this.

I really like the scrollbar in the midnight theme, maybe we should bring that over to the remastered theme....

bluegr:
I really like the changes! Subtle, but quite welcome! The buttons after the tabs are much much nicer now

Tkachov:
@criezy , about that content overlapping: I guess this could/should be fixed by adding more right padding in the layout. That's a workaround, but if this glitch happens because of overlapping, I think the easiest way to fix it is just make things not overlap.

Mataniko:
@Tkachov before merging, can you run scummvmtheme.py?

Tkachov:
@Mataniko I sure can, but why is that needed? These changes are in code, not in themes layout.

Mataniko:
My bad, I thought there was an XML change too. Merging.

PR #1786, opened 2019-08-04
SCI: Fix CAMELOT sword sheathing during room change

sluicebox:
Fixes bug #11042


bluegr:
Nice work, thanks! Quite an obscure bug. Merging

PR #1791, opened 2019-08-06
TSAGE: Add detection for Spanish ringworld2 translation

aviloria:
TSAGE: Add detection for Spanish ringworld2 translation

Fully working translation from here: http://www.pakolmo.galeon.com/RingWorld2.htm


dreammaster:
This seems like seriously impressive effort, but I'm afraid I'm going to have reject this. The translation group provided the full r2rw.rlb for download with the text changed, which is the entirety of the game. As such, it runs foul of our policy against supporting piracy. We could only officially detect and support it if they removed the full download, and came up with some kind of patcher/diff tool that, for example, transformed a known legitimate English version users would have to have into the translated Spanish version.

PR #1798, opened 2019-08-11
SCI32: Fix Mirrored Pic Drawing

sluicebox:
This fixes an off by one error in our SCI32 mirrored pic drawing. This can be seen in any room with style property bit $0400. The pic is drawn one coordinate to the left of where it should be and leaves a line of garbage on the right or left edge of the screen, depending on the compiler.

This fixes at least QFG4, which has many mirrored rooms such as 551, and LSL6 Hi Res rooms 305, 640, and 680.

Confirmed in disassembly. Fixes bug #10748


bluegr:
Nice catch, thanks! Merging

PR #1801, opened 2019-08-12
SCI: Fix SQ4CD hintbook text layout

sluicebox:
This fixes SQ4CD's hintbook layout, which is currently drawn incorrectly, due to parsing message text as if it were escaped hex codes. SQ4CD's interpreter doesn't have this feature. Converting this text to hex codes, and then applying the buggy hex conversion that appeared in later interpreters, causes newlines to appear and the text sizing to be off.,

Fixes #11070


bluegr:
Nice! Merging

PR #1802, opened 2019-08-12
TSAGE: Removed "pirated" flag from Spanish fanmade translation

aviloria:
TSAGE: Removed "pirated" flag from Spanish fanmade translation

I have been talking to the translation team... they didn't mean to distribute any "pirate" version.
At first, they just released the modified files and didn't notice they were releasing all the needed files to run the game... the same way they have done with all other translations in their webpage (http://www.pakolmo.galeon.com), but this time with a fatal result. They have presented their apologies for that

Now they have changed the downladed file to a patch requiring the original game to be applied.
http://www.pakolmo.galeon.com/RingWorld2.htm

### DIFF
--- a/engines/tsage/detection_tables.h
+++ b/engines/tsage/detection_tables.h
@@ -186,8 +186,7 @@ static const tSageGameDescription gameDescriptions[] = {
 		GF_CD | GF_ALT_REGIONS | GF_DEMO
 	},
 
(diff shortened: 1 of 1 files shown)


bluegr:
I see that they now provide an installer, which checks for the existence of the original game files before overwriting them with the translated ones. I believe that will do.

Thanks for handling this! Merging

PR #1803, opened 2019-08-13
VITA: Use activeArea rectangle also on Vita

rsn8887:
This PR changes the Vita graphics manager so that the AR correction and game screen texture scaling is done similarly to other surfaceSDL platforms. However, several issues came to light concerning aspect ratio correction now that Vita does it similarly as other platforms:

- AR doesn't work at all in Dreamweb when using 2x or 1x rendering. I observe this also on latest MacOS nightly. It works using OpenGL on MacOS though.

- AR works just fine in the main menu, and seems to work fine in Monkey Island on first glance. On closer inspection, enabling AR causes dirty rect gfx glitches in Monkey Island when the mouse is moved across the screen (warped) with the analog stick or touch input. When the cursor is close to the top edge of the screen, also significant slowdown occurs, not just flickering of rects.

- AR Is done via software scaling instead of adapting activeArea dimensions. I suspect a historical reason for this but it seems unnecessary and also slower.

These problems only appear when the aspect ratio correction tickbox is on. Otherwise everything seems to work fine.

Thanks to @ccawley2011 for pointing me towards this.

### DIFF
--- a/backends/events/psp2sdl/psp2sdl-events.cpp
+++ b/backends/events/psp2sdl/psp2sdl-events.cpp
@@ -37,6 +37,9 @@
 
 #include "math.h"
 
+#define TOUCHSCREEN_WIDTH 960
+#define TOUCHSCREEN_HEIGHT 544
+
 PSP2EventSource::PSP2EventSource() {
 	for (int port = 0; port < SCE_TOUCH_PORT_MAX_NUM; port++) {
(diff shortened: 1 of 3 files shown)


rsn8887:
Since the new stretch mode "Fit to window (4:3)" in combination with unticking the "aspect ratio correction" tickmark gives a hardware aspect ratio that works similar to how Vita used to work, I think I can merge this already.

PR #1807, opened 2019-08-14
SAGA: initial support for Japanese PC-98 version of ITE

mistydemeo:
This pull request adds initial support for the Japanese PC-98 CD-ROM version of Inherit the Earth. It's currently partially-playable, but with a number of issues. I've marked this PR as a draft right now, and I'm hoping to at least fix the voices.

* Voices are garbled. Pretty sure this is just a matter of the wrong sample rate / bit depth, because they're just barely recognizable as the English voices. I'll investigate this further.
* The intro plays with English subtitles, while the original version plays with Japanese subtitles.
* The subtitles in the rest of the game use the English font, which is obviously not prepared for Japanese text, and render as mojibake. I saw a comment elsewhere that ScummVM doesn't yet support Shift JIS in this engine.
* While most dialogue works (aside from text rendering), certain lines trigger exception similar to `ResourceContext::getResourceData() wrong resourceId 2573!`.


bluegr:
Thanks :) This is quite a rare version, which is why support for it hasn't been added up to now. Regarding your questions:

- There were several versions of ITE, and a lot of sound compression schemes were used. The engine itself tries to find the appropriate sound scheme dynamically, but some formats are specified manually, cause the respective sound files do not have any specific header. All the relevant code is in sndres.cpp, - you may be interested in lines 284 - 301. Try the following:
1) Set GF_8BIT_UNSIGNED_PCM in the detection entry flags
2) Check if the VOX format is used (resourceType = kSoundVOX, lines 291-293)
3) Check if the sound files are encoded in big endian format (resource.cpp, lines 202-223) - BE encoded files are marked as GAME_SWAPENDIAN)
- The intro strings are hardcoded. They are all in itedata.cpp, lines 655-955
- Usually, Japanese text is stored as multi-byte, which would explain why the text is shown garbled: currently, it is assumed that all game text is single-byte. The font resources for each game version are hardcoded in its detection entry (the ITE_GameFonts lines)
- The error shown means that probably some text resources are not properly loaded. Perhaps this is somehow related to the multi-byte issue above?

Feel free to ask any other questions that come along :)

mistydemeo:
Thank you! I figured out the voices after a little experimentation - appears to be 8-bit signed PCM at 10kHz. Sounds correct that way; I've pushed up a commit using that.

I still need to take a look at the text (font, encoding, etc.). I'm 99% sure it'll be Shift JIS, but I'll have to see. How were the strings that are hardcoded in ScummVM extracted from the other releases? Are they part of the executable?

bluegr:
Glad you sorted the voices out :)

Yes, the strings in the intro (and all the strings in itedata.cpp) are hardcoded in the original executable

athrxx:
Thank you for your work on this :-)
I have pulled your branch and added added two commits on top of yours.

https://github.com/athrxx/scummvm/tree/ite_pc98

The first commit does some fine tuning to the digital audio. While your 10 kHz guess seems to work fair enough this is not exactly how the original does that. The resources have header information which allows to calculate the precise rate.
I also removed the game feature flag, since the PC98 platform is not exactly a game feature and the platform can be queried via _vm->getPlatform().

The second commit adds music support. I haven't hooked up the non-digital sound effects, since I am not sure yet whether these are actually used.

If you're okay with these additions (I am not a SAGA dev, I just enjoy little side tasks from time to time) you can simply cherry-pick the commits into this PR branch.

Regarding the Japanese texts: From my tests with the Neko 2 emulator the font looks quite suspiciously like the ROM font. We have support for that in our common code: graphics/sjis.h/.cpp. But I haven't examined that part of the code. So it could also be a font from file. Just tell us if you need help figuring things out...

mistydemeo:
Thank you, @athrxx! That's really helpful!

I wrote this PR as a way of learning how to contribute to ScummVM, so if you have any suggestions like those two commits, I'd love to get pointers on how to approach those things - it'd be more helpful to me than getting the work in the form of something that's already complete. 😄

I'll look at the existing ROM font support - that definitely sounds like the right approach here. I'll look into the SJIS support used by other games too!

Unfortunately it doens't look like there are any comments in the commit that added the strings extracted from the EXE as to whether there's a convenient structure to identify/extract them, but I'll look at grabbing the equivalent strings from the Japanese version and adding them.

mistydemeo:
I located the strings; I've pushed them up. While deciding what to do with the multibyte text, I've used UTF-8 for the Japanese for clarity. That may end up changing later.

I'm missing three strings from the menu text. I haven't yet found it among the related strings in the EXE, and it's possible they're not there; one of them is explicitly marked as not used in ScummVM.

athrxx:
Looking at other engines that use Japanese ROM fonts (e. g. Scumm, Kyra, SCI) seems to be a good start for this task.

The strings actually have to stay in their original SJIS encoding, because the character drawing methods will only eat them that way. So you can basically dump them from the executable as they are.
We escape the chars >127. The lower chars stay as actual chars. So the strings should look like in this example (just 2 strings taken from the interface strings):

strings[] = {
"\x8f""E""\x82\xa6\x82\xc8\x82\xa2\x82\xcb\x81""B",
"\xcf\x82\xed\x82\xc1\x82\xbd\x82\xe0\x82\xcc\x82\xcd\x82\xc8\x82\xa2\x82\xcb\x81""B"'
};

For the readability I suggest to keep the UTF-8 strings in comments.

athrxx:
So, I can definitely confirm from disam that the PC-98 version uses the font ROM for both 2-byte and 1-byte chars.

The characters have to be drawn in 640x400 resolution, so you'll have to change the resolution for the PC-98 version. To have the extra code as non-intrusive as possible, I would suggest to merge the text with the rest of the graphics at the last possible moment in Render::drawDirtyRects(). Render the SJIS text to (a) separate hires surface(s). Write a scale2x() method, use that to scale all the dirty rects and then merge the text surface. You'll have to be careful with the dirty rect handling.
But that is only a suggestion. I can't even tell you for sure whether the SAGA engine allows such an approach. I am not too familiar with the engine myself. 

Meanwhile I have looked at the issue with the missing speech and the invalid resource error in the second part of the intro. I'll try to be more constructive this time and not just commit a fix :-P. From comparing disasm with our code I'd suggest that this has to be fixed in Script::opSpeak(). Please note that the original code does not have opSpeak() as a separate function. There is only runThread() with 87 switch cases. opSpeak is case 82.

If you're unfamiliar with reverse engineering you might want to read Dreammaster's manual:
https://wiki.scummvm.org/index.php/HOWTO-Reverse_Engineering

Please note that for this pull request noone requires you to actually get into reverse engineering, though.

mistydemeo:
> For the readability I suggest to keep the UTF-8 strings in comments.

OK, I've done this. Hope you don't mind that rather than escaping ambiguous (hex-like) sub-127 characters like `"E"`, I encoded those as hex like the 128+ characters.

I'll look into using that font and handling the 640x400 res properly.

athrxx:
Thanks! The strings are perfectly fine that way.

The PC98 extra credits ("NEC PC-9821 version by...") might want their own array eventually, since they get displayed in a different scene. But that is not an immediate issue...

bluegr:
Any news on this?

sev-:
@mistydemeo I would like to merge it in its current state, and potentially look into additional issues. Any objections?

mistydemeo:
Definitely. I'm sorry I lost track here - my last comments/commits were around the time my dad died last year, and I wasn't doing any work on personal projects for awhile. I'd love to come back to this soon, and I'm happy to follow up in a second PR after you merge this if you're happy with that.

sev-:
Could you try to rebase it to the top of current master? There are some conflicts now, as the code bitrot.

If you need help with that, I can definitely provide it.

athrxx:
Should I push a rebased version of this branch? We do have the author's consent to merge and I have a branch of this in my fork anyway...

mistydemeo:
Rebased. Looks like only the detection table conflicted, so that was a pretty easy rebase.

sev-:
Great, merging.

sev-:
Please, continue work on it, so we could announce the version as supported. If you need help, I could definitely provide it.

athrxx:
I've committed some followup commits which should further improve the intro.
So a good task to do next would probably the Japanese font rendering.
The advantage of this task is that it can be done without looking at disasm.

mistydemeo:
I could use a couple pointers on that - maybe a good implementation in another engine to look at. I saw the suggestion of using internal high res rendering but not sure how to go about it. Is FM Towns Zak the best example, or is something else better to look at?

sev-:
No, that would be a way too complex example. What you need to do is basically upscale 2x screen before rendering, you could take a look at `Screen::scale2x()` in `kyra/graphics/screen.cpp`. And then render Japanese on the resulting screen.

I vaguely remember that we had some common routines for rendering FM-TOWNS Japanese characters, need to look it up, though.

athrxx:
The common code for rendering the Japanese ROM fonts is in graphics/sjis.cpp/.h. It also supports PC98 font drawing.

As you can see SAGA has a Font class (font.cpp/.h) with not too many public methods. You can put an abstract font class there from which that existing font class and the new Japanese font class inherit. The code for the new class will be much less, since you'll just have to wrap the common code which does the actual rendering. And do the rendering on a 640x400 surface.

For the merging of the surfaces you might really want to look at Kyra like sev recommended. Even that is a bit more complex than what you need. You can ignore the hicolor code, basically all the `bpp == 2` checks. But it is certainly more agreeable than Scumm which is probably the most complex implementation. 

In KYRA the merge happens in `Screen::updateDirtyRectsOvl()`, screen.cpp. First treating the graphics surface with scale2x() so that it matches the text sureface in size, then copying the text surface on top of that, then sending everything to our "middleware" via `OSystem::copyRectToScreen()`.

For SAGA, these things happen in `Render::drawDirtyRects()`, render.cpp. In both engines you have a variant for rendering the full screen or dirty rects. The dirty rects are a common source of error. If you find that something doesn't get drawn or is cut off there is a good chance that you messed these up.

I that sufficient for a start? Just ask if you need help. Or join us on discord and ask there...

PR #1808, opened 2019-08-14
TTS: Text to speech support

vyzigold:
This pull request adds text to speech support on Linux, Windows and macOS (author of the macOS text to speech manager is @criezy ). It also uses this to add text to speech to the mortevielle engine, to the GUI (after allowing it in options) and it adds text to speech tests to the Testbed engine. This is a GSoC task described [here](https://wiki.scummvm.org/index.php?title=Summer_of_Code/GSoC_Ideas_2018#Text_to_speech).

**To run the text to speech on Linux you need:**

- Install and configure speech-dispatcher, either through your package manager, or from [github](https://github.com/brailcom/speechd)

- Install one (or more) text to speech engine, supported by speech-dispatcher are: eSpeak, Festival, Flite, Pico. I tested it (end it worked) with eSpeak-ng and Festival.

Once you are successfuly able to use the speech-dispatcher to speak from the terminal, everything should be set up correctly to work with ScummVM. For more info on configuring speech-dispatcher, look at [speech-dispatcher documentation](https://freebsoft.org/doc/speechd/speech-dispatcher.html)

**To run the text to speech on Windows you need:**
(Do all of these steps even for MinGW-w64, which has it's own SAPI. Their implementation unfortunately doesn't have all the needed features.)

- Install SAPI, for example from [here](https://www.microsoft.com/en-us/download/details.aspx?id=10121)

- From the Program Files*\Microsoft Speech SDK 5.*\Include copy the sapi.h, sapiddk.h and sperror.h to your include directory, so the compiler finds them. 

- From the Program Files*\Microsoft Speech SDK 5.*\Lib\*\ copy the sapi.lib to your lib directory, so the compiler finds them. (with MinGW-w64 delete or rename libsapi.a in the MinGW-w64's lib folder)

- You have to explicitly enable the TTS when running configure script with the --enable-tts option. Or if you are using Visual Studio, run the create_project with --enable-tts option.

The text to speech is disabled by default on Windows, because the compilation currently ends with a warning: "Warning: corrupt .drectve at end of def file", which is caused by using a msvc compiled library (sapi.lib) with MinGW. Even with the warning, everything seems to work correctly.


vyzigold:
There are 2 things, I would like to know the opinion of the community about.

1. sphelper.h
I use some functions from SAPI's sphelper.h. But because it is written to work with msvc only, I had to modifie it and create sphelper-scummvm.h so it works with MinGW too. Currently there are some warnings when compiling, which I could try to correct if needed. But I think it might be a better idea to leave the file as close to the original sphelper.h as possible, so that in the future, when there is a new version of SAPI, it is hopefuly easier to transition to that new version.

2. GUI TTS enabling
I added a new tab to the options (Accessibility), in which is currently only one checkbox, which enables or disables the TTS. I thought there might be some more accessibility features added in the future (it's kind of mentioned in the GSoC task description), so this tab could be used for that. But maybe it might be a better idea to remove the Accessibility tab and move the checkbox to the Misc tab until there are more Accessibility options.

backends/text-to-speech/linux/linux-text-to-speech.cpp
```diff
@@ -0,0 +1,393 @@
...
+	char *conv_text = SDL_iconv_string("UTF-8", charset.c_str(), str.c_str(), str.size() + 1);
+	Common::String result;
+	if (conv_text) {
+		result = conv_text;
+		SDL_free(conv_text);
+	} else if (charset != "ASCII"){
+		warning("Could not convert text from %s to UTF-8, trying ASCII", charset.c_str());
+		return strToUtf8(str, "ASCII");
+	} else
+		warning("Could not convert text to UTF-8");
```
bluegr: This should be wrapped around a translation function

backends/text-to-speech/windows/sphelper-scummvm.h
```diff
@@ -0,0 +1,2648 @@
+/*******************************************************************************
+* SPHelper.h *
+*------------*
+*   Description:
+*       This is the header file for core helper functions implementation.
```
bluegr: Does this file have any changes from the original? If yes, which ones?

configure
```diff
@@ -165,6 +165,10 @@ _taskbar=auto
 _updates=no
 _libunity=auto
 _dialogs=auto
+_tts=auto
+_linux_tts=no
+_windows_tts=no
+_macosx_tts=no
```
bluegr: Why do you need three defines here? You could just combine one define with the target platform

backends/text-to-speech/windows/windows-text-to-speech.cpp
```diff
@@ -0,0 +1,470 @@
...
+	_voice->Resume();
+	DWORD threadId;
+	if (_thread != NULL) {
+		WaitForSingleObject(_thread, INFINITE);
+		CloseHandle(_thread);
+	}
+	_speechState = SPEAKING;
+	_thread = CreateThread(NULL, 0, startSpeech, &_threadParams, 0, &threadId);
+	if (_thread == NULL) {
+		warning("Could not create speech thread");
```
bluegr: This should be wrapped around a translation function

backends/text-to-speech/windows/windows-text-to-speech.cpp
```diff
@@ -0,0 +1,470 @@
...
+	WaitForSingleObject(*params->mutex, INFINITE);
+	if (*(params->state) != WindowsTextToSpeechManager::PAUSED)
+		*(params->state) = WindowsTextToSpeechManager::READY;
+	ReleaseMutex(*params->mutex);
+	return 0;
+}
+
+bool WindowsTextToSpeechManager::say(Common::String str, Action action, Common::String charset) {
+	if (_speechState == BROKEN || _speechState == NO_VOICE) {
+		warning("The tts cannot speak in this state");
```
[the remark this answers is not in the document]
criezy: We do not usually translate strings passed to warning() and error() as there might be an encoding issue with the console/terminal in which they will appear.

criezy:
I posted this in Discord/IRC last week as a teaser, but not everybody may have seen it. So here it is in action (on macOS):
TTS in the GUI: https://www.youtube.com/watch?v=z5T3Uh0zu1I
TTS in Mortville Manor: https://www.youtube.com/watch?v=DG00_a5qVKU&t=39s

bluegr:
Overall, pretty good work :) Well done!

@criezy: which speech engine is used in the videos?

criezy:
In the video it is using the macOS speech engine (the one that comes with the system).
@vyzigold has some slightly older recordings on both Linux and Windows in [his youtube channel](https://www.youtube.com/channel/UC8BkTcfa5zLGJaci-zrgiDA).

backends/text-to-speech/linux/linux-text-to-speech.cpp
```diff
@@ -0,0 +1,393 @@
...
+void LinuxTextToSpeechManager::setLanguage(Common::String language) {
+	if (_speechState == BROKEN)
+		return;
+	spd_set_language(_connection, language.c_str());
+	_ttsState->_language = language;
+	setVoice(_ttsState->_activeVoice);
+}
+
+void LinuxTextToSpeechManager::createVoice(int typeNumber, Common::TTSVoice::Gender gender, Common::TTSVoice::Age age, char *description) {
+	SPDVoiceType *type = (SPDVoiceType *) malloc(sizeof(SPDVoiceType));
```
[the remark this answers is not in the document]
vyzigold: Because I need to pass the vaule to the TTSVoice class as it's data field, which is defined as void *

The data field in the TTSVoice is meant for platform specific information about each voice. The WindowsTextToSpeechManager for example saves there a pointer to a whole class ISpObjectToken.

So because I needed a pointer to the value and I can't just pass a pointer to the stack, because the TTSVoice that is being created in this function outlives this function, I alocate space for it on the heap and copy it there.

backends/platform/sdl/win32/win32_wrapper.cpp
```diff
@@ -81,7 +81,7 @@ bool confirmWindowsVersion(int majorVersion, int minorVersion) {
 	return VerifyVersionInfoFunc(&versionInfo, VER_MAJORVERSION | VER_MINORVERSION, conditionMask);
 }
 
-wchar_t *ansiToUnicode(const char *s, uint codePage) {
+wchar_t *ansiToUnicode(const char *s, unsigned int codePage) {
```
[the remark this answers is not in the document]
vyzigold: It didn't know the type when compiling with Visual Studio. I guess the scummsys.h wasn't included before this. But the fix seemed pretty easy (uint and unsigned int should be the same), so I changed it and didn't worry about it.

Is this something, I should change back?

backends/text-to-speech/windows/sphelper-scummvm.h
vyzigold: It has changes, I had to make them, so it works with MinGW. I added a summary of the changes to the top of the file. I think documenting each changed line would be too much and it would end up being quite confusing when reading it.

vyzigold:
@bluegr Thank you for the review, I think I addressed all of the issues you pointed out. As for the recordings on my youtube channel, even though they are older, everything on the videos should still be working the same.

backends/platform/sdl/win32/win32_wrapper.cpp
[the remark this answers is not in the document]
bluegr: Yes, please try and use the ScummVM specific types, they are defined appropriately for each platform we support

backends/text-to-speech/linux/linux-text-to-speech.cpp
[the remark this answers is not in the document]
bluegr: Thanks for the explanation. Could you please add a corresponding comment to the code?

backends/text-to-speech/windows/sphelper-scummvm.h
bluegr: Thanks

vyzigold: It is added by this commit: [e21ebee](https://github.com/scummvm/scummvm/pull/1808/commits/e21ebeeee7d8cb908d744febc97c6263f48c312d)

vyzigold:
@bluegr I added more documentation to the TTSVoice and a comment to the malloc. I also changed the unsigned int back to uint in the win32_wrapper. I hope everything is allright now.

bluegr:
Many thanks! This looks OK now, well done, and nice work overall!

+1 from me

bluegr:
One more question: I do not understand why LinuxTextToSpeechManager has been renamed to SpeechDispatcherManager. Now its name is different from the other two (the Windows and the MacOS one).

Could someone explain the rationale here?

criezy:
@bluegr I suggested to rename it as this specific implementation uses the speech dispatcher library, but we could envisage other implementations on Linux. This is a bit similar to the `UnityTaskbarManager` having been named as it is and not `LinuxTaskbarManager`. On the other hand the macOS and Windows implementations use the TTS API from the system rather than a third party library.

bluegr:
@criezy: thanks for the explanation :)

bluegr:
This is now in a good state to be merged to master, IMHO. Is there any other pending issue? If not, I would like to have this merged by next weekend (Sunday, 1st of September)

vyzigold:
It has merge conflicts because of the Common::Encoding PR #1809. And since the Common::Encoding is merged and the TTS needs to convert between encodings, I would like to use that. After that, I don't know of any issue.

I will resolve the conflicts and reimplement the encoding conversions using Common::Encoding tomorrow, since today I am quite busy.

vyzigold:
I rebased the branch on top of the current master. I replaced the character encoding conversion code with the Common::Encoding::convert(), which should support more encodings. And because this pull request adds speech synthesis to the mortevielle engine, which uses CP850 character encoding and Common::Encoding didn't support this encoding when compiling without iconv and using SDL, which is also compiled without iconv, I added a CP850 to unicode (and back) conversion table to the Common::Encoding.

bluegr:
Apart from codepage 850, there’s also 437. Perhaps this should be supported natively as well?

https://en.m.wikipedia.org/wiki/Code_page_437

neuromancer:
is it planed to offer several voices for different "actors" in the game engines? (if most engines allow such distinction)

vyzigold:
Yes, but it is not guaranteed, there might not be any voice available in some situations. This is why in the mortevielle engine we use as many voices as are available, but we also change the pitch a little bit for each character (so the voice sounds at least a little bit different, if there are not enough voices for each character). We also change the pitch a lot more, if we have to use a voice with a different gender, than the character. So if for example there are only male voices available, but we have a female character, we use a male voice, but we set the pitch really high.

bluegr:
Great work! This can be merged now

digitall:
@vyzigold : Thanks for the work on TTS. However, this has broken the OSX Intel build on the buildbot. I think the issue is that the SDK we use is for OSX 10.4 which is the earliest version we support and thus the TTS API may have changed or be not present in that version. Any ideas, apart from just dropping support for earlier OSX versions which is not optimal...
See following for details:
https://buildbot.scummvm.org/builders/master-osx_intel/builds/1896/steps/compile/logs/stdio
https://wiki.scummvm.org/index.php?title=Mac_OS_X

criezy:
Indeed, some of the API I used to implement TTS on macOS were introduced in OSX 10.5. Since this is the minimal version supported for the official releases, I did not think buildbot was using an older SDK (the PPC official release does support OS X 10.4 though, but buildbot doesn't have a OS X PPC build).

I can look at trying to fix this so that we have TTS support on OS X 10.4. But my suggestion would be instead to:
- Add a configure test to disable TTS support on older OS X versions.
- Update the toolchain and SDK on buildbot to use the 10.5 SDK. Bringing it in line with what I am using for the official releases would be a good idea anyway I think. I have also been thinking of adding the toolchain for the 64 bits / OS X 10.7 builds).
- In the meantime we could explicitly disable TTS for the macOS build on buildbot (with `--disable-tts`).

vyzigold:
@digitall I noticed, that the OSX build is broken by this pull request, but because I don't have anything with OSX and I have never seen any Objective-C++ code before this GSoC, I wasn't able to do anything about it. I hope criezy manages to fix the code, so it works with the 10.4 SDK, otherwise it will probably have to be disabled by the configure script for older versions.

digitall:
@vyzigold : Ah OK. No problem then. I am sure @criezy and I can do something to sort out the OSX builds on buildbot, so not a problem for you.

digitall:
@criezy: I think it would be best to bring the buildbot up to date to 10.5 SDK as per your release builds and add a 64-bit OSX Intel build since we are already getting bug reports from the OSX 10.15 dropping of 32-bit build support: https://bugs.scummvm.org/ticket/11136

PR #1812, opened 2019-08-18
COMMON: added support for reading ini files with different 8bit encod…

whiterandrek:
This pull request adds support for reading ini files with encodings like CP-1250, CP-1251, CP-1252, etc.
Currently, those ini files can't be read because isAlnum function returns zero for chars which are not standard ASCII characterts.

### DIFF
--- a/common/ini-file.cpp
+++ b/common/ini-file.cpp
@@ -29,10 +29,16 @@
 namespace Common {
 
 bool INIFile::isValidName(const String &name) const {
+	if (_allowNonEnglishCharacters)
+		return true;
 	const char *p = name.c_str();
 	while (*p && (isAlnum(*p) || *p == '-' || *p == '_' || *p == '.' || *p == ' '))
 		p++;
 	return *p == 0;
+}
+
+INIFile::INIFile() {
+	_allowNonEnglishCharacters = false;
 }
 
 void INIFile::clear() {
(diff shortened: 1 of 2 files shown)


common/ini-file.cpp
```diff
@@ -36,6 +36,7 @@ bool INIFile::isValidName(const String &name) {
 }
 
 INIFile::INIFile() {
+	_8bitMode = false;
```
bluegr: The name is a bit confusing. It could be renamed to something like _allowNonEnglishCharacters

common/ini-file.cpp
```diff
@@ -125,7 +126,7 @@ bool INIFile::loadFromStream(SeekableReadStream &stream) {
 			section.comment = comment;
 			comment.clear();
 
-			assert(isValidName(section.name));
+			assert(_8bitMode || isValidName(section.name));
```
bluegr: It would be better to extend the functionality of isValidName(), instead of short-circuiting its usage all over the place

whiterandrek:
@bluegr 
Because isValidName is static method.

bluegr:
Hm, I don’t remember why it’s static. Will need to check the code later on from a PC

bluegr:
I've adapted the usage of isValidName(), and it's no longer static, either. You will need to redo your changes, unfortunately

bluegr:
Great, thanks! Merging

PR #1816, opened 2019-08-21
GRAPHICS: Display Mac monochrome cursor inverted pixels

sluicebox:
This adds limited support for Macintosh monochrome cursors that have inverted pixels. Currently ScummvM treats these as transparent and so FPFP and KQ6 are missing cursor features, bug #7050.

Mac monochrome cursors have a 1 bit per pixel image and mask. Bits that are set in the image but not the mask are a fourth color other than black, white, or transparent: the underlying target pixel would be inverted. The knob on the classic mac watch cursor is an example of this:

watch

The Waiting cursor from FPFP also does this:

fpfp_wait

This PR does not add inversion. That would change the cursor interface layers, and as I understand it, require implementing inversion in each backend. That's a lot of files to touch to make a handful of mac cursors look better and I wouldn't be able to test most of those changes.

Instead, this small change to MacCursor allows an engine to specify an optional color to use for monochrome inverted pixels. For SCI that's black. FPFP and KQ6 are the only two SCI games I'm aware of that have mac monochrome cursors. Of those, only ten cursor even contain inverted pixels and most just have one or two pixels so it's indistinguishable either way, but the rest are improved by restoring their missing features:

curs

This is opt-in behavior, this PR doesn't change other engines. The full inversion effect could be added later now that the issue is known.

### DIFF
--- a/engines/sci/graphics/cursor.cpp
+++ b/engines/sci/graphics/cursor.cpp
@@ -508,7 +508,10 @@ void GfxCursor::kernelSetMacCursor(GuiResourceId viewNum, int loopNum, int celNu
 	Common::MemoryReadStream resStream(resource->toStream());
 	Graphics::MacCursor *macCursor = new Graphics::MacCursor();
 
-	if (!macCursor->readFromStream(resStream)) {
+	// use black for mac monochrome inverted pixels so that cursor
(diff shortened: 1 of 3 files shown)


bluegr:
Looks good! Well done

PR #1817, opened 2019-08-22
AUDIO: (opl2lpt) add OPL3LPT support

gnustomp:
The OPL3LPT is like the OPL2LPT with an OPL3 chip.  Add OPL3 support to the existing OPL2LPT driver.

There is no auto-detection of OPL3 capability. The user is required to explicitly select OPL3LPT as their OPL device.

Tested with Sam & Max Hit the Road.


bluegr:
Nice and cleanly written!

bluegr:
Thanks for your work! Merging

PR #1818, opened 2019-08-24
SDL: Use a non-const string for SDL_iconv_string

vyzigold:
With some older versions of SDL1, the SDL_iconv_string takes char * instead of const char * as it's argument. This should fix the build issue with gp2xwiz.


digitall:
This looks fine and self-contained, so merging to see if this fixes the builds...

PR #1819, opened 2019-08-25
Common: Fix Encoding building

vyzigold:
This should fix issues with building on the ds and osx_intel.

I tryed to include config.h, so I have the USE_ICONV define available and can include iconv.h before including the scummsys.h, thinking, that forbidden.h wouldn't allow me to include the iconv.h. This wasn't necessary and it broke the ds build, so I don't do that anymore.

I looked at how the ResidualVM works with iconv when implementing Encoding. They use a define ICONV_USES_CONST to determine the signature of iconv() function. I thought this is defined in iconv.h and used it. Now I noticed, they test this in their configure and define it by themselves. I added this test and define to configure. This should fix the osx_intel build.

### DIFF
--- a/common/encoding.h
+++ b/common/encoding.h
@@ -23,19 +23,15 @@
 #ifndef COMMON_ENCODING_H
 #define COMMON_ENCODING_H
 
-#ifdef HAVE_CONFIG_H
-#include "config.h"
-#endif // HAVE_CONFIG_H
+#include "common/scummsys.h"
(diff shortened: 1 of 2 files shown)


digitall:
This looks fine, though I think the formatting in the configure script cut and paste might be slightly off... but will fix that in a further commit after merge if need be.

@vyzigold : Thanks for this.

PR #1820, opened 2019-08-25
TESTBED: Allow forbidden symbols

tsoliman:
Since Testbed now uses backend code to test conversions, it needs to allow
forbidden symbols.

See the following comment for the rationale on why the backend code is used: https://github.com/scummvm/scummvm/pull/1809#issuecomment-522120233

### DIFF
--- a/common/encoding.h
+++ b/common/encoding.h
@@ -23,15 +23,19 @@
 #ifndef COMMON_ENCODING_H
 #define COMMON_ENCODING_H
 
-#include "common/scummsys.h"
-#include "common/str.h"
-#include "common/system.h"
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif // HAVE_CONFIG_H
 
 #ifdef USE_ICONV
 #include <iconv.h>
 #else
 typedef void* iconv_t;
 #endif // USE_ICONV
+
+#include "common/scummsys.h"
+#include "common/str.h"
+#include "common/system.h"
 
 #ifdef WIN32
 #include "backends/platform/sdl/win32/win32.h"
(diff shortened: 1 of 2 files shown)


tsoliman:
I force pushed the same change to encoding.cpp as the latest batch of changes caused that file to fail.

Here's that include chain
```
In file included from ../engines/testbed/encoding.cpp:23:
In file included from ../common/encoding.h:31:
In file included from /opt/local/include/iconv.h:110:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/wchar.h:119:```

tsoliman:
It seems reverting 96fef166 was also required to make the build succeed.

tsoliman:
Having reverted 96fef16, the changes to engines/testbed/encoding.cpp were not necessary.

tsoliman:
Further, it seems that reverting 96fef16 was necessary to fix the compile even without TESTBED being enabled at all. Reverting it might be outside the scope of this PR but its spirit is really "Fix the build" so I'll leave it in for now.

lotharsm:
I can confirm that applying this patch unbreaks the current master branch!

tsoliman:
I talked to @vyzigold and they said they'd fix it in a more correct way.

sev-:
Shall I wait for @vyzigold? Otherwise, this exception is perfectly acceptable and is good to merge

vyzigold:
I think I should have it fixed. https://github.com/vyzigold/scummvm/tree/encoding I just want to test it on windows first. If it works, I will create a pull request with it (it works on my linux, but I didn't have the issue with forbidden symbols there).

bluegr:
The other pull request (#1822) has been merged. Closing this one

PR #1821, opened 2019-08-25
CLOUD: Treat HTTP headers names case-insensitively and add more verbose error messages

Tkachov:
wolfgangr had an issue with Google Drive he told me on IRC. We spent a couple hours debugging and found out that his machine (Debian x64, though it's probably not the reason) gets HTTP headers in lowercase (while I always seen them coming in titlecase). So, as Google Drive file uploading routine in ScummVM expected header in titlecase, it didn't work in his environment. Empty error messages along with 200 OK HTTP code were not helping too.

So, I've added a new method to handle HTTP headers case-insensitively (more or less matching RFC) and updated all places that had empty error messages to provide at least some information.

### DIFF
--- a/backends/cloud/box/boxlistdirectorybyidrequest.cpp
+++ b/backends/cloud/box/boxlistdirectorybyidrequest.cpp
@@ -83,7 +83,7 @@ void BoxListDirectoryByIdRequest::responseCallback(Networking::JsonResponse resp
 	if (response.request)
 		_date = response.request->date();
 
-	Networking::ErrorResponse error(this);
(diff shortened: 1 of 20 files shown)


bluegr:
Nice work! The new HTTP header handling code is written in a much cleaner way with the usage of Common::HashMap, and the more verbose error messages will certainly aid in any future issues.

Merging

Mataniko:
Just be aware that it's acceptable to have a more than one header with the same name, so while this works now it may cause issues in the future.

PR #1822, opened 2019-08-25
COMMON: Don't include iconv.h in common/encoding.h

vyzigold:
Move #include<iconv.h> from common/encoding.h to
common/encoding.cpp and change the methods accordingly.

This resulted in not saving the iconvHandle if using the
"non-static" version of conversion, but it simplified the code
and hopefuly resolved issues with forbidden symbols on some
platforms.

I tested this on arch linux (but there was no issue with forbidden
symbols before) and on Windows with MinGW and MinGW-w64
(this had the issue and now the issue is gone). Unfortunately
I cannot test this with MacOS.

### DIFF
--- a/common/encoding.cpp
+++ b/common/encoding.cpp
@@ -20,6 +20,7 @@
  *
  */
 
+#define FORBIDDEN_SYMBOL_ALLOW_ALL
 #include "common/encoding.h"
 #include "common/textconsole.h"
 #include "common/system.h"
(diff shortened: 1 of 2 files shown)


bluegr:
Well done! The new code is much cleaner and simpler. Merging

vyzigold:
Well, I hope it finaly compiles on every platform.

PR #1826, opened 2019-08-31
BUILD: Fix GCC bogus warning by removing const

lephilousophe:
GCC emits "warning: type qualifiers ignored on cast result type" when
some types are used for HashMap because of the const modifier.
This applies especially to pointer-to-member-functions.
The const there is not really useful as it only prevents unfortunate
object editions.

### DIFF
--- a/common/hashmap.h
+++ b/common/hashmap.h
@@ -114,7 +114,7 @@ private:
 #endif
 
 	/** Default value, returned by the const getVal. */
-	const Val _defaultVal;
+	Val _defaultVal;
 
 	Node **_storage;	///< hashtable of size arrsize.
 	size_type _mask;		///< Capacity of the HashMap minus one; must be a power of two of minus one


bluegr:
It would be preferable to remove the leading const from getVal(): the compiler will throw a warning if this is used with simple types

digitall:
@bluegr: I tried removing the leading const from various declarations and templates for the getVal, but I kept getting compiler errors... If you can suggest a better solution as a patch or gist, then I will test. Thanks.

lephilousophe:
@bluegr : You can't remove ``const`` from ``getVal`` return type because it's a reference to content in ``HashMap`` and these functions are constant ones.
The non constant version of ``getVal`` don't need the ``_defaultValue`` because they simply add a new entry to the ``HashMap``

bluegr:
I'll check this again today or tomorrow - need to be at a PC with the appropriate development tools installed

bluegr:
Since my free time is quite limited these days, I'm merging this, and will work on additional enhancements at some other point in the future. Thanks for your work!

PR #1828, opened 2019-09-01
COMMON: Don't include win32.h in common/encoding.h

SupSuper:
Currently encoding.h includes win32.h to expose `switchEndian` to the backend. This has the nasty side effect of exposing poor defenseless engines to the cthulhuian horrors of Win32. In this case, Testbed doesn't compile on MSVC because Win32 requires language extensions.

I replaced the include with just making `switchEndian` public, since it's static anyways, so a friend class seems excessive. @vyzigold let me know if this is ok.

### DIFF
--- a/backends/platform/sdl/win32/win32.h
+++ b/backends/platform/sdl/win32/win32.h
@@ -27,7 +27,6 @@
 #include "backends/platform/sdl/win32/win32-window.h"
 
 class OSystem_Win32 : public OSystem_SDL {
-	friend class Common::Encoding;
 public:
 	virtual void init();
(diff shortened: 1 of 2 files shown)


vyzigold:
I didn't want the method to be public inside the Common::Encoding, because the method doesn't do anything with character encoding, so that is why I made it like this, but I didn't test this with MSVC, only with MinGW. Your solution should be ok.

sluicebox:
Confirmed that this fixes VS2015 and VS2019

bluegr:
Great work! Merging

PR #1829, opened 2019-09-02
MADE: Add Manhole detect and separate out detections list

dafioram:
1. Add an unstable manhole version (gog)
2. Move MADE detection list into separate file.

I tested the changes by running return to zork.


bluegr:
The Masterpiece Edition is not based on the MADE engine, but on another, HyperCard-like engine:
https://wiki.scummvm.org/index.php/The_Manhole

Thus, please do not add this version in this engine

dafioram:
scummvm autodetected this as MADE based on the dat file, is there a good way to inform the user that this version shouldn't be added as I have done? I imagine someone in the future will do the same as I have done.

bluegr:
You can comment this version out, and add a comment that it's not supported, since it's not using the MADE engine. This way, people won't try to re-add it

dafioram:
How that look?

bluegr:
Looks good, thanks! :) Merging

PR #1831, opened 2019-09-05
TTS: Fix crashing and freezing when compiled with msvc

vyzigold:
This pull request resolves crashing when compiling with msvc and using text to speech. The crashes were actualy caused by Common::Encoding, that returned a converted string from SDL_iconv_string(), which couldn't be freed by free(), while strings returned by Common::Encoding should be freed by free().

This also fixes the freezing caused by waiting on a thread while trying to stop speech (the freeze could last up to multiple seconds). There still is a barely noticable lag when using TTS in the GUI and moving fast with the cursor between multiple GUI widgets and thus starting and stoping multiple speeches too quickly. This is caused by a call to ISpVoice->Speak() inside the stop() method. I haven't found a better solution to stopping the speech yet, but the lag shouldn't be noticable with TTS in games (like in Mortevielle) and people with reduced sight (which should be the people, that will be using the TTS in GUI) also shouldn't notice.

### DIFF
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -28,6 +28,7 @@
 #include "common/taskbar.h"
 #include "common/textconsole.h"
 #include "common/translation.h"
+#include "common/encoding.h"
 
 #include "backends/saves/default/default-saves.h"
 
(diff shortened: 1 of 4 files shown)


criezy:
Thank you. Can you please squash commits aa5ab6f and 1420b84 together? Maybe also add a comment just before the new code that makes the copy to indicate why it is necessary.
After that I think it will be ready to be merged.

vyzigold:
The commits are squashed and I added an explanation for the copy.

PR #1835, opened 2019-09-06
BUILD: Adds -static flag when --enable-static under mingw

Joefish:
System: Win10 Version 1809, WSL (10.0.17763.615), GCC 9.2.0 (mingw), ld 2.32 (mingw)
When I compile scummvm it doesn't statically link the dependent libraries although I set
--enable-static in the configure step. Adding the -static flag to the linker fixes the issue.

### DIFF
--- a/configure
+++ b/configure
@@ -2831,6 +2831,9 @@ case $_host_os in
 		append_var DEFINES "-DWIN32"
 		# append_var DEFINES "-D__USE_MINGW_ANSI_STDIO=0"  # Modern MinGW does not need it
 		append_var LDFLAGS "-static-libgcc -static-libstdc++"
+		if test "$_static_build" = yes ; then
+			append_var LDFLAGS "-static"
+		fi
 		append_var LIBS "-lmingw32 -lwinmm -lgdi32"
 		append_var OBJS "dists/scummvm.o"
 		add_line_to_config_mk 'WIN32 = 1'


Joefish:
I talked with others on discord (rootfather, supsuper) on discord about it and I seem to be the only one so far with the issue. Nevertheless, I thought raising the issue is only appropriate once I hit it.
If changing the configure script doesn't seem necessary to the reviewers, adding a notice to the compile section in the wiki would be the next best thing so others don't waste their time trying to get it compile correctly.

sluicebox:
Which dll imports appear in your exe that you don't expect to be?

I can't tell from the discord log and I wonder if everyone was talking about the same thing. msvcrt.dll? game engines as plugins? something else?

Without modifying anything I get a scummvm.exe with engines statically compiled that imports msvcrt.dll and SDL2.dll and that's it beyond OS dlls.

lephilousophe:
> Without modifying anything I get a scummvm.exe with engines statically compiled that imports msvcrt.dll and SDL2.dll and that's it beyond OS dlls.

@sluicebox: You are talking there about MSVC while @Joefish talks about mingw. Aside from that, the package of precompiled libraries provided on the wiki only has static versions of all the dependencies except for SDL2 and WinSparkle.
That's why you don't have any problem like @Joefish.

For the docker version of the toolchain, I did the same thing as csnover did. I just compile static versions of libraries (by using --disable-shared configure flag on all of them) except for SDL2 (built with DLL) and WinSparkle (downloaded precompiled).

Anyway, using -static linker flag will prevent linking with dynamic SDL2 and WinSparkle and that may not be something we want. That must be settled before accepting this PR.

Joefish:
I just read the [compile guide for MSYS2](https://wiki.scummvm.org/index.php?title=Compiling_ScummVM/MinGW-w64#Common_Issues) again and found under Common Issues my issue of not finding the dlls (so adding another entry as I proposed before would be redundant). For now I either compile it with -static or launch scummvm from a batch file that adds x86_64-w64-mingw32/bin to windows PATH.

Nevertheless, the description of --enable-static is not accurate for mingw ("build a static binary instead of using shared objects"). I'm not sure if fixing this and potentially breaking compilation of other builds in the process is worth it since apparently I'm the only one having this issue so far.
Personally I would be fine closing this PR for now until this actually becomes an issue.

sluicebox:
Yes, I was elaborating on "works for me" by showing that my vanilla mingw setup produces an exe with the same imports as buildbot's. (Though I'm not doing MinGW-w64 or going through WSL)

I think I have my answers though, thanks

bluegr:
Closing this, since it breaks linking with the dynamic DLLs of SDL and WinSparkle, as mentioned.

SDL discourage static linking:
https://wiki.libsdl.org/Installation#Static_linking

Therefore, it would be preferable if SDL and WinSparkle are excluded from static linking. But the OP mentioned he won’t be doing any further work on this. Thus, closing

PR #1836, opened 2019-09-09
IOS: Call exit(0) when exiting to prevent hanging

sluicebox:
When quitting the ios7 backend, either through the launcher's Quit button or through a game's Quit function, the process always hangs instead of exiting. From the launcher I get a black screen, in the SCI games it hangs on the game frame. At that point I have to force kill the process.

I don't know why it isn't exiting cleanly, but the existing TODO comments suggests this wasn't quite finished. This has always happened to me on my non-jailbroken iphone se throughout its major ios versions. Seems like throwing an exit(0) at it is a good fix until the root cause is discovered, if it even matters. iOS7_main() finishes but UIApplicationMain() isn't returning, and there's no further code after that.

I have an iphone 4 running ios4 with the old 1.5 scummvm on it, and it exited cleanly, so I don't know if this problem was introduced with the ios7 backend or in a new ios version. (In my experience, probably the latter.)

### DIFF
--- a/backends/platform/ios7/ios7_osys_main.cpp
+++ b/backends/platform/ios7/ios7_osys_main.cpp
@@ -414,4 +414,7 @@ void iOS7_main(int argc, char **argv) {
 		//*stderr = NULL;
 		fclose(newfp);
 	}
+
+	// prevents hanging on exit
+	exit(0);
 }


bluegr:
Calling exit() is discouraged by Apple. Have a look here:
https://stackoverflow.com/questions/13843774/ios-is-exit0-deprecated
and here:
https://developer.apple.com/library/archive/qa/qa1561/_index.html

sluicebox:
Uh oh... there goes our spot in the app store! =)

I'm familiar with Apple's position, but it's that ending the process at all is bad, and that's already what the ios version does and has always done. It's already doing what Apple doesn't want by returning from its entry point. Adding exit() just makes difference between that succeeding or failing, and it used to succeed. The ios port already calls exit() in other codepaths.

In other words, just having a Quit button makes Apple cry, but I don't know what to do about that, so it might as well work.

Kawa-oneechan:
I guess you'd just have to...

...Think Different™ 😎

bluegr:
Indeed, we are using exit() in other places in the IOS backend, and at the point where this is called (the end of the main endpoint), it's an expected call. Since this fixes the application hang, and it's a minimal change, it's OK to add it, till the situation is properly addressed in the future

PR #1837, opened 2019-09-09
Update coktel_decoder.cpp

Dimouse:
fix Woodruff subtitles according to https://bugs.scummvm.org/ticket/10960#ticket

```

two together.

- Bullet points are okay, too

- Use a hanging indent
```

### DIFF
--- a/video/coktel_decoder.cpp
+++ b/video/coktel_decoder.cpp
@@ -2071,9 +2071,9 @@ bool VMDDecoder::readFrameTable(int &numFiles) {
 				_frames[i].parts[j].flags   = _stream->readByte();
 
 			} else if (_frames[i].parts[j].type == kPartTypeSubtitle) {
-				_frames[i].parts[j].id = _stream->readUint16LE();
+				_frames[i].parts[j].id = _stream->readByte();
 				// Speech text file name
-				_stream->skip(8);
+				_stream->skip(9);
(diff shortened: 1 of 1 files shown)


bluegr:
Has this change been tested with other games that use the Coktel Video Decoder? (e.g. SCI games, other gob games apart from Woodruff)

Dimouse:
> Thanks for your work! :)
> 
> Has this change been tested with other games that use the Coktel Video Decoder? (e.g. SCI games, other gob games apart from Woodruff)

Hi! Thank you for the reply:)
We didn't check all the games, but it seems that the only game that uses vmd with external subtitles is Woodruff.
There is the list of games here: https://wiki.multimedia.cx/index.php/VMD#Games_Using_VMD

Betrayal In Antara, The Last Dynasty, Shivers 2: Harvest of Souls and SWAT are not supported.
Gabriel Knight 2: The Beast Within, Lighthouse, Phantasmagoria, RAMA and Shivers doesn't have subtitles. Leisure Suit Larry: Love for Sail! and Torin's Passage store subtitles in msg files, and this is probably the case in other SCI games also.
In the gob engine some games are not supported, some are too old for vmd videos and some have other formats (for example, Urban Runner). 

But of course if we need to be save, we can check the game explicitly in the code?

bluegr:
Thanks for the feedback, and the work you did by checking the videos of these games. Thus, this can be merged safely.

Thanks for your work!

PR #1840, opened 2019-09-13
EDITING: more consistent errors / saves => saved games

BenCastricum:


engines/engine.cpp
```diff
@@ -326,18 +326,16 @@ void initGraphics(int width, int height, const Graphics::PixelFormat *format) {
 #endif
 
 	if (gfxError & OSystem::kTransactionModeSwitchFailed) {
-		Common::String message = _("Could not switch to video mode: '");
-		message += ConfMan.get("gfx_mode");
-		message += "'.";
+		Common::String message;
+		message = Common::String::format(_("Could not switch to video mode '%s'."), ConfMan.get("gfx_mode"));
```
criezy: This is invalid. `ConfMan.get()` returns a `Common::String`, while variadic functions, such as `format()` can only take POD arguments (some compilers may accept non-POD arguments, but this is not portable).

You can use `ConfMan.get("gfx_mode").c_str()`, which returns a `char *`, to fix the issue.

engines/engine.cpp
```diff
@@ -326,18 +326,16 @@ void initGraphics(int width, int height, const Graphics::PixelFormat *format) {
...
 		GUI::MessageDialog dialog(message);
 		dialog.runModal();
 	}
 
 	if (gfxError & OSystem::kTransactionStretchModeSwitchFailed) {
-		Common::String message = _("Could not switch to stretch mode: '");
-		message += ConfMan.get("stretch_mode");
-		message += "'.";
+		Common::String message;
+		message = Common::String::format(_("Could not switch to stretch mode '%s'."), ConfMan.get("stretch_mode"));
```
criezy: Same remark as above.

criezy:
I have pushed your two commits manually after making the necessary modifications.

PR #1844, opened 2019-09-16
Revert "WIN32: Fix getting the parent directory of non-existant file nodes"

Tkachov:
As this commit didn't fix anything, I think we should revert it just in case, and keep this assert().


sev-:
Thanks!

PR #1846, opened 2019-09-18
NETWORKING: Convert translation results to UTF-8 for local webserver pages

criezy:
Translating the text for the local web server is not working correctly when using non-ASCII characters as the html pages expect UTF-8. In commit 0c605e3 I played with the idea of injecting the current GUI encoding into the html pages instead of specifying utf-8. However I think that it might be better instead to keep utf-8 encoding for the html pages and convert all the text to utf-8, as it might work better for file lists. So this is what I am proposing in this PR.

I have however an issue with the AJAX version of the file list: I cannot get it to properly handle UTF-8. I am not familiar with AJAX, so if anybody has an idea how to fix this, please let me know.

Here is what it looks like with this PR when displaying a page in French using the AJAX handler:
image

And here is what it looks like when using the pure html handler:
image

So in the HTML case not only the translated text, but also the directory names (such as Croisière in the image above), are displayed correctly. But in the AJAX case they are not.

And I get the same behaviour with both Chrome and Safari, so this is not an issue specific to a particular web browser.

### DIFF
--- a/backends/networking/sdl_net/getclienthandler.cpp
+++ b/backends/networking/sdl_net/getclienthandler.cpp
@@ -107,7 +107,7 @@ const char *GetClientHandler::responseMessage(long responseCode) {
 
 void GetClientHandler::prepareHeaders() {
 	if (!_specialHeaders.contains("Content-Type"))
-		setHeader("Content-Type", "text/html");
+		setHeader("Content-Type", "text/html; charset=UTF-8");
 
 	if (!_specialHeaders.contains("Content-Length") && _stream)
 		setHeader("Content-Length", Common::String::format("%u", _stream->size()));
--- a/backends/networking/sdl_net/handlers/createdirectoryhandler.cpp
+++ b/backends/networking/sdl_net/handlers/createdirectoryhandler.cpp
@@ -57,35 +57,35 @@ void CreateDirectoryHandler::handle(Client &client) {
 
 	// check that <path> is not an absolute root
 	if (path == "" || path == "/") {
-		handleError(client, _("Can't create directory here!"));
+		handleError(client, HandlerUtils::toUtf8(_("Can't create directory here!")));
 		return;
 	}
 
 	// check that <path> contains no '../'
 	if (HandlerUtils::hasForbiddenCombinations(path)) {
-		handleError(client, _("Invalid path!"));
+		handleError(client, HandlerUtils::toUtf8(_("Invalid path!")));
 		return;
 	}
 
 	// transform virtual path to actual file system one
 	Common::String prefixToRemove = "", prefixToAdd = "";
 	if (!transformPath(path, prefixToRemove, prefixToAdd) || path.empty()) {
-		handleError(client, _("Invalid path!"));
+		handleError(client, HandlerUtils::toUtf8(_("Invalid path!")));
 		return;
 	}
 
 	// check that <path> exists, is directory and isn't forbidden
 	AbstractFSNode *node = g_system->getFilesystemFactory()->makeFileNodePath(path);
 	if (!HandlerUtils::permittedPath(node->getPath())) {
-		handleError(client, _("Invalid path!"));
+		handleError(client, HandlerUtils::toUtf8(_("Invalid path!")));
 		return;
 	}
 	if (!node->exists()) {
-		handleError(client, _("Parent directory doesn't exists!"));
+		handleError(client, HandlerUtils::toUtf8(_("Parent directory doesn't exists!")));
 		return;
 	}
 	if (!node->isDirectory()) {
-		handleError(client, _("Can't create a directory within a file!"));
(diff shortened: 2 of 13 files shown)


SupSuper:
I think file paths need their own conversion, since they use the system locale. I tried your branch on Windows and got different symbols on the filenames in HTML and AJAX, neither of them correct. On Linux it probably happens to work because the system locale is usually UTF-8.

The AJAX bug seems to be in the ScummVM JSON encoder, as it's mangling any non-ASCII characters.

criezy:
Yes, the path probably needs their own conversion, but that would require more work as we would need to know what encoding the system is using (and I am lucky enough to be on a system using utf-8 ;-) ). Also this is a secondary priority as the issue would only happen if the users is using non-ascii characters for his directory names, and those can always be renamed. In this PR I would like to focus on making sure that the translation of the pages is properly rendered.

And I suspect you are right with the JSON encoder being the issue. For example the following code that handles `\uXXXX` in `bool JSON::extractString(const char **data, String &str)` seems highly suspect since `next_char` is a `char` and probably overflows.
```
next_char = 0;
for (int i = 0; i < 4; i++) {
	// Do it first to move off the 'u' and leave us on the
	// final hex digit as we move on by one later on
	(*data)++;

	next_char <<= 4;

	// Parse the hex digit
	if (**data >= '0' && **data <= '9')
		next_char |= (**data - '0');
	else if (**data >= 'A' && **data <= 'F')
		next_char |= (10 + (**data - 'A'));
	else if (**data >= 'a' && **data <= 'f')
		next_char |= (10 + (**data - 'a'));
	else {
		// Invalid hex digit = invalid JSON
		return false;
	}
}
```

I will take a look at that this evening.

SupSuper:
> Yes, the path probably needs their own conversion, but that would require more work as we would need to know what encoding the system is using (and I am lucky enough to be on a system using utf-8 ;-) ). Also this is a secondary priority as the issue would only happen if the users is using non-ascii characters for his directory names, and those can always be renamed. In this PR I would like to focus on making sure that the translation of the pages is properly rendered.

Gotcha, I figured it might be a solved problem since the GUI browser shows filesystem names, but if it also just "hopes for the best", might be a good source for a future PR. :-)

> And I suspect you are right with the JSON encoder being the issue. For example the following code that handles `\uXXXX` in `bool JSON::extractString(const char **data, String &str)` seems highly suspect since `next_char` is a `char` and probably overflows.

You're right, the problem is that `\uXXXX` are 32-bit Unicode codepoints, and the code is trying to treat them as 8-bit chars. Not only is this incorrect, it's impossible to do correctly without knowing the encoding of the string. It's probably safer to just leave non-ASCII characters as-is and let the calling code deal with the encoding (JSON should be UTF-8 but we can't enforce that from within).

Removing this conversion fixes the AJAX problem (every non-ASCII was turned into `\uFF00`:
https://github.com/scummvm/scummvm/blob/b19de22053932f61d73e2cd6a17f429000ce7bc3/common/json.cpp#L1057-L1067

But since this is common code, I don't know if anything else is depending on this behavior (since it's buggy, I sincerely hope not).

criezy:
I have now pushed a commit that attempt to fix the UTF-8 string escaping and parsing in the JSON parser.

I also force pushed a change as I forgot to convert some strings to UTF-8 in my previous commit from yesterday.

And now it all works properly for me! :-)

image
(this is the AJAX version)

criezy:
Note that if the utf-8 decoding fails when escaping characters in the JSON parser, it attempt to recover by just keeping the raw character. So it may still work, but it will print a warning in the console. I tested this with a ISO-8859-1 string (involuntarily; that's when I noticed I had forgotten to convert some strings to utf-8 yesterday).

bluegr:
Nice work! :)
This is in a good state now, bit I believe we should wait for the 2.1.0 release before merging it

bluegr:
On second thought, we’ve recently added the cloud support feature for 2.1.0. So this actually addresses the UTF-8 issues with this feature. Would it be possible to merge this for the release too? (Thinking out loud)

lotharsm:
If possible, I'd prefer to see this merged in before the 2.1.0 release since the new cloud support feature may look somewhat "unfinished" when using localized ScummVM.

criezy:
For the 2.1.0 release we will either want to include the changes from this PR or remove the translation of the content for the local webserver pages.

My take on it is that the JSON parser is currently incorrect and broken. I am not completely confident with my changes as this was my first foray into the crazy world of unicode, but at worst it will still be wrong and a bit less broken. So I would be in favour of merging before the 2.1.0 release.

SupSuper:
I agree with the "at worst it will still be wrong and a bit less broken" sentiment. :) At least on Windows the non-ASCII files now show correctly on `filesAJAX` (clicking them won't work, but this was already the case).

bluegr:
I talked with @sev about this, and we agreed that it’s fine to merge. Thus, merging this. Thanks for your work!

PR #1849, opened 2019-09-20
MSVC: Don't overwrite version.h if it hasn't changed

lephilousophe:
That avoids superfluous recompilation and linkage of project when nothing
has changed.


sev-:
Very good change, thank you!

PR #1853, opened 2019-09-22
CONFIGURE: Add amigaos to the plugins family

raziel-:
I would have used the linux entry, but -ldl does not exist on amigaos4 and i still need -use-dyn-ld

### DIFF
--- a/configure
+++ b/configure
@@ -3928,6 +3928,18 @@ _mak_plugins=
 if test "$_dynamic_modules" = yes ; then
 	echo_n "Checking whether building plugins is supported... "
 	case $_host_os in
+	amigaos)
+		_plugin_prefix="lib"
+		_plugin_suffix=".so"
+		append_var CXXFLAGS "-fPIC"
+		append_var LIBS "-use-dynld"
+_mak_plugins='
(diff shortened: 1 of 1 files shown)


raziel-:
This is just for my understanding, as i'm completely puzzled as to what the correct usage of linker flags is now.
This
PRE_OBJS_FLAGS  := -Wl,-export-dynamic -Wl,-whole-archive
is sometimes done with one dash after the -Wl, and sometimes with two
PRE_OBJS_FLAGS  := -Wl,--export-dynamic -Wl,--whole-archive

I guess both work?
But why is that?
Is ld smart enough to pick the one it knows or doesn't it matter at all how many dashes there are?

sev-:
Well, as the ld(a) manual states:

> For options whose names are multiple letters, either one dash or two can precede the option name; for example, -trace-symbol and --trace-symbol are equivalent.

So, that is just a matter of personal preference.

raziel-:
@sev-

Awesome, thanks for the answer

sev-:
Merging. I'll be branching today, so it is important to have these things merged before that.

PR #1859, opened 2019-09-23
ILLUSIONS: DUCKMAN: Fix endless SFX bug #11161

yuv422:
The sfx in this scene is from the global sfx resource which doesn't get unloaded on scene change. I've added some logic to stop looping sfx when changing scenes.


sev-:
Merging and backporting

PR #1861, opened 2019-09-24
SDL: Fix build for ancient SDL versions

tsoliman:
This fixes a Maemo build failure that was introduced in commit 6dba0bbfd421121056fba0d348794ead2928c662.

The ancient custom SDL 1.2.8 in the Maemo SDK doesn't have `SDL_iconv_string()`.

I had 3 options (from easy to hard):
* Drop Maemo support
* Surround the new code with a new `#ifdef` and a configure option
* Port the Maemo SDL customizations to the latest SDL 1.2.x

I figured it won't hurt to add yet another configure option.


tsoliman:
Note: The intention is to get this merged and then cherry-picked to branch-2-1.

tsoliman:
I don't like this fix at all. I'll make a better fix that doesn't introduce configure cruft.

PR #1862, opened 2019-09-24
SDL: Fix build for older SDL versions

tsoliman:
This fixes the breakage caused by 6dba0bbfd421121056fba0d348794ead2928c662
in ancient SDL1 versions that don't have `SDL_iconv_string()`.

It was first introduced in SDL 1.2.10.

Maemo for example uses an ancient custom SDL based on 1.2.8.

### DIFF
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -760,6 +760,7 @@ int SDL_SetColorKey_replacement(SDL_Surface *surface, Uint32 flag, Uint32 key) {
 #endif
 
 char *OSystem_SDL::convertEncoding(const char *to, const char *from, const char *string, size_t length) {
+#if SDL_VERSION_ATLEAST(1, 2, 10)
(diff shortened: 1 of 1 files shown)


backends/platform/sdl/sdl.cpp
```diff
@@ -800,5 +801,8 @@ char *OSystem_SDL::convertEncoding(const char *to, const char *from, const char
 	memcpy(finalResult, result, newLength + zeroBytes);
 	SDL_free(result);
 	return finalResult;
+#else
+	return nullptr;
```
SupSuper: This should call the base implementation `ModularBackend::convertEncoding` (it only returns `nullptr` now but could always change in the future)

vyzigold:
I agree with the change proposed by SupSuper. Otherwise it looks alright.

backends/platform/sdl/sdl.cpp
tsoliman: Done

PR #1863, opened 2019-09-25
Supernova lang selection

henke37:
SUPERNOVA: Use the ADGF_ADDENGLISH flag instead of duplicate game entries.
GUI: Remove incorrect advice

Instead of using duplicate game entries in the detection, allow changing the language in the game settings dialog. This is consistent with how certain SCI games support multiple languages for the same fileset.

Also update the tooltip for the language selection dropdown menu to remove the false "advice" that this does not change the game language.

### DIFF
--- a/engines/supernova/detection.cpp
+++ b/engines/supernova/detection.cpp
@@ -62,16 +62,7 @@ static const ADGameDescription gameDescriptions[] = {
 		AD_ENTRY1s("msn_data.000", "f64f16782a86211efa919fbae41e7568", 24163),
 		Common::DE_DEU,
 		Common::kPlatformDOS,
-		ADGF_TESTING,
-		GUIO1(GAMEOPTION_IMPROVED)
-	},
-	{
-		"msn1",
-		nullptr,
-		AD_ENTRY1s("msn_data.000", "f64f16782a86211efa919fbae41e7568", 24163),
-		Common::EN_ANY,
-		Common::kPlatformDOS,
-		ADGF_TESTING,
+		ADGF_TESTING | ADGF_ADDENGLISH | ADGF_DROPLANGUAGE,
 		GUIO1(GAMEOPTION_IMPROVED)
 	},
 	// Mission Supernova 2
@@ -81,16 +72,7 @@ static const ADGameDescription gameDescriptions[] = {
 		AD_ENTRY1s("ms2_data.000", "e595610cba4a6d24a763e428d05cc83f", 24805),
 		Common::DE_DEU,
 		Common::kPlatformDOS,
-		ADGF_TESTING,
-		GUIO1(GAMEOPTION_IMPROVED)
-	},
-	{
-		"msn2",
-		nullptr,
-		AD_ENTRY1s("ms2_data.000", "e595610cba4a6d24a763e428d05cc83f", 24805),
-		Common::EN_ANY,
-		Common::kPlatformDOS,
-		ADGF_TESTING,
+		ADGF_TESTING | ADGF_ADDENGLISH | ADGF_DROPLANGUAGE,
 		GUIO1(GAMEOPTION_IMPROVED)
 	},
 	AD_TABLE_END_MARKER
--- a/gui/editgamedialog.cpp
+++ b/gui/editgamedialog.cpp
@@ -150,8 +150,8 @@ EditGameDialog::EditGameDialog(const String &domain)
 	_descriptionWidget = new EditTextWidget(tab, "GameOptions_Game.Desc", description, _("Full title of the game"));
 
 	// Language popup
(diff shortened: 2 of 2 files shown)


criezy:
There are two separate changes here. I will first comment on the change to the tooltip for the Language selection in the GUI.

What the tooltip meant is that generally speaking if you have a release for example an English release of a game, ScummVM cannot magically transform it into a Spanish version by selecting Spanish into the Language selector.

Where this get a bit complicated, and maybe confusing, as that we have different use cases:
- Some games had multilingual releases, and the Language selector allows selecting the language to use. We still cannot use a language not included in the release though.
- Some games had separate releases for each supported language, but still include some multilingual resources. For example Broken Sword has different releases for each supported language that only include the speech data for that language, but include the subtitles for all the supported language (except Portuguese subtitles that are missing from some releases). The original game did not allow using those other subtitles, but in ScummVM we decided to let the user have the ability to use them, through the Language selector. So for example if you have the French release of Broken Sword and select English in the language selector, you will get French speech with English subtitles.
 - For some games we provide translations for languages not originally supported. That is for example the case of Mission Supernova or Mortville Manor for which ScummVM provides an English translation that can be used on top of the original French or German game files. So if effect we can turn a German version of the game into an English version.

So in summary I would not say that the "advice" was false, and I am not sure about removing it. But maybe it could be reworded to be a bit less confusing in some cases.

criezy:
Regarding the second change, related to the supernova engine, this is a discussion that comes up from time to time. When in ScummVM we add a translation for a game on top of an original release in another language, we have two possible approaches:
- Use different detection entries for each language.
- Use a single detection entry for the original language + English, letting the user select the language in the Language selector.

In supernova we opted for the first approach as it makes it more visible that the game can be played in English, despite having been released only in German originally. We had also taken the same decision for Mortville Manor (with separate detection entries to play in English when using the data files from either the original French or German version).

So again I am not sure about reversing this decision and removing the separate detection entries to play in English.

henke37:
It seems like there is a conflict of interest here. Or rather one interest that is causing blindness for the real issue.

It is cool that the game is available in a new language. Of course it is something that should be promoted.

But at the same time, this is a hack. An ugly hack to abuse the manual version selector into forcing the user to acknowledge that there are more than one supported language. The manual version selector is not meant for the task of promoting the existence of new languages. It's there because the automatic version detection has failed.

For example, it gets in the way of the mass adding function. Having to manually specify which game variant it is during mass adding just doesn't work from a conceptual level. It's fire and forget.

This hack also leads to a denormalized database. Or as I like to put it: a DRY violation. Identical data is stored multiple times when it really should not.

sev-:
Oh, I don't agree, @henke37. Only a few games have multiple language resources, and historically we were displaying all available languages in the GUI. This led to repeated confusion among many users, who believed that they can switch the game language in GUI. And they were submitting bugreports.

First attempt to amend that was adding that tooltip, telling that changing the language in GUI will not affect the game.

Second, a more viable option was implemented in the form of GUI Options. From now on, the language is also stored in GUIO, and that greys out all undetected languages, thus, hinting the user which it is futile to play with that setting.

There are two kinds of multilingual games:

1. Those, which have built-in language selector. There are very few of those, and the way to process them is to specify UNK_LANG as their language. Then GUIO will not go in the way

2. Those, where there are several languages in the data files, but the selection is performed only during game setup, or installation, or via external utility. That also includes our "hacky" translations. For those games we generate several detection options, thus letting the user select the language and then lock it.

Thus, if you would like to drop path 2, then you need to implement in-game language selection. Otherwise, the users will easily complain: "I switched the game to Spanish, but it still shows German to me".

There is a third way to do it, though it will still be confusing, but anyway, more acceptable than just mixing the things or dropping the functionality: Extend GUIOs, or rather our AdvancedDetector, so you could specify a list of supported languages for a specific game.

Thoughts?

henke37:
The way I see it, the `ADGF_ADDENGLISH` flag already shows how to do it. The language selector is right there. And it only shows the languages that the game variant supports. This is how you do language switching for the second scenario.

What needs clarification is what the option does for the particular game. Because it's not consistent. For one game it switches the subtitles, for another the entire game is in a new language. The over engineered, but very formally correct, way to do this is to separate out the various game parts (subtitles, voice overs, gameplay, etc) and show which parts can be changed and to what.

sev-:
Yes, we are talking about the same thing: GUIO is a way for an engine to hint GUI which options should be on or off. However, it has one serious drawback: it requires the engine to run, otherwise, you will not be able to see that, say, the game got a new language supported.

Thus, in this current form, the approach is not quite valid. Apologies, but I'm closing it, for now, more thought/discussions and real-world examples need to be considered in order to figure out how to solve it in a better way.

PR #1865, opened 2019-09-26
ENGINES: Flag games that doesn't use midi

henke37:
Set the GUIO_NOMIDI flag for games that makes no use of MIDI playback.

This gets rid of useless MIDI and MT-32 tabs in the game edit dialog for the games, giving users less options that do nothing.


peterkohaut:
Wouldn't this be cleaner if it is the other way around? Games that use midi will have that negated flag like "GUIO_MIDI"

henke37:
Perhaps, but this is the design that already exists. Given that it's roughly half the games in either setup, I'm not spending the effort to invert the existing design.

bluegr:

lotharsm:
Should we backport this to 2.1.0?

tsoliman:
I don't think so. On one hand, it's a relatively new addition with unintended consequences. On the other hand, we will get immediate feedback/testing about it.
It all depends on how fast we are with 2.1.1

PR #1868, opened 2019-09-29
SCUMM: Detect and use correct language of Steam version of Dig, fixes bug #10697

BenCastricum:
Fixes bug #10697


sev-:
Could you please change the commit message, so it clearly states that it fixes bug #10697. Then our issue tracker will be able to pick up this commit automagically.

bluegr:
Thanks for your work! Merging

PR #1869, opened 2019-09-29
BLADERUNNER: Remove ADGF_TESTING flag

lotharsm:
Due to the high compatiblity level and Blade Runner being some sort of "flagship" game for 2.1.0, I propose to remove the ADGF_TESTING flag in order to remove the warning when starting the game. I propose to leave the flag intact for the "restored content" version.


lotharsm:
Sorry, had to do a force push: *Removing* the flag completely is a bad idea, it now uses "ADGF_NO_FLAGS" as it should be. :/

lotharsm:
Thanks for the feedback! Merging.

PR #1872, opened 2019-10-02
WII: Fix compilation and runtime with latest version of devkitPPC

lephilousophe:
This pull request lets ScummVM compile and run with the latest version of devkitPPC.
There are various fixes :
- paths to portlibs directory are added as it's where 3rd party libraries are expected to be installed
- -fpermissive is needed because reset_cb hasn't the expected signature. The proper way would be to fix the callback signature but I don't know how it would behave with an old toolchain.
- time.h is included in recent versions of network.h and must be excluded from forbidden symbols list
- Recent versions of GCC emits NONE relocations which prevents plugins to be loaded. Adding handling of them to elf-loader code make them work (and as it's a no-op it's easy to add).
- Last commit was an attempt to fix a bug which ended up to be in libgxflux. It destroys everything (and close all files) before quitting. It's not really a needed commit to have everything working.

I am not sure we want to include this in next release as I only tested PR on Dolphin Emulator and not on a real hardware. I would also like a second look from someone who knows well the platform.

### DIFF
--- a/configure
+++ b/configure
@@ -2811,10 +2811,12 @@ case $_host_os in
 		append_var CXXFLAGS "-fmodulo-sched"
 		append_var CXXFLAGS "-fuse-cxa-atexit"
 		append_var CXXFLAGS "-I$DEVKITPRO/libogc/include"
+		append_var CXXFLAGS "-I$DEVKITPRO/portlibs/ppc/include"
 		# libogc is required to link the cc tests (includes _start())
 		append_var LDFLAGS "-mogc"
 		append_var LDFLAGS "-mcpu=750"
 		append_var LDFLAGS "-L$DEVKITPRO/libogc/lib/cube"
+		append_var LDFLAGS "-L$DEVKITPRO/portlibs/ppc/lib"
 		append_var LDFLAGS "-logc"
 		if test "$_dynamic_modules" = "yes" ; then
 			# retarded toolchain patch forces --gc-sections, overwrite it
@@ -3007,10 +3009,12 @@ case $_host_os in
 		append_var CXXFLAGS "-fmodulo-sched"
 		append_var CXXFLAGS "-fuse-cxa-atexit"
 		append_var CXXFLAGS "-I$DEVKITPRO/libogc/include"
+		append_var CXXFLAGS "-I$DEVKITPRO/portlibs/ppc/include"
 		# libogc is required to link the cc tests (includes _start())
 		append_var LDFLAGS "-mrvl"
 		append_var LDFLAGS "-mcpu=750"
 		append_var LDFLAGS "-L$DEVKITPRO/libogc/lib/wii"
+		append_var LDFLAGS "-L$DEVKITPRO/portlibs/ppc/lib"
 		append_var LDFLAGS "-logc"
 		if test "$_dynamic_modules" = "yes" ; then
 			# retarded toolchain patch forces --gc-sections, overwrite it
--- a/backends/fs/wii/wii-fs-factory.cpp
+++ b/backends/fs/wii/wii-fs-factory.cpp
@@ -24,6 +24,7 @@
 
 #define FORBIDDEN_SYMBOL_EXCEPTION_printf
 #define FORBIDDEN_SYMBOL_EXCEPTION_getcwd
+#define FORBIDDEN_SYMBOL_EXCEPTION_time_h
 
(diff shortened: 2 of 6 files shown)


configure
```diff
@@ -2811,10 +2811,12 @@ case $_host_os in
 		append_var CXXFLAGS "-fmodulo-sched"
 		append_var CXXFLAGS "-fuse-cxa-atexit"
 		append_var CXXFLAGS "-I$DEVKITPRO/libogc/include"
+		append_var CXXFLAGS "-isystem $DEVKITPRO/portlibs/ppc/include"
```
dhewg: does dkppc use `-isystem` as well? (compare to just `-I`)

configure
```diff
@@ -2806,6 +2806,7 @@ case $_host_os in
 		append_var CXXFLAGS "-mcpu=750"
 		append_var CXXFLAGS "-meabi"
 		append_var CXXFLAGS "-mhard-float"
+		append_var CXXFLAGS "-fpermissive"
```
dhewg: That's unfortunate, is there a libogc define you can use to differentiate the prototype?

configure
lephilousophe: Good catch!
[devkitPro](https://github.com/devkitPro/pacman-packages/blob/master/pkgbuild-scripts/ppcvars.sh) seems to use -I.
I will correct that.

configure
lephilousophe: Hum, in fact I didn't notice that the change of prototype is quite recent and thought a change of compiler transformed a warning into error.
I see now that libogc has changed prototype between 1.8.17 and 1.8.18, I will try to use a #ifdef to avoid -fpermissive

lephilousophe:
Hi,

Thanks for your report. I wasn't sure if I could bother you on Wii stuff because I noticed you moved to other stuff. :smile: 
Not all dependencies are available through pacman but there are some (libpng, libjpeg, libmad, libogg, libvorbisidec and freetype).
I didn't see libwiiflux in there but ScummVM uses libgxflux so didn't pay that much attention.

dhewg:
Yeah sorry, I meant libgxflux... so that's the only library not available via pacman? Is my old repo at https://repo.or.cz/w/libgxflux.git still used?

Nice to see someone giving this port some love, I don't even recall in which box my good old wii is stuffed in :)
And feel free to take over as wii/gc port maintainer!

bluegr:
Does this PR supersede the older PR #1291?

lephilousophe:
My bad, I didn't see that @ccawley2011 already did similar changes.
This PR has been somewhat tested and includes fixes to ELF loader.
I will try to improve my commits to avoid mandatory toolchain upgrade like in #1291 but without the -fpermissive fix.

configure
lephilousophe: I did the fix with #ifdef.

configure
lephilousophe: Corrected.

lephilousophe:
@dhewg I still use your version of libgxflux with fixes that you can find at carstene1ns/libgxflux@bed224a399cd4179e09d5e7bda3b2a2977dd9886 and carstene1ns/libgxflux@0a636b8fd82052a17f1bc15d2a5e3b5f65bde3e6 plus one more fix from me at [my repo](https://github.com/lephilousophe/dockerized-bb/blob/master/toolchains/devkitppc/packages/libgxflux/patches/003-fix-devoptab.patch)

I can't be the port maintainer as I don't even own a Wii nor a GameCube.

dhewg:
> I still use your version of libgxflux with fixes that you can find at [carstene1ns/libgxflux@bed224a](https://github.com/carstene1ns/libgxflux/commit/bed224a399cd4179e09d5e7bda3b2a2977dd9886) and [carstene1ns/libgxflux@0a636b8](https://github.com/carstene1ns/libgxflux/commit/0a636b8fd82052a17f1bc15d2a5e3b5f65bde3e6) plus one more fix from me at [my repo](https://github.com/lephilousophe/dockerized-bb/blob/master/toolchains/devkitppc/packages/libgxflux/patches/003-fix-devoptab.patch)

Pushed those 3

lephilousophe:
Great! Thanks.
I will update my toolchain scripts.

As everyone seem OK with my last changes, I will merge the PR.
Thanks.

bluegr:
Closing PR #1291 in favor of this one

PR #1873, opened 2019-10-02
GUI: Fix launcher layout for small widths

bgK:
* Testing if a widget can be centered was ignoring the padding.
* Only resize a layout based on its content if it was not explicitly
   sized by its parent. Fixes the logo causing incorrect layout
   computations when the window width is lower than the image width.

Before:
Capture d’écran de 2019-10-02 19-32-52

After:
Capture d’écran de 2019-10-02 19-31-06

There are very obvious remaining issues (I have a fix for the wrapping logo planned for another PR), but at least the layout is correct.

### DIFF
--- a/gui/ThemeLayout.cpp
+++ b/gui/ThemeLayout.cpp
@@ -202,18 +202,11 @@ void ThemeLayoutStacked::reflowLayoutVertical() {
 
 		_children[i]->offsetY(curY);
 
-		// Center child if it this has been requested *and* the space permits it.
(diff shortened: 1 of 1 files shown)


sev-:
Nice catch, though I think, that for such low resolution the more practical approach would be landscape. (Still, the issue would remain, I know).

I'll take a deeper look in couple of days.

bgK:
I will merge this in a few days unless there are further comments.

PR #1876, opened 2019-10-03
CREATE_PROJECT: Split non engine code into additional projects

henke37:
This moves code files out from the main scummvm project and into new projects.

Specifically, it uses these new projects:
* Audio
* Graphics
* Video
* Common
* Backend
* Gui

### DIFF
--- a/devtools/create_project/create_project.cpp
+++ b/devtools/create_project/create_project.cpp
@@ -1504,33 +1504,20 @@ void ProjectProvider::createProject(BuildSetup &setup) {
 	std::string targetFolder;
 
 	if (setup.devTools) {
-		_uuidMap = createToolsUUIDMap();
+		_engineProjects = createToolsUUIDMap();
 		targetFolder = "/devtools/";
 	} else if (!setup.tests) {
-		_uuidMap = createUUIDMap(setup);
+		_engineProjects = createUUIDMap(setup);
 		targetFolder = "/engines/";
 	}
 
 	// We also need to add the UUID of the main project file.
-	const std::string svmUUID = _uuidMap[setup.projectName] = createUUID(setup.projectName);
+	const std::string svmUUID = _engineProjects[setup.projectName] = createUUID(setup.projectName);
 
 
 	StringList in, ex;
 
-	// Create project files
-	for (UUIDMap::const_iterator i = _uuidMap.begin(); i != _uuidMap.end(); ++i) {
-		if (i->first == setup.projectName)
-			continue;
-		// Retain the files between engines if we're creating a single project
-		in.clear(); ex.clear();
-
-		const std::string moduleDir = setup.srcDir + targetFolder + i->first;
-
-		createModuleList(moduleDir, setup.defines, setup.testDirs, in, ex);
-		createProjectFile(i->first, i->second, setup, moduleDir, in, ex);
-	}
-
-#define addUUID(x) _uuidMap[x] = createUUID(createUUID(x)) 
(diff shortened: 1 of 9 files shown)


dreammaster:
I verified that the new Visual Studio solution builds and runs fine. Though little nitpick.. it'd be preferable if they were all grouped under a single solution folder, rather than just being intermixed with the engine projects.

henke37:
Good point. I will look into that.

sluicebox:
What's the motivation?

There's some good code cleanup in here but I don't know why the scummvm project should be broken up.

I like the simplicity of one-engine-per-project + the-app. When I'm in sci mode I want to just see scummvm and sci projects. I frequently scope my VS searches to Current Project on scummvm when I want everything but engines. I specifically want those searches to touch graphics/backend/etc because I don't know ahead of time where they'll hit, that's why I run them in the first place! =) Why break up the mandatory project into equally mandatory projects? It's not like "graphics" is optional.

I found the existing project structure representative of the modular relationships when I was learning, extra hierarchy would have made it less clear. Now some projects map to plugins but some don't?

dreammaster:
Whilst I can't speak for Henke37, I could see two potential benefits from this:
1) Quicker compiles. When I do pulls that changes things in the ScummVM project, Visual Studio has the habit of doing a full rebuild of it. By splitting things up into separate sub-projects, it should be more insulated against this happening. Though.. it could also have been due to my frequent regeneration of the solution. The recent commit to create_project to not regenerate version.h may have already ameliorated this.. I'm not sure.

2) I too frequently use "Current Project" for searches when I'm searching for things within the ScummVM project. But despite this, I normally am only interested in finding things within a specific area. This may prove helpful with that, rather than having to select specific folders to search in, so the searches are limited to just "common/" for example.

Over all, I'm on the fence about this pull request, so long as they (the sub-projects) were grouped in a single ScummVM solution folder. The important thing, in my mind, is if having them as separate projects will help improve recompilation speed. If not, and there's not something I'm missing beyond the slight benefit to doing area specific searches, then it may not be worth splitting them up.

henke37:
One thing for sure, I definitely broke the other code generators. And I'm not equipped to fix them. This is not getting merged without help from others.

devtools/create_project/create_project.cpp
```diff
@@ -1546,20 +1534,38 @@ void ProjectProvider::createProject(BuildSetup &setup) {
 
 		createProjectFile(setup.projectName, svmUUID, setup, setup.srcDir, in, ex);
 	} else if (!setup.devTools) {
-		// Last but not least create the main project file.
 		in.clear(); ex.clear();
-		// File list for the Project file
+		createModuleList(setup.srcDir + "/video", setup.defines, setup.testDirs, in, ex);
+		createProjectFile("Video", addCommonProjectUUID("Video"), setup, setup.srcDir + "/video", in, ex);
```
sev-: I am not using Windows or MSVC, but it appears to me that you are adding here project file starting with a capital letter, while all other projects are lowercase. Is it intentional?

henke37: Just sloppy.

sev-: Great, then there is no issue with fixing those :D

athrxx:
Isn't this still broken, anyway?

> One thing for sure, I definitely broke the other code generators

bluegr:
@henke37 any news on this?

henke37:
Considering the lack of interest from others, I don't think this one has any chance of being merged any time soon.

PR #1877, opened 2019-10-07
AMIGAOS4: Fix NULL pointer access on using plugins

raziel-:
We need to explicitely open dos.library and it's IDOS interface.
Otherwise we will hit an IDOS NULL pointer after compiling a shared binary with (shared) plugins.
The hit will happen on loading a game from any engine, if more than one engine/plugin is available.

### DIFF
--- a/backends/fs/amigaos4/amigaos4-fs.cpp
+++ b/backends/fs/amigaos4/amigaos4-fs.cpp
@@ -69,12 +69,12 @@ AmigaOSFilesystemNode::AmigaOSFilesystemNode() {
 AmigaOSFilesystemNode::AmigaOSFilesystemNode(const Common::String &p) {
 	ENTER();
 
-	// We need to explicitely open dos.library and it's IDOS interface.
-	// Otherwise we will hit an IDOS NULL pointer after compiling a shared
-	// binary with (shared) plugins.
+	// We need to explicitly open dos.library and its IDOS interface.
+	// Otherwise we will hit an IDOS NULL pointer after compiling a
+	// shared binary with (shared) plugins.
 	// The hit will happen on loading a game from any engine, if more
-	// than one engine/plugin is available.
-	DOSBase=IExec->OpenLibrary("dos.library",0);
+	// than one engine/(shared) plugin is available.
+	DOSBase = IExec->OpenLibrary("dos.library", 0);
 	IDOS = (struct DOSIFace *)IExec->GetInterface(DOSBase, "main", 1, NULL);
 	
 	int offset = p.size();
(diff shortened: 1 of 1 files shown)


backends/fs/amigaos4/amigaos4-fs.cpp
```diff
@@ -62,13 +62,21 @@ AmigaOSFilesystemNode::AmigaOSFilesystemNode() {
...
 	_pFileLock = 0;
-	_nProt = 0; // Protection is ignored for the root volume
+	_nProt = 0; // Protection is ignored for the root volume.
 	LEAVE();
 }
 
 AmigaOSFilesystemNode::AmigaOSFilesystemNode(const Common::String &p) {
 	ENTER();
 
+	// We need to explicitely open dos.library and it's IDOS interface.
```
bluegr: explicitly *

backends/fs/amigaos4/amigaos4-fs.cpp
```diff
@@ -62,13 +62,21 @@ AmigaOSFilesystemNode::AmigaOSFilesystemNode() {
...
+	// Otherwise we will hit an IDOS NULL pointer after compiling a shared
+	// binary with (shared) plugins.
+	// The hit will happen on loading a game from any engine, if more
+	// than one engine/plugin is available.
+	DOSBase=IExec->OpenLibrary("dos.library",0);
```
bluegr: Spacing

backends/fs/amigaos4/amigaos4-fs.cpp
```diff
@@ -105,6 +113,10 @@ AmigaOSFilesystemNode::AmigaOSFilesystemNode(const Common::String &p) {
 		IDOS->FreeDosObject(DOS_EXAMINEDATA, pExd);
 	}
 
+	// Close dos.library and it's IDOS interface again.
```
bluegr: its *

backends/fs/amigaos4/amigaos4-fs.cpp
raziel-: Shame...

backends/fs/amigaos4/amigaos4-fs.cpp
raziel-: I´m not really sure what to do about the comments...

sev-: Apparently md5 is referring to the lack of spaces around ‘=‘

raziel-: Ah ok, that should be fixed now.
I was puzzled by the three lines (comments) above (lines 74-76).
Didn't know how i had to fix the spacing there.

bluegr:

PR #1878, opened 2019-10-08
JANITORIAL: Keep copyright "sign" upper-case in "About"

raziel-:
The two new engines Hyperspace Delivery Boy! and Mission Supernova display their copyright "sign" in lowercase, while every other engine is upper-case.


sev-:
Thanks!

raziel-:
Does launcher's "About" dialogue support special chars, so that © could be used in future?

PR #1881, opened 2019-10-09
TTS: Enable in Mingw-w64 builds

sluicebox:
This PR gets Text To Speech working in default Mingw-w64 environments.

- Removes reference to sapiddk.h which isn't used and isn't in Mingw-w64
- Defines guids whose symbols are missing from Mingw-w64
- Restores TTS detection to configure script

With these changes, TTS works out of the box in Mingw-w64 environments. Mingw32 doesn't include speech api headers or libraries and so configure still correctly detects that TTS is unavailable there.

### DIFF
--- a/backends/text-to-speech/windows/windows-text-to-speech.cpp
+++ b/backends/text-to-speech/windows/windows-text-to-speech.cpp
@@ -29,6 +29,17 @@
 #include <basetyps.h>
 #include <windows.h>
 #include <Servprov.h>
+
+// Mingw-w64 is missing symbols for two guids declared in sapi.h which are used
+//  by sphelper-scummvm.h. Mingw32 doesn't include any sapi headers or libraries
+//  so the only way to currently build there is to manually use Microsoft's, in
+//  which case the guids will be defined by their library.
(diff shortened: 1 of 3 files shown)


backends/text-to-speech/windows/windows-text-to-speech.cpp
```diff
@@ -29,6 +29,15 @@
 #include <basetyps.h>
 #include <windows.h>
 #include <Servprov.h>
+
+// Mingw-w64 is missing symbols for two guids declared in sapi.h which are used
+//  by sphelper-scummvm.h. Mingw32 doesn't include any sapi headers or libraries.
+#if defined(__MINGW32__)
```
vyzigold:
```suggestion
#if defined(__MINGW32__) && defined(__MINGW64_VERSION_MAJOR)
```
This should allow the TTS to still be compiled with the old mingw (if somebody goes through the effort of using MS SAPI files with mingw).

bluegr:
The changes look good and @vyzigold approved, so this can be merged. Thanks for your work!

PR #1882, opened 2019-10-10
HDB: Clean up music playback code

henke37:
This moves a bunch of music playback code into the Song class, which makes it easy to DRY out all the repeated code.

Specifically, the code now works with "this" Song and not "song1" or "song2". Additionally, the decoder setup code is factored out to a separate member function, eliminating some quadruple code duplicates.

As a quick bonus, I fixed [bug #11194](https://bugs.scummvm.org/ticket/11194).

### DIFF
--- a/engines/hdb/sound.h
+++ b/engines/hdb/sound.h
@@ -1474,8 +1474,8 @@ struct Song {
 	void update();
 
 private:
-	static Common::String Song::getFileName(SoundType song);
-	Audio::AudioStream* Song::createStream(Common::String fileName);
+	static Common::String getFileName(SoundType song);
+	Audio::AudioStream* createStream(Common::String fileName);
 
 	Audio::SoundHandle handle;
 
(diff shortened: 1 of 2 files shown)


henke37:
I also fixed an use before create bug with the `handle` field. The code was previously trying to set the mixer channel level before the mixer channel had even been allocated.

engines/hdb/sound.h
```diff
@@ -1457,21 +1457,38 @@ struct SoundCache {
 };
 
 struct Song {
-	bool playing;
-	SoundType song;
+
```
sev-: Shouldn't it be class then?

henke37: Doesn't really matter, either keyword declares a class. `struct` simply means I don't need a `public:` before the member functions.

digitall:
This looks good to me. @sev- : Happy to merge?

sev-:
Absolutely! Thanks!

PR #1883, opened 2019-10-10
MOHAWK: RIVEN: Modify cursor preloading to resolve crashing on 3DS

BallM4788:
On 3DS builds of ScummVM, attempting to start Riven results in an instant crash due to the way cursor preloading is coded. This commit resolves that issue.

### DIFF
--- a/engines/mohawk/cursors.cpp
+++ b/engines/mohawk/cursors.cpp
@@ -242,18 +242,20 @@ void LivingBooksCursorManager_v2::setCursor(const Common::String &name) {
 }
 
 PECursorManager::PECursorManager(const Common::String &appName) {
-	Common::PEResources exe;
-	if (!exe.loadFromEXE(appName)) {
+	_exe = new Common::PEResources();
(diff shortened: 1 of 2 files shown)


BallM4788:
Regarding the specific cause of the crash: I do not know for certain, but my uneducated guess is that it has something to do with memory addressing, as the stack dump when it crashes is completely empty. I determined the general cause by building various commits and testing them on my 3DS, then narrowing down to the exact commit where crashing began ([b9a72ff](https://github.com/scummvm/scummvm/commit/b9a72ff7a6dc7033beae4a75dc5f596264c39419)); it only has one parent ([672cfbd](https://github.com/scummvm/scummvm/commit/672cfbd518626fffc3476c2f6d0cdd391ab12e17)), under which Riven was still functional. Anyway, please let me know if anything in my above commit is incorrect.

bgK:
I was able to reproduce the issue. Allocating `Common::PEResources` on the stack was causing a stack overflow. Merged, thanks!

PR #1884, opened 2019-10-11
AMIGAOS4: Further update amigaos.mk and fix shared plugins workaround

raziel-:
I have a request to make, since i can't do it on my own.

Would someone please be so nice and move
scummvm_drawer.info
scummvm.info
from icons/
to
dists/amiga

Otherwise the amigaos.mk changes will break due to the wrong path.

Thank you very much

Reason is, that everyone with it's own platform (except windows if i see that right?) dists has their icons in their own subdir and i think it looks much more cleaner that way too. (Also amiga has it´s dists subdir aswell)


sev-:
Merging and will move the file, thanks!

PR #1885, opened 2019-10-11
GUI: Easter egg

sev-:
For a long while, I wanted to implement an easter egg to our About dialog. Here it goes.

Looking for your comments, on how suitable that is, or I am completely crazy here.

Request for Comments!

PS. Launch 'about', press 'v'. After browsing through the menu, use arrow keys for the character on the right. or enjoy AI (original one) entertaining you. ESC to quit.

### DIFF
--- a/gui/about.cpp
+++ b/gui/about.cpp
@@ -690,16 +690,18 @@ void EE::resetpt() {
 
 void EE::calcscore() {
 	if (_winner == _server) {
-		if (_soundEnabled && _mode != kModeMenu)
+		if (_soundEnabled && _mode != kModeMenu) {
 			playSound(kSndPoint);
-			_score[_winner]++;
+		}
+		_score[_winner]++;
 		if (_score[_winner] > 14 && _score[_winner] - _score[1 - _winner] > 1) {
 			init();
 			return;
 		}
 	} else {
-		if (_soundEnabled && _mode != kModeMenu)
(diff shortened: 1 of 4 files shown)


lotharsm:
Nice touch, I really like little gems like this!

Gameplay wise, I'd prefer a slightly faster character movement for the player. ;)

digitall:
Not sure about suitable, but there are a few warning fixes and general code style fixes that should be applied. I have committed them to the ee_fixup branch on my tree as:
https://github.com/digitall/scummvm/commit/67870b7b42d8feb8f60aeed2c1011f8c891a72cb

athrxx:
I think this is a nice extra for the about dialog.
Not sure whether the move speed should be increased. Maybe very slightly? Works fine for me... :)

criezy:
I think it's fun.
I don't really have any opinion on increasing or not the character movement. I suck at those kind of games anyway and it would probably need to be about 10x slower for me to manage to score a point :-P

sev-:
@digitall thanks, I applied your patch

sluicebox:
I love this! It's fun, looks great, and About boxes are my favorite place to put easter eggs. The speed is good but if it's slightly increased or decreased I expect it will still be good. I mean... it's not like I'm going to score any *less* points...

sev-:
Thanks for your feedback, merging

PR #1887, opened 2019-10-13
TOON: Add subtitles support for cutscenes

BLooperZ:
This allows subtitles during cutscenes in Toonstruck.
Attaching subtitles file for first part of intro cutscene - should be extracted to `MISC` directory
[209_1M.zip](https://github.com/scummvm/scummvm/files/3721733/209_1M.zip)

Subtitles rendered using the same text draw method from the game so they look consistent.
text color changes with the palette of the cutscene.

The original game does not have subtitles in cutscenes (new feature),
I also could not find a transcript of the cutscenes' texts.
so the hard work of transcribing and timing the subtitles is still ahead.

Subtitles files are stored  in custom simple format (`tss` for Toonstruck Subtitles) consists of:
index - start frame, end frame, text offset
texts - null terminated strings

subtitles files should have the same name as matching cutscene files (minus the extension).
for example: 209_1M.SMK (cutscene) -> 209_1M.tss (subtitles).
(When all subtitles are done, I think of putting them all under a single `SUBTITLES.PAK` file)

(I have a simple python script to convert `sbv` to `tss`... I would like to add it to `devtools`.
need explanation about those `module.mk` files, though)

### DIFF
--- a/engines/toon/movie.cpp
+++ b/engines/toon/movie.cpp
@@ -124,7 +124,7 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 					int len = frame->w * frame->h;
 					byte pixels[310000] = {0};
 					memcpy(pixels, frame->getPixels(), len);
-					for (byte i = 1; i < 256; i++) 
+					for (int i = 1; i < 256; i++) 
 					{
 						int j;
 						for (j = 0; j < len; j++) {
--- a/engines/toon/movie.cpp
+++ b/engines/toon/movie.cpp
@@ -119,22 +119,21 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 				} else {
 					_vm->_system->copyRectToScreen(frame->getPixels(), frame->pitch, 0, 0, frame->w, frame->h);
 
-    				int32 currentFrame = _decoder->getCurFrame();
+					int32 currentFrame = _decoder->getCurFrame();
 
+					// find unused color key to replace with subtitles color
 					int len = frame->w * frame->h;
-					byte pixels[310000] = {0};
-					memcpy(pixels, frame->getPixels(), len);
-					for (int i = 1; i < 256; i++) 
-					{
-						int j;
-						for (j = 0; j < len; j++) {
-							if (pixels[j] == i) {
-								break; 
-							}
-						}
-				
-						if (j == len && i != 255) {
-							unused = i;
+					const byte* pixels = (const byte *)frame->getPixels();
+					byte counts[256];
+					memset(counts, 0, sizeof(counts));
+					for (int i = 0; i < len; i++) {
+						counts[pixels[i]] = 1;
+					}
+
+					// 0 is already used for the border color and should not be used here, so it can be skipped over.
+					for (int j = 1; j < 256; j++) {
+						if (counts[j] == 0) {
+							unused = j;
 							break;
 						}
 					}
@@ -155,16 +154,11 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 				}
 			}
 
-			byte palette[768] = {0};
-			memcpy(palette, _decoder->getPalette(), 768);
-
+			byte subtitleColor[3] = {0xff, 0xff, 0x0};
+			_vm->_system->getPaletteManager()->setPalette(_decoder->getPalette(), 0, 256);
 			if (unused) {
-				palette[3 * unused] = 0xff;
-				palette[3 * unused + 1] = 0xff;
-				palette[3 * unused + 2] = 0x0;
+				_vm->_system->getPaletteManager()->setPalette(subtitleColor, unused, 1);
 			}
-
-			_vm->_system->getPaletteManager()->setPalette(palette, 0, 256);
 			_vm->_system->updateScreen();
 		}
 
--- a/engines/toon/movie.cpp
+++ b/engines/toon/movie.cpp
@@ -122,17 +122,17 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 					int32 currentFrame = _decoder->getCurFrame();
 
 					// find unused color key to replace with subtitles color
-					int len = frame->w * frame->h;
-					const byte* pixels = (const byte *)frame->getPixels();
-					byte counts[256];
-					memset(counts, 0, sizeof(counts));
-					for (int i = 0; i < len; i++) {
-						counts[pixels[i]] = 1;
+					uint len = frame->w * frame->h;
+					const byte *pixels = (const byte *)frame->getPixels();
+					bool counts[256];
+					memset(counts, false, sizeof(counts));
+					for (uint i = 0; i < len; i++) {
+						counts[pixels[i]] = true;
 					}
 
 					// 0 is already used for the border color and should not be used here, so it can be skipped over.
 					for (int j = 1; j < 256; j++) {
-						if (counts[j] == 0) {
+						if (!counts[j]) {
 							unused = j;
 							break;
 						}
--- a/engines/toon/movie.cpp
+++ b/engines/toon/movie.cpp
@@ -31,6 +31,7 @@
 #include "toon/audio.h"
 #include "toon/movie.h"
 #include "toon/toon.h"
+#include "toon/subtitles.h"
 
 namespace Toon {
 
@@ -67,6 +68,7 @@ Movie::Movie(ToonEngine *vm , ToonstruckSmackerDecoder *decoder) {
 	_vm = vm;
 	_playing = false;
 	_decoder = decoder;
+	_subtitle = new SubtitleRenderer(_vm);
 }
 
 Movie::~Movie() {
@@ -87,6 +89,7 @@ void Movie::play(const Common::String &video, int32 flags) {
 		_vm->getAudioManager()->setMusicVolume(0);
 	if (!_decoder->loadFile(video.c_str()))
 		error("Unable to play video %s", video.c_str());
+	_subtitle->load(video.c_str());
 	playVideo(isFirstIntroVideo);
 	_vm->flushPalette(true);
 	if (flags & 1)
@@ -103,6 +106,7 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 	while (!_vm->shouldQuit() && !_decoder->endOfVideo()) {
 		if (_decoder->needsUpdate()) {
 			const Graphics::Surface *frame = _decoder->decodeNextFrame();
+			byte unused = 0;
 			if (frame) {
 				if (_decoder->isLowRes()) {
 					// handle manually 2x scaling here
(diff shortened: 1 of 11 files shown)


engines/toon/toon.cpp
```diff
@@ -3286,7 +3286,18 @@ void ToonEngine::drawConversationLine() {
 	if (_currentTextLine && _showConversationText) {
 		_fontRenderer->setFontColorByCharacter(_currentTextLineCharacterId);
 		_fontRenderer->setFont(_currentFont);
-		_fontRenderer->renderMultiLineText(_currentTextLineX, _currentTextLineY, _currentTextLine, 0);
+		_fontRenderer->renderMultiLineText(_currentTextLineX, _currentTextLineY, _currentTextLine, 0, *_mainSurface);
+	}
+}
+
+void ToonEngine::drawCustomText(int16 x, int16 y, char* line, Graphics::Surface* frame, char color) {
```
sev-: code formatting

engines/toon/toon.h
```diff
@@ -212,6 +213,8 @@ class ToonEngine : public Engine {
 	void waitForScriptStep();
 	void doMagnifierEffect();
 
+	void drawCustomText(int16 x, int16 y, char* line, Graphics::Surface* frame, char color);
```
sev-: code formatting

BLooperZ:
Thank you for your feedback.
I have now fixed the problems.
(the obscure part is rewritten...
it's purpose is finding available byte not used in video frame to change its color for use as subtitles text color)

bluegr:
Nice work!

Some fixes on the subtitles:
    tommorow -> tomorrow
    I'm so DEAD! -> I'm SO dead!
    REVITILIZING -> REVITALIZING
    Here so I see it -> Here's how I see it
    A whole truckload of them is got to be -> A whole truckload of the things has got to be

And some more text afterwards (up to 5:10, around the middle of the intro):
   -  I want the new characters on my desk first thing tomorrow morning... and this time, DON'T be late!
    - Tomorrow morning, first thing, yes.
    - OK.... rabbits! More sweet, friendly, rabbits! Rabbits think nice things, rabbits say "I care". Ship!
    - You only worked with a very young, very naive Drew Blank... a kid that was going to take the world by storm... Flux. Hey there, lil' buddy. The Flux Wildly show was really going to be something... if only
    - Must... focus... on... bunnies.... Must... focus... on... bunnies! *yawn*
    - Hey cuties! Let's get hap... hap... happy! It's fluffy fluffy bun bun time!
    - Oh Drew! What are you doing at 4 o'clock in the morning anyway?
    - I've been caught with bunnies! This must be a dream! I must have fallen asleep! Oh God, sleep! I got to... Wake up Drew! Wake up! Damn it!

BLooperZ:
Thank you.
added support for loading `SUBTITLES.PAK` and updated subtitles file (extract into `MISC`)
[SUBTITLES.zip](https://github.com/scummvm/scummvm/files/3730968/SUBTITLES.zip)

engines/toon/movie.cpp
```diff
@@ -115,9 +119,30 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 				} else {
 					_vm->_system->copyRectToScreen(frame->getPixels(), frame->pitch, 0, 0, frame->w, frame->h);
 
+					int32 currentFrame = _decoder->getCurFrame();
+
+					// find unused color key to replace with subtitles color
+					int len = frame->w * frame->h;
```
sev-: This should be uint

engines/toon/movie.cpp
```diff
@@ -115,9 +119,30 @@ void Movie::playVideo(bool isFirstIntroVideo) {
...
+					const byte* pixels = (const byte *)frame->getPixels();
```
sev-: `const byte *pixels`

engines/toon/movie.cpp
```diff
@@ -115,9 +119,30 @@ void Movie::playVideo(bool isFirstIntroVideo) {
...
+					byte counts[256];
+					memset(counts, 0, sizeof(counts));
+					for (int i = 0; i < len; i++) {
```
sev-: This is also uint

engines/toon/movie.cpp
```diff
@@ -103,6 +106,7 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 	while (!_vm->shouldQuit() && !_decoder->endOfVideo()) {
 		if (_decoder->needsUpdate()) {
 			const Graphics::Surface *frame = _decoder->decodeNextFrame();
+			byte unused = 0;
```
sev-: maybe use of `bool` would be better?

engines/toon/movie.cpp
```diff
@@ -115,9 +119,30 @@ void Movie::playVideo(bool isFirstIntroVideo) {
 				} else {
 					_vm->_system->copyRectToScreen(frame->getPixels(), frame->pitch, 0, 0, frame->w, frame->h);
 
+					int32 currentFrame = _decoder->getCurFrame();
+
+					// find unused color key to replace with subtitles color
+					int len = frame->w * frame->h;
+					const byte* pixels = (const byte *)frame->getPixels();
+					byte counts[256];
```
sev-: And `bool` here too perhaps? Because you're not counting.

engines/toon/movie.cpp
```diff
@@ -115,9 +119,30 @@ void Movie::playVideo(bool isFirstIntroVideo) {
...
+							break;
+						}
+					}
+
+					_subtitle->render(*frame, currentFrame, unused);
+
 					// WORKAROUND: There is an encoding glitch in the first intro video. This hides this using the adjacent pixels.
 					if (isFirstIntroVideo) {
-						int32 currentFrame = _decoder->getCurFrame();
+						// int32 currentFrame = _decoder->getCurFrame();
```
sev-: Why this is commented out now? If you're referring to the new call above, then don't you change the semantics, as this call was only when `isFirstIntroVideo` is set?

engines/toon/subtitles.cpp
```diff
@@ -0,0 +1,94 @@
...
+	// sprintf(strf, "Time passed: %d", frameNumber);
+	// _vm->drawCostumeLine(0, 0, strf, _subSurface);
+	// _vm->_system->copyRectToScreen(_subSurface->getBasePtr(0, 0), _subSurface->pitch, 0, 0, _subSurface->w,  _subSurface->h);
+
+	if (frameNumber > _tw[_index].fend) {
+		_index++;
+		if (_index > _last) {
+			return;
+		}
+		_currentLine = (char*)_fileData + _tw[_index].foffset;
```
sev-: `char *`

engines/toon/toon.cpp
```diff
@@ -3286,7 +3289,18 @@ void ToonEngine::drawConversationLine() {
 	if (_currentTextLine && _showConversationText) {
 		_fontRenderer->setFontColorByCharacter(_currentTextLineCharacterId);
 		_fontRenderer->setFont(_currentFont);
-		_fontRenderer->renderMultiLineText(_currentTextLineX, _currentTextLineY, _currentTextLine, 0);
+		_fontRenderer->renderMultiLineText(_currentTextLineX, _currentTextLineY, _currentTextLine, 0, *_mainSurface);
+	}
+}
+
+void ToonEngine::drawCustomText(int16 x, int16 y, char *line, Graphics::Surface *frame, char color) {
```
sev-: `byte color` is better, as you may get color index >0x7f, and `char` is a signed type on number of platforms.

`line` could also perhaps be better as `byte`, because you potentially could get characters with diacritics.

engines/toon/toon.h
```diff
@@ -33,6 +33,7 @@
 #include "toon/state.h"
 #include "toon/picture.h"
 #include "toon/anim.h"
+#include "toon/subtitles.h"
```
sev-: Why do you need to include it here? It adds unneeded dependency.

engines/toon/movie.cpp
BLooperZ: this variable should later get assigned with unused value in range [1, 256] afterwards (color key) - 0 means it failed to find unused value (and already can't be unused, as it used for text border)

engines/toon/movie.cpp
BLooperZ: (sorry, I don't think I understand the question) this line was moved (line 122) so it's now unconditional as frame information should now be used for all videos and not just this workaround here.

engines/toon/toon.cpp
BLooperZ: `line` matches with the type of `_currentTextLine` (in `toon.h` 391) for consistency

BLooperZ:
Thanks again. notes marked as resolved are now fixed.

sev-:
After looking at it once more, I have one suggestion. Please, tell me what you think.

Current format for subtitles is basically binary. I was solving basically the same problem for Broken Sword 1 engine. You may get the subtitles pack here: https://www.scummvm.org/frs/extras/Broken%20Sword%20I%20and%20II/Sword1_Cutscenes_Subtitles-1.1.zip  and the code is here: https://github.com/scummvm/scummvm/blob/master/engines/sword1/animation.cpp#L131

The format is plaintext:
<start frame> <end frame> @<color> <subtitle text>

That could let anyone to create their custom subtitles for their language.

Do you think it worth changing the format, or still too much work?

sev-:
BTW, this needs to be rebased, so the merge commit goes away. Also, what questions do you have regarding the ``module.mk`` file?

BLooperZ:
I think plaintext format is better,  I wil try to work it into this as well, thanks.

Regarding `module.mk`, I can't find any information about it
I can see it is used to create makefile, so it allows creating  `make devtools/...` commands.
but I don't underatand how to use it and where to put it.
Is it well-known tool? if so, where can I find documentation?
BTW, If the subtitles are plaintext there is no need to add build tool (no need to add `module.mk`).

Thanks.

sev-:
`module.mk` is used by `makefile` utility on *nix-like systems, together with `configure`. I can add the tool for you, it is pretty straightforward. Once some tool is added, it could be built by calling `make <target-name>`, where `target-name` is normally your tool name, like `make toon-subtitles`

sev-:
And in general, I can take care of the `module.mk`, especially if you explain how your tool should be invoked. Also, I think that since it is written in python, there is nothing to build, and the only thing you need is to provide documentation next to it or at the beginning of the file.

bluegr:
I see you changed the subtitles format as per @sev- 's suggestion. Cool! :)

Could you provide a sample subtitle file with the new text format?

BLooperZ:
[SUBTITLES.zip](https://github.com/scummvm/scummvm/files/3827715/SUBTITLES.zip)
In this file the subtitles available both as separate plaintext files (`.tss`) or as single `.pak` file.
Each is enough to activate the subtitles.
When both available in game's director, the separate files take precedence.

sev-:
Nice, I'm merging this.

PR #1889, opened 2019-10-13
3DS: Embed ScummVM's support files in the package

bgK:
So users don't have to copy them to the SD card.

Also introduce a `FilesystemFactory` implementation for filesystems using the POSIX API but with a hard-coded list of drives at the top-level.
(@rsn8887 maybe you'll want to use that for the Vita)


bgK:
Merged manually.

PR #1892, opened 2019-10-14
ANDROID: Fix scaling on devices with high resolution

peterkohaut:


peterkohaut:
done in https://github.com/scummvm/scummvm/pull/1895

PR #1897, opened 2019-10-18
IOS7: enable home indicator auto-hide on iPhone X and up

stewi2:
This makes the annoying white line at the bottom of the screen go away on newer (home button-less) iPhones.

home_indicator

See also https://developer.apple.com/documentation/uikit/uiviewcontroller/2887510-prefershomeindicatorautohidden

I tested this on iOS10 to make sure it's backwards compatible.


bluegr:
Nice! Merging

PR #1898, opened 2019-10-19
BLADERUNNER: Split vector.h into h and cpp

dafioram:
All vector operations were in vector.h now its declared in header and
defined in cpp.

Removed the inlines.

Adding some needed includes.

### DIFF
--- a/engines/bladerunner/matrix.cpp
+++ b/engines/bladerunner/matrix.cpp
@@ -20,9 +20,7 @@
  *
  */
 
-#include "bladerunner/matrix.h"
-
-#include "common/scummsys.h"
+#include "bladerunner/matrix.h" // Already includes "common/scummsys.h"
 
 namespace BladeRunner {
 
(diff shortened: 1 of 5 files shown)


bluegr:
The vector related functions are called a lot, which is why they are defined as inline functions. With your changes, you've effectively made then non-inline, which could affect performance.

Have you profiled the game's performance with these changes?

dafioram:
My system is too fast to notice a difference from this change. For slower systems turning on optimization level 2 should inline the functions again.

There are probably more efficient things we can do as well to increase code performance, such as pass more by reference, and reduce the number of constructor calls.

lephilousophe:
@dafioram -O2 can't inline back functions when they are not in the same cpp file as they are called.
You need LTO for this which isn't enabled by default with gcc (no -flto flag in Makefile)

peterkohaut:
I prefer NOT to merge this, inlining is intentional. There are plenty of platforms where we have performance issues so adding unnecessary overhead is not good idea.

bluegr:
As per the above comments, I'm closing this

PR #1899, opened 2019-10-20
COMMON: Add ignoreClashes flag to SearchSet and FSDirectory

lephilousophe:
This lets engine mute warnings about name clashes when there are useless
for its case. This will be used by Versailles as the tree has
directories with the same name at various places.
Files are duplicated too in different directories but are identical.

I removed the mutable flag on _depth and _flat as they shouldn't have it. Their change on a const object would change behaviour of the object.


lephilousophe:
After talking with @sev-, we agreed that using a flag instead of adding a new argument was better.

sev-:
Looks good, thanks!

PR #1901, opened 2019-10-21
AMIGAOS4: Automate special target installation even more

raziel-:
Thank you all for the help


lotharsm:
Thanks for your contribution! Could you please change the commit message of 
e33fa38 and add the "AMIGAOS4:" prefix?

bluegr:
Thanks! I'll squash these commits in one, so that the commit message is correct

PR #1902, opened 2019-10-22
SCUMM: Bump HE version of Pajama2 GB to 99, Fixes bug #10933

BenCastricum:
The bug report error mentions that the MAXS blocksize is 52, so this game is HE99 game. Setting it in the scumm-md5.(h|txt) to 99 fixes bug #10933.


bluegr:
Nice work! Merging

PR #1904, opened 2019-10-23
SLUDGE: All games are unstable, since the engine is unstable.

henke37:
Unstable engine, unstable games.

### DIFF
--- a/engines/sludge/detection_tables.h
+++ b/engines/sludge/detection_tables.h
@@ -30,7 +30,7 @@ static const SludgeGameDescription gameDescriptions[] = {
 			AD_ENTRY1s("Welcome.slg", "50445503761cf6684fe3270d0860a4c3", 51736),
 			Common::EN_ANY,
 			Common::kPlatformUnknown,
-			ADGF_NO_FLAGS,
+			ADGF_UNSTABLE,
 			GUIO1(GUIO_NOMIDI)
 		},
 		0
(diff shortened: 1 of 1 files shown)


bluegr:
Good point. Thanks! Merging

sev-:
Why, oh why? The engine is disabled. The games are not supported. Shall we go and mark as such ALL games in engines in the development?

I am sorry but I'm reverting this. If you want to follow this, there are other better ways, bring it to the -devel.

tsoliman:
>Why, oh why? The engine is disabled. The games are not supported.

So that the warning happens on unstable builds. (e.g. from the buildbot or from any source that built it with `--enable all engines`).

>Shall we go and mark as such ALL games in engines in the development?

Yes. In fact, we did exactly that when the feature was first introduced (see #30)

>I am sorry but I'm reverting this. If you want to follow this, there are other better ways, bring it to the -devel.

No. This is how the feature works and how it was intended. If you don't like the warning you can set `enable_unsupported_game_warning=false`.

sev-:
Okay, looks like I completely forgot it. Apologies for that.

Then please revert my revert, and if time permits, do the same for all the engines in development. I'll try to do it from my side as well. And please do it directly, without PRs.

sev-:
However, I think that the better approach would be to enhance our build system and let the information about the unstable engine passed by configure. So then all games get this feature automatically. I will look also what could be done. If that will be implemented, then this flag could be restricted only to the non-working games within the supported engines.

PR #1909, opened 2019-10-29
3DS: NEW FEATURE: Magnify Mode

BallM4788:
For games with resolution higher than the hardware resolution, allow toggle-able magnification to and from scale factor 1 on the top screen; use touchscreen to choose what part of the game display to magnify.


backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -150,6 +154,9 @@ void OSystem_3DS::initSize(uint width, uint height,
 	_gameHeight = height;
 	_gameTopTexture.create(width, height, _pfGameTexture);
 	_overlay.create(getOverlayWidth(), getOverlayHeight(), _pfGameTexture);
+	_gameMagTexture.create(_magWidth, _magHeight, _pfGameTexture);
```
bgK: It would be better if the magnify mode did not use an extra texture / copy of the game screen. The 3DS is a low power device, the more processing power remains available to the game engine, the better.

`_gameTopTexture` already contains a full resolution version of the game screen. What needs to happen is showing the magnified part of the texture. This can be done by setting the UV coordinates of the vertices when rendering the sprite:
https://github.com/scummvm/scummvm/blob/af4e15ce3f6b4a22b9f7a4246e9e4e918042a085/backends/platform/3ds/sprite.cpp#L79

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -36,6 +36,8 @@
 
 namespace _3DS {
 
+extern MagnifyMode magnifyMode;
```
bgK: It's better to store this in the `OSystem_3DS` rather than having an extern declaration.

backends/platform/3ds/osystem-graphics.cpp
BallM4788: Because _gameBottomTexture pulls pixel data from _gameTopTexture, any changes to the latter's UV coords would affect that of the former as well. Because only the top screen's position should change when magnified, I opted to instead manipulate _gameTopTexture's scale and position.

BallM4788:
Well after a bit of bungling up I hope I fixed that conflict with the main branch. To be honest, I'm kind of new to merging and rebasing.

Mataniko:
Thanks for the update, there is still a little more cleanup left to do, namely the last 4 commits:

- Revert "3DS: Address conflict with 036d61c" ce2386b
- Revert "Revert "3DS: Address conflict with 036d61c"" 1512eab
- Merge branch '3ds_magnifymode' of https://github.com/ballm4788/scummvm … fedc311
- Merge branch 'master' of https://github.com/ballm4788/scummvm into 3d… d045ad0 

You can do this by doing an interactive rebase:
`git rebase HEAD~5 -i` 
In interactive rebase mode, you can either reword, drop or squash (combine with the previous commit) to clean up the commits

BallM4788:
Alright, I believe that should be everything.

bgK:
Thanks, merged.

If you want to keep working on this feature, here are some areas of improvement:
* Automatically enable both screens while magnify mode is enabled.
* Preventing magnified area updates while the GUI is open.

PR #1911, opened 2019-11-01
3DS: Remove 3dstool dependency

MrHuu:
This patch alters the way the .cia is build. It removes the need for 3dstool.

Instead of first generating a .romfs file using 3dstool, let makerom generate the romfs while building the .cia.


bgK:
Great, thanks! Didn't know this was possible.

PR #1912, opened 2019-11-01
GUI: Update ScummVM Modern Remastered theme

lotharsm:
A while ago, @Tkachov created a set of [custom ScummVM themes.](http://tkachov.ru/scummvm/themes/midnight/)

Since I really like the more "flat" style, I adapted the GFX .stx file for our current ScummVM Modern Remastered theme - and I think it really works well.

Theme-wise, full credit goes to @Tkachov - all I did was to re-color the checkbox and radio buttons so they match our color scheme and changed the formatting of the .stx file so they match our current standard. Otherwise, the theme itself, especially the .stx file, is untouched.

Here are a few screenshots:

scummvm00000
scummvm00001
scummvm00002
scummvm00003
scummvm00004
scummvm00005

So - what do you think? :)


Mataniko:
Let me preface that I think that the remastered theme still has a lot of to improve. Things like the scrollbars and buttons are the obvious big places to start.

However, these don't really work on a few levels:

- Flatness - This is too flat. Even with flat design, modals and buttons have some depth to them, otherwise they blend too much and the interaction becomes awkward. Even worse is now you have elements with some depth, and other elements with no depth, making the theme feel inconsistent
- Colors - The colors, especially Red on Orange need some work, but the white modals don't work at all, especially with red accents.

What does work:
- Scrollbars - I think they're an improvement, but would require some more refinement.

lotharsm:
Hi there, thanks for your replies - it's a good start for a discussion :)

>     * Flatness - This is too flat. Even with flat design, modals and buttons have some depth to them, otherwise they blend too much and the interaction becomes awkward. Even worse is now you have elements with some depth, and other elements with no depth, making the theme feel inconsistent

Well - I have to admit that I am a huge fan of modern, flat designs - I even tried to play around with completely square buttons ;) 

>     * Colors - The colors, especially Red on Orange need some work, but the white modals don't work at all, especially with red accents.

Yeah... it's not easy at all to work with the orange color scheme...

> 
> 
> What does work:
> 
>     * Scrollbars - I think they're an improvement, but would require some more refinement.

What do you think should be improved regarding the scrollbars?

sev-:
The scrollbars were never part of the design. Currently, they are ugliness born out of my personal lack of graphic design skills. We need to implement something which will truly fit the GUI theme.

I think this could be safely closed. @lotharsm you are welcome to revive the https://wiki.scummvm.org/index.php?title=GUI_Themes page and put your creation there.

lotharsm:
@sev- Thanks for the feedback! I'll move the theme to the wiki page as you proposed. Closing.

PR #1913, opened 2019-11-02
NETWORKING: Try loading the CA bundle from DATA_PATH

bgK:
So that platforms where libcurl cannot access the native CA store can ship a CA bundle along with the other ScummVM support files.

Also improve libcurl error handling so warnings are issued for failed requests.


bluegr:
Nice work! :)

Where is DATA_PATH declared? It’s not in the configure script

bgK:
> Where is DATA_PATH declared? It’s not in the configure script

https://github.com/scummvm/scummvm/blob/0c1a04a23cabcb5d8976f12b85576b663ec1524e/configure#L5573

bluegr:
Indeed, for some reason I mistyped it when searching through the file.

Nice work! This is quite a straightforward change, and it can be safely merged.

Merging

PR #1914, opened 2019-11-03
NETWORKING: Enter Session and SessionRequest

Tkachov:
Originally intended to be a speedup for libcurl by reusing the same handle in keep-alive connections. Turns out, libcurl automatically does that.

Anyways, this PR does some useful edits in backends/networking/curl and adds SessionRequest, which might be not only used asynchronously as before (via callbacks), but also synchronously using wait(). Session class allows to reuse these SessionRequests.


sev-:
Thanks! Merging

PR #1916, opened 2019-11-05
CLOUD: Remove the saves sync finished OSD message

bgK:
When using the cloud save sync features, It seems to me there are too many user interface elements, I feel a bit overwhelmed. There is a blinking overlay icon indicating network activity, OSD messages indicating success or failure, and a progress dialog when the save / load dialog is open.

With this PR, I propose removing the OSD message indicating the save sync finished successfully. For me, it breaks the experience when it appears while in-game after the engine autosaved. IMO, the overlay icon disappearing is a good enough indication that the saves synchronized.

### DIFF
--- a/backends/cloud/storage.cpp
+++ b/backends/cloud/storage.cpp
@@ -229,7 +229,6 @@ void Storage::savesSyncDefaultCallback(BoolResponse response) {
 
 	if (!response.value)
 		warning("SavesSyncRequest called success callback with `false` argument");
-	Common::OSDMessageQueue::instance().addMessage(_("Saved games sync complete."));
 }
 
 void Storage::savesSyncDefaultErrorCallback(Networking::ErrorResponse error) {


Tkachov:
I didn't really test the feature by completing the game yet (well, I wanted to do that on Android and there was a problem with certificates), but I thought that OSD message might be a little bit distracting.

I think instead of removing we should add some options for user to specify whether they want to see it or not. Like:
OSD: always show / show on failure / never show
icon: always show / never show

Maybe also add some icon for "failure" and show it for some time on failure. Or even turn that icon in actual interactable element. If you click a "working" icon, it shows some popup with current progress (for the "what takes you so long and when you'll be done?" use-case) and if it's "failed" icon, it shows the reason and maybe some hints on how that could be resolved, maybe "retry" button as well.

bgK:
An option for disabling the icon would be nice as the current design perhaps lacks a bit in subtlety.

However, I fail to see a case where the message discussed here would be helpful. If it worked, then it worked. No need to tell me.

lotharsm:
I agree that there should be at least an option to disable the "success" OSD message since we get a notification in case of an error anyway. Coincidentally, I got asked if there's a way to disable the message since it seems to be quite disruptive...

criezy:
I don't really see a case where having the save sync message would really be useful, so I would tend to agree with @bgK that it can be removed entirely rather than adding an option. We don't really want options for every feature that can have one, as we would end up with too many options, as well as unnecessary code that adds to the maintenance burden. And to me it seems that this is a case where we don't really need an option.

Note that this is not a strong opinion. For me the Dropbox notification for every synchronised saved was more disruptive as it was removing the focus from the ScummVM window and causing the system cursor to be displayed in addition to the ScummVM one (until I turned those notifications off). In comparison the OSD message is not as disruptive :-P

bgK:
I'm going to push the other commit in this pull request to master as it is not controversial.

Regarding the OSD message, do we need more time to reach a consensus?

lotharsm:
@bgk: No objections from my side, looking forward to see this merged as-is.

bgK:
Thanks for comments.

PR #1917, opened 2019-11-06
TTS: Improve Windows audio, remove sphelper-scummvm.h

sluicebox:
This removes sphelper-scummvm.h, a large patched Microsoft SDK header whose inclusion in the codebase raises the typical open source concerns: https://bugs.scummvm.org/ticket/11143.

Personally, I'm just offended that Microsoft ships a header that infects us with IsBadWritePtr().

In the process I found that we're setting the audio quality to SPSF_11kHz8BitMono, which is noticeably poor. I thought it had been my imagination that Windows TTS sounded bad on ScummVM but not my other programs. I don't see why we're setting a format at all. That's an OS setting so I've removed it, which conveniently removes another sphelper dependency, and makes TTS sound good.


criezy:
The idea sounds good.

I don't develop on Windows and I don't have an opinion on the details of the changes, but I noticed you replaced some `free` with `delete []` to free the result of `Win32::unicodeToAnsi()`. So I checked it and indeed that function uses `new char[]` and thus we should be using `delete []`. This change at least should be committed (and we have the same issue in `OSystem_Win32::convertEncoding`).

lotharsm:
Looks good to me too!

sluicebox:
Thanks TTS and Windows brain trusts! =) I'm a big fan of the TTS feature. Mergin'

lotharsm:
I just re-checked this again with my latest build: The TTS quality is much better now without any audible lags. Great job!

PR #1918, opened 2019-11-06
ENGINE: The Griffon Legend

sev-:
Here is a complete engine for a freeware zelda-like RPG _The Griffon Legend_.

The original game was written in FreeBASIC by Syn9 as part of some contest and published with the full source code. Then it was rewritten to C and ported to GCW0 by Dmitry Smagin, and this is the code I based this implementation on.

- The original page: https://web.archive.org/web/20160315011302/http://syn9.thehideoutgames.com/index_backup.php?table=griffonlegend

- The mirror where you still can download the game: http://www.phatcode.net/downloads.php?id=430
- @dmitrysmagin port: https://github.com/dmitrysmagin/griffon_legend

I plan to make the game downloadable from scummvm.org once the engine is merged.

Screenshot 2019-11-06 at 19 24 36
Screenshot 2019-11-06 at 19 25 50


dmitrysmagin:
Cool! I didn't know that it was possible to fit this game into the scummvm. Thumbs up!

engines/griffon/cutscenes.cpp
```diff
@@ -0,0 +1,553 @@
...
+	for (float y = 0; y < 100; y += _fpsr) {
+		_videoBuffer2->fillRect(Common::Rect(0, 0, _videoBuffer2->w, _videoBuffer2->h), 0);
+		_videoBuffer->setAlpha((int)((100.0 - y) / 100 * 255));
+		_videoBuffer->blit(*_videoBuffer2);
+
+		g_system->copyRectToScreen(_videoBuffer2->getPixels(), _videoBuffer2->pitch, 0, 0, _videoBuffer2->w, _videoBuffer2->h);
+		g_system->updateScreen();
+
+		if (g_system->getEventManager()->pollEvent(_event)) {
+			if (_event.type == Common::EVENT_QUIT) {
```
dreammaster: You might want to consider doing the quit event check as a macro. It just seems a bit redundant to keep seeing it repeated so many times in this file.

engines/griffon/dialogs.cpp
```diff
@@ -0,0 +1,761 @@
...
+				destColumn = 0;
+			else if (i == 13 && !config.effects)
+				destColumn = 0;
+			else if (i == 15 || i == 17) {
+				int vol = (i == 15 ? config.musicVol : config.effectsVol) * 9 / 255;
+				vol = CLIP(vol, 0, 9);
+
+				strcpy(line, "[----------]");
+				line[vol + 1] = 'X';
+				optionValues[i] = line;
```
dreammaster: This looks suspect.. a single static line is being assigned to different option value indexes. Is this okay?

engines/griffon/draw.cpp
```diff
@@ -0,0 +1,1230 @@
...
+							rcSrc.left = 16 * (int)(RND() * 2);
+							rcSrc.top = 80;
+							rcSrc.setWidth(16);
+							rcSrc.setHeight(16);
+
+							rcDest.left = _npcInfo[i].bodysection[10 * ff].x - 8;
+							rcDest.top = _npcInfo[i].bodysection[10 * ff].y - 8;
+
+							int x = 192 + ((int)(_itemyloc + ff * 5) % 3) * 64;
+							if (x > 255)
```
dreammaster: If there are a lot of boundary checks like this in the engine, consider maybe using the MIN macro instead of separate if checks

engines/griffon/draw.cpp
```diff
@@ -0,0 +1,1230 @@
...
+					rcDest.setWidth(24);
+					rcDest.setHeight(24);
+
+					if (_npcInfo[i].pause > _ticks && _npcInfo[i].shake < _ticks) {
+						_npcInfo[i].shake = _ticks + 50;
+						rcDest.left = rcDest.left + (int)(RND() * 3) - 1;
+						rcDest.top = rcDest.top + (int)(RND() * 3) - 1;
+					}
+					_anims[sprite]->blit(*_videoBuffer, rcDest.left, rcDest.top, Graphics::FLIP_NONE, &rcSrc);
+					// } else {
```
dreammaster: Unless there's a particular need to keep the commented out code present, it'd be better to simply remove it; it'll still be in the history if it's ever needed. Otherwise, maybe add a comment as why the block is commented out, and why it might be needed again in the future

engines/griffon/engine.cpp
```diff
@@ -0,0 +1,335 @@
...
+			_player.inventory[i] = 9;
+}
+
+void GriffonEngine::newGame() {
+	intro();
+
+	if (_shouldQuit)
+		return;
+
+	_player.px = 0;
```
dreammaster: I personally think it would be cleaner to have a separate "reset" method on your player class that sets all these, rather than just having them all here in the engine.cpp file

engines/griffon/griffon.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
```
dreammaster: I noticed this file doesn't have the additional section about the origins. Is this just an oversight, or intentional because this file has only newly created code? I've fine with either case, just want to make sure it wasn't an oversight.

engines/griffon/configure.engine
```diff
@@ -0,0 +1,3 @@
+# This file is included from the main "configure" script
+# add_engine [name] [desc] [build-by-default] [subengines] [base games] [deps]
+add_engine griffon "The Griffon Legend" no
```
dreammaster: The configuration is missing the setting.. "highres" I think? to indicate that an engine uses a high color mode

engines/griffon/griffon.h
```diff
@@ -0,0 +1,551 @@
...
+	int _ticks, _ticksPassed, _nextTicks;
+	float _fp, _fps, _fpsr; // CHECKME: _fp and _fps seems to be integers
+	int _secsingame, _secStart;
+
+	Graphics::TransparentSurface *mapImg[4];
+
+	Common::Rect rcSrc, rcDest;
+
+	// -----------special case
+	bool _dontDrawOver;   // used in map24 so that the candles don't draw over the boss, default set to 0
```
dreammaster: Nitpick, but presumably default is set to "false"

engines/griffon/logic.cpp
```diff
@@ -0,0 +1,2359 @@
...
+ 16 - armour chest
+ 17 - citadel master key
+ 18 - sword3
+ 19 - shield3
+ 20 - armour3
+
+*/
+
+// element tile locations
+const int elementmap[15][20] = {
```
dreammaster: Again a bit pedantic, but in cases like this with arrays being compiled into the executable, maybe use a smaller type where feasible, like "int8". Unless the array gets compiled in as int anyway, it'll help save a bit of needlessly wasted space.

engines/griffon/dialogs.cpp
sev-: Yes, it is a visual indicator:

Screenshot 2019-11-07 at 00 22 45

engines/griffon/griffon.h
sev-: Fixed

engines/griffon/configure.engine
sev-: Good catch. Added "16bit" there

engines/griffon/griffon.cpp
sev-: Well, in fact, it contains a few references to the original code, so added. But I'm considering leaving detection.cpp and console.cpp without this giantic header.

engines/griffon/logic.cpp
sev-: Good suggestion, fixed.

engines/griffon/engine.cpp
sev-: Good point. Fixed.

engines/griffon/draw.cpp
sev-: Done.

engines/griffon/cutscenes.cpp
sev-: Done

engines/griffon/draw.cpp
sev-: I am a bit torn on this one. Adding it will improve the code reability only slightly, but will require tons of work...

dreammaster: Don't worry about it. After the further review I did of the remainder of the engine, there weren't really enough occurrences for me to be concerned about it.

engines/griffon/detection.cpp
```diff
@@ -0,0 +1,85 @@
...
+};
+}
+
+class GriffonMetaEngine: public AdvancedMetaEngine {
+public:
+	GriffonMetaEngine() : AdvancedMetaEngine(Griffon::gameDescriptions, sizeof(ADGameDescription), griffonGames) {
+	}
+
+	virtual const char *getEngineId() const {
+		return "gnap";
```
bgK: That does not seem correct.

digitall:
@sev- : Couple of minor GCC warnings fixes are in a branch on my fork ... The commit is:
https://github.com/digitall/scummvm/commit/0fe088967bdc69936a1d8d96dbcfdaaea54ce45c

sev-:
@digitall thanks, applied

sev-:
Thanks all, merging.

PR #1919, opened 2019-11-07
Fix free/delete inconsistencies in OSystem_Win32::convertEncoding

criezy:
The current implementation of OSystem_Win32::convertEncoding causes some arrays allocated with `new []` being freed using `free()` and not `delete []`. This is an attempt to fix this.

Since I am not even able to compile the code, not being on Windows, I would appreciate if others could review and try this before this is pushed to master.

This commit also removes what appeared to be an unused variable (`wString`) and adds a sanity check as the result of `Win32::ansiToUnicode` could be null.

### DIFF
--- a/backends/platform/sdl/win32/win32.cpp
+++ b/backends/platform/sdl/win32/win32.cpp
@@ -459,9 +459,6 @@ char *OSystem_Win32::convertEncoding(const char* to, const char *from, const cha
 	}
 
 	// Add ending zeros
-	char *wString = (char *) calloc(sizeof(char), length + 2);
-	memcpy(wString, string, length);
-
 	WCHAR *tmpStr;
 	if (Common::String(from).hasPrefixIgnoreCase("utf-16")) {
 		// Allocate space for string and 2 ending zeros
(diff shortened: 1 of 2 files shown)


sluicebox:
I stepped through the changes in both codepaths on Windows with several test inputs and they are both fixed and doing what's intended.

The comment "Win32::unicodeToAnsi ses new to allocate" is missing the "u" and the project's code conventions are for no space in "delete []", which I only know from having just done this =)

backends/platform/sdl/win32/win32.cpp
```diff
@@ -492,7 +498,15 @@ char *OSystem_Win32::convertEncoding(const char* to, const char *from, const cha
 	} else {
 		result = Win32::unicodeToAnsi(tmpStr, Win32::getCodePageId(to));
 		free(tmpStr);
-		return result;
+		if (!result)
+			return nullptr;
+		// Win32::unicodeToAnsi ses new to allocate the memory. We need to copy it into an array
```
SupSuper: Typo

SupSuper:
Compiles fine here. Shame we can't just wrap them in managed Strings and avoid manual allocation, since they're Win32 types.

criezy:
Thank you to both of you. I have fixed the typo and `delete[]` and I will now merge this.

PR #1920, opened 2019-11-08
WIN32: Exit when WinSparkle runs installer

sluicebox:
Currently, when WinSparkle runs the downloaded installer, ScummVM remains open and must be manually exited or else the installer will fail. This is awkward since at this point the installer has focus.

WinSparkle has callbacks to handle this, so now ScummVM will exit as soon as WinSparkle runs the downloaded installer. At this point the user has clicked the Install Update button in the WinSparkle dialog so this isn't surprising behavior.

To test this, set SCUMMVM_VERSION to "2.0.0", or patch the version fields in scummvm.rc, and use the Check Now update button. After 2.1.0 downloads click Install Update.

https://bugs.scummvm.org/ticket/10368

### DIFF
--- a/backends/platform/sdl/win32/win32.cpp
+++ b/backends/platform/sdl/win32/win32.cpp
@@ -117,7 +117,7 @@ void OSystem_Win32::initBackend() {
 
 #if defined(USE_SPARKLE)
 	// Initialize updates manager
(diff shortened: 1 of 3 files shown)


sev-:
Nice, thanks!

PR #1923, opened 2019-11-09
POSIX: Fix compiler warning

jepael:
Parameter name isDirectory shadowed a method with same name.
The new name isDirectoryFlag has been also used elsewhere in similar situations.


bgK:
Thanks!

PR #1925, opened 2019-11-09
GUI: Remove build date from version variables

ryandesign:
See https://bugs.scummvm.org/ticket/11246

### DIFF
--- a/base/version.cpp
+++ b/base/version.cpp
@@ -60,11 +60,9 @@ const char *gScummVMVersion = SCUMMVM_VERSION SCUMMVM_REVISION;
 static const char *version_cookie __attribute__((used)) = "$VER: ScummVM " SCUMMVM_VERSION SCUMMVM_REVISION " (" AMIGA_DATE ")";
 #endif
 #ifdef __PLAYSTATION2__
-const char *gScummVMBuildDate = "Git Master"; /* ScummVM Git Master */
 const char *gScummVMVersionDate = SCUMMVM_VERSION SCUMMVM_REVISION " - PlayStation2";
 const char *gScummVMFullVersion = "ScummVM " SCUMMVM_VERSION SCUMMVM_REVISION " - PlayStation2";
 #else
-const char *gScummVMBuildDate = __DATE__ " " __TIME__;
 const char *gScummVMVersionDate = SCUMMVM_VERSION SCUMMVM_REVISION " (" __DATE__ " " __TIME__ ")";
 const char *gScummVMFullVersion = "ScummVM " SCUMMVM_VERSION SCUMMVM_REVISION " (" __DATE__ " " __TIME__ ")";
 #endif
--- a/base/version.cpp
+++ b/base/version.cpp
@@ -63,8 +63,8 @@ static const char *version_cookie __attribute__((used)) = "$VER: ScummVM " SCUMM
 const char *gScummVMVersionDate = SCUMMVM_VERSION SCUMMVM_REVISION " - PlayStation2";
 const char *gScummVMFullVersion = "ScummVM " SCUMMVM_VERSION SCUMMVM_REVISION " - PlayStation2";
 #else
-const char *gScummVMVersionDate = SCUMMVM_VERSION SCUMMVM_REVISION " (" __DATE__ " " __TIME__ ")";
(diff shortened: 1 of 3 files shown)


sev-:
And why, amy I ask?

DrMcCoy:
Instead of removing the date/time outright, ScummVM could heed [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/). I.e. if `SOURCE_DATE_EPOCH` is set, that timestamp is compiled into the binary instead of the current one.

In fact, IIRC, gcc already supports that. When `SOURCE_DATE_EPOCH` is set, `__DATE__` and `__TIME__` evaluate to the correct deterministic values.

DrMcCoy:
(And since `__DATE__` and `__TIME__` is the only thing this PR touches, I for one would say this PR is unnecessary.)

criezy:
What are the advantages of having the build date and time though? Is there any good reason not to accept this change?

@sev- For the why, you can read https://reproducible-builds.org

sluicebox:
I'd like to hear the justification for removing one of the most useful pieces of information from builds that I use all the time.

And before you say reproducible builds, which is pretty funny to even mention on this project, try reproducing the Windows 2.1.0 release right now. When version strings are the blocker, we can talk about making them less useful.

Maybe I'm in the minority here, but I don't have all of the git hashes memorized. How *isn't* the build date useful? I always want to know when a binary was built. In a bug report, that might be the only piece of information I trust a normal person to accurately report.

criezy:
I can agree that a date is easier to parse than a commit hash, but wouldn't the date of the head commit be more useful than the build date? For nightly builds it doesn't make much of a difference, but for users making their own builds and for the release builds the date being shown may currently not be a good indication of the date of the checkout used for the build.

criezy:
Also to clarify my own position on this, I do agree with @sluicebox that the date can be a useful information, so I don't think the change as proposed here should be accepted. Also as indicated by @DrMcCoy we may not even need any change to allow reproducible builds, which was the reason given in the bug tracker for this change.

However before rejecting this out of hand I am wondering if it could be acceptable in a different form, which would not degrade the current information we have in the version string, and possibly make it better. Which is why I asked if having the head commit date would not be more useful than the build date.

Also it seems to me that in release builds the release version itself, for example `2.1.0`, is the useful information and knowing when it was build doesn't tell us anything useful. So a change that only removes the date from those release builds would also be acceptable to me.

SupSuper:
I agree with @criezy that the commit date is much more useful than the build date. It's both deterministic and maintains the shorthand for "how old is this build". And since we already have a script to get the latest git hash, perhaps it could get the date as well.

A build date is only accurate in the context of nightlies. Otherwise any dev can compile any commit on any date, rendering them completely disconnected.

sev-:
Yes, I was thinking exactly that. That the commit date is a good compromise. That would in fact be a much better solution even for our release purposes, as currently we have build date differences for every platform, which are all built on top of the same commit.

Still, there are 2 other cases which need to be taken care of: building out of the source tarballs, and building so-called "dirty build", the one with local changes.

For the first case, I think, we could create some file with the revision tag and date, and feed from that when git command provides no results, and for the second case, we have to use the actual build date.

ryandesign:
I built scummvm and ran it. I noticed the window displayed the build date. [Including the build date in a program is an antipattern](https://reproducible-builds.org/docs/version-information/), so I [reported this bug](https://bugs.scummvm.org/ticket/11246). I was then asked to submit a PR to fix it, so here it is.

I am not familiar with scummvm so I cannot say whether there are other aspects of the build that make in nonreproducible. I can only say that making this change will make it more reproducible.

I was not aware that setting `SOURCE_DATE_EPOCH` will have an effect on the values of `__DATE__` and `__TIME__` when compiling with gcc. That's good, but not everyone compiles with gcc. On macOS, for example, one is most likely compiling with clang, which [does not appear to support `SOURCE_DATE_EPOCH`](https://reviews.llvm.org/D20791).

I agree that having the commit date in displayed in the program could be useful (and there is a long comment in version.cpp to that effect), but the commit date is not the same thing as the build date.

sev-:
Closing this. If somebody implements extracting commit date and including it to our build system (both makefiles and create_project), a new pull request needs to be implemented.

Simply killing useful functionality is way too hopeful for someone else coming and reimplementing it.

PR #1926, opened 2019-11-09
SDL: Fix Windows screenshots with unicode paths

sluicebox:
Use Common::FSNode to test for screenshot existence instead of
SDL_RWFromFile(). SDL does different character encoding conversions
which fail on Windows when a unicode character is in the path.

Fixes bug #11118


bluegr:
Nice catch! Merging

PR #1927, opened 2019-11-10
 MOHAWK: RIVEN: Allow for running in either 16-bit or 32-bit pixel formats - DO NOT MERGE

BallM4788:
On the 3DS, video playback and all visual effects for Riven are slow due to constant pixel format conversion; Riven's engine is currently hardcoded to use RGB565, while the 3DS backend uses RGBA8888 throughout. This pull request adds several preprocessing directives which allow for the Riven engine to run in 8888 on 3DS builds of ScummVM, but remain in 565 on all other platforms.

Important Notes:
- The WIDTHMULTI macro is necessary for 3DS builds because, without it, implementing an 8888 format causes the blend transition effect to occur only on the left half of the screen(s).
- In 3DS builds, the pixel format MUST be pulled from the 3DS backend's list of supported formats (i.e. not created with Graphics::createPixelFormat), or else video playback and visual effects will remain slow.

### DIFF
--- a/engines/mohawk/riven_graphics.cpp
+++ b/engines/mohawk/riven_graphics.cpp
@@ -285,10 +285,17 @@ public:
 				uint16 *src1 = (uint16 *) _mainScreen->getBasePtr(0, y);
 				uint16 *src2 = (uint16 *) _effectScreen->getBasePtr(0, y);
 				uint16 *dst = (uint16 *) screen->getBasePtr(0, y);
-				for (uint x = 0; x < _mainScreen->w; x++) {
+
+				uint16 widthMulti = _mainScreen->format.bytesPerPixel / 2;
+				for (uint x = 0; x < _mainScreen->w * widthMulti; x++) {
 					uint8 r1, g1, b1, r2, g2, b2;
-					Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src1++, r1, g1, b1);
-					Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src2++, r2, g2, b2);
+					if (_effectScreen->format.bytesPerPixel == 2) {
+						Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src1++, r1, g1, b1);
+						Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src2++, r2, g2, b2);
+					} else {
+						Graphics::colorToRGB< Graphics::ColorMasks<8888> >(*src1++, r1, g1, b1);
+						Graphics::colorToRGB< Graphics::ColorMasks<8888> >(*src2++, r2, g2, b2);
+					}
 
 					uint r = r1 * alpha + r2 * (255 - alpha);
 					uint g = g1 * alpha + g2 * (255 - alpha);
@@ -298,7 +305,11 @@ public:
 					g /= 255;
 					b /= 255;
 
-					*dst++ = (uint16) Graphics::RGBToColor< Graphics::ColorMasks<565> >(r, g, b);
+					if (_effectScreen->format.bytesPerPixel == 2) {
+						*dst++ = (uint16) Graphics::RGBToColor< Graphics::ColorMasks<565> >(r, g, b);
+					} else {
+						*dst++ = (uint16) Graphics::RGBToColor< Graphics::ColorMasks<8888> >(r, g, b);
+					}
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


bgK:
Did you measure the time that is actually spent doing the color conversion in the 3DS backend?

Otherwise:
* We generally don't write platform-specific code in the game engines. So, changing the 3DS backend to natively support the 565 pixel format would be better (I did it as a test a while back but the performance gains did not justify the code complexity). Or alternatively changing the Riven engine to adapt according to the preferred backend pixel format. The former would be better as it would help other engines.
* There are other places in the Riven engine that assume the pixel format is 16bits.

BallM4788:
I got a new branch going on [here](https://github.com/BallM4788/scummvm/commit/5272ad2b3612ce9b4695d2401e47375201b3d2dc) for working on converting the 3DS backend to RGB565. So far, everything seems to work except for the lack of transparency on the cursor and the menu overlay (obviously due to the absence of an alpha byte). Right now I'm thinking about how to get around that.

bgK:
> I got a new branch going on [here](https://github.com/BallM4788/scummvm/commit/5272ad2b3612ce9b4695d2401e47375201b3d2dc) for working on converting the 3DS backend to RGB565. So far, everything seems to work except for the lack of transparency on the cursor and the menu overlay (obviously due to the absence of an alpha byte). Right now I'm thinking about how to get around that.

Maybe keeping the screen in a 32 bits pixel format, and simply using the hardware to do the 565 -> 8888 conversion would be a better approach as it would not hurt other games using a 32bit pixel format directly. See this quick hack: https://github.com/bgK/scummvm/commit/a75c3006a4b2fe9cf11664c78e80055035b1e89f.

BallM4788:
I've been looking through the source code for citro3D and libctru trying to formulate a plan of action, and so far it seems to me like the best way to do this is to the following each time a game is launched based on its preferred color format:
- Change the input and output color format of the _gameTopTexture and _gameBottomTexture Sprites;
- Go in and change the color format of the frame buffer in the C3D_RenderTarget of each screen;
- Change the color formats of the screens themselves with libctru's "gfxSetScreenFormat()".

The primary goal being to keep the number and amount of 16bit-to-32bit / 32bit-to-24bit / etc. conversions per frame during gameplay at an absolute low. For drawing the cursor in games running in RGB565, I might try to borrow from the method the Riven engine uses to color blend the flies effects.

This is just the general gameplan at the moment. I'm still working on the specifics, but I'm almost certain that this is all doable.

BallM4788:
I need to understand the workings of libctru and citro3d more before I can proceed with my above plan. In the meantime, I've changed this pull request from it's original proposal (adding preprocessor macro defs and conditional inclusions for improved performance on the 3DS specifically) to the less-specialized proposal of allowing for running in either 16-bit or 32-bit pixel formats. No preprocessing like in the original pull request.

Here is the new summary:
> If platform has a list of supported formats, refer directly to the most
> preferred format on the list instead of using Graphics::PixelFormat or
> Graphics::createPixelFormat to replicate the format. This helps to reduce
> or even prevent video and effects slowdown on lower-power platforms such
> as the 3DS.

I have tested this on the 3DS, and to my knowledge everything is running properly in RGBA8888, including the flies/fireflies effect; that last one is one I had missed before. Please let me know if further modifications need to be made.

sev-:
Why did you close it?

BallM4788:
Oops.

bluegr:
The changes here are quite straightforward after the refactoring, and allow playing the game with the native pixel format.

Well done! +1 from me

@bgK: happy to merge this?

engines/mohawk/riven_graphics.cpp
```diff
@@ -285,10 +285,17 @@ class TransitionEffectBlend : public TransitionEffect {
...
+				uint16 widthMulti = _mainScreen->format.bytesPerPixel / 2;
+				for (uint x = 0; x < _mainScreen->w * widthMulti; x++) {
 					uint8 r1, g1, b1, r2, g2, b2;
-					Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src1++, r1, g1, b1);
-					Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src2++, r2, g2, b2);
+					if (_effectScreen->format.bytesPerPixel == 2) {
+						Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src1++, r1, g1, b1);
+						Graphics::colorToRGB< Graphics::ColorMasks<565> >(*src2++, r2, g2, b2);
+					} else {
+						Graphics::colorToRGB< Graphics::ColorMasks<8888> >(*src1++, r1, g1, b1);
```
bgK: This is not correct because the pixel format obtained from `g_system->getSupportedFormats().front();` could be any 32-bits pixel format, and here we assume ARGB.

To be fair, I'm not sure I want to accept the changes to Riven to make it work with any pixel format. I'm not sure moving the color conversion to the engine fixes anything. Did you test the performance of using hardware accelerated color conversion on the 3DS?

BallM4788: I've pretty much abandoned this pull request, as I'm making very good progress on getting ScummVM to switch the system output between RGBA8888 and RGB565 on a per-game basis: BallM4788/scummvm@92fa032e94966e76e616f9115d0ab481c57e9913

To answer your question though, I did test the performance of this pull request on real 3DS hardware, and it runs full-speed with no slowdown. But yeah, no hard feelings on my part if you don't merge this.

bluegr:
Since this pull request has been abandoned, I’m closing it. A new pull request can be opened with the new approach, as described above

PR #1929, opened 2019-11-10
AMIGAOS4: Bits of cleanup

raziel-:
- Cleanup, formatting and more verbose debug messages
- Change printf to snprintf
- Increase stack cookie

Tested locally and working


digitall:
Since this only touches AMIGAOS4 port specific code, and @raziel- is the port maintainer, then this should be fine to merge as-is. Merging...

PR #1930, opened 2019-11-11
POSIX: Use fstat instead of fseek / ftell to retrieve file stream sizes

bgK:
fstat is generally faster as is does not cause the IO buffer to be
invalidated / refilled.

Benchmark results for the startup time of the SCI engine with Gabriel
Knight 1 CD:
- Linux, glibc, spinning HDD, fseek/ftell: 140 ms
- Linux, glibc, spinning HDD, fstat: 100 ms
- 3DS, newlib, SD card, fseek/ftell: 68 s
- 3DS, newlib, SD card, fstat: 11 s

### DIFF
--- a/backends/fs/posix/posix-fs.cpp
+++ b/backends/fs/posix/posix-fs.cpp
@@ -33,7 +33,7 @@
 #define FORBIDDEN_SYMBOL_EXCEPTION_srandom
 
 #include "backends/fs/posix/posix-fs.h"
-#include "backends/fs/stdiostream.h"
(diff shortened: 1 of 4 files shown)


digitall:
This looks like a pretty reasonable change which should be fine on all POSIX platforms.

@bgK I assume you are going to wait a week or so in case any POSIX porters have any issues with this?

bgK:
> @bgK I assume you are going to wait a week or so in case any POSIX porters have any issues with this?

Yes, I'm planning to wait a bit before merging. I'm also not going to backport this to the 2.1 branch.

PR #1932, opened 2019-11-13
TOLTECS: Added Czech language support

AfBu:
- added Czech version definition
- reported in trac:#11263


digitall:
Looks pretty reasonable and self-contained change to TOLTECS engine to just the detection entry for this variant. Commit message fits our guidelines so will merge as-is.

@AfBu : Thank you for your code contribution.

PR #1933, opened 2019-11-14
DEVTOOLS: Add missing library to cmake-base create_project

janisozaur:


digitall:
This is to add the rpcrt4 library which is for Remote Procedure Call (RPC). I am surprised that is a necessary dependency of the devtools since I don't think any of them use network services:
https://docs.microsoft.com/en-us/windows/win32/rpc/developing-32-bit-windows-applications

@janisozaur : Any idea where this dependency is being pulled in?

janisozaur:
I find your surprise… surprising.

rpct4:
https://github.com/scummvm/scummvm/blob/2078aa732db32cc3333a8cad264a07da07a2f909/devtools/create_project/create_project.cpp#L1624-L1636

advapi32:
https://github.com/scummvm/scummvm/blob/2078aa732db32cc3333a8cad264a07da07a2f909/devtools/create_project/create_project.cpp#L1228-L1242

digitall:
@janisozaur : Ah, I see. They are dependencies for create_project which uses them for UUID generation and Windows registry access. OK. Will merge as-is.

PR #1936, opened 2019-11-15
ENGINES: Add NOMIDI GUI option flag to known games

raziel-:
I tested these games and while i could have easily overseen something, there was no MIDI output (albeit set to use in launcher) on my hardware and absolutely no SYSEX traffic.

So, i assumed (hopefully correctly) that those game variants do not feature MIDI playback et al and as such can be lifted from their MIDI options tabs in launcher.

While there might be engines where all of the game variants do not use MIDI, i did not add to those, as i cannot test them. (i.e. Gob, TsAGE, Gnap etc.) (Thanks athrxx)

Feel free to chime in


engines/mads/detection_tables.h
```diff
@@ -56,7 +56,7 @@ static const MADSGameDescription gameDescriptions[] = {
 			Common::EN_ANY,
 			Common::kPlatformDOS,
 			ADGF_NO_FLAGS,
-			GUIO5(GUIO_NOSPEECH, GAMEOPTION_EASY_MOUSE, GAMEOPTION_ANIMATED_INVENTORY, GAMEOPTION_ANIMATED_INTERFACE, GAMEOPTION_NAUGHTY_MODE)
+			GUIO6(GUIO_NOSPEECH, GAMEOPTION_EASY_MOUSE, GAMEOPTION_ANIMATED_INVENTORY, GAMEOPTION_ANIMATED_INTERFACE, GAMEOPTION_NAUGHTY_MODGUIO2, GUIO_NOMIDI)
```
bluegr: This is wrong, and won't compile

raziel-:
Thank you, fixed

bluegr:
This is kind of a mixed bag now. The flags are applied only to specific versions and languages, so the end result is inconsistent.

The no midi flag should be consistently applied for the same game variants in each engine

raziel-:
No Problem.
As i wrote i could only test those specific versions which i own.

I don't know if the whole engines, where those games belong to, are completely not using midi, but I guess not.

raziel-:
Should i close this?

bluegr:
@raziel-: These flags should be applied consistently per engine for all of its supported game variants. So, we should either apply the flags that way, or don’t apply them, to have a consistent behavior

raziel-:
@bluegr

I was told on irc that there are cases where two games from the same engine offer different audio.
iirc it was the floppy version of Inherit the Earth which features midi, but the cd release doesn't.

Same as with SCUMM, while the majority of games feature midi music, at least the cd version of Monkey Island 1 (from the Bounty Pack, which i own) does not (instead it uses cd audio tracks and was completely rid of midi).

If we are going to set every game of an engine to use midi and show the option (just because one game uses it), the bug reports will still come in from time to time, i guess.

That was the second main reason for for me *not* setting every game to NOMIDI (since i don't have them and can't test them)

But maybe i misunderstand you?

bluegr:
You misunderstand me. You are right that there are cases where floppy and CD versions use different types of music.

However, there are cases like the gob engine, where the English CD variant has been flagged as “no MIDI”, whereas the French and German CD variants haven’t.
 
So, it would be more consistent to apply these flags for the same kinds of game variants (e.g all the CD variants, regardless of language).

Consulting the engine code is also a good indicator about each case. For example, for the access engine, the demo of Amazon uses a different music type than the full version.

Having said that, if we’re unsure about the behavior of a specific game variant, the flags should not be applied. But if the music type is consistent, like in the toltecs engine, the flags should be applied to all the variants.

raziel-:
I don't know the code as good as i need to to be even remotely able to add to all game(s) variants and engines.
So i'm going to close this and apologize for the noise.

PR #1941, opened 2019-11-16
GUI: Add DropdownButtonWidget and use it in the launcher for mass add

bgK:
DropdownButtonWidget is a button split in two parts vertically. Clicking
the left part triggers a default action. Clicking the right part shows a
list of other actions the user can choose from.

Using this widget in the launcher lets 'Mass add' be a secondary action
of the 'Add' button, removing the need to press the shift key to
access the feature.

scummvm00005
scummvm00006
scummvm00007
scummvm00008

I'm still experimenting with the appearance of the dropdown menu, but would not mind feedback and suggestions.


lotharsm:
Nice improvement indeed!

However, I suggest to slightly increase the button width since at least the German translation get's trimmed and is not displayed correctly anymore.
trimmed

bgK:
> However, I suggest to slightly increase the button width since at least the German translation get's trimmed and is not displayed correctly anymore.

Ok, I've increased the width of the launcher's buttons.

lotharsm:
Thanks, looking great now:
nottrimmed

bluegr:
Nice work! I like the button and its behavior, and at a first glance, the code changes look good

bgK:
> I also tried it on iOS and it mostly works with touch screen, although the behaviour there could be improved. Tapping the arrow shows the Mass Add action, but tapping that one does not execute it - we first need to touch and drag to get the cursor on the action so that it gets highlighted, and then taping anywhere will execute it (and we already have this non intuitive behaviour with the popup widget). But that is not a blocker and can be improved later on.

I've pushed a commit to fix this (In the popup widget as well as the code for the menu is shared).

Interestingly that behavior was not happening on the 3DS because the backend injects an artificial mouse move event.

lotharsm:
It seems that the GUI translation is not working probably for the "new" Mass Add entry:
no-lang-massadd
Weblate doesn't report any fuzzy or missing strings, and I'm using the current version of translations.dat...

bgK:
> It seems that the GUI translation is not working probably for the "new" Mass Add entry.

Should be fixed now.

PR #1942, opened 2019-11-16
GUI: Use a dialog theme layout for the unknown game dialog

bgK:


criezy:
The change looks good.
One drawback though is that with the current code the width of the buttons get adapted to their label, while this is no longer the case after this change, and at least in French the Copy button is not wide enough to display its label.

sev-:
picture?

bgK:
> picture?

Before:
scummvm00009

After:
scummvm00010

sev-:
That’s great! Did you check it with 320x200? Is it still useable there?

bgK:
> That’s great! Did you check it with 320x200? Is it still useable there?

As mentioned by criezy there's an issue where the label of buttons can be cut:
scummvm00013

I haven't looked into it yet.

bgK:
Merged manually.

PR #1944, opened 2019-11-17
RFC: Remove the scummmodern GUI theme

bgK:
It seems to me the scummremastered theme has been well accepted. Perhaps the scummmodern theme is no longer necessary and can be removed. Doing so would remove the need to update it and test it when doing GUI changes.

What do you think?


tsoliman:
I might be in the minority here, but I actually prefer the old modern theme.

I didn't say anything about it before because having a new default that I strongly dislike and that others seem to really be excited about isn't a bad thing. Different people like different things.

I'm willing to help test and keep it working since I'm actively using it.

criezy:
I don't have a strong opinion there, but personally I have no issue with removing the modern theme.

From what I remember the layout files for the modern and remaster themes are actually identical though. So if we decide to keep the two, what could help would be to have a single copy of those rather than have to do the same changes to both sets whenever we change something.

Mataniko:
What would be nice if we just had theme inheritance, so maintaining both themes is simple (criezy is right, the remastered theme has virtually no changes in layout)

digitall:
It is a matter of aesthetic taste, but I prefer the colours of the older Modern theme... but that is maybe just familarity. Overall, no strong preference, but it would be nice to allow the same layout, but allow the colours to be varied as some users might prefer a different colour scheme i.e. various types of colour blindness / dsylexia or just a dislike of green / yellow :)

bgK:
Thanks all!

It seems like having multiple color schemes per theme would be a good way forward.

Closing.

PR #1945, opened 2019-11-17
SCI32: Fix QFG4 error when sliding down slope

sluicebox:
Fixes the error in the CD version when starting to automatically walk (then slide) down the top of the slope in room 800: https://bugs.scummvm.org/ticket/11268

I'm doing this as a PR so that the bug reporter can test it first since it replaces an existing script patch.


tsoliman:
Tested in the goo slide - I had 2 different saves where this happened for me and it no longer happens.

The GOG version never broke for me.

Good job!

sluicebox:
Great! That's consistent with what I'm seeing and what we know about this bug class. Mergin'

PR #1946, opened 2019-11-19
Setting revision define is applied to all sources, not just version.cpp

janisozaur:
https://github.com/scummvm/scummvm/blob/79d636ba8a34a3dd5942cd61933940bbc4c87541/Makefile.common#L214-L218
This line should apply the revision define to just version.cpp, but as you will see in the logs produced, it is done to all the sources instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ matrix:
 dist: trusty
 
 script:
-  - ./configure --enable-all-engines --enable-opl2lpt
+  - ./configure --enable-all-engines --enable-opl2lpt --enable-verbose-build
   - make -j 2
   - make test
   - make devtools


bluegr:
So, is there something to be done here?
Will this setting help to debug future builds?

criezy:
I think the idea of the PR was not so much to merge it than to illustrate, by looking at the Travis output for this PR, that the `-DSCUMMVM_REVISION=...` flags ends up being used to compile all the files and not just for the `base/version.o` target.

I am not really familiar with Makefile though, and I don't see why the target specific variable assignment doesn't work in this case. And I am unsure if it practice this issue has any unwanted side effect.

digitall:
I can replicate the issue on my local linux_x86_64 build with --enable-verbose-build. It is not all object files though ... For my build, it starts when the engine code is compiled. Very odd. Will enable this for Travis by merging this change as it is a good CI improvement anyway. Will take a look at why this is failing to match correctly in Makefile... though reason is not immediately apparent.

digitall:
Changes from this PR have been manually merged with fixup to commit message as commit 7bb28d486e0e9586f6a3b4d78301462fb6e3f596

I am going to close this PR as the remaining issue with Makefile rule matching is a minor bug, rather than a PR.

@janisozaur: If this is not fixed in the next few days, please file this as a bug at bugs.scummvm.org so we can keep track. Thanks for your code contribution.

digitall:
@janisozaur : Ah, have worked out the outline of why this is happening. It is the interaction of the two makefile rules for base/version.o:
https://github.com/scummvm/scummvm/blob/master/Makefile.common#L62
https://github.com/scummvm/scummvm/blob/master/Makefile.common#L217

The first rule makes base/version.o depend on all other files outside of the base folder to ensure this is rebuilt if any source changes are made. However, the side-effect of this is that every object outside of base gets the second rule applied resulting in the -DSCUMMVM_REVSION being appended to their CXXFLAGS used in the compilation line.

Commenting out line 62 stops this happening, but also will prevent the rebuild of version.o during incremental compilation if any source file is modified. I am not sure how we can limit the second rule to be just for that object compilation, but not for any object which recursively includes it.

Could do with a GNU Make expert here ...

digitall:
Aha... https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html#Target_002dspecific

The issue here is "There is one more special feature of target-specific variables: when you define a target-specific variable that variable value is also in effect for all prerequisites of this target, and all their prerequisites, etc. (unless those prerequisites override that variable with their own target-specific variable value). "

Hence this variable change is not just applied to base/version.o, but to all the other files which are linked by Line 62 rule :|

ykaliuta:
How about to make a separate rule for version.o and put -D in the recipe, not in the rule specific variable?

digitall:
https://github.com/scummvm/scummvm/pull/1952 has one possible solution...

digitall:
@ykaliuta : If you can suggest a better solution, then patches or pull request would be welcome.

ykaliuta:
I'm not sure about better, but I posted what I meant.

PR #1947, opened 2019-11-19
Add ccache

janisozaur:
This enables ccache on all the Travis jobs.

Note that Trusty ships with ccache 3.1.9, which is not sufficient due to https://bugzilla.samba.org/show_bug.cgi?id=10005. I took that opportunity to update the Travis CI job to use a supported distro, that also ships with updated ccache.

Addressing https://github.com/scummvm/scummvm/pull/1946 will improve direct hit rates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,11 +71,14 @@ matrix:
 dist: xenial
 
 script:
+  - ccache --show-stats > /tmp/ccache_before
   - export PATH="/usr/local/opt/ccache/libexec:/usr/lib/ccache:$PATH"
   - ./configure --enable-all-engines --enable-opl2lpt
   - make -j 2
   - make test
   - make devtools
+  - ccache --show-stats > /tmp/ccache_after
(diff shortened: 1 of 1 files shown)


digitall:
@janisozaur : Thanks for your code contribution. 

I will look at merging this manually as the commit messages do not comply with our guidelines, but this is a minor fixup i.e. they should be prefixed by "TRAVIS: " as per:
https://wiki.scummvm.org/index.php?title=Commit_Guidelines

digitall:
OK. Merged manually as 6162609..41efecc. One minor note, I did filter the side-effect change from enabling of ccache which removed the limit of just building the master branch. We have quite a few release branches and I don't think that would work very well. We can add additional branches if need be in future. Closing...

janisozaur:
> the limit of just building the master branch

Sorry, I needed this for testing locally, that was not meant to be part of the patch.

digitall:
No problem. I suspected that would be the case.

PR #1948, opened 2019-11-23
COMPOSER: add support for v1 hebrew versions (Windows + Macintosh)

BLooperZ:
Support for games and demos found in Hebrew CDs of "The Little Samurai" and "Imo and the King".

Note:
Demos from "Imo and the King" are in English.
Demos from "The Little Samurai" are in Hebrew.

Note 2: 
The games and demos from "Imo and the King" are trying to access `RETAIL2.RSC` from `DEMO` directory in CD because it is the one referenced in the `.INI`.
It does not affect the game, but triggers an error message when leaving the game.
(The exact same file is also available in actual game data directory - `IMO/MENU/RETAIL2.RSC`)

### DIFF
--- a/engines/composer/composer.cpp
+++ b/engines/composer/composer.cpp
@@ -92,7 +92,8 @@ Common::Error ComposerEngine::run() {
 				// mac version?
 				if (!_bookIni.loadFromFile("Darby the Dragon.ini"))
 					if (!_bookIni.loadFromFile("Gregory.ini"))
-						error("failed to find book.ini");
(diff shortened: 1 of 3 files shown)


BLooperZ:
The CDs are Macintosh/Windows Hybrid.
So it also supports the Macintosh version of games and demos from same CDs.

BLooperZ:
To avoid having to check many different file names for config file,
the engine gets the the config file name set in game description.

bluegr:
Nice work, thanks! Merging

PR #1949, opened 2019-11-24
COMMON: Add CJK encodings to Common::CodePage

lephilousophe:
It uses Encoding::convert to do the conversion first and try old method if it fails.
This also adds an invalid encoding which can be used when encoding is unknown (yet).
In this case encode and decode fail.

### DIFF
--- a/common/str-enc.cpp
+++ b/common/str-enc.cpp
@@ -20,8 +20,10 @@
  *
  */
 
+#include "common/encoding.h"
 #include "common/str.h"
 #include "common/ustr.h"
+#include "common/util.h"
 
 namespace Common {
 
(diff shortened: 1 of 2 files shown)


common/str-enc.cpp
```diff
@@ -303,11 +320,22 @@ void String::decodeOneByte(U32String &dst, CodePage page) const {
 }
 
 U32String String::decode(CodePage page) const {
-    U32String unicodeString;
-    if (page == kUtf8) {
+	if (page == kCodePageInvalid ||
+	        page >= ARRAYSIZE(g_codePageMap)) {
+		error("Invalid codepage");
+	}
+	char *result = Encoding::convert("UTF-32", g_codePageMap[page], _str, _size);
```
bgK: While I guess it's fine to connect ScummVM's two string encoding systems, it would be nice to merge them at some point.

As a side note, ResidualVM has it's own pre-existing system (that was lifted from ScummVM's in limbo Buried in Time engine): https://github.com/residualvm/residualvm/blob/master/common/iconv.h. So when we next merge changes from ScummVM there will be three string encoding systems.

Sorry for the ranty comment. No objections to merging from me.

lephilousophe: I agree that the current situation with two systems is far from ideal.
I must say that the Encoding class with char * stuff bothers me as it is far from optimal when used with Strings.
We allocate several blocks in convertIconv and then we have to copy back this in the String or U32String so we have multiples copies of data and I would prefer to minimize this.
Anyway, I just plumbed this to avoid dealing with a huge change as I have a pile changes needing this for Versailles.
Maybe a rework of the whole system could be considered in another PR.

bgK: I'm glad I'm not the only one to be unimpressed with the `Encoding` class.

bluegr:
Great work!
Indeed, we need to end up with one encoding method...

Merging

PR #1952, opened 2019-11-24
BUILD: Fix Version Define Being Added to All Code Objects

digitall:
Due to a quirk of target specific make variables which means they are
added to all pre-requisities, this resulted in the defines which were
meant to be applied only to the version module being applied to most
of the codebase.

This did not cause any direct issues, but was untidy and unexpected
behaviour which was reported in Pull Request 1946 by janisozaur.

Various fixes were tried, but this fix is from bgK and removes the
pre-requisities of the version object and instead builds it in all
cases using .FORCE.

However, this does cause a rebuild of the executable in all cases,
even if no source code change has occurred which is not perfect.

Better suggestions for a fix for these issues would be welcome.

```

two together.

- Bullet points are okay, too

- Use a hanging indent
```

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -56,15 +56,6 @@ CPPFLAGS := $(DEFINES) $(INCLUDES)
 DEPDIRS = $(addsuffix $(DEPDIR),$(MODULE_DIRS))
 DEPFILES =
 
-# Make base/version.o depend on all other object files. This way if anything is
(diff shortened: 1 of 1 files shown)


digitall:
This is one possible fix for the issue noted in:
https://github.com/scummvm/scummvm/pull/1946

Any suggestions for better solutions would be happily received...

digitall:
Closing this in favour of PR #1954 which looks like a better solution

PR #1953, opened 2019-11-24
SHERLOCK: Enable the use of TTS to read descriptions as a "narrator" for The Case of Serrated Scalpel

neuromancer:
SHERLOCK: Add TTS support to read descriptions 

I was recently experimenting with the new TTS feature to 
add a "narrator" to some old games (which usually contains 
a lot of text in form of descriptions of the environment). 
In particular, this PR is first attempt to add TTS support 
to Sherlock Holmes - The Case of Serrated Scalpel. 
I expect some discussion before merging this 
(if the idea makes sense for scummvm, of course)


bluegr:
Nice idea!

You should also wrap this around a game-specific option, which will be disabled by default, so that players can enable this feature at will.

As an example, check this commit: 1016838bd5e556a4882f2b07e993ea9e78d12f7f (mainly, the changes to detection.cpp and zvision.cpp)

neuromancer:
Neat!, I will implement a game-specific option to enable this and get back to you. Any suggested name for that option?

raziel-:
Just a question (but it holds true for all the GAMEOPTIONS), can this be made optional in the code?
So that it doesn't even pop up in the final build, IF TTS is not available/built in?

I know that for i.e. MIDI it has to be done by a per game decision (since some games don't support or have MIDI music), but for TTS there could be a check if it has been built in et al and only add/show the option, if that is true, couldn't there?

Just trying to understand the codebase a little bit better.

Thank you and sorry for the noise

neuromancer:
> Just a question (but it holds true for all the GAMEOPTIONS), can this be made optional in the code?
So that it doesn't even pop up in the final build, IF TTS is not available/built in?

That's a good idea. I just pushed a commit with that. The code for the option will be compiled only if `--enable-tts`  is used.

neuromancer:
A few limitations:
* It does not read the inventory descriptions: I don't know where they are exactly, I have little experience with this engine.
* It is only enabled for the english version of the game. 
* The voice is not configurable.

dreammaster:
For inventory descriptions when items are highlighted, you want the ScalpelUserInterface::lookInv method

neuromancer:
@dreammaster It seems that the `inv[temp]._description` in `lookInv` only renders the name of the item, but not the description:

https://github.com/scummvm/scummvm/blob/eae00fe6f9f0eb565f2c38447d705945db0e6ba6/engines/sherlock/scalpel/scalpel_user_interface.cpp#L721-L722

dreammaster:
If you want the full description that's shown when you click to look at an item.. that triggers a script call to doScript, which in turn calls the talkInterface method, I think it is, to actually display text embedded within the object's script.

bluegr:
Any news on this? Will the inventory descriptions be added to the mechanism as well?

neuromancer:
Sorry for the delay. I was on vacation: I'm finishing my trip on Sunday. I was planing to add inventory descriptions, but I had trouble understanding how to compute the pointer to the actual text description from the one to the name of the object. I will take a better look next week (but if you know how to do it, feel free to push that into this branch or leave a comment :smile: ).

neuromancer:
I just pushed a commit: it is working for inventory descriptions as well, @bluegr. Please review.

bluegr:
@dreammaster has done most of the work on this engine. The changes look good at a first glance, and it seems like the functionality could also be added to the Rose Tattoo. Have you tried it with that game?

neuromancer:
I have not tried to make Rose Tattoo working with voices like that yet. I worked in voice narrator of the "Lure of the Temptress" and it is working fine (it was a lot easier than this). Should I make a PR of that?

bluegr:
This is now in a good state to be merged. Thanks!

PR #1954, opened 2019-11-24
BUILD: Fix Version Define Being Added to All Code Objects

ykaliuta:
Due to a quirk of target specific make variables which means they
are added to all pre-requisities, this resulted in the defines which
were meant to be applied only to the version module being applied to
most of the codebase.

This did not cause any direct issues, but was untidy and unexpected
behaviour which was reported in Pull Request 1946 by janisozaur.

Fix it by defining a special rule for version.o with own flags
instead of using template rule.

Signed-off-by: Yauheni Kaliuta <yauheni.kaliuta@redhat.com>

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -136,6 +136,10 @@ ifdef CXX_UPDATE_DEP_FLAG
 	$(QUIET)$(MKDIR) $(*D)/$(DEPDIR)
 	$(QUIET_AS)$(CXX) $(CXX_UPDATE_DEP_FLAG) $(ASFLAGS) -c $(<) -o $*.o
 
+base/version.o: base/version.cpp
+	$(QUIET)$(MKDIR) $(*D)/$(DEPDIR)
+	$(QUIET_CXX)$(CXX) $(CXX_UPDATE_DEP_FLAG) $(CXXFLAGS) $(VERFLAGS) $(CPPFLAGS) -c $(<) -o $*.o
+
 else
 
 # Dumb compile rule, for C++ compilers that don't allow dependency tracking or
@@ -149,6 +153,9 @@ else
 	$(QUIET)$(MKDIR) $(*D)
 	$(QUIET_AS)$(CXX) $(ASFLAGS) -c $(<) -o $*.o
 
+base/version.o: base/version.cpp
+	$(QUIET)$(MKDIR) $(*D)
+	$(QUIET_CXX)$(CXX) $(CXXFLAGS) $(VERFLAGS) $(CPPFLAGS) -c $(<) -o $*.o
 endif
 
 # Build rule for assembler files
@@ -192,7 +199,7 @@ VER_EXTRA = $(shell echo $(VERSION) | cut -d. -f 3 | cut -c2-)
 ifdef AMIGAOS
 # Amiga needs date in specific format for the version cookie
 AMIGA_DATE = $(shell gdate '+%d.%m.%Y')
-base/version.o: CXXFLAGS:=$(CXXFLAGS) -DAMIGA_DATE=\"$(AMIGA_DATE)\"
+VERFLAGS += -DAMIGA_DATE=\"$(AMIGA_DATE)\"
 endif
 
 ######################################################################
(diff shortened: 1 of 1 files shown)


Makefile.common
```diff
@@ -136,19 +136,26 @@ ifdef CXX_UPDATE_DEP_FLAG
...
+	$(QUIET_CXX)$(CXX) $(CXX_UPDATE_DEP_FLAG) $(VCXXFLAGS) $(CPPFLAGS) -c $(<) -o $*.o
+
 else
 
 # Dumb compile rule, for C++ compilers that don't allow dependency tracking or
 # where it is broken (such as GCC 2.95).
 .cpp.o:
 	$(QUIET)$(MKDIR) $(*D)
-	$(QUIET_CXX)$(CXX) $(CXXFLAGS) $(CPPFLAGS) -c $(<) -o $*.o
+	$(QUIET_CXX)$(CXX) $(VCXXFLAGS) $(CPPFLAGS) -c $(<) -o $*.o
```
lephilousophe: Why are you using VCXXFLAGS here?

ykaliuta:
By accident. Thanks, updated.

Makefile.common
ykaliuta: By accident. Thanks, updated.

lephilousophe:
I personally prefer this than PR #1952.
Anyway maybe you could define VERFLAGS (or any other name) and have the version.o rule make use of $(CXXFLAGS) $(VERFLAGS) to clearly mark what is different between the standard rule and this one: the addition of version defines.

ykaliuta:
Sure. Updated.

lotharsm:
It seems that this patch helps improving the speed on consecutive builds with ccache after resetting the environment with "make" clean:

I ran a little test with ccache and went from 8b7429693d4ea490aef5202077b2e153bd04285a to 8f7df6230ea9cb6b985ffe22c2fe354704251f01. Between the runs, I cleared the ccache cache files to start from scratch. The following are the times needed for a `make clean; ./configure; make` run after I "initialized" ccache with building 8b7429693d4ea490aef5202077b2e153bd04285a.

with ccache w/o patch: real    3m59.729s
with ccache and patch: real    3m7.515s

Makefile.common
```diff
@@ -214,7 +221,7 @@ endif
 # Define the Subversion revision if available, either autodetected or
 # specified by the user, but only for base/version.cpp.
 ifneq ($(origin VER_REV), undefined)
-base/version.o: CXXFLAGS:=$(CXXFLAGS) -DSCUMMVM_REVISION=\"$(VER_REV)\"
+VERFLAGS = -DSCUMMVM_REVISION=\"$(VER_REV)\"
```
digitall: I think this might need some code to do this as append to the variable, rather than overwrite assign ... as per the original code. Since otherwise, the define for AMIGAOS is likely to get lost...

digitall:
This looks like a better solution than PR #1952 so happy to take this instead.

One point, @ykaliuta  ... The changes to the VERFLAGS look to assign to the new variable, rather than append as per the original CXXFLAGS addition. Just that I think in the AMIGAOS case this might lost that AMIGA_DATE define from VERFLAGS?

ykaliuta:
fine for me. Should I update it?

JFYI, I actually did not check the logic much. The background of the pull request is that one guy came to ##workingset channel on freenode with the question and I posted a proposal to the issue discussion. I was asked about code -- I made a demonstration of what I meant :) Feel free to change it to the needs if you think so.

digitall:
@ykaliuta : Yes, please do... ah you did :)

Yes, am aware... since that was me :) Thanks for the code contribution.

I think this should be fine as-is, but will get a bit more team review prior to merging. Will close PR #1952 in favour of this instead.

digitall:
Will leave this open until after the weekend in case we have any porters who have comments or if @sev- has time to review. If no issues, will merge early next week and we can see if the buildbot throws out any issues.

sev-:
Thanks! Merging

lotharsm:
I now have the confirmation from my buildbot: This fix really cuts the compilation time on subsequent builds _in half._ Great job!

PR #1956, opened 2019-11-28
3DS: Prevent magnify effect from updating while the GUI is active

BallM4788:
Prevents the coordinates of the magnification viewport from changing while the Options GUI is brought up.

TODO: When exiting the GUI, prevent the cursor's position within the GUI from changing the position of the magnification viewport. Possible solution: save in-game cursor coordinates separately from GUI cursor coordinates?


digitall:
bgK: This looks fairly reasonable and @BallM4788 implemented the 3DS magnify code originally so happy for me to merge this?

bgK:
Thanks!

> Possible solution: save in-game cursor coordinates separately from GUI cursor coordinates?

Sure, you can experiment with it and see if it feels right. The current behavior does not seem too bad to me though.

PR #1962, opened 2019-12-01
WINTERMUTE: Add some public demos

aviloria:
Add public demos for several wintermute games

### DIFF
--- a/engines/wintermute/detection_tables.h
+++ b/engines/wintermute/detection_tables.h
@@ -276,6 +276,18 @@ static const WMEGameDescription gameDescriptions[] = {
 		WME_ENTRY2s("data.dcp", "481b1bff44178ef2f3d879df05f65a96", 706580987,
 					"russian_language_pack.dcp", "f39ad478a711fa4b34d419ed4aac97bf", 49395113), Common::RU_RUS, ADGF_UNSTABLE, WME_1_9_3),
 
+	// Alpha Polaris (English Demo)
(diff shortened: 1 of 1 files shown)


sev-:
It would be nice, if you could also provide these demos, so we could upload them to our servers.

aviloria:
@sev: What do you prefer, origial URLs or a shared folder with all of them?

sev-:
Either. Shared folder is more convenient. Also, you could help with renaming the archives to scummvm format, and perhaps extracting those which are packed in .exe if any, so we could have multiplatform .zip files.

For the scummvm format naming, see the current demos page. Basically, it is named after autodetected target.

lolbot-iichan:
@aviloria, great job collecting all those, thank you!

>It would be nice, if you could also provide these demos, so we could upload them to our servers.
@sev- @aviloria Please note that right now IMHO only 2D games worth uploding, since 2.5D WME games are out of ScummVM's scope and are currently not supported by ResidualVM.

aviloria:
A shared folder with the wintermute demos in 2 flavours: original files and scummvm-cooked files
https://mega.nz/#F!PENWjAAR!guQofHCAz4lqLDQFp2IjJQ

I also have added 2 german demos for "Art of Murder" not in my previous commit... I'll add them later to the repository

bluegr:
Thanks for your work, and for finding these demos!
Merging

Uploading the demos is a separate task, and can be done independently of this PR

PR #1963, opened 2019-12-01
WINTERMUTE: restored commented out stack->pop()

lolbot-iichan:
Unused variables were cleaned up at
47811167d5ddd9cbf64099ad36f2da5ae413fc86, commenting out a bit too much
code, since stack->pop() has a side effect (popping a value from data
stack).


bluegr:
Nice catch! Merging

PR #1966, opened 2019-12-02
WINTERMUTE: Added more Steam releases

aviloria:
Added some missing reversion2 localizations and latests reversion1


bluegr:
Thanks! Merging

PR #1967, opened 2019-12-03
GUI: Don't show midi setting tabs if the game has no music at all.

henke37:
Games using midi always uses it for music. A game without music will not use midi. As such, don't show the midi tabs at all for that game.


bluegr:
Quite straightforward. Thanks! Merging

PR #1972, opened 2019-12-08
QUEEN: fix bellboy dialogue - bug #11272

BLooperZ:
Quick fix for https://bugs.scummvm.org/ticket/11272.
This undoes https://github.com/scummvm/scummvm/pull/1779
So the behaviour of left click during player's talk now once again does not match DOS executable.


bluegr:
Thanks for fixing the left click regression

PR #1977, opened 2019-12-15
LURE: Enable the optional use of TTS to read descriptions as a narrator in "Lure of the Temptress"

neuromancer:
LURE: Enable the optional use of TTS to read descriptions as a narrator in Lure of the Temptress

Added an option to use TTS to read descriptions in Lure of the Temptress.
Following the discussion in #1953, this option will only appear if the user
compiles with TTS support. It will read scene and object descriptions but
omit any character dialog.

### DIFF
--- a/engines/lure/detection.cpp
+++ b/engines/lure/detection.cpp
@@ -25,7 +25,8 @@
 #include "engines/advancedDetector.h"
 #include "engines/engine.h"
 #include "common/savefile.h"
-
+#include "common/system.h"
+#include "common/translation.h"
 #include "lure/lure.h"
 
 namespace Lure {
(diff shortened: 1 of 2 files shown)


engines/lure/detection.cpp
```diff
@@ -73,7 +91,11 @@ static const LureGameDescription gameDescriptions[] = {
 			Common::EN_ANY,
 			Common::kPlatformDOS,
 			ADGF_NO_FLAGS,
+                        #ifdef USE_TTS
```
mgerhardy: The formatting is off in a lot of places - mixed spaces and tabs.

neuromancer: It should be fixed. Thanks!

dreammaster:
The commit text also isn't conforming to naming requirements.. it should have a "LURE: " in front of them. At this point, it would be a good idea for you to squash your branch into a single commit, and have it have the LURE: prefix

neuromancer:
I'm ok with squashing this PR as a single commit (same for #1953 )

engines/lure/surface.cpp
```diff
@@ -467,13 +472,31 @@ Surface *Surface::newDialog(uint16 width, uint8 numLines, const char **lines, bo
 
 	Surface *s = new Surface(width, size.y);
 	s->createDialog();
+	#ifdef USE_TTS	
+	char *text = (char*) malloc(1024*sizeof(char));
+	memset(text, 0, 1024);
```
bluegr: Instead of using a char * here, you could use Common::String. The string manipulation code later on will be much easier and straightforward, and you won't need to have a hard limit of characters

neuromancer: Done!

neuromancer:
It seems that one of the Travis CI tests from the latest commit failed [fetching some packages](https://travis-ci.org/scummvm/scummvm/jobs/627702386#L736-L804). Can someone rerun it?

bluegr:
Thanks for your work, and for fixing the remaining issues. I have requested a fresh Travis build and it worked fine.

Merging

PR #1978, opened 2019-12-15
WINTERMUTE: Fix backspace handling on text input

lolbot-iichan:
WINTERMUTE: Refactor UIEdit::handleKeypress()

This fixes https://bugs.scummvm.org/ticket/11033

Removed checking printable property for system keys.
Refactored whole function to make it a bit more readable.

Reference implementations:
https://github.com/retrowork/wmelite/blob/master/src/UIEdit.cpp#L724
https://github.com/retrowork/Wintermute-Engine/blob/master/src/engine_core/wme_base/UIEdit.cpp#L708

WINTERMUTE: Handle isCurrentPrintable property for WME 1.x and WME Lite

Moved KEYCODE_SPACE to "alphanumeric or punctuation" group.
Splited "else" into before and after WME_LITE.

### DIFF
--- a/engines/wintermute/base/base_keyboard_state.cpp
+++ b/engines/wintermute/base/base_keyboard_state.cpp
@@ -26,6 +26,7 @@
  * Copyright (c) 2011 Jan Nedoma
  */
 
+#include "engines/wintermute/base/base_engine.h"
 #include "engines/wintermute/base/base_keyboard_state.h"
 #include "engines/wintermute/base/scriptables/script_value.h"
 #include "engines/wintermute/base/scriptables/script_stack.h"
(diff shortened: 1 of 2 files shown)


lolbot-iichan:
Codacy does not like `strlen(_text)` usage in refactored code, however, there are dozen of places where `strlen(_text)` is used in unmodified code - _text is assumed to always be a NULL-terminated string everywhere... 

Can we ignore it as false positive or should I refactor some files to remove `strlen(_text)` completely?

engines/wintermute/base/base_keyboard_state.cpp
```diff
@@ -246,18 +247,22 @@ bool BaseKeyboardState::readKey(Common::Event *event) {
...
 		_currentPrintable = code == Common::KEYCODE_BACKSPACE || 
-		                    code == Common::KEYCODE_TAB || 
-		                    code == Common::KEYCODE_RETURN || 
-		                    code == Common::KEYCODE_KP_ENTER || 
-		                    code == Common::KEYCODE_ESCAPE || 
-		                    code == Common::KEYCODE_SPACE;
+							code == Common::KEYCODE_TAB || 
+							code == Common::KEYCODE_RETURN || 
+							code == Common::KEYCODE_KP_ENTER || 
+							code == Common::KEYCODE_ESCAPE;
```
bluegr: It's usually a bad idea to align text using tabs, since tab spacing is a user configured setting in each editor. Nevertheless, since tabs are used for spacing throughout this file, this should be OK

bluegr:
Thanks for your work! Merging

PR #1982, opened 2019-12-19
BLADERUNNER: replace c_long with explicit c_int32

BLooperZ:
Fix subtitles creation script by converting to int32 explicitly,
(long is tranlated to int64 on some systems, which caused issues with sorting signed numbers).

Many thanks to @antoniou79 for his help solving this issue.


antoniou79:
Nice catch, thank you for this fix!

PR #1983, opened 2019-12-21
MADS: Enable the optional use of TTS to read descriptions as a narrator in Rex Nebular

neuromancer:
MADS: Enable the optional use of TTS to read descriptions as a narrator in Rex Nebular

Added an option to use TTS to read descriptions in Rex Nebular
Following the discussion in #1953, this option will only appear if the user
compiles with TTS support. It will read scene and object descriptions but
omit any character dialog.


bluegr:
Nice! Merging

PR #1984, opened 2019-12-22
DIRECTOR: Fix calling of Beep

rvanlaar:
- Convert beep called with zero arguments to beep(1)
- Beep argument starts counting at 1.

### DIFF
--- a/engines/director/lingo/lingo-funcs.cpp
+++ b/engines/director/lingo/lingo-funcs.cpp
@@ -355,7 +355,7 @@ void Lingo::func_cursor(int c) {
 }
 
 void Lingo::func_beep(int repeats) {
-	for (int r = 0; r <= repeats; r++)
+	for (int r = 1; r <= repeats; r++)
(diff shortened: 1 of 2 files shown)


engines/director/lingo/lingo-funcs.cpp
```diff
@@ -355,7 +355,7 @@ void Lingo::func_cursor(int c) {
 }
 
 void Lingo::func_beep(int repeats) {
-	for (int r = 0; r <= repeats; r++)
+	for (int r = 1; r <= repeats; r++)
```
sev-: I would rather change it back to `0` but change the `<=` condition to `<`

rvanlaar: I would prefer to keep it at `1` and `<=` , so the delay can be `if (r < repeats) delay()` instead of `if (r  < repeats - 1)`

sev-:
Thanks!

PR #1985, opened 2019-12-22
GRAPHICS: When compiling for 3DS, use RGBA8 ColorMask instead of ARGB8

BallM4788:
Because the 3DS hardware supports RGBA8 textures but not ARGB8 textures, adjust colormasks.h accordingly at compile time.

Source: [libctru documentation](https://smealum.github.io/ctrulib/enums_8h.html#a63bb2f7fceb0f356549f0be235a0f99f).

### DIFF
--- a/graphics/colormasks.h
+++ b/graphics/colormasks.h
@@ -276,10 +276,17 @@ struct ColorMasks<8888> {
 		kGreenBits  = 8,
 		kBlueBits   = 8,
 
+#ifdef __3DS__ // 3DS uses RGBA8 rather than ARGB8
+		kAlphaShift = 0,
+		kRedShift   = kGreenBits+kBlueBits+kAlphaBits,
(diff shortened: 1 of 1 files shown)


bgK:
What does this change achieves?

Having ColorMasks behave differently for different platforms is dangerous. Say you save a game screenshot thumbnail to a file using a ColorMask on your 3DS and then try to load the save on your PC. The colors will be incorrect if the ColorMasks don't match.

bluegr:
To be fair, there are similar changes in this file for other platforms.

bgK:
> To be fair, there are similar changes in this file for other platforms.

I wish platform specific code in shared files was seen as an invitation for clean-ups rather than an invitation to add more :)

sev-:
@BallM4788 any insight on the reason and effect on the saves?

bgK:
I'm closing this as the benefit is unclear.

PR #1986, opened 2019-12-23
WINTERMUTE: Fix fallback detection

bgK:
Now the singleId property has been removed, the gameIds returned by a
MetaEngine when detecting games must be accepted by the `findGame`
method. Wintermute was generating gameIds based on datafiles in the
fallback detector. With this change, the gameId for fallback detected
games is always `wintermute`. The target name is generated based on the
game caption.

The `(unknown)` mention that was previously added to the game title is gone.
The unknown game dialog now makes it very clear that the game version
is not known to ScummVM.

Fixes https://bugs.scummvm.org/ticket/11288.


bluegr:
Nice! Thanks for your work. Merging

PR #1988, opened 2019-12-28
GUI: Dynamic layouts

bgK:
Prior to this PR, a GUI layout was only affected by the screen size.
Now, a layout can additionally be influenced by the GUI dialog and widgets
that use it. This capability is leveraged to implement the following
features:

* Layout elements that are not bound to a GUI widget do not take space.
   This means that dialogs where the widgets shown depend on for example
   a feature being enabled at configure time no longer have blank spaces.
* Widgets can define a minimal required size for their contents not to be
   cut. For now this is only used for buttons so their width is always
   sufficient for their caption not to be cut. This mechanism could be
   applied to other widget types in the future.

Before 1:
scummvm00015

After 1:
scummvm00016

Before 2:
scummvm00014

After 2:
scummvm00017

These changes also fix the remaining issue in PR #1942 which I've now rebased on top of this one.


bluegr:
Very nice!
Since the new feature about the minimal size applies only to buttons for now, it would be preferable if a warning was shown, when somebody tries to use this feature with other widgets

bgK:
> Since the new feature about the minimal size applies only to buttons for now, it would be preferable if a warning was shown, when somebody tries to use this feature with other widgets

I'm not sure what you mean. The feature is implicitly enabled for all widget types that implement `Widget::getMinSize`. There is no way to use it for other widget types.

bluegr:
@bgK: great, thanks for clarifying this.

Looks good to me overall, +1 from me

bgK:
> How does it behave when you switch between games with different number of GUIO? Does it still reflow the layout?

Yes, the layout is refreshed each time the dialog is opened.

> Also, the change is big, please bump the GUI theme version.

I will merge https://github.com/scummvm/scummvm/pull/1942 at the same time as this along with a version bump.

sev-:
Cool, please do.

bgK:
Merged manually.

Tkachov:
Did anyone test how it works on Cloud tab (maybe LAN as well)? (I'm away from my desktop, so I won't be able to do it a couple more days.)

I'm a little bit worried about that tab, because it has ScrollContainer and hides and shows widgets in it depending on storage state. Also, lowres had some truncated labels, maybe "cloud.scummvm.org" button was also truncated.

bgK:
> Did anyone test how it works on Cloud tab (maybe LAN as well)?

I went trough all the dialogs that have a layout. The cloud tab does not seem to have major issues with the change.

PR #1989, opened 2019-12-30
SCUMM: support hebrew fan-translations (text are written right-to-left)

BLooperZ:
This adds support for writing texts right to left for hebrew fan translations
(by Hebrew Adventure Group).

Translations are complete for most games.
But I prefer to wait and add the detection entries after final versions are released
(after proofreading and playtesting).

Thanks.

Games already translated:
The Secret of Monkey Island
Indiana Jones and the Fate of Atlantis
Day of the Tentacle
Sam & Max Hit the Road
Full Throttle
The Dig
The Curse of Monkey Island

(Zak McKracken and the Alien Mindbenders is also  translated but is not affected by this changes)


tsoliman:
Some commit messages don't conform and some commits are "junk" that shouldn't be kept (e.g. the git push test).
You'll want to rebase/cleanup those before a merge happens.
Thanks.

(Note: this isn't a code review and I'm not a SCUMM engine dev)

BLooperZ:
@tsoliman Thank you, fixed.

engines/scumm/string.cpp
```diff
@@ -175,7 +179,7 @@ void ScummEngine_v6::drawBlastTexts() {
 				// Some localizations may override colors
 				// See credits in Chinese COMI
 				if (_game.id == GID_CMI &&	_language == Common::ZH_TWN &&
-				      c == '^' && (buf == _blastTextQueue[i].text + 1)) {
+					  c == '^' && (buf == _blastTextQueue[i].text + 1)) {
```
sev-: Unnecessary change.

engines/scumm/string.cpp
```diff
@@ -454,6 +461,88 @@ bool ScummEngine::newLine() {
 	return true;
 }
 
+void ScummEngine::fakeBidiString(char *ltext, bool ignoreVerb) {
+	int ll = 0;
```
sev-: It would be useful if you describe the purpose of this method, and perhaps even what is going on, in the comments.

engines/scumm/string.cpp
```diff
@@ -454,6 +461,88 @@ bool ScummEngine::newLine() {
...
+	if (_game.id == GID_INDY4 && ltext[ll] == 127) {
+		ll++;
+	}
+	while (ltext[ll] == -1) {
+		ll += 4;
+	}
+	int ipos = 0;
+	int start = 0;
+	char* text = ltext + ll;
```
sev-: should be `char *`, the same thing below.

engines/scumm/string.cpp
```diff
@@ -454,6 +461,88 @@ bool ScummEngine::newLine() {
...
+	char* current = text;
+	while(1) {
```
sev-: `while (1)`

engines/scumm/string.cpp
```diff
@@ -454,6 +461,88 @@ bool ScummEngine::newLine() {
...
+			// ignore the line break for verbs texts
+			if (ignoreVerb && (*(current + 1) ==  8)) {
+				*(current + 1) = *current;
+				*current = 8;
+				ipos += 2;
+				current += 2;
+				continue;
+			}
+
+			char buff[384];
```
sev-: These are quite big for the stack. I recommend to `malloc` these two buffers  at the method start and free on the exit.

engines/scumm/string.cpp
```diff
@@ -542,7 +631,7 @@ void ScummEngine::CHARSET_1() {
 		return;
 
 	if ((_game.version <= 6 && _haveMsg == 1) ||
-	    (_game.version == 7 && _haveMsg != 1)) {
+		(_game.version == 7 && _haveMsg != 1)) {
```
sev-: Unnecessary change.

engines/scumm/string.cpp
```diff
@@ -929,6 +1031,30 @@ void ScummEngine::drawString(int a, const byte *msg) {
...
 		_charset->_left -= _charset->getStringWidth(a, buf) / 2;
+	} else if (_game.version >= 4 && _game.version < 7 && _game.id != GID_SAMNMAX && _language == Common::HE_ISR) {
+		if (_game.id != GID_INDY4 || buf[0] == 127) {
+			int ll = 0;
+			if (_game.id == GID_INDY4 && buf[0] == 127) {
+				buf[0] = 32;
+				ll++;
+			}
+
+			byte* ltext = buf;
```
sev-: `byte *ltext`

engines/scumm/string.cpp
```diff
@@ -969,6 +1095,24 @@ void ScummEngine::drawString(int a, const byte *msg) {
 			case 8:
 				if (_charset->_center) {
 					_charset->_left = _charset->_startLeft - _charset->getStringWidth(a, buf + i);
+				} else if (_game.version >= 4 && _game.version < 7 && _language == Common::HE_ISR) {
+					int ll = 0;
+					byte* ltext = buf + i;
```
sev-: `byte *ltext`

engines/scumm/verbs.cpp
```diff
@@ -1010,6 +1011,9 @@ void ScummEngine_v7::drawVerb(int verb, int mode) {
 		_charset->setCurID(vs->charset_nr);
 
 		// Compute the text rect
+		if (_language == Common::HE_ISR /* && !(_game.id == GID_FT && (_charset->getCurID() == 6 || _charset->getCurID() == 7))*/) {
```
sev-: Is the commented out part going to be used in the final version?

engines/scumm/string.cpp
```diff
@@ -454,6 +461,88 @@ bool ScummEngine::newLine() {
...
+		}
+		if (*current) {
+			ipos++;
+			current++;
+			continue;
+		}
+		break;
+	}
+	if (!ignoreVerb && _game.id == GID_INDY4 && ltext[0] == 127) {
+		ltext[start + ipos + ll] = (char) 128;
```
sev-: `(char)128`, or I would rather write `'\x80'`

engines/scumm/verbs.cpp
BLooperZ: After checking again, I think it is safe to remove the commented out code (the final version will work without it)

BLooperZ:
Probably not very soon, I'm afraid.

The fake BiDi mechanism used here is custom made for SCUMM needs. it assumes all text should be written right-to-left while preserving left-to-right ordering for several formats of numbers (needed for variables) + some escaping of control codes.
(other game engines are mostly ok with just reversing the text of each line).
(The use of simple processing actually helps the translation, as it makes the ordering predictable).

The launcher GUI probably needs real BiDi processing, as it can have RTL texts (translated texts) alongside LTR text (such as paths, shaders and version info).

sev-:
Thanks, merging.

BLooperZ:
It seems the fake BiDi function does not work on system in which char is unsigned by default.
I'll soon update the code to use byte instead.

PR #1993, opened 2020-01-01
SCUMM: use reliable types in fake bidi

BLooperZ:
Fix fake BiDi (from #1989 ) issues with systems in which char is unsigned by default (e.g. Android).
By using types explicitly defined by ScummVM.


bluegr:
Thanks! Merging

PR #1996, opened 2020-01-11
Fix final boss is sluggish and crash issue

harshbawari:
Fix crash issue when Griffon.s03 map is loaded.


raziel-:
@harshbawari

Could you add the engine the fix is for to your commit message?

bluegr:
I've squashed and fixed the commit message. @harshbawari: please use the engine as part of your commits in the future, as @raziel- said

sev-:
@bluegr I wish I had a chance to review these changes.

PR #2001, opened 2020-01-19
AGI: Add pirated version of GoldRush! to detection table

ZvikaZ:
There are 2 bug reports regarding one version of GoldRush that
wasn't recognized by the detection table.

On bug #3320, it was explained to be a pirated version. Therefore,
this update, in order to avoid further reporting of the same issue.

It closes bug #11308.

### DIFF
--- a/engines/agi/detection_tables.h
+++ b/engines/agi/detection_tables.h
@@ -49,6 +49,22 @@ namespace Agi {
 		ver \
 	}
 
+#define GAME_LVFPN_PIRATED(id,extra,fname,md5,size,lang,ver,features,gid,platform,interp,guioptions) { \
+		{ \
+			id, \
+			extra, \
+			AD_ENTRY1s(fname,md5,size), \
+			lang, \
+			platform, \
+			ADGF_PIRATED, \
+			guioptions \
+		}, \
+		gid, \
+		interp, \
+		features, \
+		ver \
+	}
+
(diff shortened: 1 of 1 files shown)


engines/agi/detection_tables.h
```diff
@@ -256,6 +273,9 @@ static const AGIGameDescription gameDescriptions[] = {
 	// Gold Rush! (PC 3.5") 2.01 12/22/88 [AGI 3.002.149]
 	GAME3("goldrush", "2.01 1988-12-22 3.5\"", "grdir", "6a285235745f69b4b421403659497216", 0x3149, GID_GOLDRUSH),
 
+    // Gold Rush!  2.01 12/22/88 - pirated copy, according to https://bugs.scummvm.org/ticket/3220
```
mgerhardy: This is a whitespace error. Tabs should be used.

lotharsm:
I wonder if adding pirated versions to the detection tables is even allowed...

criezy:
@lotharsm Yes, we have the `ADGF_PIRATED` flag specially for that.
Adding this seems like a good idea to me.

lotharsm:
> 
> 
> @lotharsm Yes, we have the `ADGF_PIRATED` flag specially for that.
> Adding this seems like a good idea to me.

Alright, thanks a lot for the feedback! Looks good to me too then.

bluegr:
Thanks for your work! Merging

PR #2006, opened 2020-01-21
SDL: Use the name of the running target as a base for screenshot files

bgK:
Makes it a bit easier to recognize files in the screenshot folder, IMO.

### DIFF
--- a/backends/graphics/sdl/sdl-graphics.cpp
+++ b/backends/graphics/sdl/sdl-graphics.cpp
@@ -279,12 +279,18 @@ void SdlGraphicsManager::saveScreenshot() {
 	if (sdl_g_system)
 		screenshotsPath = sdl_g_system->getScreenshotsPath();
 
-	for (int n = 0;; n++) {
+	// Use the name of the running target as a base for screenshot file names
+	Common::String currentTarget = ConfMan.getActiveDomainName();
+
 #ifdef USE_PNG
-		filename = Common::String::format("scummvm%05d.png", n);
+	const char *extension = "png";
 #else
-		filename = Common::String::format("scummvm%05d.bmp", n);
+	const char *extension = "bmp";
 #endif
+
+	for (int n = 0;; n++) {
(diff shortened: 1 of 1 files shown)


tsoliman:
How about always prefixing `scummvm` and then adding the target as a suffix if the target name is available?

So for example `scummvm-sq4-001.png`

bgK:
> How about always prefixing scummvm and then adding the target as a suffix if the target name is available?

I don't mind either way. I'll change it if that's what people prefer. My reasoning for not having the `scummvm` prefix is that on Windows and Linux the screenshots are saved by default in a folder named `ScummVM Screenshots`. However it's true they are saved on the desktop on macOS and a `scummvm` prefix is kind of nice there. (Is saving screenshots on the desktop an accepted practice on macOS?)

tsoliman:
Thanks!

> Is saving screenshots on the desktop an accepted practice on macOS?

Yes.
I am not an expert. That's the behavior I've observed in the built-in OS screenshot function.

criezy:
> 
> 
> Thanks!
> 
> > Is saving screenshots on the desktop an accepted practice on macOS?
> 
> Yes.
> I am not an expert. That's the behavior I've observed in the built-in OS screenshot function.

Indeed, the macOS shortcuts to take screenshots (e.g. Crtl+Shift+3) saves them on the Desktop as well. See https://support.apple.com/en-us/HT201361 for reference.

For those I think prefixing with `scummvm-` and not just using the target would be better.

bgK:
> For those I think prefixing with scummvm- and not just using the target would be better.

Ok, changed it.

bluegr:
Nice work!
I do agree on the scummvm prefix, since we got that behavior on OSX.

@bgK: You mentioned that you changed the behavior, as requested, but there's no relevant commit? Perhaps you forgot to push your changes?

bgK:
> Perhaps you forgot to push your changes?

Right.

bluegr:
Nice work! Merging

PR #2009, opened 2020-01-22
KEYMAPPER: Improvements

bgK:
This PR contains various changes I've made a few years ago to the keymapper. These changes are mostly about code simplification, but also contain a few new features:
* The remap dialog uses a scroll container
* The mapped actions can have default key bindings defined in the engines
* Default key bindings are shown in the remap dialog
* Actions can have multiple key mappings

Capture d’écran de 2020-01-22 13-36-17


bluegr:
First of all, great work! A lot has been changed, but the changes are straightforward.

Could you please fix the merge conflicts?

PR #2011, opened 2020-01-22
IPHONE: Disable idle timer while an engine is running

criezy:
This prevents the iOS device from going to sleep when an engine is running, for example while the intro of a game is playing (this happened to me with DOTT for example, and both Scorp and @waltervn mentioned this today in the Discord server).

This changes here are a port of commits 4a94464 and 58f3aac from the iOS7 backend to the old iPhone backend. It replaces the use of `dispatch_async`, that is only present since SDK 4.0, with `performSelectorOnMainThread` that exists since SDK 2.0. This means all the APIs used by the commit here should work with SDK 2.0 or above.

I am however unable to test that it compiles or works, so I am submitting this here for review.

### DIFF
--- a/backends/platform/iphone/iphone_video.h
+++ b/backends/platform/iphone/iphone_video.h
@@ -101,6 +101,9 @@
 
 - (void)deviceOrientationChanged:(UIDeviceOrientation)orientation;
 
+- (void)disableIdleTimer;
+- (void)enableIdleTimer;
+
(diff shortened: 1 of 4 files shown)


waltervn:
Compiles and works on my iPad 2 with iOS 6.

criezy:
Thank you for testing. I will merge this now and monitor buildbot to check that it does not break the build there.

PR #2012, opened 2020-01-23
KEYMAPPER: Move the remap dialog to an option dialog tab

bgK:
The aim is to make it easy to discover, and possible to use without a keyboard.

Global ScummVM keymaps are shown in the `Options` dialog. Per-game keymaps are shown in the `Edit Game` dialog.

This PR builds on top of #2009 and thus includes its commits.

Mandatory screenshots:
Capture d’écran de 2020-01-23 10-53-39

Capture d’écran de 2020-01-23 10-54-06


bluegr:
Great work! Quite nice changes. Merging

PR #2014, opened 2020-01-23
DIRECTOR: LINGO: Add numberofwords implementation

rvanlaar:
We check of the current char in a string is a space,
if so check if the previous one wasn't a space, then count it as a word.
And count the last word when the text doesn't end in a space.
This is to prevent off by one errors.
    
Director 4 counts everything non-space (i.e. \n, \t, etc) as a
word separator. A word can be a single '.'.

### DIFF
--- a/engines/director/lingo/lingo-builtins.cpp
+++ b/engines/director/lingo/lingo-builtins.cpp
@@ -1807,7 +1807,7 @@ void LB::b_numberoflines(int nargs) {
 
 	d.toString();
 	int numberoflines = 1;
-	Common::String contents = d.u.s->c_str();
+	Common::String contents = *d.u.s;
 	for (uint32 i = 0; i < d.u.s->size(); i++) {
 		if (contents[i] == '\n')
 			numberoflines++;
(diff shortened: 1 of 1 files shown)


engines/director/lingo/lingo-builtins.cpp
```diff
@@ -1823,9 +1823,19 @@ void LB::b_numberoflines(int nargs) {
 void LB::b_numberofwords(int nargs) {
 	Datum d = g_lingo->pop();
 
-	warning("STUB: b_numberofwords");
-	d.toInt();
-	d.u.i = 0;
+	d.toString();
+	int numberofwords = 0;
+	Common::String contents = *d.u.s;
+	for (uint32 i=1; i < d.u.s->size(); i++) {
```
sev-: Formatting is incorrect `i = 1`

engines/director/lingo/lingo-builtins.cpp
```diff
@@ -1823,9 +1823,19 @@ void LB::b_numberoflines(int nargs) {
...
+		if (Common::isSpace(contents[i]) and not Common::isSpace(contents[i - 1]))
```
sev-: I am not sure that `not` keyword is compatible enough. Please use `!`

bluegr: Same goes for the `and` keyword, please use &&

engines/director/lingo/lingo-builtins.cpp
rvanlaar: Is there a formatter that can be automatically used for scummvm?

sev-: Yes, mentioned right on the bottom of the Code Formatting Guidelines: https://wiki.scummvm.org/index.php?title=Code_Formatting_Conventions

I highly suggest exploring our Developer Central.

PR #2020, opened 2020-01-26
WINTERMUTE: Update known prefixes of absolute paths

lolbot-iichan:
1. Fix knownPrefix for Tanya Grotter 2. Octal bytes are now used instead of incorrect pseudo-hex writing.
2. Add 5 more knownPrefixes for file seeking
3. Add known absolute EDITOR_BG_FILE paths and prefixes. EDITOR_BG_FILE path is currently ignored, so this is useless now, however it may come handy in some deep future, if something like a sprite viewer would be implemented.


bluegr:
Great work, thanks! Merging

PR #2021, opened 2020-01-26
3DS: Implement dynamic graphics modes to improve performance

BallM4788:
Because the 3DS is a low-power system, conversion from a game's pixel format to a dissimilar hardware pixel format can be taxing to the system, causing graphical slowdown. This is most evident with graphically intensive games such as Riven, which on the 3DS runs roughly half as fast as it should. Changing the pixel format of display texture surfaces when launching a game remedies this.

When launching a game, switch the graphics mode if necessary (and by extension the pixel formats used for Graphics::Surfaces and Sprites) to the one that most closely matches the pixel format used in-game.

Additional Fixes:
- Fix to prevent cursor position from changing when exiting a menu.
- Fix to prevent updating of Magnification viewport position when virtual keyboard is open.
- Cosmetic code fixes for improper whitespace and missing curly brackets.
- Remove RGB8 as a mode option because:
  - It was already commented out in the master 3DS backend.
  - There are currently no games that explicitly require it.

Notes:
- As these graphics modes are automatically implemented on a per-game basis, they are meant for backend use only and are purposefully not accessible through the Options menu.
- RGBA8 (aka RGBA8888) remains the default pixel format, being used for the launcher menu, CLUT8, and for games which do not specify a particular format.

### DIFF
--- a/backends/platform/3ds/osystem-graphics.cpp
+++ b/backends/platform/3ds/osystem-graphics.cpp
@@ -23,24 +23,44 @@
 
 #include "backends/platform/3ds/osystem.h"
 #include "backends/platform/3ds/shader_shbin.h"
+#include "backends/platform/3ds/options-dialog.h"
+#include "backends/platform/3ds/config.h"
 #include "common/rect.h"
 #include "graphics/fontman.h"
 #include "gui/gui-manager.h"
-#include "options-dialog.h"
-#include "config.h"
 
 // Used to transfer the final rendered display to the framebuffer
-#define DISPLAY_TRANSFER_FLAGS                                                 \
-	(GX_TRANSFER_FLIP_VERT(0) | GX_TRANSFER_OUT_TILED(0) |                     \
-	 GX_TRANSFER_RAW_COPY(0) | GX_TRANSFER_IN_FORMAT(GX_TRANSFER_FMT_RGBA8) |  \
-	 GX_TRANSFER_OUT_FORMAT(GX_TRANSFER_FMT_RGB8) |                            \
-	 GX_TRANSFER_SCALING(GX_TRANSFER_SCALE_NO))
+#define DISPLAY_TRANSFER_FLAGS                                                    \
+        (GX_TRANSFER_FLIP_VERT(0) | GX_TRANSFER_OUT_TILED(0) |                    \
+         GX_TRANSFER_RAW_COPY(0) | GX_TRANSFER_IN_FORMAT(GX_TRANSFER_FMT_RGBA8) | \
+         GX_TRANSFER_OUT_FORMAT(GX_TRANSFER_FMT_RGB8) |                           \
+         GX_TRANSFER_SCALING(GX_TRANSFER_SCALE_NO))
+#define TEXTURE_TRANSFER_FLAGS(fmt)                             \
+        (GX_TRANSFER_FLIP_VERT(1) | GX_TRANSFER_OUT_TILED(1) |  \
+         GX_TRANSFER_RAW_COPY(0) | GX_TRANSFER_IN_FORMAT(fmt) | \
+         GX_TRANSFER_OUT_FORMAT(fmt) | GX_TRANSFER_SCALING(GX_TRANSFER_SCALE_NO))
+#define DEFAULT_MODE _modeRGBA8
 
 namespace _3DS {
+/* Group the various enums, values, etc. needed for
+ * each graphics mode into instaces of GfxMode3DS */
+static const GfxMode3DS _modeRGBA8 = { Graphics::PixelFormat(4, 8, 8, 8, 8, 24, 16, 8, 0),
+                                       GPU_RGBA8, TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGBA8) };
+static const GfxMode3DS _modeRGB565 = { Graphics::PixelFormat(2, 5, 6, 5, 0, 11, 5, 0, 0),
+                                        GPU_RGB565, TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB565) };
+static const GfxMode3DS _modeRGB555 = { Graphics::PixelFormat(2, 5, 5, 5, 0, 11, 6, 1, 0),
+                                        GPU_RGBA5551, TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB5A1) };
+static const GfxMode3DS _modeRGB5A1 = { Graphics::PixelFormat(2, 5, 5, 5, 1, 11, 6, 1, 0),
+                                        GPU_RGBA5551, TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB5A1) };
+static const GfxMode3DS _modeCLUT8 = _modeRGBA8;
 
-void OSystem_3DS::initGraphics() {
+static const GfxMode3DS *gfxModes[] = { &_modeRGBA8, &_modeRGB565, &_modeRGB555, &_modeRGB5A1, &_modeCLUT8 };
+
+
+void OSystem_3DS::init3DSGraphics() {
+	_gfxState.gfxMode = gfxModes[CLUT8];
 	_pfGame = Graphics::PixelFormat::createFormatCLUT8();
-	_pfGameTexture = Graphics::PixelFormat(4, 8, 8, 8, 8, 24, 16, 8, 0);
+	_pfDefaultTexture = Graphics::PixelFormat(4, 8, 8, 8, 8, 24, 16, 8, 0);
 
 	C3D_Init(C3D_DEFAULT_CMDBUF_SIZE);
 
@@ -83,10 +103,12 @@ void OSystem_3DS::initGraphics() {
 	C3D_CullFace(GPU_CULL_NONE);
 }
 
-void OSystem_3DS::destroyGraphics() {
+void OSystem_3DS::destroy3DSGraphics() {
 	_gameScreen.free();
 	_gameTopTexture.free();
 	_gameBottomTexture.free();
+	_cursor.free();
+	_cursorTexture.free();
 	_overlay.free();
 	_activityIcon.free();
 
@@ -124,8 +146,37 @@ bool OSystem_3DS::getFeatureState(OSystem::Feature f) {
 	}
 }
 
+GraphicsModeID OSystem_3DS::chooseMode(Graphics::PixelFormat *format) {
+	if (format->bytesPerPixel > 2) {
+		return RGBA8;
+	} else if (format->bytesPerPixel > 1) {
+		if (format->gBits() > 5) {
+			return RGB565;
+		} else if (format->aBits() == 0) {
+			return RGB555;
+		} else {
+			return RGB5A1;
+		}
+	}
+	return CLUT8;
+}
+
+bool OSystem_3DS::setGraphicsMode(GraphicsModeID modeID) {
+	switch (modeID) {
+	case RGBA8:
+	case RGB565:
+	case RGB555:
+	case RGB5A1:
+	case CLUT8:
+		_gfxState.gfxMode = gfxModes[modeID];
+		return true;
+	default:
+		return false;
+	}
+}
+
 void OSystem_3DS::initSize(uint width, uint height,
-                                   const Graphics::PixelFormat *format) {
+                           const Graphics::PixelFormat *format) {
 	debug("3ds initsize w:%d h:%d", width, height);
 	int oldScreen = config.screen;
 	loadConfig();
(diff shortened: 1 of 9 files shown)


backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -30,32 +30,72 @@
...
+         GX_TRANSFER_RAW_COPY(0) | GX_TRANSFER_IN_FORMAT(fmt) | \
+         GX_TRANSFER_OUT_FORMAT(fmt) |                          \
+         GX_TRANSFER_SCALING(GX_TRANSFER_SCALE_NO))
 
 namespace _3DS {
 
-void OSystem_3DS::initGraphics() {
+/* Group the various enums, values, etc. needed for
+ * each graphics mode into instaces of GfxMode3DS */
+GfxMode3DS _modeRGBA8 = { Graphics::PixelFormat(4, 8, 8, 8, 8, 24, 16, 8, 0),
```
bgK: These variable can probably be  `static const`.

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -30,32 +30,72 @@
...
+                           DISPLAY_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB5A1, GX_TRANSFER_FMT_RGB5A1),
+                           TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB5A1) };
+GfxMode3DS _modeRGB5A1 = { Graphics::PixelFormat(2, 5, 5, 5, 1, 11, 6, 1, 0),
+                           Graphics::PixelFormat(2, 5, 5, 5, 1, 11, 6, 1, 0),
+                           GSP_RGB5_A1_OES, GPU_RB_RGBA5551, GPU_RGBA5551,
+                           DISPLAY_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB5A1, GX_TRANSFER_FMT_RGB5A1),
+                           TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB5A1) };
+GfxMode3DS _modeCLUT8 = _modeRGBA8;
+
+GfxMode3DS *gfxModes[6] = { &_modeRGBA8, &_modeRGB565, &_modeRGB8,
```
bgK: There is no need to set the size of the array here. The array can be `static const`.

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -30,32 +30,72 @@
...
+
+
+// Initiate 3DS graphics when starting up ScummVM
+void OSystem_3DS::init3DSGraphics() {
+	_gfxState.gfxMode = gfxModes[CLUT8];
 	_pfGame = Graphics::PixelFormat::createFormatCLUT8();
-	_pfGameTexture = Graphics::PixelFormat(4, 8, 8, 8, 8, 24, 16, 8, 0);
-
 	C3D_Init(C3D_DEFAULT_CMDBUF_SIZE);
+	assert(loadGfx() == true);
```
bgK: Please don't call functions in asserts. If assertions are disabled, the code is not executed.

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -124,27 +172,77 @@ bool OSystem_3DS::getFeatureState(OSystem::Feature f) {
...
-void OSystem_3DS::initSize(uint width, uint height,
-                                   const Graphics::PixelFormat *format) {
-	debug("3ds initsize w:%d h:%d", width, height);
-	int oldScreen = config.screen;
-	loadConfig();
-	if (config.screen != oldScreen) {
-		_screenChangeId++;
+GraphicsModeID OSystem_3DS::chooseMode(Graphics::PixelFormat *format) {
+	if (format->bytesPerPixel > 2) {
+//		if (format->aBits() > 0) {
```
bgK: Please remove commented out code.

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -124,27 +172,77 @@ bool OSystem_3DS::getFeatureState(OSystem::Feature f) {
...
 	_gameHeight = height;
-	_gameTopTexture.create(width, height, _pfGameTexture);
-	_overlay.create(400, 320, _pfGameTexture);
 	_magCenterX = _magWidth / 2;
 	_magCenterY = _magHeight / 2;
 
-	if (format) {
+	_oldPfGame = _pfGame;
+	if (!format) {
+		if (_pfGame != Graphics::PixelFormat::createFormatCLUT8()) {
```
bgK: `_pfGame` can be set to CLUT8 unconditionally here.

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -124,27 +172,77 @@ bool OSystem_3DS::getFeatureState(OSystem::Feature f) {
...
+	}
+
+	int oldScreen = config.screen;
+	loadConfig();
+	if (config.screen != oldScreen) {
+		_screenChangeId++;
+	}
+
+	_gameTopTexture.create(width, height, _gfxState.gfxMode);
+	_overlay.create(400, 320, _gfxState.gfxMode);
```
bgK: I'm not sure I understand, why use the same pixel format for the game and the overlay? Wouldn't it be simpler and less problematic to always use RGBA for the overlay (that way the overlay always has alpha, removing the need for the _screenGrab surface)?

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -239,12 +388,12 @@ void OSystem_3DS::grabPalette(byte *colors, uint start, uint num) const {
...
 
 void OSystem_3DS::copyRectToScreen(const void *buf, int pitch, int x,
-                                           int y, int w, int h) {
+                                   int y, int w, int h) {
 	Common::Rect rect(x, y, x+w, y+h);
 	_gameScreen.copyRectToSurface(buf, pitch, x, y, w, h);
 	Graphics::Surface subSurface = _gameScreen.getSubArea(rect);
 
-	Graphics::Surface *convertedSubSurface = subSurface.convertTo(_pfGameTexture, _palette);
+	Graphics::Surface *convertedSubSurface = subSurface.convertTo(_gfxState.gfxMode->svmBufferPF, _palette);
```
bgK: Wouldn't the code be clearer if it said we are converting to the pixel format of `_gameTopTexture`?

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -239,12 +388,12 @@ void OSystem_3DS::grabPalette(byte *colors, uint start, uint num) const {
 }
 
 void OSystem_3DS::copyRectToScreen(const void *buf, int pitch, int x,
-                                           int y, int w, int h) {
+                                   int y, int w, int h) {
```
bgK: So, now you've done the rest of the changes, the pixel format used by the screen texture and the game are often the same. Wouldn't it be worth it to create a fast path for that case here, bypassing the whole color conversion business?

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -609,16 +805,22 @@ void OSystem_3DS::setMouseCursor(const void *buf, uint w, uint h,
 
 	if (w != _cursor.w || h != _cursor.h || _cursor.format != _pfCursor) {
 		_cursor.create(w, h, _pfCursor);
-		_cursorTexture.create(w, h, _pfGameTexture);
+		_cursorTexture.create(w, h, _gfxState.gfxMode);
```
bgK: Why not keeping the cursor texture always in RGBA so there is always alpha bits available and there is no need to draw it in software on the screen? It should not change too often nor be too expensive to convert as it is small?

backends/platform/3ds/osystem-graphics.cpp
BallM4788: ### Short explanation:
Because the screen render targets can only accept a single "in" format, both the game and the overlay have to share the same format.

### Long Explanation:
As the 3DS backend stands currently, all surfaces are converted to RGBA8 before being passed on to the screen render targets. Additionally, the current backend uses Ctrulib's default graphics initialization, which initializes the LCD framebuffers to RGB8 (technically it's BGR8 in the 3DS's case but that's neither here nor there). In effect, games that do not already run in RGBA8, ARGB8, or whatever the backend most prefers (which in this case is also RGBA8) are being converted to two other formats each and every frame. In the case of a game like Riven, which has a game format of RGB565 (2 bytes), each frame is converted to RGBA8 (4 bytes) for the screen render targets, which is then converted into RGB8 (3 bytes) for the framebuffers. This is what causes huge slowdowns.

Unfortunately, simply changing the framebuffers to use RGBA8 as well does not solve the problem. What I found does solve the problem is to minimize instances of format conversions. This pull request does that by altering the format used throughout (surfaces, screen render targets, framebuffers, displays) to match or closely match the format used by whatever game is about to be played. Because render targets can only accept one "in" format, however, it is necessary to keep all surfaces (including the overlay and cursor) the same format.

backends/platform/3ds/osystem-graphics.cpp
BallM4788: See my comment above with regards to the game and overlay formats. Same reason.

BallM4788:
Rebased to fix conflict with the new keymapper commits (which works great on the 3DS BTW, great work guys).

backends/platform/3ds/osystem-graphics.cpp
bgK: > Because the screen render targets can only accept a single "in" format, both the game and the overlay have to share the same format.

Are you sure though? See this branch here I've just updated: https://github.com/bgK/scummvm/tree/3ds-565

There the game texture is either RGBA or RGB565 depending on the game surface pixel format. The other surfaces are always RGBA. It seems to have similar performance characteristics when compared to yours.

BallM4788: Well dang (in a good way), you're right. I'll get to work on changing the PR.

Also, I think the the keymapper commit caused a bug; after launching a game, the first touchscreen input doesn't work, causing each subsequent touchscreen input to execute the one that came before it. I thought it might be something on my branch causing it, but it's in the master branch as well.

bgK: Thanks, I'll look into the keymapper bug.

BallM4788:
@bgK Thanks for your suggestions. They allowed me excise a good chunk of code from the PR!

backends/platform/3ds/osystem-graphics.cpp
BallM4788: I've finished reworking the PR. Unless you have any other questions, it should be ready for merging.

backends/platform/3ds/sprite.h
```diff
@@ -58,23 +57,28 @@ class Sprite : public Graphics::Surface {
...
 private:
+	uint32 textureTransferFlags;
 	bool dirtyPixels;
 	bool dirtyMatrix;
 	C3D_Mtx modelview;
 	C3D_Tex texture;
 	vertex* vertices;
-	int posX;
-	int posY;
+	int posX, posLastX;
```
bgK: `posLastX` and `posLastY` don't seem to be used.

backends/platform/3ds/sprite.cpp
```diff
@@ -20,17 +20,23 @@
  *
  */
 
+#include "osystem.h"
```
bgK:
```suggestion
#include "backends/platform/3ds/osystem.h"
```

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -549,14 +666,19 @@ void OSystem_3DS::displayActivityIconOnOSD(const Graphics::Surface *icon) {
 		_activityIcon.free();
 	} else {
 		if (!_activityIcon.getPixels() || icon->w != _activityIcon.w || icon->h != _activityIcon.h) {
-			_activityIcon.create(icon->w, icon->h, _pfGameTexture);
+			_activityIcon.create(icon->w, icon->h, &DEFAULT_MODE);
+		}
+
+		if (_pfGame == _gameTopTexture.format) {
```
bgK: Is that the correct check? Shouldn't it be between the icon passed by the caller and `_activityIcon`?

backends/platform/3ds/osystem-graphics.cpp
```diff
@@ -30,17 +30,37 @@
...
 namespace _3DS {
-
-void OSystem_3DS::initGraphics() {
+/* Group the various enums, values, etc. needed for
+ * each graphics mode into instaces of GfxMode3DS */
+static const GfxMode3DS _modeRGBA8 = { Graphics::PixelFormat(4, 8, 8, 8, 8, 24, 16, 8, 0),
+                                       GPU_RGBA8, TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGBA8) };
+static const GfxMode3DS _modeRGB565 = { Graphics::PixelFormat(2, 5, 6, 5, 0, 11, 5, 0, 0),
+                                        GPU_RGB565, TEXTURE_TRANSFER_FLAGS(GX_TRANSFER_FMT_RGB565) };
+static const GfxMode3DS _modeRGB555 = { Graphics::PixelFormat(2, 5, 5, 5, 0, 10, 5, 0, 0),
```
bgK: Games using the RGB555 pixel format have glitchy colors. The pixel values probably need to be shifted by one bit to the left to be compatible with the RGBA5551 pixel format the 3DS expects.

I use the FM-Towns version of Zak McKraken from GOG for testing.

backends/platform/3ds/sprite.cpp
BallM4788: There are a few other instances like this where the full path isn't used. Do you want me to change those as well?

bgK: Your choice.

BallM4788:
That should be everything!

bgK:
Thanks again. Merged manually as b08ab0e1307460460c8923af4f5d8d134900c489.

PR #2022, opened 2020-01-27
Wme detection

lolbot-iichan:


lolbot-iichan:
Sorry, pressed something wrong and non-rebased branch with empty description was PR'd.
Replaced with #2023

PR #2025, opened 2020-01-27
ULTIMA: Ultima games engine

dreammaster:
I'm at a stage where I'm happy enough with the Ultima engine that I've been working on to open it for review for inclusion into master. This engine currently consists of work for three games in the Ultima series.

1) Ultima VIII - Pagan.
This is based on the Pentagram project. I've finally got music playback working, and am about to start a full playthrough test of the game to may sure everything in the engine still works, and the game is completeable.

2) Ultima VI - The False Prophet
This is based on the Nuvie project. I added two game detection entries, one for Nuvie's "classic" mode, and another that turns on the Nuvie enhancements like Ultima 7 style gumps and a full-screen map. I haven't really tested it much beyond the game starting up, the initial fight, and moving around the castle a bit. But everything seems to work. One downside is that I had to include the entirety of the Lua interpreter that Nuvie used in order for the cutscenes and various in-game monster AI and combat calculations to work. So testing it further, and seeing whether I can get it to use the common Lua code is going on a back burner until I've finished testing Pagan.

3) Ultima 1
This is an in progress work based on my decompilation of the first Ultima game (not counting Akalabeth). Most of the in-game towns/castles, overworld, and dungeons are supported. It's just missing the outer space sim section, and the endgame bossfight. Further work on this will wait until the other two Ultima games are done.


graphics/managed_surface.h
```diff
@@ -61,18 +61,43 @@ class ManagedSurface {
...
+	 */
+	uint _transparentColor;
+	bool _transparentColorSet;
+
+	/**
+	 * Local palette for 8-bit images
+	 */
+	uint32 _palette[256];
+	bool _paletteSet;
+protected:
```
sev-: This breaks at least Director engine. Why is this declared protected now?

sluicebox:
I would have clicked an emoji reaction but they don't have one for "jaw dropped like at the end of Parasite"

bluegr:
Fantastic work, well done!

Some questions / observations:
- I see no mention of Exult (Ultima 7). Is that considered for inclusion at some point in the future?
- I'm a bit confused with the current folder structure. In particular: 
1) there's a folder called shared/core, a shared/engine and a shared/std. What are the differences?
2) what kind of code is contained in shared/early?
3) there's also a folder called ultima0. What kind of code does it contain?
- What kind of issues were you having with the common LUA code? We use LUA 5.2.3, whereas nuvie used 5.1.3, which are quite similar
- As a neat side project, the same engine used for Ultima 8 [is also used in Crusader: No Remorse, and Crusader: No Regret](https://www.vogons.org/viewtopic.php?t=904). I suppose that's still in early development, as there's only a stub in ultima/ultima8/games/remorse_game.*

graphics/managed_surface.h
dreammaster: Whoops, sorry. I'd gotten so used to just compiling with only the engine enabled, I forgot to ensure that my changes to the core didn't break anything else. I'll do an interactive rebase this evening and fix it in the history.

dreammaster: Regarding lua, any help would be welcome. I have a branch ultima6_lua on my github repository that has a single commit at the head where I was experimenting with extensions to the core Lua to re-enable proper file loading again. That could be a good starting point for any experiments.

dreammaster: The branch is a bit out of date though, so I'll also rebase it against the current ultima branch tonight.

yuv422:
Hey,
I'm still not sure how I feel moving the Nuvie project into SummVM. I've been thinking about it a bunch recently. I think it's mainly due to losing the Nuvie project as an entity and for it to just become another game that is supported by the ScummVM project. :(

sev-:
Shall we withstand from the merging? Actually, the merge could help to bring more movement and fresh blood to Nuvie, and make it highly portable. You are more than welcome to join us if we merge.

dreammaster:
Hey @yuv422 . Good to hear that you're still alive. Obviously I initially felt upset at hearing your doubts now, after all the work I did.. particularly given that work on Nuvie hadn't been done for a while, and I tried to contact you both directly, as well as via Dominus Dragon previously to see if the project was still active. However, on reflection, I calmed down and remembered that projects like this are a labor of love, and it's best to go with whatever place of development you're most happy with if you intend to keep working on Nuvie.

One possibility is that I'm presuming that the main Ultima VI is "done" and the game is completeable, and it's just work to be done on Savage Empire and Martian Dreams. In which case, the engine would still be merged in, but we'd refrain from doing any significant development on it beyond getting it stable and supported as is. That way, players could choose to use either ScummVM or Nuvie as they see fit, And if/when Nuvie, still as a separate project, ever got support done for the other games and you considered it "done", I could reintegrate the updated codebase into ScummVM at that point again; hopefully quicker the next time due to the foundation already have been laid by my previous conversion.

But if you do decide to carry on future development directly in ScummVM, we'd be very happy, and as Sev said, it might encourage further contributions. Not to mention ScummVM's portability.

yuv422:
@dreammaster I think this could work. I also think that scummvm could bring lots of benefits to Nuvie. I just don't really want to see Nuvie die as a project. Like you said I've spend a large amount of time with the project and it holds a special place in my heart. :) Could we maybe keep the Nuvie name for the engine in scummvm and maybe a nuvie logo splash screen :)

It would be great to see more progress on the engine. Especially supporting MD and SE.

I'll ask the other Nuvie team members to get their thoughts on merging with ScummVM.

dreammaster:
>Could we maybe keep the Nuvie name for the engine in scummvm and maybe a nuvie logo splash screen :)

That could work. Generally with the other projects previously merged in, it's just been credits in the credits.pl file, and on wiki.scummvm.org.. But it would be pretty easy to add a splash screen for Nuvie as the sub-engine starts up; it would just need someone with a suitable artistic flair to draw a suitable screen. :)

bluegr:
@yuv422: as mentioned , merging Nuvie with ScummVM would bring new life to the project, and more portability, which is what happened in the past with the FreeSCI, Sarien and TrollVM projects. Check this article for more insights:

https://arstechnica.com/gaming/2012/01/maniac-tentacle-mindbenders-of-atlantis-how-scummvm-kept-adventure-gaming-alive/3/

So... merging your work definitely won't make it just another game engine. It'll benefit from the common code we got, and ScummVM's portability. And as mentioned, you're always free to join us, and work not only with Nuvie, but with other engines too :)

I don't know if it makes sense to have two separate projects for the same game in the future... imagine that you are a player who wants to play the game... why would he choose one project over the other?

Of course, all authors and references to the engine will be added to our credits:
https://www.scummvm.org/credits/#code_contrib

As for the splash screen: do you feel that this would give any extra recognition to Nuvie? Will Nuvie coexist with ScummVM in the future?

yuv422:
> As for the splash screen: do you feel that this would give any extra recognition to Nuvie? Will Nuvie coexist with ScummVM in the future?

Yeah I don't think there is much point in maintaining both versions going forward. But I would like to keep the Nuvie name going. Which I think the engine splash would accomplish.

sev-:
Compilation is still failing on Mac:

```
./engines/ultima/ultima6/files/nuvie_io_file.h:65:19: error: no member named '__1' in namespace 'Ultima::std'; did you mean '::std::__1'?
                return _file != nullptr;
                                ^~~~~~~
```
The thing is that nullptr is an object in MacOS X, and that leads to tons of side effects. Here you have a nice and nasty clash of the namespace `std` in `shared/std/*.h`. I recommend renaming that to `Std` or something like this.

sev-:
@yuv422 Is this your logo: http://nuvie.sourceforge.net/images/nuvie_logo.png ?

engines/ultima/configure.engine
```diff
@@ -0,0 +1,6 @@
+# This file is included from the main "configure" script
+# add_engine [name] [desc] [build-by-default] [subengines] [base games] [deps]
+add_engine ultima "Ultima games" no "ultima1 ultima6 ultima8"
+add_engine ultima1 "Ultima I" no "" "" "highres 16bit"
```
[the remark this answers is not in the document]
dreammaster: > Also, while we have you here ;) Dreammaster needed any disassemblies which you probably still have, so he could complete the support for other sub-engines.

I hope you don't take it that I'm stepping on your toes, so to speak. If you intend to keep working on the games, you'd remain the primary developer. And goodness knows I have plenty to get on with for the other games of the series. But it would be a good idea, if you're willing, to make a copy of your disassemblies available to us, Dominus, and any others, just for archival purposes if nothing else.

engines/ultima/detection.h
```diff
@@ -0,0 +1,90 @@
...
+	virtual const char *getEngineId() const override {
+		return "ultima";
+	}
+
+	virtual const char *getName() const override {
+		return "Ultima";
+	}
+
+	virtual const char *getOriginalCopyright() const override {
+		return "Ultima games (C) Origin Systems Inc.";
```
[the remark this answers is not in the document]
dreammaster: Okay, I'll do that tonight.

engines/ultima/module.mk
```diff
@@ -0,0 +1,544 @@
...
+	shared/gfx/visual_item.o \
+	shared/gfx/visual_surface.o \
+	shared/maps/map.o \
+	shared/maps/map_base.o \
+	shared/maps/map_tile.o \
+	shared/maps/map_widget.o \
+	shared/maps/creature.o
+endif
+
+ifdef ENABLE_ULTIMA1
```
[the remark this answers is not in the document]
dreammaster: Essentially the first block is for the "shared" code that's currently only used by Ultima 1, but will likely be reused for Ultima 2 and other early games as well. So semantically it just seemed cleaner to have it as a separate block after the shared files that all the sub-engines need, rather than just merging it into the Ultima 1 block.

engines/ultima/shared/core/str.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "ultima/shared/core/str.h"
+
+namespace Ultima {
+namespace Shared {
+
+int String::indexOf(char c) const {
```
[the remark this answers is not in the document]
dreammaster: Hmm.. I think you're right. Goodness knows that this isn't the first time I've had to implement similar functions in other engines, so they might as well go in the core to be available for other usage

dreammaster:
> Compilation is still failing on Mac:

Darn. I thought I'd got all such problems fixed. I think your idea of changing std to Std has merit if it'll help avoid such clashes, and it'll be easy enough to do a global search and replace of existing namespace qualifiers in the code. Again, I'll take care of it tonight

engines/ultima/configure.engine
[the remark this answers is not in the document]
dreammaster: > quick question. What is the reason for having individual games as the targets? I mean, usually we do it only when there are some difference in hardware requirements like resolution.

Now that I think of it, that's actually a good question. My initial though was that the sub-engines are more or less full blown engines in their own right, just consolidated together in one overall engine, with some shared code for event handling, saves, screen, etc. This is moreso than engines like TsAGE where although the different games have their own hardcoded logic, they share more of the engine in common. But in retrospect, you're right.. the sub-engines aren't so big that I can see a particular need or advantage for having them individually disable-able. I'll go ahead and remove them. This will also simplify the module.mk

> Also, is ultima1 really highres and 16bit?

Not in classic mode, but it also has even-more-in-progress code for creating an enhanced mode that will use the Ultima 6 tileset. Though if there won't be individual sub-engines listed anymore, it's a moot point.

dreammaster:
> Fantastic work, well done!

Thanks :)

> Some questions / observations:
> 
> * I see no mention of [Exult](http://exult.sourceforge.net/) (Ultima 7). Is that considered for inclusion at some point in the future?

It's definitely a possibility. For now, at least, Exult is still an active project, and is also complicated by needing to also support Exult Studio. So I'll likely end up working on the earlier games first after the current ones get officially supported. Maybe even in pull in xu4 as well.

> * I'm a bit confused with the current folder structure. In particular:
> 
> 1. there's a folder called shared/core, a shared/engine and a shared/std. What are the differences?
shared/engine is the subfolder for more directly core engine related stuff like the base engine class, the events manager, and so on.
shared/core is for more general purpose classes, like strings widgets, etc.
shared/std is for mockups/reimplementation of std library classes, so I didn't need to bulk changes to the codebase I was pulling in to use ScummVM's classes and methods directly, and potentially introduce errors.  

> 2. what kind of code is contained in shared/early?
shared/early is an interesting special case. Prior to doing the pull request, I wanted to bring all three Ultima games I'd worked on into a single branch. The shared/early was an attempt to keep some files separate from shared/core/ that would likely end up only being used by the earlier games, and not for u6+ onwards. This was only partially successful, since shared/core/ still ended up with some classes like Map and Message Target that U6 & U8 aren't using. At the point when I return to work on U1, I'll probably clean up the shared/ folder structure further

> 3. there's also a folder called ultima0. What kind of code does it contain?

It contains more tentative/skeleton code for eventually supporting Akalabeth. 

> * What kind of issues were you having with the common LUA code? We use LUA 5.2.3, whereas nuvie used 5.1.3, which are quite similar

I'm honestly not sure. By the time I had a proper look at the common lua code, and realised it was hardcoded for loading only a config.lua, I had already imported and converted Nuvie's Lua version, and had it working. I did tentatively try adding proper loading code to common lua, but when I still got a general error, I put further looking into it on a backburner. At Sev's request, I've wrapped the Lua access points in the engine with an ifdef for 'USE_COMMON_LUA', so it'll be easy to switch it over if you want to do any experiments.

> * As a neat side project, the same engine used for Ultima 8 [is also used in Crusader: No Remorse, and Crusader: No Regret](https://www.vogons.org/viewtopic.php?t=904). I suppose that's still in early development, as there's only a stub in ultima/ultima8/games/remorse_game.*

Unfortunately, according to Strangerke, only wjp and/or fingolfin had the IDBs for the games, and they're not responding. So further support for them isn't likely. But there's so little code for the games, it wasn't worth the effort to remove it.

common/str.h
```diff
@@ -306,6 +307,52 @@ class String {
...
+	size_t rfind(const char *s) const;
+	size_t rfind(const String &s) const {
+		return rfind(s.c_str());
+	}
+
+	/** Does a reverse find for a passed character */
+	size_t rfind(char c, size_t pos = npos) const;
+
+	/** Find first character in the string matching the passed character */
+	size_t find_first_of(char c, size_t pos = 0) const;
```
bgK: The existing methods in this class use a camel case naming convention.

engines/ultima/configure.engine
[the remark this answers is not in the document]
yuv422: @sev- Yes I've got some IDA dissasseblies for U6 and MD. Where should I put them?

yuv422:
> @yuv422 Is this your logo: http://nuvie.sourceforge.net/images/nuvie_logo.png ?

Yes that's the Nuvie logo. :)

engines/ultima/configure.engine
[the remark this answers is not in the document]
sev-: Please e-mail them to me at sev@scummvm.org or upload to thing like Google Drive and share the link

common/str.h
dreammaster: Good point. Those methods currently match the naming of methods from the C std::string class, but for ScummVM it'll be better for consistency to have them camelcase. Particularly since for both the current Ultima code as well as any future code we import that uses std::string, it will be pretty simple to do a global search and replace to correct the method names to the ScummVM version

dreammaster:
> > @yuv422 Is this your logo: http://nuvie.sourceforge.net/images/nuvie_logo.png ?
> Yes that's the Nuvie logo. :)

Maybe someone familiar with Lua programming could, post merge, change the Lua scripts for the game intro sequences to additionally load the image in the corner of the first shown screen. Both adding the image and updating the existing Lua scripts would both require updating ultima.dat, so it could be done at the same time.

yuv422:
> > > @yuv422 Is this your logo: http://nuvie.sourceforge.net/images/nuvie_logo.png ?
> > > Yes that's the Nuvie logo. :)
> 
> Maybe someone familiar with Lua programming could, post merge, change the Lua scripts for the game intro sequences to additionally load the image in the corner of the first shown screen. Both adding the image and updating the existing Lua scripts would both require updating ultima.dat, so it could be done at the same time.

I can do this in another PR if you like.

dreammaster:
> I can do this in another PR if you like.

Thanks. Having the expert take care of it will make things easier :)

devtools/credits.pl
```diff
@@ -1015,10 +1015,14 @@ sub add_paragraph {
 			end_section();
 
 			begin_section("Ultima");
-				add_person("Dominik Reichardt", "Dominus Dragon", "(Pentagram)");
-				add_person("Eric Fry", "yuv422", "(Nuvie)");
 				add_person("Paul Gilbert", "dreammaster", "");
+				add_person("Eric Fry", "yuv422", "(Nuvie)");
```
dreammaster: I added users with 100+ commits to either project, based on Openhub (thanks Lightkey for pointing out I use that to check contribution totals). If any of the users have publically known names, let me know and I'll flesh out their entries

common/str.cpp
```diff
@@ -434,7 +434,7 @@ void String::deleteChar(uint32 p) {
 }
 
 void String::erase(uint32 p, uint32 len) {
-	if (p == npos)
+	if (p == npos || len == 0)
```
dreammaster: Credit to yuv422 for noticing the erase problems

yuv422:
image
The icons on the inventory view are broken

yuv422:
Also the game crashes when you enter and exit the main Nuvie menu using the esc key.
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   scummvm                       	0x000000010155263c Ultima::Shared::EventsManager::getMousePos() const + 12 (events.h:337)
1   scummvm                       	0x0000000101552611 Ultima::Nuvie::Screen::get_mouse_location(int*, int*) + 33 (screen.cpp:1599)
2   scummvm                       	0x00000001014d7edd Ultima::Nuvie::GUI::Display() + 189 (gui.cpp:207)
3   scummvm                       	0x00000001014d8067 Ultima::Nuvie::GUI::CleanupDeletedWidgets(bool) + 343 (gui.cpp:166)
4   scummvm                       	0x00000001014d859a Ultima::Nuvie::GUI::HandleEvent(Common::Event*) + 1034 (gui.cpp:343)
5   scummvm                       	0x000000010146c3d7 Ultima::Nuvie::Events::update() + 183 (events.cpp:183)
6   scummvm                       	0x0000000101487ea1 Ultima::Nuvie::Game::play() + 129 (game.cpp:646)
7   scummvm                       	0x00000001015c1f87 Ultima::Nuvie::NuvieEngine::run() + 87 (nuvie.cpp:179)
8   scummvm                       	0x00000001011f8311 runGame(Plugin const*, OSystem&, Common::String const&) + 3361 (main.cpp:300)
9   scummvm                       	0x00000001011f6888 scummvm_main + 3000 (main.cpp:582)
10  scummvm                       	0x00000001011d8b06 main + 182 (posix-main.cpp:45)
11  libdyld.dylib                 	0x00007fff6dd823d5 start + 1
```

dreammaster:
> Also the game crashes when you enter and exit the main Nuvie menu using the esc key.

Fixed. There was an incorrect forced cast between GUI_CallBack and CallBack.

dreammaster:
> The icons on the inventory view are broken

There's something screwy with the buttons; likely a format or palette style mismatch between the colour32 array and the resulting ScumMVM surface pixel format. I'll have to look into it further post-merge

michas0602:
I can find no way to jump from U6 back to ScummVM's menu...would it be possible to make all Ultima-games more conformant to others (F5 to open menu)?

But my most important problem is the mouse - I can't play U6 at all because the mousepointer is so sluggish and constantly moves into one direction.

It would also be very cool if you manage to support the german patched version...

neuromancer:
> But my most important problem is the mouse - I can't play U6 at all because the mousepointer is so sluggish and constantly moves into one direction.

I can confirm this happens in the main menu.

dreammaster:
Cool. I've been preparing to announce testing for both Ultima VI and Ultima IV in the next few days,, and it looks like you beat me to it :)

1) The standard key across all the engines is Ctrl+F5 for the ScummVM Global Main Menu. You're right in that F5/F7 would be a nice shortcut, but unfortunately for Ultima VI, the game uses the function keys as a shortcut to look at specific companions' inventories. So the key is unavailable. Of course, the numeric keys are also in use already for solo mode character selection. I briefly mulled having a scheme like Ctrl-number to open a user's inventory, but in the end decided it'd too far a departure from what people will expect.

At least pressing Escape will give you then in-game options dialog, which contains options for Save and Load as well. Not quite as quick as a single F5/F7 keypress, but hopefully still reasonable.

2) As for the cursor moving, I suspect I know what the problem is. In recent times ScummVM startup was changed so that joystick functionality is turned on by default. Since then, I noticed the mouse not drifting  in Windows, but it was in my VirtualBox VM when I tried to performance testing of the code. I was able to work around it by specifying --joystick=-1 on the command line. See if that helps, and if so, you might want to raise it as a general issue; since I'm not as familiar with the joystick handling code, and it may affect other players playing other games.

As for German, sure, as long as the Nuvie codebase already supports it. Can you provide a link to where the patch is available?

michas0602:
Hi and many thanks for your quick reply,
if I start ScummVM with --joystick=-1, there is no change in the behaviour - I get stuck in the menu as the mousepointer is uncontrolable and ESC does not do anything. I can just kill ScummVM through the taskmanager.

So whatever is causing it, the mousepointer definately needs some work in my opinion.

Regarding german patch, here is a link:
https://sirjohn.de/ultima-6/

With best regards, 
Michael

dreammaster:
Sorry, I can't offer any other suggest. You should report your problem to the Help and Support group in the Forums. Maybe someone will have better suggestions about your problem

michas0602:
Already wrote this in my (closed) bugreport, unfortunately the mouse-problem is still actual with a wireless-mouse here, at least in the menu.

PR #2027, opened 2020-01-29

jepael:
Use a cast to fix a compiler warning about comparison between signed
and unsigned integer expressions.


sev-:
Thanks! Merging.

PR #2028, opened 2020-01-30
GOB: Add workaround for incorrect Goblins 3 CD audio track strings

lawadr:
Fix for https://bugs.scummvm.org/ticket/11335

Added the fix directly below an existing precedent for this type of issue.

One potential problem might be if there is a version of the game out there in the wild without this bug, but given the naive approach of the aforementioned precedent and the fact that I can't find any other versions of the game to test it...


bluegr:
The workaround proposed is done in a similar way as fix done for Fascination (right above this one).

So, this can be merged. Thanks for your work!

PR #2029, opened 2020-01-31
GRAPHICS: MACGUI: Fix compiler warning

jepael:
Change variable to unsigned to fix a compiler warning about
comparison between signed and unsigned integer expressions.


sev-:
Thanks! merging

PR #2032, opened 2020-02-02
KYRA: Add detection for Polish fan translation of kyra1

xeros:
Add checksum of "GEMCUT.PAK" for Polish text translation to kyra1 engine.

### DIFF
--- a/engines/kyra/detection_tables.h
+++ b/engines/kyra/detection_tables.h
@@ -382,6 +382,22 @@ const KYRAGameDescription adGameDescs[] = {
 		},
 		KYRA1_CD_FLAGS
 	},
+	{ // Polish fan translation of Macintosh CD version
+		{
+			"kyra1",
+			"CD",
+			{
+				{ "GEMCUT.PAK", 0, "8ddf09fd6bfafdb27b4cf31c5a6fc91d", -1 },
+				{ "BEAD.CPS", 0, "3038466f65b7751451844707187aa401", -1 },
+				{ NULL, 0, NULL, 0 }
+			},
+			Common::EN_ANY,
+			Common::kPlatformMacintosh,
+			ADGF_CD,
+			GUIO0()
+		},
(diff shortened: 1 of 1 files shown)


sev-:
Thanks, could you please fix the commit log message in accordance with our [Commit Guidelines](https://wiki.scummvm.org/index.php?title=Commit_Guidelines)?

xeros:
Is it better now?

lotharsm:
Please use something along the lines of `KYRA: Add detection for Polish fan translation of kyra1` for the commit message. Please note that you should change the commit message itself, not only the PR title.

raziel-:
@xeros

Take a look at the other PR's

Something like this should fit:
KYRA: Add detection for Kyrandia 1 polish fan translation

xeros:
Done, thank you.

lotharsm:
Please change the commit message itself, not the PR title.

athrxx:
Have you tested this? Does it work as expected?

Having this detected as English will lead to various strings (including the intro) being displayed in English. Setting it to Polish won't work, since we don't have that language resources. But maybe this fan translation is supposed to be like that anyway?

xeros:
Yes, I have tested it and played most of game locations on PlayStation Classic, having ScummVM compiled as RetroArch core.

Everything works fine for me, but you're right about having some strings not translated.
Intro dialogs and item names are in English (Polish translation for them is integrated in MAIN.EXE file), but game dialogs & location names, loaded from PAK files are in Polish.

It's a fan translation where translated strings to Polish language are mapped to "ENG" and it's different than official Polish version of Legend of Kyrandia 2: Hand of Fate, where Polish strings are properly mapped as "POL". I've started some work on this version too, but it has to be handled differently due many differences to English version.

xeros:
Commit message corrected.

sev-:
Thanks, merging.

PR #2034, opened 2020-02-02
ENGINES: add GUI warning when encountering a pirated game

ZvikaZ:

### DIFF
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -30,6 +30,7 @@
 #include "common/textconsole.h"
 #include "common/translation.h"
 #include "gui/EventRecorder.h"
+#include "gui/message.h"
 #include "engines/advancedDetector.h"
 #include "engines/obsolete.h"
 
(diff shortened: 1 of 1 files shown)


engines/advancedDetector.cpp
```diff
@@ -133,6 +134,8 @@ bool cleanupPirated(ADDetectedGames &matched) {
 		// We ruled out all variants and now have nothing
 		if (matched.empty()) {
 			warning("Illegitimate game copy detected. We provide no support in such cases");
+			GUI::MessageDialog dialog("Illegitimate game copy detected. We provide no support in such cases", "OK");
```
bgK: This crashes when executed from the command line (`scummvm --detect` for example). A simple way to fix it is to check is the GUI manager has been instantiated already, if not then the call comes from the command line. Another way, is to do the same thing as with the unknown game dialog. Store in the `DetectedGame` struct that the game is pirated and display the GUI message when the detection code is called from the GUI.

ZvikaZ: Thanks for the detailed explanations.
I implemented it with `if (GUI::GuiManager::hasInstance())`, and updated the PR.

bluegr:
Looks good now, thanks! Squashing + merging

PR #2035, opened 2020-02-02
SKY: Keymapper integration

bgK:
scummvm-00008

### DIFF
--- a/engines/sky/detection.cpp
+++ b/engines/sky/detection.cpp
@@ -25,6 +25,9 @@
 
 #include "base/plugins.h"
 
+#include "backends/keymapper/action.h"
+#include "backends/keymapper/keymap.h"
+#include "backends/keymapper/standard-actions.h"
 #include "common/config-manager.h"
 #include "engines/advancedDetector.h"
 #include "engines/metaengine.h"
@@ -84,6 +87,7 @@ public:
 	const ExtraGuiOptions getExtraGuiOptions(const Common::String &target) const override;
 	PlainGameDescriptor findGame(const char *gameid) const override;
 	DetectedGames detectGames(const Common::FSList &fslist) const override;
+	Common::KeymapArray initKeymaps(const char *target) const override;
 
 	Common::Error createInstance(OSystem *syst, Engine **engine) const override;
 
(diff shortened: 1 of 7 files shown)


engines/sky/detection.cpp
```diff
@@ -201,6 +205,80 @@ DetectedGames SkyMetaEngine::detectGames(const Common::FSList &fslist) const {
...
+	act->setCustomEngineActionEvent(kSkyActionToggleFastMode);
+	act->addDefaultInputMapping("C+f");
+	shortcutsKeymap->addAction(act);
+
+	act = new Action("RFAST", _("Toggle really fast mode"));
+	act->setCustomEngineActionEvent(kSkyActionToggleReallyFastMode);
+	act->addDefaultInputMapping("C+g");
+	shortcutsKeymap->addAction(act);
+
+	act = new Action(kStandardActionOpenDebugger, _("Open debugger"));
```
sev-: After looking at this. Could it possibly make a straightforward helper method, which would process the following:

```
keymap = {
  { kStandardActionOpenMainMenu,  _s("Open control panel"), kSkyActionOpenControlPanel, { "F5", "JOY_X", NULL } },
  { "RFAST", _s("Toggle really fast mode"), kSkyActionToggleReallyFastMode, { "C+g", NULL } },
 ...
};
```

bgK: It would certainly be possible to make such a helper. However I did not choose this approach because:
* It makes the code harder to read. In the struct literal the names of the fields are not immediately visible.
* It makes the code harder to edit. If you want to change one of the action in a way that does not fit the struct / helper to experiment, you have to rewrite in to use the Action class directly anyway.

rsn8887:
Why not map skip line to a joy button? I find it essential.

rsn8887:
These keymaps should include default game controller mappings too. Otherwise users will find their controllers not working in all the games that have custom mappings.

bgK:
> Why not map skip line to a joy button by default? I find it essential. Same goes for global default mapping in metaengine.

Can you suggest a button?

> These keymaps should include default game controller mappings too. Otherwise users will find their controllers not working in all the games that have custom mappings.

What do you mean? That keymap is including game controller binding, is it not?

rsn8887:
> > Why not map skip line to a joy button by default? I find it essential. Same goes for global default mapping in metaengine.
> 
> Can you suggest a button?

Globally I would suggest "JOY_X". But you are right here in this game that doesn't work because X opens the menu.

> > These keymaps should include default game controller mappings too. Otherwise users will find their controllers not working in all the games that have custom mappings.
> 
> What do you mean? That keymap is including game controller bindings, is it not?

I was going by the screenshot. Yes it is there.

bluegr:
This looks to be in a good state to be merged. Are there any pending TODOs?

bgK:
Ideally we could map skip and skip line on the same game controller button. However someone with a good knowledge on the engine would need to do the required changes. For now, users will have to remap skip line to the button of their choice.

PR #2036, opened 2020-02-02
SCI: fix segfault upon detection of pirated game, without fallback

ZvikaZ:
Since ~2 weeks ago, when adding a pirated SCI game that has no fallback, there is a segfault.
Fixing it here.
(this is not related to my other PR, since it was introduced earlier, I just encountered it while working on the other PR ;-) )

### DIFF
--- a/engines/sci/resource.cpp
+++ b/engines/sci/resource.cpp
@@ -982,7 +982,7 @@ void ResourceManager::init() {
 #ifdef ENABLE_SCI32
 	_currentDiscNo = 1;
 #endif
-	if(g_sci) {
+	if (g_sci) {
 		_patcher = new ResourcePatcher(g_sci->getGameId(), g_sci->getLanguage());
 		addSource(_patcher);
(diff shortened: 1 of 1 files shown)


engines/sci/resource.cpp
```diff
@@ -980,8 +982,12 @@ void ResourceManager::init() {
 #ifdef ENABLE_SCI32
 	_currentDiscNo = 1;
 #endif
-	_patcher = new ResourcePatcher(g_sci->getGameId(), g_sci->getLanguage());
-	addSource(_patcher);
+	if(g_sci) {
```
bluegr: Please fix spacing here, to conform to our coding guidelines

ZvikaZ: We're talking about the `if(g_sci)` that should be `if (g_sci)`, right?
Fixed.

bluegr:
Thanks for your work! Squashing + merging

PR #2037, opened 2020-02-03
ENGINES: Savegame code simplifications

dreammaster:
This pull request aims to simply provide some simplification and improvements for the savegame code, particularly going forward for new engines added in the future. It consists of:

1. Showing the save/load dialogs due to in-game actions
A common feature of a lot of the existing games is that they need to trigger showing the save/restore dialogs from in-game buttons, menus, etc., in addition to from the GMM. As such, engines have gotten into the habit of continuously repeating code copied from the GMM to instantiate the dialog, get the result, and call saveGameState/loadGameState. Two new methods are added to the Engine class - loadGameDialog and saveGameDialog. I've also converted several of the engines to use the new methods rather than their own code. I wasn't pedantic about it.. some engines I wasn't familiar with, or had extra code directly in their save/load methods. It could perhaps be a future task for GSOC applicants to identify other existing engines that could be similarly simplified.

2. Simplified integration of the new extended savegame headers for future engines
Recently adding support for the new extended savegame format to Nuvie and Pentagram, I found it a bit cumbersome to properly integrate. I had to get help to find out the method to call to add the extra info to the savegame, and afterwards I only coincidentally noticed a problem with total play time resetting because I wasn't calling setTotalPlayTime in the load code.

So, to combat this, I've added several new methods to the base Engine class, and changed the implementation of loadGameState and saveGameState. First of all, this shouldn't affect any of the existing engines, since they all override loadGameState and saveGameState, so they'll remain unaffected. When not overridden, the assumption is that new engines will be using the new extended savegame code. The loadGameState and saveGameState methods now contain the code for:
* Opening up the savefiles
* Calling new methods readGameStream/writeGameStream which new engines would now override instead of loadGameState/saveGameState. These methods take in the already opened savefile streams, so there's no need for the engines to provide their own code anymore.
* Finally, either appending the extended savegame info for saving, or setting total play time for loading.

3. Autosaves
Many games have logic for when a save is done, such as displaying a "saved" message of some sort. To facilitate this, the saveGameStream method has an isAutoSave parameter to indicate if the game was being saved due to autosave kicking in, or via a player action. New engines will thus be able to do whatever UI notification they want in the method, and have it be done for saves done both via the GMM, or if the game itself manually called the saveGameDialog method.

Speaking of autosaves, I think this is something that should really be considered for being moved into the core as well. My latest commit added autosave support for Griffon, but it occurred to me that that's somewhat inefficient. Currently, not even a quarter of the engines call the "shouldPerformAutoSave" method. I was thinking maybe we could do the checks in the events managers - it could access the global engine pointer to call the method as well as canSaveGameStateCurrently, and maintain the last autosave time. That way, we could do away with engines having to do it themselves. Thoughts?

### DIFF
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -661,12 +661,16 @@ bool Engine::canLoadGameStateCurrently() {
 }
 
 Common::Error Engine::saveGameState(int slot, const Common::String &desc) {
+	return saveGameState(slot, desc, false);
+}
+
+Common::Error Engine::saveGameState(int slot, const Common::String &desc, bool isAutosave) {
 	Common::OutSaveFile *saveFile = _saveFileMan->openForSaving(getSaveStateName(slot));
 
 	if (!saveFile)
 		return Common::kWritingFailed;
 
-	Common::Error result = saveGameStream(saveFile);
+	Common::Error result = saveGameStream(saveFile, isAutosave);
 	if (result.getCode() == Common::kNoError) {
 		MetaEngine::appendExtendedSave(saveFile, getTotalPlayTime() / 1000, desc);
 
@@ -677,7 +681,7 @@ Common::Error Engine::saveGameState(int slot, const Common::String &desc) {
 	return result;
 }
 
-Common::Error Engine::saveGameStream(Common::WriteStream *stream) {
+Common::Error Engine::saveGameStream(Common::WriteStream *stream, bool isAutosave) {
 	// Default to returning an error when not implemented
 	return Common::kWritingFailed;
 }
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -31,6 +31,7 @@
 #include "engines/engine.h"
 #include "engines/dialogs.h"
 #include "engines/util.h"
+#include "engines/metaengine.h"
 
 #include "common/config-manager.h"
 #include "common/events.h"
@@ -40,6 +41,7 @@
 #include "common/error.h"
 #include "common/list.h"
 #include "common/memstream.h"
+#include "common/savefile.h"
 #include "common/scummsys.h"
 #include "common/taskbar.h"
 #include "common/textconsole.h"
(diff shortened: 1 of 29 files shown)


engines/engine.cpp
```diff
@@ -641,15 +661,73 @@ bool Engine::canLoadGameStateCurrently() {
...
+	GUI::SaveLoadChooser *dialog = new GUI::SaveLoadChooser(_("Load game:"), _("Load"), false);
+	pauseEngine(true);
+	int slotNum = dialog->runModalWithCurrentTarget();
+	pauseEngine(false);
+	delete dialog;
+
+	if (slotNum != -1)
+		return loadGameState(slotNum).getCode() == Common::kNoError;
+
+	return false;
```
bgK: I like showing a GUI error message when loading / saving failed.

```suggestion
	if (slotNum < 0)
		return false;
	
	Common::Error loadError = loadGameState(slot);
	if (loadError.getCode() != Common::kNoError) {
		GUI::MessageDialog dialog(loadError.getDesc());
		dialog.runModal();
		return false;
	}
	
	return true;
```

engines/engine.cpp
```diff
@@ -641,15 +661,73 @@ bool Engine::canLoadGameStateCurrently() {
...
+	return Common::kWritingFailed;
 }
 
 bool Engine::canSaveGameStateCurrently() {
 	// Do not allow saving by default
 	return false;
 }
 
+bool Engine::loadGameDialog() {
+	if (!canLoadGameStateCurrently())
```
bgK: I'm not sure where we are going with this check. It seems to me it can create situations where the player clicks on the load button and nothing happens because loading is not allowed currently. So maybe this should be an assertion to force the caller handle the case or maybe this should display a GUI error message telling the player loading is no possible at the moment.

engines/engine.cpp
```diff
@@ -631,8 +634,25 @@ void Engine::flipMute() {
 }
 
 Common::Error Engine::loadGameState(int slot) {
-	// Do nothing by default
-	return Common::kNoError;
+	Common::InSaveFile *saveFile = _saveFileMan->openForLoading(getSaveStateName(slot));
```
sev-: It seems to me, the indentation here is two spaces instead of a tab.

engines/engine.cpp
sev-: Or an OSD message?

bluegr:
I really like the implementation too. Nice!

henke37:
This for sure is a real issue. We have some missing code in `pink` that I could easily implement with these new features.

engines/engine.cpp
dreammaster: Good point both of you; that'd make more sense to users that simply ignoring the call.

engines/engine.cpp
dreammaster: I'll double check the indentation

PR #2039, opened 2020-02-03
GUI: add optional 'alignment' parameter to 'MessageDialog', and SCI: GK2: if subtitles patch is missing, explain how to install it

ZvikaZ:

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -30,7 +30,12 @@
 #include "sci/engine/guest_additions.h"
 #endif
 
+#include "common/config-manager.h"
 #include "common/util.h"
+#include "common/translation.h"
+#include "common/system.h"
+#include "gui/message.h"
+
 
 namespace Sci {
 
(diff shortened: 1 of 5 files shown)


engines/sci/engine/script_patches.cpp
```diff
@@ -18922,6 +18939,8 @@ void ScriptPatcher::processScript(uint16 scriptNr, SciSpan<byte> scriptData) {
 				// Enable subtitle compatibility if a sync resource is present
 				if (g_sci->getResMan()->testResource(ResourceId(kResourceTypeSync, 10))) {
 					enablePatch(signatureTable, "subtitle patch compatibility");
+				} else {
+					suggestDownloadGK2SubTitlesPatch();
```
bluegr: This should not be enforced for all users. The dialog should only be shown if the user has chosen to display "speech and subtitles" or "subtitles" in ScummVM's audio options

ZvikaZ: OK.
Fixed that.
But before committing and pushing - a question.

Currently, the "edit game" menu for GK2 has "speech and subtitles"/"subtitles"/"speech" choosing disabled (probably because in the past it wasn't relevant).
I have changed that in my global "options" menu to verify my fix.

I'd like to enable that option in the GK2 "edit game" menu.

Is it better to be done on a separate PR?
Or in this PR as a separate commit?
Or even do it all on one commit?

bluegr: You need to check the "Override global audio settings" checkbox in order to override the defaults and change the audio settings for an individual game

ZvikaZ: yeah, I know...

I already fixed that.
It was because of GUIO_NOSUBTITLES in GK2 detection table.
I've removed it, and it's fine now.

The question is technical-aesthetical, whether it's preffered to have the GUIO_NOSUBTITLES removal as a different commit, or even as a different PR, or send now one commit with everything.

bluegr: It would be best to end up with two commits, one in the common code, and one in the SCI engine

ZvikaZ:
Done.
Thanks.

engines/sci/engine/script_patches.cpp
```diff
@@ -18435,6 +18440,43 @@ void ScriptPatcher::applyPatch(const SciScriptPatcherEntry *patchEntry, SciSpan<
...
+			downloadMessage = "";
+		}
+
+		GUI::MessageDialog dialog(_("GK2 has a fan made subtitles, available thanks to the good persons at SierraHelp.\n\n"
+			"Installation:\n"
+			"- download http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip\n" +
+			downloadMessage +
+			"- extract zip file\n"
+			"- no need to run the .exe file\n"
+			"- extract the .exe file with 7-zip (don't worry regarding resource.cfg, you can just replace it, it's not used anyway)\n"
```
bluegr: This is too verbose. There's no need to mention resource.cfg here. Also, not all people use 7-zip

```suggestion
			"- extract the .exe file with a file archiver like 7-zip\n"
```

engines/sci/engine/script_patches.cpp
```diff
@@ -18435,6 +18440,43 @@ void ScriptPatcher::applyPatch(const SciScriptPatcherEntry *patchEntry, SciSpan<
...
+			"- copy the content of GK2Subtitles\\SUBPATCH to:\n"
+			"  = if your game directory has a PATCHES directory, to there\n"
+			"  = if not, to the game directory\n"
```
bluegr: It's easier to just copy the contents into the PATCHES directory, and create it if it doesn't exist

```suggestion
			"- create a PATCHES subdirectory inside your GK2 directory\n"
			"- copy the content of GK2Subtitles\\SUBPATCH to the PATCHES subdirectory\n"
```

ZvikaZ:
Removed yet another redundant line, and squashed commits

bluegr:
Great work, thanks! Merging

PR #2041, opened 2020-02-05
PINK: Implement save & load in the menubar

henke37:
Fairly obvious and straightforward implementation.

Seems to expose a bug with the pause logic, scenes don't resume correctly, either freezing entirely or loosing the BGM. #NotMyFault


bluegr:
Thanks for your work. Could you please add a FIXME regarding the pause logic in the code, so that it won't be forgotten?

sev-:
Thanks, merging.

PR #2042, opened 2020-02-05
TESTBED: Provide correct engine feature implementation data

henke37:
The testbed "engine" doesn't do any of the possible engine features, mark it as such.


bluegr:
Nice work! Merging

PR #2043, opened 2020-02-05
ENGINES: Display a GUI error message when loading / saving failed

bgK:
Without this change, loading / saving fails silently (none of the callers check the return code).


dreammaster:
Looks straightforward and good to me

bluegr:
Looks good to me too

bluegr:
As mentioned, this is quite a good and straightforward change, and the behavior introduced is correct.

Merging

PR #2045, opened 2020-02-07
COMMON: Fix compiler warning

jepael:
Fixes warning comparison between signed and unsigned integer expressions


sev-:
Thanks! Merging.

PR #2046, opened 2020-02-07
BACKEND: Allow SDL2 mapping of L2/R2, fix psp2/switch mapping

rsn8887:
This PR
- implements SDL2 game controller mapping of L2/R2 trigger buttons via the keymapper, similar to how it is done in DevilutionX.
- fixes the default button names for switch and psp2, so that "cross", "circle" etc. is displayed in the keymapper.


rsn8887:
Tested working on Switch

bgK:
I'd prefer if we followed the SDL2 event model as far as standard game controller buttons / axes are concerned. For them left and right triggers are axes, not buttons. I know it does not fit the Switch very well. I've pushed to your branch a commit that enables the keymapper to remap half-axes to actions. It should accomplish the same end result, but without having a hardcoded mapping from an axis to a button in the SDL backend. Is that Ok with you?

rsn8887:
That’s even better, thanks.

rsn8887:
Tested working on Switch. I think this is ready to merge if you agree @bgK.

rsn8887:
Hmm something  is wrong the analog trigger seems to get stuck. The button down is recognized but not button up it seems

rsn8887:
Ok everything works now, thanks! Ready to merge.

PR #2047, opened 2020-02-07

jepael:
Fixes warning of comparison between signed and unsigned integer expressions


sev-:
Thank you :)

PR #2049, opened 2020-02-08
WIN32: `scummvm.log` should not roam.

henke37:
Log files do not need be in the roaming part of an user profile. Especially not ones that are overwritten on each launch.


Kirben:
The log files were placed in that location, to match the config file location.

I don't think it is a good idea to split the ScummVM related files into two completely different locations.

sev-:
Indeed. People even now strongly confused and cannot find the logs easily, and now it is proposed to become even worse.

bluegr:
I agree with the aforementioned comments. Closing this

PR #2050, opened 2020-02-08
ENGINES: Standardization of autosaves

dreammaster:
This pull request is intended to help deal with the fact that the bulk of engines don't implement autosaves, by making it part of the standard base Engine class. There are several provisios, which I'll discuss below:

* First, and foremost, the new autosave functionality relies on an engine implementing savegames via the saveGameState method. This means that even with this, engines like DreamWeb and Tinsel which don't allow saving via the GMM won't benefit from this.

* As a requirement for autosaves to work with the existing engines, I had to standardize saveGameState to always have the extra isAutosave parameter, whereas before in my previous pull request for extended save improvements, there was a separate overloaded version of saveGameState to the existing one. This will slightly affect any existing engines that implement saveGameState and aren't yet merged into master, such as the Blazing Dragons engine. Luckily it'll be easy to simply add in the extra needed parameter to the method, even if unused.

* At it's core, the changes work by adding a new "handleAutosave" method to the base engine class, which the DefaultEventManager's pollEvent method now calls. This seemed the best place, and the method was already accessing the engine via g_engine anyway (for EVENT_MAINMENU handling), so it fits in well. Now, as engines regularly call pollEvent, when the autosave point is reached, it will trigger an autosave, so long as the engine's canSaveGameStateCurrently method returns true.

* To avoid clashes, I removed existing autosave code from the engines, with the exception of scumm.. it's autosaving handling seems somewhat more complicated, so to be on the safe side I left it intact, and instead flag to the Engine that it's autosaves should be turned off. I've added a new method canSaveAutosaveCurrently, which by default simply calls canSaveGameStateCurrently.

* I also added a method getAutosaveSlot for engines to specify which slot is for autosaves. One engine had it in slot 99, and another in slot 999. Griffon, for example, I put the autosave in slot 4, since the engine (based on the in-game save dialog) was previously limited to 4 slots (0-3). Most of the existing engines already supporting autosaves, however, use slot 0, which is the default.

* Some of the engines, already with autosaves in mind, map in-game saves to start at slot 1 so there's no chance of overlap. But since some don't, I've added some logic to mark slot 0 as write protected.. which some engines were also doing already. I also did a minor change to the save load chooser to call a new overloaded listSaves method in the Meta Engine that takes in a 'saveMode' parameter. When in save mode, if autosaves are turned on and there isn't yet a save in the autosave slot, then it will show a dummy 'Autosave' entry that is marked as read only (slots can't be marked as write protected without having an entry for that slot). This won't show in Load mode to avoid confusion. However, obviously, there may still be an outstanding problem if an in-game/origional save dialog puts the first save as slot 0.. potentially they could end up saving a game quickly that later gets overwritten by autosaves.

For the above, one thing I'd like to do is use the querySaveMetaInfos method of the meta engine to generically get the basic information of any existing save, and only save the autosave if the slot is empty, or the description of an existing save matches "Autosave". This would prevent existing saves accidentally getting overwritten. However, I'm not sure how to cleanly get a reference to the meta engine in the engine's saveAutosaveIfEnabled method. The SaveLoadChooser has a lot of code for finding plugins based on engine ids, and getting the meta engine from there.. I'm wondering if there's any existing methods accessible to the engine that simplifies/abstracts the process.

For that matter, I currently have a duplication of getAutosaveSlot in both the engine and meta engine base classes. If there's an easy standardized way for an engine to access the meta engine that launched it, I could do away with the duplication.

### DIFF
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -707,7 +707,7 @@ Common::Error Engine::saveGameState(int slot, const Common::String &desc, bool i
 
 	Common::Error result = saveGameStream(saveFile, isAutosave);
 	if (result.getCode() == Common::kNoError) {
-		MetaEngine::appendExtendedSave(saveFile, getTotalPlayTime() / 1000, desc);
+		MetaEngine::appendExtendedSave(saveFile, getTotalPlayTime() / 1000, desc, isAutosave);
 
 		saveFile->finalize();
 	}
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -148,7 +148,9 @@ Engine::Engine(OSystem *syst)
 		_saveSlotToLoad(-1),
 		_engineStartTime(_system->getMillis()),
 		_mainMenuDialog(NULL),
-		_debugger(NULL) {
+		_debugger(NULL),
+		_autosaveInterval(ConfMan.getInt("autosave_period")),
+		_lastAutosaveTime(_system->getMillis()) {
 
 	g_engine = this;
 	Common::setErrorOutputFormatter(defaultOutputFormatter);
@@ -482,10 +484,16 @@ void Engine::checkCD() {
 #endif
 }
 
-bool Engine::shouldPerformAutoSave(int lastSaveTime) {
-	const int diff = _system->getMillis() - lastSaveTime;
-	const int autosavePeriod = ConfMan.getInt("autosave_period");
-	return autosavePeriod != 0 && diff > autosavePeriod * 1000;
+void Engine::handleAutoSave() {
+	const int diff = _system->getMillis() - _lastAutosaveTime;
+
+	if (_autosaveInterval != 0 && diff > (_autosaveInterval * 1000)) {
+		// Save the autosave
+		saveGameState(getAutosaveSlot(), _("Autosave"), true);
+
+		// Reset the last autosave time
+		_lastAutosaveTime = _system->getMillis();
+	}
 }
 
 void Engine::errorString(const char *buf1, char *buf2, int size) {
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -495,13 +495,23 @@ void Engine::handleAutoSave() {
 
 void Engine::saveAutosaveIfEnabled() {
 	if (_autosaveInterval != 0) {
-		bool canSave = canSaveAutosaveCurrently();
+		bool saveFlag = canSaveAutosaveCurrently();
+		Common::String saveName = _("Autosave");
 
-		if (!canSave || saveGameState(getAutosaveSlot(), _("Autosave"), true).getCode() != Common::kNoError) {
+		if (saveFlag) {
+			// First check for an existing savegame in the slot, and if present, if it's an autosave
+			SaveStateDescriptor desc = getMetaEngine().querySaveMetaInfos(
+				_targetName.c_str(), getAutosaveSlot());
+			saveFlag = desc.getSaveSlot() == -1 || desc.getDescription() == saveName;
+		}
+
+		if (saveFlag && saveGameState(getAutosaveSlot(), saveName, true).getCode() != Common::kNoError) {
 			// Couldn't autosave at the designated time
-			if (!canSave)
-				g_system->displayMessageOnOSD(_("Error occurred making autosave"));
+			g_system->displayMessageOnOSD(_("Error occurred making autosave"));
+			saveFlag = false;
+		}
 
+		if (!saveFlag) {
 			// Set the next autosave interval to be in 5 minutes, rather than whatever
 			// full autosave interval the user has selected
 			_lastAutosaveTime = _system->getMillis() + (5 * 60 * 1000) - _autosaveInterval;
@@ -799,3 +809,9 @@ EnginePlugin *Engine::getMetaEnginePlugin() const {
 }
 
 */
+
+MetaEngine &Engine::getMetaEngine() {
+	const Plugin *plugin = EngineMan.findPlugin(ConfMan.get("engineid"));
+	assert(plugin);
+	return plugin->get<MetaEngine>();
+}
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -489,7 +489,8 @@ void Engine::handleAutoSave() {
 
 	if (_autosaveInterval != 0 && diff > (_autosaveInterval * 1000)) {
 		// Save the autosave
-		saveGameState(getAutosaveSlot(), _("Autosave"), true);
+		if (canSaveAutosaveCurrently())
+			saveGameState(getAutosaveSlot(), _("Autosave"), true);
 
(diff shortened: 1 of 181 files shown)


bluegr:
Excellent work! I really like the revised autosave logic. And the general cleanup of the save code is great, too!

engines/mohawk/riven.cpp
```diff
@@ -761,23 +756,10 @@ void MohawkEngine_Riven::saveGameStateAndDisplayError(int slot, const Common::St
...
-
-	if (!_saveLoad->isAutoSaveAllowed()) {
-		return; // Can't autosave ever, try again after the next autosave delay
-	}
-
-	Common::Error saveError = saveGameState(RivenSaveLoad::kAutoSaveSlot, "Autosave", true);
-	if (saveError.getCode() != Common::kNoError)
-		warning("Attempt to autosave has failed.");
+bool MohawkEngine_Riven::canSaveAutosaveCurrently() {
+	return canSaveGameStateCurrently() && !_gameEnded;
```
bgK: `_saveLoad->isAutoSaveAllowed()` is no longer called?

The thing is in Myst and Riven, we check whether the save in the autosave slot can be overwritten by the autosave (maybe the user somehow put their own save in there). To do so, we open the save and check if it has the autosave flag set.
Reading the file is a somewhat costly operation, and we were careful to do it at most once per autosave interval. Now, in the worst case, it is done every frame and for each event in the queue. So maybe `canSaveAutosaveCurrently` should allow to tell the autosave system "no, autosaving is not possible currently, and don't try again immediately".

engines/engine.cpp
```diff
@@ -480,10 +482,21 @@ void Engine::checkCD() {
...
+
+	if (_autosaveInterval != 0 && diff > (_autosaveInterval * 1000)) {
+		// Save the autosave
+		saveAutosaveIfEnabled();
+	}
+}
+
+void Engine::saveAutosaveIfEnabled() {
+	if (_autosaveInterval != 0 && canSaveAutosaveCurrently())
+		saveGameState(getAutosaveSlot(), _("Autosave"), true);
```
bgK: Should we warn the player if autosaving failed? With an OSD message perhaps ? At least there should be a warning message in the log/console.

bgK:
> The SaveLoadChooser has a lot of code for finding plugins based on engine ids, and getting the meta engine from there.. I'm wondering if there's any existing methods accessible to the engine that simplifies/abstracts the process.

I don't think there is a method to retrieve the MetaEngine from the engine. We can add one the the Engine class:
```c++
MetaEngine &Engine::getMetaEngine() const {
	const Plugin *plugin = EngineMan.findPlugin(ConfMan.get("engineid"));
	return plugin->get<MetaEngine>();
}
```

engines/mohawk/riven.cpp
dreammaster: Actually, that's not the case. The way I set things up in Engine is that it waits for the autosave interval to expire, and only then calls saveAutosaveIfEnabled. It's there that I check canAutosaveCurrently, and even if it returns false, I still reset the autosave interval.

But in retrospect, it does have the downside that if autosaves were, say, set for half a hour, if you just happen to have a dialog or cutscene open on the exact minute the autosave check triggers, it'd wait another half hour before it tried saving again. What I think I'll do is have that if it can't do the autosave, it'll reset the next interval to try in 5 minutes time, irrespective of what the normal autosave interval is.

engines/engine.cpp
dreammaster: Thanks for reminding me. I'd originally decided against showing a message when all I knew was the error message call, but with the OSD display, it'll give the player a non-blocking informative display.

dreammaster:
> I don't think there is a method to retrieve the MetaEngine from the engine. We can add one in the Engine class:

Thanks. I added it in, and used it to add code that avoids saving over a slot if it has a custom save.

engines/mohawk/riven.cpp
bgK: Oh, right, I mixed things up with the previous behavior, sorry.

That call to `_saveLoad->isAutoSaveAllowed()` still needs to be restore though.

engines/engine.cpp
```diff
@@ -493,13 +493,23 @@ void Engine::handleAutoSave() {
...
 	if (_autosaveInterval != 0) {
-		bool canSave = canSaveAutosaveCurrently();
+		bool saveFlag = canSaveAutosaveCurrently();
+		Common::String saveName = _("Autosave");
+
+		if (saveFlag) {
+			// First check for an existing savegame in the slot, and if present, if it's an autosave
+			SaveStateDescriptor desc = getMetaEngine().querySaveMetaInfos(
+				_targetName.c_str(), getAutosaveSlot());
+			saveFlag = desc.getSaveSlot() == -1 || desc.getDescription() == saveName;
```
bgK: Is testing on the save name really a good idea? What if the user changes the language of the GUI?

dreammaster: There is that possibility, yes. But testing the name is the best solution I could come up. I think still allowing for the 'Autosave' to be translateable is better for foreign language users then forcing the text to always be English.

bgK: Have you considered adding a `isAutoSave` flag to `SaveStateDescriptor`? Engines that save the information that a save is an autosave could valorize it. For the other engines it could default to the description being `Autosave`. That way, newer engines can be more robust.

engines/mohawk/riven.cpp
dreammaster: Sorry I forgot it. Fixed.

engines/engine.cpp
dreammaster: The trouble with that is all the different historical engines provide their own querySaveMetaInfos method implementations. So even if I added an isAutosave property to SaveStateDescriptor, I'd have to add the comparison to "Autosave" to every one of them to set it as necessary. Which is too much code duplication for my liking.

bgK: What about:
```c++
enum SaveType {
    kSaveTypeUndetermined,
    kSaveTypeRegular,
    kSaveTypeAutosave
};

SaveType _saveType { kSaveTypeUndetermined };

void setAutosave(bool autosave) {
    _saveType = autosave ? kSaveTypeAutosave : kSaveTypeRegular;
}

bool isAutosave() const {
    if (_saveType != kSaveTypeUndetermined) {
        return _saveType == kSaveTypeAutosave;
    } else {
        return _description == _("Autosave");
    }
}
```

dreammaster: Now that I like. And an autosave property could also be fairly easily added to the extended savegame format to as well, which could set the property.

bluegr:
Are there any pending tasks regarding this pull request? If not, it can be merged this Sunday, the 16th of February

bluegr:
@dreammaster: unfortunately, this has conflicts now. Could you please resolve them?

dreammaster:
Sure. I'll look into it later today

PR #2053, opened 2020-02-11
SCI32: fix bug #9752 (for all games)

ZvikaZ:
Some SCI32 games have an issue that the saved games thumbnails is useless because it's hidden
by the configuration panel (QFG4) or by the keyboard (Shivers).
Maybe this problems appears on other SCI32 games as well, should be checked.

We avoid that by keeping the thumbnail everytime the menu bar is hidden,
and when time comes to save the game, we use that thumbnail instead of taking
a new one (which would be with the control panel/keyboard).

This fix is for QfG4 only.

### DIFF
--- a/engines/sci/engine/kfile.cpp
+++ b/engines/sci/engine/kfile.cpp
@@ -1331,6 +1331,11 @@ reg_t kGetSaveFiles(EngineState *s, int argc, reg_t *argv) {
 #ifdef ENABLE_SCI32
 
 reg_t kSaveGame32(EngineState *s, int argc, reg_t *argv) {
+	// fix bug #9752 - make sure that control panel (in case of QFG4),
+	// keyboard (in case of Shivers), or any other obstacle (in other unknown SCI32 games...)
+	// will be hidden before saving
+	kFrameOut(s, 0, NULL);
+
 	const Common::String gameName = s->_segMan->getString(argv[0]);
 	int16 saveNo = argv[1].toSint16();
 	Common::String saveDescription = argv[2].isNull() ? "" : s->_segMan->getString(argv[2]);


tsoliman:
Currently the bug doesn't exist if you save by pressing F5 to save since that doesn't bring up the control panel.

How does this behave if I press F5 to save the game or using the GMM or some other way that doesn't involve the menu bar hide action?

I'm guessing a stale image will be used from when the menu bar was last hidden. That would be actually worse IMO as it is silently broken.

If that's the case, then a better solution might be a script patch for the save game handing script to hide the control panel just before saving (since SCI32 saving in ScummVM seem to be script driven and not like SCI16 which explicitly hides the menu before saving)

tsoliman:
I've actually verified that this approach introduces a new bug when saving with F5.

To reproduce:
- Move the mouse to show the menubar
- Leave the room
- Press F5 to save
- Check out the thumbnail and find that it is of the old room

The fact that the thumbnail is silently wrong makes this new bug much worse IMO (the old thumbnail was of the correct viewport state, just with a dialog on it, whereas this one is of the last time you happened to show the menubar)

ZvikaZ:
Nice catch @tsoliman , thanks for noticing this bug...
Investigating your idea now.

ZvikaZ:
@tsoliman , I've implemented your idea, thanks. It's simpler and better, and we also avoid adding a new kernel command.

tsoliman:
This seems to work in the one version of QFG4 I tested with (the GOG version)

I'm looking forward to feedback on the script code by an SCI dev.

Good work! I'm excited that this is finally getting fixed.

tsoliman:
I've tested this with both the GOG version and the Anthology CD version. Both of these are CD versions.

It's likely that this script patch will work on the standalone CD version.
It's unlikely that this script patch will work on the floppy version.

It might be safer to mark this as a QFG4CD patch rather than a generic QFG4 patch unless you've tested the floppy.

ZvikaZ:
You were correct, it indeed didn't work on floppy version.
But now I've added support for floppy as well ;-)

ZvikaZ:
If it's possible to consolidate the CD and Floppy patches to one patch with some magic, I'd be happy if someone will explain how to.

ZvikaZ:
I think I have a better idea, let's put it on hold.

ZvikaZ:
I'm getting concise from commit to commit...
Now this PR contains only one additional line:
`kFrameOut` at the beginning of `kSaveGame32`.

This solves the problem for QFG4 CD/Floppy and Shivers, and probably for other games, if this bug exists for other games (as @csnover suspected at [bug #9752](https://bugs.scummvm.org/ticket/9752) )

I could have put that line under `if` and make it run only for QFG4 and Shivers; however, I don't think that it makes any harm, and it makes sure that we won't encounter this problem in any other game.

However, I would like some SCI expert (@sluicebox ?) to make sure that it's OK.

bluegr:
Excellent work! I've tried this with QFG4, Shivers and SQ6, among others, and they all work correctly. The fix is quite straightforward, and the issue is resolved. Thanks!

Squashing + merging (to change the description somewhat)

ZvikaZ:
Thanks. But you had a typo in the commit header, in the bug number fixed by this...

sluicebox:
Sorry I'm so late the party! Everything happens while on vacation...

@ZvikaZ, I am amazed that you fixed this with a single call to FrameOut, I know others have looked at this and I was pessimistic it could be resolved at all. Great job!

PR #2054, opened 2020-02-11
TESTBED: Set the NOLAUNCHLOAD GUI option

henke37:
This continues the work of #2042 to provide correct metadata for the `testbed` "engine".


sev-:
Thanks a lot!

PR #2056, opened 2020-02-12
DIRECTOR: LINGO: Fix compiler warning

jepael:
Fixes a compiler warning about comparing signed and unsigned variables


sev-:
Thanks!

PR #2057, opened 2020-02-12
SCI: debugger: disasm - added 'bcc' parameter, to use in C code

ZvikaZ:
`bcc` is based on `bc`, but prints in format that's can be used in
C code for patches - just copy and paste (without the address column).


bluegr:
Nice! Merging

PR #2058, opened 2020-02-12
PINK: Open local webpages from the menu bar

henke37:
While the online pages are long dead, the menu bar commands refer to local pages in the game folder.

### DIFF
--- a/engines/pink/gui.cpp
+++ b/engines/pink/gui.cpp
@@ -365,7 +365,7 @@ void PinkEngine::executeMenuCommand(uint id) {
 void PinkEngine::openLocalWebPage(const Common::String &pageName) const {
 	Common::FSNode gameFolder= Common::FSNode(ConfMan.get("path"));
 	Common::FSNode filePath = gameFolder.getChild("INSTALL").getChild(pageName);
-	Common::String fullUrl = Common::String::format("file://%s", filePath.getPath().c_str());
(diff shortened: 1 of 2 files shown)


engines/pink/gui.cpp
```diff
@@ -361,7 +363,10 @@ void PinkEngine::executeMenuCommand(uint id) {
 }
 
 void PinkEngine::openLocalWebPage(const Common::String &pageName) {
-	warning("Trying to open local webpage \"%s\", but webpage opening isn't implemented yet",pageName.c_str());
+	auto gameFolder= Common::FSNode(ConfMan.get("path"));
```
dreammaster: We don't yet allow for 'auto' variables in the codebase. Please change these to explicit types

dreammaster: I'm also concerned whether just sticking a file:// in front of the filename is portable across all browsers, particularly for Windows. What system did you test the code on?

henke37: I test on Windows.

dreammaster: That makes me a bit happier. Though according to
http://help.collab.net/index.jsp?topic=/faq/winletterdrive.html
The proper formatting for a Window path/filename is like the following:
file:///d:/some/path/to/repos/on/d/drive.txt

So just putting a file:// in front of the standard path/filename getPath() returns may not be portable for all possible browsers on Windows.

sev-:
Thanks!

PR #2060, opened 2020-02-13
PINK: Use existing member variable

henke37:
Using an argument for what is already in the class definition is pointless, especially when there is no chance of an alternative value being used.


bluegr:
Good catch. Merging

PR #2061, opened 2020-02-13
PINK: Mark stuff const

henke37:
Const correctness for a bit of everything in the engine.

Notably `InventoryMgr::isPinkOwnsAnyItems` is not const, because it has a side effect that will cause a crash if removed.


sev-:
Thanks, merging

PR #2063, opened 2020-02-15
SCUMM: Tweak script dumping file management

henke37:
A separate folder per game makes it impossible to mix up scripts from different games.
And add a warning when the file couldn't be opened so that the user is told of the failure.


engines/scumm/resource.cpp
```diff
@@ -1393,11 +1393,13 @@ void ScummEngine::dumpResource(const char *tag, int id, const byte *ptr, int len
 	else
 		size = READ_BE_UINT32(ptr + 4);
 
-	sprintf(buf, "dumps/%s%d.dmp", tag, id);
+	sprintf(buf, "dumps/%s/%s%d.dmp", _game.gameid, tag, id);
 
-	out.open(buf);
-	if (out.isOpen() == false)
+	out.open(buf, true);
+	if(out.isOpen() == false) {
```
sev-: Bad formatting.

sev-:
Well, this is suboptimal. As you would always need to create two directories. Why not move htm manually after each scummvm run?

henke37:
The "having to create the folder yourself" part was already addressed with `34a40e6`.

sev-:
Good. Then, we need to change README, the built-in help, and fix the formatting. Then we're good to go

henke37:
You know what? I might as well clean up all usages of `DumpFile` in one sweep.

sev-:
Yes, cleaning up usages in more engines would bring this to consistency.

Also, the README.md file is still unchanged.

bluegr:
Any news on this?

henke37:
The entire `DumpFile` class belongs in the dumpster. The only value it adds is creating folders for you, which should be added as a feature to the `File` class. There is nothing to salvage here.

sev-:
You simply did not pay enough attention to the reason why our `File` class is read-only and `DumpFile` is the only write-capable file class.

PR #2064, opened 2020-02-15
DIRECTOR: LINGO: Debug print before string fix

rvanlaar:
Debug print statement that ran before the stack showed 'after', should be before.


sev-:
Thanks, merging.

PR #2065, opened 2020-02-15
DIRECTOR: LINGO: Implement ARRAY variable

rvanlaar:
Arrays can be:
- assigned to variables and 
- those variables can be fetched.

Includes implementation to `put`, i.e. print the array.

### DIFF
--- a/engines/director/lingo/lingo-code.cpp
+++ b/engines/director/lingo/lingo-code.cpp
@@ -226,6 +226,9 @@ void LC::c_printtop(void) {
 	case OBJECT:
 		warning("#%s", d.u.s->c_str());
 		break;
+	case ARRAY:
+		warning("ARRAY: %s", d.toString()->c_str());
+		break;
 	default:
 		warning("--unknown--");
 	}
@@ -373,7 +376,7 @@ void LC::c_assign() {
 }
 
 bool LC::verify(Symbol *s) {
-	if (s->type != INT && s->type != VOID && s->type != FLOAT && s->type != STRING && s->type != POINT && s->type != SYMBOL) {
+	if (s->type != INT && s->type != VOID && s->type != FLOAT && s->type != STRING && s->type != POINT && s->type != SYMBOL && s->type != ARRAY) {
 		warning("attempt to evaluate non-variable '%s'", s->name.c_str());
 
 		return false;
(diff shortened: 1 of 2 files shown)


engines/director/lingo/lingo-code.cpp
```diff
@@ -226,6 +226,9 @@ void LC::c_printtop(void) {
 	case OBJECT:
 		warning("#%s", d.u.s->c_str());
 		break;
+	case ARRAY:
+		warning("ARRAY: %s", d.toString()->c_str());
```
sev-: That should be just toString(). The thing is that I am not sure that I implemented toString() for arrays correctly. Somebody needs to test that. And here it should produce the output as it is now, e.g. "[<el1>, <el2>. ..]"

engines/director/lingo/lingo-codegen.cpp
```diff
@@ -527,6 +527,8 @@ void Lingo::varAssign(Datum &var, Datum &value) {
 			sym->u.s = value.u.s;
 		} else if (value.type == VOID) {
 			sym->u.i = 0;
+		} else if (value.type == ARRAY) {
+			sym->u.farr = new DatumArray(*value.u.farr);
```
sev-: are you sure we need to copy the whole thing and not just pass the reference here?

engines/director/lingo/lingo-code.cpp
rvanlaar: This array: ["one", "two", "three"] is printed as: WARNING: ARRAY: [one, two, three]! 
Without the `->c_str()` part its: WARNING: ARRAY: !

The `toString()` works correctly when showing the debug stack after message: 
Stack after: <[one, two, three]>

What would you like to see? A re-implementation of toString() for Arrays, or a custome print function in this code?

engines/director/lingo/lingo-codegen.cpp
rvanlaar: Pass by reference works.

engines/director/lingo/lingo-code.cpp
sev-: Sorry for the created cofusion. I meant:
```
warning("%s", d.toString()->c_str());
```

sev-:
Thanks!

PR #2066, opened 2020-02-15
COMMON: Add find() method for character to U32String

lephilousophe:
This is like strchr but for uint32 based strings


sev-:
Just commit things like this directly. Thanks!

PR #2067, opened 2020-02-15
ENGINES: Fix incoherent buffer size

lephilousophe:
Buffer is too small compared to what snprintf tries to put in it.


sev-:
Ugh, nice catch!

PR #2068, opened 2020-02-15
DIRECTOR: LINGO: Implement repeat with var in array

rvanlaar:
Lingo implements the for each loop as `repeat with VAR in ARRAY`.
    1) The `ARRAY` is evaluated,
    2) a loop counter is created of size(ARRAY),
    3) consume the loop counter, starting at 0.
    3) the ARRAY[counter] item is assiged to `VAR` and
    4) the loop body executed,
    5) update the loop counter with + 1 and
    6) goto 3 till the loop counter is exhausted.


sev-:
Great work!

PR #2069, opened 2020-02-15
DIRECTOR: LINGO: Fix compiler warning

jepael:
Fix compiler warning about comparison between signed and unsigned variables


sev-:
Nice catch.

PR #2070, opened 2020-02-16
SCI: fix MasterVolume to modify music volume, fixing #11012 for KQ6

ZvikaZ:
Issue #11012 reports that Music volume isn't reduced when characters are speaking.
But it's worse than that. The "Volume" slider didn't affect music at all, for GM.

This is fixing the latter issue, and fixes also original #11012 for KQ6.
Strangely, it doesn't fix KQ7 as well...
It has yet to be investigated.

### DIFF
--- a/engines/sci/sound/drivers/midi.cpp
+++ b/engines/sci/sound/drivers/midi.cpp
@@ -469,9 +469,18 @@ void MidiPlayer_Midi::setVolume(byte volume) {
 	if (!_playSwitch)
 		return;
 
-	for (uint i = 1; i < 10; i++) {
-		if (_channels[i].volume != 0xff)
-			controlChange(i, 0x07, _channels[i].volume & 0x7f);
+	if (_mt32Type == kMt32TypeNone) {
+		// the per channel volume change doesn't work for GM - therefore bug #11012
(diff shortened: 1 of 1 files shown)


ZvikaZ:
Well, on another check, I think it improved KQ7 as well, maybe even solved it.
I also think that it's solving [issue #9578](https://bugs.scummvm.org/ticket/9578).

I'd be happy if someone can verify me on these two...

waltervn:
This probably warrants a closer look

sluicebox:
I can't speak to this code change, but for context, the Narrator class is the script responsible for lowering the volume during speaking and then restoring it. (At least in KQ6 and LB2)

There does seem to be a problem with setting the volume when using General MIDI. It has an effect, but it isn't immediate. In KQ6 and LB2 if I change the volume from the game's control panel while music is playing it doesn't change the volume of the current music but it does affect music that starts afterwards, such as when changing rooms.

I wonder how much it's worth going after these MIDI things in the current codebase versus trying to dust off and integrate the MIDI work csnover did, which might have already taken care of them. (I don't know anything about the state of that)

ZvikaZ:
Let's put this PR on hold, as my new PR #2074 makes this redundant.
If we'll decide to not accept #2074, let's get back to this one.

PR #2071, opened 2020-02-16
Fix building when selected engines don't have their own debugger

lephilousophe:
In this case, the linker optimizes out Debugger which get required
later by Engine due to @dreammaster rework. Building with all engines
prevents optimization because several engines subclass Debugger and they
are included before gui.a in linker command line


bluegr:
Nice catch, thanks! Merging

PR #2072, opened 2020-02-16
3DS: Improve cursor behavior in Drag Mode

BallM4788:
Some games such as Riven perform an action when a mouse button is pressed but update the cursor when the mouse button is released. As a result, the way touch controls are coded in the 3DS port can result in the cursor image updating too late. Continuing with the Riven example, performing a touch screen click on a hotspot in Drag Mode will not update the cursor image until after the scene transition completes, when the cursor update should occur BEFORE the scene transition. In Hover Mode, however, the cursor updates before the scene transition as desired.

The reason for this is that, in Drag Mode, a touch screen press immediately sends an EVENT_LBUTTONDOWN event, while an EVENT_LBUTTONUP event is only sent once the touch screen is released. This allows the Riven subengine of Mohawk to schedule a scene transition between the EVENT_LBUTTONDOWN and EVENT_LBUTTONUP events, thus causing the cursor update to happen after the transition has occurred.

This commit makes changes to how touch screen events happen in Drag Mode. Instead of immediately sending an EVENT_LBUTTONDOWN when the touch screen is pressed, a 50 millisecond grace period is set in place which allows the backend to differentiate between clicks and drags. If the touch screen is released during the 50ms period, the touch is registered as a click and a 3-event sequence is sent (MOUSEMOVE, LBUTTONDOWN, and LBUTTONUP), not allowing game engines to interrupt that sequence. If the touch screen is still being held once the 50ms has elapsed, the touch is registered as a drag, and an EVENT_LBUTTONDOWN is sent, with an EVENT_LBUTTONUP sent when the touch screen is released.

### DIFF
--- a/backends/platform/3ds/osystem-events.cpp
+++ b/backends/platform/3ds/osystem-events.cpp
@@ -159,7 +159,7 @@ static void eventThreadFunc(void *arg) {
 				pushEventQueue(eventQueue, event);
 			}
 
-			if ((inputMode == MODE_DRAG) & !isDragHeld) {
+			if ((inputMode == MODE_DRAG) & (!isDragHeld)) {
 				if (osys->getMillis() - touchStartTime >= 50) {
 					isDragHeld = true;
 					event.type = Common::EVENT_LBUTTONDOWN;
--- a/backends/platform/3ds/osystem-events.cpp
+++ b/backends/platform/3ds/osystem-events.cpp
@@ -159,7 +159,7 @@ static void eventThreadFunc(void *arg) {
 				pushEventQueue(eventQueue, event);
 			}
 
-			if (inputMode == MODE_DRAG & !isDragHeld) {
+			if ((inputMode == MODE_DRAG) & !isDragHeld) {
 				if (osys->getMillis() - touchStartTime >= 50) {
 					isDragHeld = true;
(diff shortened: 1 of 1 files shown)


bgK:
I'd need some more time to think about the 3DS changes, but it seems to me the Riven engine needs to be fixed. It has the information the mouse has moved and should update the cursor before / when performing the scene transition. I want Riven to work well for all the backends regardless of their mouse event quirks.

BallM4788:
I might just be dense, but I've had no luck fixing the issue from within the Riven engine. In addition, I've discovered that what I did here in the 3DS backend doesn't actually solve it either; when making a quick, very precise tap, the errant behavior seen in the master branch occurs.

I'll continue to try finding a solution, but you'd probably have much better chances of doing so, given your much greater understanding of the Mohawk engine's inner workings.

BallM4788:
@bgK Figured out how to fix the cursor quirk in the Riven engine, turns out I was indeed being very dense. Submitted a PR for it here: #2129, let me know if I should close this one.

bgK:
Ok, thanks. In PR #2110, I've changed a bit the behavior of the mouse cursor in the 3DS backend so that the cursor only moves once the engine has seen the corresponding mouse event. That way it matches the other backends. It helps with this issue. I'll take a look at your Riven PR a bit later. Closing this.

PR #2073, opened 2020-02-16
SCI: fix bug #9735 by avoiding remap to percussion channel

ZvikaZ:
Channel 9 (in the code, which starts counting from 0; and it's channel 10 in the MIDI spec, which starts counting from 1) is reserved for percussion instruments.
Fixing remapping channel code to avoid remapping to that channel.

This is fixing bug #9735, and probably other sound issues.


ZvikaZ:
This solves also bug #10297

ZvikaZ:
Let's put this PR on hold, as my new PR #2074 makes this redundant.
If we'll decide to not accept #2074, let's get back to this one.

PR #2074, opened 2020-02-17
SCI: using only channels 1-8 on GM devices

ZvikaZ:
There are many situations that GM sounds worse than MT, even for games that were written for GM.
I think this is related to this.

And, there is a note in midi.cpp, added by Walter van Niftrik on his commit from 2010:
```// NOTE: SSCI uses channels 1 through 8 for General MIDI as well, in the drivers I checked```
Thus, I took @waltervn's note and made GM use the same channels as MT-32.

Also, issue #6686 reports that using the "higher" channels causes problem on a real SC-55.

This change solves at least bugs: #6686, #9578, #9735, #10297 and #11012.

Notes:
- I have verified all these bugs, except for #6686, as I don't have SC-55
- On PR #2070 , @sluicebox mentioned that the master volume is changing, but slowly. That's the cause of issues #9578 and #11012 , so it's better to have it fixed.

If we accept this PR, it makes my other PRs: #2070 and #2073 redundant, as it fixes the same issues, and more, with simpler solution.

### DIFF
--- a/engines/sci/sound/drivers/midi.cpp
+++ b/engines/sci/sound/drivers/midi.cpp
@@ -452,15 +452,11 @@ void MidiPlayer_Midi::send(uint32 b) {
 // We return 1 for mt32, because if we remap channels to 0 for mt32, those won't get played at all
 // NOTE: SSCI uses channels 1 through 8 for General MIDI as well, in the drivers I checked
 int MidiPlayer_Midi::getFirstChannel() const {
-	if (_mt32Type != kMt32TypeNone)
-		return 1;
-	return 0;
+	return 1;
 }
 
 int MidiPlayer_Midi::getLastChannel() const {
-	if (_mt32Type != kMt32TypeNone)
-		return 8;
-	return 15;
+	return 8;
 }
 
 void MidiPlayer_Midi::setVolume(byte volume) {


bluegr:
I don't understand the logic behind this.

The MT-32 had [8 melodic and 1 rhythm channel](https://en.wikipedia.org/wiki/Roland_MT-32#Features), whereas General MIDI includes support for [15 melodic and 1 percussion channel](https://en.wikipedia.org/wiki/General_MIDI#General_MIDI_1_Requirements).

With your commit, you're essentially killing half of the GM channels. How does that make the music sound better?

bluegr:
And here's an example: this is track 120 in Leisure Suit Larry 6, which is the theme music. The MIDI track uses channels 2, 3, 4, 5, 10 and 16

image

ZvikaZ:
Well, @bluegr, that's a good question that I can't give good answer for...
However, I claim that it's better because:
- It solves 5 bugs, that actually cover 3 different topics (volume doesn't take affect, missing sounds, wrong sounds) in different games, which gives an impression of random instability. All these bugs are GM only, and working well for MT32.
- As @waltervn noted in the code, that's also how Sierra originally implemented it, therefore, they don't expect to use more than 8 channels in parallel. Thus, I assume that it makes no harm.
- Indeed, I'd be very glad if someone can show me a place that sounds worse with this commit.

ZvikaZ:
My previous comment was answer to your first question.
Regarding the second one, the channel numbers in the resource file have no significance, as they are replaced by the remap mechanism.
That track is using 6 channels. Therefore, it'd be mapped to channels 3-8, leaving tracks 1,2 free. Is there a scenario that we need more than that?

ZvikaZ:
I think I had a small mistake in my channels explanation. In the LSL track, one of the channels is 10, which is reserved for percussion (note, I spoke of channels as they are in the code, starting from 0, in SCICompanion they are starting from 1, as defined in the MIDI spec), and shouldn't be remapped.
Therefore, the track is supposed to use channels 4-9 (9 is "10"), leaving 1,2,3 free.
(in other words, I claim that I've killed less than half the channels :-) )

But I'm not near computer yo verify myself.

waltervn:
As an aside, I do want to mention that channel 10 being reserved for percussion is part of the **General** MIDI specification, not the original MIDI specification.

ZvikaZ:
What's more interesting, I've just discovered that even before my commit, we're not using all the 16 channels!
I've checked GK1. Its main theme, called in the game's scripts `gkMusic1`, is sound resource #370, and shown at SCICompanion to use channels 2,3,4,5,6,7,9,10 (starting count from 1) for GM - i.e., 7 melodic channels and 1 percussion channel.
However, with `define DEBUG_REMAP`, I get the following output (again, before my changes):

	 Song 0 (08EA2FC8) [gkMusic1], prio 100
	  Channel 1: prio 0, 1 voice
	   trying to map to 15
		OK
	  Channel 2: prio 0, 1 voice
	   trying to map to 14
		OK
	  Channel 3: prio 0, 1 voice
	   trying to map to 13
		OK
	  Channel 4: prio 0, 1 voice
	   trying to map to 12
		OK
	  Channel 5: prio 0, 1 voice
	   trying to map to 11
		OK
	  Channel 6: prio 0, 1 voice
	   trying to map to 10
		OK
	  Channel 9: prio 0, 1 voice, dontRemap
	   trying to map to 9
		OK
	Remap results:
	 Unmapping song 0, channel 15
	 Mapping (dontRemap) song 0, channel 9 to device channel 9
	 Mapping song 0, channel 6 to device channel 15
	 Mapping song 0, channel 5 to device channel 14
	 Mapping song 0, channel 4 to device channel 13
	 Mapping song 0, channel 3 to device channel 12
	 Mapping song 0, channel 2 to device channel 11
	 Mapping song 0, channel 1 to device channel 10

I see here the following interesting facts:

1. Somehow, the original 7+1 channels got reduced to 5+1.
2. The code knows to keep original percussion channel.
3. Beside percussion channel, the code doesn't care about original track's channel numbers.

If we're at it, let me show the problematic Score Jingle sound dump:

		Remap: avail chans: 0-15
		 Song 0 (08EA37E0) [gkSound3], prio 100
		  Channel 1: prio 0, 0 voices
		   trying to map to 15
			OK
		 Song 1 (08EA3908) [gkMusic1], prio 100
		  Channel 1: prio 0, 1 voice
		   trying to map to 14
			OK
		  Channel 3: prio 0, 1 voice
		   trying to map to 13
			OK
		  Channel 4: prio 0, 1 voice
		   trying to map to 12
			OK
		  Channel 6: prio 0, 1 voice
		   trying to map to 11
			OK
		  Channel 2: prio 2, 1 voice
		   trying to map to 10
			OK
		  Channel 5: prio 1, 1 voice
		   trying to map to 9
			OK
		Remap results:
		 Keeping song 1, channel 5 on device channel 15
		 Keeping song 1, channel 2 on device channel 14
		 Keeping song 1, channel 6 on device channel 13
		 Keeping song 1, channel 4 on device channel 12
		 Keeping song 1, channel 3 on device channel 11
		 Keeping song 1, channel 1 on device channel 10
		 Mapping song 0, channel 1 to device channel 9

`Song 1` is the older song - gkMusic1- the game's theme - sound #370.
`Song 0` is the new song - gkSound3 - the score jingle - sound #10. 
Sound #10 is originally using channel #2, but it got mapped to channel #9 (=percussion) , and therefore, didn't play.

BTW, if we play gkSound3 from inside debugger, it play well - because nothing else is playing, all channels are free, and it get the first free channel - 15 - and therefore, no problems.

Bottom line of the long comment
-------------------------------------
- In the current situation, we already don't enjoy all the 16 channels, there seem to be some code out there that drops out some channels if there are too many...
- That matches @athrxx note on Discord, that the remapping code behaves like Sierra's code (and as we said before, originally Sierra used only channels 1-8), and bad things might happen if I change the original behaviour.
- These bad things already happened! See the list of open bugs I wrote above.
- Therefore, my suggested fix shouldn't cause any harm, only return to the original state.
- In the future, after things will stabilize, we might want to write new code, that will really use all 16 possible channels - and then we'll undo my changes here, but will make sure that:
** No channels are "thrown out" (like GK's gkMusic1 example I just gave)
** There is no regression in the issues I've supplied. They can serve as nice "unit test" for anyone trying to improve music.

waltervn:
I don't know why there's one less channel for that song compared to what SCI Companion says, but it seems to be a bit premature to call this a bug. We'd have to look into that a bit closer.

But anyway, it does indeed appear that increasing the last channel for GM from 8 to 15 is incompatible with the channel remapper.

Personally, I never liked this change to begin with, mainly because the games were not designed and tested playing on that many channels.

athrxx:
I have just taken another look at the KQ6 and the QFG4 genmidi.drv drivers, since the details had left my memory.

The drivers do report a first channel of 1, a last channel of 8 and a polyphony of 32. So this change seems to be right, at least for the two drivers I checked.

r-lind:
Being the one who first started making some GM translation layer in FreeSCI... I do remember (if I'm not mistaken) that you could reprogram the MT-32 to begin with channel 1 on a hardware level.

r-lind:
> There are many situations that GM sounds worse than MT, even for games that were written for GM.
Some games were during a transition period written for MT-32 and used a GM translation layer with instrument mapping and 4 different velocity envelopes (which is pretty cost effective), so even though they support GM they were made for MT-32 and it should be the preferred device driver. Perhaps a flag in detection_tables.h stating this would be good?

ZvikaZ:
> There are many situations that GM sounds worse than MT, even for games that were written for GM.

I wasn't clear enough. In this sentence I meant ` GM sounds worse than MT` on ScummVM - not on original Sierra's interpreter.
For example - GK1 missing score jingle (https://bugs.scummvm.org/ticket/9735) , QFG1VGA playing whistles instead of guitars on Erana's Peace (https://bugs.scummvm.org/ticket/10297) or music volume not being decreased when characters are speaking (https://bugs.scummvm.org/ticket/11012).

All these problems are GM only, and don't exist on MT-32. But they don't exist on DosBOX as well. They exist only on ScummVM GM.

This PR fixes all of them, by returning to the original implementation.

After we'll accept this PR, and return stability, if someone is willing, it's possible to try other changes and improvements. But I think that we should first incorporate this PR and fix all these issues.

ZvikaZ:
> Some games were during a transition period written for MT-32 and used a GM translation layer with instrument mapping and 4 different velocity envelopes (which is pretty cost effective), so even though they support GM they were made for MT-32 and it should be the preferred device driver. Perhaps a flag in detection_tables.h stating this would be good?

I agree with that - it's a good idea. But it should be done as a separate PR.

bluegr:
Since this change matches the original behaviour, it should be merged, and the other two pull requests will be closed. Thanks for working on this! Merging

PR #2075, opened 2020-02-17
BLADERUNNER: Set 'ditheringFactor' as static const object.

carlo-bramini:
In my opinion, there is no need to initialize this array every time drawShadowPolygon() is called, so I would suggest to declare it as "static const".


antoniou79:
Hi! The change looks good to me and it makes sense. Well spotted!
However, to keep with the commit guidelines the commit message has to include (ie. start with) the subsystem name (in this case it should be "BLADERUNNER:"). Could you fix this?

dreammaster:
You changed the title of the pull request, but not the commit itself, which is the important one

PR #2076, opened 2020-02-17
DIRECTOR: LINGO: handle break from loops

rvanlaar:
Lingo has two ways to break from loops: `return` and `exit repeat`.
Both are now handled in repeat with VAR in ARRAY.
Example Lingo code:

```
on exitRepeat
  set aList = [1,2,3,4]
  repeat with a in aList
    if a = 3 then
      exit repeat
    end if
  end repeat  
  return a
end exitRepeat

on returnRepeat
  set aList = [1,2,3,4]
  repeat with a in aList
    if a = 3 then
      return a
    end if
  end repeat  
end return

put exitRepeat()
put returnRepeat()
```


sev-:
Looks fine, merging. However, we would better add tests for all of this. Could you please commit it directly?

PR #2077, opened 2020-02-18
BUILD: Apply LDFLAGS on modules/plugins

besser82:
Some distributions populate their build environment with special linker flags, e.g. for hardening purposes such as immediate binding and/or read-only relocations.

Those flags should be honored when linking dynamic libraries.

### DIFF
--- a/rules.mk
+++ b/rules.mk
@@ -47,7 +47,7 @@ ifdef PLUGIN
 PLUGIN-$(MODULE) := plugins/$(PLUGIN_PREFIX)$(notdir $(MODULE))$(PLUGIN_SUFFIX)
 $(PLUGIN-$(MODULE)): $(MODULE_OBJS-$(MODULE)) $(PLUGIN_EXTRA_DEPS)
 	$(QUIET)$(MKDIR) plugins
-	$(QUIET_PLUGIN)$(CXX) $(filter-out $(PLUGIN_EXTRA_DEPS),$+) $(PLUGIN_LDFLAGS) -o $@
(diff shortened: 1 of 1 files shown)


sev-:
Could you please be a bit more specific, which distribution do that, and if possible provide some links where we can read more info about this hardening.

besser82:
The specific distributions are Fedora and RHEL/CentOS.

When building a rpm package for those distributions `-Wl,-z,relro -Wl,-z,now` is passed to the linker using the `LDFLAGS` environment variable, to achieve binary hardening.

From the [GCC `ld` documentation](https://linux.die.net/man/1/ld):

  * *relro*: Create an ELF "PT_GNU_RELRO" segment header in the object.
  * *now*: When generating an executable or shared library, mark it to tell the dynamic linker to resolve all symbols when the program is started, or when the shared library is linked to using dlopen, instead of deferring function call resolution to the point when the function is first called.

Further information about binary hardening can be found on the [redhat blog](https://www.redhat.com/en/blog/hardening-elf-binaries-using-relocation-read-only-relro).

sev-:
Thanks for the explanation. merging.

sev-:
I reverted it for the time being, because it broke tons of buildbot ports:

https://buildbot.scummvm.org/builders/master-gamecube/builds/3318
https://buildbot.scummvm.org/builders/master-ps2/builds/3320
https://buildbot.scummvm.org/builders/master-psp/builds/3319
https://buildbot.scummvm.org/builders/master-pspfull/builds/1959
https://buildbot.scummvm.org/builders/master-wii/builds/3314

sev-:
If you could figure it out, that would be quite helpful.

besser82:
@sev- A fixed version is in https://github.com/scummvm/scummvm/pull/2082.

PR #2078, opened 2020-02-19
ENGINES: Show the keymaps tab in the in-game options dialog

bgK:
This adds the keymaps tab to the in-game options dialog accessed through the GMM, making it possible to remap keys without exiting the game.

At the moment, only the game keymaps are shown. I'm not sure whether to show the global keymaps as well. The issue is the global keymaps are saved globally whereas the game keymaps are saved per target. Thoughts?

scummvm-riven-00000

scummvm-dig-00000


dreammaster:
I concur that it's best not to show global keymaps here. It would only cause confusion, particularly since everything else in the options dialog is game specific.

bluegr:
I like it, and also agree with @dreammaster

dreammaster:
Did some testing, and it all looks good

gui/ThemeEngine.cpp
```diff
@@ -876,6 +876,10 @@ void ThemeEngine::drawDD(DrawData type, const Common::Rect &r, uint32 dynamic, b
 		extendedRect.clip(_clip);
 	}
 
+	if (extendedRect.isEmpty()) {
+		return;
+	}
```
bluegr: Under what circumstances could this occur?

gui/ThemeParser.cpp
```diff
@@ -740,7 +740,18 @@ bool ThemeParser::parserCallback_dialog(ParserNode *node) {
 			return false;
 	}
 
-	_theme->getEvaluator()->addDialog(name, node->values["overlays"], enabled, inset);
+	Common::String overlays = node->values["overlays"];
+
+	int width = -1, height = -1;
+	if (node->values.contains("size")) {
+		if (overlays != "screen_center")
+			return parserError("Dialogs can only have an explicit size if they overlay 'screen_center'.");
```
bluegr: Could you add a comment here, explaining why?

bgK: Okay.

gui/ThemeEngine.cpp
bgK: When the scroll container widget is scrolled down, some the widgets it contains are outside of the visible area.

bluegr:
Many thanks for your work, and the explanations!

This is good to be merged now

PR #2079, opened 2020-02-21
CONFIGURE: Add compatibility for RPM's configure macro

besser82:
RPM-based distributions come with a configure macro, that sets up the whole build environment.  In order to be able to use this macro, some options, which are valid for the configure script generated by GNU Autotools, should not error-out when preparing the build stage.

Unfortunately the configure script, as used by ScummVM uses the `--host` option to determine the target system it will actually be build for.  Autotools based configure scripts have a `--target` option for such a purpose, and use the `--host` option to determine the system the build is performed on.

For that reason there should be a way to discard the parameters passed to the configure script with the `--host` option.  The easiest approach to achieve this goal, is to have an environment variable, which when set influences the configure script to ignore the parameters of the `--host` option.

Thus we introduced a variable called `CONFIGURE_NO_HOST`, that will when set to anything, but an empty value, before invoking the configure script, have the parameters of the `--host` option take no influence on the configure stage (and the build stage as well).

### DIFF
--- a/configure
+++ b/configure
@@ -1192,6 +1192,19 @@ EOF
 
 for ac_option in $@; do
 	case "$ac_option" in
+	# Silently ignore options valid for Autotools configure.
+	--build=*)                                           ;;
+	--program-prefix=*)                                  ;;
+	--sbindir=*)                                         ;;
+	--sysconfdir=*)                                      ;;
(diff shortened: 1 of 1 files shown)


sev-:
Thanks, merging.

PR #2080, opened 2020-02-22
ULTIMA8: Fix blends to SDL surface (corrects lamps, smoke, etc)

mduggan:
As part of the porting work from Pentagram, the format masks used during blend
operations are not set correctly, resulting in grey shadows under lamps (rather
than lighting the underlying surface).  This sets the masks correctly so lamps
now look right.

attn @dreammaster to save you trying to find this.


dreammaster:
Good catch. Merging it in now

PR #2082, opened 2020-02-22
BUILD: Use unmodified SAVED_LDFLAGS from env for linking plugins.

besser82:
In commit dce6b012 we added the `LDFLAGS`, which were modified by the configure script, to the linker stage of the plugins.  Doing so turns out not to work well for all system architectures.

This approach uses the unmodified `LDFLAGS`, stored in the `SAVED_LDFLAGS` variable by the configure script, from the system environment for linking dynamic plugins, which is very likely not to cause any harm.

See https://github.com/scummvm/scummvm/pull/2077 for further reference.


sev-:
okay, let's try it.

PR #2083, opened 2020-02-23
DIRECTOR: Set the backgroundcolor

rvanlaar:
The background color is called stageColor in Director.
Director uses a palette with 256 colors.

The colors in the Palette have an offset of 255.


sev-:
Cool, thanks!

PR #2084, opened 2020-02-24
COMMON: Fix conversion to UTF32 for short strings

lolbot-iichan:
Testcase: convert 1-byte encoded single character string (e.g. "=") from 125X to
UTF32.

Expected result: got widestring, first dword contains "=", next dword
contains "\0"

Actual result: got widestring, first dword contains "=", second dword
contains garbage, next dword contains "\0"

Reference implementation:
http://www-personal.umich.edu/~bazald/l/api/_s_d_l__iconv_8c_source.html
, line 887


sev-:
Thanks! Merging.

PR #2088, opened 2020-02-24
ULTIMA8: Fix buggy UCList.remove

mduggan:
The code for `UCList.remove` is buggy in 2 ways:
* The loop resets the value of `equal` for each byte
* The removal check is reversed, removing items which are `!equal`

In practice this function is only called from `substractList` for engine opcode `0x1B`.  The comment in `uc_engine.cpp` says `0x1B` only happens in Crusader, so this code is probably never called in U8 but it feels better to fix it anyway and match the behavior of `substractStringList`.


mduggan:
Sorry for the force-push, I originally had `equal &=` but decided I prefer `equal = equal &&` better for bools.

PR #2090, opened 2020-02-24
ULTIMA8: Clean up unused functions in istring

mduggan:
The `istring` class file and header redefine a few common string functions, but they're not used any more now that it's using the `Common::String` functions.


bluegr:
Nice catch, quite straightforward changes. Thanks for your work! Merging

PR #2091, opened 2020-02-24
AUDIO: Added dump-midi mechanism

ZvikaZ:
This mechanism is enabled by `--dump-midi` command line parameter.
The midi events are printed to screen, and dumped to 'dump.mid' file.

It doesn't handle SysEX events, which can be printed to screen by adding:
`  #define DEBUG_SYSEX`

### DIFF
--- a/audio/mididrv.h
+++ b/audio/mididrv.h
@@ -26,6 +26,7 @@
 #include "common/scummsys.h"
 #include "common/str.h"
 #include "common/timer.h"
+#include "common/array.h"
 
 class MidiChannel;
 
@@ -86,7 +87,9 @@ enum MidiDriverFlags {
  */
 class MidiDriver_BASE {
 public:
-	virtual ~MidiDriver_BASE() { }
+	MidiDriver_BASE();
+
+	virtual ~MidiDriver_BASE();
 
 	/**
 	 * Output a packed midi command to the midi stream.
@@ -96,6 +99,7 @@ public:
 	 */
 	virtual void send(uint32 b) = 0;
 
+
 	/**
 	 * Output a midi command to the midi stream. Convenience wrapper
 	 * around the usual 'packed' send method.
@@ -103,10 +107,8 @@ public:
 	 * Do NOT use this for sysEx transmission; instead, use the sysEx()
 	 * method below.
 	 */
-	void send(byte status, byte firstOp, byte secondOp) {
-		send(status | ((uint32)firstOp << 8) | ((uint32)secondOp << 16));
-	}
-
+	void send(byte status, byte firstOp, byte secondOp);
+	
 	/**
 	 * Transmit a sysEx to the midi device.
 	 *
@@ -121,6 +123,39 @@ public:
 
 	// TODO: Document this.
 	virtual void metaEvent(byte type, byte *data, uint16 length) { }
+
+protected:
+
+	/**
+	 * Enables midi dumping to a 'dump.mid' file and to debug messages on screen
+	 * It's set by '--dump-midi' command line parameter
+	 */
+	bool _midiDumpEnable;
+
+	/** Used for MIDI dumping delta calculation */
+	uint32 _prevMillis;
+
+	/** Stores all MIDI events, will be written to disk after an engine quits */
+	Common::Array<byte> _midiDumpCache;
+
+	/** Initialize midi dumping mechanism, called only if enabled */
+	void midiDumpInit();
+
+	/** Handles MIDI file variable length dumping */
+	int midiDumpVarLength(const uint32 &delta);
+
+	/** Handles MIDI file time delta dumping */
+	void midiDumpDelta();
+
+	/** Performs dumping of MIDI commands, called only if enabled */
+	void midiDumpDo(uint32 b);
+
+	/** Performs dumping of MIDI SysEx commands, called only if enabled */
+	void midiDumpSysEx(const byte *msg, uint16 length);
+
+	/** Writes the captured MIDI events to disk, called only if enabled */
+	void midiDumpFinish();
+
 };
 
 /**
@@ -165,6 +200,13 @@ public:
 
 	/** Get the device description string matching the given device handle and the given type. */
 	static Common::String getDeviceString(DeviceHandle handle, DeviceStringType type);
+
+	/** Common operations to be done by all drivers on start of send */
+	void midiDriverCommonSend(uint32 b);
+
+	/** Common operations to be done by all drivers on start of sysEx */
+	void midiDriverCommonSysEx(const byte *msg, uint16 length);
+
 
 private:
 	// If detectDevice() detects MT32 and we have a preferred MT32 device
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -131,6 +131,8 @@ static const char HELP_STRING[] =
 	"                           supported by some MIDI drivers)\n"
 	"  --multi-midi             Enable combination AdLib and native MIDI\n"
 	"  --native-mt32            True Roland MT-32 (disable GM emulation)\n"
+	"  --dump-midi              Dumps MIDI events to 'dump.mid', until quitting from game\n"
+	"                           (if file already exists, it will be overwritten)\n"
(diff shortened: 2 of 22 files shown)


ZvikaZ:
Hi.
We discussed it yesterday on Discord, and you suggester to me to do this from command line console.
I started to do it, but stopped because:

- I found it more difficult to implement
  (This change is generic, how can I reach an unknown's engine's midi driver from debugger context?)
- Starting from debugger console will miss some initial MIDI events, which might be related to the problem at hand
- I've found a solution how to it with command line parameter...

audio/mididrv.h
```diff
@@ -121,6 +123,34 @@ class MidiDriver_BASE {
...
+	 * Enables midi dumping to a 'dump.mid' file and to debug messages on screen
+	 * It's set by '--dump-midi' command line parameter
+	 *
+	 * Note that SysEx aren't dumped by this mechanism,
+	 * but then can be printed to screen by
+	 * #define DEBUG_SYSEX
+	 */
+	bool _midiDumpEnable;
+
+	/** used for midi dumping delta calculation */
```
bluegr:
```suggestion
	/** Used for MIDI dumping delta calculation */
```

audio/mididrv.h
```diff
@@ -121,6 +123,34 @@ class MidiDriver_BASE {
...
+	uint32 _prevMillis;
+
+	/** keeps all midi events, will be written only upon game quite */
+	Common::Array<byte> _midiDumpCache;
+
+	/** initialized midi dumping mechanism, called only if enabled */
```
bluegr:
```suggestion
	/** Initialize midi dumping mechanism, called only if enabled */
```

audio/mididrv.h
```diff
@@ -121,6 +123,34 @@ class MidiDriver_BASE {
...
+	void midiDumpInit();
+
+	/** does midi dumping, called only if enabled */
```
bluegr:
```suggestion
	/** Performs dumping of MIDI commands, called only if enabled */
```

audio/mididrv.h
```diff
@@ -121,6 +123,34 @@ class MidiDriver_BASE {
...
+	void midiDumpDo(const byte &status, const byte &firstOp, const byte &secondOp);
+
+	/** finises midi dumping mechanism, called only if enabled */
```
bluegr:
```suggestion
	/** Writes the captured MIDI events to disk, called only if enabled */
```

base/commandLine.cpp
```diff
@@ -131,6 +131,10 @@ static const char HELP_STRING[] =
 	"                           supported by some MIDI drivers)\n"
 	"  --multi-midi             Enable combination AdLib and native MIDI\n"
 	"  --native-mt32            True Roland MT-32 (disable GM emulation)\n"
+	"  --dump-midi              Dumps MIDI events to 'dump.mid', until quitting from game\n"
+	"                           (if file already exist, it will be overwritten)"
```
bluegr:
```suggestion
	"                           (if file already exists, it will be overwritten)"
```

base/commandLine.cpp
```diff
@@ -131,6 +131,10 @@ static const char HELP_STRING[] =
...
+	"                           Note: it doesn't record SysEX events\n"
```
bluegr:
```suggestion
	"                           SysEX events are not handled,\n"
```

base/commandLine.cpp
```diff
@@ -131,6 +131,10 @@ static const char HELP_STRING[] =
...
+	"                           therefore, MT-32 MIDI will be recorded, but sound distorted\n"
```
bluegr:
```suggestion
	"                           therefore, MT-32 MIDI will be recorded, but will sound distorted\n"
```

base/main.cpp
```diff
@@ -449,6 +449,12 @@ extern "C" int scummvm_main(int argc, const char * const argv[]) {
 		return res.getCode();
 	}
 
+	if (settings.contains("dump-midi")) {
+		// keep that command line setting (all other transient will be destroyed), without using a global variable
```
bluegr:
```suggestion
		// Store this command line setting in ConfMan, since all transient settings are destroyed
```

ZvikaZ:
I'm not a C++ expert, so maybe it's a stupid question...

Can I add code to the `actual send() function, that gets overridden in each MIDI driver` in one common place?
Wouldn't it get overridden?

Or are you talking about duplicating code to each engine's `send()` implementation?

bluegr:
@ZvikaZ: No, if you add code to a virtual function, child classes that override it will not execute the code in the base virtual function:

https://www.geeksforgeeks.org/virtual-function-cpp/

However, an idea would be to split send() into two discreet functions: a virtual one that gets overriden in each driver, as it's done now, and a wrapper function that gets called in client code, like for example, the MidiParser, and calls the virtual method and the MIDI dumping code inside it.

Same goes for the sysEx() function.

ZvikaZ:
I assumed that this is the purpose of `MidiDriver_BASE::send` - a common wrapper that calls the different `send`s.

ZvikaZ:
Well, no problem.
I will implement your suggestion.

bluegr:
Here's a PoC:

    class MidiDriver_BASE {
    public:
        /* The public method, available in all the client code */
        void send(uint32 b) {
            // Add any logic here
            sendImpl(b);
        }
        /* The virtual method, overriden in each driver */
        virtual void sendImpl(uint32 b) = 0;
    }

ZvikaZ:
I've squashed all your changes, and implemented your POC.
Pushed now my code.

It's not complete yet:
- I haven't handled SysEx
- It'll probably not compile for non SCI engines

But I'd like the changes that I've already made to be reviewed before I continue.

ZvikaZ:
Travis failed, and that's expected, as I said - I need to add support for all engines.
But still, I'd like that code that I pushed to be reviewed first, if possible.

ZvikaZ:
I've implemented the `sendImpl mechanism`. It was quite complicated, involved modifying 93 files, all over ScummVM.
It worked well for SCI games, DOTT and MI2.
But then I've tried an AGI game (SQ2) - and it was stuck.
Fixed that.
Then I've tried Loom, and the music got into endless loop.

Then I realized that this `send()` mechanism is very complicated (remember we spoke that it appears both for real drivers, and for players?), and I won't be 100% sure that I haven't inserted audio bugs until I will verify all engines, because some of them has peculiar way of handling midi send()-ing.

Therefore, I've dropped that attitude, and replaced it with something more simpler for our scenario - a common method, called by the drivers' `send()`.

audio/mididrv.h
```diff
@@ -166,6 +201,13 @@ class MidiDriver : public MidiDriver_BASE {
 	/** Get the device description string matching the given device handle and the given type. */
 	static Common::String getDeviceString(DeviceHandle handle, DeviceStringType type);
 
+	/** Common operations to be done by all drivers on start of send */
+	void sendCommonDriving(uint32 b);
```
bluegr: I would suggest a better name for this, "driving" is not associated with "driver", in this case.
midiDriverCommonSend() sounds better

audio/mididrv.h
```diff
@@ -166,6 +201,13 @@ class MidiDriver : public MidiDriver_BASE {
...
+
+	/** Common operations to be done by all drivers on start of sysEx */
+	void sysExCommonDriving(const byte *msg, uint16 length);
```
bluegr: Same as above. midiDriverCommonSysEx() sounds better

bluegr:
Very nice! This is a much cleaner approach, and fully working :)

My only leftover comment is with the naming of the two common functions.

ZvikaZ:
Done.
I've also discovered that I forgot to support fluidsynth emulation (since I compile without it...).
Added that as well (it's similar to MT32 emulation, which I've locally tested before)

bluegr:
Thanks for addressing all the remaining issues, and for rewriting this to be fully functional.

Great work, merging

ZvikaZ:
Thanks for the good feedback!
That's a good example of the virtues of code review - I think my original code was good, but because of your review it's now better...

PR #2092, opened 2020-02-24
ULTIMA8: Don't save while starting up

mduggan:
This fixes the direct loading of save games from the ScummVM menu by
blocking autosave during the startup.

There may be some more elegant way to fix this so I'm not sure if this is the nicest solution - it does the job though.


dreammaster:
Agreed. Looks good to me.

PR #2095, opened 2020-02-25
WINTERMUTE: Initial support for 3 games by HeroCraft

lolbot-iichan:
HeroCraft is a Russian game developer. They made at least 3 games with their own fork of WME:
* Papa's Daughters 1 & 2 ( two nice story-driven hidden object games themed after russian TV series, a sitcom about a divorced man and his 5 daughters )
* Pole Chudes ( word guessing and money spending simulator game themed after popular russian TV show similar to Wheel of Fortune )

Good news: all those games are now completable with this PR (with some known visual glitches though, similar to https://bugs.scummvm.org/attachment/ticket/9683/helga.gif and several other known issues, such as scrollable list viewed outside of viewport and incorrent filling of bar widgets).
Bad news: those games are not sold anymore (except for flea markets and similar places).

HeroCraft's fork of WME is based of WME Kinjal Edition 1.4, a community fork of WME 1.9.0. So, it's technically a sub-sub-engine =)

All HeroCraft games are heavily using entity reordering methods from WME Kinjal Edition 1.4, which allows to put an entity in front or behind another entity of the same layer. Those methods are fully reimplemented.

Other Kinjal Edition 1.4 features are currently not reimplemented and it's unknown if any of them are required. Kinjal Edition changelog can be found here: http://forum.dead-code.org/index.php?topic=4025.0

It seems that developers wanted to obfuscate their work: most of HeroCraft additional methods have misleading names, and a lot of code is added to crash everything if self-check fail. 

Pole Chudes requires 3 additional APIs to run:
* custom randomization methods that allows to provide a random seed. Reimplemented.
* GetCharCode() method for strings to get ord(N-th character). Reimplemented.
* GetSpriteControl() method that return some internal state enum value. No idea what it originaly did, hardcoded 44332211L to make game scripts run correctly.

Papa's Daughters 1 require 1 additional method to run:
* GetImageInfo method should return image size (if "fsdata\\splash%d.jpg" exists, they are resized to 1024x768 and displayed in game intro). Hardcoded to "1024;768".

Papa's Daughters 2 require 2 additional APIs to run:
* There are LOTS of methods with misleading names, that are actually used to calculate CRC-like function of game binary parts. They are called here and there, and game exits if CRC does not match expected value (there are 100+ CRCs of different exe file parts hardcoded in different parts of game scripts to be checked with those methods). Reverse engineered CRC algirithm and reimplemented.
* There is another self-check, that expects Scene.GetNode("papasEntity") to return 777 integer, while Scene.GetNode() usually returns native object or null. Hardcoded.

Other things in this PR:
* Game.RegReadString() contained a bug, because of which game progress was not saved at all HeroCraft games, while other WME games were not affected much
* Some tidy up work is done at File and Directory classes: warning() when needed, comments on known TODOs for various non-HeroCraft games
* Call warning() on runtime error instead of LOG(), so error would always be logged
* Use vsnprintf() for runtime errors


engines/wintermute/base/base_game.cpp
```diff
@@ -4249,6 +4344,40 @@ char *BaseGame::getKeyFromStringTable(const char *str) const {
...
+	}
+
+	if (size < end) {
+		warning("File '%s' is too small for checksum check", filename);
+		delete[] buffer;
+		return 0;
+	}
+
+	uint8 result = 0;
+	for (size_t i = begin; i < end; i++) {
```
sev-: Please use ScummVM types, not `size_t`

engines/wintermute/base/base_game.cpp
```diff
@@ -4249,6 +4344,40 @@ char *BaseGame::getKeyFromStringTable(const char *str) const {
...
+		uint8 tmp = buffer[i];
+		result += tmp;
+		if (result < tmp) {
+			result ++;
```
sev-: Code formatting, should be `result++`

engines/wintermute/ad/ad_entity.cpp
```diff
@@ -838,6 +840,78 @@ bool AdEntity::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
...
+	//////////////////////////////////////////////////////////////////////////
+	// [WME Kinjal 1.4] SetBeforeEntity / SetAfterEntity
+	// Usage at HeroCraft games: ent.SetBeforeEntity("redDuskaEntity")
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "SetBeforeEntity") == 0 || strcmp(name, "SetAfterEntity") == 0) {
+		stack->correctParams(1);
+		const char *layerName = stack->pop()->getString();
+
+		if (strcmp(getName(), layerName) == 0) {
+			warning("%s(%s): source and target has the same name", name, layerName);
```
sev-: has=have

engines/wintermute/base/scriptables/script_ext_directory.cpp
```diff
@@ -95,13 +98,20 @@ bool SXDirectory::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisSt
...
-		stack->pop()->getString();
+		const char *dirName = stack->pop()->getString();
 		stack->pop()->getString();
 
 		stack->pushInt(0);
 		BaseScriptable *array = makeSXArray(_gameRef, stack);
 
-		warning("Directory.%s is not implemented! Returning empty array...", name);
+		// used in secret scene of "Art of Murder 1: FBI Confidential"
+		if (strcmp(dirName, "X:\\FBI\\data\\scenes\\17-magic\\") == 0 && strcmp(name, "GetDirectories") == 0) {
```
sev-: This looks like it belongs to another game, e.g. the commit mixes things, but I guess, it is fine.

engines/wintermute/ad/ad_entity.cpp
```diff
@@ -838,6 +840,78 @@ bool AdEntity::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
...
+			AdLayer *layer = ((AdGame *)_gameRef)->_scene->_layers[i];
+			for (uint32 j = 0; j < layer->_nodes.size(); j++) {
+				if (layer->_nodes[j]->_type == OBJECT_ENTITY && this == layer->_nodes[j]->_entity) {
+					for (uint32 k = 0; k < layer->_nodes.size(); k++) {
+						if (layer->_nodes[k]->_type == OBJECT_ENTITY && strcmp(layer->_nodes[k]->_entity->getName(), layerName) == 0) {
+							if (j < k && strcmp(name, "SetBeforeEntity") == 0) {
+								k--;
+							} else if (j > k && strcmp(name, "SetAfterEntity") == 0) {
+								k++;
+							}
```
bluegr: Comparing strings in each for loop iteration is costly. I suggest you store the result of the comparison in a boolean variable outside the for loops, and use that instead, i.e. something like:

    bool isSetBeforeEntity = !strcmp(name, "SetBeforeEntity");    // before the for loops
    bool isSetAfterEntity = !strcmp(name, "SetAfterEntity");    // before the for loops
    ....
    if (j < k && isSetBeforeEntity)
        k--;
    else if (j > k && isSetAfterEntity)
        k++;

engines/wintermute/ad/ad_entity.cpp
```diff
@@ -838,6 +840,78 @@ bool AdEntity::scCallMethod(ScScript *script, ScStack *stack, ScStack *thisStack
...
+	// [WME Kinjal 1.4] GetLayer / GetIndex
+	//////////////////////////////////////////////////////////////////////////
+	else if (strcmp(name, "GetLayer") == 0 || strcmp(name, "GetIndex") == 0) {
+		stack->correctParams(0);
+
+		for (uint32 i = 0; i < ((AdGame *)_gameRef)->_scene->_layers.size(); i++) {
+			AdLayer *layer = ((AdGame *)_gameRef)->_scene->_layers[i];
+			for (uint32 j = 0; j < layer->_nodes.size(); j++) {
+				if (layer->_nodes[j]->_type == OBJECT_ENTITY && this == layer->_nodes[j]->_entity) {
+					if (strcmp(name, "GetLayer") == 0) {
```
bluegr: Same here, there's no reason to do this on each loop iteration

bluegr:
Nice work, well done! I agree with @sev, I believe that the changes are not big enough to warrant a separate subengine define for these games

engines/wintermute/base/base_game.cpp
lolbot-iichan: Fixed. But let me ask, what is wrong with `size_t`? It's widely used in tons of code, including all kinds of engines, backends and even `common/str.h`.

engines/wintermute/base/base_game.cpp
lolbot-iichan: Fixed, thanks!

engines/wintermute/ad/ad_entity.cpp
lolbot-iichan: Fixed, thanks!

engines/wintermute/base/scriptables/script_ext_directory.cpp
lolbot-iichan: That's correct, 73b669a9d32802cd8efe5d1fe1db7b6e9113c1b1 is unrelated to main subject (except for `error()` changed to `warning()` on `File.Delete()`, which is requied for HeroCraft games). 

Those notes were added while researching on why user profiles were not loaded correctly at HeroCraft games, before noticing that `Game.RegReadString()` is broken, I looked through usage of Directory and File APIs. Since I have decompiled game scripts for almost all games, I decided to run `grep` on them all, not only HeroCraft game scripts only.

engines/wintermute/ad/ad_entity.cpp
lolbot-iichan: There is `return` statement after those `strcmp`s, so they are actually called only once.
Added some comments since this function is too complex and confusing.

engines/wintermute/ad/ad_entity.cpp
lolbot-iichan: Same here, it's done only once, then `return` is called

lolbot-iichan:
> I wonder, though, what is the benefits of keeping it as a subengine?
>I agree with @sev-, I believe that the changes are not big enough to warrant a separate subengine define for these games

I got 2 ideas in mind while adding FoxTail and HeroCraft as subengines:
1. There are projects like https://jennibee.itch.io/dreamcat, which expects ScummVM's WME to be a a tiny point-and-click engine and an alternative runtime for games made with original WME devkit. So I didn't want to bloat ScummVM's WME with undocumented and studio-specific features like self-check functions of Papa's Daughters 2.
2. There are projects that are using ScummVM for GOG releases. I guess it may be kind of strange if things like `papasEntity` would be integrated in their build of ScummVM. I thought it may be better for them not to redistribute reverse engineered code unrelated to their game.

bluegr:
Thanks for the explanation. Yes, your reasoning makes sense, regarding the subengine.

Thanks for your work! Merging

PR #2096, opened 2020-02-26
DRAGONS: Fix several compiler warnings

jepael:
This PR contains several simple and unrelated commits that address compiler warnings about comparison between signed and unsigned variables.


sev-:
Nice!

PR #2097, opened 2020-02-26
DRAGONS: Fix compiler warnings

jepael:
This PR contains two commits to fix compiler warnings about signed vs unsigned variables.

In both files/classes, the private member variable  _count is of type int16, but it does not need negative values so it is changed to type uint16.


sev-:
Thanks!

PR #2098, opened 2020-02-26
ENGINES: Keep track of the unavailable files in AdvancedDetector

bgK:
Checking if a file exists is somewhat costly on the 3DS. Remembering which files do not exist instead of repeatedly checking for them reduces the time it takes to detect games from ~1 minute to ~10 seconds in some cases.


bluegr:
Quite straightforward. It would be good to document in a comment why we need to cache negative results - the description you gave above would be sufficient

bgK:
> It would be good to document in a comment why we need to cache negative results - the description you gave above would be sufficient

I prefer keeping information about context in the commit message. However I've expanded a bit the newly added comment to explain the reason.

PR #2100, opened 2020-02-28
MACVENTURE: Fix compiler warnings

jepael:
This PR fixes multiple compiler warnings about comparison between signed and unsigned variables, as it is reported each time this header is included.

The type int should be large enough to handle calculations between uint16 surface height and int font height.


bluegr:
Nice catch! Merging

PR #2101, opened 2020-02-28
ULTIMA8: Clean up header guards

mduggan:
A trivial change to standarize some header guards, making them ALLCAPS and match the name of the files they're in.


bluegr:
Nice! Merging

PR #2102, opened 2020-02-28
ULTIMA8: Reset palette when showing menu

mduggan:
If the avatar dies while tripping on mushrooms or the menu is opened
when the palette is changed (eg, during a scripted fight with the
guards where the screen blackens) then the tombstone and menu palette
are also changed.

This resets the palette temporarily while showing the menu, and
permanently on avatar death.

I'm not totally sure if this is the cleanest way to do it, there may be some better way I'm missing, or some side-effect of doing it this way I haven't noticed - for now it seems to work well in my testing.

### DIFF
--- a/engines/ultima/ultima8/graphics/palette_manager.cpp
+++ b/engines/ultima/ultima8/graphics/palette_manager.cpp
@@ -71,7 +71,8 @@ void PaletteManager::resetTransforms() {
 		Palette *pal = _palettes[i];
 		if (!pal) continue;
 		pal->_transform = Transform_None;
-		for (int j = 0; j < 12; j++) pal->_matrix[j] = matrix[j];
+		for (int j = 0; j < 12; j++)
+			pal->_matrix[j] = matrix[j];
 		_renderSurface->CreateNativePalette(pal); // convert to native format
 	}
 }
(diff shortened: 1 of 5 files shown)


engines/ultima/ultima8/graphics/palette_manager.cpp
```diff
@@ -155,6 +155,15 @@ void PaletteManager::untransformPalette(PalIndex index) {
 	transformPalette(index, matrix);
 }
 
+bool PaletteManager::getTransformMatrix(int16 matrix[12], PalIndex index) {
+	Palette *pal = getPalette(index);
+
+	if (!pal) return false;
+
+	for (int i = 0; i < 12; i++) matrix[i] = pal->_matrix[i];
```
bluegr: Please follow our coding guidelines, and don't put these in one line

mduggan: Sure. I don't like this style either and was just following the other loops in the file. Since there's only 2 others I'll just fix them all.

dreammaster: Thanks. Properly seperating lines like that from the original Pentagram source is an outstanding cleanup. I'm hopeful that it can done in bulk, but at least running astyle using the provided parameters on the Wiki didn't do it. And I'm wary of accidentally introducing errors if I tried to do it manually file by file.

PR #2104, opened 2020-02-29
ULTIMA8: Fix apperance of minimap

mduggan:
The minimap appears incorrectly for 2 reasons:
* The memset_{32|16} implementations are buggy - uses a logical or instead
  of a binary or, and it fails to move the buffer pointer after doing
  the aligned section. This becomes more obvious in 16bpp mode.
* The top line is not long enough

This corrects both issues.

### DIFF
--- a/engines/ultima/ultima8/misc/memset_n.h
+++ b/engines/ultima/ultima8/misc/memset_n.h
@@ -55,7 +55,7 @@ inline void memset_32(void *buf, uint32 val, uint32 dwords) {
 		if ((reinterpret_cast<uintptr>(buf) & 1)) {
 			*reinterpret_cast<uint8 *>(buf) = static_cast<uint8>(val & 0xFF);
 			buf = (reinterpret_cast<uint8 *>(buf)) + 1;
-			val = ((val & 0xFF) << 24) || ((val & 0xFFFFFF00) >> 8);
+			val = ((val & 0xFF) << 24) | ((val & 0xFFFFFF00) >> 8);
 			align --;
 		}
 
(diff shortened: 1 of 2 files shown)


engines/ultima/ultima8/misc/memset_n.h
```diff
@@ -73,6 +73,7 @@ inline void memset_32(void *buf, uint32 val, uint32 dwords) {
 
 	// Do the unaligned data
 	if (align) {
+		buf = (reinterpret_cast<uint8 *>(buf)) + dwords*4;
```
bluegr:
```suggestion
		buf = (reinterpret_cast<uint8 *>(buf)) + dwords * 4;
```

Spacing

mduggan:
Force-pushed a clean branch because the commit message was the wrong format and Travis had timed out for some reason anyway.

mduggan:
Added a fix for the memset_16 implementation which was also buggy (just hadn't noticed as pointers tended to be 16-bit aligned at least)

dreammaster:
One problem I noticed was that the content of the minimap wasn't showing, and I've since committed a fix for that. The cause seems to be a deficiency in the original Pentagram, though I haven't yet double-checked in the original Pentagram when running in 16-bit mode to to confirm.

Basically, the Texture uses an uint32 buffer for pixels, but force cases the buffer point to a 16-bit pointer when the running bits-per-pixel is 2 (16 bit video mode). So basically,we end up with a mismatch.. Texture is expending the buffer to be a 16-bit array when in 16-bit mode, whereas the minimap expects the buffer to always be 32-bits, even though it puts 16-bit values into each buffer entry.

This may well be a problem for anything else using Texture similarly. Ideally, Texture should really be refactored to use a Graphics::ManagedSurface of it's own, and not have separate NATIVE/STANDARD modes. The current method of casting a uint32 * to uint16 *, is just nasty.

mduggan:
Oh, thanks for fixing that - I didn't even notice that the whole thing was black, stupidly assumed it was just something about the area I was in at the time!
Agree that the whole thing could probably do with a rework - I kind of wanted to find a way to delete these functions entirely but decided to just do the smaller fix for now.

PR #2105, opened 2020-02-29
ULTIMA8: Add some unit tests for base U8 classes

mduggan:
I've been slowly writing these unit tests, which helped me find the cause of #2088, #2094, and #2104.

There is also one change to the engine code in `uc_stack.h` to make a previously implicit header dependency explicit.


mduggan:
Of course.. the Travis build failed because the test was checking that #2104 was fixed.  It's merged now so pushed a rebase.

PR #2107, opened 2020-03-01
ILLUSIONS: Fix wide text handling logic for big endian systems. (#11236)

yuv422:
This should fix the issue with wide text processing on big endian systems.

Could someone try this on a big endian system for me to see if I've fixed it.

### DIFF
--- a/engines/illusions/illusions.cpp
+++ b/engines/illusions/illusions.cpp
@@ -61,17 +61,25 @@
 
 namespace Illusions {
 
-char *debugW2I(byte *wstr) {
+char *debugW2I(uint16 *wstr) {
 	static char buf[65];
 	char *p = buf;
 	uint i = 0;
 	while (*wstr != 0 && i < sizeof(buf) - 1) {
-		*p++ = *wstr;
-		wstr += 2;
+		*p++ = (byte)*wstr;
+		wstr++;
 		i++;
 	}
 	*p = 0;
 	return buf;
+}
+
+void swapBytesInWideString(byte * wstr) {
+#if defined(SCUMM_BIG_ENDIAN)
+	for (byte *ptr = wstr; *ptr != 0; ptr += 2) {
+		WRITE_UINT16(ptr, SWAP_BYTES_16(READ_UINT16(ptr)));
+	}
+#endif
 }
 
 IllusionsEngine::IllusionsEngine(OSystem *syst, const IllusionsGameDescription *gd) :
--- a/engines/illusions/illusions.cpp
+++ b/engines/illusions/illusions.cpp
@@ -74,7 +74,7 @@ char *debugW2I(uint16 *wstr) {
 	return buf;
 }
 
-void swapBytesInWideString(byte * wstr) {
+void swapBytesInWideString(byte *wstr) {
 #if defined(SCUMM_BIG_ENDIAN)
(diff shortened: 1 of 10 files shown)


bgK:
Hi,

I've tried this branch on a PS3, with the demo version. I believe the game goes a bit further, however it still eventually gets stuck in an infinite loop. This time at the pool scene, after a bit of dialog (right before getting control?).

engines/illusions/illusions.cpp
```diff
@@ -61,19 +61,27 @@
...
-		wstr += 2;
+		*p++ = (byte)*wstr;
+		wstr++;
 		i++;
 	}
 	*p = 0;
 	return buf;
 }
 
+void swapBytesInWideString(byte * wstr) {
```
bluegr: Formatting (byte *wstr)

engines/illusions/resources/talkresource.cpp
```diff
@@ -45,11 +45,18 @@ void TalkEntry::load(byte *dataStart, Common::SeekableReadStream &stream) {
...
 	_tblPtr = dataStart + tblOffs;
 	_voiceName = dataStart + voiceNameOffs;
 	debug(0, "TalkEntry::load() _talkId: %08X; textOffs: %08X; tblOffs: %08X; voiceNameOffs: %08X",
 		_talkId, textOffs, tblOffs, voiceNameOffs);
+
+#if defined(SCUMM_BIG_ENDIAN)
+	for (byte *ptr = (byte *)_text; ptr != _tblPtr; ptr += 2) {
+		WRITE_UINT16(ptr, SWAP_BYTES_16(READ_UINT16(ptr)));
+	}
+#endif
```
bluegr: Why don't you use swapBytesInWideString() here?

sev-:
Any update on this?

yuv422:
I got a bit stuck trying to get a bigendian environment setup. I'll take another look this weekend.

sev-:
No news?

yuv422:
I still haven't tested this on a big endian system. I think it is safe to merge though. I'll just check it against master first.

engines/illusions/resources/talkresource.cpp
yuv422: This one doesn't check for the null terminator

raziel-:
@yuv422

Sorry, but I've  just seen this PR.
I have a BE machine (AmigaOS4) and could do some testing.

Which games use the illusion engine?
Is that Duckman?
If it is then there's already a bug item in the tracker from me.

If not, what games and what exactly would you need me to test?

I can start testing at the end of this week

yuv422:
Hi @raziel- 
Thanks that would be great.
There is an issue in the tracker. https://bugs.scummvm.org/ticket/11236
The game is Duckman. You can use the demo version to test.

raziel-:
@yuv422

I can test in a few days, but I think I already found a duplicate or at least something similar.
See here:
https://bugs.scummvm.org/ticket/11528

raziel-:
@yuv422
https://bugs.scummvm.org/ticket/11236#comment:1

Crash on BE AmigaOS4, crashlog attached to the bug item

sev-:
@yuv422 do you have time to take a look into this?

yuv422:
Ok I tested this against master and it looks ok. I still need to do a bit more to support big endian systems but I'll do that in another PR

PR #2108, opened 2020-03-02
ULTIMA8: Free the correct pointer in ~OAutoBufferDataSource

mduggan:
Also add a unit test for the class to make sure the problem
is fixed.
The unit test doesn't directly detect the problem, it just exercises the class - but it will show this sort of error:
```
Running 282 tests...........................................................................................................................................................................................................................................................................runner(58520,0x113aac5c0) malloc: *** error for object 0x7f8f094004ca: pointer being freed was not allocated
runner(58520,0x113aac5c0) malloc: *** set a breakpoint in malloc_error_break to debug
make: *** [test] Abort trap: 6
```

`~OAutoBufferDataSource()` is freeing the wrong pointer - `bufPtr` is the moving pointer to the next byte to write, but `buf` is the pointer to the start of the range.

### DIFF
--- a/engines/ultima/ultima8/filesys/odata_source.h
+++ b/engines/ultima/ultima8/filesys/odata_source.h
@@ -359,7 +359,7 @@ public:
 	}
 
 	~OAutoBufferDataSource() override {
-		delete [] _bufPtr;
+		delete [] _buf;
 	}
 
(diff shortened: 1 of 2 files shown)


bluegr:
Nice catch! Merging

PR #2109, opened 2020-03-02
SCI: Added new VM hook mechanism, allowing new instructions

ZvikaZ:
The current patch mechanism allows only replacing intructions in current
scripts, but not adding new instructions out of nowhere.
Thus, issue like #9646 is about to be closed as WONTFIX.

This new mechanism adds a hook on vm.cpp, before executing opcodes, and
if required, executs a new code.

It solves one of the issues of #9646, the others can be solved as well using that mechanism.

### DIFF
--- a/engines/sci/engine/vm_hooks.cpp
+++ b/engines/sci/engine/vm_hooks.cpp
@@ -1,0 +1,165 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "common/hashmap.h"
+#include "common/array.h"
+#include "sci/engine/vm_hooks.h"
+#include "sci/engine/vm.h"
+#include "sci/engine/state.h"
+#include "sci/engine/kernel.h"
+#include "sci/engine/scriptdebug.h"
+
+namespace Sci {
+
+/*****************************************************************************************************************
+ *
+ * This mechanism allows inserting new instructions, and not only replace existing code.
+ * Note that when using hooks, the regular PC is frozen, and doesn't advance.
+ * Therefore, 'jmp', 'bt' and 'bnt' are used to locally move around inside the patch.
+ * call* opcodes can be used - but they should be last executed opcode, in order to successfully transfer control.
+ *
+ ******************************************************************************************************************/
+
+
+// solves the issue described at #9646:
+// "
+// When in room 58, and type "run", the hero will fall and his HP will decrease by 1 point. This can be repeated, but will never cause the hero to die.
+// When typing "run" the ego will be assigned with the script egoRuns.
+// egoRuns::changeState calls proc0_36 in script 0 which is deducing damage from the hero's HP.
+// This procedure returns TRUE if the hero is still alive, but the return value is never observed in egoRuns.
+// "
+// we solve that by calling the hook before executing the opcode following proc0_36 call
+// and check the return value. if the hero should die, we kill him
+
+const byte qfg1_die_after_running_on_ice[] = {
+	// if shouldn't die, jump to end
+	0x2f, 22,					   // bt +22
+
+	// should die - done according to the code at main.sc, proc0_29:
+	// 			(proc0_1 0 59 80 {Death from Overwork} 82 800 1 4)
+	0x39, 0x08,                    // pushi 8		-- num of parameters
+	0x39, 0x00,                    // pushi 0
+	0x39, 59,					   // pushi 59
+	0x39, 0,                       // pushi 0		-- modified, not using {Death from Overwork}
+	0x36,                          // push
+	0x39, 82,					   // pushi 82
+	0x38, 32,  3,				   // push 800
+	0x39, 1,                       // pushi 1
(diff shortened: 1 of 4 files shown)


engines/sci/engine/vm_hooks.cpp
```diff
@@ -0,0 +1,143 @@
...
+		int framesize = 16;
+		int temp = ((framesize >> 1) + 0 + 1);
+		StackPtr s_temp = s->xs->sp;
+		s->xs->sp -= temp;
+
+		ExecStack *xs_new = execute_method(s, 0, 1, s_temp, s->xs->objp, s->xs->sp[0].getOffset(), s->xs->sp);
+		if (xs_new) { // in case of error, keep old stack
+			s->_executionStackPosChanged = true;
+			s->xs = xs_new;
+		};
```
bluegr: I'm not sure if I like this implementation. We're implementing logic which should be in scripts into engine code.
Up to now, we have gone into great lengths to keep this logic into script patches. Why can't we do the same in this case?

ZvikaZ: Are we talking about these 4 lines, or the general idea of this PR?

If it's these lines, do you have better suggestion for them?
If it's about the PR in general - see the discussion at https://bugs.scummvm.org/ticket/9646 for justification of the PR's approach.

bluegr: We are talking about the approach taken. I am aware of the bug itself, but I'm asking why are we mixing script logic and engine code with this approach.

ZvikaZ: I assumed you meant regarding the approach. Strangely, GitHub I/F showed it as if you were speaking regarding the last 4 lines, which seemed weird.

Never mind that, let's talk about the approach.
Indeed, mixing script logic and engine code isn't recommended, I agree. But I think that sometimes, we have to take inferior design decisions, to solve problems.

And as far as I have understand, we are already mixing script logic and engine code to solve some bugs that are impossible/difficult to solve in the regular ways.

Just looked through the code, and copied some random examples:

kgraphics.cpp, lines 1171-1183:

	reg_t kAnimate(EngineState *s, int argc, reg_t *argv) {
		reg_t castListReference = (argc > 0) ? argv[0] : NULL_REG;
		bool cycle = (argc > 1) ? ((argv[1].toUint16()) ? true : false) : false;

		g_sci->_gfxAnimate->kernelAnimate(castListReference, cycle, argc, argv);

		// WORKAROUND: At the end of Ecoquest 1, during the credits, the game
		// doesn't call kGetEvent(), so no events are processed (e.g. window
		// focusing, window moving etc). We poll events for that scene, to
		// keep ScummVM responsive. Fixes ScummVM "freezing" during the credits,
		// bug #3101846
		if (g_sci->getGameId() == GID_ECOQUEST && s->currentRoomNumber() == 680)
			g_sci->getEventManager()->getSciEvent(kSciEventPeek);


kpathing.cpp, lines 1188-1196:

	// WORKAROUND: broken polygon in lsl1sci, room 350, after opening elevator
	// Polygon has 17 points but size is set to 19
	if ((size == 19) && g_sci->getGameId() == GID_LSL1) {
		if ((s->currentRoomNumber() == 350)
		&& (readPoint(pointList, 18) == Common::Point(108, 137))) {
			debug(1, "Applying fix for broken polygon in lsl1sci, room 350");
			size = 17;
		}
	}

kscripts.cpp, lines 291-299:

	// WORKAROUND: Bugfix for the intro speed in PQ2 version 1.002.011.
	// This is taken from the patch by NewRisingSun(NRS) / Belzorash. Global 3
	// is used for timing during the intro, and in the problematic version it's
	// initialized to 0, whereas it's 6 in other versions. Thus, we assign it
	// to 6 here, fixing the speed of the introduction. Refer to bug #3102071.
	if (g_sci->getGameId() == GID_PQ2 && script == 200 &&
		s->variables[VAR_GLOBAL][kGlobalVarSpeed].isNull()) {
		s->variables[VAR_GLOBAL][kGlobalVarSpeed] = make_reg(0, 6);
	}

ZvikaZ: Anyway, I've got an idea that maybe will improve things, so let's put this in hold, and I will check it.

bluegr: The code you mention is out of place indeed, but most of it was added in early stages of the SCI engine development, before the script patcher was implemented. It should be moved into script patches.

ZvikaZ: I've separated script logic from engine code.
But it comes with a price...

I had to modify opcode behaviour in `vm.cpp`, see `case op_bt`.
If it's OK, I will need to change also `op_bnt`.

(regarding `jmp` - I will make it just illegal in these patches, without touching `vm.cpp`)

That's because the hook's instructions aren't placed in the interpreter memory, but come from a "bypass" - and most critical - the PC has the same value during their execution.

ZvikaZ: I'm working on improvement that will allow `bnt` with negative offset, and `jmp` will be supported as well.

But still, I'd like to hear your opinion regarding the last commit I've pushed.

ZvikaZ: Well, I've fixed the issues I've previously mentioned (after separating script logic and engine code).

I think that now the code is ready.

ZvikaZ: Found a problem.
I've debugged other issue, and noticed that some instruction was initially at `0040:01e7`, but on another debugging session it was at `003e:01e7`.
So, the `segment` might change, and I probably shouldn't check it.

I hope that we *can* count on `offset` being constant (as I saw until now), and use the combination of `script number + offset`.

ZvikaZ: Fixed that.
Ready for a new review.

engines/sci/engine/vm_hooks.h
```diff
@@ -0,0 +1,103 @@
...
+#include "sci/engine/vm.h"
+
+namespace Sci {
+
+
+#define ARRAYSIZE(x) ((int)(sizeof(x) / sizeof(x[0])))
+#define HOOKARRAY(x) x, ARRAYSIZE(x)
+
+/** _hooksMap keys are built from script number and PC's offset */
+struct HookHashKey {
```
bluegr: This is not needed, the exact same functionality is provided by reg_t's

ZvikaZ: You mean to replace `HookHashKey` usage with `reg_t`?

- But in `reg_t` it's `_segment` while in `HookHashKey` it's `scriptNumber`
(of course, I *can* use `_segment`  for `scriptNumber`, but it'd be confusing, as anyone who reads that would think that it's really a segment
-  `reg_t` doesn't have `hash` function, which is the main reason for this struct existence

bluegr:
Nice work! This can be merged now. Thanks

PR #2111, opened 2020-03-03
SCI: debugger: disasm_addr - fixed 'bcc' parameter

ZvikaZ:


bluegr:
Nice catch, thanks! Merging

PR #2115, opened 2020-03-04
ULTIMA8: Add unit tests for IBufferDataSource

mduggan:
Adding some basic unit tests for IBufferDataSource.  Didn't find any bugs in here.


mduggan:
Codacy is complaining about my static test string potentially not being null terminated, which seems unlikely :)
I'll fix it by hard-coding the string length if anyone cares but honestly for a unit test I think the code is cleaner as-is?

dreammaster:
It should be fine as is

PR #2116, opened 2020-03-04
ULTIMA8: Fix menu/credits music behavior

mduggan:
There are 2 problems with the way combat music interacts with the menu (and the
credits you can view from there):

* When the menu is closed, the combat music becomes the "last request", so it
  never stops even if combat finishes.
* If the creidts are viewed, the requested credits music is ignored because
  combat is active, and the music player is checking the actor's status to
  decide whether to change tracks.

This fixes both those problems, and restores the original game behavior of just
turning off the music when the menu is opened.  I also removed the coupling of
actor state from the music process.  I think this now works properly in all
situations.

### DIFF
--- a/engines/ultima/ultima8/audio/music_process.cpp
+++ b/engines/ultima/ultima8/audio/music_process.cpp
@@ -26,9 +26,6 @@
 #include "ultima/ultima8/audio/music_flex.h"
 #include "ultima/ultima8/audio/midi_player.h"
 #include "ultima/ultima8/audio/audio_mixer.h"
-#include "ultima/ultima8/kernel/object_manager.h"
-#include "ultima/ultima8/world/get_object.h"
-#include "ultima/ultima8/world/actors/main_actor.h"
 #include "ultima/ultima8/filesys/idata_source.h"
 #include "ultima/ultima8/filesys/odata_source.h"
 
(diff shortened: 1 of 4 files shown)


engines/ultima/ultima8/audio/music_process.cpp
```diff
@@ -96,9 +88,23 @@ void MusicProcess::unqueueMusic() {
...
 void MusicProcess::restoreMusic() {
 	_queuedTrack = 0;
+	_combatMusicActive = false;
 	playMusic_internal(_lastRequest);
 }
 
+void MusicProcess::getTrackState(int trackState[3]) const {
+	trackState[0] = _wantedTrack;
+	trackState[1] = _lastRequest;
+	trackState[2] = _queuedTrack;
```
mduggan: If I'm honest this feels a little hacky, I considered creating a struct to make it cleaner.  Happy to take input on it.

dreammaster: I agree. A structure would be better. You could also have the structure have a default constructor to initialize the fields. Also, Common::fill should be used in preference to direct memset calls generally, though in this case neither will be needed if you make the structure.

mduggan: Thanks. I updated it to be a struct.  Feels better.

engines/ultima/ultima8/audio/music_process.h
```diff
@@ -37,11 +37,27 @@ class MidiPlayer;
 class MusicProcess : public Process {
 	friend class Debugger;
 
-	enum MusicStates {
-		MUSIC_NORMAL = 1,
-		MUSIC_TRANSITION = 2,
-		MUSIC_PLAY_WANTED = 3
+	enum PlaybackStates {
```
mduggan: I changed the name of this enum (and entries) to be more clear now that we also have "track" state.

PR #2118, opened 2020-03-05
ULTIMA8: Clean up code for pentagram menu

mduggan:
This removes the code for generating the Pentagram menu (which is replaced by the ScummVM menu) and command line arguments which are no longer used in ScummVM.

There's still a bunch of work to do in cleaning up how the settings work to be more ScummVM standard, so this is only a first step of deleting the easiest bits.


dreammaster:
Looks good

PR #2129, opened 2020-03-08
MOHAWK: RIVEN: Refresh cursor before transitions

BallM4788:
This fixes a quirk on the 3DS port where touching a transition-triggering hotspot would move the cursor, but not update the cursor image until after the transition finished.

### DIFF
--- a/engines/mohawk/riven_card.cpp
+++ b/engines/mohawk/riven_card.cpp
@@ -928,6 +928,7 @@ RivenHotspot *RivenCard::getCurHotspot() const {
 
 RivenScriptPtr RivenCard::onMouseDown(const Common::Point &mouse) {
 	RivenScriptPtr script = onMouseMove(mouse);
+	updateMouseCursor();
 
 	_pressedHotspot = _hoveredHotspot;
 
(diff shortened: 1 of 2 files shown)


bgK:
Thanks. To be fair, I'm no too keen on drawing extra frames, especially not on each click. The speedrunning community people are counting the individual frames for TASes: https://www.youtube.com/watch?v=DU87V29EBho.

How about updating the mouse cursor sooner, when receiving the mouse move event rather than on click: https://gist.github.com/bgK/d4507984cfd1aea3ec690b93fbee1541

BallM4788:
That could definitely work, but your implementation causes the moved cursor not to render until the next card is loaded for the transition. I'll need to work out how to also show the moved cursor before the next card begins loading, as it does in master.

bgK:
How about making image decoding much faster so it does not really matter: https://gist.github.com/bgK/bd75d7dfab32e875ce49582324e9d980

BallM4788:
@bgK Moving the onMouseUpdate lines (per you first diff) into the onMouseMove function was breaking drag controls, so I opted for adding a call to updateMouseCursor into the onMouseDown function.

bgK:
Thanks, merged manually as 047545cc1b..3b005ac890

PR #2130, opened 2020-03-08
ENGINES: Exposed ConfigDialog

dreammaster:
This is a fairly minor change, but since it touches the GMM, I thought it worthwhile doing as a pull request for review.

Essentially, it moves the class declaration of ConfigDialog, used for the GMM Options/keymapper dialog, from the code file to the header file. This lets it be instantiated by engines that have in-game options buttons, but whose settings are all now handled by the ScummVM GUI. As is the case for Ultima VIII. I briefly considered moving the class's code into an entirely new file, but for now it still remains in the engines/dialogs.cpp/.h

Additionally, the TsAGE also used the dialog, but made the mistake of creating it's own sublcass of GUI::OptionsDialog rather than exposing the existing one. Due to various changes over time, this "mini" config dialog is now erroring out. Luckily, the error seems to have previously gone unnoticed up till now. I've removed the custom ConfigDialog from the TsAGE code and changed it to use the now exposed ConfigDialog as well.

### DIFF
--- a/engines/dialogs.cpp
+++ b/engines/dialogs.cpp
@@ -48,19 +48,6 @@
 #include "gui/KeysDialog.h"
 #endif
 
-class ConfigDialog : public GUI::OptionsDialog {
-protected:
-#ifdef GUI_ENABLE_KEYSDIALOG
-	GUI::Dialog		*_keysDialog;
-#endif
-
-public:
-	ConfigDialog(bool subtitleControls);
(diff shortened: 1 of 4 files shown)


bluegr:
Nice work! This change is straightforward and the rationale behind it is correct.

Since there aren't any functionality changes here, this can safely be merged.

Merging

PR #2133, opened 2020-03-09
ULTIMA8: add mouse buttons and controller defaults to keymap

rsn8887:
The custom keymap used in Ultima 8 broke mouse click that are triggered via game controller buttons A and B.

This PR adds remappable mouse click events that can be mapped to game controller buttons. It also adds some other default game controller mappings to the custom keymap used in Ultima 8.


dreammaster:
Thanks for doing the work for this. Much appreciated.

PR #2134, opened 2020-03-09
KEYMAPPER: Map Joy_X to skip lines by default

rsn8887:
This maps game controller X button to '.' (skip line of dialog) by default. I don't see a reason not to map it like this, since that is how it used to be before the keymapper, see here: https://wiki.scummvm.org/index.php?title=Nintendo_Switch


rsn8887:
Never mind, Joy_X is the game menu by default.

rsn8887:
I am merging this because it is trivial I think. No need to waste anybody's time reviewing.

PR #2135, opened 2020-03-10
ULTIMA8: Replace 0s with nullptr where applicable

mduggan:
The U8 engine mixes item IDs with pointers, so it's often hard to know what I'm looking at when code sets or compares something to 0.  I went through and replaced all `0` literals which actually meant null pointer with the `nullptr` literal - I think I got most of them.

Sorry for the monster patch.  There are a couple of instances of whitespace fixes and me moving things from constructors to initializer lists, so it's not a totally "pure" patch.  I can go back and pull those out if they bother you.

### DIFF
--- a/engines/ultima/ultima8/audio/audio_channel.cpp
+++ b/engines/ultima/ultima8/audio/audio_channel.cpp
@@ -42,7 +42,7 @@ namespace Ultima8 {
 #define RANGE_REDUX(x)  (((x) * 27) >> 5)
 
 AudioChannel::AudioChannel(Audio::Mixer *mixer, uint32 sampleRate, bool stereo) :
-		_mixer(mixer), _decompressorSize(0), _frameSize(0), _loop(0), _sample(0),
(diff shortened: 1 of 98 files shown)


dreammaster:
Just reviewing it now; looking good. Though one problem I've previously encountered is that nullptr is technically a C++11 feature, and at least one of our compilers, for AmigaOS4 if I remember correctly, doesn't support it. We do have a manual #define for such cases in common/c++11-compat.h, but I've previously run into trouble where particular header files using nullptr weren't pulling in enough of the system files to include it.

Upshot is,  ideally, any place where you've put in nullptr in any inline methods or initializers specified in header .h files would be better moved into the cpp files were feasible. Though single line methods that simply return nullptr would still be cleaner to leave in the header files.. I'll just manually add in inclusion of the compat file post-merge if the AmigaOS4 compiler barfs on any of them.

mduggan:
Done.  I removed the references from headers where I could.

PR #2138, opened 2020-03-12
SDL: Enable joystick input by default

bgK:
With the keymapper, game controllers are more configurable. I believe it's time to enable them by default in the SDL port. This pull request also includes a change to reload the keymapper's keymaps when a joystick is plugged or unplugged so the the default joystick bindings are picked up if necessary.


bluegr:
Good idea! +1 from me

PR #2141, opened 2020-03-14
ULTIMA8: Remove redundant cleanup in destructors (emptying vectors etc)

mduggan:
Cleaned up a few destructors which unnecessarily set member pointers to null or empty vectors just before the object is destroyed.

I double-checked that all of these were members of their respective classes, so the change should be safe.

### DIFF
--- a/engines/ultima/ultima8/audio/music_flex.cpp
+++ b/engines/ultima/ultima8/audio/music_flex.cpp
@@ -41,7 +41,6 @@ MusicFlex::~MusicFlex() {
 	uint32 i;
 	for (i = 0; i < 128; i++) {
 		delete _info[i];
-		_info[i] = 0;
 	}
 
 	Archive::uncache();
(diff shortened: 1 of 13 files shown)


bluegr:
This commit is a bit dangerous. In all these cases, the objects are deleted and their pointers are then set to null. This is good practice, since in some systems, like Windows, deleting an object does not necessarily set the pointer that was pointing at it to null, and the pointer will point to garbage after the deletion.

Thus, I'm not in favor of merging this.

mduggan:
@bluegr that's true, it will point to null, but all of these are inside destructors of the objects holding the members - setting the pointer to null is just vacuuming the carpet before the house is demolished.

dreammaster:
I lean on the side of mduggan, if only that I feel that explicitly setting fields to null can give the incorrect impression that they'll be accessed again later. Since of all these are in the complete teardown of containing parent objects, it's better to remove the redundant nulls.

PR #2143, opened 2020-03-14
PINK: fixed pausing when menu bar disappears

whiterandrek:
This pull request makes pink panther scenes to resume correctly after using menu bar

### DIFF
--- a/engines/pink/gui.cpp
+++ b/engines/pink/gui.cpp
@@ -155,8 +155,6 @@ void PinkEngine::executeMenuCommand(uint id) {
 
 	case kSaveAction:
 	case kSaveAsAction:
-		//FIXME: Somehow messes up the pause system causing issues such as
-		//frozen animations and BGM disappearing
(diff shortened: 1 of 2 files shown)


henke37:
This is a "yiekes" from me. This is steamrolling over a counter to get a desired result without regards to the count being correct.

henke37:
If anything, the best solution in my eyes is to rewrite the engine pause system from the ground up. No more manual pause/unpause. Use RAII to automatically increment and decrement the counter based on object lifespans.

whiterandrek:
The problem is not in pause system itself, but in menu which pauses the engine 2 times, but  unpauses 1 time
E.g
Menu appeared -> engine paused 1 time
Mouse button down -> engine paused 2 time
Mouse button up -> engine unpaused 1 time

henke37:
Then the menu system should be fixed. The engine is innocent and should not be attempting to cover for the menu system.

PR #2145, opened 2020-03-15
FULLPIPE: Shift loadGameState() and saveGameState()

harshbawari:


sev-:
What is the purpose of this exercise? The code change makes no sense.

harshbawari:
https://github.com/scummvm/scummvm/commit/0ae4cb1ea7907c7d558cb89ad101f762a0d64c8c

sev-:
352915aa?

harshbawari:
Can you provide the link? I can't find the commit

harshbawari:
I have spent this whole week figuring out what you want me to do exactly. I read the wiki, saw the code, and what I could make out is that you've already implemented MetaEngine in fullpipe. I don't understand what else you want me to do. Problem is 'what to do' not 'how to do'. And you are most welcome to block me on GitHub too

sev-:
The task was the following: We need to move Fullpipe engine save system to the new MetaEngine extended saves format, similarly like I did so for the HDB engine. I implemented the extended saves format precisely after fullpipe, so this should be a relatively straightforward task of removing code and reusing the MetaEngine parts.

This code covers nothing from it. Closing.

PR #2147, opened 2020-03-15
ALL: Pause Token

henke37:
This PR introduces a new system for managing the pause counter in engines. By using RAII errors with forgotten or duplicate calls to `Engine::pause` wont happen.

In particular, this appears to solve the issue in PINK where opening the save and load dialogs from the menu fails to correctly resume the engine.

### DIFF
--- a/backends/events/default/default-events.cpp
+++ b/backends/events/default/default-events.cpp
@@ -171,12 +171,11 @@ bool DefaultEventManager::pollEvent(Common::Event &event) {
 #endif
 	case Common::EVENT_RTL:
 		if (ConfMan.getBool("confirm_exit")) {
+			PauseToken pt;
 			if (g_engine)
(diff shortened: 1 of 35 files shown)


henke37:
As a bonus, it also cleans up several engine specific functions that duplicate existing code.

henke37:
Codacy has clearly never heard of variables whose very lifetime is semantically important.

henke37:
Took care of the conflict that happened.

sev-:
A simple glimpse at the code shows that you do not follow our code formatting guidelines in many places. Please fix.

sev-:
...and not following the code naming conventions. This needs to be fixed as well.

henke37:
Hopefully this should take care of all the stupid violations.

sev-:
Thanks! merging, great change!

PR #2150, opened 2020-03-16
SWITCH: Use ScummVM file buffering instead of newlib's implementation

bgK:
Fixes #11384.

 I don't have a Switch, this is untested. Based on my experience with the 3DS, ScummVM's implementation of file buffering has better performance than newlib's when seeks are involved. The buffer size probably needs some tuning.


rsn8887:
I just tested this with The Dig and Blade Runner, it seems to work great already as-is. The buffer seems fine! Merging...

PR #2154, opened 2020-03-20
GUI: Predictive dialog box can handle multiple keypad press

aryanrawlani28:
Fixes #7125

The predictive dialog box is unable to handle multiple presses via the keypad. This fixes it, only one press is allowed, and the simultaneous button presses are ignored - only the fastest one is registered.

### DIFF
--- a/gui/predictivedialog.h
+++ b/gui/predictivedialog.h
@@ -143,6 +143,8 @@ private:
 
 	bool _navigationWithKeys;
 	bool _needRefresh;
+	bool _isPressed;
+
 private:
 	EditTextWidget *_editText;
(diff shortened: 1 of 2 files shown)


gui/predictivedialog.h
```diff
@@ -143,6 +143,7 @@ class PredictiveDialog : public GUI::Dialog {
 
 	bool _navigationWithKeys;
 	bool _needRefresh;
+	bool _isPressed;
```
sev-: Please add an empty line before `private:`

aryanrawlani28: Done :)

PR #2155, opened 2020-03-20
COMMON: Dedicated function for in buffer check

henke37:
It's UB to compare pointers that aren't from the same array. Cast to uintptr_t to reduce the issue to IB.

### DIFF
--- a/common/str.cpp
+++ b/common/str.cpp
@@ -284,7 +284,7 @@ String &String::operator=(char c) {
 }
 
 String &String::operator+=(const char *str) {
-	if (_str <= str && str <= _str + _size)
+	if (pointerInOwnBuffer(str))
 		return operator+=(String(str));
 
 	int len = strlen(str);
@@ -295,6 +295,16 @@ String &String::operator+=(const char *str) {
 		_size += len;
 	}
 	return *this;
+}
+
+bool String::pointerInOwnBuffer(const char *str) const {
+	//compared pointers must be in the same array or UB
+	//cast to intptr however is IB
+	//which includes comparision of the values
+	uintptr ownBuffStart = (uintptr)_str;
+	uintptr ownBuffEnd = (uintptr)(_str + _size);
+	uintptr candidateAddr = (uintptr)str;
+	return ownBuffStart <= candidateAddr && candidateAddr <= ownBuffEnd;
 }
 
 String &String::operator+=(const String &str) {
--- a/common/str.h
+++ b/common/str.h
@@ -398,6 +398,8 @@ protected:
 	void decRefCount(int *oldRefCount);
 	void initWithCStr(const char *str, uint32 len);
 
+	bool pointerInOwnBuffer(const char *str) const;
+
 	void decodeUTF8(U32String &dst) const;
 	void decodeOneByte(U32String &dst, CodePage page) const;
 };


common/str.cpp
```diff
@@ -26,6 +26,7 @@
 #include "common/str.h"
 #include "common/util.h"
 #include "common/mutex.h"
+#include <stdint.h>
```
sev-: It is not allowed to include standard headers outside of backends/

common/str.cpp
```diff
@@ -279,7 +280,7 @@ String &String::operator=(char c) {
 }
 
 String &String::operator+=(const char *str) {
-	if (_str <= str && str <= _str + _size)
+	if(pointerInOwnBuffer(str))
```
sev-: Code formatting is wrong.

common/str.cpp
```diff
@@ -292,6 +293,13 @@ String &String::operator+=(const char *str) {
 	return *this;
 }
 
+bool String::pointerInOwnBuffer(const char *str) const {
+	uintptr_t ownBuffStart = (uintptr_t)_str;
```
sev-: It is not allowed to use standard types outside of `backends/` dure to portability reasons.

henke37:
You are fully right in that this is implementation defined behavior. The thing is, this is an _improvement_ over the current state. The previous code is depending on **undefined behavior**.

The code is comparing pointers that sometimes do not belong to the same array. That is undefined behavior.

If this new code doesn't work on a platform then the previous behavior may not have worked either. `uintptr_t` is indeed optional. It is only supplied on platforms where you can do this sort of code.

bgK:
> This code relies on a specific C++ implementation/standard behaviour and is not portable per our standards. uintptr_t is not defined on all platforms.

To be fair, we do define `uintptr`.

> The code is comparing pointers that sometimes do not belong to the same array. That is undefined behavior.

Unpopular opinion: The standard being silly is no reason to have weird workarounds for theoretical problems in the code.

Ultimately, don't all pointers belong to the byte array that covers the whole program address space?

henke37:
Like with most intentional UB in the standard, there is a motivation behind the rule. Some platforms do not have simple pointers. Segmentation, tiny and large pointers, all that stuff is why this rule exists.

bgK:
> Like with most intentional UB in the standard, there is a motivation behind the rule. Some platforms do not have simple pointers. Segmentation, tiny and large pointers, all that stuff is why this rule exists.

Right, segmented memory architectures. Do we care about those in ScummVM? Do the proposed changes improve the situation on those platforms?

henke37:
That takes care of all the guideline issues.

sev-:
`uintptr` is a simple `uint32` or `uint64`, depending on the architecture. So, it seems to me that what you do now is a no-op on the compilation time. Or am I missing something?

henke37:
It was just like that with the standard type. It's casting a pointer to an integer type. From a language lawyer standpoint, this is semantically important. In pretty much all implementations it's a nop that the compiler will filter out even on low optimization levels.

The point here is that the typecast is IB, while the original pointer comparison is UB. By doing the cast, it removes the UB from the following comparisons.

criezy:
I think one reason we might wand to accept the change is that even on flat memory architectures the current code may fail. The way I understand it, since comparing two pointers to different objects is UB, the compiler may assume that the two pointers point to the same object and optimise out the  `_str <= str` part and thus not catch cases where `str` is smaller than `_str`. Casting to integer types ensure this does not happen.

But since that check is done in only one place, do we need a new function for that?

Also the change may not help on segmented memory architectures (it depends how the pointer to integer conversion implementation is defined), but if it doesn't work the current code would probably not work either.

henke37:
A separate function at least documents the intent by giving the operation a name. And pretty much all compilers will inline the function anyway.

lephilousophe:
Why not just cast all operands and put a comment above to explain the cast?

Else, we could start defining templated operators less, greater, ... with pointers specialization as it's done in STL :smiley:
In this case that would need to be located in a header to let inlining fully work.

henke37:
To me, the function name and variable names in the function are doing a better job than comments.

criezy:
Using good function and variable names can go a long way in helping make the code easy to understand. In this case though it seems to me the intent of the code (check if the given pointer is within the String buffer) is quite clear without needing a separate function or a comment.

What is not so clear is why the cast to `uintptr` is needed (the UB for pointer checks is not a well known part of the standard), and the function name doesn't help with that. This is the part that could benefit from being clarified, and I think a comment would be more effective than trying to find a function name that would explain this.

Another minor point i that the commit message indicates it does a cast to `uintptr_t` but the code does a cast to `uintptr`. 

@henke37 can you please correct the commit message, and maybe add a comment in the code explaining that comparing pointers to different arrays is UB and that is why the cast to `uintptr` is needed?

After that I think it will be good to merge.

criezy:
Thank you.
I either missed the github notification for the force push, and there wasn't one. So I didn't see the change before today.
I will merge this now.

PR #2157, opened 2020-03-21
ULTIMA: Split and center text when too large

dulsi:
Character creation screens on Martian Dream are larger than the screen.
The text needs to be split into multiple lines.

I tried to be conservative to not accidentally break anything. It falls back to
the original core if it is shorter than the screen size.


dulsi:
Fixed complaints about code quality I believe.

PR #2158, opened 2020-03-21
ULTIMA8: Some minor shape and render cleanups

mduggan:
Some changes to make the code a bit clearer for `Shape` and `SoftRenderSurface`.

This is a first small PR of a longer-term plan I have to unify `Shape` and `Surface` code inside the U8 codebase and get it all a bit more using native ScummVM surfaces.

The rough plan is:
1. Tidy up the shape and render code - remove unused features, make it easier to analyze (this and maybe some other PRs)
1. Remove the in-memory use of RLE shapes - uncompress them on load so that we can remove 1 of the 2 separate render paths.
1. Change shapes internal representation more - from having row-level data to instead use RGBA surfaces so they can be blitted directly to the render surface
1. Change the `SoftRenderSurface` to always be a managed surface and use the line-draw/fill/blit provided by that instead
1. Remove the `ScalerGump` - make it all blt to a single surface 320x200 (tbd: fix the aspect ratio somehow) so that built-in ScummVM scalers can be used.

Looking for input on the overall idea as well as the changes in this PR.

### DIFF
--- a/engines/ultima/ultima8/convert/convert_shape.cpp
+++ b/engines/ultima/ultima8/convert/convert_shape.cpp
@@ -631,7 +631,7 @@ bool ConvertShape::CheckUnsafe(IDataSource *source, const ConvertShapeFormat *cs
 	bool result = true;
 
 	// Just to be safe
-	int start_pos = source->getPos();
+	const uint32 start_pos = source->getPos();
 
 	// Read the ident
 	if (csf->_bytes_ident)
@@ -739,7 +739,7 @@ bool ConvertShape::CheckUnsafe(IDataSource *source, const ConvertShapeFormat *cs
 void ConvertShape::Write(ODataSource *dest, const ConvertShapeFormat *csf, uint32 &write_len)
 {
 	// Just to be safe
-	uint32 start_pos = dest->getPos();
+	const uint32 start_pos = dest->getPos();
 
 	// Write the ident
 	if (csf->_bytes_ident) dest->write(csf->_ident, csf->_bytes_ident);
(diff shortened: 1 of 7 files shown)


dreammaster:
My comments on your points

> 1. Tidy up the shape and render code - remove unused features, make it easier to analyze (this and maybe some other PRs)
This is definitely a desirable long term goal. Everything that can simplify the codebase will be worth it long term, not only to make the further changes you do easier, but also for others in the future. It may even make it easier for someone to support Crusader: No Remorse, which shares the same engine.

> 2. Remove the in-memory use of RLE shapes - uncompress them on load so that we can remove 1 of the 2 separate render paths.
Definitely yes. It's not the old DOS days anymore when every single byte mattered, and if can allow the ditching of an entire render path, then so much the better.

> 3. Change shapes internal representation more - from having row-level data to instead use RGBA surfaces so they can be blitted directly to the render surface
As above. I've actually done likewise with other games. It's much easier and cleaner to encapsulate any decompression in the loading code, and then simply be able to blit any kind of image. Particularly in such cases where different variations of drawing code has be simplified into simple blit calls.

> 4. Change the SoftRenderSurface to always be a managed surface and use the line-draw/fill/blit provided by that instead
This is another thing I'd already briefly considered back when I was first converting the engine, so it definitely gets my approval. I was wary at the time of doing such changes to the code until I had a version that properly worked in ScummVM as a baseline. Now that we do, I'm all for having it simplified.

> 5. Remove the ScalerGump - make it all blt to a single surface 320x200 (tbd: fix the aspect ratio somehow) so that built-in ScummVM scalers can be used.
This has already been mentioned in-channel, so I concur with cleaning it up for removal. And frankly, having a scaler surface when the engine itself isn't meant to be doing any scaling is definitely not ideal.


I haven't yet reviewed the actual code of this PR, so I'll be done that next. I'll also request Dominus in-channel to see if he has any comments on the PR before it gets merged

dreammaster:
Okay, so the PR as it stands is in a non-compilable state because of the PaintMasked removal when I compiled it locally

mduggan:
Ugh, sorry - somehow I missed adding the header file where I  removed `PaintMasked` too.  Added it to the commit, rebased on master, and did a clean build locally to be sure.

DominusExult:
I really have no opinion. I never worked on that stuff.
The person with the most insight would be @wench https://github.com/wench

PR #2159, opened 2020-03-21
GUI: Scroll buttons properly display at edge.

aryanrawlani28:
Fixes #7123.

Previously the scroll buttons at extreme edges were clickable but did nothing. For example, if it were on the rightmost side, ">" would be clickable but useless. This fix addresses that.


sev-:
You still do not conform to our commit guidelines. Please fix.

sev-:
Thanks, merging.

sev-:
I changed the code to have the buttons enabled/disabled, not visible/invisible, though.

aryanrawlani28:
I see. It would be better this way I guess. I thought enabling/disabling were a little more expensive and could have unintended effects if other properties depended on them, so I went with visibility instead.

PR #2160, opened 2020-03-21
ULTIMA8: Use ScummVM PixelFormat struct

mduggan:
The Ultima8 engine defined its own `PixelFormat` struct with very similar members to the ScummVM one.  I refactored it to extend that struct instead of redefining it.

This refactor should help the planned changes that I mentioned in #2158.


dreammaster:
All looks good. This was something I briefly experimented with when I was doing the initial conversion, but got bogged down in the minor variations, so put it on a backburner.

PR #2162, opened 2020-03-22
ULTIMA: Split and center text when too large

dulsi:
Character creation screens on Martian Dream are larger than the screen.
The text needs to be split into multiple lines.

I tried to be conservative to not accidentally break anything. It falls back to
the original core if it is shorter than the screen size.

### DIFF
--- a/engines/ultima/nuvie/fonts/font.cpp
+++ b/engines/ultima/nuvie/fonts/font.cpp
@@ -54,19 +54,54 @@ uint16 Font::drawString(Screen *screen, const char *str, uint16 x, uint16 y, uin
 uint16 Font::drawString(Screen *screen, const char *str, uint16 string_len, uint16 x, uint16 y, uint8 color, uint8 highlight_color) {
 	uint16 i;
 	bool highlight = false;
-	uint16 font_len = 0;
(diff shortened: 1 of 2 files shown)


yuv422:
Screen Shot 2020-03-22 at 10 19 57 pm

ok I think this is a bug while porting over to ScummVM. I just tried it in Nuvie and it is working correctly.

yuv422:
I think we need to look at why the original code wasn't working as expected first.

dulsi:
s->text_align is 0 at 1576 in script_cutscene.cpp so it isn't running display_wrapped_text. Not sure why yet.

dreammaster:
Thanks for having looked into this. I'll have to apologize in this case.. it seems like, somehow, the Lua scripts I pulled in to make the ultima.dat file weren't the absolute latest version. I've now committed in the correct latest ones. And in the older version, the Lua script didn't have the line for text alignment.

However, I'm briefly holding off generating a new ultima.dat file, so you can see the result by editing your Martian Dreams game entry in the launch, and pointing the Extra Path to the devtools/create_ultima folder of your ScummVM installation. This will allow it to use the raw files directly.

dulsi:
Yes. It works with that change. Thanks.

PR #2163, opened 2020-03-22
ULTIMA8: Make ShapeFrame pointers const

mduggan:
A mostly trivial change to make all the use of `ShapeFrame` pointers const (and also remove the never-used parameter `need_bytes_rle` on `ShapeFrame::getConvertShapeFrame`), so that I can more confidently decompress the frames after loading and avoid separate render paths (as mentioned in #2158).

I think this will conflict with #2158 if one is merged first, I'll rebase if needed.


bluegr:
Const-correctness is always welcome. Thanks!

PR #2164, opened 2020-03-22
GUI: Custom engine settings tab + in-game game settings tab

bgK:
This PR aims at making various improvements to the per-engine settings. The following observations are the basis for the changes:

Some per-engine / game settings are available in the edit game dialog:
scummvm-00011
These are restricted to boolean values.

Some engines have a custom settings dialog:
scummvm-myst-win-1-00000
In the case of Myst the option from the edit game dialog is not available there.

The Global Game Menu has a settings dialog as well:
scummvm-myst-win-1-00001
There are no custom game options there.

This PR makes the following changes:
* Engines can define a custom tab to be shown in the edit game dialog. The full capabilities of ScummVM's GUI toolkit can be used to create an arbitrarily complex panel. This is done by implementing `MetaEngine::buildEngineOptionsWidget`. The default implementation is the existing "extra GUI options" system.
* Engines can choose whether their custom options tab should be shown in the in-game options dialog. This is done by declaring the `kSupportsChangingOptionsDuringRuntime` engine feature. `Engine::applyGameSettings` is called after the settings are saved to allow the engine to apply the changes the user made to the configuration.

The end result is as follows for Myst:
scummvm-00012
scummvm-myst-win-1-00002


sev-:
Great idea, a huge improvement.

However, if I would work on it, I would do It slightly differently. The thing which concerns me is that you're constructing the dialog manually right in the code.

What if the engines would be able to provide XML blocks which then would be assembled by the theme generator? Then we would have things in the same format, for the devs to learn one format, and more control and understanding over potential theme redesigns.

Of course, this is just an idea, and it will imply more work (though I'm willing to help with the implementation), so it is up to you about following it. The PR is OK to merge as-is.

bgK:
> However, if I would work on it, I would do It slightly differently. The thing which concerns me is that you're constructing the dialog manually right in the code.
>
> What if the engines would be able to provide XML blocks which then would be assembled by the theme generator? Then we would have things in the same format, for the devs to learn one format, and more control and understanding over potential theme redesigns.

Yes, indeed, I'm trying the approach of constructing the dialog layouts in the cpp source with this PR.  The motivation is as follows:
* The dialog layouts being defined in the theme files essentially prevents people from making and sharing their own themes. As soon as we make a change to a dialog we have to invalidate all existing themes by incrementing the theme format version.
* The dialog layouts are duplicated across all the themes. At the moment when changing a dialog we have to update 6 layout files with essentially the same modifications.
* The setting dialogs are not very interesting to theme when compared to say the launcher.

However you'll notice the new `OptionsContainerWidget` base class is flexible when it comes to layouts:
* It's possible to manually lay out the widgets as in `RemapWidget`
* It's possible to use a XML layout as in `ExtraGuiOptionsWidget`
* It's possible to define a layout in code as in the Myst and Riven. When the layout is defined in the code, if there is a layout with the same name defined in the theme XML files it takes precedence, allowing the dialogs to be customized.

As a result, I'm not really interested in building a system where engines can provide their XML themes; I would prefer converting existing XML definitions of uninteresting dialog layouts to code to enable users to create and share more durable themes. Perhaps it's not the correct approach, maybe the theme system needs to be made more powerful instead so there are "default layouts" defined in XML, embedded in the source, and that can be shared across the themes. I don't know.

bluegr:
Really nice improvement!

I too agree that changing the theme files just to add new game-specific options is cumbersome. Thus, handling these options in the code is more straightforward for developers.

The idea of having standard layouts / templates is also good, perhaps this could be done as a future improvement, to avoid unnecessary boilerplate code for simple cases.

sluicebox:
This introduces a minor bug: changing any SCI game ID with the GUI now causes an error(), https://bugs.scummvm.org/ticket/11416

(I still like the PR though!)

bgK:
Yes, I'll fix the issue, sorry for the inconvenience.

PR #2165, opened 2020-03-22
SCI: SQ4 intro - stop puker alien's sound

ZvikaZ:
solves one of the issues described at #11265 - the puker alien's sound is still heard on next room

### DIFF
--- a/engines/sci/engine/vm_hooks.cpp
+++ b/engines/sci/engine/vm_hooks.cpp
@@ -121,10 +121,23 @@ static const byte sci0_hebrew_input_prompt[] = {
 };
 
 
+// solves one of the issues described at #11265 - the puker alien's sound is still heard on next room
+// it seems like a script bug, which assumes the sound to be always 4 seconds, while it has varying length
+// depending on whether it's digital or midi (both are longer than 4 seconds...)
+// we solve that by calling:
+// (gLongSong2 stop:)
+const byte sq4_puker_stop_sound[] = {
+	0x38, 0xa7, 0x00,               // pushi    #stop
+	0x76,                           // push0
+	0x81, 0x9a,                     // lag      gLongSong2
+	0x4a, 0x04                      // send     4
+};
+
+
 
 /** Write here all games hooks
  *  From this we'll build _hooksMap, which contains only relevant hooks to current game
- *  The match is performed according to PC, script number, opcode (only opcode name, as seen in ScummVM debugger),
+ *  The match is performed according to PC (of next instruction to execute), script number, opcode (only opcode name, as seen in ScummVM debugger),
  *  and either:
  *  - objName and selector	(and then externID is -1)
  *  - external function ID  (and then selector is "")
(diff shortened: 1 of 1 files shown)


sluicebox:
Is this a script bug that needs patching, or is this a discrepancy between the way the interpreters handle audio?

Why does Sierra's interpreter stop the audio when the room changes and ScummVM doesn't? It seems like that needs to be discovered and explained first, and then implemented in the interpreter so that all similar scenes are fixed. (that reminds me, I have a patch comment along those lines I still haven't updated, oops)

ZvikaZ:
That's a good question!
I'd say that it's Sierra's interpreter's bug ;-)

And more seriously, note that in Dosbox the music stops *before* changing rooms, and I have no idea why.
It's in room 15, and that's the end of `barScript::changeState`:

			(17
				(gLongSong2 number: 886 loop: 1 play:)
				(puker
					posn: 287 143
					setLoop: 6
					setCel: 0
					cycleSpeed: 12
					setCycle: End self
				)
			)
			(18
				(puker posn: 287 115 setLoop: 8 setCel: 0 setCycle: Fwd)
				(= seconds 4)
			)
			(19 (global2 newRoom: 16))


State 17's `(gLongSong2 number: 886 loop: 1 play:)` starts the alien's sound.
State 18 is waiting 4 seconds, and state 19's `(global2 newRoom: 16)` will cause the room change.

I don't see in the code anything that stops the music, and I don't understand how Sierra's interpreter knew that it should stop it.

ZvikaZ:
Well, upon further inspection, maybe it's related to `newRoom` implementation.
Defined at `game.sc` (script #994),

	(method (newRoom newRoomNumber &tmp [temp0 5] temp5)
		(gOldATPs
			eachElementDo: #dispose
			eachElementDo: #delete
			release:
		)
		(gOldFeatures eachElementDo: #perform fDC release:)
		(gOldCast eachElementDo: #dispose eachElementDo: #delete)
		(gTimers eachElementDo: #delete)
		(gRegions eachElementDo: #perform DNKR release:)
		(gTheDoits release:)
		(Animate 0)
		(= gWest gNumber)
		(= gNumber newRoomNumber)
		(= gNewRoomNumber newRoomNumber)
		(FlushResources newRoomNumber)
		(self startRoom: gNumber)
		(while ((= temp5 (Event new: 3)) type?)
			(temp5 dispose:)
		)
		(temp5 dispose:)
	)

Maybe one of these `#dispose` or `#delete` stops the music at Dosbox, and probably should have done it also in our case. 
(and it stops the music a moment before visually changing rooms, and that's why I claimed that it happens before...)

I'll investigate this.

ZvikaZ:
Or more reasonable, the `(FlushResources newRoomNumber)` is the point of difference.

ZvikaZ:
I've traced the difference.
`newRoom` calls `(gRegions eachElementDo: #perform DNKR release:)`, which gets to `Rgn::dispose`, which calls `(gSounds eachElementDo: #clean self)`.

If I comment the last line in Dosbox, the sound doesn't stop.
I checked in ScummVM debugger, and got to `Sound::clean`:

	(method (clean param1)
		(if (or (not owner) (== owner param1))
			(self dispose:)
		)
	)

but it failed that `if` check, and didn't got to call `dispose`.

The `active_object` (for `[0001:2138] longSong2`) shows that `owner = 0001:2224 (sq4)`, therefore, it's probably a problem with `param1` being something different.

ZvikaZ:
Forgot to mention in my previous comment, that I also disabled the narrator's voice on room 16.

In Dosbox, that `clean` method is calling `dispose`, because `longSong2` has no owner (at that point, I mean;  I've `display`ed the `sq4` owner it had before [*] ).

Now the question is, how come in SSCI it has no owner, and in our SCI it still has its original owner?

[*] with:

	(Display
		((gLongSong2 owner?) name?)
		dsCOORD 60 80
		dsCOLOUR clRED
		dsBACKGROUND clWHITE
		dsALIGN alCENTER
	)

ZvikaZ:
Well, a really interesting story is being developed here...

The object that didn't had an owner before, isn't `gLongSong2` (as I thought), but `pukeSFX`!
i.e, we played first `pukeSFX` until end, then played `gLongSong2`, then disposing `pukeSFX` (which shouldn't have any effect) kills somehow  `gLongSong2` (at Dosbox only).

ScummVM's behaviour makes sense now, I have no idea why SSCI did it.
I'm back to Sierra's bug theory...

It can be easily demonstrated by adding `(DoSound sndDISPOSE pukeSFX)`  at `barScript::changeState`  state 17, immediately after `(gLongSong2 number: 886 loop: 1 play:)` - then  `gLongSong2` won't play at all.

BTW, if anyone wants to check this, changing script #1 `(if (== gNumber gNewRoomNumber) (global2 newRoom: 6))` to `(if (== gNumber gNewRoomNumber) (global2 newRoom: 15))` will jump there quickly.

ZvikaZ:
I have asked also at http://sciprogramming.com/community/index.php?topic=1886.0 , but got no response.

It really seems like a weird bug in original Sierra SCI. It also occurs only with this resource pair, and not with other resources...

Bottom line, I think that the current ScummVM SCI behaviour is correct, while SQ4 'by luck' behaved well because of that Sierra's bug. Therefore, it's best to leave ScummVM SCI as is, and fix this SQ4 bug.

ZvikaZ:
We had here a small conflict, fixed it from GitHub GUI, and it got ugly..
Fixing in my local git, and pushing

bluegr:
I just read through this. Nice work! It's quite odd that the sound disposal behaves like that in the intro. Good to know that our implementation is correct here

Merging

PR #2166, opened 2020-03-22
SDL2: silence 'window magic' assert

athrxx:
This is a one line fix to silence the weird 'window && window->magic == &this->window_magic' assert that gets triggered at the start of ScummVM with some SDL2 builds (for me it does that with the debug Win64 version from vcpkg).

This shouldn't have any negative impact.  Calling SDL_GetWindowWMInfo() with a nullptr parameter can't really return anything useful. Nonetheless I commit this as a pull request, so that any interested team member can comment on it.


sev-:
Thanks!

PR #2168, opened 2020-03-27
DIRECTOR: LINGO: Implement list math

djsrv:
This implements Lingo list math. e.g.
```
set a to [1, 2, 3]
set b to [4, 5, 6, 7]
set floats to [4.0, 5.0, 6.0, 7.0]
set strings to ["4", "5", "6", "7"]

put a + b
-- [5, 7, 9]
put a + floats
-- [5.0, 7.0, 9.0]
put a + strings
-- [5.0, 7.0, 9.0]
put a + 1
-- [2, 3, 4]
```

This also fixes a bug in Datum::toString() which caused arrays' contents to be overwritten with strings. Since Datum::toString() is called frequently, this prevented list math from working properly.

### DIFF
--- a/engines/director/lingo/lingo.cpp
+++ b/engines/director/lingo/lingo.cpp
@@ -518,7 +518,8 @@ Common::String *Datum::toString() {
 		for (uint i = 0; i < u.farr->size(); i++) {
 			if (i > 0)
 				*s += ", ";
-			*s += *u.farr->operator[](i).toString();
+			Datum d = u.farr->operator[](i);
+			*s += *d.toString();
 		}
 
 		*s += "]";
--- a/engines/director/lingo/lingo.cpp
+++ b/engines/director/lingo/lingo.cpp
@@ -316,34 +316,22 @@ int Lingo::alignTypes(Datum &d1, Datum &d2) {
 
 	if (d1.type == STRING) {
 		char *endPtr = 0;
-		int i = strtol(d1.u.s->c_str(), &endPtr, 10);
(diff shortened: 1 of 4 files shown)


engines/director/lingo/lingo.cpp
```diff
@@ -518,7 +506,8 @@ Common::String *Datum::toString() {
 		for (uint i = 0; i < u.farr->size(); i++) {
 			if (i > 0)
 				*s += ", ";
-			*s += *u.farr->operator[](i).toString();
+			Datum d = u.farr->operator[](i);
```
sev-: Why? For readability?

djsrv: `u.farr->operator[](i)` returns a reference to the original array item, not just its value. Since `toString()` mutates the datum, `*u.farr->operator[](i).toString()` was mutating every array item into a string datum. Every time I tried to create an array of ints, it would suddenly become an array of strings when the stack was printed and `toString()` was called.

`Datum d = u.farr->operator[](i);` copies just the value into `d`. So then just the copy is cast to a string, and that copy is discarded after.

djsrv: To be honest I think mutating the datum in `toString()`, `toInt()`, etc. is an overall bad idea because it leads to issues like this one. Changing that would require a lot of fixes throughout the interpreter, though, which I haven't done that.

sev-: Nice catch! And I have no other notes, great work, merging.

PR #2169, opened 2020-03-28
ULTIMA8: Uncompress shapes on load for simpler render path

mduggan:
Creating a PR for this since it's a slightly larger change and I'd like some other eyes on it.  I changed it to unpack/un-RLE the shape data on load into simpler buffers, allowing me to simplify the render code a lot as a result.

### DIFF
--- a/engines/ultima/ultima8/graphics/raw_shape_frame.cpp
+++ b/engines/ultima/ultima8/graphics/raw_shape_frame.cpp
@@ -1,0 +1,164 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "ultima/ultima8/misc/pent_include.h"
+
+#include "ultima/ultima8/graphics/raw_shape_frame.h"
+#include "ultima/ultima8/convert/convert_shape.h"
+#include "ultima/ultima8/convert/u8/convert_shape_u8.h"
+#include "ultima/ultima8/filesys/idata_source.h"
+
+#include "common/memstream.h"
+
+namespace Ultima {
+namespace Ultima8 {
+
+/*
+  parse data and fill class
+ */
+RawShapeFrame::RawShapeFrame(const uint8 *data, uint32 size, const ConvertShapeFormat *format,
+                       const uint8 special[256], ConvertShapeFrame *prev) : _line_offsets(0) {
+	// Load it as u8
+	if (!format || format == &U8ShapeFormat || format == &U82DShapeFormat)
+		loadU8Format(data, size);
+	else if (format == &PentagramShapeFormat)
+		loadPentagramFormat(data, size);
+	else if (format == &U8CMPShapeFormat)
+		loadU8CMPFormat(data, size, format, special, prev);
+	else
+		loadGenericFormat(data, size, format);
+}
+
+RawShapeFrame::~RawShapeFrame() {
+	delete [] _line_offsets;
+}
+
+// This will load a u8 style shape 'optimized'.
+void RawShapeFrame::loadU8Format(const uint8 *data, uint32 size) {
+	Common::MemoryReadStream stream(data, size + 8);
+
+	stream.skip(8); // skip header
+	_compressed = stream.readByte();
+	stream.skip(1);
+	_width = stream.readUint16LE();
+	_height = stream.readUint16LE();
+	_xoff = stream.readUint16LE();
+	_yoff = stream.readUint16LE();
+
+	if (_height == 0)
+		return;
+
+	_line_offsets = new uint32[_height];
+
+	for (int32 i = 0; i < _height; i++) {
+		_line_offsets[i] = stream.readUint16LE() - ((_height - i) * 2);
+	}
+
+	_rle_data = data + stream.pos();
+}
+
+// This will load a pentagram style shape 'optimized'.
+void RawShapeFrame::loadPentagramFormat(const uint8 *data, uint32 size) {
+	Common::MemoryReadStream stream(data, size);
+
+	_compressed = stream.readByte();
+	stream.skip(3);
+	_width = stream.readSint32LE();
+	_height = stream.readSint32LE();
+	_xoff = stream.readSint32LE();
+	_yoff = stream.readSint32LE();
+
+	if (_height == 0)
+		return;
+
+	_line_offsets = new uint32[_height];
+
+	for (int32 i = 0; i < _height; i++) {
+		_line_offsets[i] = stream.readSint32LE();
+	}
+
+	_rle_data = data + stream.pos();
+}
+
+// This will load any sort of shape via a ConvertShapeFormat struct
+void RawShapeFrame::loadGenericFormat(const uint8 *data, uint32 size, const ConvertShapeFormat *format) {
+	IBufferDataSource ds(data + format->_bytes_frame_unknown, size);
+
+	_compressed = ds.readX(format->_bytes_frame_compression);
+	_width = ds.readXS(format->_bytes_frame_width);
+	_height = ds.readXS(format->_bytes_frame_height);
+	_xoff = ds.readXS(format->_bytes_frame_xoff);
+	_yoff = ds.readXS(format->_bytes_frame_yoff);
+
+	if (_height == 0)
+		return;
+
+	_line_offsets = new uint32[_height];
+
+	for (int32 i = 0; i < _height; i++) {
+		if (format->_line_offset_absolute) {
+			_line_offsets[i] = ds.readX(format->_bytes_line_offset);
+		} else {
+			_line_offsets[i] = ds.readX(format->_bytes_line_offset) - ((_height - i) * format->_bytes_line_offset);
+		}
+	}
+
+	_rle_data = data + format->_len_frameheader2 + _height * format->_bytes_line_offset;
+}
+
+// This will load an U8-compressed shape
+void RawShapeFrame::loadU8CMPFormat(const uint8 *data, uint32 size, const ConvertShapeFormat *format, const uint8 special[256], ConvertShapeFrame *prev) {
+	IBufferDataSource ds(data, size);
+
+	ConvertShapeFrame f;
+
+	f.ReadCmpFrame(&ds, format, special, prev);
+
+	uint32 to_alloc = f._height + (f._bytes_rle + 3) / 4;
+	_line_offsets = new uint32[to_alloc];
+	_rle_data = reinterpret_cast<uint8 *>(_line_offsets + f._height);
+
+	_compressed = f._compression;
+	_height = f._height;
+	_width = f._width;
+	_xoff = f._xoff;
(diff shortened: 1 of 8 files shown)


engines/ultima/ultima8/graphics/raw_shape_frame.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+// Some macros to make things easier
+#define READ1(data,offset) (data[offset])
+#define READ2(data,offset) (data[offset] + (data[offset+1] << 8))
+#define READ4(data,offset) (data[offset] + (data[offset+1] << 8) + (data[offset+2] << 16) + (data[offset+3] << 24))
+
+// This will load a u8 style shape 'optimized'.
+void RawShapeFrame::LoadU8Format(const uint8 *data, uint32 /*size*/) {
+	_compressed = READ1(data, 8);
+	_width = static_cast<int16>(READ2(data, 10));
```
sev-: We have our `FROM_LE_16(ptr)` etc, you could use that. Or even better, write the following code:
```
   Common::MemoryReadStream str(data, size);

   stream->skip(8); // skip header (?)
   _compressed = stream->readByte();
   _width = stream->readUint16LE();
   _height = stream->readUint16LE();
   _xoff = stream->readUint16LE();
   _yoff = stream->readUint16LE();
```

It is much more readable and uses our helper utilities.

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
```diff
@@ -0,0 +1,162 @@
...
+		data += 2;
+	}
+
+	_rle_data = data;
+}
+
+// This will load a pentagram style shape 'optimzed'.
+void RawShapeFrame::LoadPentagramFormat(const uint8 *data, uint32 /*size*/) {
+	_compressed = READ1(data, 0);
+	_width = static_cast<int32>(READ4(data, 4));
```
sev-: Same thing here with the stream reading

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
```diff
@@ -0,0 +1,162 @@
...
+	_xoff = ds.readXS(format->_bytes_frame_xoff);
+	_yoff = ds.readXS(format->_bytes_frame_yoff);
+
+	if (_height == 0) return;
+
+	_line_offsets = new uint32[_height];
+
+	if (format->_line_offset_absolute) for (int32 i = 0; i < _height; i++) {
+			_line_offsets[i] = ds.readX(format->_bytes_line_offset);
+		}
```
sev-: Formatting is wrong here

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
```diff
@@ -0,0 +1,162 @@
...
+	else for (int32 i = 0; i < _height; i++) {
```
sev-: Formatting is wrong here

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
```diff
@@ -0,0 +1,162 @@
...
+			_line_offsets[i] = ds.readX(format->_bytes_line_offset) - ((_height - i) * format->_bytes_line_offset);
+		}
+
+	_rle_data = data + format->_len_frameheader2 + _height * format->_bytes_line_offset;
+}
+
+// This will load an U8-compressed shape
+void RawShapeFrame::LoadU8CMPFormat(const uint8 *data, uint32 size, const ConvertShapeFormat *format, const uint8 special[256], ConvertShapeFrame *prev) {
```
sev-: These methods are not named after our code naming conventions. Only constants, classes and namespaces can start with an upper case letter. It should be `RawShapeFrame::loadU8CMPFormat()`

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
```diff
@@ -0,0 +1,162 @@
...
+	_line_offsets = new uint32[to_alloc];
+	_rle_data = reinterpret_cast<uint8 *>(_line_offsets + f._height);
+
+	_compressed = f._compression;
+	_height = f._height;
+	_width = f._width;
+	_xoff = f._xoff;
+	_yoff = f._yoff;
+
+	Std::memcpy(_line_offsets, f._line_offsets, f._height * 4);
```
sev-: Your `Std::memcpy` class is just a wrapper over `memcpy`. I'd drop that altogether.

engines/ultima/ultima8/graphics/raw_shape_frame.h
```diff
@@ -0,0 +1,79 @@
...
+
+	uint32              _compressed;
+	int32               _width, _height;
+	int32               _xoff, _yoff;
+
+	uint32              *_line_offsets;      // Note these are offsets into rle_data
+	const uint8         *_rle_data;
+
+	// This will load a u8 style shape 'optimized'.
+	void LoadU8Format(const uint8 *data, uint32 size);
```
sev-: Yep, these must be renamed to start with a lowercase lettr.

engines/ultima/ultima8/graphics/shape.cpp
```diff
@@ -36,63 +37,56 @@ DEFINE_RUNTIME_CLASSTYPE_CODE_BASE_CLASS(Shape)
...
-	uint8 *d = new uint8[this->_size];
-	this->_data = d;
-	src->read(d, this->_size);
+	loadFrames(data, size, format);
 
+	delete[] data;
+}
+
+Shape::~Shape() {
+	for (unsigned int i = 0; i < _frames.size(); ++i)
```
sev-: Consider using `uint` for brevity, we have that type defined.

engines/ultima/ultima8/graphics/shape.cpp
```diff
@@ -36,63 +37,56 @@ DEFINE_RUNTIME_CLASSTYPE_CODE_BASE_CLASS(Shape)
...
-		format = DetectShapeFormat(_data, _size);
+		format = DetectShapeFormat(data, size);
 
 	if (!format) {
 		// Should be fatal?
 		perr << "Error: Unable to detect shape format" << Std::endl;
 		return;
 	}
 
+	Std::vector<RawShapeFrame *> rawframes;
```
sev-: Since it is a new code, consider using `Common::Array` directly and avoid the wrapper.

engines/ultima/ultima8/graphics/shape.cpp
```diff
@@ -36,63 +37,56 @@ DEFINE_RUNTIME_CLASSTYPE_CODE_BASE_CLASS(Shape)
...
-		LoadGenericFormat(_data, _size, format);
-}
+		rawframes = LoadGenericFormat(data, size, format);
 
-Shape::~Shape() {
-	for (unsigned int i = 0; i < _frames.size(); ++i)
-		delete _frames[i];
-
-	delete[] const_cast<uint8 *>(_data);
+	for (unsigned int i = 0; i < rawframes.size(); i++) {
```
sev-: `uint`

engines/ultima/ultima8/graphics/shape_frame.cpp
```diff
@@ -23,128 +23,60 @@
...
-		LoadU8Format(data, size);
-	else if (format == &PentagramShapeFormat)
-		LoadPentagramFormat(data, size);
-	else if (format == &U8CMPShapeFormat)
-		LoadU8CMPFormat(data, size, format, special, prev);
-	else
-		LoadGenericFormat(data, size, format);
-}
+ShapeFrame::ShapeFrame(const RawShapeFrame *rawframe)
+: _xoff(rawframe->_xoff), _yoff(rawframe->_yoff),
```
sev-: Missing indentation on the continuation strings.

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
mduggan: Thanks! Those macros were in the pentagram code but I also wanted to get rid of them so I updated with your suggestion.

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
mduggan: Done.

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
mduggan: Thanks, I copied that from the old code and didn't notice.. fixed.

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
mduggan: Also leftover from before - but no reason to duplicate the existing error, so I updated.

engines/ultima/ultima8/graphics/raw_shape_frame.cpp
mduggan: Yeah, also leftover that I should go through and clean all those up.  For now I have fixed this branch so it's not adding new ones at least.

engines/ultima/ultima8/graphics/shape.cpp
mduggan: Thanks for the suggestion, done.

engines/ultima/ultima8/graphics/shape.cpp
mduggan: Done, another thing that I should go through and clean up everywhere.

engines/ultima/ultima8/graphics/shape_frame.cpp
mduggan: Ok, this mistake was all me :)

dreammaster:
Looks good to me too. And thanks Sev for reviewing the pull request as well.

dreammaster:
Hmm.. realised that I hadn't actually tried to compile and run it yet. Did so and I'm now getting a brunch of crashes in ShapeFrame. Not to be merged yet.

dreammaster:
It seems like the last frame of the u8mouse.shp file, frame #41. I did some basic investigation, and it looks like the frame is 36 bytes in size, but size the height is 12, it's expecting the frame to be 42 bytes. In the file itself, it looks like there's 44 bytes remaining in the file so it worked, but since you're now wrapping the data acess in a MemoryReadStream, the read past end isn't giving the data anymore.

I've run out of time tonight to try a patch to fix it, but if you have time tomorrow, I'd suggest doing a hack in the frame loop to set the frame size for the last frame of a file to be the overall size_ - the frame offset. And put a // WORKAROUND: comment above the change mentioning that it's to fix invalid data in the u8mouse.shp, just so it's marked for posterity why it's present. Or it you want to wait, I'll tackle it myself tomorrow or Tuesday night and then merge it back to master.

mduggan:
Thanks for the headsup @dreammaster, I should have run with address sanitizer to notice this. I added this code to the frame loading loop:
```
		if (i == framecount - 1 && framesize < size_ - frameoffset) {
			uint32 remaining = size_ - frameoffset;
			perr << "WARN: Overriding framesize " << framesize << " with remaining bytes "
				 << remaining << " on last frame." << Std::endl;
			framesize = remaining;
		}
```
Interestingly it triggers on a whole bunch of shapes, and it's always 8 bytes short.  It suggests there's something more generally wrong in the interpretation of the data, but it seems to work fine as a workaround.  I'll dig in a bit more before I merge this to see if I can get a better understanding of what exactly is wrong and fix it properly.

mduggan:
I realised that the 8 bytes corresponds to the `_bytes_frame_unknown` in the shape format description. Instead of skipping those bytes at the start, I just offset the pointer by that many bytes and assumed they were not part of the "size", which seems to always work.  I think this is probably good to merge now and will do so unless there's anything else.

wjp:
What is the impact on rendering performance?

The original point of the use of RLE was to make masked/transparent rendering faster, and the perspective of U8 means that large parts of many shapes are impacted by this. Of course systems now are very different than in the 90's, so maybe this is no concern, but this should be evaluated, especially on our slower platforms.

mduggan:
@wjp great question, I had the same concern.  I did a pretty unscientific comparison of cpu usage on my machine (a 5-year-old i5 macbook pro) for the 2 versions, and I couldn't really detect any difference.  I will try to do a bit more comparison though before I merge though.

mduggan:
I did a bit of profiling.  The un-RLE'd version spends slightly less % of its time in `Gump::Paint` than the `master` branch, so it seems to be a very slight improvement.  I would guess this is due to simpler branches allowing better prediction, but I haven't investigated too deeply.  At least I can be fairly sure it's not worse.

Master Branch:
profile_u8_master

This branch:
profile_u8_unrle

PR #2171, opened 2020-03-31
GRIFFON: Simplify rendering calls

ixote:
GRIFFON: Simplify rendering calls

Griffon takes many setAlpha calls in its loops to set Alpha . That wastes 
CPU resources and make the game slow. Change uilt call's params will
help.

### DIFF
--- a/engines/griffon/dialogs.cpp
+++ b/engines/griffon/dialogs.cpp
@@ -256,7 +256,6 @@ void GriffonEngine::configMenu() {
 
 	int ticks1 = _ticks;
 
-	_cloudImg->setAlpha(128, true);
 
 	do {
 		_videoBuffer->fillRect(Common::Rect(0, 0, _videoBuffer->w, _videoBuffer->h), 0);
@@ -267,14 +266,14 @@ void GriffonEngine::configMenu() {
 		rcDest.setWidth(320);
 		rcDest.setHeight(240);
 
-		_cloudImg->blit(*_videoBuffer, 0, 0, Graphics::FLIP_NONE, &rcDest);
+		_cloudImg->blit(*_videoBuffer, 0, 0, Graphics::FLIP_NONE, &rcDest, TS_ARGB(128, 255, 255, 255));
 
 		rcDest.left = 256;
 		rcDest.top = 192;
 		rcDest.setWidth(320);
 		rcDest.setHeight(240);
 
-		_cloudImg->blit(*_videoBuffer, 0, 0, Graphics::FLIP_NONE, &rcDest);
+		_cloudImg->blit(*_videoBuffer, 0, 0, Graphics::FLIP_NONE, &rcDest, TS_ARGB(128, 255, 255, 255));
 
 		configwindow->blit(*_videoBuffer);
 
@@ -467,7 +466,6 @@ void GriffonEngine::configMenu() {
 		g_system->delayMillis(10);
 	} while (!_shouldQuit && !exitMenu && _gameMode != kGameModeNewGame && _gameMode != kGameModeLoadGame);
 
-	_cloudImg->setAlpha(64, true);
 
 	configwindow->free();
 	_itemTicks = _ticks + 210;
@@ -559,7 +557,6 @@ void GriffonEngine::saveLoadNew() {
 	int ticks1 = _ticks;
 	int tickPause = _ticks + 150;
 
-	_cloudImg->setAlpha(128, true);
 
 	renderSaveStates();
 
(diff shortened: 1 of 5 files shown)


engines/griffon/dialogs.cpp
```diff
@@ -256,7 +256,7 @@ void GriffonEngine::configMenu() {
 
 	int ticks1 = _ticks;
 
-	_cloudImg->setAlpha(128, true);
+	// _cloudImg->setAlpha(128, true);
```
sev-: Since you're replicating the behaviour it's better to remove these lines than comment out.

engines/griffon/dialogs.cpp
```diff
@@ -267,14 +267,14 @@ void GriffonEngine::configMenu() {
 		rcDest.setWidth(320);
 		rcDest.setHeight(240);
 
-		_cloudImg->blit(*_videoBuffer, 0, 0, Graphics::FLIP_NONE, &rcDest);
+		_cloudImg->blit(*_videoBuffer, 0, 0, Graphics::FLIP_NONE, &rcDest,TS_ARGB(128,255,255,255));
```
sev-: The new code piece does not conform to our formatting guidelines: there must be spaces after commas.

engines/griffon/resources.cpp
```diff
@@ -293,26 +293,36 @@ void GriffonEngine::loadMap(int mapnum) {
 					rcDest.setWidth(16);
 					rcDest.setHeight(16);
 
+					int flag=0;
```
sev-: The line does not conform to our code formatting guidelines. It must be `int flag = 0;` Note spaces around the operator.

engines/griffon/resources.cpp
```diff
@@ -293,26 +293,36 @@ void GriffonEngine::loadMap(int mapnum) {
...
-								_tiles[curtilel]->setAlpha(192, true);
+								flag=2;
+								// _tiles[curtilel]->setAlpha(192, true);
 							}
 						}
 					}
-
-					_tiles[curtilel]->blit(*_mapBg, rcDest.left, rcDest.top, Graphics::FLIP_NONE, &rcSrc);
-					_tiles[curtilel]->setAlpha(255, true);
+					if(flag==1){
```
sev-: All keywords must have spaces after them, so are operators and curly braces. `if (flag == 1) {`

engines/griffon/resources.cpp
```diff
@@ -293,26 +293,36 @@ void GriffonEngine::loadMap(int mapnum) {
...
+						_tiles[curtilel]->blit(*_mapBg, rcDest.left, rcDest.top, Graphics::FLIP_NONE, &rcSrc,TS_ARGB(128,255,255,255));
+					}
+					else if (flag ==2){
```
sev-: The `else` statement must stick to the closing bracket as per our code formatting guidelines.

engines/griffon/resources.cpp
```diff
@@ -293,26 +293,36 @@ void GriffonEngine::loadMap(int mapnum) {
...
 					rcDest.setHeight(16);
 
+					int flag=0;
 					if (l == 2 && curtilel == 1) {
 						for (int ff = 0; ff <= 5; ff++) {
 							int ffa = 20 * 5 - 1 + ff * 20;
 							int ffb = 20 * 5 + 4 + ff * 20;
 							if (curtile > ffa && curtile < ffb) {
-								_tiles[curtilel]->setAlpha(128, true);
+								flag=1;
```
sev-: Why don't you just instead of this flag do something like `alphaValue = 128` and then use it directly?

ixote: According to the context, I think that his alpha rendering is triggered under certain conditions, and I hope that the logic of the source code is not changed when possible.

engines/griffon/resources.cpp
```diff
@@ -293,26 +293,31 @@ void GriffonEngine::loadMap(int mapnum) {
...
 							if (curtile == ffa) {
-								_tiles[curtilel]->setAlpha(192, true);
+								flag = 2;
 							}
 						}
 					}
-
-					_tiles[curtilel]->blit(*_mapBg, rcDest.left, rcDest.top, Graphics::FLIP_NONE, &rcSrc);
-					_tiles[curtilel]->setAlpha(255, true);
+					if(flag == 1){
```
sev-, line 314: Formatting is still incorrect, no space before opening curly brace.

sev-: Please, simplify this code by using `alpha = <vale>` instead of flag. Just think about this logically. You change flag only under certain conditions, and otherwise, you're using previous flag value. Exactly the same thing will be with `alpha` but you will avoid those three calls to `blit()` at the bottom.

ixote: i see, i will modify it soon and fix my git usage  errors

sev-:
Great, thank you, all is fine, merging!

PR #2176, opened 2020-04-02
SCI: suggest patching if GK2 version is 1.0

ZvikaZ:
Refactored `VERSION` reading to one function,
check if GK2 is version 1.0 and suggest patching - implementing
Sev's idea on end of #11400.

### DIFF
--- a/engines/sci/sci.cpp
+++ b/engines/sci/sci.cpp
@@ -436,7 +436,21 @@ Common::Error SciEngine::run() {
 	}
 
 	if (getGameId() == GID_GK2 && ConfMan.getBool("subtitles") && !_resMan->testResource(ResourceId(kResourceTypeSync, 10))) {
-		suggestDownloadGK2SubTitlesPatch();
+		suggestDownloadPatch(_("GK2 has a fan made subtitles, available thanks to the good persons at SierraHelp.\n\n"
+			"Installation:\n"
+			"- download http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip\n"
+			"- extract zip file\n"
+			"- no need to run the .exe file\n"
+			"- extract the .exe file with a file archiver, like 7-zip\n"
+			"- create a PATCHES subdirectory inside your GK2 directory\n"
+			"- copy the content of GK2Subtitles\\SUBPATCH to the PATCHES subdirectory\n"
+			"- replace files with similar names\n"
+			"- restart the game\n"), "http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip");
+	}
+
+	if (getGameId() == GID_GK2 && getGameVersion() == "1.0") {
+		suggestDownloadPatch(_("Your current GK2 version is 1.0 , which is known to suffer from few problems.\n"
+			"You might want to install the official Sierra patch, available at http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2(GK2PAT).zip\n"), "http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2(GK2PAT).zip");
 	}
 
 	runGame();
@@ -509,32 +523,16 @@ bool SciEngine::gameHasFanMadePatch() {
 	return false;
 }
 
-void SciEngine::suggestDownloadGK2SubTitlesPatch() {
+void SciEngine::suggestDownloadPatch(const Common::String &message, const char *url) {
 	const char* altButton;
-	Common::String downloadMessage;
 
-	if (g_system->hasFeature(OSystem::kFeatureOpenUrl)) {
+	if (g_system->hasFeature(OSystem::kFeatureOpenUrl)) 
 		altButton = _("Download patch");
-		downloadMessage = _("(or click 'Download patch' button. But note - it only downloads, you will have to continue from there)\n");
-	}
-	else {
+	else
 		altButton = nullptr;
-		downloadMessage = "";
-	}
 
-	int result = showScummVMDialog(_("GK2 has a fan made subtitles, available thanks to the good persons at SierraHelp.\n\n"
-		"Installation:\n"
-		"- download http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip\n" +
-		downloadMessage +
-		"- extract zip file\n"
-		"- no need to run the .exe file\n"
-		"- extract the .exe file with a file archiver, like 7-zip\n"
-		"- create a PATCHES subdirectory inside your GK2 directory\n"
-		"- copy the content of GK2Subtitles\\SUBPATCH to the PATCHES subdirectory\n"
-		"- replace files with similar names\n"
-		"- restart the game\n"), altButton, false);
+	int result = showScummVMDialog(message, altButton, false);
 	if (!result) {
-		char url[] = "http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip";
 		g_system->openUrl(url);
 	}
 }
@@ -778,6 +776,17 @@ Console *SciEngine::getSciDebugger() {
 
 const char *SciEngine::getGameIdStr() const {
 	return _gameDescription->gameId;
+}
+
+const Common::String SciEngine::getGameVersion() {
+	Common::String gameVersion = "N/A";
+
+	Common::File versionFile;
+	if (versionFile.open("VERSION")) {
+		gameVersion = versionFile.readLine();
+		versionFile.close();
+	}
+	return gameVersion;
 }
 
 Common::Language SciEngine::getLanguage() const {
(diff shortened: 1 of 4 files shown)


engines/sci/sci.cpp
```diff
@@ -436,7 +436,21 @@ Common::Error SciEngine::run() {
 	}
 
 	if (getGameId() == GID_GK2 && ConfMan.getBool("subtitles") && !_resMan->testResource(ResourceId(kResourceTypeSync, 10))) {
-		suggestDownloadGK2SubTitlesPatch();
+		suggestDownloadPatch(_("GK2 has a fan made subtitles, available thanks to the good persons at SierraHelp.\n\n"
```
Thunderforge, line 439: My personal preference would be to spell out "Gabriel Knight 2" instead of "GK2", but I'm not stuck on the idea.

Thunderforge, line 439: > GK2 has a fan made subtitles, available thanks to the good persons at SierraHelp

This sounds strange to my ears as a native-speaker of American English. Two things:

* "a […] subtitles": the first word implies singular, the second is plural. I think it should just be "GK2 has fan made subtitles"
* The good persons at SierraHelp" just sounds awkward. Maybe just "hosted at SierraHelp"?

engines/sci/sci.cpp
```diff
@@ -436,7 +436,21 @@ Common::Error SciEngine::run() {
...
+			"Installation:\n"
+			"- download http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip\n"
```
Thunderforge, line 441: Does ScummVM allow us to make this a hyperlink instead of a bare URL? So it would say something like

> - download [the subtitles](http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip)

engines/sci/sci.cpp
```diff
@@ -436,7 +436,21 @@ Common::Error SciEngine::run() {
...
+			"- no need to run the .exe file\n"
+			"- extract the .exe file with a file archiver, like 7-zip\n"
+			"- create a PATCHES subdirectory inside your GK2 directory\n"
+			"- copy the content of GK2Subtitles\\SUBPATCH to the PATCHES subdirectory\n"
+			"- replace files with similar names\n"
+			"- restart the game\n"), "http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip");
+	}
+
+	if (getGameId() == GID_GK2 && getGameVersion() == "1.0") {
+		suggestDownloadPatch(_("Your current GK2 version is 1.0 , which is known to suffer from few problems.\n"
```
Thunderforge, line 452: > which is known to suffer from few problems

This makes it sound like version 1.0 only has a few problems, which isn't the case. Either it should be "a few problems" (don't ask me why "few problems" and "a few problems" have different grammatical meaning in English) or rephrase to something like "which has some problems" or "which has known issues".

engines/sci/sci.cpp
```diff
@@ -436,7 +436,21 @@ Common::Error SciEngine::run() {
...
+			"You might want to install the official Sierra patch, available at http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2(GK2PAT).zip\n"), "http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2(GK2PAT).zip");
```
Thunderforge, line 453: As I recall, installing the patch means that save games for older versions are invalidated. Maybe it's worth adding that so that players don't start playing, then patch it and be surprised that they have to restart?

engines/sci/sci.cpp
```diff
@@ -509,32 +523,16 @@ bool SciEngine::gameHasFanMadePatch() {
 	return false;
 }
 
-void SciEngine::suggestDownloadGK2SubTitlesPatch() {
+void SciEngine::suggestDownloadPatch(const Common::String &message, const char *url) {
```
Thunderforge, line 526: I like the method rename here to make it more reusable!

engines/sci/sci.cpp
```diff
@@ -436,7 +436,21 @@ Common::Error SciEngine::run() {
 	}
 
 	if (getGameId() == GID_GK2 && ConfMan.getBool("subtitles") && !_resMan->testResource(ResourceId(kResourceTypeSync, 10))) {
-		suggestDownloadGK2SubTitlesPatch();
+		suggestDownloadPatch(_("GK2 has a fan made subtitles, available thanks to the good persons at SierraHelp.\n\n"
+			"Installation:\n"
+			"- download http://www.sierrahelp.com/Files/Patches/GabrielKnight/GK2Subtitles.zip\n"
+			"- extract zip file\n"
+			"- no need to run the .exe file\n"
+			"- extract the .exe file with a file archiver, like 7-zip\n"
```
Thunderforge, line 444: [According to the official website](https://www.7-zip.org), it's "7-Zip", not "7-zip". I think that should be corrected. Also, isn't it a "file unarchiver" instead of "archiver" in this case, since we're extracting not packing?

Thunderforge, line 444: 7-Zip is Windows-only, and I wish there was a good way that we could suggest tools for a number of platforms. But I think that's difficult. If there's some way that we can be more platform-independent in these instructions, that would be great, but I won't hold up this PR over that.

sluicebox:
:horse_racing:  Whoa whoa whoa whoa. Whoa. =)

Well first off, the version string isn't "1.0" it's "1.00", so this new message box will never appear, but lets ignore that.

This is the warning when you try to load the bug reporter's save game in 1.00:

`WARNING: Save game was created for game version 1.1, but the current game version is 1.00!`

Does that sound like we should be warning users with a version file that says 1.00?

This PR is in response to ticket 11400. Ticket 11400's bug reporter admitted they had a hot mess of files downloaded from a torrent or something. The reporter never said they had version 1.00. The reporter couldn't even provide a save game that loads in 1.00, which @dafioram mentioned. The uploaded save games have the version string "1.1". And so we've concluded from this pirate setup we can't reproduce, which the reporter solved by replacing with files from original media, and didn't involve 1.00, that 1.00 is broken and we need a mandatory message box? Oops!

But let's assume this is a real GK2 bug, which it *might* be...

We prefer to fix original bugs so that all versions can be played wonderfully by ScummVM.

If someone takes the time to load their old game media in ScummVM then it should just work. Most won't have any idea what version they have or that there even are versions. We shouldn't ask them to do any work we can't do ourselves. This is especially true when it comes to Sierra patches. They tend to fix a small number of bugs that most players are unlikely to encounter in the first place. A message box like this implies that the game is going to go badly for them in ScummVM when it almost certainly won't, otherwise we'd already have warnings like this for ever single game. Installing Sierra patches is not a trivial task for many players. Even with all the instructions some will mess it up, and we have the legion of invalid tickets to show for it.

I consider any message box warning when starting a game an extreme measure for extreme circumstances. Otherwise it's an admission of failure. I only know of one SCI game that warranted this at startup: Conquests of the Longbow 1.0 which had around a 40% random chance of every game being impossible to complete. That's the bar to clear for a broken-game message. That message lasted something like seven years, and check out at what I went through to remove it: https://github.com/scummvm/scummvm/pull/1576

I don't see any mention of all the GK2 script patches I wrote that backport all the fixes from later versions to 1.00 and 1.1 so that patches are unnecessary. That's... a pretty big detail to miss before asking users to go get a patch. =) I might as well have not done all that if we're just going to say "go find a better version". We should have a firm understanding of what we're asking users to do.

I see patch notes referenced in the ticket as part of the troubleshooting. Sierra patch notes can't be solely relied upon to determine what's in a patch. You have to reverse engineer it yourself. Some fixes are script/resource bugs but just as many are interpreter fixes, which are irrelevant to ScummVM. Sierra patch notes and anything written about them by fans and not reverse engineers should be read as well-intentioned rumors. Patch notes omit changes and the GK2 patch notes even list a fix they didn't include.

Adventure game bugs tend to be edge cases; they don't happen all the time. When someone says "this room in a popular 1990s game always crashes" that's suspect because someone would have noticed. "I upgraded and started over and it didn't happen" doesn't tell us that the bug was fixed. Starting over would have been enough without the upgrade, assuming the files weren't corrupt. There still might be a bug here but we'll have to reproduce it first.

BTW, plenty of patches don't patch the VERSION file, and Mac versions often exclude them, so this is not a good way to programmatically detect versions. It's just for display. 

I don't mean to rag on this, it's just a good opportunity to go over some SCI stuff and mouth off, which are both things I love dearly. I also happen to have audited all the GK2 patches and version differences back when I was subsided. I'll try taking a look using real 1.1 media later. But for any of the reasons above, we should not be adding this message box, and it's kind of weird it got this far without reverse engineering or reproduction of the underlying bug.

Thunderforge:
> But for any of the reasons above, we should not be adding this message box, and it's kind of weird it got this far without reverse engineering or reproduction of the underlying bug.

So close the PR without merging and don’t reopen unless it can be demonstrated that the issue really exists?

ZvikaZ:
@sluicebox , I really appreciate all the great works that you've invested in all Sierra games, including of course, GK2.
As you said, I haven't investigated enough the background of the acclaimed "1.0(0 ;-) ) problem", and therefore, this PR isn't needed.

Sorry...

Also, thanks for the detailed explanation.

@Thunderforge , thanks for the improvement suggestions. I know that my English grammar isn't perfect, and please feel free to suggest fixes in the future as well.

Anyway, I'm closing this.

Thunderforge:
@ZvikaZ Thanks for understanding. Looking forward to future contributions from you! If you have questions about future work, you can always discuss it on IRC/Discord.

PR #2177, opened 2020-04-02
GUI: added `warningGuiAndConsole`

ZvikaZ:
Add `warningGuiAndConsole` function to send warnings with
regular `warning` and to GUI - if available.

Used for example in SCI failures to save/load games, which used only
console `warning`s - and it's not clear enough to users.

### DIFF
--- a/gui/message.cpp
+++ b/gui/message.cpp
@@ -22,6 +22,7 @@
 
 #include "common/str.h"
 #include "common/system.h"
+#include "common/translation.h"
 #include "gui/message.h"
 #include "gui/gui-manager.h"
 #include "gui/ThemeEval.h"
@@ -119,6 +120,25 @@ void TimedMessageDialog::handleTickle() {
 	MessageDialog::handleTickle();
 	if (g_system->getMillis() > _timer)
 		close();
+}
+
+void warningGuiAndConsole(const char *s, ...) {
+	Common::String output;
+	va_list va;
+
+	va_start(va, s);
+	output = Common::String::vformat(Common::TranslationManager::getLastOrigMessage().c_str(), va);
+	va_end(va);
+
+	warning(output.c_str());
+
+	if (GuiManager::hasInstance()) {
+		va_start(va, s);
+		output = Common::String::vformat(s, va);
+		va_end(va);
+		MessageDialog dialog(output, "OK");
+		dialog.runModal();
+	};
 }
 
 } // End of namespace GUI
--- a/common/translation.cpp
+++ b/common/translation.cpp
@@ -107,11 +107,13 @@ void TranslationManager::setLanguage(const String &lang) {
 	}
 }
 
-const char *TranslationManager::getTranslation(const char *message) const {
+const char *TranslationManager::getTranslation(const char *message) {
 	return getTranslation(message, nullptr);
 }
 
-const char *TranslationManager::getTranslation(const char *message, const char *context) const {
+const char *TranslationManager::getTranslation(const char *message, const char *context) {
+	_lastOrigMessage = message;
+
 	// If no language is set or message is empty, return msgid as is
 	if (_currentTranslationMessages.empty() || *message == '\0')
 		return message;
@@ -175,11 +177,17 @@ String TranslationManager::getCurrentLanguage() const {
 	return _langs[_currentLang];
 }
 
-String TranslationManager::getTranslation(const String &message) const {
+String TranslationManager::_lastOrigMessage = "";
+
+String TranslationManager::getLastOrigMessage() {
+	return _lastOrigMessage;
+}
+
+String TranslationManager::getTranslation(const String &message) {
 	return getTranslation(message.c_str());
 }
 
-String TranslationManager::getTranslation(const String &message, const String &context) const {
+String TranslationManager::getTranslation(const String &message, const String &context) {
 	return getTranslation(message.c_str(), context.c_str());
 }
 
--- a/common/translation.h
+++ b/common/translation.h
@@ -111,25 +111,14 @@ public:
 	 * message. In case the message isn't found in the translation catalog,
 	 * it returns the original untranslated message.
 	 */
-	const char *getTranslation(const char *message) const;
+	const char *getTranslation(const char *message);
 
 	/**
 	 * Returns the translation into the current language of the parameter
 	 * message. In case the message isn't found in the translation catalog,
 	 * it returns the original untranslated message.
 	 */
-	String getTranslation(const String &message) const;
-
-	/**
-	 * Returns the translation into the current language of the parameter
-	 * message. In case the message isn't found in the translation catalog,
-	 * it returns the original untranslated message.
-	 *
-	 * If a translation is found for the given context it will return that
-	 * translation, otherwise it will look for a translation for the same
-	 * massage without a context or with a different context.
-	 */
-	const char *getTranslation(const char *message, const char *context) const;
+	String getTranslation(const String &message);
 
 	/**
 	 * Returns the translation into the current language of the parameter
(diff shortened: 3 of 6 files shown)


bgK:
The issue with this is translations. We want GUI messages to be translated, and console messages to be in english. One way to fix it would be to mark the message as translatable using `_s()` at the call site, and then use `TransMan.getTranslation()` when calling `GUI::MessageDialog`. However it's quite error prone, as we use `_()` to mark strings at translatable and translate them most of the time when passing strings to the GUI.

Mostly unrelated, but perhaps it's time to revise the policy of having the console messages in English? I mean it seems to be possible to solve the encoding issues. Somewhat recent versions of Windows allow setting the console codepage to UTF-8, and it would be possible to convert the strings from the current ScummVM translation charset to UTF-8.

ZvikaZ:
1.
I have addressed the translation issue in my latest update - `warning` is called with original `s`, while `MessageDialog` is using `_(s)`.

2.
There was a discussion at Discord lately regarding this, and the argument to keep console messages untranslated is to ease debugging. We want to make sure that devs can read users' log files.

bgK:
> I have addressed the translation issue in my latest update - `warning` is called with original `s`, while `MessageDialog` is using `_(s)`.

Thank you, however for the translations to work, `_s()` needs to be placed around the translatable strings at the calling sites of the function you introduced. It is a marker used to extract the strings that need to be translated. `TransMan.getTranslation()` should be then used to translate the strings when passing them to the GUI (without marking for extraction).

gui/message.cpp
```diff
@@ -26,6 +26,7 @@
 #include "gui/gui-manager.h"
 #include "gui/ThemeEval.h"
 #include "gui/widget.h"
+#include <common/translation.h>
```
bgK, line 29: Please use `""` style includes for in-project files instead of `<>`.

ZvikaZ: Agreed.
I added this line with some VS auto magic - weird that it used the `<>` style.

ZvikaZ:
> Thank you, however for the translations to work, `_s()` needs to be placed around the translatable strings at the calling sites of the function you introduced. It is a marker used to extract the strings that need to be translated. `TransMan.getTranslation()` should be then used to translate the strings when passing them to the GUI (without marking for extraction).

I'm afraid that it'd be hard to convince people to call `warningGuiAndConsole(_s("something"))` instead of `warningGuiAndConsole(_("something"))`. Or, to make sure that everybody remembers in the future that `warningGuiAndConsole` needs the `_s()` instead of the usual `_()`.

Therefore, I've tried another approach.
We can call `warningGuiAndConsole(_("something"))`, as feels natural.

I've added a small "memory" to the translation manager, that remembers the original, untranslated string. Thus `warningGuiAndConsole` can use that for its console warning.

ZvikaZ:
A side note - there is a problem in GitHub-Travis integration. It says here `Some checks haven’t completed yet`, and `continuous-integration/travis-ci/pr Pending — The Travis CI build is in progress`.

However, when clicking on the `Details` link - https://travis-ci.org/github/scummvm/scummvm/builds/671193186?utm_source=github_status&utm_medium=notification we can see that travis builds have passed.

bgK:
> I'm afraid that it'd be hard to convince people to call warningGuiAndConsole(_s("something")) instead of warningGuiAndConsole(_("something")). Or, to make sure that everybody remembers in the future that warningGuiAndConsole needs the _s() instead of the usual _().

Yes, agreed.

> I've added a small "memory" to the translation manager, that remembers the original, untranslated string. Thus warningGuiAndConsole can use that for its console warning.

Don't you think that's quite a hack, and seems rather fragile?

athrxx:
Whats the point of throwing a warning and display a dialog at the same time?
If you have located a specific warning in SCI which you think is insufficient then you might just replace that with a dialog.

ZvikaZ:
> Don't you think that's quite a hack, and seems rather fragile?

Well, I don't think so...
Unless, there's a possibility of race condition.
Is it possible that while function A is calling `warningGuiAndConsole(foo)` there is a function B that calls `_(bar)`, and control passes from from A to B in the middle of `warningGuiAndConsole`?

That's the only scenerio that I expect problems. But is it realistic?

ZvikaZ:
> If you have located a specific warning in SCI which you think is insufficient then you might just replace that with a dialog.

Good point.
See in this PR, my changes to `engines/sci/engine/kfile.cpp`.
All these save/load failures, are currently `warning`s. I think that it's not enough, and the user should get the message in a GUI dialog box.

But, is GUI dialog box enough? Don't we want to leave trace in the log file?
When the user will open a bug, or request support, because of problems in the savefile, we could see the problem immediately in the log file with the `warning`s. 
Isn't this better than just showing the user a GUI message?

bgK:
> > Don't you think that's quite a hack, and seems rather fragile?
> 
> Well, I don't think so...
> Unless, there's a possibility of race condition.
> Is it possible that while function A is calling `warningGuiAndConsole(foo)` there is a function B that calls `_(bar)`, and control passes from from A to B in the middle of `warningGuiAndConsole`?
> 
> That's the only scenerio that I expect problems. But is it realistic?

Let's say I write some code like this;
```c++
Common::String message = _("Unable to write the save file.") + " ";
switch (errorCode) {
case kDiskFullError:
    message += _("The destination disk is full.");
default:
    message += _("Unknown error.");
}

GUI::warningGuiAndConsole(message);
```

What is going to happen?

ZvikaZ:
> What is going to happen?

OK, I Agree...

So, maybe @athrxx is right? It's better to give up this mechanism, and just choose either a console warning, or (*) a GUI dialog box?
And in the occasion that someone wants both of them, do that manually?

(*) XOR

In that case, is it agreed that the SCI save/load failures I changed in that PR should be just GUI dialog boxes, without console warnings?

bgK:
> So, maybe @athrxx is right? It's better to give up this mechanism, and just choose either a console warning, or (*) a GUI dialog box?
> And in the occasion that someone wants both of them, do that manually?

That's what we usually do.

> In that case, is it agreed that the SCI save/load failures I changed in that PR should be just GUI dialog boxes, without console warnings?

That's how errors are handled when saving from the GMM, so I suppose it's fine: https://github.com/scummvm/scummvm/blob/bd992479b331941594194520580704fd1b0e4a85/engines/dialogs.cpp#L222

A simple improvement would be to allow users to copy the error message shown in the error dialog to the clipboard, for bug reporting purposes.

ZvikaZ:
> A simple improvement would be to allow users to copy the error message shown in the error dialog to the clipboard, for bug reporting purposes.

Is that simple?
I assumed that allowing copy-paste from our dialogs is problematic, and required a lot of work...
Well, if it's feasible, maybe it's better to always allow copying, from all dialog boxes?

criezy:
> Is that simple?
> I assumed that allowing copy-paste from our dialogs is problematic, and required a lot of work...
> Well, if it's feasible, maybe it's better to always allow copying, from all dialog boxes?

It is actually quite simple as copying text to the clipboard is one of the features provided by the backends (see `OSystem::kFeatureClipboardSupport` and `OSystem::setTextInClipboard()`). The unknown game variant dialog has for example a button to allow copying the report to the clipboard. We could had a similar button to the `MessageDialog` has an option.

ZvikaZ:
Cool.

I'm using ScummVM (as a user) for a long time, and wasn't aware that it's possible to copy from (some of) ScummVM dialogs.

Is it possible to allow selecting text (with the mouse, or shift-arrow-keys) and copy the selected text?
It can be useful, for example, in the `about` dialog, to copy the version string to bug reports (most users don't know how to run `scummvm -v` from shell)
Or actually to allow copying any message the user gets, and wants to Google, or paste in bug report, even if we didn't expect him wanting to copy it (my previous `version` example is good for here too...)

criezy:
> Is it possible to allow selecting text (with the mouse, or shift-arrow-keys) and copy the selected text?

Selecting text is not really possible currently. In some places (such as the `EditableWidget`) we support using Ctrl+C and Ctrl+V to copy and past text but the user has no control over which text is copied (in the `EditableWidget` for example it copies the full text from the widget).

But that question actually raises an issue related to what has been discussed in this PR. In the unknown game dialog, using the _Copy to Clipboard_ button actually copies the original english text of the report and not the translated text visible in the dialog, since we want our bug reports to be in English. If we want to do the same here we get back to the issue of having access to the original English text from the dialog. One possibility would be to delegate that responsibility to the code using the dialog. It would have to pass both the text to display and the text to copy to the clipboard (which could actually be different to the text to display, e.g. a subset, even when not using translation).

ZvikaZ:
Well, this PR gets out of my intended scope...
I think it'd be be to close it.

Thanks for all your comments and ideas!

PR #2178, opened 2020-04-04
ANDROID: Multiple fixes for the new Gradle based build

lephilousophe:
With the merge of the new Gradle build system, builds running on Android 10 don't have access to SD card anymore.
When playing around, I noticed that gradle.build in android_project build directory is not updated when it is changed in source tree and various variables in configure are overriden.
These commits fix all of this and should not break existing setups (ie. buildbot should still be happy).

### DIFF
--- a/configure
+++ b/configure
@@ -1966,31 +1966,45 @@ esac
 
 # Toolchain for Android is in NDK and is using different naming convention
 if test "$_host_os" = android; then
-	case $_host_cpu in
-	arm)
-		_android_target="armv7a-linux-androideabi16"
-		;;
-	aarch64)
-		# Platform version 21 is needed as earlier versions of platform do not support this architecture.
-		_android_target="aarch64-linux-android21"
-		;;
-	i686)
-		_android_target="i686-linux-android16"
-		;;
-	x86_64)
-		# Platform version 21 is needed as earlier versions of platform do not support this architecture.
-		_android_target="x86_64-linux-android21"
-		;;
-	esac
+	# Try to use a known to work (on linux) toolchain
+	_android_toolchain="$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64"
+	if test -n "$ANDROID_TOOLCHAIN"; then
+		_android_toolchain="$ANDROID_TOOLCHAIN"
+	fi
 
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -1966,31 +1966,45 @@ esac
...
-		;;
-	esac
+	# Try to use a known to work (on linux) toolchain
+	_android_toolchain="$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64"
+	if test -n "$ANDROID_TOOLCHAIN"; then
+		_android_toolchain="$ANDROID_TOOLCHAIN"
+	fi
+
+	# If CXX environment variable is not set, try to set known to work defaults
+	if test -z "$CXX"; then
```
peterkohaut, line 1976: i was intentionally trying to avoid this, as i didnt find any real-case scenario. i'm not strictly against, but if you know such case, please tell me :)

antoniou79:
The first commit looks good to me -- updating gradle.build in the builld folder "android_project" when it is modified in the source tree dist folder.

The second commit seems necessary for now (targeting API 28), although the point is to support Android 10 as soon as possible anyway so we could get away with API 29 as long as we address the issue before uploading to store (probably still as a beta as a first phase). 

I am a bit unsure about the third commit changes. It seems that, even if the user has set $ANDROID_NDK_ROOT, this would now be overridden if there was a $ANDROID_TOOLCHAIN set already. Wouldn't this be confusing?

configure
lephilousophe: It is really useful for the buildbot as we use ccache there to speed up rebuilds.
To make it work we define CXX="ccache <real compiler>"

lephilousophe:
Main trouble with ANDROID_NDK_ROOT is that it doesn't point at the toolchain and @peterkohaut hardcoded a path valid on Linux only.
Currently, anyone on MacOSX or Windows couldn't build project.
I took inspiration on Google [documentation](https://developer.android.com/ndk/guides/other_build_systems#autoconf) where they define TOOLCHAIN variable.
Defining HOST_TAG and building the path ourselves could make it too.

configure
peterkohaut: that is very good use case which i didnt think of, thanks:)

PR #2179, opened 2020-04-04
CREATE_PROJECT: Windows: Add ignore flag for byte packing

stevenhoefel:
Otherwise VS2019 throws:
C2338: Windows headers require the default packing option. Changing this can lead to memory corruption. This diagnostic can be disabled by building with WINDOWS_IGNORE_PACKING_MISMATCH defined.


SupSuper:
This is fixed by updating SDL2 to the latest version: http://libsdl.org/download-2.0.php

PR #2180, opened 2020-04-04
DIRECTOR: Adds debug option to automate tests.

rvanlaar:
The debugflag fewframesonly exits the program after running 10 frames.
This is used when running a large amount of Director programs in an
automated fashion. The example script below saves the output for further
inspection.

CAVEAT: A score is ran for 10 frames. The counter is reset if it is
switched before that.

An example script in fishshell:
```
  for movie in (cat list_of_files_to_check.txt)
      ./scummvm --debugflags=runfirstframes -p DIRECTORY/ --start-movie=$movie workshop > output/$movie.txt 2>&1
      if test $status -eq 0
        echo $movie >> output/passed.txt
      else
        echo $movie >> output/failed.txt
      end
    end
```

### DIFF
--- a/engines/director/score.cpp
+++ b/engines/director/score.cpp
@@ -104,6 +104,9 @@ Score::Score(DirectorEngine *vm) {
 	_loadedCast = nullptr;
 
 	_numChannelsDisplayed = 0;
+
+	_framesRan = 0; // used by kDebugFewFramesOnly
+
 }
 
 void Score::setArchive(Archive *archive) {
@@ -1555,6 +1558,11 @@ void Score::startLoop() {
 
 		if (_currentFrame < _frames.size())
 			_vm->processEvents();
+
+		if (debugChannelSet(-1, kDebugFewFramesOnly) && _framesRan > 9) {
+			warning("Score::startLoop(): exiting due to debug few frames only");
+			break;
+		}
 	}
 
 	_lingo->processEvent(kEventStopMovie);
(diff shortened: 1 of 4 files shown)


engines/director/score.cpp
```diff
@@ -1555,6 +1559,11 @@ void Score::startLoop() {
 
 		if (_currentFrame < _frames.size())
 			_vm->processEvents();
+
+		if (debugChannelSet(-1, kDebugFewFramesOnly) && _framesRan > 9) {
+			warning("exiting due to debug few frames only");
```
sev-, line 1564: Please add `Score::startLoop()` prefix

engines/director/score.cpp
```diff
@@ -1539,6 +1539,10 @@ void Score::startLoop() {
 	_stopPlay = false;
 	_nextFrameTime = 0;
 
+	if (debugChannelSet(-1, kDebugFewFramesOnly)){
```
sev-, line 1542: Code formatting is wrong. Missing space before the curly bracket. Or better drop the brackets.

sev-:
Are you aware of `directortest-all` target? It loops over all files in directory, thus, your shell script is not needed.

rvanlaar:
I wasn't aware of `directortest-all`. It has one big caveat. It won't allow me to get the output on a per program basis and will stop the run when a segfault is triggered.

sev-:
Great, thank you!

PR #2185, opened 2020-04-09
GUI: Sort games list ignoring articles

sev-:
This PR introduces dictionary sort to the list of games.

Thus, 'The Dig' is inserted between 'Deja Vu' and 'Discworld'.

Quick search by typing also works properly.

### DIFF
--- a/common/str.cpp
+++ b/common/str.cpp
@@ -1299,6 +1299,8 @@ const char *scumm_skipArticle(const char *s1) {
 		o1 = 4;
 	else if (!scumm_strnicmp(s1, "a ", 2))
 		o1 = 2;
+	else if (!scumm_strnicmp(s1, "an ", 3))
+		o1 = 3;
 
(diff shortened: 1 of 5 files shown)


common/str.cpp
```diff
@@ -1284,6 +1293,20 @@ int scumm_strnicmp(const char *s1, const char *s2, uint n) {
 	return l1 - l2;
 }
 
+const char *scumm_skipArticle(const char *s1) {
+	int o1 = 0;
+	if (!scumm_strnicmp(s1, "the ", 4))
+		o1 = 4;
+	else if (!scumm_strnicmp(s1, "a ", 2))
+		o1 = 2;
+
```
criezy, line 1302: You probably want to skip "an " as well.

criezy:
The idea sounds good. It's a thumb up from me.

sev-:
OK, merging it.

PR #2188, opened 2020-04-12
SDL: Add --logfile command line argument to specify logfile path

criezy:
This adds a `-l PATH` or `--logfile=PATH` command line argument to specify a custom log file. Since log files are overwritten with each launch, this could help get a more persistent one when we want to (this remove the need to copy the default one after closing ScummVM, and this also removes the need to know where the default one is ;-) ).

This may also be useful to get a more portable application as mentioned in [bug #11412](https://bugs.scummvm.org/ticket/11412).

This only works with the SDL-based  backends.

Note that for this to work I had to delay a bit the creation of the log file. It is now done in the initBackend() function instead of `OSystem_SDL::init()`. This means for example that using a command line command such as `--detect` no longer overwrites the log file (since ScummVM returns before it is created), and that any logged message during the command line processing phase will only appear in stdout/stderr and will no longer be logged to the log file. In my opinion this is not necessarily a bad thing, but maybe others have a different opinion?

### DIFF
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -150,16 +150,6 @@ void OSystem_SDL::init() {
 	// Disable OS cursor
 	SDL_ShowCursor(SDL_DISABLE);
 
-	if (!_logger)
-		_logger = new Backends::Log::Log(this);
-
-	if (_logger) {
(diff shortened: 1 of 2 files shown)


sev-:
Nice change.

criezy:
Thank you for the review. Given the positive feedback I think I can merge this now.

PR #2189, opened 2020-04-15
DIRECTOR: LINGO: Implement moveableSprite of sprite

npjg:
This pull request implements the `moveableSprite of sprite`
property. We must thus track sprite positions based upon their 
current point and not just their start point.

### DIFF
--- a/engines/director/events.cpp
+++ b/engines/director/events.cpp
@@ -80,6 +80,19 @@ void DirectorEngine::processEvents() {
 			case Common::EVENT_MOUSEMOVE:
 				sc->_lastEventTime = g_director->getMacTicks();
 				sc->_lastRollTime =	 sc->_lastEventTime;
+
+				if (_draggingSprite) {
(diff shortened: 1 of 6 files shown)


engines/director/events.cpp
```diff
@@ -80,6 +80,19 @@ void DirectorEngine::processEvents() {
 			case Common::EVENT_MOUSEMOVE:
 				sc->_lastEventTime = g_director->getMacTicks();
 				sc->_lastRollTime =	 sc->_lastEventTime;
+
+				if (_draggingSprite) {
+					Sprite* draggedSprite = currentFrame->_sprites[_draggingSpriteId];
```
sev-, line 85: Code formatting is incorrect. It should be `Sprite *draggedSprite`

sev-:
Thanks, there is only one small nitpick with formatting, I'll fix it in-tree. Merging

PR #2190, opened 2020-04-15
MACOS: Changing default macOS savegame location

Thunderforge:
Addresses [#11427](https://bugs.scummvm.org/ticket/11427) by preventing future installations from accessing Documents (which the user can deny access to), but doesn't warn users if their existing installation still points to this folder and access isn't granted.

My understanding is that existing users will still be pointing to `~/Documents/ScummVM Savegames`, and so ScummVM will prompt them for access to the Documents folder.

### DIFF
--- a/backends/saves/posix/posix-saves.cpp
+++ b/backends/saves/posix/posix-saves.cpp
@@ -48,7 +48,7 @@ POSIXSaveFileManager::POSIXSaveFileManager() {
 	const char *home = getenv("HOME");
 	if (home && *home && strlen(home) < MAXPATHLEN) {
 		savePath = home;
-		savePath += "/Documents/ScummVM Savegames";
+		savePath += "/Library/Application Support/ScummVM/Savegames";
 
 		ConfMan.registerDefault("savepath", savePath);
 	}


criezy:
I think this change will also impact existing users who have not set a custom save path in the ScummVM options, and those (which I suspect to be the majority of users) will suddenly and silently lose all their savegames in ScummVM.

This needs to be checked, but if that is the case then I don't think we can accept this change without having a migration mechanism in place for the existing savegames.

criezy:
Also as a note, if we change the default save path on macOS (which I think is a good idea if done properly) we probably also want to remove the `NSDocumentsFolderUsageDescription` Info.plist.

And I have some reservations about using something under ~/Library/ as this folder is hidden to the user and this hinders his ability to share savegames. The cloud synchronisation of savegames mitigate that though. And it is definitely better than the crashes described in the bug report (although we might want to fix those crashes in the engines anyway as there may be other reasons why a save fails).

sev-:
I think, that we will have way too many problems with changing the default save path. The only possible approach would be to copy over all contents of the current save directory, which will anyway require access to the `Documents/` directory. We can detect the previous state by checking 'version' in the `ScummVM Preferences` aka `scummvm.ini` file.

We already changed default save directory on Android, and that had disastrous as users are just whining and shouting "my saves are gone".

I think the alternative solution would be to extend the `SaveFileManager` class and make sure that the `Documents/ScummVM` directory is writeable or otherwise nicely show banner to the user with an explanation of how to enable the write permission (via Control Panel).

Thunderforge:
Seeing that things are more complicated than I've initially assumed, I'm going to close this PR since changing the default seems premature. 

I've split the bug into a general "bad stuff happens if you can't access the savegames location, find a way to let the user know" ([#11427](https://bugs.scummvm.org/ticket/11427)) and "change the default macOS location so that users don't accidentally deny access to Documents" ([#11428](https://bugs.scummvm.org/ticket/11428)). I still hope that changing the default is a possibility, but we'll need a better way of handling migrations than with Windows and Android.

PR #2194, opened 2020-04-17
MACOSX: Remove NSRequiresAquaSystemAppearance from Info.plist

dwatteau:
Bug #11430.

Enabling Dark Mode compatibility on macOS, either by using `NSRequiresAquaSystemAppearance` on older SDKs, or by building with 10.14+ SDKs, makes the display blurry on Retina screens on Mojave (and probably Catalina).

This happens in OpenGL or SDL2 mode, filtered or not, with any rendering mode, etc.

This is possibly a macOS bug in OpenGL:

  https://bugzilla.libsdl.org/show_bug.cgi?id=5087
  https://communities.vmware.com/thread/618831
  https://code.videolan.org/videolan/VLCKit/issues/82#note_30971
  https://openradar.appspot.com/45895864 

This has been reported to Apple, but I don't think it's going to be fixed soon. SDL2 developers are hesitant about adding even more NSOpenGLView workarounds.

The only workaround that looks OK today is:

* to continue building official ScumMVM on older macOS releases (possibly any pre-10.14 SDK)
* and to stop providing any `NSRequiresAquaSystemAppearance` content in Info.plist.

So, revert commit 485e8bee17230ed8e17a27eaf148a3330f305a6f for now.

Missing Dark mode for now is probably less important that having incorrect output. It looks like VLC still doesn't enable Dark mode for the same reason.

### DIFF
--- a/dists/macosx/Info.plist
+++ b/dists/macosx/Info.plist
@@ -72,8 +72,6 @@
 	<string>faYfM+DGPgJCrRzpxSHoFD0rzHa6ZnnEXuzz2E7ZRUk=</string>
 	<key>NSDockTilePlugIn</key>
 	<string>scummvm.docktileplugin</string>
-	<key>NSRequiresAquaSystemAppearance</key>
-	<false/>
 	<key>NSDocumentsFolderUsageDescription</key>
 	<string>ScummVM saves and loads savegames in the Documents folder by default.</string>
 	<key>SUPublicDSAKeyFile</key>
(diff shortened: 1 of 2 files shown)


criezy:
Thank you.
I don't really like the idea of removing the Dark Mode support, but unfortunately I don't have any better idea, and I agree that the blurry display is more problematic than not supporting the dark mode. So I will merge this now.

PR #2195, opened 2020-04-17
WINTERMUTE: Added new reversion3 release (Steam)

aviloria:
Added new "Reversion 3 - The Return" release (Steam), including Italian language
Fixed WME version of previous release


sev-:
Thanks!

PR #2196, opened 2020-04-17
DIRECTOR:LINGO: implements run only one lingo test

rvanlaar:
Lingo tests can be run with:
`./scummvm -d11 -p engines/director/lingo/tests/ directortest`
This runs all the .lingo files in that directory to run in sequence.
When one script fails, the others won't run.

With this commit it's now possible to use the commandline option `--start-movie=` to run a specific lingo script.
`start-movie` already made it possible to run a specific director
movie.
The following command will only run the `c2.lingo` file.
`./scummvm -d11 -p engines/director/lingo/tests/ --start-movie=c2.lingo directortest`

### DIFF
--- a/engines/director/lingo/lingo.cpp
+++ b/engines/director/lingo/lingo.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "common/file.h"
+#include "common/config-manager.h"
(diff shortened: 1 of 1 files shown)


sev-:
Nice

PR #2197, opened 2020-04-17
Add kFeatureNoQuit in System to remove the Quit buttons from the GUI

criezy:
I am creating this Pull Request mostly for a review on the first commit (commit 27a4ad0) since it touches common code.

The main idea is that for various reasons on some platform we do not want to allow quitting the application programatically, for example through the use of a Quit button. One such platform is iOS. The first commit in this pull request does the following:
- It adds a kFeatureNoQuit in OSystem.
- If this feature is supported by a backend
  - it removes the Quit button from the launcher
  - it removes the Quit button from the GMM
  - it remaps EVENT_QUIT to EVENT_RTL in the event manager.

The rest of the commits are specific to the iOS backend to use this feature and further remove all use of exit() as well as cleanup the error reporting.

### DIFF
--- a/backends/events/default/default-events.cpp
+++ b/backends/events/default/default-events.cpp
@@ -95,6 +95,10 @@ bool DefaultEventManager::pollEvent(Common::Event &event) {
 	event = _eventQueue.pop();
 	bool forwardEvent = true;
 
+	// If the backend has the kFeatureNoQuit, replace Quit event with RTL
+	if (event.type == Common::EVENT_QUIT && g_system->hasFeature(OSystem::kFeatureNoQuit))
+		event.type = Common::EVENT_RTL;
+
 	switch (event.type) {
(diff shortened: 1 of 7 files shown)


henke37:
How does this interact with games that terminate the process on completion?

criezy:
They return to the launcher (as the quit event is remapped to a RTL event). Or at least that is the idea, but I was too lazy to play a game to the end so I only tested that with a hack.

sluicebox:
This is great! I hoping this would happen, I tossed in the iOS7_main exit() a while back since that was better than bricking when a game exited, but this is the right behavior for an ios app. I had imagined this would be more involved, I really like the remapping approach. You made it look easy!

criezy:
Thank you for the feedback. Since there was no objection I will merge this now and send an email to -devel in case other ports could benefit from the feature.

PR #2198, opened 2020-04-18
ULTIMA4: Enable keymapper actions to optionally trigger on kbd repeats

dreammaster:
This is a minor change to the keymapper to optionally allow actions to specify that they can be triggered by keyboard repeats, to allow the arrow keys in Ultima IV to repeatedly trigger movement whilst they're held down. Unless the method is called by engines for any actions when they set up their keymapper, the existing behaviour still happens, and keyboard reports won't trigger actions. Since it touches the core keymapper, I thought I'd get opinion on it from anyone familiar with the keymapper rather than simply committing it in master.

### DIFF
--- a/backends/keymapper/action.h
+++ b/backends/keymapper/action.h
@@ -115,6 +115,14 @@ public:
 	}
 
 	/**
+	 * Allows an action bound to a keyboard event to be repeatedly
+	 * triggered by key repeats
+	 */
+	void allowKbdReapets() {
+		event.kbdRepeat = true;
+	}
+
+	/**
 	 * Add a default input mapping for the action
 	 *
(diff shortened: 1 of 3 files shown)


bgK:
Thanks! I've pushed to this branch an additional commit to complete the keyboard repeat event handling in the keymapper: https://github.com/bgK/scummvm/commit/8e0b749b5ab7a4dc490204e3b90793ce7766986b.

To be fair, I have some reservations regarding relying on keyboard repeat events for character movement, as those do not trigger when the action is bound to a game controller.

dreammaster:
Thanks for the assistance cleaning it up. Looks good, I'll merge both commits in shortly.

PR #2199, opened 2020-04-18
JANITORIAL: fix the name of original author of cge and cge2

t3d:
The first name is Janusz, not Janus.
The correct name was used in AUTHORS and credits.
In case of doubts, see his personal webpage: https://www.jbw.pl/ - name is in the page footer


sev-:
Thanks

PR #2200, opened 2020-04-18
ULTIMA8: Fix compiler warnings

jepael:
These commits fix compiler warnings that come from headers.
They are all about shadowing a member variable.
Since these headers are included in many other files, the warnings are reduced by about 600 lines.


mduggan:
Thanks! Pretty straight forward so I went ahead and merged.

PR #2201, opened 2020-04-19
NUVIE: Fix compiler warnings

jepael:
These commits remove compiler warnings about variable shadowing that come from headers.
As the headers are included in many other files, the warnings are reduced by about 200 lines.


digitall:
These patches look pretty simple and self explanatory. Merging as-is

PR #2202, opened 2020-04-19
DISTS: install the AppStream file to the canonical location

pinotree:
The canonical location for AppStream XML files has been changed to `/usr/share/metainfo` four years ago at least, with `/usr/share/appdata` left as legacy location. It is time to switch to the right location.

### DIFF
--- a/ports.mk
+++ b/ports.mk
@@ -20,8 +20,8 @@ install:
 	$(INSTALL) -c -m 644 $(DIST_FILES_THEMES) $(DIST_FILES_NETWORKING) $(DIST_FILES_VKEYBD) $(DIST_FILES_ENGINEDATA) "$(DESTDIR)$(datadir)/"
 	$(INSTALL) -d "$(DESTDIR)$(datarootdir)/applications"
 	$(INSTALL) -c -m 644 "$(srcdir)/dists/scummvm.desktop" "$(DESTDIR)$(datarootdir)/applications/scummvm.desktop"
-	$(INSTALL) -d "$(DESTDIR)$(datarootdir)/appdata"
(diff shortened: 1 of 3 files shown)


ports.mk
```diff
@@ -21,7 +21,7 @@ install:
 	$(INSTALL) -d "$(DESTDIR)$(datarootdir)/applications"
 	$(INSTALL) -c -m 644 "$(srcdir)/dists/scummvm.desktop" "$(DESTDIR)$(datarootdir)/applications/scummvm.desktop"
 	$(INSTALL) -d "$(DESTDIR)$(datarootdir)/appdata"
```
digitall, line 23: Should this be amended to /metainfo as well?

ports.mk
```diff
@@ -43,7 +43,7 @@ install-strip:
...
```
digitall, line 45: Should this be amended to /metainfo as well?

digitall:
This is as per https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html I assume?

Looks fairly reasonable. @pinotree : Couple of minor extra changes might be required, See my line notes

ports.mk
pinotree: Ups sorry, brown paper bag, fixing in a moment. Thanks for the notice!

ports.mk
pinotree: Ups sorry, brown paper bag, fixing in a moment. Thanks for the notice!

pinotree:
> This is as per https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html I assume?

That's correct.

Thanks for the quick review! MR updated.

lotharsm:
Looks good to me. +1

digitall:
Right then... merging.

PR #2203, opened 2020-04-19
DIRECTOR: LINGO: Improve `chars` implementation

rvanlaar:
Improvements to chars implementation:
- handle pass by reference next to pass in of string
     Makes the D4 dictionary movie `chars` not exit with a segfault.
- handle when from_id is larger than to_id and return an empty string.

The lingo function chars(string, from_id, to_id) returns the
from_id'th char till and including the to_id'th char of the string, 1 indexed.


sev-:
Nice.

PR #2207, opened 2020-04-26
CREATE_PROJECT: Add c++11 warnings to xcode project

mduggan:
I broke some builders with a trailing comma in an enum, but xcode was totally ok with it and didn't even issue a warning.

I suggest we add `-Wc++11-extensions` to the build in xcode so others will be warned about the same.  Details on the warnings this adds [are here](https://clang.llvm.org/docs/DiagnosticsReference.html#wc-11-extensions).  Adding this also warns about some uses of `long long`, which we should be mostly avoiding anyway.

I don't have very strong feelings on this, so happy to hear opinions on whether this is a good idea.


criezy:
This seems fine to me. The only potential issue I can see is that it will break compiling with old Xcode version, although I don't know how old. I have tried with Xcode 6.2, and it does support the option. So you would have to go really far back in the past to find a version that does not.

criezy:
Thank you. I will merging this now.

PR #2208, opened 2020-04-26
GUI: Added (partial) BiDi support

ZvikaZ:
Added GNU FriBidi, thus allowing Hebrew (or other future RTL languages)
to be displayed correctly.
It's been implemented for all ScummVM GUI (as far as I have noticed),
and can be further used by the engines as needed.

This work is only partial, because for complete BiDi support we'll need
to mirror the widgets, and support input text areas (which currently
don't even support Hebrew text input at all).

Some changes are required in order to to use this:
- Visual Studio:
  -- add FriBidi lib from https://github.com/ShiftMediaProject/fribidi
     (and place the files in the other libs location)
  -- add fribidi.dll to the current directory
  -- add fribidi.lib to VS library list
     (in GUI it's: Project -> scummvm Properties -> Linker ->
                   input -> Additional Dependencies)
  -- Add USE_FRIBIDI define
         either to:
             dists/msvc/ScummVM_Global.props   <PreprocessorDefinitions>
         or in GUI:
             Project -> scummvm Properties ->
             C/C++ -> Preprocessor -> Preprocessor Definitions
- GCC:
  -- install FriBidi
     e.g.,
        Ubuntu: `apt-get install libfribidi-dev`
        CentOS: `yum install fribidi-devel`
  -- rerun `configure`

### DIFF
--- a/common/translation.cpp
+++ b/common/translation.cpp
@@ -36,6 +36,10 @@
 
 #ifdef USE_TRANSLATION
 
+#ifdef USE_FRIBIDI
+#include <fribidi/fribidi.h>
+#endif
+
 namespace Common {
 
 DECLARE_SINGLETON(TranslationManager);
@@ -449,6 +453,68 @@ bool TranslationManager::checkHeader(File &in) {
 
 	return true;
 }
+
+String TranslationManager::convertBiDiString(const String &input) {
+	if (getCurrentLanguage() != "he")		//TODO: modify when we'll support other RTL languages, such as Arabic and Farsi
+		return input;
+
+	if (getCurrentCharset() != "iso-8859-8") {
+		warning("convertBiDiString: Unexpected charset is used with %s language: %s", getCurrentLanguage().c_str(), getCurrentCharset().c_str());
+		return input;
+	};
+
+	return TranslationManager::convertBiDiString(input, HE_ISR);
+}
+
+#ifdef USE_FRIBIDI
+String TranslationManager::convertBiDiString(const String &input, const Common::Language lang) {
+	if (lang != HE_ISR)		//TODO: modify when we'll support other RTL languages, such as Arabic and Farsi
+		return input;
+
+	int buff_length = (input.size() + 2) * 2;		// it's more than enough, but it's better to be on the safe side
+	FriBidiChar *input_unicode = (FriBidiChar *)malloc(buff_length * sizeof(FriBidiChar));
+	FriBidiChar *visual_str = (FriBidiChar *)malloc(buff_length * sizeof(FriBidiChar));
+	char *output = (char *)malloc(buff_length);
+
+	FriBidiCharType pbase_dir = FRIBIDI_TYPE_ON;
+	FriBidiCharSet char_set = FRIBIDI_CHAR_SET_ISO8859_8;
+
+	FriBidiStrIndex length = fribidi_charset_to_unicode(char_set, input.c_str(), input.size(), input_unicode);
+
+	if (!fribidi_log2vis(
+		/* input */
+		input_unicode,
+		length,
+		&pbase_dir,
+		/* output */
+		visual_str,
+		NULL,			// position_L_to_V_list,
+		NULL,			// position_V_to_L_list,
+		NULL			// embedding_level_list
+	)) {
+		warning("convertBiDiString: calling fribidi_log2vis failed");
+		free(input_unicode);
+		free(visual_str);
+		free(output);
+		return input;
+	}
+
+	fribidi_unicode_to_charset(char_set, visual_str, length, output);
+
+	String result = String(output);
+	free(input_unicode);
+	free(visual_str);
+	free(output);
+
+	return result;
+}
+#else
+String TranslationManager::convertBiDiString(const String &input, const Common::Language lang) {
+	return input;
+}
+#endif
+
+
 
 } // End of namespace Common
 
--- a/gui/ThemeEngine.cpp
+++ b/gui/ThemeEngine.cpp
@@ -919,7 +919,11 @@ void ThemeEngine::drawDDText(TextData type, TextColor color, const Common::Rect 
 		restoreBackground(dirty);
 
 	_vectorRenderer->setFgColor(_textColors[color]->r, _textColors[color]->g, _textColors[color]->b);
+#ifdef USE_TRANSLATION
+	_vectorRenderer->drawString(_texts[type]->_fontPtr, TransMan.convertBiDiString(text), area, alignH, alignV, deltax, ellipsis, dirty);
+#else
 	_vectorRenderer->drawString(_texts[type]->_fontPtr, text, area, alignH, alignV, deltax, ellipsis, dirty);
+#endif
 
 	addDirtyRect(dirty);
 }
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -39,6 +39,7 @@ parts:
       - libfreetype6-dev
       - zlib1g-dev
       - libunity-dev
+      - libfribidi-dev
     stage-packages:
       - libicu55
       - libasound2
--- a/common/translation.h
+++ b/common/translation.h
@@ -28,6 +28,7 @@
 #include "common/str.h"
 #include "common/singleton.h"
 #include "common/str-array.h"
+#include "common/language.h"
 
 #ifdef USE_TRANSLATION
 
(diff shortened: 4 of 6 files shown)


gui/ThemeEngine.cpp
```diff
@@ -919,7 +919,7 @@ void ThemeEngine::drawDDText(TextData type, TextColor color, const Common::Rect
 		restoreBackground(dirty);
 
 	_vectorRenderer->setFgColor(_textColors[color]->r, _textColors[color]->g, _textColors[color]->b);
-	_vectorRenderer->drawString(_texts[type]->_fontPtr, text, area, alignH, alignV, deltax, ellipsis, dirty);
+	_vectorRenderer->drawString(_texts[type]->_fontPtr, TransMan.bidiAlgo(text), area, alignH, alignV, deltax, ellipsis, dirty);
```
criezy, line 922: This will break the compilation when translations is disabled (i.e. if `USE_TRANSLATION` is not defined).

ZvikaZ: Done.
I've also updated the initial PR description a little to much the updated compilation instructions.

snapcraft.yaml
```diff
@@ -39,6 +39,7 @@ parts:
       - libfreetype6-dev
       - zlib1g-dev
       - libunity-dev
+	  - libfribidi-dev
```
sev-, line 42: This has extra spaces.

common/translation.cpp
```diff
@@ -450,6 +454,55 @@ bool TranslationManager::checkHeader(File &in) {
...
+	String result = String(output);
+	free(input_unicode);
+	free(visual_str);
+	free(output);
+
+	return result;
+}
+#else
+String TranslationManager::bidiAlgo(const String input) {
+	return input;
```
sev-, line 500: Does it compile or it says that it loses `const` qualifier?

common/translation.cpp
```diff
@@ -450,6 +454,55 @@ bool TranslationManager::checkHeader(File &in) {
 	return true;
 }
 
+#ifdef USE_FRIBIDI
+String TranslationManager::bidiAlgo(const String input) {
```
sev-, line 458: I am not a big fan of this name. Perhaps `convertBiDiString` could be better?

common/translation.cpp
```diff
@@ -450,6 +454,55 @@ bool TranslationManager::checkHeader(File &in) {
...
+	if (getCurrentLanguage() != "he")		//TODO: modify when we'll support other RTL languages, such as Arabic and Farsi
+		return input;
+
+	if (getCurrentCharset() != "iso-8859-8") {
+		warning("Unexpected charset is used with %s language: %s", getCurrentLanguage().c_str(), getCurrentCharset().c_str());
```
sev-, line 463: It is better to put name of the method as prefix to the warning, so it would be easier to find it.

common/translation.cpp
ZvikaZ: It compiled, both in Windows 10 VS, and Ubuntu.

ZvikaZ:
I will fix your notes tomorrow.

Meanwhile, I didn't understand your first question.
What's (non) AA fonts?

criezy:
@ZvikaZ It looks better now. Thank you. I will look at this in details in a few days.

@sev- I think it might find the font in `dists/engine-data/fonts.dat`.

digitall:
@ZvikaZ : AA in this context is Anti-Aliasing i.e. https://en.wikipedia.org/wiki/Font_rasterization and https://en.wikipedia.org/wiki/Spatial_anti-aliasing

ZvikaZ:
OK, but I still don't understand the potential risk with non AA fonts.

And how can I test that?
I don't remember seeing a setting for AA or non AA fonts...

common/translation.cpp
SupSuper: You should be using `const String &` instead of `const String` in function parameters. This avoids needlessly copying immutable strings.

ZvikaZ: @SupSuper , done

snapcraft.yaml
ZvikaZ: Done

common/translation.cpp
ZvikaZ: Done

common/translation.cpp
ZvikaZ: Done

ZvikaZ:
Implemented all the suggestions.
Thanks.

Also, I've changed the API a little bit - in the previous version, it assumed that the text is in the GUI language, but game's language can be different than the GUI's.

Regarding the non AA fonts - I tried running with `GUI renderer` equals `Antialiased renderer` and `Standard rendered`, in Hebrew and in English, and didn't see any change.
Anyway, as I said before, I don't see why it should make any difference.

sev-:
The difference for the non-AA renderer is fonts. Where does the Hebrew font come from?

ZvikaZ:
Honestly, I don't know.
But it's out if the scope of this PR.
Hebrew was displayed before, it was just (almost) mirrored.
This PR adds code manipulates the string to display - it takes an input string, and changes the order of the characters inside - returning another string.

ZvikaZ:
I have re-read my initial comment here, and maybe I wasn't clear enough.
When I spoke about 'display correctly' I didn't mean in the sense  of render correctly, but in the sense of displayed in the correct order.

criezy:
Also I confirmed that it is getting the font from our `font.dat` file. If I rename that file I can no longer use the Hebrew translation (neither with the AA renderer, nor with the non-AA renderer). But if the `font.dat` file is here, then it works (both with the AA renderer and with the non-AA renderer).

ZvikaZ:
> I have one question: if we build without fribidi, do we want to still show Hebrew in the language list in the options? I am guessing it is probably hard to read when all the text is backward. So maybe it should be skipped in the `TranslationManager::getSupportedLanguageNames()` implementation in such a case?

That's a good question...
I took some time to think about it.
Personally, I'm annoyed by the current backward Hebrew.
However, maybe there are users that speak only Hebrew, or feel much more comfortable in Hebrew than English, that they would prefer backward Hebrew to anything else.
Therefore, I think that it's OK to leave Hebrew in even if bidi is disabled, and let the user decide whether he wants that or not.

sev-:
Great, thank you, merging.

BLooperZ:
Regarding that last question, I suggest allowing users to choose they prefered language for ScummVM GUI.

ZvikaZ:
> Regarding that last question, I suggest allowing users to choose they prefered language for ScummVM GUI.

Of course the user can choose.
The question was whether we want to insert Hebrew as one of the options, in platforms that won't support this new BiDi PR.

i.e.,
In most platforms, there will BiDI support, and Hebrew will be displayed correctly, and can be chosen.
But there might be platforms that GNU's FriBidi library isn't available for them, and Hebrew will be displayed reversed. 
So - the question is, do we want to leave Hebrew as an option in that case.
The current decision is yes - leave Hebrew as an option, even if it's reversed, and let the user decide.

antoniou79:
Wouldn't it be possible to integrate the FriBiDi source code in ScummVM so as to avoid the third party library dependency (and the cross-compilation it will require per port)?

For example, there's currently work done for Hebrew subtitles for Blade Runner by rzil. Ideally, I'd like to make use of the BiDi support in ScummVM without the extra implication that some of our ports will not be displaying the subtitles correctly in Hebrew.

ZvikaZ:
> Wouldn't it be possible to integrate the FriBiDi source code in ScummVM so as to avoid the third party library dependency (and the cross-compilation it will require per port)?
> 
> For example, there's currently work done for Hebrew subtitles for Blade Runner by rzil. Ideally, I'd like to make use of the BiDi support in ScummVM without the extra implication that some of our ports will not be displaying the subtitles correctly in Hebrew.

Well, that's a question of policy, I think.
After all, why didn't we integrate other libs inside ScummVM?

@sev- when do we prefer to have a library as an external dependancy, and when do we integrate it as part of ScummVM?

And as for rzil (@BLooperZ ) work, we've been in touch in the recent days, and we're coordinated. As far as I understand, he's happy that we have now a good BiDi library available.

lephilousophe:
IMO it would be a bad idea to include Fribidi in the main tree as it's an external dependency receiving bugfixes on a regular basis. Including it would make it necessary to follow upstream.
In addition, as @ZvikaZ said, why don't we include other libraries too?
The only library I see which is included in the tree is the Lua one and it seems we patched it.

I already added fribidi to almost all platforms supported by [dockerized-bb](https://github.com/lephilousophe/dockerized-bb/). Builds are still compiling but it seems OK so far.
It was really easy to add the dependency.

antoniou79:
Hmm, I didn't think FriBiDi was receiving regular updates.

It seems to me that most of the other third party libraries that we don't include are either optional or affecting specific (small or just one) other engine(s). 

I am also unsure about their dependencies, meaning that in the case of FriBiDi they state that it does not depend on any other library, so it would be straight forward to integrate. I don't think this is the case for some of our other third party libs.

lotharsm:
Maybe we should move this out of this PR since this could lead to a more "general" discussion - scummvm-devel@ might be a better place.

Have we ever bundled third-party libs in the past despite `munt`? I'm not a big fan of this idea since I think this will introduce a pretty high workload due to having to keep track about updates.

PR #2211, opened 2020-04-29
SCUMM: Change NES Palette Used for Maniac Mansion to Sony CXA2025AS

digitall:
NES games have a 48 color and 6 gray palette, but this does not include
how the colors are mapped to the screen which will vary by system / TV
display. Each NES Emulator has used slightly different values.

These values are submitted as a better mapping for the flesh tones in
Maniac Mansion by the user "LaurentLaSalle" from a forum post.


digitall:
The relevant forum post was https://forums.scummvm.org/viewtopic.php?f=2&p=88732

sev-:
Thanks! Thanks a lot for taking care of it.

laurentlasalle:
I would advise *not* to change the NES color palette yet with what I suggested back on the SCUMM VM board as it needs further testing.

Some scenes will inappropriately render grey pixels as black, a bug caused by a modification I did on the first color of the tableNESPalette. Here is a comparison shot between the modified Sony CXA2025AS palette and the default palette with the occuring problem : 

Maniac Mansion NES - Garage color bug with modified Sony CXA2025AS palette (vs Default palette)

Yes, two shades of grey gets affected by this one change. The reason for this change was motivated by a rendering problem on the intro scene of the game. Once again, a comparison shot between the modified Sony CXA2025AS palette and the default palette : 

Maniac Mansion NES - Sky render with modified Sony CXA2025AS palette (vs Default palette)

The modification was done on the first color of the tableNESPalette, from 0x58, 0x58, 0x58 (of the original Sony CXA2025AS palette) to 0x00, 0x00, 0x00. If you keep it to 0x58, 0x58, 0x58, not only the sky render problem of the intro screen gets more visible, but the kids selection screen is much worse. Here is a comparison of that screen between the modified Sony CXA2025AS palette and the original Sony CXA2025AS palette : 

Maniac Mansion NES - Kids selection with modified Sony CXA2025AS palette (vs Sony CXA2025AS original palette)

So, from what I understand, Maniac Mansion on NES is one rare games that render that first black color inappropriately. Here is what I could find on that topic from the NES Dev Wiki : 

> On an NES, the palette color $0D causes the signal to drop below the normal black level. This low voltage signal is sometimes mistaken by televisions for blanking signals, which can cause an unstable picture, or total picture loss on some devices. Other devices seem to process with signal without problem.

Looks to me like the shades of grey occuring in the garage screen is caused by a substraction of the color information of these grey sprites with that first black color. A fix *could* be to revert the first color back to the default 0x24, 0x24, 0x24 used by SCUMM VM, while keeping the rest of the Sony CXA2025AS palette.

Mataniko:
I think we should strive for accuracy, and not use a custom palette, unless it's behind a setting or a custom palette entered by the user via config.

laurentlasalle:
Unfortunately, there is no such thing as an accurate NES color palette (NTSC TV models are configured differently, with different gammas, brightness, saturation, and hues adjustments). There seems to be no general consensus on what color palette to use in the NES emulation community. 

If SCUMM VM had a user option to choose different color palette ([that can be grabbed in the Download section of this page](http://emulation.gametechwiki.com/index.php/Famicom_Color_Palette) or elsewhere on the web), without having to recompile a custom build just for this, that would be great. But I know that implementing such an option for only one game is not that valuable.

Mataniko:
By accurate I mean "send the same colors to the screen if the NES would send a digital signal" The way they appear on the TV is due to the TV itself and should not be changed by ScummVM (or a NES emulator) unless the user manually changes the palette.

sev-:
And this is the problem: NES had no digital output, only PAL or NTSC analog signal.

laurentlasalle:
I tested the garage screen using Sony CXA2025AS original palette on both Bizhawk emulator (as a reference; it's accurate) and SCUMM VM. Without my modification, sprites that are supposed to be grey are black on SCUMM VM, and the background (that is supposed to be black) is grey. No wonder changing the first color to `0x00, 0x00, 0x00` made some sprites disappeared.

Maniac-Mansion-NES---Comparison-of-Bizhawk-vs-SCUMM-VM-using-Sony-CXA2025AS-original-palette

I just found out that SCUMM VM is based in large part on the (much more saturated) [RGB palette](http://emulation.gametechwiki.com/index.php/Famicom_Color_Palette) from the PlayChoice-10 arcade cabinet, thinking that this palette was an accurate representation of what NES developers worked with. It's not. AFAIK, only games developed for the PlayChoice-10 were encoded using this palette.

I say "based on large part" because there is an inconsistency with `$00` and `$1D` (seems like they have switched places) which might be the root of this whole black vs grey bug I keep mentioning. Here is the RGB palette [according to the Emulation General Wiki](http://emulation.gametechwiki.com/index.php/Famicom_Color_Palette) : 

RGB Palette

It would make sense, if indeed `$00` and `$1D` have been interchanged, that the SCUMM VM is programmed to look for those colors in the wrong place, and render the garage screen incorrectly when using the Sony CXA2025AS as shown above. Here is the default SCUMM VM palette (where `$00` and `$1D` are switched when compared to the RGB palette above) and the Sony CXA2025AS original palette : 

NES-Palette-SCUMM-vs-Sony

UNLESS... Unless having removed the annoted `/*    0x1D     */` and `/*    0x00     */` in the original code when I suggested changing the palette broke something. That wouldn't make much sense IMO.

Mataniko:
It looks identical to the Nestopia RGB palette, and it looks terrible because it's meant for an RGB screen.

I think as a reference we should look at either Mesen's NTSC palette or for a more natural look, FireBrandX's palettes: http://www.firebrandx.com/nespalette.html 
Those look much more natural and close to the original hardware. He even has a palette ripped from the NES Classic, which could be considered as the "official palette of a digital NES"

image

Which is way less saturated than the NTSC or CXA2025AS palettes

laurentlasalle:
Regardless of what palette we end up with, the mix up between position `$00` and `$1D` will have to be solved.

Mataniko:
The mix up in the two locations was done because the initial buffer for the screen is initialized with 0's so black needed to be the first color. As a hack, every time we draw, we'd reverse the two colors whenever the color is 0x00 or 0x1d.

The correct solution is to undo the hack, fix the palette, and for NES MM initialize the background with 0x1d which is what the NES does for the game. We also incorrectly set the color of the top text (it shouldn't be white)

I have things mostly fixed, with a decent way to toggle between NTSC and NES Classic palettes, but can't find where the surface is set initially with all 0's. Someone with some engine/graphics knowledge can pitch in

sev-:
Do you mean with hack this? https://github.com/scummvm/scummvm/blob/master/engines/scumm/gfx.cpp#L2594

The background reset is this: https://github.com/scummvm/scummvm/blob/master/engines/scumm/gfx.cpp#L426

Mataniko:
Yes, that needs to be undone, and instead the initial surface should be initialized to 0x1d instead of 0 as it never gets redrawn (the gray borders in the screenshot above)

laurentlasalle:
I switched both colors on my version, and it fixes everything. Of course, if the objective is to allow user to choose any .pal file for their color palette, reverting the hack would be wiser.

fix

Mataniko:
@laurentlasalle can you please look at #2218 and compare the garage scene now?

laurentlasalle:
I'm unable to build the latest source files using Visual Studio 2017. The only error seems to be related to fribidi.h?

- **Warning C4715** 'Mohawk::MohawkEngine_Myst::canDoAction': not all control paths return a value mohawk `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\engines\mohawk\myst.cpp` 705	
- **Warning C4389** '==': signed/unsigned mismatch prince `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\engines\prince\archive.cpp` 101	
- **Warning C4530** C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc (compiling source file ..\..\audio\softsynth\mt32\c_interface\c_interface.cpp) scummvm `d:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.16.27023\include\xlocale` 319	
- **Warning C4611** interaction between '_setjmp' and C++ object destruction is non-portable scummvm `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\common\lua\ldo.cpp` 143	
- **Error C1083** Cannot open include file: 'fribidi/fribidi.h': No such file or directory scummvm `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\common\translation.cpp` 40	
- **Warning C4018** '>=': signed/unsigned mismatch scummvm `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\graphics\macgui\mactext.cpp` 63	
- **Warning C4018** '>=': signed/unsigned mismatch scummvm `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\graphics\macgui\mactext.cpp` 498	
- **Warning C4018** '<': signed/unsigned mismatch scummvm `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\graphics\macgui\mactext.cpp` 910	
- **Warning C4018** '<': signed/unsigned mismatch scummvm `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\graphics\macgui\mactext.cpp` 968	
- **Warning C4701** potentially uninitialized local variable 'nsx' used scummvm `d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\graphics\macgui\mactext.cpp` 747	
- **Warning C4805** '!=': unsafe mix of type 'bool' and type 'int16' in operation scummvm	`d:\retro gaming\emulators\scummvm-master-2.2-nesfix_stable\gui\options.cpp` 1140

laurentlasalle:
I just applied the changes on the files I previously downloaded prior to the FriBidi implementation, and was able to build the project in Visual Studio 2017. My comments have been posted on #2218.

criezy:
> I'm unable to build the latest source files using Visual Studio 2017. The only error seems to be related to fribidi.h?

FriBiDi is a new dependency for ScummVM. You can either get it from https://github.com/ShiftMediaProject/fribidi/releases or use `--disable-fribidi` when running `create_project`.

laurentlasalle:
> FriBiDi is a new dependency for ScummVM. You can either get it from https://github.com/ShiftMediaProject/fribidi/releases or use `--disable-fribidi` when running `create_project`.

Thanks! Shouldn't it be included with the ScummVM Libs?

laurentlasalle:
How exactly do we toggle between NTSC and NES Classic palette @Mataniko? I can't find any options regarding this in the SCUMM VM UI.

Mataniko:
it is a checkbox in the Engine tab, in the Edit game dialog

digitall:
OT, but in TV broadcast, NTSC is nicknamed "Never The Same Color" :)

laurentlasalle:
@Mataniko I was expecting it to be in the Options, but I got it mixed up with Edit Game. #Facepalm

PR #2212, opened 2020-04-30
DEVTOOLS: add fribidi library support to create_project

ZvikaZ:
This PR follows @criezy 's suggestion in https://github.com/scummvm/scummvm/pull/2208 

Note:
I have checked it in VS. However, I don't have a Mac, so XCODE changes weren't verified.


criezy:
Thank you.
I have tested the Xcode changes and they work as expected.
I will merge this now.

PR #2214, opened 2020-05-01
IMAGE: Add support for Crusader: No Remorse  movie decoding

mduggan:
The movies for Crusader: No Remorse have a unique decoder which is not too hard
to implement.  Unfortunately, they don't properly implement the "compression"
FourCC, and instead put their ID in the "Stream Handler".  Since supporting
them requires a change to the existing Image API, I thought I should make a
pull request for comments.

With this change, the movies in Crusader can all be played nicely.

### DIFF
--- a/video/avi_decoder.cpp
+++ b/video/avi_decoder.cpp
@@ -201,6 +201,7 @@ bool AVIDecoder::parseNextChunk() {
 	case ID_STRH:
 		handleStreamHeader(size);
 		break;
+	case ID_HDRL: // Header list.. what's it doing here? Probably ok to ignore?
 	case ID_STRD: // Extra stream info, safe to ignore
 	case ID_VEDT: // Unknown, safe to ignore
 	case ID_JUNK: // Alignment bytes, should be ignored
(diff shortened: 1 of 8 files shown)


video/avi_decoder.cpp
```diff
@@ -201,6 +201,7 @@ bool AVIDecoder::parseNextChunk() {
 	case ID_STRH:
 		handleStreamHeader(size);
 		break;
+	case ID_HDRL: // Header list.. what's it doing here? Probably ok to ignore?
```
bluegr, line 204: Why is this a question?

bluegr:
Nice work overall!

I assume that Crusader: No Regret uses the same movie format?

video/avi_decoder.cpp
mduggan: Ah, because I'm honestly not sure of the answer - this pattern pops up in a few of the crusader videos (not all) and it didn't seem to affect anything to just skip.  If anyone else understands AVI containers better than me I'd be happy to get advice.

mduggan:
Thanks, unfortunately No Regret has its own weird format :(
There is some documentation available for that one here: https://wiki.multimedia.cx/index.php/Origin_Xan_Codec

sev-:
Thanks! merging.

PR #2215, opened 2020-05-01
SCI: add support for Hebrew translation of Torin

ZvikaZ:
- Detect fan made Hebrew translation of Torin's Passage
- Use BiDi algorithm in SCI engine

### DIFF
--- a/engines/sci/graphics/text32.cpp
+++ b/engines/sci/graphics/text32.cpp
@@ -22,6 +22,7 @@
 
 #include "common/util.h"
 #include "common/stack.h"
+#include "common/unicode-bidi.h"
 #include "graphics/primitives.h"
 
 #include "sci/sci.h"
@@ -340,10 +341,18 @@ void GfxText32::drawTextBox() {
 		uint length = getLongest(&nextCharIndex, textRectWidth);
 		int16 textWidth = getTextWidth(charIndex, length);
 
-		if (_alignment == kTextAlignCenter) {
-			_drawPosition.x += (textRectWidth - textWidth) / 2;
-		} else if (_alignment == kTextAlignRight) {
-			_drawPosition.x += textRectWidth - textWidth;
+		if (!g_sci->isLanguageRTL()) {
+			if (_alignment == kTextAlignCenter) {
+				_drawPosition.x += (textRectWidth - textWidth) / 2;
+			} else if (_alignment == kTextAlignRight) {
+				_drawPosition.x += textRectWidth - textWidth;
+			}
+		} else {
+			if (_alignment == kTextAlignCenter) {
+				_drawPosition.x += (textRectWidth - textWidth) / 2;
+			} else if (_alignment == kTextAlignLeft) {
+				_drawPosition.x += textRectWidth - textWidth;
+			}
 		}
 
 		drawText(charIndex, length);
@@ -364,7 +373,18 @@ void GfxText32::drawText(const uint index, uint length) {
 	// This draw loop implementation is somewhat different than the
 	// implementation in SSCI, but is accurate. Primarily the changes revolve
 	// around eliminating some extra temporaries and fixing the logic to match.
-	const char *text = _text.c_str() + index;
+
+	Common::String textString;
+	const char *text;
+	if (!g_sci->isLanguageRTL()) {
+		text = _text.c_str() + index;
+	} else {
+		const char *textOrig = _text.c_str() + index;
+		Common::String textLogical = Common::String(textOrig, (uint32)length);
+		textString = Common::convertBiDiString(textLogical, g_sci->getLanguage());
+		text = textString.c_str();
+	}
+
 	while (length-- > 0) {
 		char currentChar = *text++;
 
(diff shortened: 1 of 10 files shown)


engines/sci/graphics/text32.cpp
```diff
@@ -364,7 +377,22 @@ void GfxText32::drawText(const uint index, uint length) {
 	// This draw loop implementation is somewhat different than the
 	// implementation in SSCI, but is accurate. Primarily the changes revolve
 	// around eliminating some extra temporaries and fixing the logic to match.
-	const char *text = _text.c_str() + index;
+
+	Common::String textString;
+	const char *text;
+	if (g_sci->getLanguage() == Common::HE_ISR) {
+#ifdef USE_TRANSLATION
```
sev-, line 384: That is incorrect and will not work when translation is enabled but BiDi is not. You have to wrap it in both conditions.

ZvikaZ: But see in `common\translation.cpp`:

```
#ifdef USE_FRIBIDI
String TranslationManager::convertBiDiString(const String &input, const Common::Language lang) {

...

}
#else
String TranslationManager::convertBiDiString(const String &input, const Common::Language lang) {
	return input;
}
#endif
```

Therefore, in case that translation is enabled but BiDi is not, it should just return the original input.

If you're worried about the `(g_sci->getLanguage() == Common::HE_ISR)` - it's OK, because `SciEngine::getLanguage()` is always defined, and not dependent on any `ifdef`.

Anyway, I have just verified it without `USE_TRANSLATION`, and it worked well.

criezy: Maybe the `convertBiDiString()` that takes a language could be moved out of `TranslationManager` so that it is available also when `USE_TRANSLATION` is not defined?

I am also wondering if assuming `iso8859-8` is a good idea, and maybe we might want to pass the charset to the function as well (e.g. maybe we support some Hebrew games that use Unicode encoding and would benefit from this, or we might want to change TranslationManager to use Unicode in the future but without breaking BiDi in SCI).

ZvikaZ: @criezy , I agree.
Let's wait with this PR until #2217 is approved, and then I'll update the code here.

Kawa-oneechan: For a second there I misread that as "SCI games that use Unicode encoding" and thought someone had beaten me to it.

engines/sci/graphics/text32.cpp
```diff
@@ -340,10 +344,19 @@ void GfxText32::drawTextBox() {
 		uint length = getLongest(&nextCharIndex, textRectWidth);
 		int16 textWidth = getTextWidth(charIndex, length);
 
-		if (_alignment == kTextAlignCenter) {
-			_drawPosition.x += (textRectWidth - textWidth) / 2;
-		} else if (_alignment == kTextAlignRight) {
-			_drawPosition.x += textRectWidth - textWidth;
+		if (g_sci->getLanguage() != Common::HE_ISR) {  // i.e. regular Left To Right direction
```
bluegr, line 347: It would be better if this check was wrapped into a separate function, e.g. isRTLLanguage(), as we might support other RTL languages in the future

ZvikaZ: Agreed. I have even already implemented `isLanguageLTR` (I prefer positive checks rather than `if not something`.
I will update his after PR #2217 will be accepted, along with other required changes.

sev-: Now #2217 is accepted. Please, update this PR, so we could merge it.

ZvikaZ:
Continuing the work at PR #2236 , which moved `convertBiDiString(..page)` to unicode-bidi.h, now moving to there also the `(..lang)` flavour.

Thus, translation.h has only the SVM-GUI related function, and the two util functions page+code are 
 in unicode-bidi.

Also fixed a bug reported in Discord.
Copying here it as well:

Problem:

> In this code:
> 
>     char c[] = {0xee, 0xe5, 0xe3, 0x21, 0x20, 0xee, 0xe9, 0x20, 0xe6, 0xe4, 0x20, 0xf9, 0xed, 0x3f, 0x22, 0};
>     String s_in = c;
>     Common::U32String us = s_in.decode(Common::kWindows1255);
>     String s_out = us.encode(Common::kWindows1255);
> 
> I expect s_in and s_out to be equal.
> But s_out is just a bunch of question marks.
> Somehow, the .encode failed, but I don't understand why.
> These chars are perfect Windows 1255 encoding, as can be seen at https://en.wikipedia.org/wiki/Windows-1255

Solution:

> In the file backends\platform\sdl\win32\win32.cpp, line 441:
> string = Common::convertUtf32ToUtf8(UTF32Str).c_str();
> There is a problem.
> convertUtf32ToUtf8 returns a String, and .c_str() returns a pointer to its internal storage, and that's assigned to string.
> But this returned  String, along with its storage, is later deallocated, and now string (note the S and s in my explanations...) is pointing to garbage.
> 
> Fixed it by declaring:
>     Common::String tempString; outside of the scoped if, and changing that line to:
> 
>         tempString = Common::convertUtf32ToUtf8(UTF32Str);
>         string = tempString.c_str();

ZvikaZ:
Note: the builds aren't really failed.
2 passed, and the Mac one also went well, and finished without errors, but then it timed out after 50 minutes (what is it doing there so long, after finishing?...)

engines/sci/graphics/text32.cpp
ZvikaZ: @sev- , @bluegr , @criezy , a reminder to check my updates.
@sluicebox , you might want to review this.
Thanks.

sev-:
Now it is clean and tidy. Thank you!

PR #2216, opened 2020-05-02
AUDIO: Add support for loading Crusader AMF files

mduggan:
Creating a PR for comments - I want to hear opinions as I'm adding functionality to something which was previously a fairly clean import of the micromod/ibxm player code.

This plays the Asylum AMF files *pretty* well, but there's still a bug with some instruments I haven't worked out.  The drums in the menu music sound a bit wrong.

The ibxm player also doesn't support loop points, so it would be nice to add them.  That would be a different change though.


dreammaster:
Looks okay; as with the previous PR, it looks mostly self contained, but I'll leave it for hopefully someone more familiar with the audio code to also comment.

sev-:
This is a good change. Thanks, merging.

PR #2217, opened 2020-05-02
COMMON: Base BiDi support on U32String

BLooperZ:
Following #2208
Apply Fribidi convertion on U32String and use ScummVM's charset/unicode conversion rather than Fribidi's.
The BiDi convertion itself now works on U32String which is already encoded in UTF-32, so when using it, it should not require additional parameters (lang) and can be used unconditionally as-is (if conversion is not needed, it should return the original string).

This will simplify RTL support for engines using U32String, such as Director, Pink and Blade Runner. (Maybe some parts of the GUI?)


lephilousophe:
As @criezy proposed yesterday on #2215, shouldn't these functions moved to str.h and ustr.h?
That would make them independent of USE_TRANSLATION and as they are more string manipulation functions than translation things that seems more logical.

ZvikaZ:
I like these changes. And also support moving out of transman, to avoid all these ifdefs.

BLooperZ:
I think it is good idea to move it out of transman, but putting the function on different files can be confusing.
Perhaps should put it in str-enc.h or some new file (e.g. str-bidi.h)?

BLooperZ:
Moved to str-bidi.h.
Methods which already were in transman are left there for (probably short-term) backwards compatibility.

ZvikaZ:
> Moved to str-bidi.h.
> Methods which already were in transman are left there as stubs for (probably short-term) backwards compatibility.

I think it's better to move everything to the same place. Anyway, there is currently only one place that calls `convertBiDiString` (in `gui/ThemeEngine.cpp`), so it should be easy.
And ThemeEngine.cpp would benefit from the removal of `#ifdef USE_TRANSLATION`.

ZvikaZ:
> > Moved to str-bidi.h.
> > Methods which already were in transman are left there as stubs for (probably short-term) backwards compatibility.
> 
> I think it's better to move everything to the same place. Anyway, there is currently only one place that calls `convertBiDiString` (in `gui/ThemeEngine.cpp`), so it should be easy.
> And ThemeEngine.cpp would benefit from the removal of `#ifdef USE_TRANSLATION`.

On the other hand, maybe it doesn't matter now, and I will do it in my https://github.com/scummvm/scummvm/pull/2215/ , after this PR will be accepter.

sev-:
@criezy are you happy with having `str-bidi.h` or better still stuff it into `str.h` and `ustr.h`? I'm personally more inclined towards the latter.

criezy:
I don't think we need to have all the BiDi conversion functions defined in the same place. In my opinion it would make more sense to have `convertBiDiString` in `str.h` and `convertBiDiU32String` in `ustr.h`. And keeping the one for the GUI in `TranslationManager` is also fine I think as it has to check the current GUI language (in in theory encoding), which is handled by the `TranslationManager` instance.

bgK:
IMO the PR is fine as it is, with all the code in `str-bidi.h`. We should avoid adding more stuff to the `String` class and prefer utility fonctions so we can simply change them to take a `StringView` instead of a `String` when we have one (see https://en.cppreference.com/w/cpp/string/basic_string_view).

bluegr:
I also agree that the PR is fine as it is

BLooperZ:
As requested here and on #2236 I have moved the conversion to `str.h` and `ustr.h`.
But in retrospect I think it wasn't a good change and it was better to use the separate `str-bidi.h` for number of reason:
1. `str.h` is very widely used, so every change to this file require building almost everything again (long build when updating code base).
2. bidi support might introduce more changes which should have context (for example, supporting accelerator underline in macgui needed a way to figure out the visual position of nth character of the original (logical) string).
3. it is better to use this convertion functions in specific "key point" through the code... and working with the logical string as much as possible to avoid confusion. putting it in a separate file helps minimizing the use of this methods to the necessary abstractions or search which files uses methods for BiDi handling.

sev-:
Closing this in favor of #2238

PR #2220, opened 2020-05-04
SCUMM: Add Apple IIgs custom color palette

laurentlasalle:
I've added a unique palette for Apple IIgs games, as suggested by a comment in `palette.cpp`. The previous workaround was to use `tableC64Palette` as default.

Color information was pulled [from the bottom of this page](https://mrob.com/pub/xapple2/colors.html), which list converted 12 bits values documented in [Apple IIgs Technical Notes: IIgs 2523063 Master Color Values](https://archive.org/details/IIgs_2523063_Master_Color_Values/mode/2up). Rearranged the palette order to match `tableC64Palette` and tested it with Maniac Mansion (by comparing the scenes[ with screenshots from MobyGames](https://www.mobygames.com/game/apple2/maniac-mansion/screenshots/gameShotId,43107/). 

The only hickup is that, using real hardware, all text is displayed white regardless of context. ScummVM actual behaviour of matching text color with actors, and showing the parser, verbs and inventory using different colours, is more practical for the end user IMO. This behaviour is still present with this new color palette.

### DIFF
--- a/engines/scumm/palette.cpp
+++ b/engines/scumm/palette.cpp
@@ -164,6 +164,17 @@ void ScummEngine::resetPalette() {
 	static const byte tableHercGPalette[] = {
 		0x00, 0x00, 0x00, 	0x00, 0xFF, 0x00
 	};
+	
+	// Palette based on Apple IIgs Technical Notes: IIgs 2523063 Master Color Values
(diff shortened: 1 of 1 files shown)


sev-:
Sweet, thank you!

PR #2223, opened 2020-05-04
WINTERMUTE: Move plugins and externals to separate folder + one more stub

lolbot-iichan:
Original Wintermute Engine 1.x supports 2 types of extentions:
1. Importing external functions from DLLs, calling them from game script. This way DLLs like kernel32.dll and system32.dll can be used. See https://wiki.scummvm.org/index.php?title=Wintermute/Externals for details
2. Implementing additional functionality as WME object, created at game script with operator "new". Those objects should be exported by DLLs named "wme_%s.dll". See https://wiki.scummvm.org/index.php?title=Wintermute/Plugins for details.

Most known extentions were reimplemented at ScummVM in previous PRs, but they were hardcoded deeply into /base/ folder files. This PR moves them all to a separate /ext/ folder.

Also, a stub for "wme_3fstatistics.dll" is implemented to remove on-screen script error messages during playing 3f Interactive games (the Reversion trilogy).


engines/wintermute/base/base_game.cpp
```diff
@@ -3413,6 +3394,12 @@ bool BaseGame::externalCall(ScScript *script, ScStack *stack, ScStack *thisStack
 	}
 #endif
 
+#if 1
```
sev-, line 3397: Is this `#if` a leftover?

lolbot-iichan: It marks the block that should be removed to disable plugins. Similar `#if 1` marks enabling/disabling externals.

I thought of something similar to `#ifdef ENABLE_FOXTAIL` and `#ifdef ENABLE_HEROCRAFT`, so that build configuration can be switched between a vanilla engine for guys like https://jennibee.itch.io/dreamcat and extended engine for ScummVM and ResidualVM. However, I don't think that additional subengine is a good metaphor for this (since those extentions are from a pack of completely different games actually). 

What do you think I should do with those?

sev-: I see a little benefit in disabling plugins in general or striving for "vanilla" engine. The game-specific changes are well-contained, and even if they change the engine behaviour, they are guarded by checks, that is normal.

We implement subengines only when there are _substantial_ code additions or different hardware requirements such as a bigger screen. This is not applicable to WM because it is a high-end engine anyway.

So, either remove those ifs or put a comment above them.

lolbot-iichan: That sounds reasonalble. Removed those `#if` at 899c569

sev-:
Great, thank you! BTW, maybe you would like to join the team and commit your changes directly?

PR #2226, opened 2020-05-07
DIRECTOR: return early where textCast is null

rvanlaar:
Fixes a segfault for the rectangle not being valid and
resolves multiple test crashes in the buildbot.


engines/director/frame.cpp
```diff
@@ -857,6 +857,11 @@ void Frame::renderButton(Graphics::ManagedSurface &surface, uint16 spriteId) {
 
 void Frame::renderText(Graphics::ManagedSurface &surface, uint16 spriteId, Common::Rect *textRect) {
 	TextCast *textCast = (TextCast*)_sprites[spriteId]->_cast;
+	if (textCast == nullptr){
```
sev-, line 860: This formatting is incorrect. I will merge it and then fix. But please, follow our code formatting conventions.

PR #2227, opened 2020-05-07
ULTIMA8: Remove custom memory manager from engine

OMGPizzaGuy:
ULTIMA8: Remove custom memory manager from engine

Back in 2005 I added custom memory allocation to Pentagram because I thought it would be a cool project (it was), and maybe it would improve performance on older systems (unlikely).
With the inclusion of the Ultima 8 engine in ScummVM this code doesn't belong in the engine, and I would be more comfortable an active developer adding performance optimizations based on actual requirements rather than my unfounded speculations.

Question: Should pent_include.cpp be removed now since it contains only the namespace stubs?


mduggan:
Thanks! I had been eyeing that code and wondering if it was better to remove, so glad you thought the same.

I think yeah, `pent_include.cpp` can be removed entirely now. If you want to add that to this PR that'd be great, otherwise I'm happy to do it.

mduggan:
Cool, thanks!

PR #2228, opened 2020-05-08
SCI: add missing but implemented help commands for script related debug commands

deckarep:
For SCI debug console,

- Adds missing help doc for script related commands
- These commands were already implemented by missing the help console documentation

### DIFF
--- a/engines/sci/console.cpp
+++ b/engines/sci/console.cpp
@@ -409,6 +409,9 @@ bool Console::cmdHelp(int argc, const char **argv) {
 	debugPrintf(" dissect_script - Examines a script\n");
 	debugPrintf(" backtrace / bt - Dumps the send/self/super/call/calle/callb stack\n");
 	debugPrintf(" trace / t / s - Executes one operation (no parameters) or several operations (specified as a parameter) \n");
+	debugPrintf(" script_objects / scro - Shows all objects inside a specfied script\n");
+	debugPrintf(" script_steps - Shows the number of executed SCI operations\n");
+	debugPrintf(" script_strings / scrs - Shows all strings inside a specified script\n");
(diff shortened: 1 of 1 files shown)


digitall:
Pretty self-explanatory change, limited to SCI engine console... Commit message meets formatting conventions. Will merge as-is.

@deckarep : Thanks for your code contribution!

deckarep:
@digitall - 🎉 woohoo my very first ScummVM contribution...even though it was tiny! 😉

sluicebox:
LOL well I get up for a minute and come back and it's merged =)

Great catch! I'm laughing at how much time I've spent in that debugger without knowing about `script_objects` and `script_strings`. I would have used `script_objects` a lot!

`script_steps` is already in the help text although, it's the first under the "VM" section. In reviewing this I noticed `script_said` is also missing. Can you remove this now duplicate `script_steps` and add `script_said`? Maybe there are more?

I think these commands would be more at home under the "VM" section in the help text. That's also how they're grouped when registering. I know they start with the word script, because they're querying script objects, but the "Script" section seems more focused on debugging script code. The word script means several things in SCI. "VM" has query commands for things in scripts like objects, variables, etc. Just a thought, I don't feel strongly about that. `script_said` might go in the "Parser" section though.

sluicebox:
Sorry, I typed that prior to this getting merged, I'll just do those two touch-ups myself. Thank you again for the contribution!

deckarep:
@sluicebox - I know this may not be the forum for a follow up question but I’ll ask anyway: do you know if the `disasm` debug command is working correctly? I ask because the opcodes emitted don’t seem to match the opcode numbers in the ScummVM itself.


I am trying to figure out how to patch code for a game the way you have been doing it.

Thanks for the cleanup!

PR #2229, opened 2020-05-08
ALL: clang-format

janisozaur:
Apply clang-format to the sources


janisozaur:
Similar to https://github.com/scummvm/scummvm/pull/1776

sev-:
...and similarly to #1776 I have to close it. It must be done in small pieces and with sanity checking of every file manually.

PR #2231, opened 2020-05-08
SCI32: Add GK1 Italian translation patch detection

trembyle:
This patch can be found here, on the left side:
https://gkpatches.vogons.org/

It can technically be applied to any English DOS/Windows Floppy or CD release, and so I'm not sure how we want to handle the version checks. I've confirmed that this correctly detects and runs the game against the floppy that I patched.

If anyone has contact info for Enrico Rolfi, I'd be happy to send him a note.

### DIFF
--- a/engines/sci/detection_tables.h
+++ b/engines/sci/detection_tables.h
@@ -816,6 +816,14 @@ static const struct ADGameDescription SciGameDescriptions[] = {
 		AD_LISTEND},
 		Common::FR_FRA, Common::kPlatformDOS, ADGF_NO_FLAGS, GUIO_GK1_FLOPPY },
 
+	// Gabriel Knight - Italian DOS Floppy (translation patch by Enrico Rolfi)
+	// SCI interpreter version 2.000.000
+	{"gk1", "", {
+		{"alt.map", 0, "ecf55868f9327f2abaf237710a657db7", 1557},
+		{"resource.alt", 0, "b2198dc83b1d030b332e78380e2ad00c", 3058970},
+		AD_LISTEND},
+		Common::IT_ITA, Common::kPlatformDOS, ADGF_NO_FLAGS, GUIO_GK1_FLOPPY },
+
 	// Gabriel Knight - English DOS CD (from jvprat)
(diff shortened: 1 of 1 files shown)


sluicebox:
That's a cool fan translation, I like that they packaged it as a RESOURCE.ALT/ALT.MAP so you can drop it into any version. I really dig the wave file of the theme.

I'm not sure if this works as a single detection entry since the patch is extra resource files which don't replace existing ones. There's a detection entry for each of the five GK1 floppy versions: 1.0, 1.0a, 1.0b, German, and French. The patch files could be placed with any of these, even if they don't work, and this entry would match. It would also match corrupt or missing original files. For this to work and do the job of validating expected files I think we'd need to duplicate at least each of the three existing English entries and add these patch files on top of that. And then do that again for CD versions.

It's attractive to show the language as Italian upon detection because that's the language you see when you play, but right now this prompts the user if this is English or Italian since multiple entries match, but that answer doesn't affect the language you get, so we've added an extra step that looks like it does something but doesn't. Maybe. We have to be careful because that language field is what ScummVM passes on to the scripts so they can ask "what language am I?" See `SciEngine::setSciLanguage` and `SciEngine::getSciLanguage`. In the original interpreter that answer would be English unless overridden by RESOURCE.CFG or game scripts, but the RESOURCE.CFG included with the Italian patch doesn't override this. I don't know how that affects all the versions of GK1 but I know some localized versions depend on printLang. Other games break when it's not what they shipped with.

I'm not against adding detection entries for this, but given that the fan patch already works without error messages my enthusiasm drains when I think about what it would take to do this comprehensively and verify that it wouldn't have side effects. There's a lot of GK1 versions, localizations, Sierra patches, and NRS patches on top of those which GOG shipped. I would at least like to see this figured out for the CD version before it's merged.

trembyle:
@sluicebox Thanks for the feedback. Ideally we would want to see the resource.alt and alt.map files with those hashes and assume that it's the Italian translation regardless of other files. Based on your feedback, I can see that is not an option.

PR #2233, opened 2020-05-09
ULTIMA8: Fix detection of FR_FRA & DE_DEU from Gold Edition CD installs

OMGPizzaGuy:
The English usecode file eusecode.flx was not included in the installs for French or German versions of the game when installed from the Gold Edition CD. The checksums remain the same for the language specific usecode.


mduggan:
LGTM, thanks!

PR #2234, opened 2020-05-09
JANITORIAL: clang-format video/

janisozaur:
As discussed on discord, starting clang-format with `video/`


video/avi_decoder.cpp
```diff
@@ -43,46 +43,43 @@ namespace Video {
...
-#define ID_REC  MKTAG('r','e','c',' ')
-#define ID_VEDT MKTAG('v','e','d','t')
-#define ID_IDX1 MKTAG('i','d','x','1')
-#define ID_STRD MKTAG('s','t','r','d')
-#define ID_INFO MKTAG('I','N','F','O')
-#define ID_ISFT MKTAG('I','S','F','T')
-#define ID_DISP MKTAG('D','I','S','P')
-#define ID_PRMI MKTAG('P','R','M','I')
-#define ID_STRN MKTAG('s','t','r','n')
+#define ID_RIFF MKTAG('R', 'I', 'F', 'F')
```
sev-, line 46: This change needs to be reverted. We do not want FourCCs to be bloated like this.

video/avi_decoder.cpp
```diff
@@ -43,46 +43,43 @@ namespace Video {
...
+#define ID_ISFT MKTAG('I', 'S', 'F', 'T')
+#define ID_DISP MKTAG('D', 'I', 'S', 'P')
+#define ID_PRMI MKTAG('P', 'R', 'M', 'I')
+#define ID_STRN MKTAG('s', 't', 'r', 'n')
 
 // Stream Types
 enum {
 	kStreamTypePaletteChange = MKTAG16('p', 'c'),
-	kStreamTypeAudio         = MKTAG16('w', 'b')
+	kStreamTypeAudio = MKTAG16('w', 'b')
```
sev-, line 75: Why it did not allow enum to be aligned? Please revert.

video/avi_decoder.cpp
```diff
@@ -1038,13 +1035,12 @@ void AVIDecoder::AVIVideoTrack::setDither(const byte *palette) {
...
 }
 
-AVIDecoder::AVIAudioTrack::AVIAudioTrack(const AVIStreamHeader &streamHeader, const PCMWaveFormat &waveFormat, Audio::Mixer::SoundType soundType) :
-		AudioTrack(soundType),
-		_audsHeader(streamHeader),
-		_wvInfo(waveFormat),
-		_audioStream(0),
-		_packetStream(0),
-		_curChunk(0) {
+AVIDecoder::AVIAudioTrack::AVIAudioTrack(const AVIStreamHeader &streamHeader, const PCMWaveFormat &waveFormat, Audio::Mixer::SoundType soundType) : AudioTrack(soundType),
```
sev-, line 1038: This now looks terrible, needs to be reverted. Probably we need extending .clang-format to prevent continuation string be so extreme.

video/avi_decoder.cpp
```diff
@@ -1142,7 +1138,7 @@ AVIDecoder::TrackStatus::TrackStatus() : track(0), chunkSearchOffset(0) {
 AVIDecoder::OldIndex *AVIDecoder::IndexEntries::find(uint index, uint frameNumber) {
 	for (uint idx = 0, frameCtr = 0; idx < size(); ++idx) {
 		if ((*this)[idx].id != ID_REC &&
-				AVIDecoder::getStreamIndex((*this)[idx].id) == index) {
+		    AVIDecoder::getStreamIndex((*this)[idx].id) == index) {
```
sev-, line 1141: It was correct and let the developer distinguish the continuation string from the loop body. Needs to be reverted and rule extended if possible.

video/avi_decoder.h
```diff
@@ -28,13 +28,13 @@
 #include "common/rect.h"
 #include "common/str.h"
 
-#include "video/video_decoder.h"
 #include "audio/mixer.h"
+#include "video/video_decoder.h"
```
sev-, line 32: Why it sorts the includes?

video/bink_decoder.cpp
```diff
@@ -76,8 +76,8 @@ bool BinkDecoder::loadStream(Common::SeekableReadStream *stream) {
 	if ((id != kBIKfID) && (id != kBIKgID) && (id != kBIKhID) && (id != kBIKiID))
 		return false;
 
-	uint32 fileSize         = stream->readUint32LE() + 8;
-	uint32 frameCount       = stream->readUint32LE();
+	uint32 fileSize = stream->readUint32LE() + 8;
```
sev-, line 79: We want to leave this to developers. E.g. to not force alignment or not-alignment.

video/bink_decoder.cpp
```diff
@@ -287,39 +287,41 @@ BinkDecoder::BinkVideoTrack::BinkVideoTrack(uint32 width, uint32 height, const G
 	_surface.w = width;
 
 	// Compute the video dimensions in blocks
-	_yBlockWidth   = (width  +  7) >> 3;
-	_yBlockHeight  = (height +  7) >> 3;
-	_uvBlockWidth  = (width  + 15) >> 4;
+	_yBlockWidth = (width + 7) >> 3;
```
sev-, line 290: Another case of nice alignment which should be left intact.

video/bink_decoder.cpp
```diff
@@ -287,39 +287,41 @@ BinkDecoder::BinkVideoTrack::BinkVideoTrack(uint32 width, uint32 height, const G
...
+	_yBlockHeight = (height + 7) >> 3;
+	_uvBlockWidth = (width + 15) >> 4;
 	_uvBlockHeight = (height + 15) >> 4;
 
 	// The planes are sized according to the number of blocks
-	_curPlanes[0] = new byte[_yBlockWidth  * 8 * _yBlockHeight  * 8]; // Y
+	_curPlanes[0] = new byte[_yBlockWidth * 8 * _yBlockHeight * 8];   // Y
```
sev-, line 296: And here too.

video/bink_decoder.cpp
```diff
@@ -287,39 +287,41 @@ BinkDecoder::BinkVideoTrack::BinkVideoTrack(uint32 width, uint32 height, const G
...
 
 	initBundles();
 	initHuffman();
 }
 
 BinkDecoder::BinkVideoTrack::~BinkVideoTrack() {
 	for (int i = 0; i < 4; i++) {
-		delete[] _curPlanes[i]; _curPlanes[i] = 0;
-		delete[] _oldPlanes[i]; _oldPlanes[i] = 0;
+		delete[] _curPlanes[i];
```
sev-, line 321: Controversial change. Rules need to be enhanced.

video/bink_decoder.cpp
```diff
@@ -1086,15 +1083,22 @@ void BinkDecoder::BinkVideoTrack::readDCTCoeffs(VideoFrame &video, int32 *block,
...
-	int coefList[128];      int modeList[128];
-	coefList[listEnd] = 4;  modeList[listEnd++] = 0;
-	coefList[listEnd] = 24; modeList[listEnd++] = 0;
-	coefList[listEnd] = 44; modeList[listEnd++] = 0;
-	coefList[listEnd] = 1;  modeList[listEnd++] = 3;
-	coefList[listEnd] = 2;  modeList[listEnd++] = 3;
-	coefList[listEnd] = 3;  modeList[listEnd++] = 3;
+	int listEnd = 64;
+
+	int coefList[128];
```
sev-, line 1088: This also ruined the authors' idea.

I'm stopping here. Could you please extend the rules, so we have fewer changes, such as #include sorting, allowing several statements on the same line, allowing optional alignment etc.

video/avi_decoder.cpp
Mataniko: I don't think any of the built-in formatters support the type of wrapping we'd want here

video/avi_decoder.h
Mataniko: To group related includes, not necessarily a bad thing IMO on files that import a lot of headers

video/bink_decoder.cpp
Mataniko: That would be hard to determine what was intended vs unintended - if we want to improve code formatting in other areas this might be a place where we need to be more lax.

Mataniko:
I think we need to sort the changes into a few buckets:

1. Bad formatting - for example: things that don't comply with the ScummVM coding standard, constructor initialize lists, etc. These need to be addressed with either config changes to clang-format or extending some of the rules
2. Unformattable code - These are places where the author decided they want to layout code in a certain way, and that way is objectively easier to read - these are where we'll either need a special rule, or wrap around with a DNF block
3. Acceptable changes - These are format changes that aren't necessarily bad - just different than what the author had in mind. I think we need to be flexible here in order to benefit from clang-format as a whole. Some of the changes in `bink_decoder.cpp` fall under that IMO
4. Improved formatting - Obvious formatting changes that are better, like pointer operator consistency, extra whitespace, trailing spaces, indentation, etc.

sev-:
@Mataniko yep, nice summary. Want me to classify my notes?

Mataniko:
I think as a first step @janisozaur can address the current feedback

video/bink_decoder.cpp
ZvikaZ: Can we rethink about this?
I really think that having two statements in two line is risky, especially since we don't have mandatory braces.

Therefore, we might have something like:
`i++; j++;`
And someone says - hey - that needs to be only on specific conditions, and changes it to:

    if (nice_condition)
        i++; j++;

And here goes nothing...

Solution?
Either enforce having a `{` after every `if`, or disallow two statements in one line.
IMO the second is better:
- The first change is even more controversial than the second change
- Not permitting two statements in one line is a common practice (at least, in my experience, but YMMV)

Mataniko:
@janisozaur It's been 3 weeks since the last comment - are you still working on improving this? Thanks!

Mataniko:
Closing due to lack of activity

PR #2236, opened 2020-05-09
PINK: support hebrew version

BLooperZ:
Based on #2217
Changes include:
- BiDi text in macmenu and mactext
- RTL layout for macmenu (which is activated when running the hebrew version)
- Accelerator underline position
- Fix encoding for text in pages

I have also created a new `TextRenderer` as suggested in the opening comment of `font.h`.
this was made to allow to only think about the logical string and let the renderer abstract the conversion to visual text (so no need to track if currently working with logical string or visual, which I found very confusing).


graphics/text_renderer.cpp
```diff
@@ -0,0 +1,52 @@
...
+#ifdef USE_FRIBIDI
+#include <fribidi/fribidi.h>
+#endif
+
+namespace Graphics {
+
+namespace TextRenderer {
+
+void drawString(Surface *dst, const Graphics::Font &font, const Common::String &str, const Common::CodePage page, int x, int y, int w, uint32 color, TextAlign align, int deltax, bool useEllipsis) {
+    font.drawString(dst, convertBiDiString(str, page), x, y, w, color, align, deltax, useEllipsis);
```
sev-, line 35: I do not understand the need for these methods, and they're also poorly named. It makes much more thing to put convertBiDiString() right in place where it is called.

common/str-bidi.h
```diff
@@ -0,0 +1,45 @@
...
+#include "common/str-enc.h"
+
+namespace Common {
+
+/*
+ * Wrapper for GNU FriBidi implementation of the Unicode Bidirectional Algorithm
+ * For LTR (Left To Right) languages, returns the original input
+ * For RTL (Right To Left) languages, returns visual representation of a logical single-line input
+ */
+U32String convertBiDiU32String(const U32String &input);
```
sev-, line 37: No, I am with @criezy here, please move these to str.cpp and ustr.cpp respectively.

graphics/text_renderer.cpp
BLooperZ: I have found it hard to track if I'm using the string in logical order (before conversion) or visual order (after conversion) so I needed abstraction, and moving it directly to Font::drawString might introduce unexpected effects... I think this "high-level draw bidi string method" might help using bidi in other places (only have to think about logical order as the drawing abstract it away)... please tell me if you are still not convinced, and I will revert this API change.

common/str-bidi.h
BLooperZ: Ok, I will change it on #2217 and rebase

BLooperZ:
Introduced Common::UnicodeBiDiText which helps
to make it clear when using visual or logical string,
and to be able to use additional info from FriBiDi
(e.g. logical to visual position, which is used in underlineAccelerator)

If the API is ok, #2217 can be closed in favour of this changes.

sev-:
Thank you, now it is a good solution. Merging.

sev-:
Hi @BLooperZ  After recent rewrites and improvements (we added Win95-like rendering), the Hebrew rendering is off again. Will you have time to take a look into it? It should be relatively simple.

PR #2238, opened 2020-05-10
SDL: Return unicode keycode for Hebrew input

ZvikaZ:
Before this commit, when typing in Hebrew, SDL reported the equivalent
English keycode. Thus, if user typed Hebrew letter Aleph, it was
reported as ascii of 't' (or 'T', I don't remember...).

It's still impossible to type Hebrew letters in the GUI's input dialogs,
but it's a step forward.

This was supposed to be part of a bigger PR, adding support for Hebrew in SQ3 (SCI0), but I'm waiting with this PR until there will be a decision in PR #2217. However, it's taking time, and I realized it might help @aryanrawlani28 's work on the GUI, and @BLooperZ 's work on other engines, therefore, it's submitted alone.


sev-:
Thanks!

PR #2239, opened 2020-05-10
Ports: Pause token mk2

henke37:
Fix some missed uses of `Engine::pauseEngine` from #2147.

### DIFF
--- a/backends/events/default/default-events.cpp
+++ b/backends/events/default/default-events.cpp
@@ -160,11 +160,10 @@ bool DefaultEventManager::pollEvent(Common::Event &event) {
 		if (_vk->isDisplaying()) {
 			_vk->close(true);
 		} else {
+			PauseToken pt;
 			if (g_engine)
-				g_engine->pauseEngine(true);
+				pt = g_engine->pauseEngine();
 			_vk->show();
-			if (g_engine)
-				g_engine->pauseEngine(false);
 			forwardEvent = false;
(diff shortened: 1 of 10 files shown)


backends/events/default/default-events.cpp
```diff
@@ -160,11 +160,10 @@ bool DefaultEventManager::pollEvent(Common::Event &event) {
 		if (_vk->isDisplaying()) {
 			_vk->close(true);
 		} else {
+			PauseToken pt;
```
bluegr, line 163: You're not handling pt anywhere, so why are you introducing it as a new variable?

backends/events/ps3sdl/ps3sdl-events.cpp
```diff
@@ -37,9 +37,10 @@
  */
 void PS3SdlEventSource::preprocessEvents(SDL_Event *event) {
 	if (event->type == SDL_APP_DIDENTERBACKGROUND) {
+		PauseToken pt;
```
bluegr, line 40: Same. you're not handling pt anywhere, so why are you introducing it as a new variable?

henke37:
The lifetime of the `PauseToken` object is significant. The destructor has semantic value, it unpauses the engine once, matching the pausing required to get the token in the first place.

sev-:
Thanks, let's see the buildbot.

PR #2241, opened 2020-05-11
AGOS: Fix compatibility with strict-alignment architectures

dwatteau:
Proposed fix for a bug I've reported some years ago:
https://bugs.scummvm.org/ticket/6220

SIMON2-FR would crash really early with a SIGBUS on OpenBSD/loongson, which is a mips64el system with strict alignment constraints.

Aligning to `sizeof(byte *)` in alignTableMem(), and calling alignTableMem() in loadTextIntoMem() seems to be enough. It just took me a bit of time to test this :)

(Maybe alignTableMem() could be optimised a bit, but I'm not sure it's worth it.)

Tested with SIMON2-FR and SIMON1-FR on OpenBSD/loongson and on macOS Mojave (x86-64).

I don't really know how to test the "running out of memory" possible case that lordhoto reported above, though. Don't know which platform or use case would be more likely to hit it.

### DIFF
--- a/engines/agos/string.cpp
+++ b/engines/agos/string.cpp
@@ -332,6 +332,7 @@ void AGOSEngine::loadTextIntoMem(uint16 stringId) {
 
 			_tablesHeapPtr += size;
 			_tablesHeapCurPos += size;
+			alignTableMem();
 
 			if (_tablesHeapCurPos > _tablesHeapSize) {
 				error("loadTextIntoMem: Out of table memory");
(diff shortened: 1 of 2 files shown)


digitall:
@dwatteau : Thank you for the patch for this bug in the AGOS engine which has been open for quite a while. Hopefully an AGOS engine dev will review this shortly. If not, then I will give your patch a basic test with my copies of AGOS games to see if this seems good...

sev-:
Thanks

PR #2242, opened 2020-05-12
WINTERMUTE: Add FoxTail Chapter 3 support

lolbot-iichan:
FoxTail Chapter 3 was released several days ago:
https://store.steampowered.com/app/581360/FoxTail/
https://www.gog.com/game/foxtail

There were 2 builds released: 1.2.896.4370 and 1.2.896.4371. There is a very minor difference in english localization between them, other package files are the same.

FoxTail Engine 1.2.896 for some reason is using different opcodes enum than usual WME bytecode, so new scripts are not runnable by unmodified Wintermute at all. However, simple remapping of instruction just after `getDWORD()` is enough to fix everything.

Chapter 3 features a kalimba - playable musical instrument. Most of the chapter events are related to playing some melodies & catching tunes by watching how kalimba is resonating to a melody. However, there is an easter egg - kalimba can be also played with keyboard and result may be saved to a file and even open & replayed by other FoxTail player.

It took some time to support features related to replayer functionality, since ScummVM does read files from disk and SavefileManager and saves files to SavefileManager. `BaseFileManager::listMatchingFiles()` method was added to support "*.mb" files lookup in both places. During this fix I found out that `getNodeForRelativePath` was broken. It was tokenizing using "\\" delimeter, while text input was always containing "/" instead. I hope now when it's fixed, it won't fire back in some unrelated WME game.

All new keys, mostly related to kalimba, are added to keymapper. While adding keymaps for this, I noticed that keymapper_tables.h is not in POTFILES, so keymaps were out of localization for some time... Ooops. Fixed.

Refactored and fixed Split() code. This FoxTail method is now used not only with 1-character delimeters only, but with 2-character delimeters as well, so I replaced `*it == sep` with `strncmp(it, sep.c_str(), sep.size()) == 0`. A few issues related to this method were fixed as well.

Music crossfade code is extended to work with custom volume numbers and to work correctly if channel1==channel2. This refactored code was used to implement FoxTail's `MusicCrossfadeVolume()` method.

Half a dozen of other small FoxTail-related methods introduced in Engine 1.2.896 were also added here and there: `ValidSaveSlotVersion()` was stubbed, `IsNumber()`, `Trim()`, `lTrim()`, `rTrim()`, `GetSoundFilename()`, `GetDay()`, `HeightToFit()` were added, and also a `_lineHeight -= 1` hack is added while TTF display.

With those commits, FoxTail Chapter 3 is now playable and completable.
All the books in the game should be displayed correctly as well.


sev-:
Awesome, and excellent documentation.

PR #2244, opened 2020-05-12
ULTIMA8: Custom type checking cleanup

OMGPizzaGuy:
ULTIMA8: Custom type checking cleanup

Use template methods for type checking instead of referencing static ::ClassType variable directly.
Alter Gump::FindGump to check matches against a function instead of referencing the RunTimeClassType and remove no_inheritence option as it is never used. These changes should help migration from p_dynamic_cast if desired.

### DIFF
--- a/engines/ultima/ultima8/gumps/gump.cpp
+++ b/engines/ultima/ultima8/gumps/gump.cpp
@@ -533,8 +533,7 @@ bool Gump::GetLocationOfItem(uint16 itemid, int32 &gx, int32 &gy,
 }
 
 // Find a child gump that matches the matching function 
-Gump *Gump::FindGump(const FindPredicate predicate, bool recursive)
-{
+Gump *Gump::FindGump(const FindPredicate predicate, bool recursive) {
 	if ((this->*predicate)())
 		return this;
 
--- a/engines/ultima/ultima8/gumps/gump.cpp
+++ b/engines/ultima/ultima8/gumps/gump.cpp
@@ -532,13 +532,10 @@ bool Gump::GetLocationOfItem(uint16 itemid, int32 &gx, int32 &gy,
 	return false;
 }
 
-// Find a child gump of the specified type
-Gump *Gump::FindGump(const RunTimeClassType &t, bool recursive,
-                     bool no_inheritance) {
-	// If that is our type, then return us!
-	if (GetClassType() == t)
-		return this;
-	else if (!no_inheritance && IsOfType(t))
+// Find a child gump that matches the matching function 
+Gump *Gump::FindGump(const FindPredicate predicate, bool recursive)
+{
+	if ((this->*predicate)())
 		return this;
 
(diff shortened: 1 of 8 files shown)


engines/ultima/ultima8/gumps/gump.h
```diff
@@ -108,23 +108,20 @@ class Gump : public Object {
 	//! \param takefocus If true, set parent's _focusChild to this
 	virtual void                InitGump(Gump *newparent, bool take_focus = true);
 
-	//! Find a gump of the specified type (this or child)
-	//! \param t Type of gump to look for
+	typedef bool (Gump::*FindPredicate)() const;
```
OMGPizzaGuy, line 111: Should I consider a Common::UnaryFunction for this instead of member function pointer?

OMGPizzaGuy:
Another use of RunTimeClassType is for debugging information - specifically for the class name - in a number of console commands through the code. I'm curious to know what thoughts are on these commands.

OMGPizzaGuy:
The "bool IsOfType(const char * type)" functions are currently unused and I'm considering removing them as well.

engines/ultima/ultima8/gumps/gump.cpp
```diff
@@ -532,13 +532,10 @@ bool Gump::GetLocationOfItem(uint16 itemid, int32 &gx, int32 &gy,
...
-// Find a child gump of the specified type
-Gump *Gump::FindGump(const RunTimeClassType &t, bool recursive,
-                     bool no_inheritance) {
-	// If that is our type, then return us!
-	if (GetClassType() == t)
-		return this;
-	else if (!no_inheritance && IsOfType(t))
+// Find a child gump that matches the matching function 
+Gump *Gump::FindGump(const FindPredicate predicate, bool recursive)
+{
```
sev-, line 537: The opening braces should stick to the preceding lines by our Code formatting conventions

engines/ultima/ultima8/gumps/gump.h
mduggan: I don't have a strong feeling on this either way. Since the usage is fairly limited it may be easier to just do it like this?

mduggan:
> Another use of RunTimeClassType is for debugging information - specifically for the class name - in a number of console commands through the code. I'm curious to know what thoughts are on these commands.

I think now that most of the stuff is explicitly defined in the debugger class, the usefulness of this is lower than it was in Pentagram?
 
> It also appears to be used on save objects and processes, but load class names appears to be hard coded for object manager and the kernel's process loaders.

Yeah, the save/load stuff uses the names hard coded there.  I know because I broke my savegames by renaming the music class and had to fix it :)

mduggan:
Thanks for doing this cleanup! I'm also in favor of removing the unused code.  Since it's just on the U8 engine we probably don't need to leave open for comments so I'll merge now.

PR #2245, opened 2020-05-12
DRAGONS: Implementing midi music player

yuv422:
I'm starting to implement music for Blazing Dragons. It uses PS1 SEQ files for music. These files are almost identical to standard SMF MIDI but use specific sound bank for instruments.

I've copied the method that is used by the tinsel engine to support SEQ files.

I was thinking that we could load the sound bank samples automatically in the engine and convert them into the soundfont format to be used by midi drivers that support loading samples.

I haven't implemented the soundfont loading bit yet. It would probably require changes to base audio classes.

What does everyone think?

This PR implements the midi player. I've tested it with a manually converted VAB to SF2 file and it sounds good to my ears.

I could also apply this technique to the tinsel engine to add proper sound samples there too.

### DIFF
--- a/engines/dragons/dragons.cpp
+++ b/engines/dragons/dragons.cpp
@@ -654,7 +654,9 @@ void DragonsEngine::updateHandler() {
 
 	// 0x8001b200
 	if (isFlagSet(ENGINE_FLAG_8000) && !_sound->isSpeechPlaying()) {
+		//dialog finished playing.
 		clearFlags(ENGINE_FLAG_8000);
+		_sound->resumeMusic();
 	}
 
 	//TODO logic here
--- a/engines/dragons/midimusicplayer.cpp
+++ b/engines/dragons/midimusicplayer.cpp
@@ -1,0 +1,132 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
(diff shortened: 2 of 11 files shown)


yuv422:
Bump, anyone have any thoughts on this one? 
The gist of what I want to do. 

Convert PS1 sound bank data (VAB) into soundfont data in the engine and send it to the midi driver. Assuming the driver supports soundfont.

sev-:
I think unification would make sense, and moving these to the audio/ is a good thing.

athrxx:
But which Midi devices would run with that? If I understand correctly this would be limited to devices which support SF2 (like fluidsynth, timidity). It's a bit hard to tell, since that part is not included in the PR.

yuv422:
> But which Midi devices would run with that? If I understand correctly this would be limited to devices which support SF2 (like fluidsynth, timidity). It's a bit hard to tell, since that part is not included in the PR.

Yes this would be dependent on midi drivers that support soundfont. I haven't implemented that bit yet. I was just seeing what people though of the idea first.

I think I'll give it a go. 😄   I feel that requiring a soundfont capable midi driver isn't too much of an ask to support music in the game.

I don't really have the desire to re-implement a SEQ player from scratch at the moment. I guess someone could always come along later an implement it without using the midi system.

yuv422:
Ok I'll merge this now and do the refactor + soundfont loading in separate PRs

PR #2246, opened 2020-05-12
CONFIGURE: Add -fno-operator-names for clang/gcc

mduggan:
The use of operator names works fine on gcc and clang, but breaks MSVC.  Make
them error on gcc/clang too.

According to this random link, gcc and clang have supported this flag for a
really long time, so it shouldn't need an extra compile check:
https://phabricator.kde.org/D3850

Since I've never changed the `configure` file before I wanted to get more eyes on this change, so doing as a PR.

Background: My last PR (#2214) contained an `and`, which worked on all the buildbots, but broke when someone tried to compile with MSVC.  This should avoid similar breakage in the future.

### DIFF
--- a/configure
+++ b/configure
@@ -2275,6 +2275,7 @@ if test "$have_gcc" = yes ; then
 	else
 		append_var CXXFLAGS "-Wconversion"
 	fi;
+	append_var CXXFLAGS "-fno-operator-names"
 elif test "$have_icc" = yes ; then
 	add_line_to_config_mk 'CXX_UPDATE_DEP_FLAG = -MMD -MF "$(*D)/$(DEPDIR)/$(*F).d" -MQ "$@" -MP'
 fi;
--- a/engines/wintermute/wintermute.cpp
+++ b/engines/wintermute/wintermute.cpp
@@ -161,7 +161,7 @@ int WintermuteEngine::init() {
 	}
 
 	// check dependencies for games with high resolution assets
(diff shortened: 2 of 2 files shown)


mduggan:
Hmm.. this revealed another use of these operators..
```
engines/wintermute/wintermute.cpp:154:10: error: token is not a valid binary
      operator in a preprocessor subexpression
        #if not defined(USE_PNG) || not defined(USE_JPEG) || not defined...
            ~~~ ^
engines/wintermute/wintermute.cpp:165:10: error: token is not a valid binary
      operator in a preprocessor subexpression
        #if not defined(ENABLE_FOXTAIL)
            ~~~ ^
engines/wintermute/wintermute.cpp:176:10: error: token is not a valid binary
      operator in a preprocessor subexpression
        #if not defined(ENABLE_HEROCRAFT)
            ~~~ ^
```
I'll fix them too and add to the PR.

digitall:
This looks like a reasonable change since "#if !defined()" is more common code anyway... though maybe not as readable. This PR has been open for more than 4 weeks with no comments so I assume this is not a major issue?

Any objections to fixup and manual merge of this PR? @sev- ?

mduggan:
Rebased to save you the trouble of fixing up :)

digitall:
@mduggan : Thanks for the rebase... Will see if there are any objections from team. If not, I will look at merging this.

digitall:
Well looks like no objections forthcoming so will merge this as-is and deal with any buildbot breakage...

PR #2249, opened 2020-05-14
DIRECTOR: LINGO: Array comparison fixes

djsrv:
This fixes property array comparison and implements the special case when a longer array is compared with a shorter array.


sev-:
Excellent!

PR #2250, opened 2020-05-15
ULTIMA8: Fix loading of game saves after music process rename.

OMGPizzaGuy:
Both the old class name and the new class name for a process or object must be registered for load compatibility


mduggan:
Oops, thanks! I guess I hadn't actually saved+loaded any new games since I made that change..

PR #2251, opened 2020-05-15
ULTIMA8: More custom type checking cleanup

OMGPizzaGuy:
Remove unused IsOfType string-based methods and p_dynamic_cast template function.

Remaining IsOfType methods likely can be replaced with dynamic_cast as well


OMGPizzaGuy:
Attempts to simply make IsOfType use dynamic_cast aren't as straightforward as hoped - might need to brush up more on templates

mduggan:
All looks great, thanks!

PR #2253, opened 2020-05-15
DIRECTOR: LINGO: Remove quirk handling from <> op

djsrv:
Upon further testing, this behavior only applies to the = operator.


sev-:
Thanks!

PR #2254, opened 2020-05-16
ULTIMA8: Even more custom type checking cleanup 

OMGPizzaGuy:
Replace remaining uses of IsOfType with dynamic_cast.
Consolide DEFINE_RUNTIME_CLASSTYPE_CODE_* macros. 
Remove RunTimeClassType from classes that no longer need it.

The RunTimeClassType code is down only the classname used for saving and debugger info.

### DIFF
--- a/engines/ultima/ultima8/gumps/gump.h
+++ b/engines/ultima/ultima8/gumps/gump.h
@@ -37,6 +37,10 @@ class Shape;
 class Item;
 class GumpNotifyProcess;
 
+class Gump;
+typedef bool (*FindGumpPredicate)(Gump *g);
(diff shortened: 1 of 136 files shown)


engines/ultima/ultima8/gumps/gump.h
```diff
@@ -37,6 +37,10 @@ class Shape;
 class Item;
 class GumpNotifyProcess;
 
+class Gump;
+typedef bool (*FindGumpPredicate)(Gump *g);
+template<class T> inline bool IsOfType(Gump *g) { return dynamic_cast<T*>(g) != nullptr; }
```
OMGPizzaGuy, line 42: This perhaps could be more generic or be placed in a different file.

mduggan:
All looks good to me, merging.

PR #2257, opened 2020-05-16
PETKA: Engine doesn't use midi, mark the games as such

henke37:
Yet another "that's so henke37" detection entry update.


sev-:
Merging

PR #2259, opened 2020-05-17
BASE: Added 'test_new_standards.cpp'

ZvikaZ:
Adding C++11 compatibility tests, disabled by default.
See instructions inside the file.

### DIFF
--- a/base/module.mk
+++ b/base/module.mk
@@ -1,6 +1,7 @@
 MODULE := base
 
 MODULE_OBJS := \
+	test_new_standards.o \
 	main.o \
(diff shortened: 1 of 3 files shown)


henke37:
I'm a little surprised that it doesn't test move semantics. Think you can add a test for that?

ZvikaZ:
> I'm a little surprised that it doesn't test move semantics. Think you can add a test for that?
@henke37 , you got it.

sev-:
Cool, let's see how our ports will behave. Merging.

PR #2260, opened 2020-05-17
DIRECTOR: Begin rework of render pipeline

npjg:
This PR moves from clearing the entire screen each time to rendering only the parts of the screen that changed, as well as relocating drawing methods to the score.


sev-:
Very nice.

PR #2261, opened 2020-05-17
ULITIMA8: Improve basic avatar keyboard movement.

OMGPizzaGuy:
ULITIMA8: Improve basic avatar keyboard movement.

This adds continuous movement while the movement keys are depressed. I'm an unsure how Crusader should move, so this was just experimenting until the controls felt good for the Avatar.  I'm thinking a few adjustments might be keyboard jumping being a jump forward instead of towards mouse, and maybe shift modifier to run and control modifier to step.


mduggan:
Thanks! I had been meaning to do improvements like this so this is awesome.

This will change them to how they should work in crusader basically - it also uses shift to run btw.

Crusader is a little bit more subtle in that it effectively has a continuous target direction (with a target reticule showing the that direction) and the avatar rotates to face the closest he can.. also avatar has 16 combat frames and 8 walking ones so there's still a bunch of changes to do there, but that's all out of scope here.

For jump key, I feel the "targeted" jump button should go to the mouse cursor ideally, it makes getting through the jumping puzzles a lot less frustrating (and makes them possible on platforms without two mouse buttons).  The running jump doesn't work with the jump key at the moment though, and that should just go in current direction.

PR #2265, opened 2020-05-19
GUI: Add RTL support

aryanrawlani28:
This PR introduces RTL layout to ScummVM.  

A few screens to demonstrate the main parts of GUI, you can compare it directly with your build. Each screen is followed by a few comments about it.

It worked as a 2 step process:
- Flipping the GUI
- Flipping individual widgets where necessary.

In order to enable RTL in your build, you would need to change code in:
- In [`GuiManager.cpp`](https://github.com/scummvm/scummvm/blob/43ae5c0e738909ede449f82f83bec1deb2aa0670/gui/gui-manager.cpp)
A new function named `setLanguageRTL()` is added. See there for enabling RTL. For now, RTL is *enabled* by default, for *Hebrew*

1 - Main Screen
The home screen of ScummVM. Everything is right-aligned, including the tooltips. 

-----------------------------------------------------------------------------------

0 home
An example of individual widget flipping is the dropdown arrow you see on the left.

-----------------------------------------------------------------------------------

2 loadgame
Load Game menu. ~~A possible issue here is that the numbers need to be drawn on the right, which is not the case here.~~

-----------------------------------------------------------------------------------

3 loadgame icons
Other load game menu. ~~This is bugged, as you can see. The thing is - they are being drawn in the background correctly and even being loaded correctly. The only problem is the icon does not show for some reason.~~

-----------------------------------------------------------------------------------

Tabs

This shows the reversal of tabs, wherein the first tab starts from the right, and "next" tabs are located to the left. Individual widget flips of checkboxes and popup lists are also shown. ~~You'll notice that the last characters are being cut off, but this may be a local testing issue (because I'm testing in English)~~ (Now fixed)

-----------------------------------------------------------------------------------

4 scrollbars and keymap
~~A bug here is that the mouse events are being offset by widths of the button or something like that, and thus - mouse events are unstable for this menu.~~ (Now fixed)

-----------------------------------------------------------------------------------

10 - updated sliders
~~SliderWidget is a problem here. The values are correct (rightmost is 0 and going left increases the value), but the color should be reversed too.~~ Sliders are LTR by default for now, but the support is available. To change one slider to RTL, simply need to update its flag in the layout file

-----------------------------------------------------------------------------------

8 - Options - Editable
~~EditableWidgets are a problem here, and to be honest - I'm a little lost here. I have no idea how these should work and how the carets should be positioned. Will the input be changed, Will it start from the right side, which language will be used are some questions I don't know the answer to. Some help in this mini-task would be really helpful.~~
**For now, Editable and related tasks are parked for later.** They are currently aligned to the right with the normal functionalities, with input in English.

11 big screen
12 small screen

Bigger screens & smaller screens are now supported, across all themes. All themes also are now supported.
-----------------------------------------------------------------------------------
Some **Hebrew specific screens**

----------------------------------------------------------------------------------
Issues that need to be fixed or pending tasks, in order of their importance:

- [x] - AGI - Predictive Dialogs, the numbers are also being flipped with the GUI. Only the numbers need to be reversed again and we're good.
- [x] - LoadGame Icons do not show correctly
- [x] - Navigation within the options menus - `navLeft` & `navRight`. These are, for the moment, only being displayed correctly. No event handling is done when clicked.
- [x] - `SliderWidgets` - inverting the colors.
- [x] - If testing for the current theme passes, extending the RTL functionality to other [themes](https://github.com/scummvm/scummvm/tree/master/gui/themes) and also extending support for screens with a resolution of <400. (Currently, I've only made changes to the >399 resolution.)

Any suggestions or ways to enhance RTL are very welcome!


aryanrawlani28:
Codacy issues show the experimental code I did to see if the optional flag `rtl = 'no'` passes through to code or not, and one additional issue in which I incorrectly placed same value on both sides of the ternary operator in `tooltip.cpp`. Do I drop the former code issue? I'll fix it together with the latter.

gui/ThemeEngine.cpp
```diff
@@ -988,60 +1008,80 @@ void ThemeEngine::drawLineSeparator(const Common::Rect &r) {
...
+		r2.right = r2.left + radioButtonSize;
+	}
 
 	drawDD(dd, r2);
 
-	r2.left = r2.right + checkBoxSize;
-	r2.right = MAX(r2.left, r.right);
+	if (rtl) {
+		r2.left = r.left;
+		r2.right = r.right - (radioButtonSize * 2);
```
sev-, line 1077: Why `* 2`?

gui/ThemeLayout.cpp
```diff
@@ -242,6 +242,14 @@ void ThemeLayoutMain::reflowLayout(Widget *widgetChain) {
 		}
 	}
 
+	if (g_gui.useRTL()) {
+		if (this->_name == "GameOptions" || this->_name == "GlobalOptions") {
```
sev-, line 246: As discussed, this is not a good idea to hardcode names like this. What if you compare the dialog width with the overlay width?

gui/ThemeParser.cpp
```diff
@@ -684,6 +684,7 @@ bool ThemeParser::parserCallback_widget(ParserNode *node) {
 		var = node->values["name"];
 		int width = -1;
 		int height = -1;
+		bool useRTL = true;
```
sev-, line 687: Where is this used now?

gui/editgamedialog.cpp
```diff
@@ -344,6 +344,7 @@ EditGameDialog::EditGameDialog(const String &domain)
 
 	// Activate the first tab
 	tab->setActiveTab(0);
+
```
sev-, line 347: This needs to be removed.

gui/widget.cpp
```diff
@@ -111,6 +115,16 @@ void Widget::draw() {
 		_x = getAbsX();
 		_y = getAbsY();
 
+		if (g_gui.useRTL()) {
```
sev-, line 118: You also need to check against `nortl` flag for the current widget

gui/widget.cpp
```diff
@@ -54,6 +54,10 @@ void Widget::init() {
 	_next = _boss->_firstWidget;
 	_boss->_firstWidget = this;
 	_needsRedraw = true;
+
+	if (TransMan.getCurrentLanguage() == "C" && false) {		// GUI TODO: Change this and GuiManager::GuiManager() to use RTL.
```
sev-, line 58: I do not see you taking `nortl` XML flag into account

gui/ThemeEngine.cpp
aryanrawlani28: We're using RTL.
The line you mentioned is modifying the rectangle to draw a text.

Before that, we draw a RadioButton.
`if (rtl) {
		r2.left = r.right - radioButtonSize;
		r2.right = r.right;
	}
`
It looks something like this. The rectangle subtracts a radiobuttonsize and draws itself to the extreme right.

Next, if we use RTL, we subtract the original, already drawn radiobutton. Next, another one is used is subtracted for spacing.

Thus, `*2`.

gui/widgets/scrollcontainer.cpp
```diff
@@ -141,6 +141,13 @@ void ScrollContainerWidget::reflowLayout() {
 }
 
 void ScrollContainerWidget::drawWidget() {
+	// GUI TODO: Recheck what the below line does.
+	if (g_gui.useRTL()) {
+		if (this->_name.contains("GameOptions") || this->_name.contains("GlobalOptions")) {
```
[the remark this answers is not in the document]
aryanrawlani28: This was not really doing anything, so this code has been removed. I am still working on removing the hardcoded nature.

gui/ThemeLayout.cpp
aryanrawlani28: So, I looked into this and your idea is good, but it doesn't work here.

Here, I am trying to set the dialog paddings _only_ when I encounter an entire dialog. `ThemeLayoutMain::reflowLayout()` is being called many times for the widgets that are present inside the dialog menu, thus disturbing and breaking other things.

Eg - Once "GameOptions" is being called, then widgets and containers inside like "GameOptions_Graphics", etc are also being called. dialog width and overlay width are different for them, so they also fall true against the condition you suggested - breaking my intention.

gui/widget.cpp
aryanrawlani28: The drawing function is here for flipping the GUI.. the `noRTL` flag will only disable an internal widget from not flipping around. They have been added to each individual widgets where necessary.

aryanrawlani28:
@sev- I believe all the necessary changes have been made.

> "JANITORIAL: Code cleanup"
> 
> This commit is badly named: JANITORIAL means no functional changes. Please, change to 'GUI: Code cleanup' or something like this

This is now reworded: https://github.com/scummvm/scummvm/pull/2265/commits/a5dc1c87e638b6f23285ab11222e3a5352f17f96

gui/ThemeParser.cpp
aryanrawlani28: This is now fixed :)

gui/widget.cpp
aryanrawlani28: This is fixed now, every widget gets their internal flags from the function in `gui/object.cpp`, in `GuiObject::reflowLayout()`

gui/ThemeEngine.cpp
```diff
@@ -988,63 +1009,83 @@ void ThemeEngine::drawLineSeparator(const Common::Rect &r) {
...
+		r2.left = r.left;
+		r2.right = r.right - (checkBoxSize * 2);
+	} else {
+		r2.left = r2.right + checkBoxSize;
+		r2.right = r.right;
+	}
 
 	if (r2.right > r2.left) {
-		drawDDText(getTextData(dd), getTextColor(dd), r2, str, true, false, _widgets[kDDCheckboxDefault]->_textAlignH,
+		drawDDText(getTextData(dd), getTextColor(dd), r2, str, true, false, _widgets[dd]->_textAlignH,
```
aryanrawlani28, line 1046: Not sure if changing the default to using a particular dd's align is the best choice? Maybe I should choose from `kDDCheckboxDefault` or `kDDCheckboxDefaultRTL` to avoid some potential crashes?

gui/ThemeEngine.cpp
```diff
@@ -988,63 +1009,83 @@ void ThemeEngine::drawLineSeparator(const Common::Rect &r) {
...
+		r2.left = r.left;
+		r2.right = r.right - (radioButtonSize * 2);
+	} else {
+		r2.left = r2.right + radioButtonSize;
+		r2.right = MAX(r2.left, r.right);
+	}
 
-	drawDDText(getTextData(dd), getTextColor(dd), r2, str, true, false, _widgets[kDDRadiobuttonDefault]->_textAlignH,
-	           _widgets[dd]->_textAlignV);
+	drawDDText(getTextData(dd), getTextColor(dd), r2, str, true, false, _widgets[dd]->_textAlignH,
```
aryanrawlani28, line 1084: the same thing here (as with checkboxes)

gui/ThemeLayout.cpp
```diff
@@ -242,6 +246,14 @@ void ThemeLayoutMain::reflowLayout(Widget *widgetChain) {
 		}
 	}
 
+	if (g_gui.useRTL()) {
+		if (this->_name == "GameOptions" || this->_name == "GlobalOptions" || this->_name == "Browser") {
```
aryanrawlani28, line 250: TODO: Add comments about why particular names

gui/gui-manager.cpp
```diff
@@ -69,12 +69,19 @@ GuiManager::GuiManager() : _redrawStatus(kRedrawDisabled), _stateIsSaved(false),
 
 	_launched = false;
 
+	_useRTL = false;
+	_isAnotherDialogOverlayed = false;
```
aryanrawlani28, line 73: This needs to be cleaned out, I thought I removed this code

gui/widget.cpp
```diff
@@ -210,6 +226,10 @@ bool Widget::isVisible() const {
 	return !(_flags & WIDGET_INVISIBLE);
 }
 
+bool Widget::useRTL() const{
```
aryanrawlani28, line 229: space after const

gui/widgets/editable.cpp
```diff
@@ -46,6 +46,9 @@ void EditableWidget::init() {
 	_caretInverse = false;
 
 	_editScrollOffset = 0;
+	_caretStartPos = 0;
+	_forceOffset = false;
+	_rtlOffset = 0;
```
aryanrawlani28, line 51: Maybe _rtlCaretOffset would be a better name for this

gui/widgets/tab.cpp
```diff
@@ -119,6 +128,7 @@ int TabWidget::addTab(const String &title, const String &dialogName) {
 	if (newWidth < _minTabWidth)
 		newWidth = _minTabWidth;
 	newTab._tabWidth = newWidth;
+	_widthTillLastTab += newWidth;
```
aryanrawlani28, line 131: This wouldn't be a problem, but maybe it would be best if I were to wrap this around `if(useRTL)` also?

gui/ThemeEngine.cpp
```diff
@@ -988,63 +1009,83 @@ void ThemeEngine::drawLineSeparator(const Common::Rect &r) {
...
-	DrawData dd = kDDCheckboxDefault;
+	DrawData dd = rtl ? kDDCheckboxDefaultRTL : kDDCheckboxDefault;
 
 	if (checked)
-		dd = kDDCheckboxSelected;
+		dd = rtl ? kDDCheckboxSelectedRTL : kDDCheckboxSelected;
 
 	if (state == kStateDisabled)
-		dd = checked ? kDDCheckboxDisabledSelected : kDDCheckboxDisabled;
+		dd = checked ? rtl ? kDDCheckboxDisabledSelectedRTL : kDDCheckboxDisabledSelected : rtl ? kDDCheckboxDisabledRTL : kDDCheckboxDisabled;
```
aryanrawlani28, line 1023: Possibly an unreadable statement

gui/ThemeEval.cpp
```diff
@@ -50,7 +50,7 @@ void ThemeEval::reset() {
 	_layouts.clear();
 }
 
-bool ThemeEval::getWidgetData(const Common::String &widget, int16 &x, int16 &y, int16 &w, int16 &h) {
+bool ThemeEval::getWidgetData(const Common::String &widget, int16 &x, int16 &y, int16 &w, int16 &h, bool &useRTL) {
```
Mataniko, line 53: @sev- Should useRTL have a default value here since this method is used in Engines to create dialogs.

BLooperZ: IMO we should use default Graphics::TextAlignment in widgets, and only use the useRTL flag in GUI code to tell the GUI to set the default to kTextAlignRight

sev-: Guys, this is a _getter_, e.g. it tells what is specified in the XML theme. Of course, there is a default present, in the constructor of ThemeEval.

sev-: However, if have RTL ignored, then a viable alternative is to overload this method and have one without RTL parameter.

gui/ThemeEngine.cpp
sev-: Yes, split it in 2,` if (checked)`

gui/ThemeEval.cpp
sev-: ...and to clarify further. I was specifically asked Aryan to implement a flag, so any widget could be marked as `noRTL`

gui/ThemeEngine.cpp
```diff
@@ -988,63 +1009,87 @@ void ThemeEngine::drawLineSeparator(const Common::Rect &r) {
...
+		r2.right = r.right - (radioButtonSize * 2);
+	} else {
+		r2.left = r2.right + radioButtonSize;
+		r2.right = MAX(r2.left, r.right);
+	}
 
-	drawDDText(getTextData(dd), getTextColor(dd), r2, str, true, false, _widgets[kDDRadiobuttonDefault]->_textAlignH,
-	           _widgets[dd]->_textAlignV);
+	drawDDText(getTextData(dd), getTextColor(dd), r2, str, true, false, _widgets[dd]->_textAlignH,
+				_widgets[dd]->_textAlignV);
```
sev-, line 1089: It seems that here is an unneeded whitespace change. Could it be reverted?

gui/ThemeEval.cpp
sev-: After further discussions, here is the conclusion. @aryanrawlani28 could you please add here the following:

```
bool ThemeEval::getWidgetData(const Common::String &widget, int16 &x, int16 &y, int16 &w, int16 &h) {
   bool useRTL;

   return getWidgetData(widget, x, y, w, h, useRTL);
}
```

As a result, we will remove the necessity to modify many places which call this method. I will mark those.

gui/object.cpp
```diff
@@ -41,12 +41,13 @@ GuiObject::~GuiObject() {
 void GuiObject::reflowLayout() {
 	if (!_name.empty()) {
 		int16 w, h;
-
-		if (!g_gui.xmlEval()->getWidgetData(_name, _x, _y, w, h) || w == -1 || h == -1) {
+		bool useRTL = true;
```
sev-, line 44: Something is wrong with whitespace here.

engines/scumm/dialogs.cpp
```diff
@@ -309,7 +309,7 @@ void HelpDialog::reflowLayout() {
 
 	assert(lineHeight);
 
-	g_gui.xmlEval()->getWidgetData("ScummHelp.HelpText", x, y, w, h);
+	g_gui.xmlEval()->getWidgetData("ScummHelp.HelpText", x, y, w, h, _useRTL);
```
sev-, line 312: This could be reverted. Read my comment in ThemeEval.cpp

gui/options.cpp
```diff
@@ -2777,10 +2782,10 @@ void GlobalOptionsDialog::setupCloudTab() {
 	int16 shiftUp = 0;
 	if (!showingCurrentStorage || enabled) {
 		// "storage is disabled" hint is not shown, shift everything up
-		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDisabledHint", x, y, w, h))
+		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDisabledHint", x, y, w, h, _useRTL))
```
sev-, line 2785: This could be reverted

gui/options.cpp
```diff
@@ -2777,10 +2782,10 @@ void GlobalOptionsDialog::setupCloudTab() {
...
 			warning("GlobalOptions_Cloud_Container.StorageUsernameDesc's position is undefined");
 		shiftUp = y;
-		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageUsernameDesc", x, y, w, h))
+		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageUsernameDesc", x, y, w, h, _useRTL))
```
sev-, line 2788: This could be reverted.

gui/options.cpp
```diff
@@ -2830,10 +2835,10 @@ void GlobalOptionsDialog::setupCloudTab() {
 
 	int16 disconnectWidgetsAdditionalShift = 0;
 	if (!showDownloadButton) {
-		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDownloadHint", x, y, w, h))
+		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDownloadHint", x, y, w, h, _useRTL))
```
sev-, line 2838: This could be reverted.

gui/options.cpp
```diff
@@ -2830,10 +2835,10 @@ void GlobalOptionsDialog::setupCloudTab() {
...
 			warning("GlobalOptions_Cloud_Container.StorageDownloadHint's position is undefined");
 		disconnectWidgetsAdditionalShift = y;
-		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDisconnectHint", x, y, w, h))
+		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDisconnectHint", x, y, w, h, _useRTL))
```
sev-, line 2841: This could be reverted.

gui/options.cpp
```diff
@@ -2880,10 +2885,10 @@ void GlobalOptionsDialog::setupCloudTab() {
 	}
 
 	if (!shownConnectedInfo) {
-		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDisabledHint", x, y, w, h))
+		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageDisabledHint", x, y, w, h, _useRTL))
```
sev-, line 2888: This could be reverted.

gui/options.cpp
```diff
@@ -2880,10 +2885,10 @@ void GlobalOptionsDialog::setupCloudTab() {
...
 			warning("GlobalOptions_Cloud_Container.StorageUsernameDesc's position is undefined");
 		shiftUp = y;
-		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageWizardNotConnectedHint", x, y, w, h))
+		if (!g_gui.xmlEval()->getWidgetData("GlobalOptions_Cloud_Container.StorageWizardNotConnectedHint", x, y, w, h, _useRTL))
```
sev-, line 2891: This could be reverted.

gui/options.cpp
```diff
@@ -2903,7 +2908,7 @@ void GlobalOptionsDialog::shiftWidget(Widget *widget, const char *widgetName, in
 
 	int16 x, y;
 	int16 w, h;
-	if (!g_gui.xmlEval()->getWidgetData(widgetName, x, y, w, h))
+	if (!g_gui.xmlEval()->getWidgetData(widgetName, x, y, w, h, _useRTL))
```
sev-, line 2911: This could be reverted.

gui/saveload-dialog.cpp
```diff
@@ -491,7 +493,7 @@ void SaveLoadChooserSimple::reflowLayout() {
 		int16 x, y;
 		int16 w, h;
 
-		if (!g_gui.xmlEval()->getWidgetData("SaveLoadChooser.Thumbnail", x, y, w, h))
+		if (!g_gui.xmlEval()->getWidgetData("SaveLoadChooser.Thumbnail", x, y, w, h, _useRTL))
```
sev-, line 496: This could be reverted.

gui/saveload-dialog.cpp
```diff
@@ -923,7 +934,7 @@ void SaveLoadChooserGrid::reflowLayout() {
 
 	int16 x, y;
 	int16 w;
-	if (!g_gui.xmlEval()->getWidgetData("SaveLoadChooser.List", x, y, w, availableHeight))
+	if (!g_gui.xmlEval()->getWidgetData("SaveLoadChooser.List", x, y, w, availableHeight, _useRTL))
```
sev-, line 937: This could be reverted.

gui/Tooltip.cpp
```diff
@@ -71,22 +75,25 @@ void Tooltip::drawDialog(DrawLayer layerToDraw) {
...
-			Graphics::kTextAlignLeft,
+			textAlignment,
 			ThemeEngine::kTextInversionNone,
 			0,
 			false,
 			ThemeEngine::kFontStyleTooltip,
 			ThemeEngine::kFontColorNormal,
 			false
 		);
+		}
```
sev-, line 96: This is a no-op whitespace change, please revert, same goes with the change below.

gui/gui-manager.cpp
```diff
@@ -574,7 +580,11 @@ void GuiManager::processEvent(const Common::Event &event, Dialog *const activeDi
 		return;
 	int button;
 	uint32 time;
+
```
sev-, line 583: Please revert.

gui/gui-manager.cpp
```diff
@@ -647,6 +661,27 @@ void GuiManager::setLastMousePos(int16 x, int16 y) {
 	_lastMousePosition.time = _system->getMillis(true);
 }
 
+void GuiManager::setLanguageRTL() {
+	if (ConfMan.hasKey("guiRTL")) {
```
sev-, line 665: Please add a comment here, something like "put guiRTL=yes to your scummvm.ini to enforce RTL"

gui/predictivedialog.cpp
```diff
@@ -72,16 +72,42 @@ PredictiveDialog::PredictiveDialog() : Dialog("Predictive") {
...
-	_button[kButton0Act] = new ButtonWidget(this, "Predictive.Button0", "0"           , nullptr, kBut0Cmd);
+
+	if (g_gui.useRTL()) {
+		/** If using RTL, swap the internal name of odd rows, to be flipped again when drawing.
+			We flip them back to orignal, because the keyboard layout stays the same in LTR & RTL.
+			The rest, like okButton, cancel, etc are all flipped.
+		*/
+
+		// Row 1
+		_button[kButton1Act] = new ButtonWidget(this, "Predictive.Button3", "1  `-.&", nullptr, kBut1Cmd);
```
sev-, line 83: Would it be possible to sort it by row, so, the layout is clear?

3 2 1
6 5 4
9 8 7

gui/widgets/edittext.cpp
```diff
@@ -55,6 +55,10 @@ EditTextWidget::EditTextWidget(GuiObject *boss, const String &name, const String
 void EditTextWidget::setEditString(const String &str) {
 	EditableWidget::setEditString(str);
 	_backupString = str;
+
+	if (str.equals(""))
```
sev-, line 59: it is better to write `if (str.empty())`

gui/widgets/edittext.cpp
```diff
@@ -74,6 +78,10 @@ void EditTextWidget::handleMouseDown(int x, int y, int button, int clickCount) {
 
 	x += _editScrollOffset;
 
+	if (g_gui.useRTL()) {
+		x = g_gui.getStringWidth(_editString) - x + 14;
```
sev-, line 82: Why `+ 14` ?

gui/widgets/edittext.cpp
```diff
@@ -85,8 +93,11 @@ void EditTextWidget::handleMouseDown(int x, int y, int button, int clickCount) {
 			break;
 		last = cur;
 	}
+
 	if (setCaretPos(i))
 		markAsDirty();
+	else if (g_gui.useRTL())
```
sev-, line 99: What about writing
```
  if (setCaretPos(i) || g_gui.useRTL())
     markAsDirty();
```

gui/widgets/list.cpp
```diff
@@ -571,16 +574,40 @@ void ListWidget::drawWidget() {
...
+		}
+
 		if (_selectedItem == pos && _editMode) {
 			buffer = _editString;
 			color = _editColor;
 			adjustOffset();
-			g_gui.theme()->drawText(Common::Rect(_x + r.left, y, _x + r.right, y + fontHeight - 2), buffer, _state,
-			                        Graphics::kTextAlignLeft, inverted, pad, true, ThemeEngine::kFontStyleBold, color);
+			g_gui.theme()->drawText(r1, buffer, _state,
+									alignment, inverted, pad, true, ThemeEngine::kFontStyleBold, color);
```
sev-, line 594: Both of these `g_gui.theme()->drawText()` calls are equivalent, thus, they could be moved out of the `if` statement and only one copy are left.

gui/widgets/popup.cpp
```diff
@@ -387,14 +391,35 @@ void PopUpDialog::drawMenuEntry(int entry, bool hilite) {
...
+
+	if (g_gui.useRTL()) {
+		if (_twoColumns) {
+			r1.translate(this->getWidth() - w, 0);
+		}
+
+		r2.left = g_system->getOverlayWidth() - r2.left - w + g_gui.getOverlayOffset();
+		r2.right = r2.left + w;
+
+		if (_boss->getFlags() == 305) {
```
sev-, line 406: What is `305`? Please use the constants, not a magic value.

gui/widgets/popup.cpp
```diff
@@ -510,7 +535,11 @@ void PopUpWidget::drawWidget() {
 	Common::String sel;
 	if (_selectedItem >= 0)
 		sel = _entries[_selectedItem].name;
-	g_gui.theme()->drawPopUpWidget(Common::Rect(_x, _y, _x + _w, _y + _h), sel, _leftPadding, _state);
+
+	if (g_gui.useRTL() && _useRTL)
+		_leftPadding = 0;
```
sev-, line 540: Maybe it should be something like `padding = _rightPadding` for RTL and  `padding = _leftPadding` for LTR?

gui/widgets/tab.cpp
```diff
@@ -360,23 +384,59 @@ void TabWidget::reflowLayout() {
...
 }
 
 void TabWidget::drawWidget() {
 	Common::Array<Common::String> tabs;
 	Common::Array<int> widths;
-	for (int i = _firstVisibleTab; i <= _lastVisibleTab; ++i) {
-		tabs.push_back(_tabs[i].title);
-		widths.push_back(_tabs[i]._tabWidth);
+	if (g_gui.useRTL()) {
+		for (int i = _firstVisibleTab; i <= _lastVisibleTab; ++i) {
```
sev-, line 396: Please, rewrite it, so you're using only `push_back`. Use trick with `int idx = _lastVisibleTab - i` which I showed you recently. `insert_at(0, ...)` is ugly and expensive.

gui/widgets/popup.cpp
aryanrawlani28: This ties back to the `x += deltax;` in `graphics/font.cpp`.
I have to put paddings as 0, otherwise they will go out of bound.

In order to make these work, I would have to reintroduce the thing earlier with
`if (align == kTextAlignRight) then x -= deltax`

I know you gave approval on this, but all the last char issues are now fixed. If I were to enable the above code, it would allow me to add some padding. I thought it would be best if I asked again.

gui/widgets/popup.cpp
aryanrawlani28: I haven't figured out a way to center align texts from dropdownbutton widgets. I've left a FIXME note just below this line for now

gui/widgets/popup.cpp
aryanrawlani28: Fixed now, all right aligns now have a neat little padding :)

gui/widgets/edittext.cpp
aryanrawlani28: Added comment explaining why. Quick overview - the click needs to match the exact position where the text starts drawing (when right aligned)

gui/object.cpp
aryanrawlani28: I couldn't find out the issue here. I deleted this and wrote it again, but git diff didn't show any change

gui/widgets/popup.cpp
aryanrawlani28: this was wrong code, now removed

sev-:
Thanks, merging.

PR #2268, opened 2020-05-22
SCI: fix bug #11476 (QFG1: Incomplete Music)

athrxx:
This fixes a regression from 3154d57. The fix is based on my findings in the QFG1 DOS disasm (counterchecked a bit with PQ2 DOS disasm).


Deledrius:
I can confirm this fixes the problems I reported in [bug #11476](https://bugs.scummvm.org/ticket/11476).

bluegr:
Great work, thanks! Merging

PR #2270, opened 2020-05-22
DIRECTOR: Fix file path loading

rvanlaar:
Filepaths are converted between mac and windows filepaths. In some cases the filenames were converted correctly, without doing a pass. A filepath conversion would return the same filepath. This wasn't recognized as a conversion that succeeded.

The fix is check if the resulting filename can be opened.


sev-:
Yes, that is the most straightforward way to fix it.

PR #2271, opened 2020-05-23
ULTIMA8: Refactor responsibility of class name save for objects and processes

OMGPizzaGuy:
ULTIMA8: Refactor responsibility of class name save for objects and processes

Objects and processes were responsible for saving the name that identifies them but cannot be responsible for the load of that class name as it is used to determine what object to construct. The better choice is to force the kernel and object manager to be responsible for both saving and loading of these class type headers.

Ideally, this will lead into a future change were the registration of the class for loading would also be used for registering known types for saving and possibly eliminate errors that occur when renaming or adding a class.

### DIFF
--- a/engines/ultima/ultima8/kernel/kernel.cpp
+++ b/engines/ultima/ultima8/kernel/kernel.cpp
@@ -327,6 +327,14 @@ void Kernel::save(Common::WriteStream *ws) {
 		const char *cname = (*it)->GetClassType()._className; // virtual
 		uint16 clen = strlen(cname);
 
+		Std::map<Common::String, ProcessLoadFunc>::iterator iter;
+		iter = _processLoaders.find(cname);
+
+		if (iter == _processLoaders.end()) {
+			perr << "Process class cannot save without registered loader: " << cname << Std::endl;
+			continue;
+		}
+
 		ws->writeUint16LE(clen);
 		ws->write(cname, clen);
 		(*it)->saveData(ws);
--- a/engines/ultima/ultima8/kernel/kernel.cpp
+++ b/engines/ultima/ultima8/kernel/kernel.cpp
@@ -324,7 +324,12 @@ void Kernel::save(Common::WriteStream *ws) {
 	_pIDs->save(ws);
 	ws->writeUint32LE(_processes.size());
 	for (ProcessIterator it = _processes.begin(); it != _processes.end(); ++it) {
-		(*it)->save(ws);
+		const char *cname = (*it)->GetClassType()._className; // virtual
+		uint16 clen = strlen(cname);
+
+		ws->writeUint16LE(clen);
+		ws->write(cname, clen);
+		(*it)->saveData(ws);
 	}
 }
 
--- a/engines/ultima/ultima8/kernel/kernel.cpp
+++ b/engines/ultima/ultima8/kernel/kernel.cpp
@@ -324,14 +324,13 @@ void Kernel::save(Common::WriteStream *ws) {
 	_pIDs->save(ws);
 	ws->writeUint32LE(_processes.size());
 	for (ProcessIterator it = _processes.begin(); it != _processes.end(); ++it) {
-		Std::string classname = (*it)->GetClassType()._className; // virtual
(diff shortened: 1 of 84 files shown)


OMGPizzaGuy:
I'm considering having ObjectManager::saveObject and Kernel::save check the _objectLoaders & _processLoaders for a matching class name and fail save if missing. This would mean saves are still dependent on GetClassType()._className and wouldn't solve the class renaming issue, but at least would catch a missing load on a new class.

Alternately, we could have a list of registered structs with the RunTimeClassType and the type name used on save so they could potentially be disassociated.

engines/ultima/ultima8/kernel/kernel.cpp
```diff
@@ -324,7 +324,20 @@ void Kernel::save(Common::WriteStream *ws) {
...
-		(*it)->save(ws);
+		const char *cname = (*it)->GetClassType()._className; // virtual
+		uint16 clen = strlen(cname);
+
+		Std::map<Common::String, ProcessLoadFunc>::iterator iter;
+		iter = _processLoaders.find(cname);
+
+		if (iter == _processLoaders.end()) {
+			perr << "Process class cannot save without registered loader: " << cname << Std::endl;
+			continue;
```
OMGPizzaGuy, line 335: This would prevent a completely broken save but would be incomplete. Is it better to fail the save entirely or save the object even though load will fail?

engines/ultima/ultima8/kernel/object_manager.cpp
```diff
@@ -288,6 +288,22 @@ bool ObjectManager::load(Common::ReadStream *rs, uint32 version) {
...
 
+void ObjectManager::saveObject(Common::WriteStream *ws, Object *obj) const {
+	const char *cname = obj->GetClassType()._className; // note: virtual
+	uint16 clen = strlen(cname);
+
+	Std::map<Common::String, ObjectLoadFunc>::iterator iter;
+	iter = _objectLoaders.find(cname);
+	if (iter == _objectLoaders.end()) {
+		perr << "Object class cannot save without registered loader: " << cname << Std::endl;
+		return;
```
OMGPizzaGuy, line 299: Same issue as save process without load ability

engines/ultima/ultima8/kernel/kernel.cpp
OMGPizzaGuy: Skipping the process /object save causes other assertions to fail, so what's currently here is definitely not a good answer.

engines/ultima/ultima8/kernel/kernel.cpp
```diff
@@ -324,7 +324,19 @@ void Kernel::save(Common::WriteStream *ws) {
 	_pIDs->save(ws);
 	ws->writeUint32LE(_processes.size());
 	for (ProcessIterator it = _processes.begin(); it != _processes.end(); ++it) {
-		(*it)->save(ws);
+		Std::string classname = (*it)->GetClassType()._className; // virtual
```
mduggan, line 327: Could probably avoid a copy here by making it const & I think

engines/ultima/ultima8/kernel/object_manager.cpp
```diff
@@ -288,6 +288,21 @@ bool ObjectManager::load(Common::ReadStream *rs, uint32 version) {
 	return true;
 }
 
+void ObjectManager::saveObject(Common::WriteStream *ws, Object *obj) const {
+	Std::string classname = obj->GetClassType()._className; // note: virtual
```
mduggan, line 292: `const &` would be nicer here too.

engines/ultima/ultima8/kernel/object_manager.cpp
mduggan: Since this is the save process, maybe best to assert here with `error()`? Unlike load (where maybe the file is corrupt or it comes from an incompatible version), it suggests a coding error if save fails.

mduggan:
Overall I really like this refactor, it feels cleaner if the save and load are registered in the same list.

It makes me wonder if we need a better way to signal errors in the save calls, but that can be out of scope for this change.

PR #2273, opened 2020-05-23
MIDI/KYRA: Miles channel locking and GM support

NMIError:
This PR improves the XMIDI channel locking implementation from the KYRA engine
and moves it to the generic Miles MIDI driver to reduce duplication and for
possible re-use. It also adds GM support to this driver. The KYRA engine has
been updated to use the improved driver.

To support channel locking, the XMIDI parser and generic MIDI driver can now
pass a source number, which allows MIDI driver implementations to distinguish
different MIDI sources (like background music and sound effects).

The generic MIDI driver now has functions for initializing MT-32 and GM/GS
devices, including compatibility for playing one type of MIDI data on the
other type of device.

This fixes issues in Legend of Kyrandia where volume levels were not properly
set after unlocking a channel:
- In the intro, when Brandon is lifted up to the tree house, several sound
effects were silent due to volume being 0.
- At the start of the game, when speaking to the tree, after the "creaking"
sound effect, some music channels are too loud.

It also adds support for the "enable GS" option so the game can play MT-32
music on the MT-32 instrument banks of a Roland GS device.
I've tested these changes on Legend of Kyrandia, Hand of Fate, Lands of Lore
and The 7th Guest (for unintended changes). I don't have games from the other
engines which make use of the Miles driver, but I've tried to avoid any
significant changes to the current functionality.


NMIError:
Regarding the final Codacy issue: I did not change that code and I don't see what's wrong with it, so if someone could help me out, that would be great. :-)

athrxx:
Wow, that's a lot to test/review.
Don't worry about codacy. We're not paricularly interested in that.

athrxx:
Seems like a great improvement. All seems to work as intended so far.

I just noticed that I had killed the Kyra1 PC Speaker driver. I was blaming it on you, of course :-)
But now that one also seems to run fine on the new code (the driver basically pretends to be an mt-32 driver). I'll do some more testing tomorrow.

athrxx:
Do you happen to know whether your changes would have a positive impact on the Simon2 issues, too? The driver has been disabled there and the author never got around to fix it (agos/midi.cpp, line 117).

NMIError:
Hi, thanks for looking at my PR. I noticed too yesterday that the PC speaker support was no longer working compared to the last stable version. I was very confused as to how it had ever worked in the first place, considering that the PC speaker sound resources aren't in XMIDI format... :-)

Regarding Simon2, I don't have the game, so I'll have to go by the comment in midi.cpp. The biggest problems seem to be with the AdLib driver, which hasn't changed. MT-32 apparently already works, so I guess that would work with the new version of the driver as well.

athrxx:
I fixed the PC speaker issue yesterday. It was some experimental/debugging nonsense I must have left in the code when I made the PC speaker driver for EOB. The Kyra driver does use XMIDI format files, if you actually do load the correct file... :-)

athrxx:
This would be much easier to review if it weren't all squashed into one single commit (why not first make the changes to the Kyra version of the Midi driver, then make the changes to the common driver and then do the renaming). But I don't want to complain, since it really is great work 👍 

We'll wait a couple of days, since other devs whose engines are affected might be interested in this, too.

NMIError:
Yeah sorry about the large commit. The problem was that this PR went through a couple of revisions, and the older commits contained a lot of code that was later removed or reworked. I figured that was just confusing, so I put it into a single large commit...

NMIError:
BTW the build keeps breaking on something in Blade Runner apparently; not sure what's going on there.

PR #2275, opened 2020-05-25
ULTIMA8: Add directional movement controls in addition to tank movement

OMGPizzaGuy:
ULTIMA8: Add directional movement controls in addition to tank movement

Experimenting with directional movement as tank controls felt awkward on a controller.


engines/ultima/ultima8/world/actors/avatar_mover_process.h
```diff
@@ -89,6 +93,10 @@ class AvatarMoverProcess : public Process {
 	bool _tryTurnRight;
 	bool _tryMoveForward;
 	bool _tryMoveBack;
+	bool _tryMoveLeft;
+	bool _tryMoveRight;
+	bool _tryMoveUp;
+	bool _tryMoveDown;
```
OMGPizzaGuy, line 99: There's starting to be quite a few of these. Should these be switched flags for a movement state variable?

engines/ultima/ultima8/meta_engine.cpp
```diff
@@ -69,12 +73,12 @@ static const KeybindingRecord CHEAT_KEYS[] = {
...
-	{ ACTION_MOVE_RIGHT, "MOVE_RIGHT", "Move Right", "QuickAvatarMoverProcess::startMoveRight", "QuickAvatarMoverProcess::stopMoveRight", "A+RIGHT", nullptr },
+	{ ACTION_QUICK_MOVE_ASCEND, "ASCEND", "Ascend", "QuickAvatarMoverProcess::startAscend", "QuickAvatarMoverProcess::stopAscend", "HOME", nullptr },
+	{ ACTION_QUICK_MOVE_DESCEND, "DESCEND", "Descend", "QuickAvatarMoverProcess::startDescend", "QuickAvatarMoverProcess::stopDescend", "END", nullptr },
+	{ ACTION_QUICK_MOVE_UP, "MOVE_UP", "Move Up", "QuickAvatarMoverProcess::startMoveUp", "QuickAvatarMoverProcess::stopMoveUp", "A+UP", nullptr },
+	{ ACTION_QUICK_MOVE_DOWN, "MOVE_DOWN", "Move Down", "QuickAvatarMoverProcess::startMoveDown", "QuickAvatarMoverProcess::stopMoveDown", "A+DOWN", nullptr },
+	{ ACTION_QUICK_MOVE_LEFT, "MOVE_LEFT", "Move Left", "QuickAvatarMoverProcess::startMoveLeft", "QuickAvatarMoverProcess::stopMoveLeft", "A+LEFT", nullptr },
+	{ ACTION_QUICK_MOVE_RIGHT, "MOVE_RIGHT", "Move Right", "QuickAvatarMoverProcess::startMoveRight", "QuickAvatarMoverProcess::stopMoveRight", "A+RIGHT", nullptr },
```
OMGPizzaGuy, line 81: I think I need to rename the id here too for saving of bindings, right?

engines/ultima/ultima8/world/actors/avatar_mover_process.h
mduggan: Yeah it might be nicer - that would also avoid impossible combinations (like trymoveup + trymovedown at the same time).  Doesn't have to be part of this change.

engines/ultima/ultima8/meta_engine.cpp
mduggan: Yeah, I think these should not have the same name as the ones above.

OMGPizzaGuy: I've confirmed that the cheats section is on a separate keymap. For the ini all the cheats have the prefix keymap_ultima8_cheats_ and should be fine without renaming the id. I still can if preferred and have locally.

mduggan: If it's not ambiguous I don't have a strong opinion.. maybe just leave it like this then.

PR #2276, opened 2020-05-25
BACKENDS: Discord rich presence support

SupSuper:
Reworked version of @Mataniko 's PR: https://github.com/scummvm/scummvm/pull/1247

This adds support for [Discord rich presence](https://discord.com/developers/docs/rich-presence/how-to), showing the currently running game in a user's status (if they have Game Activity enabled):
image

Based on the similar taskbar integration. Tested on Windows, should also work on OSX and Linux (welcome testing). You'll need to install [discord-rpc](https://github.com/discord/discord-rpc/releases), it's deprecated but still works (the new Discord SDK isn't GPL-compatible).

The game images have to be provided by us via the [Discord Developer Portal](https://discord.com/developers/applications), instead of provided by the player. I created a ScummVM team with @sev- as admin so he has full control over it.

In my testing, game icons look better than screenshots, but most of the links in this old thread are broken: https://forums.scummvm.org/viewtopic.php?f=1&t=4127
Plus I don't know if it's kosher to use community resources as official.

### DIFF
--- a/configure
+++ b/configure
@@ -169,6 +169,7 @@ _iconv=auto
 _tts=auto
 _gtk=auto
 _fribidi=auto
+_discord=auto
 _test_cxx11=no
 # Default option behavior yes/no
 _debug_build=auto
@@ -1158,6 +1159,9 @@ Optional Libraries:
   --with-iconv-prefix=DIR    prefix where libiconv is installed (optional)
   --disable-iconv        disable libiconv encoding conversion library [autodetect]
 
+  --with-discord-prefix=DIR   prefix where discord-rpc is installed (optional)
+  --disable-discord       disable Discord rich presence integration [autodetect]
+
 Some influential environment variables:
   AR                 archiver command
   AS                 assembler command
@@ -1289,6 +1293,8 @@ for ac_option in $@; do
 	--disable-gtk)                _gtk=no                ;;
 	--enable-bink)                _bink=yes              ;;
 	--disable-bink)               _bink=no               ;;
+	--enable-discord)             _discord=yes           ;;
+	--disable-discord)            _discord=no            ;;
 	--opengl-mode=*)
 		_opengl_mode=`echo $ac_option | cut -d '=' -f 2`
 		;;
@@ -1424,6 +1430,11 @@ for ac_option in $@; do
 		arg=`echo $ac_option | cut -d '=' -f 2`
 		ICONV_CFLAGS="-I$arg/include"
 		ICONV_LIBS="-L$arg/lib"
+		;;
+	--with-discord-prefix=*)
+		arg=`echo $ac_option | cut -d '=' -f 2`
+		DISCORD_CFLAGS="-I$arg/include"
+		DISCORD_LIBS="-L$arg/lib"
 		;;
 	--backend=*)
 		_backend=`echo $ac_option | cut -d '=' -f 2`
@@ -5440,6 +5451,25 @@ if test "$_pandocext" = "default"; then
 		_pandocext=".$_pandocformat"
 	fi
 fi
+
+#
+# Check for Discord
+#
+echocheck "Discord RPC"
+if test "$_discord" = auto ; then
+	_discord=no
+	cat > $TMPC << EOF
+#include <discord_rpc.h>
+int main(void) { Discord_Shutdown(); return 0; }
+EOF
+	cc_check $DISCORD_CFLAGS $DISCORD_LIBS -ldiscord-rpc && _discord=yes
+fi
+if test "$_discord" = yes ; then
+	append_var LIBS "$DISCORD_LIBS -ldiscord-rpc"
+	append_var INCLUDES "$DISCORD_CFLAGS"
+fi
+define_in_config_if_yes "$_discord" 'USE_DISCORD'
+echo "$_discord"
 
 #
 # Enable vkeybd / event recorder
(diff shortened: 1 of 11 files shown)


backends/discord/discord.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#define FORBIDDEN_SYMBOL_EXCEPTION_time_h
+#include "backends/discord/discord.h"
+#include "common/encoding.h"
+#include "common/translation.h"
+
+#include <discord_rpc.h>
+#include <time.h>
+
+#define DISCORD_CLIENT_ID "714287866464698470"
```
sev-, line 33: Where does this come from? Is it kind of secret? If so, we probably could regenerate it and put at compile time, not from the source.

backends/discord/discord.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+Discord::Discord() {
+	DiscordEventHandlers handlers;
+	memset(&handlers, 0, sizeof(handlers));
```
sev-, line 37: Is it safe to initialize like this? Same question regarding other memsets down the code.

configure
```diff
@@ -5583,6 +5594,25 @@ if test "$_pandocext" = "default"; then
...
+# Check for Discord
+#
+echocheck "Discord RPC"
+if test "$_discord" = auto ; then
+	_discord=no
+	cat > $TMPC << EOF
+#include <discord_rpc.h>
+int main(void) { Discord_Shutdown(); return 0; }
+EOF
+	cc_check $DISCORD_CFLAGS $DISCORD_LIBS -ldiscord-rpc && _discord=yes
```
sev-, line 5607: Do you require some specific minimal version of the library?

backends/discord/discord.cpp
```diff
@@ -0,0 +1,72 @@
+/* ScummVM - Graphic Adventure Engine
```
sev-, line 1: Could we call it probably not "backends/discord" but "backends/presence" and the Discord. The current API looks more or less generic, so it could be potentially extended with other things like Facebook status 😉

lotharsm:
I wonder if we can (and should) include the discord-rpc sources to our own sources - otherwise, we have yet another dependency to take care of. :)

Mataniko:
i recommend this approach by Dolphin: https://github.com/dolphin-emu/dolphin/tree/master/Externals/discord-rpc

criezy:
> In my testing, game icons look better than screenshots, but most of the links in this old thread are broken: https://forums.scummvm.org/viewtopic.php?f=1&t=4127
> Plus I don't know if it's kosher to use community resources as official.

If needed I have a local backup of most, if not all, the icons from that thread (I have 166 icons from that thread).But indeed I think we would need agreement from the authors to use them for the discord integration.

Mataniko:
Also mostly found here:
https://s437.photobucket.com/user/MapleySyrup/library/?sort=3&page=1

From https://web.archive.org/web/20160816091729/http://www.megamonkey.org/icons/:
"These icons were created for use with the 2D point-point-and-click adventure game virtual machine ScummVM, as well as ScummVM's sister projects ResidualVM (for 3D adventure games) and the gargoyle engine from ScummVM-Misc (for text adventures)."

I think this falls under the intent of "for use with ScummVM" and would not require explicit permission.

SupSuper:
@sev- 
> Nice, very nice.
> 
> We still need to clarify the usage of the icons. Perhaps ask on the icons thread and send PMs to the authors?

I PMed the icon authors, but only got a few responses (they were positive though). I suspect nobody in that thread visits the forum anymore, and PM notifications aren't enabled by default.
Maybe we could set up a repository similar to scummvm-media for people to contribute?

@lotharsm 
> I wonder if we can (and should) include the discord-rpc sources to our own sources - otherwise, we have yet another dependency to take care of. :)

There's a few obstacles to that: dependencies, lots of platform-specific code, and our worst enemy: C++11. :)
The library isn't getting more updates and Discord is only supported on desktop, so I don't think it'll be that annoying to maintain. But we could also try to strip it down to only the feature we want.

Mataniko:
@SupSuper Status check on this

SupSuper:
@Mataniko We need game icons. I got a few positive replies from the forum thread, the question is how do we organize this to make it easy to contribute and maintain going forward? Create a repository for them?

While we can't automate adding icons to Discord, it could also serve as a basis for an official ScummVM icon pack.

sev-:
@SupSuper the icons should go into a separate repo.

I think it will be a fair use of the IP and we still could host it.

Could you please rebase your PR so it is mergeable again?

backends/discord/discord.cpp
SupSuper: It's just a pointer struct so it should be fine. They do the same in their code: https://github.com/discord/discord-rpc/blob/master/examples/send-presence/send-presence.c

backends/discord/discord.cpp
SupSuper: Will do!

backends/discord/discord.cpp
SupSuper: This comes from the Discord Developer Portal, so it can't be regenerated. Is there a better place to put it?

Here's an example from another OSS: https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/UICommon/DiscordPresence.cpp#L195

configure
SupSuper: No.

SupSuper:
@sev- Done.

sev-:
Thanks! merging.

PR #2281, opened 2020-05-27
KEYMAPPER: Allow mapped actions without modifier keys

OMGPizzaGuy:
KEYMAPPER: Allow mapped actions without modifier keys when no mapped actions exist for current modifier keys

While experimenting with keymaps for Ultima 8 I found that pressing unmapped modifier keys (shift or control) while holding a mapped key would prevent the release event from occurring for the mapped key if released first. The result for the movement controls could be that the movement gets stuck on and player continues moving unexpectedly until the key is press and released again with no modifier

Additionally, I found it desirable to always use the key mapping without the modifier if there was no exact match for the key with a modifier. I am considering mapping left shift to run and would like to be able to press it before or during movement.

I'm currently unaware how this might effect other engines and this may be an incomplete concept as it could try other mappings if multiple modifiers are in use. I thought this would be a good venue for discussion.

### DIFF
--- a/backends/keymapper/keymap.cpp
+++ b/backends/keymapper/keymap.cpp
@@ -124,74 +124,108 @@ const Action *Keymap::findAction(const char *id) const {
 	return nullptr;
 }
 
-Keymap::ActionArray Keymap::getMappedActions(const Event &event) const {
+Keymap::KeymapMatch Keymap::getMappedActions(const Event &event, ActionArray &actions) const {
 	switch (event.type) {
 	case EVENT_KEYDOWN:
 	case EVENT_KEYUP: {
 		KeyState normalizedKeystate = KeyboardHardwareInputSet::normalizeKeyState(event.kbd);
 		HardwareInput hardwareInput = HardwareInput::createKeyboard("", normalizedKeystate, "");
-		return _hwActionMap[hardwareInput];
+		actions.push_back(_hwActionMap[hardwareInput]);
+		if (!actions.empty()) {
+			return kKeymapMatchExact;
+		}
+
+		if (normalizedKeystate.flags & KBD_NON_STICKY) {
+			// If no matching actions and non-sticky keyboard modifiers are down,
+			// check again for matches without the exact keyboard modifiers
+			for (HardwareActionMap::const_iterator itInput = _hwActionMap.begin(); itInput != _hwActionMap.end(); itInput++) {
+				if (itInput->_key.type == kHardwareInputTypeKeyboard && itInput->_key.key.keycode == normalizedKeystate.keycode) {
+					int flags = itInput->_key.key.flags;
+					if (flags & KBD_NON_STICKY && (flags & normalizedKeystate.keycode) == flags) {
+						actions.push_back(itInput->_value);
+						return kKeymapMatchPartial;
+					}
+				}
+			}
+
+			// Lastly check again for matches no non-sticky keyboard modifiers
+			normalizedKeystate.flags &= ~KBD_NON_STICKY;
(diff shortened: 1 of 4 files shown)


bgK:
Hi, thanks for your contribution!

> While experimenting with keymaps for Ultima 8 I found that pressing unmapped modifier keys (shift or control) while holding a mapped key would prevent the release event from occurring for the mapped key if released first. The result for the movement controls could be that the movement gets stuck on and player continues moving unexpectedly until the key is press and released again with no modifier

Yes indeed, this is a problem. As a matter of fact it's part of a larger problem that there is currently no guarantee the keymapper will emit paired (START, END) events. Your proposed fix if fine for the case you mention but will be insufficient for some others:
1. The user presses `SHIFT+A`, an action is found. A `XXX_START` event is emitted.
2. The user releases `SHIFT` but keeps pressing `A`.
3. The user releases `A`.
An event `XXX_END` would need to be emitted at step 2 or 3. Not sure which one would be best, but currently no matching end event is produced. The idea I had in mind was to keep a list of currently started actions, and to somehow check if end events need to be emitted when events are received.

I am fine with merging your fix as is as it resolves some issues, though perhaps the limitations should be stated in a comment.

> Additionally, I found it desirable to always use the key mapping without the modifier if there was no exact match for the key with a modifier. I am considering mapping left shift to run and would like to be able to press it before or during movement.

That sounds like a positive change. If the behavior is discovered to be undesirable in some cases, it'd be easy to add a flag to opt out either at the keymap or at the action level.

OMGPizzaGuy:
Thinking about this further and I believe this proposed change could break the global keymap.
I'll check this later and consider some adjustments.

OMGPizzaGuy:
For the current handling the keymapper needs to process an exact match before attempting alternate matches in order to properly handle the global keymap.

Another option to handle alternate matches could be returning a match priority with the mapped actions so the combination of SHIFT+A could still be returned lower priority if CTRL+SHIFT+A is being held. With that an exact match priority could be checked before asking the global keymap, and if no exact match priority is found there, try again with lower priorities on each set until a match is found.

OMGPizzaGuy:
Latest attempt at this enhancement will handle a subset of held modifier keys. Now mapped combination of SHIFT+A could still be returned with the partial match status if CTRL+SHIFT+A is being held and would still take priority of a mapped A key.

I think I'm happy with the current version. Let me know what changes you would like to see.

backends/keymapper/keymap.cpp
```diff
@@ -124,74 +124,108 @@ const Action *Keymap::findAction(const char *id) const {
...
+			return kKeymapMatchExact;
+		}
+
+		if (normalizedKeystate.flags & KBD_NON_STICKY) {
+			// If no matching actions and non-sticky keyboard modifiers are down,
+			// check again for matches without the exact keyboard modifiers
+			for (HardwareActionMap::const_iterator itInput = _hwActionMap.begin(); itInput != _hwActionMap.end(); ++itInput) {
+				if (itInput->_key.type == kHardwareInputTypeKeyboard && itInput->_key.key.keycode == normalizedKeystate.keycode) {
+					int flags = itInput->_key.key.flags;
+					if (flags & KBD_NON_STICKY && (flags & normalizedKeystate.keycode) == flags) {
```
bgK, line 144: Doing a bitwise and between the flags and the keycode does not seem correct.

OMGPizzaGuy: Yeah, that does seem wrong. Good eye.

bgK:
Thanks! Merged as b4083476d5a8645e781052e68918a32a2e821dbf (with a minor formatting fix).

PR #2283, opened 2020-05-27
BASE: Fix and enable INITIALIZIER_LIST1 c++11 test

criezy:
I am creating a pull request for this so that a second pair of eyes can take a look before this is pushed (I am not very confident with my c++11 skills since we are not allowed to use it yet where I work).

### DIFF
--- a/base/test_new_standards.cpp
+++ b/base/test_new_standards.cpp
@@ -43,15 +43,14 @@
 //
 // We're not testing `nullptr` and `override`, since they're defined in common/c++11-compat.h
 
-// INITIALIZIER_LIST1 test disabled:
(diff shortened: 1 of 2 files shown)


criezy:
Following some discussion on IRC/Discord today I have added a second branch to the INITIALIZER_LIST1 test with a replacement to std:: initializer_list.

With this addition all the tests now pass with my macOS i386 toolchain targeting macOS 10.6. Without it the INITIALIZER_LIST1 test fails because while the compiler is c++11 compliant, the standard library doesn't have std::initializer_list (and I am not the first one to find that out, see for example https://stackoverflow.com/questions/42675632/emulating-braced-list-initialization-before-c11-stl ).

PR #2285, opened 2020-05-28
MIDI/ULTIMA8: Add XMIDI sequence branch indexes

NMIError:
This adds support for the "sequence branch index" feature of the XMIDI format. 
These "branches" (positions in the MIDI bytestream) are defined in the track 
header. I've added two generic functions to the MidiParser interface to set the 
current parsing position to one of these positions, or to check for the 
existence of a position with a certain number.

Additionally, I've added an option to the MIDI parser to prevent the parser 
from starting playback immediately after parsing, so that a starting position 
(or a track) can be selected first.

Ultima 8 uses sequence branches to mark different playback starting points in 
the music files. I've added support for starting playback at these points.

### DIFF
--- a/audio/midiparser.h
+++ b/audio/midiparser.h
@@ -281,7 +281,7 @@ protected:
 	bool   _centerPitchWheelOnUnload;  ///< Center the pitch wheels when unloading a song
 	bool   _sendSustainOffOnNotesOff;   ///< Send a sustain off on a notes off event, stopping hanging notes
 	bool   _disableAllNotesOffMidiEvents;   ///< Don't send All Notes Off MIDI messages
-	bool   _doNotAutoStartPlayback;  ///< Do not automatically start playback after parsing MIDI data or setting the track
+	bool   _disableAutoStartPlayback;  ///< Do not automatically start playback after parsing MIDI data or setting the track
 	byte  *_tracks[MAXIMUM_TRACKS];    ///< Multi-track MIDI formats are supported, up to 120 tracks.
 	byte   _numTracks;     ///< Count of total tracks for multi-track MIDI formats. 1 for single-track formats.
 	byte   _activeTrack;   ///< Keeps track of the currently active track, in multi-track formats.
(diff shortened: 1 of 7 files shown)


mduggan:
This looks great to me, and it seems to be totally backwards-compatible for other engines, but to be honest I'm not super familiar with the midi code so I'll leave it open a couple more days to see if anyone else has comments before merging.

OMGPizzaGuy:
Ran this locally and it works well. 

Starting a combat run consistently gives the following warnings but continues to operate without issue:
WARNING: MidiParser-XMIDI: jumpToIndex called with invalid sequence branch index 0!
WARNING: play() invalid branch index 0!

NMIError:
Thank you for your feedback.

I could not reproduce the issue, but I did see that the code did not properly handle tracks without sequence branches, resulting in the warnings you saw. It did fall back to playing the track from the beginning, so it did work as intended.

I've added an explicit check to see if the track does not have any branches, and then start from the beginning of the track. This should get rid of the warnings.

OMGPizzaGuy:
Makes sense and fixes the issue. Thanks!

audio/midiparser.h
```diff
@@ -377,7 +381,14 @@ class MidiParser {
...
 		  */
-		 mpDisableAllNotesOffMidiEvents = 6
+		 mpDisableAllNotesOffMidiEvents = 6,
+
+		 /**
+		  * Does not automatically start playback after parsing MIDI data
+		  * or setting the track. Use startPlaying to start playback.
+		  * Note that not every parser implementation might support this.
+		  */
+		  mpDoNotAutoStartPlayback = 7
```
OMGPizzaGuy, line 391: How would you feel about mpDisableAutoStartPlayback name for consistency? Too much of a nitpick?

NMIError: Yes, that's a better name.

OMGPizzaGuy:
Well, this PR has been open for a few weeks and looks like it shouldn't have effects outside engines that plan to use the feature. Merging!

NMIError:
Thanks!

PR #2287, opened 2020-05-29
MIDI: Fix XMIDI hanging notes

NMIError:
XMIDI uses Note On events with a length which specifies when a corresponding
Note Off event should be generated. Some XMIDI data contains notes with length
0, which are played very briefly by AIL. However, the ScummVM MidiParser will
treat Note On events with length 0 as "active notes", meaning it will wait for
a Note Off event before turning off the note. The Note Off will never come and
the note will hang.

This is particularly noticeable in Ultima 8 (first section of the track that
plays at the beginning of the game, at the shore) and The 7th Guest (the track
that plays while solving the cake puzzle).

I've fixed this by changing the length from 0 to 1. This will cause the notes
to be active for the minimum length.

### DIFF
--- a/audio/midiparser_xmidi.cpp
+++ b/audio/midiparser_xmidi.cpp
@@ -120,6 +120,14 @@ void MidiParser_XMIDI::parseNextEvent(EventInfo &info) {
 		info.basic.param1 = *(_position._playPos++);
 		info.basic.param2 = *(_position._playPos++);
 		info.length = readVLQ(_position._playPos);
+		if (info.length == 0) {
+			// Notes with length 0 are played with a very short duration by the AIL driver.
+			// However, the MidiParser will treat notes with length 0 as "active notes"; i.e.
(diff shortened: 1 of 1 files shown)


mduggan:
Thanks for fixing this!

OMGPizzaGuy:
Oh merged! Quick question though - I was reading / experimenting with this and wanted to make sure it wasn't a virtual note off.
I'm very new to audio and not sure I understand if length is the same a velocity or if that is what info.basic.param2 is in this case.

Here the resource I was able to find on the subject: http://www.palserv.com/XMidi/docs/UsersGuide.html#NOTE_OFF

NMIError:
Param2 is velocity i.e. the strength with which the note is played (usually maps to volume). Length is used by XMIDI to specify the duration of the note. So there are no explicit Note Off events; they must be generated by the player after the duration of a note has expired.

OMGPizzaGuy:
Excellent. Thanks for clearing that up for me!

PR #2288, opened 2020-05-30
AUDIO: Add ability for engine to supply its own soundfont data.

yuv422:
* Add ability for engine to supply its own soundfont data.
* Add MDT_PREFER_FLUID to allow engines to indicate they prefer the fluidsynth driver.

The fluidsynth driver will only be given preference if the midi driver is set to auto.

This PR allows engines to specify their own soundFont data. This will be used by Blazing Dragons and Tinsel to support PSX SEQ midi music.

I will work on a midiPlayer to extracts the soundFont data in another PR.

I added the following new virtual methods to the midi driver class.
```
// Allow an engine to supply its own soundFont data. This stream will be destroyed after use.
virtual void setEngineSoundFont(Common::SeekableReadStream *soundFontData) { }

// Does this driver accept soundFont data?
virtual bool acceptsSoundFontData() { return false; }
```

FluidSynth has a weird way of accepting in memory soundfont data. I followed the example given here. http://www.fluidsynth.org/api/fluidsynth_sfload_mem_8c-example.html

### DIFF
--- a/audio/softsynth/fluidsynth.cpp
+++ b/audio/softsynth/fluidsynth.cpp
@@ -82,7 +82,7 @@ public:
 // MidiDriver method implementations
 
 MidiDriver_FluidSynth::MidiDriver_FluidSynth(Audio::Mixer *mixer)
-	: MidiDriver_Emulated(mixer), _engineSoundFontData(NULL) {
+	: MidiDriver_Emulated(mixer), _engineSoundFontData(nullptr) {
 
 	for (int i = 0; i < ARRAYSIZE(_midiChannels); i++) {
 		_midiChannels[i].init(this, i);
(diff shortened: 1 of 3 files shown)


audio/softsynth/fluidsynth.cpp
```diff
@@ -119,13 +124,49 @@ void MidiDriver_FluidSynth::setStr(const char *name, const char *val) {
 	free(val2);
 }
 
+// Soundfont memory loader callback functions.
+
+static void *SoundFontMemLoader_open(const char *filename)
```
sev-, line 129: It would be nice to follow our code formatting guidelines in this whole added block

yuv422: Whoops I missed that. Done.

yuv422:
Is everyone happy for me to merge this?

yuv422:
ok I'll merge this now.

PR #2289, opened 2020-05-30
WINTERMUTE: Add Alavi Animator

aviloria:


aviloria:
It seems that "Alavi Animator" is the beta version of "Informer Alavi" acording to the webpage of the developers:
http://web.archive.org/web/20140724142640/http://alavi.aanimator.com/html/download_try.php

PR #2291, opened 2020-05-30
COMMON: Fix build when uint32_t and uint32 are not defined the same

lephilousophe:
FriBidiChar is mapped to uint32_t (from stdint.h) while U32String is
composed of uint32 (from scummsys.h/config.h). On some platforms (like in 3DS) those
two types aren't defined using the same base type (long int vs int) and
compilation errors occur.


lephilousophe:
As it's a simple change, I will merge this tomorrow if nobody is against it.

lephilousophe:
Merging now

PR #2293, opened 2020-05-31
SCI: Allow channel 9 on all matching audio drivers

sluicebox:
ScummVM doesn't play several Adlib sounds in The Colonel's Bequest when using the Adlib driver: https://bugs.scummvm.org/ticket/10734 

The current logic in SoundResource::getChannelFilterMask() is intended to always play channel 9 if a driver has a rhythm channel, even if the driver's play-flag isn't set in the sound resource. The way it's written though, it also bans channel 9 from ever playing if a driver doesn't have rhythm channel, such as Adlib, which is incorrect.

This PR adds the normal play-flag check to channel 9 when the driver doesn't have a rhythm channel. (This switch hasn't been touched in 10 years!)

For example, here are the channels in sound 48, when knocking on the door in room 14:
channel: 01 voices: 00 flags: 30
channel: 09 voices: 82 flags: 0F [ Adlib flag 04 ]

Other sounds in the same situation:
Turning the crank in room 52
Turning the valve in room 65
Prying open a crypt vault in room 57
Skull falling out of a crypt vault in room 57

ADL.DRV does not filter out channel 9, which I've confirmed in all the other SCI0 Late ADL.DRVs I have.

### DIFF
--- a/engines/sci/resource_audio.cpp
+++ b/engines/sci/resource_audio.cpp
@@ -968,7 +968,7 @@ int SoundResource::getChannelFilterMask(int hardwareMask, bool wantsRhythm) {
 			break;
 		case 9:
 			// Play rhythm channel when requested
-			play = wantsRhythm;
+			play = wantsRhythm || (flags & hardwareMask);
 			break;
 		default:
 			// Otherwise check for flag


waltervn:
I needed this exact same thing for the Casio driver (which hasn't been merged yet): https://github.com/waltervn/scummvm/commit/057814ff0d08d88cd346d4c3954ec3d0ba632075

waltervn:
Alternatively, with the initTrack mechanism we currently have, an alternative fix might be to always request the rhythm channel for SCI0 within the adlib driver? I'm not 100% sure about this though, maybe @athrxx can comment on that.

athrxx:
Well, fixing the issues of the wrong parsing / channel filtering code with hacks in the affected drivers (probably only those that actually have that many voices) doesn't strike me as the most intuitive method of doing this. 

I'd rather go with sluicebox's solution. Or really limit the special treatment of channel 9 to the Midi devices that use it.  I can't say whether sluicebox's change breaks anything. Why not add a comment to the line so that it is easy to spot if something goes wrong.

If another possible solution were to have the drivers always return`true `on calls to `hasRhythmChannel() ` , then why not get rid of that function altogether. By returning `true` we would basically make the drivers tell a lie just to appease the broken parsing / channel code. This sort of thing will make the code even less comprehensible for someone not directly involved.

I am not the author of the AdLib driver. Why is there a method `hasRhythmChannel()` that always returns false and a `useRhythmChannel()` which returns different things depending on the patch format / SCI version (how can it "use" a rhythm channel it doesn't "have")? Maybe just a naming issue...

sluicebox:
@waltervn you wrote this *exact same commit* 5 years ago?? Um, would you describe yourself as a patient person? =)

I can't speak to the hasRhythmChannel() method's accuracy or purpose, though I had the same questions about it.

I think this change is the best option, having tried some of the alternatives first, because what's there now is only correct if Sierra's drivers contain code that's hard-coded to reject channel 9, which ADL.DRV does not, and I would be a little surprised if any of them do. If there are drivers *do* explicitly block 9 then we'd want to update the driver interface and/or return values to reflect that. My bet is that blocking channel 9 on !hasRhythmChannel() is an unintentional side effect of the way this switch was written, and that the author's intent was just to override the play-flag and not completely ignore it.

I can expand the existing comment but I think the line that follows in the next default block makes everything clear, it's just that the github preview stops right before that line. I think more words get in the way on this one and it's the current behavior that doesn't match the comment and needed more justifying text.

For what it's worth, I only detected two pieces of play-flag filtering in the SCI0 Late ADL.DRV, and they are only when a channel has voices. The LoadSound export tests the play-flag when there are voices, just we we do in initTrack(), and the function that's equivalent to our voiceMapping() tests the play-flag again and immediately returns if it's not set. We don't do that but because we filter all channels by play-flag before they reach the driver we don't need to.

This is my first time cracking open a Sierra driver so factor that in when evaluating my boisterous claims.

waltervn:
> @waltervn you wrote this _exact same commit_ 5 years ago?? Um, would you describe yourself as a patient person? =)

Heh. I have way too much old code still collecting dust in my repo. I'm trying to address some of that now, starting with the Amiga and Mac sound drivers.

> I can't speak to the hasRhythmChannel() method's accuracy or purpose, though I had the same questions about it.

I haven't verified this, but I think the most logical explanation for this function would be that there are drivers (and I'm thinking e.g. MT-32) that play the rhythm channel even when the play flag is clear.

sluicebox:
Since the SCI audio brain-trust has weighed in and @waltervn came up with the same fix and I've been testing with with this for a week, I'm now comfortable merging this.

I scanned more games for sounds that this behavior has suppressed and found them in almost every one. For example, sound 60 in SQ3, which can be triggered by "search closet" in the room 90, doesn't play in ScummVM when emulating Casio or Tandy or PC Speaker, but does now with this fix:

```
channel: 09 flags: 38 [08: Casio, 10: Tandy, 20: PC Speaker]
channel: 11 flags: 06 [02: Yamaha, 04: Adlib]
channel: 12 flags: 06 [02: Yamaha, 04: Adlib]
channel: 13 flags: 06 [02: Yamaha, 04: Adlib]
channel: 14 flags: 06 [02: Yamaha, 04: Adlib] 
```

PR #2295, opened 2020-05-31
CONFIGURE: Undefine __STRICT_ANSI__ for newlib based platforms

lephilousophe:
This allows backends to use non-ANSI functions.
As it was the main reason to not set -ansi on these platforms, add it
back to ensure strict adherence to C++ standard.

This commit makes compilation work when using --enable-c++11 on tested platforms.
Adding -ansi is where the risk is and we could add -U__STRICT_ANSI__ only when enabling C++11 on newlib platforms.
Though, I think using -ansi everywhere would be nice to have.

:warning: This commit has only be tested on some platforms impacted by this change and some are still missing.
Most tests have been done using [dockerized-bb](https://github.com/lephilousophe/dockerized-bb).
- [x] 3ds
- [x] amigaos (thanks to @raziel-)
- [x] android
- [ ] androidsdl (though I don't expect any error on it)
- [x] dreamcast (thanks to @zeldin)
- [x] ds
- [x] gamecube
- [x] mingw64
- [ ] mint
- [ ] n64
- [ ] ps2
- [x] ps3
- [x] psp
- [x] psp2
- [x] switch
- [ ] tizen
- [x] wii

For DS, PR #2294 is needed to compile without C++11 mode.

@sev- I don't really know how to ask to porters to test this change, hence I request your review :)

### DIFF
--- a/configure
+++ b/configure
@@ -1786,6 +1786,8 @@ switch)
 	datarootdir='${prefix}/data'
 	datadir='${datarootdir}'
 	docdir='${prefix}/doc'
+	# Switch SDK has C++11 constructs so we must enable it
+	_use_cxx11=yes
 	;;
 tizen)
  	_host_os=tizen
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -2251,16 +2253,17 @@ if test "$have_gcc" = yes ; then
...
-				;;
-			*)
-				append_var CXXFLAGS "-ansi"
-				;;
-			esac
+			append_var CXXFLAGS "-ansi"
 		fi
+		case $_host_os in
+		# newlib-based system include files suppress non-C89 function
+		# declarations under __STRICT_ANSI__, don't undefine it
```
zeldin, line 2260: I think this comment contains an accidental double negative...  :smiley_cat:

lephilousophe: oops! :smile:

raziel-:
Checked and working with AmigaOS4.

Compilation error mentioned in the c++11 wiki is fixed by this.
https://wiki.scummvm.org/index.php?title=Compiling_ScummVM/C%2B%2B11_Readiness#cite_note-3

raziel-:
Forgot to add.
This PR also fixes item #6988
https://bugs.scummvm.org/ticket/6988

zeldin:
The fix is ok for dreamcast too.  (So that leaves 5 platforms to tick off. :smile: )

lephilousophe:
Thanks for the testing, I modified my initial comment to take your results into account.

sev-:
Okay, let's give it a try and see if build bot blows up

lephilousophe:
It seems buildbot only blew up for PS2. I am not really sure why as it works well for others.
I suspect gcc 3.2.3 to not take -U__STRICT_ANSI__ into account when passing -ansi but I will need further tests.
As gcc 3.2.3 is really too old maybe we should do a special case for it? It won't support anything from C++11 anyway.
@sev-: what do you think?

zeldin:
I guess one solution could be to move the block that adds `-ansi` from the gcc version >= 3 scope into the gcc version >= 4.3 scope instead?

lephilousophe:
After checking a PS2 compiler, it seems that adding -ansi, makes g++ call the preprocessor with -D__STRICT_ANSI__ after our -U so we can't do anything in command line.
Anyway, @sev- did fix the bugs so ps2 now compiles fine. I think it could even compile without -U__STRICT_ANSI__.

PR #2298, opened 2020-05-31
COMMON: Use assert for CLIP() if bounds are not properly ordered

antoniou79:
If the boundary arguments of CLIP function are not ordered as expected (the first should the min of the two), then the function may return an invalid result. This is more likely to happen if the boundary arguments are variables rather than literals.

### DIFF
--- a/common/util.h
+++ b/common/util.h
@@ -49,7 +49,18 @@ template<typename T> inline T ABS(T x)		{ return (x >= 0) ? x : -x; }
 template<typename T> inline T MIN(T a, T b)	{ return (a < b) ? a : b; }
 template<typename T> inline T MAX(T a, T b)	{ return (a > b) ? a : b; }
 template<typename T> inline T CLIP(T v, T amin, T amax)
-		{ if (amin > amax) {T tmp = amin; amin = amax; amax = tmp;} if (v < amin) return amin; else if (v > amax) return amax; else return v; }
+	{
+#if !defined(RELEASE_BUILD)
+		// debug builds use this assert to pinpoint
+		// any problematic cases, where amin and amax
+		// are incorrectly ordered
+		// and thus CLIP() would return an invalid result
+		assert(amin <= amax);
+#endif
+		if (v < amin) return amin;
+		else if (v > amax) return amax;
+		return v;
+	}
 
 /**
  * Template method which swaps the values of its two parameters.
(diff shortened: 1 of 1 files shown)


lskovlun:
I had a similar problem with a boundary case in Common::Rect::extend that broke something in the SCI engine. I didn't fix it; I was afraid I'd have to go and re-test every supported game from start to finish.

antoniou79:
> I had a similar problem with a boundary case in Common::Rect::extend that broke something in the SCI engine. I didn't fix it; I was afraid I'd have to go and re-test every supported game from start to finish.

I understand; I still think that the change in this commit will not be causing any side effects that every engine using it will have to be retested. It's in line with the purpose of CLIP and makes sure that an invalid value won't be returned.

lephilousophe:
Shouldn't CLIP the simplest possible to let it be fast?
I can't find an implementation in real life which takes care of inverting bounds.
libstdc++ (for std::clamp) seems to assert when bounds are inverted. I think it is a more sane choice.
This check should be enabled when debugging only to let us spot code errors but disabled for release to be faster.

antoniou79:
> Shouldn't CLIP the simplest possible to let it be fast?
> I can't find an implementation in real life which takes care of inverting bounds.
> libstdc++ (for std::clamp) seems to assert when bounds are inverted. I think it is a more sane choice.
> This check should be enabled when debugging only to let us spot code errors but disabled for release to be faster.

In principle I agree with this, there will be a small speed impact. This is the reason this was submitted as a PR.

So do we have a defined macro specifically for debug builds? If I just add an assert will it eval to nop in release builds by default? From a quick google search it seems that this is not guaranteed, some macro has to be defined as well ie. NDEBUG ?

lephilousophe:
we can put a `#ifdef DEBUG` to ensure it is not run in release?
Something like that:
```
#ifdef DEBUG
assert(min <= max)
#endif
```

antoniou79:
Based on our current configure script the macro that is used to distinguish a release build (that is a build configured with **---enable-release**) is "**RELEASE_BUILD**". 

The DEBUG macro does not work (not for local builds anyway, I haven't checked our buildbot scripts). It is not automatically set nor is it set with our configure script.

lephilousophe:
Seems great to me. That will avoid unexpected behaviors.

antoniou79:
Great! If there are no objections or further feedback I'll probably be merging this in the weekend or early next week.

PR #2299, opened 2020-05-31
New engine - KINGDOM

Strangerke:
This engine was mostly written 3 years ago, and went to hibernation because I felt unable to write a video player for the MVE format used, despite TMM wrote a player for ffmpeg.
Some days ago, madmoose started writing a player, and miraculously finished it in a matter of days. The player is even in better shape than the one in ffmpeg, which underwent serious regressions those past 3 years.

It's with a real joy that I'm starting this PR. This engine targets the game "Kingdom: The Far Reaches". The full DOS version is supported, as well as the DOS Demo. 

This game is a FMV game, with an interface and an inventory, making it almost a standard point'n'click game.

Your comments are more than welcome. 
Have fun!


dreammaster:
Right off the bat, it seems like the pull request needs some cleanup. In particular
1) It doesn't seem like it was correctly rebased. Lots of non-Kingdom stuff there
2) Lots of the commits don't have an engine prefix, they just have names like "Small compilation error fix"
3) May be related to point 1, but consider whether anything remaining for backends, keymapper, or audio should be part of the pull request or not

sev-:
Rebase issues addressed, git log messages amended.

engines/kingdom/GPLogic1.cpp
```diff
@@ -0,0 +1,1501 @@
+/* ScummVM - Graphic Adventure Engine
```
dreammaster, line 1: It's more standard to underscores in filenames rather than camel-case, though I'm not sure if this is fine with regard to the project's coding conventions

engines/kingdom/GPLogic1.cpp
```diff
@@ -0,0 +1,1501 @@
...
+		_vm->processMap(5, _vm->_zoom);
+		_vm->_userInput = 0;
+		break;
+	case 0x43E:
+		_vm->_sound = _vm->_lastSound;
+		_vm->playMovie(_vm->_pMovie);
+		break;
+	default:
+		if (_vm->_userInput)
+			debug("Skipped UserInput %d(0x%04X) for _StatPlay %d", _vm->_userInput, _vm->_userInput, _statPlay);
```
dreammaster, line 296: For all these warning messages you might want to consider using warning rather than debug. That'll give more clarity that the message is a warning

engines/kingdom/console.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "gui/debugger.h"
+
+#include "kingdom/console.h"
+#include "kingdom/kingdom.h"
+
+namespace Kingdom {
+
+Console::Console(Kingdom::KingdomGame *vm) : _vm(vm) {
+	registerCmd("showHotspots",	WRAP_METHOD(Console, Cmd_showHotspots));
```
dreammaster, line 31: _allowSkip isn't being initialized in constructor, and isn't used by the console. Remove if it's unused

engines/kingdom/constants.cpp
```diff
@@ -0,0 +1,5484 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+#include "kingdom/kingdom.h"
+
+namespace Kingdom {
+
+const char *_rezNames[] = {
```
dreammaster, line 27: There's enough constants that you should consider setting up a data file for them, though this could wait until post-merge

engines/kingdom/detection.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+	return desc != nullptr;
+}
+
+int KingdomMetaEngine::getMaximumSaveSlot() const {
+	return MAX_SAVES;
+}
+
+SaveStateList KingdomMetaEngine::listSaves(const char *target) const {
+	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
```
dreammaster, line 121: Given the more recent improvements for save files, you might want to consider taking advantage of it, and avoid having to worry about implementing list files, savegame meta, etc.

engines/kingdom/GPLogic1.cpp
Strangerke: done

engines/kingdom/console.cpp
Strangerke: Gone :)

engines/kingdom/constants.cpp
Strangerke: I'm currently reluctant to switch to a DAT file, as I don't manage to work on an engine after that. This engine only support one of the ports of the game, I guess it could be interesting to have a look at at least the MAC port which has a different interface and uses Quicktime (including to display the fix image between two videos). So, yes, I totally agree it would be a lot cleaner... but right now I prefer to delay a bit, just in case.

engines/kingdom/GPLogic1.cpp
Strangerke: you're right, I renamed to files to logic* (it wasn't able to notice a change of uppercase, maybe because it's not case sensitive under Windows?)

PR #2301, opened 2020-06-01
DIRECTOR: Move transitions to the score

npjg:
To facilitate later refactoring, transitions are moved out of the frame and into the score.


sev-:
Thanks.

PR #2303, opened 2020-06-02
ULTIMA8: Improve keyboard movement

OMGPizzaGuy:
Add movement state and refactor player movement to unify behavior between mouse and keyboard movement.

A few thoughts: 
1. Note that targetted jump now only occurs when not moving in a direction. I've considered renaming the AvatarMoverProcess::jump method to AvatarMoverProcess::targetJump and moving the settings check out of that method
2. I could potentially combine all the AvatarMoverProcess debugger command into one or two if we would like to use args.
4. The movement state could be a struct to separate it from the AvatarMoverProcess
5. ACTION_MOVE_RUN or ACTION_RUN? same for others
6. I made STEP (LCTRL) and Left mouse perform a jump as it seemed to make sense on a controller. This could replace the Jump action keymap -  although moving with keyboard plus space also works nicely.


OMGPizzaGuy:
BasicU8Mouse

Here's a flow chart for mouse movement I made years back.
I'm going through it now to check correct behavior - I'm noticing to jump or walk I need to hold the left and right mouse, while the document indicates a click should be enough to do it once - may need to confirm that again in original game

OMGPizzaGuy:
Combat run issue: start combat run, reduce mouse cursor to a step, let go -> Avatar will take another unneeded walk animation

OMGPizzaGuy:
I opted to correct the combat run behavior as it seems like the original game behavior was also a bug.
Attempts to change the click vs. hold movement appeared to not work well with combat kicks, so I think it would be out of scope for the PR. Instead I noted the original game behavior in a comment.

mduggan:
Awesome, this is much cleaner than the previous setup.  Sorry for the lack of response - I was deep in disassembling crusader. I'll play with it and see if I notice any issues, but otherwise I think it's good to merge.

I think having targeted jump only happen when not moving is fine, it fits with how it is used.

Separating movement state from the process might be nice, I still have to do some more work in Crusader to understand what it will need to be different.

PR #2304, opened 2020-06-03
DIRECTOR: LINGO: Implement kTheImmediate

djsrv:
This implements the sprite property `immediate`, which causes sprite scripts to run on mouseDown instead of mouseUp.


sev-:
Thanks.

PR #2305, opened 2020-06-03
DIRECTOR: LINGO: Fix loadScriptText

djsrv:
This fixes a faulty whitespace check in loadScriptText. Script text only needs to start with "--", no newline necessary, and the faulty check broke Simple Factory, which has a script starting with
```
--  Simple Factory
```

### DIFF
--- a/engines/director/lingo/lingo-preprocessor.cpp
+++ b/engines/director/lingo/lingo-preprocessor.cpp
@@ -165,6 +165,7 @@ Common::String Lingo::codePreprocessor(const char *s, ScriptType type, uint16 id
 	const char *lineStart, *prevEnd;
 	int iflevel = 0;
 	int linenumber = 1;
+	bool defFound = false;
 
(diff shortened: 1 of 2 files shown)


engines/director/lingo/lingo-preprocessor.cpp
```diff
@@ -191,6 +192,19 @@ Common::String Lingo::codePreprocessor(const char *s, ScriptType type, uint16 id
 			res1 = preprocessSound(res1);
 		}
 
+		if (type == kMovieScript && _vm->getVersion() <= 3 && !defFound) {
```
sev-, line 195: This should be the first thing to do, right after the line is acquired, e.g. move it right after the `debugC`. Otherwise, you could potentially do unnecessary work with other preprocessing steps.

djsrv: Fixed in #2306

PR #2307, opened 2020-06-04
MACGUI: don't activate menu if it is active

whiterandrek:
This pull request fixes activation of menu if it is active.

If menu is activated more than 1 time, it restores screen with menu drawn and moreover leads to error due to assigning to busy PauseToken, which leads to bugs in Pink Panther games .


sev-:
Thanks!

henke37:
Seems more like suppressing the bad behavior rather than locating the cause of it. In particular, the caller already checks if the menu is active:

```cpp
if (!_active)
	_wm->activateMenu();
```

PR #2310, opened 2020-06-06
MT32: Update munt emulation code to mt32emu 2.4.0

lotharsm:
This patch updates mt32emu used for our MT32 emulation to version 2.4.0. Additionally, I re-applied 369ba0c4b7f1278c415f7c7655cfa6cc5ae2b5fb and (partially) 21450facee606e8303fb06b94b9c2a3306254442 to silence some GCC and MSVC warnings.


sev-:
Nice, thanks for doing it!

PR #2311, opened 2020-06-06
DIRECTOR: LINGO: Implement factories

djsrv:
This implements factories, as well as primary event handlers, so that the Lingo in the D2 Simple Factory example is now fully working.


sev-:
Excellent. And you've addressed all my earlier notes.

PR #2312, opened 2020-06-07
KINGDOM: Fix DS compilation

lephilousophe:
Without specializing MIN compiler gets confused because the two arguments are of different types: uint32 and uint.
delayMillis is using an uint as argument so fixing the type of delay in the same time.


lephilousophe:
Already fixed by @sev-

PR #2313, opened 2020-06-07
I18N: Fixed typo in wintermute/POTFILES. Updated translations.dat

yuv422:
Fixed typo in Wintermute POTFILE. It was missing the folders. The file also needed a carriage return at the end.

I haven't updated the translations.dat before so I've make a PR just to check I'm following the right process.

### DIFF
--- a/engines/wintermute/POTFILES
+++ b/engines/wintermute/POTFILES
@@ -1,3 +1,3 @@
 engines/wintermute/detection.cpp
 engines/wintermute/wintermute.cpp
-keymapper_tables.h
+engines/wintermute/keymapper_tables.h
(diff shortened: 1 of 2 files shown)


lotharsm:
Thanks for your contribution!

One sidenote related to the translations.dat: Updating the translations.dat at this point makes no real difference, since it only contains the translated content. No matter if the data is missing from translations.dat only or from the entire translations system, in this case, the strings will always default to the original english strings.

Modifying just the .pot file is enough at this point, since our weblate instance (aka. our translations system) will pick up the changes in the .pot file and do most of the remaining magic.

yuv422:
ah k cool. I'll remove the translations.dat update from this PR then.

lotharsm:
Thanks, merging!

PR #2320, opened 2020-06-10
DIRECTOR: Move toward widget-based rendering

npjg:
This PR implements many rich text features in MacEditableText and uses the Mac GUI framework to render all cast types except QuickDraw shapes.


thebpc:
Retry DeepCode

PR #2321, opened 2020-06-11
SCUMM: do not reverse strings in HE

BLooperZ:
Apparently, my previous PR #1989 for supporting fan-translated hebrew version of SCUMM games broke official translation of HE games (HE version 73 and 99).
It made the games crash with `handleNextCharsetCode: invalid code ...` (even with subtitles off)
I have added a condition so HE games are not affected by the reversal of strings.
Now HE games are working again, sorry.


sev-:
Thanks

PR #2322, opened 2020-06-12
TESTBED: Bugfix & Achievements

lolbot-iichan:
1. Minor refactoring to have a single function to create testsuites list both at TestbedEngine() constructor and metaengine's getAchievementsInfo() method. Some code was moved away from GFXTestSuite() constructor to avoid running it inside getAchievementsInfo() from main menu.

2. Fixed a bug that chashed TestBed after Return-To-Launcher

3. Give achievement for each testsuite and a secret achievement for completing them all.

4. Added some files required for FS testsuite to pass (test1/file.txt, etc)

### DIFF
--- a/dists/engine-data/testbed-audiocd-files/TEST3/FILE.txt
+++ b/dists/engine-data/testbed-audiocd-files/TEST3/FILE.txt
@@ -1,0 +1,1 @@
+It works!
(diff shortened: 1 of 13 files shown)


sev-:
Thanks!

PR #2327, opened 2020-06-14
WINTERMUTE: Add save thumbnails to Corrosion

lolbot-iichan:
This implements enhancement #11032 WME: Add save thumbnails to Corrosion: https://bugs.scummvm.org/ticket/11032

1. Move some code to separate methods to make it accessible from other code points.
2. Clean cached thumbnail on game load (I believe it's original WME engine bug that this thumbnail may be not cleared after game load)
3. Save thumbnail to cache on Game.FadeOut(750) for Corrosion
4. Delete thumbnail on Game.UnloadObject(<interface\\system\\mainmenu.window>) for Corrosion

This has some unsupported scenarios like "opening the save menu, then opening ScummVM menu, save in second one, save in first one", but works in all normal use-cases.


sev-:
Thanks!

PR #2328, opened 2020-06-14

jepael:
Fix compiler warning about comparison between signed and unsigned expressions


sev-:
Thanks

PR #2329, opened 2020-06-14
WINTERMUTE: Implement walking with arrow keys for Corrosion

lolbot-iichan:
This implements enhancement  #11503 WME: Implement walking with arrow keys for Corrosion: https://bugs.scummvm.org/ticket/11503

 "Corrosion: Cold Winter Waiting" (both original and enchanted edition) requires a lot of walking.

However, this is originally controlled with mouse, by clicking on the edges or the center of the screen, which requires a lot of mouse moving and is not fun.

Mapping some arrow keys to mouse clicks on given regions may be a nice way to improve this game experience.

### DIFF
--- a/engines/wintermute/base/base_game.cpp
+++ b/engines/wintermute/base/base_game.cpp
@@ -3792,6 +3792,56 @@ bool BaseGame::handleMouseWheel(int32 delta) {
 	return true;
 }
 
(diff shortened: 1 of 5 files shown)


lolbot-iichan:
Retry DeepCode

sev-:
Thank you!

PR #2330, opened 2020-06-15
GLK: COMPREHEND: Fix compiler warning

jepael:
This commit removes a compiler warning about discarding qualifiers.


digitall:
Reasonable fix for the compiler warning and self-contained. Merging as-is.

PR #2333, opened 2020-06-17
KYRA/MIDI: Fix minor MT-32/GM issues

NMIError:
This PR fixes a few minor issues with MIDI playback in KYRA engine games:

- SysExes embedded in MIDI files would cause an OSystem:delayMillis call, which 
would affect the timing of the MIDI parser and cause MIDI events to "pile up" 
and be executed in a shorter timespan than intended. I've changed this so the 
MIDI parser will pause playback for the necessary time between SysExes, which 
avoids this problem. This restores a sound effect after the MT-32 
initialization of Legend of Kyrandia.
- I've added the instrument fallback feature of the Roland SC-55 that corrects 
invalid bank selects. Lands of Lore depends on this in some cases, f.e. the 
sound effect of the ring in the introduction.
- Lands of Lore GM initialization was not done. These SysExes only affect the 
display of the Roland SC-55 and are not important, but it is a quite unique 
and fun feature.
- The KYRA engine games use MIDI files with multiple tracks in them. Because 
the MIDI parser would automatically start playback of the first track after 
loading the MIDI data, the engine would load the data, stop playback, then set 
the correct track and start playback. I've now made use of a recently added 
feature of the parser to disable the start of playback after loading the data, 
so that stopping playback first is no longer necessary.
- I've changed the way All Notes Off events are used by the engine. In some 
places these were sent on all channels after stopping playback of the music, 
but this risks cutting off notes of MIDI sound effects playing at the same 
time. I've added them when quitting the game.
- I've added pausing and resuming to the MIDI parser, so that the engine no 
longer needs to remove and re-add the driver to the parser when pausing the 
MIDI playback.

### DIFF
--- a/audio/mididrv.cpp
+++ b/audio/mididrv.cpp
@@ -631,6 +631,80 @@ void MidiDriver::sendGMReset() {
 	g_system->delayMillis(100);
 }
 
+byte MidiDriver::correctInstrumentBank(byte outputChannel, byte patchId) {
+	if (_gsBank[outputChannel] == 0 || patchId >= 120 || _gsBank[outputChannel] >= 64)
+		// Usually, no bank select has been sent and no correction is
+		// necessary.
+		// No correction is performed on banks 64-127 or on the SFX
+		// instruments (120-127).
+		return 0xFF;
+
+	// Determine the intended bank. This emulates the behavior of the
+	// Roland SC-55 v1.2x. Instruments have 2, 1 or 0 sub-capital tones.
+	// Depending on the selected bank and the selected instrument, the
+	// bank will "fall back" to a sub-capital tone or to the capital
+	// tone (bank 0).
+	byte correctedBank = 0xFF;
+
+	switch (patchId) {
+	case 25:  // Steel-String Guitar / 12-string Guitar / Mandolin
+		// This instrument has 2 sub-capital tones. Bank selects 17-63 
+		// are corrected to the second sub-capital tone at 16.
+		if (_gsBank[outputChannel] >= 16) {
+			correctedBank = 16;
+			break;
+		}
+		// Corrections for values below 16 are handled below.
+
+		// fall through
+	case 4:   // Electric Piano 1 / Detuned Electric Piano 1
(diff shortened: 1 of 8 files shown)


athrxx:
Thanks for your work on this. The code changes in the KYRA engine are fine for me. 
I didn't come across any regressions with my Kyra 1, Kyra 2 and LoL tests.
For the SC-55 things, I trust that they work as intended as I can't test them.
Let's wait a couple of days before merging this. I want to do some more tests and maybe other devs are affected by the miles driver changes, too.

athrxx:
664fb91 causes a regression (e. g. LOL outro Draracle scene with MT-32).

Instead of doing this:
`_music->setTrack(track);`
`_music->startPlaying();`

please do this:
`if (_music->setTrack(track))`
`	_music->startPlaying();`

Same for sfx...

NMIError:
Thanks; I did not anticipate the game trying to play invalid tracks...

athrxx:
Don't ask me why they do that :) They do have  a valid track for AdLib.

Time to merge this...
Thanks for your work!

digitall:
@athrxx / @NMIError : Minor issue with this code:
audio/mididrv.cpp: In member function ‘byte MidiDriver::correctInstrumentBank(byte, byte)’:
audio/mididrv.cpp:699:30: warning: comparison is always true due to limited range of data type [-Wtype-limits]
   if (_gsBank[outputChannel] >= 0)

_gsBank has base type of byte... but that is unsigned and thus this will always return true. Not sure what fix should be applied here.

digitall:
See here for reference: https://github.com/scummvm/scummvm/blob/master/common/scummsys.h#L432

NMIError:
Sorry about that... I intended > 0, but the check is unnecessary anyway. All the other cases have been checked above (SFX instruments, bank >= 64, sub-capitals) so anything else should use bank 0. The if can be removed and correctedBank can just be set to 0.
Do you want me to create a PR for that?

NMIError:
I created PR #2341 for this and missing initialization of the new _gsBank variable.

PR #2335, opened 2020-06-18
DIRECTOR: implement debugflag for screenshotting

rvanlaar:
It's now possible to screenshot all played frames by using:
    `--debugflags=screenshot`

Note: it's PNG.

### DIFF
--- a/engines/director/score.cpp
+++ b/engines/director/score.cpp
@@ -20,12 +20,19 @@
  *
  */
 
+#include "common/config-manager.h"
+#include "common/file.h"
 #include "common/system.h"
 
 #include "engines/util.h"
 #include "graphics/primitives.h"
 #include "graphics/macgui/macwindowmanager.h"
 #include "graphics/macgui/mactext.h"
+#include "graphics/surface.h"
+
+#ifdef USE_PNG
+#include "image/png.h"
+#endif
 
 #include "director/director.h"
 #include "director/cast.h"
@@ -34,6 +41,7 @@
 #include "director/sound.h"
 #include "director/sprite.h"
 #include "director/stxt.h"
+#include "director/util.h"
 #include "director/lingo/lingo.h"
 
 namespace Director {
@@ -144,7 +152,7 @@ Score::Score(DirectorEngine *vm) {
 
 	_numChannelsDisplayed = 0;
 
-	_framesRan = 0; // used by kDebugFewFramesOnly
+	_framesRan = 0; // used by kDebugFewFramesOnly and kDebugScreenshot
 
 	_window = nullptr;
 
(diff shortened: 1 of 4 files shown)


engines/director/score.cpp
```diff
@@ -623,6 +629,23 @@ void Score::renderFrame(uint16 frameId, bool forceUpdate, bool updateStageOnly)
...
+
+void Score::screenShot(){
+	Graphics::Surface rawSurface = _surface->rawSurface();
+	const Graphics::PixelFormat requiredFormat_4byte(4, 8, 8, 8, 8, 0, 8, 16, 24);
+	Graphics::Surface newSurface = *(rawSurface.convertTo(requiredFormat_4byte, _vm->getPalette()));
+
+	Common::String filename;
+	Common::String currentTarget = ConfMan.getActiveDomainName();
+	filename = Common::String::format("./dumps/%s-%05d.png",
+									  currentTarget.c_str(), _framesRan);
```
sev-, line 643: Nope, this will not work with the current approach: you're running same target (director) hundreds of times for Workshop movies.

You'd rather use `dumpScriptName(archName.c_str(), type, id, "png")`

engines/director/score.cpp
```diff
@@ -623,6 +629,23 @@ void Score::renderFrame(uint16 frameId, bool forceUpdate, bool updateStageOnly)
 	}
 
 	g_system->copyRectToScreen(_surface->getPixels(), _surface->pitch, 0, 0, _surface->getBounds().width(), _surface->getBounds().height());
+
+}
+
+void Score::screenShot(){
```
sev-, line 635: Code formatting is wrong here.

engines/director/score.cpp
```diff
@@ -623,6 +629,23 @@ void Score::renderFrame(uint16 frameId, bool forceUpdate, bool updateStageOnly)
...
+	Graphics::Surface rawSurface = _surface->rawSurface();
+	const Graphics::PixelFormat requiredFormat_4byte(4, 8, 8, 8, 8, 0, 8, 16, 24);
+	Graphics::Surface newSurface = *(rawSurface.convertTo(requiredFormat_4byte, _vm->getPalette()));
+
+	Common::String filename;
+	Common::String currentTarget = ConfMan.getActiveDomainName();
+	filename = Common::String::format("./dumps/%s-%05d.png",
```
sev-, line 642: You could write `Common::String filename = ....`

sev-:
Which part of the functionality is SDL-only? I cannot see it.

rvanlaar:
> Which part of the functionality is SDL-only? I cannot see it.

It was my understanding that there's a special SDL backend in backend/platforms.
Due to your comment, I take it that those are two seperate things.

I lifted the code from here:
https://github.com/scummvm/scummvm/blob/master/backends/graphics/sdl/sdl-graphics.cpp#L284
Do you want the BMP/PNG IFs and little/big endian once also in here?

sev-:
Well, what you have to do is wrap it into USE_PNG, for sure. Whether you want to produce BMP screenshots? probably not.

I would add something like:
```
#ifdef USE_PNG
writePng();
#else
warning("Screenshot requested, but PNG support is not compiled in");
#endif
```

Also, I would not make the engine _requiring_ PNG, because this is completely optional functionality.


My comment regarding screenshots in SDL-only, is that we have hardcoded Alt+S to make screenshot there. It is engine-independent, being part of the backend. In your case, you have all engine-specific code, thus, it is universal.

PR #2337, opened 2020-06-20
XEEN: Apply string input capitalization behavior from the original

FrostyHitoshura:
Should not change the game logic. This can be tested easily with the mirror or mine cart dialog.


dreammaster:
Thanks for that. I was able to simplify your changes using existing common methods, so I'm closing the pull request

FrostyHitoshura:
I prefer your change over mine. I didn't do enough investigation of the Common code before making this change. Thanks!

PR #2339, opened 2020-06-20
AGI: Apple2GS: Add `Speed` menu and commands

ZvikaZ:
Following the suggestion at https://bugs.scummvm.org/ticket/7026, added a `Speed` menu to Apple2GS AGI games.
This menu is configurable, and can be disabled.
The speed can also be changed by text parser commands.
`Normal` speed is as before, with the speed set by the apple2gs workaround mechanism that was added in the past to solve sound issues.
Other speeds are used only if there isn't a specific workaround speed for that room.

Edit
====
It's disabled by default.


sev-:
Cool stuff!

PR #2341, opened 2020-06-21
MIDI: Fix _gsBank init and bad if statement

NMIError:
Fixes a missing variable initialization and unnecessary if statement introduced by 
my latest KYRA changes.


athrxx:
Thanks for that :)

PR #2343, opened 2020-06-22
Glk tts implementation

criezy:
This is a first pass at implementing TTS for the Glk Engine.

There are some issues remaining. For example I have noted:
- There is no tts flush done for the text in `GlkAPI::glk_exit` and has a result it is not spoken.
- There is no tts flush done for the introductory text at the start of Beyond Zork, and as a result it is not spoken. This can be partially solved by doing an automatic flush on new line characters (see commented out code in speech.cpp).
- The text in the top half of the window for Adventureland is not spoken. This does not appear to be using a `TextBufferWindow`?
- Zork zero uses drawing for the first letter of some sentences, and as a result the spoken text is missing that first letter.
- Pausing the engine to show the GMM interrupts the current speech, and it is not resumed when closing the GMM.

And that is only one the three games I have tested. Any hint on how to best solve some of those would be great.

Also I have not tested any game that have multiple `TextBufferWindow`, so commit 30a1f5b is purely speculative. Any suggestion on which game I could test?

But at least this is a start.


dreammaster:
Excellent work on the pull request

My comments, in no particular order:
* The first commit for "Glk: Simplify SpeechManager" needs GLK in uppercase
* There's conflicts in engines/glk/detection.* that need to be resolved
* Engine change for pop tts looks good, but the prefix should be "ENGINES"
* You currently only initialize TTS for text buffer windows. If we want to speak any text printed in the top/status bar areas, it would need to be done likewise for TextGridWindow as well.

> There is no tts flush done for the text in GlkAPI::glk_exit and has a result it is not spoken.
Is this something that we could easily add in, I presume?

> There is no tts flush done for the introductory text at the start of Beyond Zork, and as a result it is not spoken. This can be partially solved by doing an automatic flush on new line characters (see commented out code in speech.cpp).
That could work, as long as there aren't any issues with any of the earlier games putting line breaks in the middle of sentences.

> The text in the top half of the window for Adventureland is not spoken. This does not appear to be using a TextBufferWindow?
Right, uses a TextGridWindow, as do many of the other games for their status bar line. TextGridWindow is very similarly to TextBufferWindow except that it's for fixed width text. So it's good for such things. It also allows for text input, but none of the sub-engines currently use it for anything else than display.

Zork zero uses drawing for>  the first letter of some sentences, and as a result the spoken text is missing that first letter.
Yeh, Zork Zero is a mess in that regard. I remember from when I was spending way too much time on it, that I noticed the game has special scripts for when transcripts are turned on to actually output the correct character to the output file. Maybe an option to be looked into in the future could be to hook up the ZCode transcript output to tts rather than the screen text display. But given that I plan to eventually dump all the current Frotz code, any changes like should be kept on the backburner.

> Pausing the engine to show the GMM interrupts the current speech, and it is not resumed when closing the GMM.
Not too big an issue in my opinion.

criezy:
Thank you. I rebased the code and fixed the two commit messages.
I will take a look at the TextGridWindow over the weekend.

criezy:
@dreammaster I pushed a few more commits, including one that implements TTS for the `TextGridWindow`, and rebased all the commits. It seems to work quite well now with the few games I tried. Are you OK if we merge the PR now or do you want more changes to be made?

dreammaster:
I'd say go ahead and merge the PR now.

criezy:
I tested this with about 10 different games, although I only played a few minutes of each, and after another minor change this seems to work quite well (at least on macOS). I will merge this now. Any further improvements can be done directly in master.

PR #2344, opened 2020-06-22
DIRECTOR: Move flag loading to casts itself for D4.

rvanlaar:
unk1 in loadCastDataVWCR is probably a flag, based on D4Player.
https://github.com/renaldobf/D4Player/blob/master/src/director/cast.c#L90


engines/director/cast.cpp
```diff
@@ -79,6 +79,7 @@ BitmapCast::BitmapCast(Common::ReadStreamEndian &stream, uint32 castTag, uint16
 		if (_pitch % 16)
 			_pitch += 16 - (_initialRect.width() % 16);
 	} else if (version == 4) {
+		_flags = stream.readByte();
```
sev-, line 82: when you move it further, does it mean that pitch is now read from this byte?

sev-:
Thanks!

PR #2345, opened 2020-06-23
WINTERMUTE: Theora and BaseFileManager::hasFile residualvm fixes

GunnarBirke:
Some things that popped up during implementing the WIntermute 3D port in ResidualVM.

The theora video player should use the appropriate factory method to create a surface instead of constructing it directly.

The file manager changes forward to backward slashes when opening files from a dcp package. However, when it's only looking for a file, slashes are not flipped, which is inconsistent and causes bugs in Alpha Polaris.


sev-:
Thanks!

PR #2347, opened 2020-06-24
WINTERMUTE: Remove some comments for better ResidualVM base

lolbot-iichan:


engines/wintermute/ad/ad_scene.cpp
```diff
@@ -1116,16 +1116,14 @@ bool AdScene::display() {
 //////////////////////////////////////////////////////////////////////////
 bool AdScene::updateFreeObjects() {
 	AdGame *adGame = (AdGame *)_gameRef;
-	// 3D-code removed
-	// bool is3DSet;
+	bool is3DSet = false;
```
mgerhardy, line 1119: This would produce a warning about unused variable, no?

lolbot-iichan: Well, yes. But @aquadran and @GunnarBirke were going to use this variable anyway.

lolbot-iichan: Discussed with @aquadran, removing this line.

PR #2351, opened 2020-06-28
DEVTOOLS: Unhardcode scummvm-name in create_project for XCode.

somaen:
The existing code broke create_project in ResidualVM, as
the loop that creates the actual build configurations
generated the hash from the actual object name, which
for ResidualVM would be XCConfigurationList_residualvm.

The same logic was not applied when generating the reference,
as that would use the hardcoded name XCConfigurationList_scummvm.

This patch makes a minimal fix for that, by using the
PROJECT_NAME define instead of the hardcoded name.


sev-:
Makes sense.

PR #2353, opened 2020-06-29
GLK: ZCODE: Fix moving cursor in TextGridWindow

criezy:
Currently Beyond Zork does not display correctly. It appears that the text is not correctly placed in the Text Grid Window. I don't know if this is the correct change, but just removing the division by the font width and height in `Glk:ZCode::Window::setCursor()` seems to fix the issue. I have not tested with any other games though.

Here are comparaison between current master with and without this change.

Copyright screen without the change:
image

And with the change:
image

Character setup without the change:
image

And with the change:
image

Hilltop without the change:
image

And with the change:
image

### DIFF
--- a/engines/glk/zcode/windows.cpp
+++ b/engines/glk/zcode/windows.cpp
@@ -228,8 +228,8 @@ void Window::setCursor(const Point &newPos) {
 
 void Window::setCursor() {
 	if (dynamic_cast<TextGridWindow *>(_win)) {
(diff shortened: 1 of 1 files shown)


dreammaster:
Many thanks for this. I'd planned to look into the problem today, but got distracted by still working on Comprehend. As expected, it was one of the changes I done to try and support the V6 games. I'm happy merging this in now, since I'm not counting on the v6 games as being supported for this release.

PR #2354, opened 2020-06-29
GUI: Convert GUI to use U32Strings

aryanrawlani28:
This PR basically converts everything drawing to the screen to use U32Strings and use unicode.

There are many files modified, and I will give a quick overview of what has been changed.

--------------------------

### **ScummVM:**
- All widget's Labels and Tooltips converted to U32
- When drawing in ThemeEngine, all widgets now accept U32Strings to draw
- `drawString` is overloaded to accept U32Strings (graphics/VectorRendererSpec.cpp)
- `drawDDText` is overloaded (gui/ThemeEngine.cpp)
- `getStringWidth` is overloaded for u32
- U32Strings can now also use deltaX and handleEllipsis like the String counterparts
- `handleEllipsis` (graphics/font.cpp) now uses a template
- U32Strings have a new utility - format and vformat (similiar to common/str.cpp). These take in U32Strings, and use iterators to replace. Find it in common/ustr.cpp.
-  About dialog uses U32Strings, u32iterators to function..
- Clipboards (Windows, MacOS) use U32. 
- `ttsMan->say` takes in a U32Strings. (Windows, Linux, MacOS)
- BiDi strings now use the U32 Constructor directly
- KeysDialog uses u32

-----------------------------

### **Engines:**

Engine changes in code mostly tie back to ScummVM GUI and mostly does not modify any internal behavior. Furthermore, most changes are occurring when something is absent and a warning in console + message dialog is shown.

For example, going engine by engine alphabetically, from Access to Dragons are just MessageDialog changes.

- Widget creations now always have u32 descriptions, labels, or tooltips
- Message dialogs make use of default arguments instead of providing the same argument explicitly more often
- Modify Hugo utils (yesNoBox and notify box) to use u32
- Above also have fake constructors for the above. meaning, they just redirect to the u32 constructor.
- Convert all keymap descriptions to u32 across all engines
- showConfirmationDialog in mohawk now uses u32
- showScummVMDialog also uses u32
- Scumm engine has dialogs now which use u32
- Wrapping Common::convertToU32String(msg) wherever necessary.
- MessageDialog also has a fake constructor, i.e a redirect to the u32 constructor.
- GUIErrorMessage & GUIErrorMessageFormat now use U32Strings.
- error messages across some engines use U32Strings. they are changed because they show a message dialog.

#### **GLK**
- GLK engine supports the printing of Unicode strings, so some helper functions were added to enable the proper usage of U32Strings.
- Tested for `Scott` - Game: Adventureland. No idea if the characters were correct though, but it did look proper.
greek

--------------------------------

### **Devtools:**
- Modify credits.pl to no longer write ASCII strings to credits.h
- Regenerate new credits.h file, and adjust accordingly in about dialog.

------------------------------------

### **Devtools & Translations:**
- `po/` directory has all the translation files. All these files are now encoded in UTF-8
- devtools/create_translations tool is now modified to reflect these changes.

1) Translation version 4
2) Remove all code related to codepages, charsets, charmaps, etc from the tool which writes to the .dat file.
3) Remove cp_parser.cpp, also remove the object file from modules makefile.
4) Remove iso-8859-x codepages in po/
5) Remove dependant code on the above 2 points
6) All block-sizes (the information which helps reader to jump at a specific point of language start) is written in uint32BE.
7) Remove _charset for PoMessageEntryList, and replace it with _useUTF8. A check is added to ensure all po files have utf8. if not, error out.

- Subsequently, the reading part for these files (in common/translation.cpp) has the following changes:

1) Remove reading of codepages, charsets, charmaps, etc
2) Store messages (_currentTranslationMessages.msgstr) as U32String. messages are decoded from utf8 before being stored.
3) All block-sizes read in uint32BE (see point 6 of above)

- Encode all po files in UTF-8
- Regenerate .dat file to include all the above changes.

----------------------------------------

### **Platform-specific changes**
This PR also makes changes to some platform-specific things, which I will list below. These are changed to work with U32Strings.

#### Native browser dialogs
- GTK | Windows | MacOS

#### Clipboard
- Windows | MacOS | Linux | iOS | Android

#### OSD Messages Display
- General

#### TTS
- Windows | MacOS | Linux

---------------------------------------------

### **New Translations in**
- engines/agos/POTFILES (engines/agos/midi.cpp)
- engines/glk/POTFILES (engines/glk/hugo/hugo.cpp)
- engines/lab/POTFILES (3 files, newly added)
- engines/mohawk/myst_stacks/POTFILES (engines/mohawk/myst_stacks/preview.cpp)
- engines/saga/POTFILES (engines/saga/saga.cpp)
- engines/sci/POTFILES (engines/sci/engine/kmisc.cpp)
- engines/zvision/POTFILES (engines/zvision/zvision.cpp)
- engines/supernova/supernova.cpp

### **Removed Translations in**
- engines/ultima/nuvie/nuvie.cpp, line 324 -> (`scroll->display_string("\nGame Saved\n\n");`)

-----------------------------------------

### **Misc**
- ~Add a hi_IN.po file. (Hindi translations). I have translated~~50 or so strings. I am not sure if this PR is the best place for this commit. Perhaps a new PR would be better for this?~ Removed for now. Will be added in a new PR.
- I tested some messages with counters to see how many calls are being made to drawTexts with String as parameter. This count is 0. Should they be removed, then, because everything draws with U32?
- Travis-CI passes on Linux & MacOS. Check https://travis-ci.org/github/aryanrawlani28/scummvm/builds. I have also rebuilt the .dat file, so you can clone and build to check out the changes without doing anything extra.

---------------------------------------

### **Possible issues, remaining things**
- A while back, I had removed some temp code from this branch. My ScummVM took around 20 seconds, just to launch. the .ini file was corrupted with weird symbols. I don't think this should happen again, but in case this does, delete the .ini file - this should work.
- U32String as a template of String to reduce code duplication. (This will probably be shifted to another PR and at a later time, because this PR has already a lot of changes, I think)
- ~The PR is now modifying many files. I have tried my best to not write problematic code, but have not tested everything. Especially the MacOS / Linux commits, the engine specific commits I have not really seen myself how it looks. Some help in testing to see if everything is okay would be helpful.~
- ~ScummVM project used translations like:
`_("message", "context")` or somewhat similar. A while back, I removed some temporary code, but forgot to remove the parentheses, so it looks now like: `(_("msg", "ctx"))`. I will soon remove them.~
- ~Remove .travis.yml commit before merge~

**Wiki updates TODO:**
- When near merge, the [translation information in the wiki](https://wiki.scummvm.org/index.php?title=Supporting_GUI_Translation/Translations_DAT_Format) will be outdated. Remove those. 
- create_translations tool will need to be rebuilt and uploaded for the [HOWTO-Translate](https://wiki.scummvm.org/index.php?title=HOWTO-Translate_ScummVM_GUI#Using_the_translation_in_ScummVM) ScummVM page
---------------------------------------------

All suggestions / ways to improve very welcome!

### DIFF
--- a/backends/dialogs/gtk/gtk-dialogs.cpp
+++ b/backends/dialogs/gtk/gtk-dialogs.cpp
@@ -41,13 +41,14 @@ Common::DialogManager::DialogResult GtkDialogManager::showFileBrowser(const Comm
 
 	DialogResult result = kDialogCancel;
 
-	Common::U32String choose = _("Choose");
-	Common::U32String cancel = _("Cancel");
-
 	// Convert labels to UTF-8
-	char *utf8Title = (char *)title.encode().c_str();
-	char *utf8Choose = (char *)choose.encode().c_str();
-	char* utf8Cancel = (char *)cancel.encode().c_str();
+	Common::String titleStr = title.encode();
+	Common::String choose = _("Choose").encode();
+	Common::String cancel = _("Cancel").encode();
+
+	char *utf8Title = (char *)(titleStr.c_str());
+	char *utf8Choose = (char *)(choose.c_str());
+	char *utf8Cancel = (char *)(cancel.c_str());
 
 	GtkFileChooserAction action = GTK_FILE_CHOOSER_ACTION_OPEN;
 	if (isDirBrowser) {
@@ -60,9 +61,6 @@ Common::DialogManager::DialogResult GtkDialogManager::showFileBrowser(const Comm
 	GtkWidget *dialog = gtk_file_chooser_dialog_new(utf8Title, NULL, action, utf8Choose, GTK_RESPONSE_ACCEPT, utf8Cancel, GTK_RESPONSE_CANCEL, NULL);
 	GtkFileChooser *chooser = GTK_FILE_CHOOSER(dialog);
 #endif
-	free(utf8Cancel);
-	free(utf8Choose);
-	free(utf8Title);
 
 	// Customize dialog
 	gtk_file_chooser_set_show_hidden(chooser, ConfMan.getBool("gui_browser_show_hidden", Common::ConfigManager::kApplicationDomain));
--- a/backends/dialogs/gtk/gtk-dialogs.cpp
+++ b/backends/dialogs/gtk/gtk-dialogs.cpp
@@ -42,23 +42,19 @@ Common::DialogManager::DialogResult GtkDialogManager::showFileBrowser(const Comm
 	DialogResult result = kDialogCancel;
 
 	// Convert labels to UTF-8
-	Common::String titleStr = title.encode();
-	Common::String choose = _("Choose").encode();
-	Common::String cancel = _("Cancel").encode();
-
-	char *utf8Title = (char *)(titleStr.c_str());
-	char *utf8Choose = (char *)(choose.c_str());
-	char *utf8Cancel = (char *)(cancel.c_str());
+	Common::String utf8Title = title.encode();
+	Common::String utf8Choose = _("Choose").encode();
+	Common::String utf8Cancel = _("Cancel").encode();
 
 	GtkFileChooserAction action = GTK_FILE_CHOOSER_ACTION_OPEN;
 	if (isDirBrowser) {
 		action = GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER;
 	}
 #if GTK_CHECK_VERSION(3,20,0)
-	GtkFileChooserNative *native = gtk_file_chooser_native_new(utf8Title, NULL, action, utf8Choose, utf8Cancel);
+	GtkFileChooserNative *native = gtk_file_chooser_native_new(utf8Title.c_str(), NULL, action, utf8Choose.c_str(), utf8Cancel.c_str());
 	GtkFileChooser *chooser = GTK_FILE_CHOOSER(native);
 #else
-	GtkWidget *dialog = gtk_file_chooser_dialog_new(utf8Title, NULL, action, utf8Choose, GTK_RESPONSE_ACCEPT, utf8Cancel, GTK_RESPONSE_CANCEL, NULL);
+	GtkWidget *dialog = gtk_file_chooser_dialog_new(utf8Title.c_str(), NULL, action, utf8Choose.c_str(), GTK_RESPONSE_ACCEPT, utf8Cancel.c_str(), GTK_RESPONSE_CANCEL, NULL);
 	GtkFileChooser *chooser = GTK_FILE_CHOOSER(dialog);
 #endif
 
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -659,7 +659,10 @@ bool OpenGLSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 		endGFXTransaction();
 
 #ifdef USE_OSD
-		Common::U32String message = _("Stretch mode") + Common::U32String(": ") + _(stretchModes[index].description);
+		Common::U32String message = Common::U32String::format(Common::U32String("%S: %S"),
+			_("Stretch mode").c_str(),
+			_(stretchModes[index].description).c_str()
+		);
 		displayMessageOnOSD(message);
 #endif
 
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -2500,16 +2500,16 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 		Common::U32String message;
 		if (_videoMode.aspectRatioCorrection)
 			message = Common::U32String::format(Common::U32String("%S\n%d x %d -> %d x %d"),
-											 _("Enabled aspect ratio correction").c_str(),
-			                                 _videoMode.screenWidth, _videoMode.screenHeight,
-			                                 _hwScreen->w, _hwScreen->h
-			);
+			                                    _("Enabled aspect ratio correction").c_str(),
+			                                    _videoMode.screenWidth, _videoMode.screenHeight,
+			                                    _hwScreen->w, _hwScreen->h
+			          );
 		else
 			message = Common::U32String::format(Common::U32String("%S\n%d x %d -> %d x %d"),
-											 _("Disabled aspect ratio correction").c_str(),
-			                                 _videoMode.screenWidth, _videoMode.screenHeight,
-			                                 _hwScreen->w, _hwScreen->h
-			);
+			                                    _("Disabled aspect ratio correction").c_str(),
+			                                    _videoMode.screenWidth, _videoMode.screenHeight,
+			                                    _hwScreen->w, _hwScreen->h
+			          );
 		displayMessageOnOSD(message);
 #endif
 		internUpdateScreen();
@@ -2553,9 +2553,9 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 
 #ifdef USE_OSD
 		Common::U32String message = Common::U32String::format(Common::U32String("%S: %S"),
-														_("Stretch mode").c_str(),
-														_(s_supportedStretchModes[index].description).c_str()
-									);
+		                                                      _("Stretch mode").c_str(),
+		                                                      _(s_supportedStretchModes[index].description).c_str()
+		                            );
 		displayMessageOnOSD(message);
 #endif
 		_forceRedraw = true;
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -2465,7 +2465,8 @@ void SurfaceSdlGraphicsManager::handleScalerHotkeys(int scalefactor, int scalerT
 		}
 		if (newScalerName) {
 			const Common::U32String message = Common::U32String::format(
-				_("Active graphics filter:") + Common::U32String("%s\n%d x %d -> %d x %d"),
+				Common::U32String("%S %s\n%d x %d -> %d x %d"),
+				_("Active graphics filter:").c_str(),
 				newScalerName,
 				_videoMode.screenWidth, _videoMode.screenHeight,
 				_hwScreen->w, _hwScreen->h);
@@ -2498,12 +2499,14 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 #ifdef USE_OSD
 		Common::U32String message;
 		if (_videoMode.aspectRatioCorrection)
-			message = Common::U32String::format(_("Enabled aspect ratio correction") + Common::U32String("\n%d x %d -> %d x %d"),
+			message = Common::U32String::format(Common::U32String("%S\n%d x %d -> %d x %d"),
+											 _("Enabled aspect ratio correction").c_str(),
 			                                 _videoMode.screenWidth, _videoMode.screenHeight,
 			                                 _hwScreen->w, _hwScreen->h
 			);
 		else
-			message = Common::U32String::format(_("Disabled aspect ratio correction") + Common::U32String("\n%d x %d -> %d x %d"),
+			message = Common::U32String::format(Common::U32String("%S\n%d x %d -> %d x %d"),
+											 _("Disabled aspect ratio correction").c_str(),
 			                                 _videoMode.screenWidth, _videoMode.screenHeight,
 			                                 _hwScreen->w, _hwScreen->h
 			);
@@ -2549,7 +2552,10 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 		endGFXTransaction();
 
 #ifdef USE_OSD
-		Common::U32String message = _("Stretch mode") + Common::U32String(": ") + _(s_supportedStretchModes[index].description);
+		Common::U32String message = Common::U32String::format(Common::U32String("%S: %S"),
+														_("Stretch mode").c_str(),
+														_(s_supportedStretchModes[index].description).c_str()
+									);
 		displayMessageOnOSD(message);
 #endif
 		_forceRedraw = true;
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -2503,7 +2503,7 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 			                                 _hwScreen->w, _hwScreen->h
 			);
 		else
-			message = Common::U32String::format(_("Disabled aspect ratio correction") + Common::U32String("%s\n%d x %d -> %d x %d"),
+			message = Common::U32String::format(_("Disabled aspect ratio correction") + Common::U32String("\n%d x %d -> %d x %d"),
 			                                 _videoMode.screenWidth, _videoMode.screenHeight,
 			                                 _hwScreen->w, _hwScreen->h
 			);
--- a/backends/platform/ios7/ios7_osys_misc.mm
+++ b/backends/platform/ios7/ios7_osys_misc.mm
@@ -48,37 +48,45 @@ bool OSystem_iOS7::hasTextInClipboard() {
 	return [[UIPasteboard generalPasteboard] containsPasteboardTypes:UIPasteboardTypeListString];
 }
 
-Common::String OSystem_iOS7::getTextFromClipboard() {
+Common::U32String OSystem_iOS7::getTextFromClipboard() {
 	if (!hasTextInClipboard())
-		return Common::String();
+		return Common::U32String();
 
 	UIPasteboard *pb = [UIPasteboard generalPasteboard];
 	NSString *str = pb.string;
 	if (str == nil)
-		return Common::String();
+		return Common::U32String();
 
 	// If translations are supported, use the current TranslationManager charset and otherwise
 	// use ASCII. If the string cannot be represented using the requested encoding we get a null
 	// pointer below, which is fine as ScummVM would not know what to do with the string anyway.
-#ifdef USE_TRANSLATION
-	NSString* encStr = [NSString stringWithCString:TransMan.getCurrentCharset().c_str() encoding:NSASCIIStringEncoding];
-	NSStringEncoding encoding = CFStringConvertEncodingToNSStringEncoding(CFStringConvertIANACharSetNameToEncoding((CFStringRef)encStr));
+#ifdef SCUMM_LITTLE_ENDIAN
+	NSStringEncoding stringEncoding = NSUTF32LittleEndianStringEncoding;
 #else
-	NSStringEncoding encoding = NSISOLatin1StringEncoding;
+	NSStringEncoding stringEncoding = NSUTF32BigEndianStringEncoding;
 #endif
-	return Common::String([str cStringUsingEncoding:encoding]);
+	NSUInteger textLength = [str length];
+	uint32 *text = new uint32[textLength];
+
+	if (![str getBytes:text maxLength:4*textLength usedLength:NULL encoding: stringEncoding options:0 range:NSMakeRange(0, textLength) remainingRange:NULL]) {
+		delete[] text;
+		return Common::U32String();
+	}
+
+	Common::U32String u32String(text, textLength);
+	delete[] text;
+
+	return u32String;
 }
 
-bool OSystem_iOS7::setTextInClipboard(const Common::String &text) {
-#ifdef USE_TRANSLATION
-	NSString* encStr = [NSString stringWithCString:TransMan.getCurrentCharset().c_str() encoding:NSASCIIStringEncoding];
-	NSStringEncoding encoding = CFStringConvertEncodingToNSStringEncoding(CFStringConvertIANACharSetNameToEncoding((CFStringRef)encStr));
+bool OSystem_iOS7::setTextInClipboard(const Common::U32String &text) {
+#ifdef SCUMM_LITTLE_ENDIAN
+	NSStringEncoding stringEncoding = NSUTF32LittleEndianStringEncoding;
 #else
-	NSStringEncoding encoding = NSISOLatin1StringEncoding;
+	NSStringEncoding stringEncoding = NSUTF32BigEndianStringEncoding;
 #endif
 	UIPasteboard *pb = [UIPasteboard generalPasteboard];
-	[pb setString:[NSString stringWithCString:text.c_str() encoding:encoding]];
-	return true;
+	return [pb setString:[[NSString alloc] initWithBytes:text.c_str() length:4*text.size() encoding: stringEncoding] forType:NSStringPboardType];
 }
 
 bool OSystem_iOS7::openUrl(const Common::String &url) {
--- a/backends/platform/ios7/ios7_osys_misc.mm
+++ b/backends/platform/ios7/ios7_osys_misc.mm
@@ -86,7 +86,10 @@ bool OSystem_iOS7::setTextInClipboard(const Common::U32String &text) {
 	NSStringEncoding stringEncoding = NSUTF32BigEndianStringEncoding;
 #endif
 	UIPasteboard *pb = [UIPasteboard generalPasteboard];
-	return [pb setString:[[NSString alloc] initWithBytes:text.c_str() length:4*text.size() encoding: stringEncoding] forType:NSStringPboardType];
+	NSString *nsstring = [[NSString alloc] initWithBytes:text.c_str() length:4*text.size() encoding: stringEncoding];
+	[pb setString:nsstring];
+	[nsstring release];
+	return true;
 }
 
 bool OSystem_iOS7::openUrl(const Common::String &url) {
--- a/engines/bbvs/dialogs.cpp
+++ b/engines/bbvs/dialogs.cpp
@@ -86,11 +86,11 @@ MainMenu::~MainMenu() {
 }
 
 void MainMenu::init() {
-	_buttons[0] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
-	_buttons[1] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
-	_buttons[2] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
-	_buttons[3] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
-	_buttons[4] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
+	_buttons[0] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""));
+	_buttons[1] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""));
+	_buttons[2] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""));
+	_buttons[3] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""));
+	_buttons[4] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""));
 	gotoMenuScreen(kMainMenuScr);
 }
 
--- a/engines/drascula/drascula.cpp
+++ b/engines/drascula/drascula.cpp
@@ -928,9 +928,11 @@ bool DrasculaEngine::loadDrasculaDat() {
 	in.open(filename.c_str());
 
 	if (!in.isOpen()) {
-		Common::U32String errorMessage = Common::U32String::format(_("Unable to locate the '%s' engine data file."), filename.c_str());
+		const char *msg = _s("Unable to locate the '%s' engine data file.");
+
+		Common::U32String errorMessage = Common::U32String::format(_(msg), filename.c_str());
 		GUIErrorMessage(errorMessage);
-		warning("Unable to locate the '%s' engine data file.", filename.c_str());
+		warning(msg, filename.c_str());
 
 		return false;
 	}
@@ -942,9 +944,10 @@ bool DrasculaEngine::loadDrasculaDat() {
 	buf[8] = '\0';
 
 	if (strcmp(buf, "DRASCULA") != 0) {
-		Common::U32String errorMessage = Common::U32String::format(_("The '%s' engine data file is corrupt."), filename.c_str());
+		const char *msg = _s("The '%s' engine data file is corrupt.");
+		Common::U32String errorMessage = Common::U32String::format(_(msg), filename.c_str());
 		GUIErrorMessage(errorMessage);
-		warning("The '%s' engine data file is corrupt.", filename.c_str());
+		warning(msg, filename.c_str());
 
 		return false;
 	}
@@ -952,12 +955,10 @@ bool DrasculaEngine::loadDrasculaDat() {
 	ver = in.readByte();
 
 	if (ver != DRASCULA_DAT_VER) {
-		Common::U32String errorMessage = Common::U32String::format(
-			_("Incorrect version of the '%s' engine data file found. Expected %d.%d but got %d.%d."),
-			filename.c_str(), DRASCULA_DAT_VER, 0, ver, 0);
+		const char *msg = _s("Incorrect version of the '%s' engine data file found. Expected %d.%d but got %d.%d.");
+		Common::U32String errorMessage = Common::U32String::format(_(msg), filename.c_str(), DRASCULA_DAT_VER, 0, ver, 0);
 		GUIErrorMessage(errorMessage);
-		warning("Incorrect version of the '%s' engine data file found. Expected %d.%d but got %d.%d.",
-			filename.c_str(), DRASCULA_DAT_VER, 0, ver, 0);
+		warning(msg, filename.c_str(), DRASCULA_DAT_VER, 0, ver, 0);
 
 		return false;
 	}
--- a/engines/glk/adrift/os_glk.cpp
+++ b/engines/glk/adrift/os_glk.cpp
@@ -2811,7 +2811,7 @@ static int gsc_startup_code(Common::SeekableReadStream *game_stream, int restore
 		 * Display a brief loading game message; here we have to use a timeout
 		 * to ensure that the text is flushed to Glk.
 		 */
-		g_vm->glk_put_string("Loading game...\n");
+		g_vm->glk_put_string_uni(_("Loading game...\n").c_str());
 		if (g_vm->glk_gestalt(gestalt_Timer, 0)) {
 			event_t event;
 
--- a/engines/glk/advsys/vm.cpp
+++ b/engines/glk/advsys/vm.cpp
@@ -22,6 +22,7 @@
 
 #include "glk/advsys/vm.h"
 #include "common/translation.h"
+#include "common/ustr.h"
 
 namespace Glk {
 namespace AdvSys {
@@ -593,7 +594,7 @@ bool VM::getWord(Common::String &line) {
 		_words.push_back(iw);
 		return true;
 	} else {
-		Common::String msg = Common::String::format("I don't know the word \"%s\".\n", iw._text.c_str());
+		Common::U32String msg = Common::U32String::format(_("I don't know the word \"%s\".\n"), iw._text.c_str());
 		print(msg);
 		return false;
 	}
--- a/engines/glk/comprehend/game.cpp
+++ b/engines/glk/comprehend/game.cpp
@@ -1330,7 +1330,7 @@ void ComprehendGame::read_input() {
 
 		// Empty line, so toggle picture window visibility
 		g_comprehend->toggleGraphics();
-		g_comprehend->print("Picture window toggled\n");
+		g_comprehend->print(_("Picture window toggled\n"));
 
 		_updateFlags |= UPDATE_GRAPHICS;
 		update_graphics();
--- a/engines/glk/glk_api.cpp
+++ b/engines/glk/glk_api.cpp
@@ -60,7 +60,7 @@ GlkAPI::GlkAPI(OSystem *syst, const GlkGameDescription &gameDesc) :
 }
 
 void GlkAPI::glk_exit(void) {
-	glk_put_string("[ press any key to exit ]");
+	glk_put_string_uni(_("[ press any key to exit ]").c_str());
 	_events->waitForPress();
 
 	// Trigger a ScumMVM shutdown of game
--- a/engines/glk/scott/scott.cpp
+++ b/engines/glk/scott/scott.cpp
@@ -24,6 +24,7 @@
 #include "glk/quetzal.h"
 #include "common/config-manager.h"
 #include "common/translation.h"
+#include "common/ustr.h"
 
 namespace Glk {
 namespace Scott {
@@ -164,6 +165,18 @@ void Scott::display(winid_t w, const char *fmt, ...) {
 	va_end(ap);
 
 	glk_put_string_stream(glk_window_get_stream(w), msg.c_str());
+}
+
+void Scott::display(winid_t w, const Common::U32String fmt, ...) {
+	Common::U32String msg;
+
+	va_list ap;
+
+	va_start(ap, fmt);
+	Common::U32String::vformat(fmt.begin(), fmt.end(), msg, ap);
+	va_end(ap);
+
+	glk_put_string_stream_uni(glk_window_get_stream(w), msg.c_str());
 }
 
 void Scott::delay(int seconds) {
@@ -385,13 +398,18 @@ void Scott::output(const Common::String &a) {
 		display(_bottomWindow, "%s", a.c_str());
 }
 
+void Scott::output(const Common::U32String &a) {
+	if (_saveSlot == -1)
+		display(_bottomWindow, Common::U32String("%S"), a.c_str());
+}
+
 void Scott::outputNumber(int a) {
 	display(_bottomWindow, "%d", a);
 }
 
 void Scott::look(void) {
-	const char *const ExitNames[6] = {
-		_("North").encode().c_str(), _("South").encode().c_str(), _("East").encode().c_str(), _("West").encode().c_str(), _("Up").encode().c_str(), _("Down").encode().c_str()
+	const uint32 *const ExitNames[6] = {
+		_("North").c_str(), _("South").c_str(), _("East").c_str(), _("West").c_str(), _("Up").c_str(), _("Down").c_str()
 	};
 	Room *r;
 	int ct, f;
@@ -403,9 +421,9 @@ void Scott::look(void) {
 	if ((_bitFlags & (1 << DARKBIT)) && _items[LIGHT_SOURCE]._location != CARRIED
 			&& _items[LIGHT_SOURCE]._location != MY_LOC) {
 		if (_options & YOUARE)
-			display(_topWindow, _("You can't see. It is too dark!\n").encode().c_str());
+			display(_topWindow, _("You can't see. It is too dark!\n"));
 		else
-			display(_topWindow, _("I can't see. It is too dark!\n").encode().c_str());
+			display(_topWindow, _("I can't see. It is too dark!\n"));
 		if (_options & TRS80_STYLE)
 			display(_topWindow, TRS80_LINE);
 		return;
@@ -415,27 +433,27 @@ void Scott::look(void) {
 		display(_topWindow, "%s\n", r->_text.c_str() + 1);
 	else {
 		if (_options & YOUARE)
-			display(_topWindow, _("You are in a %s\n").encode().c_str(), r->_text.c_str());
+			display(_topWindow, _("You are in a %s\n"), r->_text.c_str());
 		else
-			display(_topWindow, _("I'm in a %s\n").encode().c_str(), r->_text.c_str());
+			display(_topWindow, _("I'm in a %s\n"), r->_text.c_str());
 	}
 
 	ct = 0;
 	f = 0;
-	display(_topWindow, _("\nObvious exits: ").encode().c_str());
+	display(_topWindow, _("\nObvious exits: "));
 	while (ct < 6) {
 		if (r->_exits[ct] != 0) {
 			if (f == 0)
 				f = 1;
 			else
 				display(_topWindow, ", ");
-			display(_topWindow, "%s", ExitNames[ct]);
+			display(_topWindow, Common::U32String("%S"), ExitNames[ct]);
 		}
 		ct++;
 	}
 
 	if (f == 0)
-		display(_topWindow, _("none").encode().c_str());
+		display(_topWindow, _("none"));
 	display(_topWindow, ".\n");
 	ct = 0;
 	f = 0;
@@ -444,10 +462,10 @@ void Scott::look(void) {
 		if (_items[ct]._location == MY_LOC) {
 			if (f == 0) {
 				if (_options & YOUARE) {
-					display(_topWindow, _("\nYou can also see: ").encode().c_str());
+					display(_topWindow, _("\nYou can also see: "));
 					pos = 18;
 				} else {
-					display(_topWindow, _("\nI can also see: ").encode().c_str());
+					display(_topWindow, _("\nI can also see: "));
 					pos = 16;
 				}
 				f++;
--- a/engines/glk/scott/scott.cpp
+++ b/engines/glk/scott/scott.cpp
@@ -408,8 +408,8 @@ void Scott::outputNumber(int a) {
 }
 
 void Scott::look(void) {
-	const uint32 *const ExitNames[6] = {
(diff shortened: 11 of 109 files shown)


gui/credits.h
```diff
@@ -461,8 +434,6 @@ static const char *credits[] = {
 "C2""(Pentagram)",
 "C0""Dominik Reichardt",
 "C2""(Pentagram)",
-"C0""Matthew Jimenez",
-"C2""(Pentagram)",
```
aryanrawlani28, line 465: an author is being removed from here. is this expected? I modified the tool to not write any ascii strings.

ccawley2011: It's expected - the relevant commit is this one: 9c4e6244100f89b0e078006de7ba73dbb36d46a4.

aryanrawlani28: Ah, I see. thanks!

engines/hugo/hugo.cpp
```diff
@@ -433,9 +433,9 @@ bool HugoEngine::loadHugoDat() {
 	in.open(filename.c_str());
 
 	if (!in.isOpen()) {
-		Common::String errorMessage = Common::String::format(_("Unable to locate the '%s' engine data file."), filename.c_str());
+		Common::U32String errorMessage = Common::U32String::format(_("Unable to locate the '%s' engine data file."), filename.c_str());
```
sev-, line 436: This needs to be reverted because we do not support Unicode in `warning()`

engines/hugo/hugo.cpp
```diff
@@ -454,7 +454,7 @@ bool HugoEngine::loadHugoDat() {
 
 	if ((majVer != HUGO_DAT_VER_MAJ) || (minVer != HUGO_DAT_VER_MIN)) {
 		Common::String errorMessage = Common::String::format(
```
sev-, line 456: shouldn't this be U32?

engines/kyra/engine/eobcommon.cpp
```diff
@@ -341,7 +341,7 @@ EoBCoreEngine::~EoBCoreEngine() {
 }
 
 Common::KeymapArray EoBCoreEngine::initKeymaps(const Common::String &gameId) {
-	Common::Keymap *const engineKeyMap = new Common::Keymap(Common::Keymap::kKeymapTypeGame, kKeymapName, "Eye of the Beholder");
+	Common::Keymap *const engineKeyMap = new Common::Keymap(Common::Keymap::kKeymapTypeGame, kKeymapName, Common::convertToU32String("Eye of the Beholder"));
```
sev-, line 344: I would overload `Common::Keymap` and revert this.

engines/lab/processroom.cpp
```diff
@@ -326,7 +326,7 @@ void LabEngine::doActions(const ActionList &actionList) {
 				// This is a Wyrmkeep Windows trial version, thus stop at this
 				// point, since we can't check for game payment status
 				_graphics->readPict(getPictName(true));
-				GUI::MessageDialog trialMessage("This is the end of the trial version. You can play the full game using the original interpreter from Wyrmkeep");
+				GUI::MessageDialog trialMessage(Common::convertToU32String("This is the end of the trial version. You can play the full game using the original interpreter from Wyrmkeep"));
```
sev-, line 329: Actually, this is a mistake, and it must be wrapped in `_()`, e.g. marked for translation.

engines/lab/speciallocks.cpp
```diff
@@ -145,7 +145,7 @@ void SpecialLocks::changeTile(uint16 col, uint16 row) {
 
 	if (scrolltype != -1) {
 		if (_vm->getFeatures() & GF_WINDOWS_TRIAL) {
-			GUI::MessageDialog trialMessage("This puzzle is not available in the trial version of the game");
+			GUI::MessageDialog trialMessage(Common::convertToU32String("This puzzle is not available in the trial version of the game"));
```
sev-, line 148: Same thing here. Wrap it in `_()`

engines/lure/surface.cpp
```diff
@@ -491,7 +491,7 @@ Surface *Surface::newDialog(uint16 width, uint8 numLines, const char **lines, bo
 		Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
 		if (ttsMan != nullptr) {
 			ttsMan->stop();
-			ttsMan->say(text.c_str());
+			ttsMan->say(Common::convertToU32String(text.c_str()));
```
sev-, line 494: I would make `say()` accept `const char *` and revert this

engines/mohawk/dialogs.cpp
```diff
@@ -122,16 +122,16 @@ MystOptionsWidget::MystOptionsWidget(GuiObject *boss, const Common::String &name
 		assert(vm);
 
 		// I18N: Drop book page
-		_dropPageButton = new GUI::ButtonWidget(widgetsBoss(), "MystOptionsDialog.DropPage", _("~D~rop Page"), nullptr, kDropCmd);
+		_dropPageButton = new GUI::ButtonWidget(widgetsBoss(), "MystOptionsDialog.DropPage", _("~D~rop Page"), Common::U32String(""), kDropCmd);
```
sev-, line 125: This needs to be reverted

engines/saga/saga.cpp
```diff
@@ -243,7 +243,7 @@ Common::Error SagaEngine::run() {
 
 	// Detect game and open resource files
 	if (!initGame()) {
-		GUIErrorMessage("Error loading game resources.");
+		GUIErrorMessage(Common::convertToU32String("Error loading game resources."));
```
sev-, line 246: Wrapt this too into `_()`

criezy:
I will try to review this this weekend.

But I would like already to flag that whoever merges this will need to be careful with the translations. The workflow I would recommend to avoid any issue is:
1. Commit and push any pending changes from weblate.
2. Lock weblate.
3. Rebase this PR and deal with any conflict on the po files by reconverting the version from master from its original encoding to utf-8.
4. Merge the PR.
5. Unlock weblate.

engines/mohawk/dialogs.cpp
aryanrawlani28: for what reason?
here is the function definition for button widget:

```
ButtonWidget(GuiObject *boss, int x, int y, int w, int h, const Common::U32String &label, const Common::U32String &tooltip = Common::U32String(""), uint32 cmd = 0, uint8 hotkey = 0);
	ButtonWidget(GuiObject *boss, const Common::String &name, const Common::U32String &label, const Common::U32String &tooltip = Common::U32String(""), uint32 cmd = 0, uint8 hotkey = 0);
```

engines/glk/advsys/vm.cpp
```diff
@@ -593,8 +593,8 @@ bool VM::getWord(Common::String &line) {
 		_words.push_back(iw);
 		return true;
 	} else {
-		Common::String msg = Common::String::format(_("I don't know the word \"%s\".\n"), iw._text.c_str());
-		print(msg);
+		Common::U32String msg = Common::U32String::format(_("I don't know the word \"%s\".\n"), iw._text.c_str());
+		print(msg.encode());
```
aryanrawlani28, line 597: should things like these not use translations? print would not support chars from other than ascii (I think). i will also mark some more places like these, which I have some doubt

engines/glk/comprehend/game.cpp
```diff
@@ -1339,7 +1339,7 @@ void ComprehendGame::read_input() {
 
 		// Empty line, so toggle picture window visibility
 		g_comprehend->toggleGraphics();
-		g_comprehend->print(_("Picture window toggled\n"));
+		g_comprehend->print(Common::convertFromU32String(_("Picture window toggled\n")).c_str());
```
aryanrawlani28, line 1342: this

engines/glk/glk_api.cpp
```diff
@@ -60,7 +60,7 @@ GlkAPI::GlkAPI(OSystem *syst, const GlkGameDescription &gameDesc) :
 }
 
 void GlkAPI::glk_exit(void) {
-	glk_put_string(_("[ press any key to exit ]"));
+	glk_put_string(_("[ press any key to exit ]").encode().c_str());
```
aryanrawlani28, line 63: this

engines/glk/scott/scott.cpp
```diff
@@ -403,9 +403,9 @@ void Scott::look(void) {
 	if ((_bitFlags & (1 << DARKBIT)) && _items[LIGHT_SOURCE]._location != CARRIED
 			&& _items[LIGHT_SOURCE]._location != MY_LOC) {
 		if (_options & YOUARE)
-			display(_topWindow, _("You can't see. It is too dark!\n"));
+			display(_topWindow, _("You can't see. It is too dark!\n").encode().c_str());
 		else
-			display(_topWindow, _("I can't see. It is too dark!\n"));
+			display(_topWindow, _("I can't see. It is too dark!\n").encode().c_str());
```
aryanrawlani28, line 408: this

engines/glk/zcode/zcode.cpp
```diff
@@ -141,7 +141,7 @@ Common::Error ZCode::saveGameState(int slot, const Common::String &desc, bool is
 	bool success = q.save(*file, this, desc);
 
 	if (!success)
-		print_string(_("Error writing save file\n"));
+		print_string(_("Error writing save file\n").encode().c_str());
```
aryanrawlani28, line 144: this

engines/glk/zcode/zcode.cpp
```diff
@@ -123,7 +123,7 @@ Common::Error ZCode::loadGameState(int slot) {
 			|| h_screen_cols != old_screen_cols))
 			erase_window(1);
 	} else {
-		error("%s", _("Error reading save file"));
+		error("%s", ("Error reading save file"));
```
aryanrawlani28, line 126: this

engines/scumm/scumm.cpp
```diff
@@ -2470,13 +2470,13 @@ void ScummEngine::scummLoop_handleSaveLoad() {
 		}
 
 		if (!success) {
-			displayMessage(0, errMsg, filename.c_str());
+			displayMessage(0, errMsg.encode().c_str(), filename.c_str());
```
aryanrawlani28, line 2473: this

engines/scumm/scumm.cpp
```diff
@@ -2470,13 +2470,13 @@ void ScummEngine::scummLoop_handleSaveLoad() {
...
 		} else if (_saveLoadFlag == 1 && _saveLoadSlot != 0 && !_saveTemporaryState) {
 			// Display "Save successful" message, except for auto saves
 			char buf[256];
-			snprintf(buf, sizeof(buf), _("Successfully saved game in file:\n\n%s"), filename.c_str());
+			snprintf(buf, sizeof(buf), _("Successfully saved game in file:\n\n%s").encode().c_str(), filename.c_str());
```
aryanrawlani28, line 2477: this

audio/mididrv.cpp
```diff
@@ -145,7 +145,7 @@ MusicType MidiDriver::getMusicType(MidiDriver::DeviceHandle handle) {
 	return MT_INVALID;
 }
 
-Common::String MidiDriver::getDeviceString(DeviceHandle handle, DeviceStringType type) {
+Common::U32String MidiDriver::getDeviceString(DeviceHandle handle, DeviceStringType type) {
```
sev-, line 148: This should be kept at String, the whole file except the places which use `_()`. No need to return U32String where there is ASCII-only

audio/musicplugin.cpp
```diff
@@ -29,33 +29,33 @@ MusicDevice::MusicDevice(MusicPluginObject const *musicPlugin, Common::String na
 	_name(_(name)), _type(mt) {
 }
 
-Common::String MusicDevice::getCompleteName() {
-	Common::String name;
+Common::U32String MusicDevice::getCompleteName() {
```
sev-, line 32: Same thing here. We do not translate driver names, thus, leave it as String

.travis.yml
```diff
@@ -57,7 +57,7 @@ addons:
 
 branches:
  only:
-   - master
```
sev-, line 60: Why these changes are here? Perhaps, you would need to rebase.

audio/musicplugin.h
```diff
@@ -42,29 +43,29 @@ class MusicDevice {
 public:
 	MusicDevice(MusicPluginObject const *musicPlugin, Common::String name, MusicType mt);
 
-	const Common::String &getName() const { return _name; }
-	const Common::String &getMusicDriverName() const { return _musicDriverName; }
-	const Common::String &getMusicDriverId() const { return _musicDriverId; }
+	const Common::U32String &getName() const { return _name; }
```
sev-, line 46: Obviously, this must be also reverted. Those names are ASCII-only.

audio/softsynth/mt32.cpp
```diff
@@ -67,17 +67,17 @@ class ScummVMReportHandler : public MT32Emu::IReportHandler {
 
 	// Callbacks for reporting various errors and information
 	void onErrorControlROM() {
-		GUI::MessageDialog dialog("MT32Emu: Init Error - Missing or invalid Control ROM image", "OK");
+		GUI::MessageDialog dialog("MT32Emu: Init Error - Missing or invalid Control ROM image");
```
sev-, line 70: Why is this change?

audio/softsynth/mt32.cpp
```diff
@@ -67,17 +67,17 @@ class ScummVMReportHandler : public MT32Emu::IReportHandler {
...
 		dialog.runModal();
 		error("MT32emu: Init Error - Missing or invalid Control ROM image");
 	}
 	void onErrorPCMROM() {
-		GUI::MessageDialog dialog("MT32Emu: Init Error - Missing PCM ROM image", "OK");
+		GUI::MessageDialog dialog("MT32Emu: Init Error - Missing PCM ROM image");
 		dialog.runModal();
 		error("MT32emu: Init Error - Missing PCM ROM image");
 	}
 	void showLCDMessage(const char *message) {
```
sev-, line 79: This is ASCII-only, needs to be reverted.

backends/cloud/cloudmanager.cpp
```diff
@@ -153,8 +153,8 @@ void CloudManager::replaceStorage(Storage *storage, uint32 index) {
 	_currentStorageIndex = index;
 	if (_storages[index].username == "") {
 		// options' Cloud tab believes Storage is connected once it has non-empty username
-		_storages[index].username = _("<syncing...>");
-		_storages[index].lastSyncDate = _("<right now>");
+		_storages[index].username = Common::convertFromU32String(_("<syncing...>"));
```
sev-, line 156: This looks to me as a double conversion. You're getting U32String from `_()`, right?

backends/dialogs/win32/win32-dialogs.cpp
```diff
@@ -134,7 +134,7 @@ Common::DialogManager::DialogResult Win32DialogManager::showFileBrowser(const ch
 		hr = dialog->SetTitle(str);
 		delete[] str;
 
-		str = Win32::ansiToUnicode(_("Choose"), Win32::getCurrentCharset());
+		str = Win32::ansiToUnicode(Common::convertFromU32String(_("Choose")).c_str(), Win32::getCurrentCharset());
```
sev-, line 137: So, you're converting from Unicode, to Unicode, and then back to Unicode? Please, rethink this.

backends/graphics/opengl/opengl-graphics.cpp
```diff
@@ -796,8 +796,8 @@ void OpenGLGraphicsManager::displayMessageOnOSD(const char *msg) {
 #ifdef USE_OSD
 void OpenGLGraphicsManager::osdMessageUpdateSurface() {
 	// Split up the lines.
-	Common::Array<Common::String> osdLines;
-	Common::StringTokenizer tokenizer(_osdMessageNextData, "\n");
+	Common::Array<Common::U32String> osdLines;
+	Common::StringTokenizer tokenizer(Common::convertFromU32String(_osdMessageNextData), "\n");
```
sev-, line 800: Will you lose the U32 data here during the converting? Why?

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -659,11 +659,8 @@ bool OpenGLSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 		endGFXTransaction();
 
 #ifdef USE_OSD
-		Common::String message = Common::String::format("%s: %s",
-			_("Stretch mode"),
-			_(stretchModes[index].description)
-			);
-		displayMessageOnOSD(message.c_str());
+		Common::U32String message = _("Stretch mode") + Common::U32String(": ") + _(stretchModes[index].description);
```
sev-, line 662: Why not use `Common::U32String::format`?

backends/graphics/sdl/sdl-graphics.cpp
```diff
@@ -435,7 +435,7 @@ Common::Keymap *SdlGraphicsManager::getKeymap() {
 	};
 
 	for (uint i = 0; i < ARRAYSIZE(filters); i++) {
-		act = new Action(filters[i].id, filters[i].description);
+		act = new Action(filters[i].id, _(filters[i].description));
```
sev-, line 438: Why this change? If you really need to translate description, this is a completely incorrect way of doing it and will break the translation subsystem.

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -2225,11 +2225,11 @@ void SurfaceSdlGraphicsManager::drawMouse() {
...
 
 #ifdef USE_OSD
-void SurfaceSdlGraphicsManager::displayMessageOnOSD(const char *msg) {
+void SurfaceSdlGraphicsManager::displayMessageOnOSD(const Common::U32String &msg) {
 	assert(_transactionMode == kTransactionNone);
-	assert(msg);
+	assert(!msg.empty());
 #ifdef USE_TTS
-	Common::String textToSay = msg;
+	Common::U32String textToSay = msg;
```
sev-, line 2232: Is the TTS now works with U32Strings?

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -2240,15 +2240,17 @@ void SurfaceSdlGraphicsManager::displayMessageOnOSD(const char *msg) {
...
-	Common::Array<Common::String> lines;
-	const char *ptr;
-	for (ptr = msg; *ptr; ++ptr) {
-		if (*ptr == '\n') {
-			lines.push_back(Common::String(msg, ptr - msg));
-			msg = ptr + 1;
+	Common::Array<Common::U32String> lines;
+	Common::U32String::const_iterator itr = msg.begin(), strLineItrBegin = msg.begin();
+
+	for ( ; itr != msg.end(); itr++) {
```
sev-, line 2246: Please reduce the scope of `itr` to the array only.

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -2466,13 +2468,12 @@ void SurfaceSdlGraphicsManager::handleScalerHotkeys(int scalefactor, int scalerT
 			g++;
 		}
 		if (newScalerName) {
-			const Common::String message = Common::String::format(
-				"%s %s\n%d x %d -> %d x %d",
-				_("Active graphics filter:"),
+			const Common::U32String message = Common::U32String::format(
+				_("Active graphics filter:") + Common::U32String("%s\n%d x %d -> %d x %d"),
```
sev-, line 2472: Why are you constructing the format string like this? Please revert to the original code.

backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -2499,20 +2500,18 @@ bool SurfaceSdlGraphicsManager::notifyEvent(const Common::Event &event) {
 			setFeatureState(OSystem::kFeatureAspectRatioCorrection, !_videoMode.aspectRatioCorrection);
 		endGFXTransaction();
 #ifdef USE_OSD
-		Common::String message;
+		Common::U32String message;
 		if (_videoMode.aspectRatioCorrection)
-			message = Common::String::format("%s\n%d x %d -> %d x %d",
-			                                 _("Enabled aspect ratio correction"),
+			message = Common::U32String::format(_("Enabled aspect ratio correction") + Common::U32String("\n%d x %d -> %d x %d"),
```
sev-, line 2505: This is where I am pausing my review.

ccawley2011:
It might be a good idea to split some of the more standalone changes like marking strings as translatable into a separate PR so that this one has fewer changes, making reviewing it slightly easier.

.travis.yml
aryanrawlani28: I left a note regarding this at the top of PR - 

> PS: I have also modified .travis.yml to build my branch. I plan to remove this when the PR is near merge.

travis builds are very helpful and help checking if any engine code breaks, i can just drop this commit when near-merge

engines/scumm/scumm.cpp
```diff
@@ -2759,7 +2758,10 @@ bool ScummEngine::startManiac() {
 		eventMan->pushEvent(event);
 		return true;
 	} else {
-		displayMessage(0, "%s", _("Usually, Maniac Mansion would start now. But for that to work, the game files for Maniac Mansion have to be in the 'Maniac' directory inside the Tentacle game directory, and the game has to be added to ScummVM."));
```
aryanrawlani28, line 2762: all this did, was go into displayMessage, use vsnprintf, and call MessageDialog. to make translations work, I just shifted it out and used u32::format

backends/dialogs/win32/win32-dialogs.cpp
aryanrawlani28: this was my bad! rethought about this, new code pushed and tested. thanks!

backends/graphics/opengl/opengl-graphics.cpp
aryanrawlani28: A U32Tokenizer has been implemented.

criezy:
There is some issue with this code, but I have not tracked the exact cause.
What I noticed initially is that I am getting a crash when opening the Add game dialog on macOS using French GUI.
```
frame #11: 0x0000000109c2492f scummvm`Common::Encoding::stringLength(string="Afficher les fichiers cachés뺾????????????????????????????????????????????????????????????????????????????????????????", encoding=0x00007fff5fbf9880) at encoding.cpp:465
frame #12: 0x000000010000d1ff scummvm`OSystem_SDL::convertEncoding(this=0x0000613000001a80, to="UTF-8", from="UTF-32LE", string="A", length=116) at sdl.cpp:763
frame #13: 0x0000000109c21960 scummvm`Common::Encoding::conversion(to=0x00007fff5fbfaa80, from=0x00007fff5fbfa4c0, string="A", length=116) at encoding.cpp:180
frame #14: 0x0000000109c20bca scummvm`Common::Encoding::convertWithTransliteration(to=0x00007fff5fbfaa80, from=0x00007fff5fbfaad0, string="A", length=116) at encoding.cpp:168
frame #15: 0x0000000109c20d6d scummvm`Common::Encoding::convert(to=0x00007fff5fbfaa80, from=0x00007fff5fbfaad0, string="A", size=116) at encoding.cpp:84
frame #16: 0x0000000109b83751 scummvm`Common::U32String::encode(this=0x00007fff5fbfad60, page=kUtf8) const at str-enc.cpp:422
frame #17: 0x0000000109316596 scummvm`::-[BrowserDialogPresenter showOpenPanel:](self=0x00006030009759a0, _cmd="showOpenPanel:", panel=0x00006150000ca100) at macosx-dialogs.mm:85
```

The code is calling U32String::encode() on a string coming from TransMan, but it appears to have an additional strange character at the end.

The po file contains the following text: `Afficher les fichiers cachés` with a length of 28 characters.
The _translations.dat_ file appears to be correct (the first line below is the hexadecimal content of the _translations.dat_ file, and the second line is the equivalent UTF-8 text):
```
41 66 66 69 63 68 65 72 20 6C 65 73 20 66 69 63 68 69 65 72 73 20 63 61 63 68 C3A9 73 00
A  f  f  i  c  h  e  r     l  e  s     f  i  c  h  i  e  r  s     c  a  c  h  é    s
```
However the U32String I get from TransMan has a length of 29 and not 28, and it has one additional strange character at the end. Below is the hexadecimal content of that string; notice the BEBE0000 after the 's' at the end:
```
00000041 00000066 00000066 00000069 00000063 00000068 00000065 00000072 00000020 0000006C 00000065 00000073 00000020 00000066 00000069 00000063 00000068 00000069 00000065 00000072 00000073 00000020 00000063 00000061 00000063 00000068 000000E9 00000073 BEBE0000 00000000
```

I have not investigated further where that issue came from, but it looks like there might be an issue with the UTF-8 decoding in TransMan when it reads the _translations.dat_ file and creates the U32String.

backends/text-to-speech/linux/linux-text-to-speech.h
```diff
@@ -59,7 +60,8 @@ class SpeechDispatcherManager : public Common::TextToSpeechManager {
 	SpeechDispatcherManager();
 	virtual ~SpeechDispatcherManager() override;
 
-	virtual bool say(Common::String str, Action action, Common::String charset = "") override;
+	virtual bool say(const Common::U32String &str, Action action, Common::String charset = "") override;
```
criezy, line 63: What is the charset argument for here? Surely the string is in UTF-32 since it is passed as a `Common::U32String`?

backends/text-to-speech/linux/linux-text-to-speech.h
```diff
@@ -59,7 +60,8 @@ class SpeechDispatcherManager : public Common::TextToSpeechManager {
...
+	virtual bool say(const Common::String &str, Action action, Common::String charset = "") override;
```
criezy, line 64: This one doesn't need to be virtual. It could be implemented just in the base class converting the string to UTF-32 and calling the one above.

backends/text-to-speech/linux/linux-text-to-speech.cpp
```diff
@@ -241,6 +226,10 @@ bool SpeechDispatcherManager::say(Common::String str, Action action, Common::Str
 	return false;
 }
 
+bool SpeechDispatcherManager::say(const Common::String &str, Action action, Common::String charset) {
+	return say(Common::U32String(str), action, charset);
```
criezy, line 230: That doesn't seem correct. The given `charset` is ignored to create the `U32String`. Doesn't it assume that this is ASCII then? This should be converted from the given `charset` to UTF-32 instead.

backends/text-to-speech/macosx/macosx-text-to-speech.mm
```diff
@@ -112,12 +113,16 @@ - (void)ignoreNextFinishedSpeaking:(BOOL)ignore {
...
 
-	_messageQueue.push(SpeechText(text, encoding));
+	_messageQueue.push(SpeechText(strToSpeak, encoding));
 	if (!isSpeaking())
 		startNextSpeech();
 	return true;
 }
 
+bool MacOSXTextToSpeechManager::say(const Common::String &text, Action action, Common::String encoding) {
+	return say(Common::U32String(text), action, encoding);
```
criezy, line 123: Why reimplement that function in `MacOSXTextToSpeechManager`? The implementation is identical to the one in the base class so I don't see the point.
(and it is identically incorrect - see my comment for the base class implementation)

backends/dialogs/macosx/macosx-dialogs.mm
```diff
@@ -128,22 +128,22 @@ - (IBAction) showHiddenFiles : (id) sender {
...
 
 	// Get current encoding
 #ifdef USE_TRANSLATION
-	CFStringRef encStr = CFStringCreateWithCString(NULL, TransMan.getCurrentCharset().c_str(), kCFStringEncodingASCII);
+	CFStringRef encStr = CFStringCreateWithCString(NULL, TransMan.getCurrentCharset().c_str() == "UTF-32" ? "UTF-8" : "ASCII", kCFStringEncodingASCII);
 	CFStringEncoding stringEncoding = CFStringConvertIANACharSetNameToEncoding(encStr);
 	CFRelease(encStr);
 #else
 	CFStringEncoding stringEncoding = kCFStringEncodingASCII;
 #endif
```
criezy, line 142: This whole block is overly complicated. You are using `U32String` and calling `encode()`, so we know the text is in UTF-8. And ASCII is also valid UTF-8 so there is no need to make a distinction.

Thus you could just be using
`CFStringEncoding stringEncoding = kCFStringEncodingUTF8;`

backends/text-to-speech/macosx/macosx-text-to-speech.mm
```diff
@@ -112,12 +113,16 @@ - (void)ignoreNextFinishedSpeaking:(BOOL)ignore {
 #endif
 	}
 
-	_messageQueue.push(SpeechText(text, encoding));
+	_messageQueue.push(SpeechText(strToSpeak, encoding));
```
criezy, line 116: That is incorrect.
You have
```
const Common::U32String &text
Common::String strToSpeak = text.encode();
```
So `strToSpeak` is UTF-8 and cannot be anything else.

But you also use
```
	if (encoding.empty()) {
#ifdef USE_TRANSLATION
		encoding = TransMan.getCurrentCharset();
#endif
	}
```
So UTF-32 encoding. There is a discrepancy between the two.

I think you can just remove the encoding field in `SpeechText`, and in `startNextSpeech` assume `kCFStringEncodingUTF8`.

backends/graphics/openglsdl/openglsdl-graphics.cpp
aryanrawlani28: uhh, yes, you are absolutely correct. This should have been obvious, I don't remember what I must have been thinking back then! Will change it accordingly, a `%S` sounds good.

backends/platform/ios7/ios7_osys_misc.mm
```diff
@@ -48,37 +48,45 @@ static inline void execute_on_main_thread_async(void (^block)(void)) {
...
+#ifdef SCUMM_LITTLE_ENDIAN
+	NSStringEncoding stringEncoding = NSUTF32LittleEndianStringEncoding;
 #else
-	NSStringEncoding encoding = NSISOLatin1StringEncoding;
+	NSStringEncoding stringEncoding = NSUTF32BigEndianStringEncoding;
 #endif
 	UIPasteboard *pb = [UIPasteboard generalPasteboard];
-	[pb setString:[NSString stringWithCString:text.c_str() encoding:encoding]];
-	return true;
+	return [pb setString:[[NSString alloc] initWithBytes:text.c_str() length:4*text.size() encoding: stringEncoding] forType:NSStringPboardType];
```
criezy, line 89: That line will cause a memory leak as any objects allocated also need to be released. I missed that issue when looking at the macOS code a few days ago. I will fix both.

engines/glk/adrift/os_glk.cpp
```diff
@@ -2811,7 +2811,7 @@ static int gsc_startup_code(Common::SeekableReadStream *game_stream, int restore
 		 * Display a brief loading game message; here we have to use a timeout
 		 * to ensure that the text is flushed to Glk.
 		 */
-		g_vm->glk_put_string(_("Loading game...\n"));
+		g_vm->glk_put_string("Loading game...\n");
```
criezy, line 2814: This code doesn't even work in current master (non-ASCII characters are not displayed properly). However the GLK engine actually handles unicode text, and there is a `glk_put_string_uni(const uint32 *)` function for example. So with the change to use u32 strings for the translation, we could actually make this code work I think.

engines/glk/scott/scott.cpp
criezy: I don't think using UTF-8 will work correctly. To make this work I think we would need to keep U32Strings here and add a display() function that takes `const uint32 *` instead of `const char *` and calls `glk_put_string_stream_uni()` instead of `glk_put_string_stream()`.

engines/hugo/dialogs.cpp
```diff
@@ -56,15 +56,15 @@ void TopMenu::init() {
...
+	_whatButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("What is it?"),				kCmdWhat);
+	_musicButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Music"),						kCmdMusic);
+	_soundFXButton =	new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Sound FX"),					kCmdSoundFX);
+	_saveButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Save game"),					kCmdSave);
+	_loadButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Load game"),					kCmdLoad);
+	_recallButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Recall last command"),		kCmdRecall);
+	_turboButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Turbo"),						kCmdTurbo);
+	_lookButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Description of the scene"),	kCmdLook);
+	_inventButton =		new GUI::PicButtonWidget(this,	x, y, kButtonWidth, kButtonHeight,	Common::U32String("Inventory"),					kCmdInvent);
```
criezy, line 67: Are those lines supposed to be aligned? They are not aligned for me as I probably have a different setting than you for the tab size (e.g 2 characters, 4 characters, or 8 characters). Aligning those kind of line works better if you use spaces rather than tabs.

engines/scumm/dialogs.cpp
```diff
@@ -609,15 +613,15 @@ void SubtitleSettingsDialog::cycleValue() {
 		_value = 0;
 
 	if (_value == 1 && g_system->getOverlayWidth() <= 320)
-		setInfoText(_sc("Speech & Subs", "lowres"));
+		setInfoText(Common::U32String(_sc("Speech & Subs", "lowres")));
```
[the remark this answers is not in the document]
criezy: I suspect this is a bug. I think it was meant to use `_c(...)` and not `_sc(...)`

backends/graphics/sdl/sdl-graphics.cpp
criezy: Actually your change is correct here. The strings are marked for translation using the `_s()` macro above in the `filters` array definition. However in current master they are not translated (you can see that in the Options dialog) and this is a bug. With your change they are now properly translated in the GUI.

engines/bbvs/dialogs.cpp
```diff
@@ -86,11 +86,11 @@ MainMenu::~MainMenu() {
...
-	_buttons[0] = new GUI::ButtonWidget(this, 0, 0, 1, 1, "", 0, 0);
-	_buttons[1] = new GUI::ButtonWidget(this, 0, 0, 1, 1, "", 0, 0);
-	_buttons[2] = new GUI::ButtonWidget(this, 0, 0, 1, 1, "", 0, 0);
-	_buttons[3] = new GUI::ButtonWidget(this, 0, 0, 1, 1, "", 0, 0);
-	_buttons[4] = new GUI::ButtonWidget(this, 0, 0, 1, 1, "", 0, 0);
+	_buttons[0] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
+	_buttons[1] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
+	_buttons[2] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
+	_buttons[3] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
+	_buttons[4] = new GUI::ButtonWidget(this, 0, 0, 1, 1, Common::U32String(""), Common::U32String(""), 0);
```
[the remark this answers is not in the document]
aryanrawlani28: I had some changes like those before with my modifications, but I received a comment saying that I shouldn't unnecessarily change stuff (like using default args), so I was avoiding things like these. Since you've pointed it out though, I will clean it up.

backends/graphics/sdl/sdl-graphics.cpp
aryanrawlani28: Glad that you've cleared it up. I saw by examples before that strings can use `_s` or `_sc`, and then when I actually want to use the translation, I can use `_(...)` like normal, so I was confused as to why this was wrong. Thanks, Resolving this convo.

engines/glk/scott/scott.cpp
```diff
@@ -385,13 +398,18 @@ void Scott::output(const Common::String &a) {
...
+
 void Scott::outputNumber(int a) {
 	display(_bottomWindow, "%d", a);
 }
 
 void Scott::look(void) {
-	const char *const ExitNames[6] = {
-		_("North").encode().c_str(), _("South").encode().c_str(), _("East").encode().c_str(), _("West").encode().c_str(), _("Up").encode().c_str(), _("Down").encode().c_str()
+	const uint32 *const ExitNames[6] = {
+		_("North").c_str(), _("South").c_str(), _("East").c_str(), _("West").c_str(), _("Up").c_str(), _("Down").c_str()
```
criezy, line 412: You have a memory issue here.
TranslationManager::getTranslation() used to return a `const char *` pointer to a data in its translation messages array. But now it returns a `U32String` by copy. And here since you don't keep this `U32String` after getting the pointer to its data, it is destroyed. And you are thus keeping pointers to freed memory in the `ExitNames` array.

This is a similar issue to the one I highlighted in gtk-dialogs.cpp where `U32String::encode()` returns a `String` by copy and you are keeping a pointer to its data without keeping the `String` itself.

aryanrawlani28: Ah yes, makes sense. I will change it accordingly in a little while. Also, regarding your new commits, I will sweep the backends directory and search for `_(...)` usage, so I can catch any more compilation errors.

criezy:
With the two commits for IOS7 and IPHONE I was able to compile and test this branch on an old iPad, and it works well with no noticeable performance issues.

@aryanrawlani28 There will be similar compilation errors in the various backends. For example `git grep "_(" backends/platform/` highlighted many issues with the tooltip argument in code like this:
```
backends/platform/3ds/options-dialog.cpp:.   new GUI::ButtonWidget(this, 120, 180, 72, 16, _("~C~lose"), 0, GUI::kCloseCmd);
```

You might want to take a look to limit the number of errors and fix we have to make post-merge. I expect builedbot to break for some platforms and highlight those issues, but the more issues are fixed beforehand, the faster it will be to fix the remaining ones.

engines/scumm/dialogs.cpp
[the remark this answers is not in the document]
aryanrawlani28: Thanks for the verification. Have changed it.

backends/dialogs/gtk/gtk-dialogs.cpp
```diff
@@ -35,25 +35,19 @@
...
 
 	// Convert labels to UTF-8
-	char *utf8Title = utf8.convert(title, strlen(title));
-	Common::String choose = _("Choose");
-	char *utf8Choose = utf8.convert(choose.c_str(), choose.size());
-	Common::String cancel = _("Cancel");
-	char* utf8Cancel = utf8.convert(cancel.c_str(), cancel.size());
+	char *utf8Title = (char *)title.encode().c_str();
+	char *utf8Choose = (char *)choose.encode().c_str();
+	char* utf8Cancel = (char *)cancel.encode().c_str();
```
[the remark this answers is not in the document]
aryanrawlani28: Even though it makes sense that a copy is generated by `encode()` and this shouldn't work, it still did. I manually browsed the memory and the titles were present. Still, to be on the safe side, I have changed it around.

Also, this helped me point out another obvious wrong usage, which was a memory corruption because I was freeing memory that wasn't mine. Have fixed them both in a new commit.

engines/glk/scott/scott.cpp
```diff
@@ -408,8 +408,8 @@ void Scott::outputNumber(int a) {
 }
 
 void Scott::look(void) {
-	const uint32 *const ExitNames[6] = {
-		_("North").c_str(), _("South").c_str(), _("East").c_str(), _("West").c_str(), _("Up").c_str(), _("Down").c_str()
+	const Common::U32String ExitNames[6] = {
+		_("North"), _("South"), _("East"), _("West"), _("Up"), _("Down")
 	};
```
criezy, line 413: A possibly better alternative to this would have been to keep using an array of `char *` and only translate them when actually used (note the use of `_s()` instead of `_()` in the array definition, and then `_()` being used when calling `display()`):
```c++
const char *const ExitNames[6] = {
		_s("North"), _s("South"), _s("East"), _s("West"), _s("Up"), _s("Down")
	};
...
	display(_topWindow, Common::U32String("%S"), _(ExitNames[ct]).c_str());
```

aryanrawlani28: That is indeed a better alternative. It also closely matches the original code. Have changed it.

gui/filebrowser-dialog.h
```diff
@@ -45,7 +45,7 @@ class FileBrowserDialog : public Dialog {
 
 	void handleCommand(CommandSender *sender, uint32 cmd, uint32 data) override;
 
-	const char *getResult() { return Dialog::getResult() ? _fileName->getEditString().c_str() : nullptr; }
+	const char *getResult() { return Dialog::getResult() ? _fileName->getEditString().encode().c_str() : nullptr; }
```
criezy, line 48: You have another memory issue here.
`_fileName->getEditString()` returns a reference to to the string stored in the `_fileName` widget, so getting its `c_str()` is fine (as long as you don't use it after the widget has been deleted.
However the addition of `encode()` means it is now creating a temporary `Common::String` and you are returning a pointer to the content of this temporary string. And by the time the call of this function can use the pointer, the string will already have been destroyed.

Maybe it should return a `Common::String` instead.

criezy (changes requested):
I have now reviewed all the changes, although I will do another quick pass this weekend.

Once you have fixed the remaining issues, please also rebase on current master and fix the merge conflict if any.

gui/message.cpp
```diff
@@ -83,18 +83,28 @@ MessageDialog::MessageDialog(const Common::String &message, const char *defaultB
...
+		new ButtonWidget(this, cancelButtonPos, _h - buttonHeight - 8, buttonWidth, buttonHeight, altButton, Common::U32String(""), kCancelCmd, Common::ASCII_ESCAPE);
+	}
+}
+
+MessageDialog::MessageDialog(const char *message, const char *defaultButton, const char *altButton, Graphics::TextAlign alignment)
+	: Dialog(30, 20, 260, 124) {
 
-	if (altButton)
-		new ButtonWidget(this, cancelButtonPos, _h - buttonHeight - 8, buttonWidth, buttonHeight, altButton, nullptr, kCancelCmd, Common::ASCII_ESCAPE);	// Cancel dialog
+	MessageDialog(Common::U32String(message), Common::U32String(defaultButton), Common::U32String(altButton), alignment);
```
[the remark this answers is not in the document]
criezy: I tried testbed quickly, and indeed the MessageDialog there is empty, which confirms my analysis:
image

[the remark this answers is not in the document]
aryanrawlani28: Yes, you are correct. It of course won't do the calls for the `this` object, and would rather create a new temporary object that would immediately be destroyed. I think I've made this mistake in a few other places, I will check for all of them and correct accordingly.

gui/filebrowser-dialog.h
aryanrawlani28: I checked for usage of this function
(Examples: `engines/director/lingo/xlibs/fileio.cpp:140`, `engines/bbvs/minigames/bbairguitar.cpp:1220:` 
and it seems like they use a `Common::String` for assignment anyway. So, making the return value a `const char *` is useless anyway, since it promotes to a `Common::String`. So, your suggestion is correct, and I will change this to return a string instead.

engines/drascula/drascula.cpp
```diff
@@ -928,9 +928,9 @@ bool DrasculaEngine::loadDrasculaDat() {
...
 
 	if (!in.isOpen()) {
-		Common::String errorMessage = Common::String::format(_("Unable to locate the '%s' engine data file."), filename.c_str());
+		Common::U32String errorMessage = Common::U32String::format(_("Unable to locate the '%s' engine data file."), filename.c_str());
 		GUIErrorMessage(errorMessage);
-		warning("%s", errorMessage.c_str());
+		warning("Unable to locate the '%s' engine data file.", filename.c_str());
 
 		return false;
 	}
```
aryanrawlani28, line 936: Should I do something like below to avoid copy paste? (here and some other files)
```suggestion
		if (!in.isOpen()) {
		Common::String msg = _s("Unable to locate the '%s' engine data file.");

		Common::U32String errorMessage = Common::U32String::format(_(msg), filename.c_str());
		GUIErrorMessage(errorMessage);
		warning(msg.c_str(), filename.c_str());

		return false;
	}
```

criezy: I saw this yesterday, and I was wondering if it should be changed. The current code does not shock me, but I think your proposed change is a good idea.

criezy (approved):
I am quite happy with the state of the PR now.

There is probably some legacy code that is no longer used and can be removed (e.g. maybe the text drawing code for a `Common::String`) but I would prefer to handle this separately. There is already many changes in this PR and we don't need to add more.

With the amount of changes I am pretty sure I missed some things, and I expect that the compilation will break for some ports. But buildbot is here for that and will hopefully let us know what breaks (and the error should be easy to fix).

I think what will need to be done when we merge this:

1. Commit and push any pending changes from weblate.
2. Lock weblate.
3. Rebase this branch and convert the latest po files to UTF8.
4. Generate the translations.dat file and commit.
5. Merge the PR.
6. Unlock weblate.
7. Update the wiki (update description of translations.dat format, instructions to translator about not using UTF-8 encoding needs to be removed).
8. Monitor buildbot and fix compilation errors.

backends/dialogs/gtk/gtk-dialogs.cpp
```diff
@@ -35,25 +35,20 @@
...
-	char *utf8Choose = utf8.convert(choose.c_str(), choose.size());
-	Common::String cancel = _("Cancel");
-	char* utf8Cancel = utf8.convert(cancel.c_str(), cancel.size());
+	Common::String titleStr = title.encode();
+	Common::String choose = _("Choose").encode();
+	Common::String cancel = _("Cancel").encode();
+
+	char *utf8Title = (char *)(titleStr.c_str());
+	char *utf8Choose = (char *)(choose.c_str());
+	char *utf8Cancel = (char *)(cancel.c_str());
```
aryanrawlani28, line 51: GTK documentation for dialogs (https://developer.gnome.org/gtk3/stable/GtkFileChooserDialog.html) shows that the functions that use these titles are of type `const gchar*`, and `gchar` is just a typedef for `char`. So, I'm pretty sure we could further remove these lines, and below (where we actually use strings) we can pass `str.c_str()` and everything should work fine. Just wanted to get a second opinion before I push this out though.

criezy: When I saw those lines I assumed that the `gtk_file_chooser_dialog_new()` function needed `char *` pointers, while `String::c_str()` returns a `const char *`. But you are right and indeed it takes a `const gchar *`. So I have no idea why those casts to `char *` are here. Maybe the `gtk_file_chooser_dialog_new()` was changed and in the past it was indeed taking non-const `gchar *` strings? But even in the gtk2 documentation (https://developer.gnome.org/gtk2/stable/GtkFileChooserDialog.html) it takes const ones.
So I think you are right and this code can be simplified further.

criezy:
The current plan is to merge this pull request after we have created the branch for the 2.2.0 ScummVM release as it seems risky to include all those changes just before a release. According to the current release schedule this might happen as soon as this Sunday.

backends/dialogs/gtk/gtk-dialogs.cpp
aryanrawlani28: Thanks for the response. I have changed it.

aryanrawlani28:
I agree that it would indeed be risky to merge before a release. It's best for later.
The plan criezy has suggested is perfect in my opinion, just 2 minor points more:
- There's a big red box that says not to use UTF-8 on the weblate, that should be replaced with a green box or something, saying you should use UTF-8.
- https://wiki.scummvm.org/index.php?title=HOWTO-Translate_ScummVM_GUI#Generate_Translations_Data_file_outside_of_ScummVM_code_repository has `create_translations` binaries, which will be outdated after merge.

gui/unknown-game-dialog.cpp
```diff
@@ -122,7 +122,7 @@ void UnknownGameDialog::rebuild() {
 Common::U32String UnknownGameDialog::encodeUrlString(const Common::U32String &string) {
 	Common::U32String encoded;
 	for (uint i = 0 ; i < string.size() ; ++i) {
-		char c = string[i];
+		uint32 c = string[i];
```
criezy, line 125: Encoding non-ASCII characters that way probably doesn't work? I think it needs to be converted to UTF-8 first.

aryanrawlani28: `Common::U32String report = generateUnknownGameReport(_detectedGame, false, false);`
I didn't think it mattered because these strings are ASCII only (the above function, second argument is for `translate`, and since it's `false`, the returned string is ascii (https://github.com/aryanrawlani28/scummvm/blob/3fe5e20b1482a7b2e84c6489eb65b284418e9a53/engines/game.cpp#L168)
I changed it just to match consistency, I'll change the function to accept a `Common::String` instead.

sev-:
MERGING!

PR #2355, opened 2020-06-30
DIRECTOR: LINGO: implement scummvmAssert

rvanlaar:
scummvmAssert is a scummvm specific Lingo function.
Useful for getting feedback on lingo code tests without
having to check if the output is as expected.

Example:

    set x = True
    scummvmAssert(x)    // will pass

    set x = False
    scummvmAssert(x)    // will fail

The core of the implementation is: `assert(d.asInt() == 1)`.
It's defined as a Lingo builtin function for D2.

strings.lingo has been modified to leverage the new scummvmAssert.


sev-:
Good work, thank you!

PR #2359, opened 2020-07-05
XEEN: Fix "remove X to equip Y" display issue + minor fixes

FrostyHitoshura:


dreammaster:
Looks good. Thanks for identifying and fixing the problem

PR #2361, opened 2020-07-08
DOXYGEN: Add doxygen groups to header files in the common folder

b-gent:
All the header files from the ``common`` folder now have doxygen groups added to them. Thanks to that, the usability of API doc is greatly improved and crucial information can be found. Compare:
Old:
image
New:
image

### DIFF
--- a/common/array.h
+++ b/common/array.h
@@ -31,6 +31,16 @@
 namespace Common {
 
 /**
+ * @defgroup common_array Arrays
+ * @ingroup common
+ *
+ * @brief  Functions for working on arrays.
+ *
+ * @{
+ */
+
+
+/**
  * This class implements a dynamically sized container, which
  * can be accessed similar to a regular C++ array. Accessing
  * elements is performed in constant time (like with plain arrays).
@@ -455,6 +465,8 @@ private:
 
 	Comparator _comparator;
 };
+
+/** @} */
 
 } // End of namespace Common
 
--- a/common/md5.h
+++ b/common/md5.h
@@ -27,6 +27,15 @@
 
 namespace Common {
 
+/**
+ * @defgroup common_md5 MD5 checksum
+ * @ingroup common
+ *
+ * @brief API for computing the MD5 checksum.
+ *
+ * @{
+ */
+
 class ReadStream;
 class String;
 
@@ -53,6 +62,8 @@ bool computeStreamMD5(ReadStream &stream, uint8 digest[16], uint32 length = 0);
  * @return the MD5 as a hex string on success, and an empty string if an error occurred
  */
 String computeStreamMD5AsString(ReadStream &stream, uint32 length = 0);
+
+/** @} */
 
 } // End of namespace Common
 
--- a/common/memory.h
+++ b/common/memory.h
@@ -28,6 +28,14 @@
 namespace Common {
 
 /**
+ * @defgroup common_memory Memory
+ * @ingroup common
+ *
+ * @brief Functions for managing the memory.
+ * @{
+ */
+
+/**
  * Copies data from the range [first, last) to [dst, dst + (last - first)).
  * It requires the range [dst, dst + (last - first)) to be valid and
  * uninitialized.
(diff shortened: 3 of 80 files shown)


b-gent:
@sev-

criezy:
> 1. I couldn't find the doxygen config file used for your current API doc anywhere in the repository 

You can find the current doxygen config file in https://github.com/scummvm/scummvm-sites/tree/doxygen

And indeed there is a lot of work needed to organise  our doxygen documentation and make it usable. This looks like a good start. Using groups is a good idea and the main page could for example be improved with some basic information and link to https://wiki.scummvm.org/index.php?title=Developer_Central.

Mataniko:
Thanks for the PR, Looking forward to seeing your application!

b-gent:
> > 1. I couldn't find the doxygen config file used for your current API doc anywhere in the repository
> 
> You can find the current doxygen config file in https://github.com/scummvm/scummvm-sites/tree/doxygen

I wasn't aware of the scummvm-sites fork, thanks :)

> And indeed there is a lot of work needed to organise our doxygen documentation and make it usable. This looks like a good start. Using groups is a good idea and the main page could for example be improved with some basic information and link to https://wiki.scummvm.org/index.php?title=Developer_Central.

Yes, and not only the main page, I bet there is much more 'conceptual' documentation that could be added to the doxygen docset that would explain how the API works. The empty placeholder pages that are there right now that are supposed to explain  how to implenent a plugin are an example of this.

common/array.h
```diff
@@ -30,6 +30,16 @@
 
 namespace Common {
 
+/**
+ * @defgroup common_array Arrays
+ * @ingroup common
+ *
+ * @brief  Functions for working on arrays.
```
Mataniko, line 37: these are functions to replace some std arrays

common/md5.h
```diff
@@ -27,6 +27,15 @@
 
 namespace Common {
 
+/**
+ * @defgroup common_md5 MD5 checksum
+ * @ingroup common
+ *
+ * @brief API for computing the MD5 checksum.
```
Mataniko, line 34: This is less of an API and more of a method/class - @sev- WDYT on terminology in general

common/memory.h
```diff
@@ -27,6 +27,14 @@
 
 namespace Common {
 
+/**
+ * @defgroup common_memory Memory
+ * @ingroup common
+ *
+ * @brief Functions for managing the memory.
```
Mataniko, line 34: Same as earlier - we should have consistent terminology - functions, api, class, methods

Mataniko:
Overall looks like a massive improvement - left a couple comments on terminology that we should standardize before proceeding

common/md5.h
b-gent: That's a good point about having consistent terminology. As I see it, 'API' is a superset of whatever we might have in these header files - classes, methods, functions, structures etc. If you agree with this, then probably using the term 'API' is safest and most future-proof.

common/memory.h
b-gent: I guess in this case the most accurate description is 'Class templates for managing the memory' ?

common/array.h
b-gent: I'll correct this (as soon as we agree on terminology).

b-gent:
I would like to get this in as soon as possible. I'm already making additional changes to these files on another branch and I would like to avoid some crazy merge conflicts.

b-gent:
The output is now available for preview on https://b-gent.github.io/PR2361/group__common.html

@sev , it would be great if you could have a look at the comments here and the PR as a whole.

common/array.h
criezy: Actually these are classes and not functions.

common/md5.h
criezy: In this specific case I would probably have used `Functions` if I had written the documentation. But API is also fine with me. And indeed we need consistency.

common/memory.h
criezy: Referring to classes here seems confusing to me. They look like `Template functions`.

sev-:
Great work, merging

PR #2362, opened 2020-07-09
MIDI/GROOVIE: The 7th Guest MIDI fixes

NMIError:
This PR contains several fixes for MT-32 and GM support for The 7th Guest:

- Fixed MT-32 support. This had several problems:
  - XMIDI custom timbres were sent after starting playback, so this overlapped
with the start of the track. This would cause wrong instruments and bad
timing. Necessary delays between the SysEx messages were generated using
OSystem::delayMillis, which caused more timing issues. Finally, this
functionality was refactored to use an interface rather than a callback.
XMIDI Final Data controller was incorrectly implemented and did not send
the final data byte of the SysEx.
  - GM and MT-32 initialization is now performed when loading a game from the
launcher.
- MIDI music is now paused when the ScummVM menu is open, in order to keep the
music in sync with cutscenes.
- Removed duplicate and unnecessary MIDI messages for better performance on
hardware MIDI modules. This includes duplicate All Notes Off messages and
messages on unused channels when stopping playback, and duplicate volume
messages and volume messages on unused channels during fade-outs.
- Looping XMIDI music could overflow the MIDI parser tracker after a long time
(bugs #6275 and #4354). This should be fixed.
- The Groovie engine had its own implementation of the Miles AIL functionality,
which was later implemented in the generic Miles drivers. I've removed this
dead code.


sev-:
Very good code quality. Thanks, merging.

PR #2248, opened 2020-05-13
Add support for MSVC AArch64 and vcpkg

janisozaur:
[vcpkg](https://github.com/microsoft/vcpkg) is a C++ Library Manager, primarily for Windows. It's fairly popular as Windows is missing a proper package manager built into the system. Vcpkg also provides support for targetting different architectures supported by MSVC: x86, AMD64, arm64.

Last year I built ScummVM for Windows on Arm as a one-off binary and created relevant ports in vcpkg upstream. The last has finally landed and I want to contribute a more sustainable way of producing vcpkg-based ScummVM builds and extend support for MSVC arm64.

My initial plan was to use cmake, the cross-platform meta build system, but ScummVM seems to take the cross-platform out of it, so instead I restorted to using the msbuild-based system.

I tried keeping with the code style, but given how many contradicting formatters are provided, I used none.

I tried keeping current libraries working without any changes, but haven't tested that. I had to add a new option to `create_project`: `--use-canonical-lib-names` to map the libraries' names from how they're called in `SCUMMVM_LIBS` to how they are usually called when built with respective build systems. The bulk of work to support vcpkg went into extending the expected paths for includes and libraries.

As a non-Windows user I may have mislabelled something or violated some convention, but the builds were tested and all work.

One issue I encountered was I couldn't get the symbols for `FLAC__StreamDecoderStateString` and `FLAC__StreamDecoderErrorStatusString` to link properly, so I removed them in hope this is a rarely reported issue and the error code will suffice.

The builds are by default using dynamic linking, unlike the binaries usually produced by the project, that's primarily to accommodate one of the libs' (I think it was freetype, but I'm not sure) dependency on glib, which needs to be compiled to a dynamic library and it should be relatively easy to change it to static linkage, if said library was to be dropped from build. I don't know about mix-n-match situation, but as this is not intended to replace the current builds, I don't consider this a problem.

~~On top of it all sits a commit a bit out of scope for this PR, enabling GitHub Actions-based CI exercising the vcpkg-provided libs. There's caching of vcpkg libraries implemented inside the action, however my tests suggest it works for x86 and arm64 only, with AMD64 job rebuilding them over and over again. I don't know why that is and would rather see the job removed instead of investigating why. The jobs all produce relevant artifacts: libs as produced by vcpkg, scummvm binaries with all used libs, debug symbols.~~

CI stuff moved out to https://github.com/scummvm/scummvm/pull/2369

Finally, a screenshot:

teenagent

PR #2369, opened 2020-07-15
CI: Add GitHub Actions-based CI to excercise vcpkg and MSVC AArch64

janisozaur:
CI integration extracted from https://github.com/scummvm/scummvm/pull/2248 which uses GitHub Actions to provide builds.

Not expected to work until #2248 gets merged, but at the same time #2248 won't get verified with GitHub Actions due to missing workflow definitions in the upstream.

There's caching of vcpkg libraries implemented inside the action, however my tests suggest it works for x86 and arm64 only, with AMD64 job rebuilding them over and over again. I don't know why that is and would rather see the job removed instead of investigating why. The jobs all produce relevant artifacts: libs as produced by vcpkg, scummvm binaries with all used libs, debug symbols.

You can see how it works over at https://github.com/janisozaur/scummvm/actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,86 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  windows:
+    name: Windows
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: win32
+            triplet: x86-windows
+            arch: x86
+            vcpkgPackages: 'curl faad2 fluidsynth freetype fribidi libflac libjpeg-turbo libmad libmpeg2 libogg libpng libtheora libvorbis sdl2 sdl2-net zlib'
+            useNasm: 'true'
+          - platform: x64
+            arch: x64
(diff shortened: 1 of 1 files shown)


sev-:
Thanks! Let's see how it works. Merging.

sev-:
@janisozaur This thing timed out after 6 hours: https://github.com/scummvm/scummvm/runs/884711089

janisozaur:
Yes, I noticed. Cursory look at logs suggests the timeout is indeed happening while preparing packages. I can't verify this right now, but I think there are a few more packages to target x86 and amd64 than there are four arm64?

The caching works only after first successful build and can be possibly build up on - first build a subset of packages and only then use cached ones and add new ones. That pollutes the history a bit and doesn't scale well.

Vcpkg can export compiled libraries, maybe provide a multistage approach where each stages builds a subset of packages until all are built?

I can't really look into this right now and would like to discuss the options first, so for now I guess you could drop the builds that timeout?

But hey, arm64 works fine, which was my ultimate goal, so yay!

PR #2370, opened 2020-07-16
SCI: Add more detection entries for Hebrew SQ3

ZvikaZ:


ZvikaZ:
Hi.
I've discussed this with @sluicebox , and he explained to me that it's not a good idea.
Different SQ3 releases have inconsistent text resources numbers, and some of the strings  are hard coded in the scripts files.
Therefore, the translation should match a specific detection entry.

Currently it matches the GOG release. It's best to leave it as is.
Therefore, closing.

PR #2373, opened 2020-07-21
Support 4-bit windows cursors

phcoder:


henke37:
You missed to update the comment right above the if statement that checks for support. Might as well remove the maintenance hazard there.

phcoder:
> You missed to update the comment right above the if statement that checks for support. Might as well remove the maintenance hazard there.

Updated

sev-:
Thanks, do you have any game or target which uses it?

phcoder:
> Thanks, do you have any game or target which uses it?

Hades Challenge by Disney's Interactive

PR #2374, opened 2020-07-21
DIRECTOR: LINGO: Implement b_label

rvanlaar:
b_label converts a label into a frameId.
There can be multiple labels of the same name.
It was found that the first label is the one director uses.


sev-:
I've had it implemented and it also should accept integers.

PR #2375, opened 2020-07-22
SCI: Make menus BiDi

ZvikaZ:
Adding BiDi function call for menus if the language is RTL.
Before this fix, the menu names were written reversed in Hebrew.


sluicebox:
Does this affect currently supported games or fan-translations, or is this for a translation that's in development?

ZvikaZ:
> Does this affect currently supported games or fan-translations, or is this for a translation that's in development?

There are currently only two SCI RTL titles:
- Torin, which is finished, and not affected by this PR
- SQ3, which is work in progress, and this PR fixes a problem found in it

ZvikaZ:
@sluicebox , what do you say?
It's a simple and short PR, I don't think it can make any harm...

PR #2376, opened 2020-07-22
GUI: Small patches.

aryanrawlani28:
- Fix a small issue of 2 chars being wobbled together if pressing home/end in EditableWidgets.
- Fix warnings being repeatedly printed in the console, when using an RTL language with FriBiDi disabled.


sev-:
Thanks. Merging.

PR #2377, opened 2020-07-23
GUI: Accurately reflect paddings in the RTL widgets.

aryanrawlani28:
A while back, https://github.com/scummvm/scummvm/pull/2265 added RTL support to GUI. In this PR, we modified many files, including the theme files to have new RTL-based widgets.

When the paddings were set, some place's I had mistakenly changed things which I now realize were wrong.

In the wiki, there was a line which stated:
> padding [int, int, int, int]: Pixel padding for Left, Right, Top and Bottom sides, respectively.

which I first thought was the case, and made the mistake. The difference is ever so subtle (highlighted in the pictures below), but the new changes should be now perfectly accurate and reflect/mirror the normal GUI padding (only effects PopUp and DropDownButton).

---------------------------------

**English language:**

English - Stable

----------------------------------

**Current master branch, Hebrew language:**

Hebrew - Previous

---------------------------------------

**This PR, Hebrew language:**

Hebrew - New


sev-:
Thanks!

PR #2378, opened 2020-07-24
DIRECTOR: Extend --start-movie to start at frame

rvanlaar:
A frame number can be given to --start-movie and it will start
at that frame.
Examples:
    --start-movie=movie.dir:45  # start movie.dir at frame 45
    --start-movie=:23           # start the default movie frame 23
    --start-movie=movie.dir     # start movie.dir as usual


sev-:
Thanks!

PR #2379, opened 2020-07-26
DIRECTOR: LINGO: Add FlushXObject

rvanlaar:
FlushXObj is a Mac only XObject to call the underlying FlushEvents function
from the Macintosh Toolbox. Its purpose is to flush, i.e. remove, events
that happened while loading code.

Implemented as a no-op, there's no need to flush events because:
 - ScummVM handles them and
 - computers were slower and queued events when the program was loading.

More information about the Toolbox and the flush events can be found here:
 https://en.wikipedia.org/wiki/Macintosh_Toolbox
 https://developer.apple.com/legacy/library/documentation/mac/pdf/MacintoshToolboxEssentials.pdf


sev-:
Closing this in favour of #2380 which contains bot XPal and XFlush

PR #2380, opened 2020-07-26
DIRECTOR: LINGO: STUB XObject FixPalette

rvanlaar:
FixPalette is an XObject for Mac that patches the palette.
According to the docs it's either needed after an 8 palette switch
or when using QuickTime movies with more than one custome palette.
Afterwards the colors will be wrong.
    
The mPatchIt function that correctly sets the palette is stubbed.
    
Note: It's at the moment of writing this commit unknown if the
bug will show itself on the mac versions of games.


sev-:
Great work! There are only a couple of minor issues with indentation, which I'll fix after the merge.

PR #2381, opened 2020-07-26
DIRECTOR: LINGO: Add scripts to the correct cast

rvanlaar:
Repair the castIds of script when the scripts are loaded.
The castIds that are defined in the scripts can be wrong.
A better way to link scripts with casts is via the scriptId.

To achieve this casts need to be fetched by their scriptId.
This commit implements the necessary getCastMemberByScript functions
to enable fetching a cast by it's scriptId.


sev-:
Thanks a lot!

PR #2383, opened 2020-07-27
AUDIO: SoundFont2 creation support

yuv422:
This PR adds support for converting PS1 VAB sample data into soundfont2 format. This can then be used by fluidsynth to play music from playstation games. 

These include Blazing Dragons and tinsel engines.

I've taken the VAB to SF2 logic from the VGMTrans project. I've removed all  C++11 code and converted everything over to the ScummVM code style.

example usage.
```
	MemFile *memFile = new MemFile(vabData, vabDataSize);
	debug("Loaded vab file size: %lu", memFile->size());
	Vab *vab = new Vab(memFile, 0);
	vab->LoadVGMFile();
	VGMColl vabCollection;
	SF2File *file = vabCollection.CreateSF2File(vab);
	const byte *bytes = (const byte *)file->SaveToMem();
```

I would like to use this to play PS1 SEQ files in the blazing dragons engine. It should also work for vab data used by the PS1 version in Tinsel.

### DIFF
--- a/audio/soundfont/common.h
+++ b/audio/soundfont/common.h
@@ -19,6 +19,11 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  *
  */
+/*
+ * VGMTrans (c) 2002-2019
+ * Licensed under the zlib license,
+ * refer to the included LICENSE.txt file
+ */
 #ifndef AUDIO_SOUNDFONT_COMMON_H
 #define AUDIO_SOUNDFONT_COMMON_H
 
--- a/audio/soundfont/common.h
+++ b/audio/soundfont/common.h
@@ -1,0 +1,66 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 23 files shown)


Mataniko:
I think that the new code should be marked with the ScummVM header and GPL license, but comply with the vgmtrans license and represent it as the origin. 

@sev- thoughts?

audio/soundfont/rawfile.h
```diff
@@ -0,0 +1,77 @@
...
+/*
+ * VGMTrans (c) 2002-2019
+ * Licensed under the zlib license,
+ * refer to the included LICENSE.txt file
+ */
+
+#ifndef AUDIO_SOUNDFONT_RAWFILE_H
+#define AUDIO_SOUNDFONT_RAWFILE_H
+
+#include <climits>
```
sev-, line 31: we do not allow standard headers outside of baclends/ code. Could these be avoided?

audio/soundfont/common.h
```diff
@@ -0,0 +1,71 @@
...
+
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+/*
+ * VGMTrans (c) 2002-2019
+ * Licensed under the zlib license,
+ * refer to the included LICENSE.txt file
```
sev-, line 25: This should be changed to `VGMTrans_LICENSE.txt`

yuv422:
Is it ok for me to merge this one?

sev-:
Thanks a lot!

PR #2384, opened 2020-07-27
AGI: Apple2GS: Add `Speed` menu configuration to other games

ZvikaZ:
Previous commits added an optional Speed menu to few Apple2GS
games. Now adding it to all other games, except:
- manhunter
- agi demo
- winnie the pooh (that seems like a wrong detection entry)

Thanks to @trembyle for testing this PR :-)


sev-:
Thanks, makes sense.

PR #2385, opened 2020-07-28
GROOVIE: The 7th Guest MIDI fixes

NMIError:
This PR contains fixes for 3 small MIDI issues in The 7th Guest:

- Music fade-outs would sometimes cause the background music to start playing 
while there would be no music in the original interpreter, and it would not 
always restore volume when playing a new track, causing it to play at volume 0.
- Loading a savegame would start or continue playing background music for a few 
seconds before the music for the loaded room would start playing.
- Checking if the game should start background music would only happen if the 
game was receiving input.

### DIFF
--- a/engines/groovie/groovie.h
+++ b/engines/groovie/groovie.h
@@ -111,6 +111,7 @@ protected:
 
 public:
 	void waitForInput();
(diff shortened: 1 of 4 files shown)


sev-:
Thank you, merging.

PR #2386, opened 2020-07-29
PINK: Warn when the timestamps mismatch

henke37:
Write to the log when the timestamps are mismatched so that the issue has a chance of being spotted.


sev-:
Thanks!

PR #2388, opened 2020-07-30
AGI: Issue GUI warning when there is a missing Hercules font

ZvikaZ:

### DIFF
--- a/engines/agi/font.cpp
+++ b/engines/agi/font.cpp
@@ -21,6 +21,10 @@
  */
 
 #include "common/config-manager.h"
+#include "common/translation.h"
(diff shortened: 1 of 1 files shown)


raziel-:
@ZvikaZ 

I like that, though i have one question:
Would it be possible to add a general path for the font pickup?
Like the extra path?
It´s probably out of scope of this PR, but it would be awesome to not being forced to copy the font file to all game directories where i´d like to use it?
With the above mentioned, how about a GUI option (use hires font), if found? :-)

sev-:
Thanks! In fact, if you put this file into extrapath, it will be picked up successfully.

PR #2389, opened 2020-07-30
AGI: Fix bug #11209 (wrong `horizon` value saved)

ZvikaZ:
`_game.horizon` is defined as uint16, but was casted to uint8 before saving
thus, caused problems for values bigger than 128 - they were interpreted
as negative values


bluegr:
Nice catch! Merging

PR #2390, opened 2020-07-30
CINE: Add _animDataTable bounds checkings.

karisal:
Hopefully a fix for Coverity check CID 1430586
("Untrusted value as argument").


sev-:
Thanks!

PR #2393, opened 2020-07-31
PLUGINS: Big refactoring DO NOT MERGE

henke37:
The plugin code needed an overhaul, so I refactored it.

Key points:
* EngineManager moved to separate files
* Engine related code moved from PluginManager to EngineManager
* Clearer names for various member functions
* Removed stateless plugin enumeration member functions from PluginManager
* The Plugin class now knows if the plugin is loaded or not


henke37:
Gonna need some feedback on this. And testing. Definitely testing, since I don't use dynamic plugins. There are definitely spots that depend on plugins being loaded without checking that they are.

henke37:
Ok, I feel finished. I did a full check for all usage of plugins and I feel confident that I've examined every spot.

henke37:
So... There is one problem left to solve. That I know of. And as it so happens, I don't know how.

Dynamic plugins don't know their type until loaded. Yet I need to know their type to decide when to load them. Catch 22.

henke37:
Resolved the dynamic type issue by just deferring determining the type to the plugin provider. Which, by default, will detect all plugins as engine plugins. Which is fine for now, since all dynamic plugins are engine plugins right now.

sev-:
Okay, time to look into it. Could you please rebase the PR?

henke37:
It's going to require more than a simple rebase. I'm waiting for the dust to settle after the other PR got merged before I deal with this PR.

sev-:
What is status of this (besides being bitrot)? Still "DO NOT MERGE"?

henke37:
I think I need to do some further debugging to ensure that everything is correct with regards to plugin types.

bluegr:
5 months later... any news?

bluegr:
@henke37 any news?

henke37:
This one would likely require more work than it is worth to merge, it'd be faster to redo the work.

PR #2394, opened 2020-07-31
CINE: FW: Fix Roland MT-32 sample playing.

karisal:
Fix bug #11547 ("CINE: FW: Game crashes during intro with MT-32 music").
Future Wars PC disassembly passes 256 to writeInstrument and
writeInstrument overwrites it with value 246 here. So the value
becomes 246 in the end.


karisal:
@eriktorbjorn Here's a fix for the bug you reported.

eriktorbjorn:
That fixes the crash for me. (Not surprising, since it no longer has any way that I can see of writing outside the buffer on this code path, but I thought I'd confirm it anyway.)

sev-:
Thanks

PR #2395, opened 2020-08-01
Stream warnings

athrxx:
COMMON: fix stream class related warnings

This should take care of the warnings introduced with 1aea38cc.
It is a bit of a delicate matter, since different compilers seem to have slightly different reactions.
I've locally tested MSVC 2019, gcc and clang.

So, let's see what travis says and then we can try the buildbot...


athrxx:
Okay, that was a quick fail.

PR #2396, opened 2020-08-01
CINE: Fix disabled action menu and autosave thumbnail.

karisal:
This pull request has two bugfixes:

#1:
Future Wars PC disassembly does not save disableSysteMenu variable.
After adding versioning to Future Wars saves this value was also
saved and loaded but using it on load seems to have caused the
action menu (EXAMINE, TAKE, INVENTORY, ...) to be disabled sometimes
when it is not supposed to be disabled. Thus, now skipping the loading
of disableSystemMenu variable for versioned Future Wars saves.

#2:
Using the current screen contents for autosaves now.
For non-autosaves the screen contents before opening
the system menu are used.


sev-:
Thanks!

PR #2399, opened 2020-08-02
KYRA: Automatically suggest current time when adding a new savegame

neuromancer:
This small quality-of-life change suggests a new savegame name with the current time and date for the Land of Lore games. This allows to play the games with less usage of a keyboard (e.g. only qith mouse or remapping keys to a gamepad).

### DIFF
--- a/engines/kyra/gui/gui_lol.cpp
+++ b/engines/kyra/gui/gui_lol.cpp
@@ -2673,6 +2673,12 @@ int GUI_LoL::clickedSaveMenu(Button *button) {
 	_saveDescription[0] = 0;
 	if (_saveMenu.item[-s - 2].saveSlot != -3)
 		strcpy(_saveDescription, _saveMenu.item[-s - 2].itemString);
+	else {
+		TimeDate td;
+		g_system->getTimeAndDate(td);
+		Common::String ts = Common::String::format("%d:%d:%d on %d/%d/%d", td.tm_hour, td.tm_min, td.tm_sec, td.tm_mday, td.tm_mon + 1, td.tm_year + 1900);
+		strcpy(_saveDescription, ts.c_str());
+	}
 
 	return 1;
 }


athrxx:
Thanks for your contribution. I understand how this might make sense on some platforms.
I am okay with it as a a GUI option (in the Engine tab, like "smooth scrolling" and "floating cursors"), but not as a default for desktop platforms. Different ConfMan.registerDefault() writes for this option depending on the backend would be fine for me (via #ifdefs).

Do you think you can add that? If it is too much trouble I'll just merge this PR and add the things I want...

athrxx:
Another thing I just noticed: the whole string would have to go through translation manager just because of the use of "on". Maybe just use a comma (or hyphen or whatever) instead.

neuromancer:
Thanks for the feedback, I think I can fix both requests, let me try it.

neuromancer:
I just realize that implementing a new option in Kyra for some of the games (e.g. LoL) is an absolute MACRO nightmare, the code is very difficult to maintain because the macros between games and variations. And adding just one option will cause code duplication and a lot of changes (e.g. adding `false` as default for all the game variations). In summary, this could be implemented, but it will take me longer than expected.. :cry:

athrxx:
Okay, don't worry about it. I can do that myself. It isn't that bad for someone who is familiar with the code.

neuromancer:
Thanks! :) Btw, if this improvement is accepted, I'm interested to add the same feature to all the other games (starting with the ones I'm playing). Is that a feature you could be interested in general? (but as you said, disabled by default).

athrxx:
I'll merge this and add my modifications on top of that. Please check whether these are okay for you or whether you want anything changed.

I can't tell you in general if an option like this would be a welcome addition for all games. That is basically the engine authors' decision. It certainly makes sense here. Game variants that were originally designed for consoles/gamepads like EOB I Sega-CD and EOB II FM-Towns already have this as a default.

neuromancer:
You wanted to remove "on", that's fine, I think you can use "-". In the future, we could have a standard function to generate a proper save description using the current time.

athrxx:
Since this always seems to cause confusion: For the new gui option to become available in the launcher you have to start the game once and then exit. After that the new option should be there.

athrxx:
I've made another modification to the save description, since I didn't really like the level numbers in there. It should now be a little more transparent to the user, but still fit in the text slot.

neuromancer:
Looks good to me. I'm thinking in adding a key for showing the automap (M), instead of clicking the icon. This is not available in the original game (AFAIK, however the manual doesn't show all the key bindings), but it doesn't give any advantage and it is very useful for the user. Are you interested in this change for a new PR?

athrxx:
I've just checked. There is no key binding for the map button. I wouldn't be opposed to a PR for that if it is done right.

Let me give you some pointers, since the process might be not as simple as you think:
- In `kyra_v1::setupKeyMap()` you can find the key codes that you have to assign to the buttons. For 'm' you have an entry `{ KC(m), 52, 52, 77 },` . That's code 52 for DOS and PC-98 and code 77 for FM-Towns.
. In the kyra.dat resource files _lol_dos.h_, _lol_dos_cd.h_ etc. you have entries named  `kLoLButtonDefsDOS`,  `kLoLButtonDefsDOSCD,` etc. The second and third struct members are the keycode bindings. You just have to replace the zero value for the second entry with your key code. The map button seems to be no. 92 (third from bottom).

neuromancer:
Uhm, I did a quick try and this should be enough:

```
diff --git a/engines/kyra/gui/gui_lol.cpp b/engines/kyra/gui/gui_lol.cpp
index 42b22d6..67a86f4 100644
--- a/engines/kyra/gui/gui_lol.cpp
+++ b/engines/kyra/gui/gui_lol.cpp
@@ -784,6 +784,8 @@ void LoLEngine::gui_updateInput() {
 
        if (inputFlag == _keyMap[Common::KEYCODE_SPACE] || inputFlag == _keyMap[Common::KEYCODE_RETURN]) {
                snd_stopSpeech(true);
+       } else if (inputFlag == _keyMap[Common::KEYCODE_m]) {
+               clickedAutomap(0x0);
        } else if (inputFlag == _keyMap[Common::KEYCODE_SLASH]) {
                if (_weaponsDisabled || _availableSpells[1] == -1)
                        return;
```

Of course, this won't update the scummvm key mapping to allow remapping using other keys/gamepad, but at least it will work.

athrxx:
Okay, that's the sort of hack I tried to avoid by giving you the pointers :-P

neuromancer:
I will check the mappings as you said, but that will be enough to implement this feature? Is there a dispatcher to call the `clickedAutomap` function for certain keys?

athrxx:
Well, to answer that question whether the change is enough or not I had to test it myself. And yes, it does. But now that I have done these code changes for my test I've just pushed them. So, no more need for a PR :)
Unless you want to add something to LoLEngine::initKeymaps(). Feel free to have a go at that...

neuromancer:
> Well, to answer that question whether the change is enough or not I had to test it myself. And yes, it does. But now that I have done these code changes for my test I've just pushed them. So, no more need for a PR :)

Thanks a lot! I will check your code so I can submit future PRs without annoying you.

> Unless you want to add something to LoLEngine::initKeymaps(). Feel free to have a go at that...

I think I can do that.

neuromancer:
Ok, the PR for updating the keymap is ready. Btw, looking at the source code I found that if you press 'c' when in the map mode, all the map will be revealed. If this happen as an accident, it could spoil the game, so perhaps it is a good idea to have a 'cheats enabled' options with this (and perhaps other) keys for making the game less difficult.

athrxx:
I prefer not to change this, since it is the original game's behavior and not something added by ScummVM.

PR #2403, opened 2020-08-12
ENGINES: ALL: Split MetaEngines into 2 parts, statically link detection features across all engines.

aryanrawlani28:
# Updates:
- MetaEngine is now renamed to MetaEngineStatic (has static parts - detection, etc)
- MetaEngineConnect is now renamed to MetaEngine (has dynamic parts)
- Wintermute & Sci fallback detection disabled by default for systems which use UncachedPluginManager

# Remaining things:
- [ ] File structuring issues for Visual Studio.
- [ ] Dynamic detection plugin not loading for some systems, needs some location reworking.

---------------------------------------------------

This PR splits apart the MetaEngines into a way that detection features are always available* to the executable.

This PR also extends the support for the same to all of ScummVM's engines (stable & unstable). Work for ResidualVM's engines is in progress and a PR for the same will be opened soon. It should probably get merged around the same time when ResidualVM syncs up with ScummVM.

(*Some engines fallback detection is not linked statically. More info in a bit.)

# New Structure

- Addition of 2 new classes, MetaEngineConnect & AdvancedMetaEngineConnect.

    - MetaEngine: Houses all methods that are always linked statically. As before, each engine provides their own implementation of a MetaEngine, but only provides detection and related methods.

    - MetaEngineConnect: Houses all things that "can" be linked dynamically. These are bridge functions, that connect an Engine with MetaEngine, thus the name. It houses methods like `createInstance`, which instantiates an actual game engine.

    - AdvancedMetaEngine: Inherits from MetaEngine and provides a default detection method.

   - AdvancedMetaEngineConnect: Inherits MetaEngineConnect to derive a structure like before.

- Most changes are centered around plugins. Two main new helpers are `giveEngineFromMetaEngine` and `giveMetaEngineFromEngine` which do the same thing as they sound like. These help out in various scenarios where you would need the matching engine-metaengine to do something.

The format of each individual engine is largely similar to before, but I will explain in a bit how some things will change.

# Changes to existing engines

Most changes are pretty simple to follow. The new structure will need to be kept in mind when working on new engines,
To keep things short, I'll use _"ME" for MetaEngine,_ _"AME" for AdvancedMetaEngine_, _MEC & AMEC for the connect counterparts_.
I'll also use the `AGI` engine as an example, so it's clear what everything means.

- If an engine inherits ME, it should also inherit MEC. 
- The class ME for AGI lives in `agi/detection.cpp`, and inherits ME. The detection-related code goes here, detection tables, etc.
Example:
`class AgiMetaEngine : public MetaEngine {...}`
- The same engine should use the equivalent meta-counterpart in `agi/metaengine.cpp`, a file specifically for MEC/AMEC.
Example:
`class AgiMetaEngineConnect : public MetaEngineConnect {...}`

- If an engine uses AME, it should also use AMEC.

-----------------------------

That's the basic engine split. As for the engines themselves,
- Many engines have a custom description for the game. For example, in addition to the `ADGameDescription` struct, they would have their own, something like:
```
struct AGIGameDescription {
	ADGameDescription desc;

	int gameID;
	int gameType;
	uint32 features;
	uint16 version;
};
```
- These go into `agi/detection.h`.
- Sometimes the detection-tables will need some enumeration values present in the main engine header file, for example, `agi.h`. To avoid including the whole header which would contain irrelevant information for the executable, a copy of these enums is provided to both - the engine & metaengine.
Simply shift these enumeration values to `agi/detection.h`. Include it in the main engine header, so everything returns to normal. However, now we can include it in the `detection.cpp` translation unit, while avoiding the engine header file.

- Engines inheriting from `MetaEngine` will provide their own detection method. An example of this is Scumm. However, detection is not only used during detecting of games, but also during creating an instance. Scumm provides some internal static methods, called `detectionImpl` which is used both - by `MetaEngine::detectGames(...)` & by `MetaEngineConnect::createInstance(...)`. The solution for this is to shift the common code inside another header file, `detection_internal.h`. Then, this is included by both - `detection.cpp` & `metaengine.cpp`
   - Remember to mark the methods as static inside the header, so the visibility is limited to the TU.
- MetaEngines now go into the executable. They provide some naming functions - `getName`, `getEngineId` & `getOrignalCopyright`. Because MetaEngines will need to find a relevant MetaEngineConnect, the implemented MetaEngineConnect should override the `getName` method, and provide it a name similar to the one found in `getEngineId` of MetaEngine.
- Lastly, after the creation of your files, you need to register them as plugins.

- For `detection.cpp`, use `REGISTER_PLUGIN_STATIC(...)` macro, as they always go into the executable.
   - Keep in mind that the ID should be suffixed with "DETECTION". This is because the ID cannot conflict with the one in metaengine. The type should be a metaengine. Example:
```
REGISTER_PLUGIN_STATIC(AGI_DETECTION, PLUGIN_TYPE_METAENGINE, AgiMetaEngine);
```
- For `metaengine.cpp` the story is as-is, if the plugin is dynamic, use the dynamic macro - otherwise the static version.
-------------------------------
## Module objects / Building detection features
- Each engine's module.mk file will be adding object files to `DETECTION_OBJS`. It looks something like this:
```
...
...
# Include common rules
include $(srcdir)/rules.mk

# Detection objects
DETECT_OBJS += $(MODULE)/detection.o
```

- Remember to *add* `metaengine.o` and *remove* `detection.o` from the `enginename` module.
- Additional dependencies for detection are also *ok*. However, remember that engines could also use those files, so it wouldn't always be the case to separate the module-object completely. I will give some examples:
  - AGI uses something called `wagparser` to help in its fallback detection. The engine itself doesn't need it, so we can separate it completely i.e remove from `MODULE_OBJS` in `engines/agi` module and add it to `DETECT_OBJS` inside the same module list.
   - Scumm uses 2 files - `file.cpp`, `file_nes.cpp` to help with detection. However, the engines themselves need those files too, so we can't separate it completely. Hence, the updated module will look like
```
MODULE := engines/scumm

...
# Rest of the file
...

# Detection objects
DETECT_OBJS += $(MODULE)/detection.o

# Skip building the following objects if a static
# module is enabled, because it already has the contents.
ifneq ($(ENABLE_SCUMM), STATIC_PLUGIN)
DETECT_OBJS += $(MODULE)/file.o
DETECT_OBJS += $(MODULE)/file_nes.o
endif
```
- The makefiles then take care of adding all the `DETECT_OBJS` from each of engine module files, regardless of which engine is being enabled or not.
## Some engines that would require review
- While most of the engines adapted pretty nicely to this structure, some had a somewhat different scenario because of which the base structure was changed a bit. I will list all of them below.

### Sci & Wintermute (fallback-detection)
- These engines have a struct similar to `ADGameDescription`, whose contents are overwritten by the fallback detection method. The problem is that these fallback detection methods are heavily dependent on engine resources.
   - I tried to make a "lite" version for these classes just to be helpful in detection, but that was a fast fail because it's too dependent on engine resources.
- Thus, a workaround would need to be added.
- A solution for this is that you would still override `fallbackDetect` in MetaEngine (which lives in `detection.cpp`, but instead of detecting games here, you provide a hook to the relevant engine plugin and call a relevant method from there.
  - Override a function called `fallbackDetectExtern` in the `MetaEngineConnect` class which lives in `metaengine.cpp` and provide the actual detection here, which is engine-resources dependant. See https://github.com/aryanrawlani28/scummvm/commit/eb960d0814a3dd1d6cd2e7f91d53518f88176476 for more info, and an in-code example.

### Mohawk (dialogs)
- A while back https://github.com/scummvm/scummvm/pull/2164 added support for in-game custom dialogs.
- If you look at some of the engines currently scummvm master tree, which inherit from AME, you'll see that they provide some extra gui options via an array, `ADExtraGuiOptionsMap gameGuiOptions[]`  and then pass this along to the AME constructor.
  - To not disturb the engine-structure too much, these are still as-is, meaning that without an engine plugin loaded, you'll be able to see these options. (Other things like Keymaps, achievements are still coming from the engine.)
- Mohawk however, has the new dialogs in which they could query information from the game, then setup dialog accordingly. As such, it's probably best that they remain in the engine plugin itself.
- So, ME's & MEC's both have the ability to create custom dialogs. ME's only provide the "Engine" tab inside "Edit Game" option, while MEC's provide in-game dialogs, etc.
- For now, to clearly distinguish between them, the name `buildEngineOptionsWidgetDynamic` is given, but this is not really proper. `extern & intern` like the ones I did with fallback detection were ok, but it could be confusing in this scenario. Hence, this name can probably be changed into something better.

# Detection features as a plugin themselves
- On some low-hardware end platforms, the executable size bloat would prove difficult.
- Making detection features available as a plugin itself would be beneficial in that scenario.

## How is it done?
- We already had the detection features separated from Engines. They are normally supposed to link directly with the executable, but since we have to have a plugin, we can group all the detection files themselves into a `DETECTION` type plugin.
- In the ScummVM `engines` directory, look for `detection.cpp`. This is where we include the detection tables and mark the plugin as a `PLUGIN_TYPE_DETECTION`.
- When using configure, `--enable-detection-dynamic` will be available, which enables the detection objects to be packed into a library.
  - When using this, helper methods in `PluginManagerUncached` will help in loading the detection library, linking the MetaEngines, and eventually clearing those out of memory once we don't need them.
   - If not using that, they will be linked statically, but still have an explicit option - `--enable-detection-static` available.

## Modules objects / Building detection features for dynamic detection plugins
- As said above, each engine is already populating the detect_objs variable.
- In the configuration script, when `enable-detection-dynamic` is passed, the `Makefiles.common` includes the below module:
```
MODULE := detection // Only added if dynamic-detection enabled

DETECT_OBJS_DYNAMIC=$(addprefix ../,$(DETECT_OBJS)) // Since it lives outside the engines subdirectory, add "../" prefix.

MODULE_OBJS := \
	../engines/detection.o \
	$(DETECT_OBJS_DYNAMIC) // Reuse the already existing list of detection objects

# Reset detect objects, so none of them build into the executable.
DETECT_OBJS :=

PLUGIN := 1

# Include common rules
include $(srcdir)/rules.mk
```

---------------
As always, any suggestions, improvements, or anything really, is appreciated and helpful.


henke37:
Not sure why you chose to use the word "give" instead of "get" as is custom in programming.

aryanrawlani28:
> Not sure why you chose to use the word "give" instead of "get" as is custom in programming.

There's already a "get" present. A "give" might be better for readability purposes and avoids ambiguity perhaps.

henke37:
I think there is a greater risk of confusion if you mix verbs. And if there would be a conflict, it suggests that the feature exists already.

base/plugins.cpp
```diff
@@ -256,6 +265,69 @@ void PluginManager::addPluginProvider(PluginProvider *pp) {
...
+			if (metaEnginePluginName.equalsIgnoreCase(enginePluginName)) {
+				enginePlugin = (*itr);
+				found = true;
+				break;
+			}
+		}
+	} while (!found && PluginMan.loadNextPlugin());
+
+	if (enginePlugin) {
+		warning("MetaEngine: %s \t matched to \t Engine: %s", plugin->getName(), enginePlugin->getFileName());
```
mgerhardy, line 294: Is that really a warning?

aryanrawlani28: Ah, sorry, I guess I didn't point it out. Throughout the tree, I have placed many warning messages. These just exist for review/test purposes, because it will be way easier to track down what's happening if anyone faces any issues.

I plan to remove all unnecessary comments, warning statements once the PR looks nice and near merge. So, I'd request you to ignore these warning statements, they're just there temporarily (most of them anyway).

sev-:
@aryanrawlani28  I am not sure who suggested you moving files around and creating detection/ directories here and there.

What is the purpose of this exercise?

sev-:
Hi Aryan,

I tested, it works nicely. Will start looking into the code deeper. These are the generic notes based on a cursory look and attentive reading of your comprehensive description above.

The PR got rot, again, sigh. I promise to merge it this week (if you have time to address my notes, of course).

First of all, I second @henke37 's opinion: the classic verb for is 'get', not 'give', thus, `getEngineFromMetaEngine()` and `giveMetaEngineFromEngine()` should be used.

Second, I was quite confused by naming of the MetaEngineConnect() and AdvancedMetaEngineConnect(). When reading your description and the code, it is not quite following their logic. E.g. you're asking to keep all `MetaEngineConnect()` methods in `metaengine.cpp` file. And `MetaEngine()` methods are kept normally in `detection.cpp`. My proposal here would be to call the detection part (`MetaEngine` now) as `MetaEngineStatic` or `MetaEngineDetection`. And the remaining dynamic methods will be in a class called plainly `MetaEngine`. Thus, current `AgiMetaEngineConnect` becomes `AgiMetaEngine`, and current `AgiMetaEngine` becomes `AgiMetaEngineStatic`. How does it sound?

Then, regarding the complex fallback detection in Wintermute and SCI engines now. I do not like the proposed solution of loading the plugin for detection, as this would clash with the purpose of this whole exercise. However, we do really need trying detecting it. Thus, I would do it this way: Have a setting in `ConfMan.getBool("always_run_fallback_detection")`, and set it to true on big platforms but on smaller platforms I would set it to false and do not attempt to load the plugin even if present. Also, the message could be improved like: "Engine plugin for Wintermute not present. Fallback detection is disabled.". The current message gives the impression that some _known_ games are not detected.

sev-:
Putting detection into plugin does not compile.

I used

`./configure  --enable-detection-dynamic --enable-plugins --default-dynamic  --disable-all-engines --enable-c++11 --enable-engine=director,scumm`

Getting the following:
```
make: *** No rule to make target `detection/detection.o', needed by `plugins/detection.plugin'.  Stop.
```

This is after my attempt to fix it, as previously it was `detection/../engines/detection.o`

engines/detect_modules.mk
```diff
@@ -0,0 +1,15 @@
+MODULE := detection
+
+DETECT_OBJS_DYNAMIC=$(addprefix ../,$(DETECT_OBJS))
+
+MODULE_OBJS := \
+	../engines/detection.o \
```
sev-: This will not compile. Apparently, you had an empty directory `../detection`, this is why this path was still valid for you. More work is needed here for removing the `detection` prefix.

engines/dialogs.cpp
```diff
@@ -273,7 +273,7 @@ ConfigDialog::ConfigDialog() :
 	assert(g_engine);
 
 	const Common::String &gameDomain = ConfMan.getActiveDomainName();
-	const MetaEngine &metaEngine = g_engine->getMetaEngine();
+	const MetaEngineConnect &metaEngineConnect = g_engine->getMetaEngineConnect();
```
sev-, line 276: With renaming back, all of these become not needed.

engines/metaengine.cpp
```diff
@@ -37,15 +37,15 @@
 #include "graphics/managed_surface.h"
 #include "graphics/thumbnail.h"
 
-const char *MetaEngine::getSavegameFile(int saveGameIdx, const char *target) const {
+const char *MetaEngineConnect::getSavegameFile(int saveGameIdx, const char *target) const {
```
sev-, line 40: And these renames too become obsolete.

engines/neverhood/module.mk
```diff
@@ -4,16 +4,16 @@ MODULE_OBJS = \
...
 	entity.o \
 	gamemodule.o \
 	gamevars.o \
 	graphics.o \
 	klaymen.o \
 	menumodule.o \
+	metaengine.o \
 	microtiles.o \
-	module.o \
+	module_scene.o \
```
sev-, line 16: Why this rename?

gui/saveload-dialog.cpp
```diff
@@ -103,7 +103,7 @@ void SaveLoadCloudSyncProgressDialog::handleTickle() {
 #endif
 
 #ifndef DISABLE_SAVELOADCHOOSER_GRID
-SaveLoadChooserType getRequestedSaveLoadDialog(const MetaEngine &metaEngine) {
```
sev-, line 106: These also will be reverted after rename.

gui/saveload.cpp
```diff
@@ -77,14 +77,24 @@ Common::String SaveLoadChooser::createDefaultSaveDescription(const int slot) con
 
 int SaveLoadChooser::runModalWithCurrentTarget() {
 	const Plugin *plugin = EngineMan.findPlugin(ConfMan.get("engineid"));
```
sev-, line 79: You could avoid many renames if you would keep `plugin` variable and introduce `metaEnginePlugin` or something. This is for all these files in the GUI.

sev- (changes requested):
Very good quality. We are so close to the merge.

Please, do the class renames and address those few nit-picking comments and we merge it.

sev-:
I could help with the renames if you like.

engines/neverhood/module.mk
aryanrawlani28: My commit notes for this say:
- The name was clashing with one of the audio files.
- New changes were spread to the detection files, so no idea how this file is involved.
- In the meanwhile, this commit should fix a build error, if any.

VStudio was having a problem with this file and one in the main scummvm subproject because they had the same name. I know we have the tool that checks for clashing names and renames it acc in dev tools, but this was still an issue and was the cause of compilation error.

Makefile
```diff
@@ -15,6 +15,10 @@ DEPDIR      := .deps
 MODULES     :=
 MODULE_DIRS :=
 
+# All game detection-related object files for engines
+DETECT_OBJS :=
+USE_RULES   := 1
```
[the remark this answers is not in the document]
aryanrawlani28: Have changed it.

Makefile.common
```diff
@@ -52,6 +52,36 @@ CPPFLAGS := $(DEFINES) $(INCLUDES)
 # Include the build instructions for all modules
 -include $(addprefix $(srcdir)/, $(addsuffix /module.mk,$(MODULES)))
 
+# Store original info
+MODULES_ORIG:= $(MODULES)
+MODULE_DIRS_ORIG := $(MODULE_DIRS)
+KYRARPG_COMMON_OBJ_ORIG := $(KYRARPG_COMMON_OBJ)
```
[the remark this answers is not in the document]
aryanrawlani28: Can you clarify what you mean?
By the time flow reaches here:
- `engines.mk` file has been included and everything is going according to configuration scripts.
- these lines stores the original information, preparing for **all** module.mk files to be included.
- below, all module.mk files are included:
```
# Include all engine's module files, which populate DETECT_OBJS
-include $(srcdir)/engines/*/module.mk
```
which populate `DETECT_OBJS`, but they also fill other values like `MODULES`, `MODULE_DIRS`.
- So, after my work of populating detect objs is done, I would need to reset values to the original configuration (how the conf script intended to do) like
```
# Reset stuff
MODULES := $(MODULES_ORIG)
```

Makefile.common
```diff
@@ -52,6 +52,36 @@ CPPFLAGS := $(DEFINES) $(INCLUDES)
...
+MODULE :=
+MODULE_OBJS :=
+MODULE_DIRS := $(MODULE_DIRS_ORIG)
+PLUGIN :=
+KYRARPG_COMMON_OBJ := $(KYRARPG_COMMON_OBJ_ORIG)
+
+# Enable-rules again
+USE_RULES := 1
+
+ifneq "$(DETECTION_STATIC)" "1"
```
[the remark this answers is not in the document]
aryanrawlani28: Taken care of.

base/commandLine.cpp
```diff
@@ -884,19 +887,30 @@ static Common::Error listSaves(const Common::String &singleTarget) {
...
+				if (!singleTarget.empty())
+					return Common::Error(Common::kEnginePluginNotFound,
+				                     	 Common::String::format("target '%s'", i->c_str()));
+				printf("EnginePlugin could not be loaded for target '%s'\n", i->c_str());
+				continue;
+			}
 		}
 
-		const MetaEngine &metaEngine = plugin->get<MetaEngine>();
+		const MetaEngineConnect &metaEngine = enginePlugin->get<MetaEngineConnect>();
```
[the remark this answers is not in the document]
aryanrawlani28: Thanks for your advice! I've finished with the renames, so will resolve all conversations related to it.

PR #2404, opened 2020-08-12
SCI: New Amiga and Mac sound drivers

waltervn:
The current AmigaMac driver was created by playing around with values in sound banks and observing what the effect was in an emulator. There are quite a few issues with this driver, especially in SCI1. This PR replaces it with drivers that were RE'd from scratch. This is based on work done together with Matthew Hoops back in 2013/2014.


sluicebox:
*GASP* Oh.. My... Gosh......

Congratulations on seeing this through, this is tremendously exciting!!

engines/sci/sound/drivers/macmixer.h
```diff
@@ -0,0 +1,268 @@
...
+	ufrac_t _nextTick;
+	ufrac_t _samplesPerTick;
+	bool _isPlaying;
+	const Mode _mode;
+	Channel _mixChannels[kChannels];
+	byte _mixVolume;
+};
+
+template <typename T>
+Mixer_Mac<T>::Mixer_Mac(Mixer_Mac::Mode mode) :
```
sluicebox, line 93: VS2015 can't compile this line, it gets confused by the "Mixer_Mac::" qualifier on the parameter because it's an enum in the template class. Removing "Mixer_Mac::" avoids this while keeping the correct scope on the Mode type.

I can't find a VS bug report for this, but I verified it in a standalone test case and that it was fixed in at least VS2019.
Sorry I use an old one! =)

error C2988: unrecognizable template declaration/definition

bluegr:
Great work, well done!

sev-:
Thank you! This an excellent addition.

PR #2405, opened 2020-08-13
CREATE-PROJECT: Set MSVC subsystem

henke37:
There is no need to show an unusable console window, so only show that if the text console is enabled.

### DIFF
--- a/devtools/create_project/create_project.cpp
+++ b/devtools/create_project/create_project.cpp
@@ -2138,3 +2138,16 @@ void error(const std::string &message) {
 	std::cerr << "ERROR: " << message << "!" << std::endl;
 	std::exit(-1);
 }
+
+bool BuildSetup::featureEnabled(std::string feature) const {
+	return getFeature(feature).enable;
+}
+
+Feature BuildSetup::getFeature(std::string feature) const {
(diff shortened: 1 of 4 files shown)


bluegr:
Nice work! A good and easy way to manage the console

SupSuper:
Upon further examination, the `text-console` parameter is kinda deceptive. The full name is `USE_TEXT_CONSOLE_FOR_DEBUGGER`.
Personally I like having the text console for debug output but **not** as a debugger, so I'd prefer if the two weren't tied together.

sev-:

henke37:
As far as I know, ready to merge unless someone has an objection.

dreammaster:
I previously overlooked this pull request. Since I'm at work right now and can't test it, and I'm a bit confused about "text window" vs "console window" in the description. Does this mean that builds done in MSVC won't have the console window open by default?

henke37:
That's the idea yes, if the os level console isn't used, don't show it.

dreammaster:
I've no objection then. As long as it's been tested, a merge is fine by me

henke37:
I've been using it for my builds for like half a year now.

PR #2406, opened 2020-08-15
CREATE_PROJECT: Update MSVC libraries

SupSuper:
One of the biggest challenges of updating the prebuilt MSVC libraries has been the hardcoded libraries in CREATE_PROJECT, as they don't match modern package managers like vcpkg and require manual work. In the process of updating them, I refactored several things:

- MSVC libraries can have different names depending on configuration, so now the `MSVCProvider` has a table of library names for every configuration (Debug/Release/etc) and dependencies. This makes it much easier to maintain, and is similar to the other project providers.
- This removed the need for the BuildSetup `libraries` property, as it was only really used by MSVC. The other providers have their own logic for handling libraries, which makes sense as there is no "global" list of libraries, every platform and tool will have different names and formats. Now each provider is responsible for turning Features into libraries.

So here's what needs testing:
- Make sure I didn't break any of the other project providers and all features still build correctly. 
- For the MSVC users: at long last, [an updated library zip](https://www.dropbox.com/s/erh2i5e0ls1ct59/scummvm_libs_2015.zip?dl=1)! All static, all up-to-date, for all configurations!
To test them just replace your `SCUMMVM_LIBS` folder with the one in my ZIP and enable `--use-canonical-lib-names` in `create_msvc.bat`. If all goes well, these will be the new default.

### DIFF
--- a/devtools/create_project/codeblocks.cpp
+++ b/devtools/create_project/codeblocks.cpp
@@ -24,6 +24,7 @@
 #include "codeblocks.h"
 
 #include <fstream>
+#include <cstring>
 
 namespace CreateProjectTool {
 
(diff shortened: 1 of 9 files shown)


sluicebox:
This is great!

I tested this as much as I could on VS2015, I figure you've got VS2019 covered. I only tested 32-bit, I can try do some specific 64 stuff if you want.

Wow so it's been a while since I've tried building all engines in VS in Release mode. The linker died with a stack overflow. I tried disabling titanic and ultima and it still failed so I gave up and just did sci since that's what I already had set up, but for Debug tests I did all engines.

First I verified that the new create_project continued to work with the existing set of static libraries, and it did.

Next I used the new static libraries and enabled everything. It worked! My command line was:

`create_project ..\..\ --enable-all-engines --msvc --use-canonical-lib-names --msvc-version 14 --enable-text-console --enable-tts --enable-mpeg2 --enable-updates --enable-faad`

(mpeg 2 and faad aren't enabled by default)

The one cosmetic thing is that every project now produces a warning about arm64 when loaded:

`warning  : Platform '[arm64, 0]' referenced in the project file 'sci' cannot be found.
`

sev-:
Thanks a lot! merging.

PR #2407, opened 2020-08-15
KYRA: Fixed typos in keymap entries and added map key

neuromancer:
This removes the extra ')' in the first two entries of the keymap description and add the map button (m).


sev-:
Thanks!

PR #2408, opened 2020-08-17
SCI: Add support for Roland D-110

waltervn:
This PR adds support for the Roland D-10/D-20 keyboards and the rack-mounting Roland D-110. These devices are similar to the MT-32. There was unofficial support by Sierra through aftermarket patches (interestingly PQ2 usually has the patch included, probably by accident).

This PR replaces the Yamaha FB-01 checkbox in the engine options with a popup widget, allowing a list of devices to be shown.

Short overview of what we're dealing with:
- Early SCI0: There's a driver for KQ4 with an embedded patch.
- Late SCI0: The driver reads patch 0, which is basically identical to the MT-32 patch format, except that the text strings contained within it are 32 bytes instead of 20.
- Early SCI1: No driver was made, instead SysEx files were created that users had to manually upload to their device before starting the game with the MT-32 driver.
- Late SCI1: Here they switched to using GENMIDI.DRV. Unfortunately the patches for these games have not been found. They were originally distributed on the Sierra BBS in a self-extracting archive called GEND110.EXE. Hopefully they will turn up some day.


sluicebox:
I love adding SCI engine support for even more obscure sound devices. This comment is just about renaming existing source files to work around the MSVC #include conflict.

I tried to figure out a formal way to deal with the MSVC conflict, but I couldn't do any better, and I even roped @SupSuper into it. I use detection_tables.h a lot so if we have to add a subdirectory and rename it all because of MSVC, I just want to first make sure that isn't avoidable, otherwise I'll grumble about it every time I access it in the future.

I couldn't figure out anything better. If I understand the problem, it's that MSVC's rules for resolving #include "___" are different from other compilers. MSVC starts at the initial file with #include (ThemeEval.h) and tries its directory ("gui", which fails) and then looks at the directory of the parent file (detection.cpp) and tries that... which unexpectedly succeed because the "sci" directory happens to have "graphics/font.h" too, and so the wrong header file gets used. Other compilers just do the first step and then move on to the list of include/search paths.

I wish there were a formal way to address the naming conflict since it's likely to happen again. I'm not optimistic since this is implementation specific behavior, and I don't see any MSVC options to alter it, but this is all still a welcome reminder as to why I don't write C++.

waltervn:
I thought about several other ways to get around this, but in the end settled on a 'detection' directory. Gob has this directory too so there is precedent for having one. That being said, I'd be happy to change it to some other workaround if people don't like this one.

sluicebox:
I keep thinking of better searches for this problem only to come up with more results that acknowledge it without offering solutions, other than avoiding it. I'm surprised this problem doesn't come up more often, but then I'm often surprised by the state of C++. I don't want this to distract from this cool PR. The main MSVC nerds have taken a look, which is all I wanted, so it's all good.

sev-:
Thanks, merging.

PR #2409, opened 2020-08-17
DIRECTOR: Add detection for more games

trembyle:


engines/director/detection_tables.h
```diff
@@ -117,12 +117,18 @@ static const DirectorGameDescription gameDescriptions[] = {
...
+	WINGAME("bookshelf94", "", "BS94.EXE", "e42c0effd8182079e8b6c93fe04a545f", 637552, 300),
+
 	MACGAME("chaos", "", "The C.H.A.O.S. Continuum", "2ce360c9ea2da80a2c6d1040b0ad92dd", 384481, 300), // original name is The C.H.A.O.S. Continuum™
 	WINGAME("chaos", "", "CHAOS.EXE", "7c8230a804abf9353b05627a675b5ffb", 375294, 300),
 
 	// almost all folders and file names use Japanese encoding, including executable
 	MACGAME_l("easternmind", "", "Tong Nou", "276bee761e48a6fd709df77d5c2f60dd", 395037, Common::JA_JPN, 300),
 
+	WINGAME("encarta94", "", "ENCART94.EXE", "98b82159158fe0cbb51e2083c3322c69", 1342976, 300),
+	WINGAME("encarta95", "", "ENCART95.EXE", "7a895781299abfd0493d975f9ac64160", 2064512, 300),
```
sev-, line 129: This should be version 310. And we need all entries to be revisited.

trembyle: Windows Encarta entries are removed for now as there is no executable or first movie to point to. I've added minor version numbers for all of the new entries. Some versions still translate to even numbers, such as `4.0b1 = 400` or `5.0r134 = 500`.

sev-:
Thanks! Merging.

PR #2410, opened 2020-08-18
Dragons credits

yuv422:
This PR adds Blazing Dragons credits.

I needed to increase the text max length from 29 to 30. Hopefully that is ok.
I also needed to add support for capital acute A


sev-:
Thanks!

PR #2413, opened 2020-08-19
DIRECTOR: Fix info entry double frees

henke37:
The InfoEntry struct owns a buffer, so the default copy constructor is wrong. This oversight lead to a double free.


sev-:
Thanks!

PR #2415, opened 2020-08-19
GLK: ADRIFT - Move known v5 game to unsupported list

raziel-:
If this is wrong, please tell and reject


sev-:
Thanks for catching it

PR #2418, opened 2020-08-20
iOS: Added LaunchScreen.storyboard to support full screen on iPad Pro

andshrew:
Apple requires an iOS app to use a Launch Screen storyboard - instead of a LaunchImage - to enable the app to run in full screen on devices like an iPad Pro. The Launch Screen file I have created just replicates what was in the current LaunchImage file (ie. centered ScummVM logo, orange background)

When run on an iPad Pro the app currently looks like this:- note black borders on either side
Screen Shot 2020-08-19 at 13 57 24

With a Launch Screen storyboard added the app can now display in full screen:
Screen Shot 2020-08-19 at 13 59 36

The procedure for updating create_project so that this new file was properly included in the Xcode project it creates is not entirely clear, so how I have implemented here was the only way I was able to reliably get the file included in the project and properly assigned to the iOS target only. This was done by adding the new file path in xcode.cpp so that it is included in the project, and naming it with an \_ios suffix so that it is only assigned the iOS target (ios7\_ prefix does not seem to work correctly, with the file assigned to both the iOS and macOS targets).

Tested on an iPad Pro 13.6.1, Xcode 11.6 (no jailbreak), following build instructions from https://wiki.scummvm.org/index.php?title=Compiling_ScummVM/iPhone#Using_Xcode

Andrew


devtools/create_project/xcode.cpp
```diff
@@ -806,6 +806,7 @@ XcodeProvider::ValueList& XcodeProvider::getResourceFiles() const {
 		files.push_back("dists/engine-data/wintermute.zip");
 		files.push_back("dists/engine-data/macventure.dat");
 		files.push_back("dists/engine-data/xeen.ccs");
+        files.push_back("dists/ios7/LaunchScreen_ios.storyboard");
```
mgerhardy, line 809: This is not correctly indented - use tabs, not spaces

criezy:
There are two ways to build ScummVM for iOS, one is with `create_project` and Xcode, and the other is using `make` on the command line. This PR only updates `create_project` currently, and will need to be modified to also handle the `make` process.

I am not really familiar with `create_project`, but I will have a look at the change in the next few days to see if there is a better way to handle this.

For the `make` process, the `ios7bundle` target in the `ports.mk` file needs to be updated. Do you want to take a look at this? Otherwise I can do it myself.

andshrew:
Thanks for your reply. I can certainly give modifying the `ports.mk` file a go but would appreciate your assistance if I get stuck as I'm not overly familiar with that (in particular ensuring it actually builds from the command line once the file has been updated as I've only used Xcode for iOS development so far).

andshrew:
I have had a look at amending this so that it can be built via `make` from the command line as well as via the Xcode project, but I need a bit of assistance from someone with more experience.

The first issue is that the [instructions](https://wiki.scummvm.org/index.php?title=Compiling_ScummVM/iPhone#Using_the_command_line) for building from the command line seem to be a bit out of date. I can follow it down to `./cctools-ld64.sh` which then fails to download `http://download.gna.org/gnustep/libobjc2-1.7.tgz` as it no longer exists.

The second issue is that the ios-toolchain-based-on-clang-for-linux [seems to suggest](https://code.google.com/archive/p/ios-toolchain-based-on-clang-for-linux/wikis/HowTo_en.wiki) that it is unable to compile storyboard files so it may not work even if I were able to follow the above steps. Given its age I'm unsure if there is another alternative that should now be used.

> NOTE: storyboard and xib files can not be compiled by this toolchain, since the format of these files is closed and undocumented, only nibtool shipped with xcode can handle these files and we had no chance to port it.

As the code in this PR currently stands (from testing in Xcode only on a non-jailbroken device), if you compile it without the storyboard file then when the app is first launched you will get a blackscreen for a moment before it loads (instead of the usual ScummVM logo on orange screen). The only way I can think of which would keep the launch screen correct on builds from both Xcode and make (while maintaining the new full screen support on builds from Xcode) would be to create a separate `Info.plist` for each build option so that the one for make builds omits references to the storyboard launch screen - unless anyone can see another solution to this?

criezy:
So it looks like we might indeed want to exclude the storyboard when doing cross-compilation. One way to do that without duplicating the `Info.plist` would be to add a `sed` command in the `ios7bundle` target in `ports.mk` to strip the `UILaunchStoryboardName` and following value in the `Info.plist` file.

criezy:
I have manually merged this with a small change (the `Info.plist` file is generated from the `Info.plist.in` file, so we actually need to make the change to the `Info.plist.in` file).

I have also added a commit to remove the UILaunchStoryboardName from the Info.plist while when running the `ios7bundle` make target.

Thanks again for your contribution.

PR #2423, opened 2020-08-24
AUDIO: fix uninitialized variable

darkstar:
Fix some uninitialized variables in the soundfont code that trigger a warning in MSVC


sev-:
This is a false positive from MSVC because all bits are covered by the switch. Still, it is a good practice to init variables, but I do not understand why did you change scope of those two variables? They're not used outside of the `while()` loop and get fully initialized at every cycle.

I would rather simply add a `default:` case for shutting the warning. I'll do it separately.

Thanks for the heads up.

PR #2424, opened 2020-08-24
NEVERHOOD: Big demo

henke37:
This adds support for a new demo version of the game.

In addition to the detection entry, it fixes the game logic to allow for fully accessing the demo.


sev-:
Thanks!

PR #2425, opened 2020-08-25
SCI: Hebrew - add prefix rule (second letter is Yud)

ZvikaZ:


sluicebox:
Looks good and it's contained to the Hebrew processing code, merging.

PR #2426, opened 2020-08-26
AMIGAOS4: Reorder .mk file to fix readme.guide

raziel-:
The created readme.guide file features some in-guide links that point to local files (I.e. AUTHORS), which need to be in the same directory level.
It will otherwise silently fail, when a user clicks on said links.

Due to all documents (except the readme.guide) being installed into the doc/ subdirectory since a few months now, the in-guide links stopped working.
We fix that by restructure the install order to
1) place the .guide creation into the documents loop
2) make sure the /doc subdirectory is available before the guide creation is processed and
3) place the readme.guide into the /doc subdir to secure working in-guide links


sev-:
Thanks

PR #2427, opened 2020-08-26
TESTBED: Test networking

henke37:
There seems to be a need for a test suite for non webserver related networking functionality.

To start off it checks if the `OSystem::isConnectionLimited` member function works.

### DIFF
--- a/engines/testbed/networking.cpp
+++ b/engines/testbed/networking.cpp
@@ -1,0 +1,56 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


engines/testbed/networking.cpp
```diff
@@ -0,0 +1,34 @@
+#include "testbed/networking.h"
```
sev-, line 1: This file is missing ScummVM copyright

engines/testbed/networking.cpp
```diff
@@ -0,0 +1,56 @@
...
+	if (ConfParams.isSessionInteractive()) {
+		if (Testsuite::handleInteractiveInput("Testing the IsConnectionLimited API implementation", "Continue", "Skip", kOptionRight)) {
+			Testsuite::logPrintf("Info! IsConnectionLimited test skipped by the user.\n");
+			return kTestSkipped;
+		}
+
+		bool isLimited = g_system->isConnectionLimited();
+
+		if (ConfParams.isSessionInteractive()) {
+			// Directly verify date
```
criezy, line 41: Is that a copy/paste from another test? This comment doesn't seem to make much sense here ;-)

criezy (commented):
Thank you. Having network tests could indeed be useful. I just have one minor comment on a comment that seems out of place. Otherwise it looks good.

Do you have other network tests in mind that would make sense to add?

engines/testbed/networking.cpp
sev-: I'll clean it up after the merge.

sev-:
Thank you! Hope you will have time/enthusiasm for implementing deeper tests. Merging.

henke37:
I have been meaning to clean it up myself, but thanks.

As for more tests, I was thinking of at least pilfering the get ip test from the webserver group. Bit tricky when the underlying code there is part of the webserver code.

PR #2431, opened 2020-08-30
WME: Added Commandline helper plugin for "Pizza Morgana: Episode 1" game

aquadran:
Added basic support for Commandline helper plugin (wme_commandlinehelper.dll) used by "Pizza Morgana: Episode 1" game.
This allow start game in free demo mode.

### DIFF
--- a/engines/wintermute/ext/wme_commandlinehelper.cpp
+++ b/engines/wintermute/ext/wme_commandlinehelper.cpp
@@ -1,0 +1,102 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
+ * file distributed with this source distribution.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ *
+ */
+
+/*
+ * This file is based on WME Lite.
+ * http://dead-code.org/redir.php?target=wmelite
+ * Copyright (c) 2011 Jan Nedoma
+ */
+
+#include "engines/metaengine.h"
+#include "engines/wintermute/wintermute.h"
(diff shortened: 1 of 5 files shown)


engines/wintermute/ext/wme_commandlinehelper.cpp
```diff
@@ -0,0 +1,102 @@
...
+//////////////////////////////////////////////////////////////////////////
+ScValue *CommandLineHelper::scGetProperty(const Common::String &name) {
+	_scValue->setNULL();
+
+	//////////////////////////////////////////////////////////////////////////
+	// Parameters (RO)
+	// Used to launch demo: "Pizza Morgana: Episode 1 - Monsters and Manipulations in the Magical Forest"
+	//////////////////////////////////////////////////////////////////////////
+	if (name == "Parameters") {
+		_scValue->setString("Pizza.exe FULLDEMO");
```
lolbot-iichan, line 79: I guess we could do something like this:

```
if (BaseEngine::instance().getFlags() & ADGF_DEMO) {
    _scValue->setString("Pizza.exe FULLDEMO")
} else {
    _scValue->setString("Pizza.exe FULL")
}
```

lolbot-iichan:
Looks like there is one place outside /ext/ folder that I forgot to move to plugins.h:

https://github.com/scummvm/scummvm/blob/78b36f248b4d11446b4614decf7064eabceff1b4/engines/wintermute/persistent.cpp#L163

For now you can add REGISTER_CLASS(CommandLineHelper, false) line to persistent.cpp.
I can move REGISTER_CLASS'es for plugin classes to /ext/ later.

engines/wintermute/ext/wme_commandlinehelper.h
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef WINTERMUTE_COMMANDLINEHELPER_H
+#define WINTERMUTE_COMMANDLINEHELPER_H
+
+#include "common/str.h"
+#include "engines/wintermute/base/base_scriptable.h"
+
+namespace Wintermute {
+
+class CommandLineHelper : public BaseScriptable {
```
lolbot-iichan, line 37: I suggest using SX prefix, since it's a scriptable object like SXSteamAPI, SXDirectory, SXString, SXArray, etc.

engines/wintermute/ext/wme_commandlinehelper.cpp
aquadran: It seems game is both full and demo same time. So this will not work. Full game require some authentication which is not part of scope here.

engines/wintermute/ext/wme_commandlinehelper.h
aquadran: done

engines/wintermute/ext/wme_commandlinehelper.cpp
lolbot-iichan: OK. It seems we can't do much in this case.

engines/wintermute/ext/wme_commandlinehelper.h
lolbot-iichan: Thanks!

PR #2432, opened 2020-08-30
COMMON: U32: Fix different types of string pointers for u32

aryanrawlani28:
Build broke for AmigaOS due to varying types of pointers of the internal string of U32. This attempts to fix that.


criezy:
Thanks. The change looks correct. Merging.

PR #2436, opened 2020-08-30
Use native language names

criezy:
Now that the GUI uses UTF-32, we no longer need to restrict language names to ASCII characters. This means we can use accentuated characters and non-latin alphabets.

This pull request proposes to do that.
Possible issues:
- I pulled the language names from the wikipedia language selection menu. Hopefully this is correct, but it would be good to check.
- I only implemented a basic sorting of `U32String` based on the character values. This is not ideal, so if somebody wants to implement proper unicode string sorting, that would be welcome. But I considered that it was overkill just for a language selection menu. I only considered doing the sorting based on the language code, but this is not ideal either (for example we end up having Suomi between Euskara and Français).

image

### DIFF
--- a/common/translation.cpp
+++ b/common/translation.cpp
@@ -42,7 +42,7 @@ namespace Common {
 DECLARE_SINGLETON(TranslationManager);
 
 bool operator<(const TLanguage &l, const TLanguage &r) {
-	return strcmp(l.name, r.name) < 0;
+	return l.name < r.name;
 }
 
 TranslationManager::TranslationManager() : _currentLang(-1) {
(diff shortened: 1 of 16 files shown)


ZvikaZ:
That's a good idea, IMO. And I've verified that Hebrew is indeed spelled correctly.

jepael:
I confirm Suomi and Svenska are written correctly.

bluegr:
I conform that Greek is written correctly.

I also propose to merge this, if there are any typos they can be fixed later on.

Kawa-oneechan:
Nederlands klopt.

sev-:
I was looking for this for so long... finally we can do it.

raziel-:
Deutsch ist korrekt.

Hehehe :-D

SupSuper:
Português looks good to me!

lotharsm:
Español, Català, Galega and Euskara are correct too.

sev-:
Let me just finish this :P. BTW, Belarussian, Ukrainian, Russian and Polish are all correct.

PR #2438, opened 2020-08-31
CINE: Fix for bug #10643 to branch-2-2 ("Fix overlay rendering with NULL source mask").

karisal:
In French Amiga Future Wars from bug #10643 when walking left from the
scene with the open manhole cover a call to renderOverlay with a
type 0 (color sprite) overlay happens. The source mask for the overlay
is NULL and a memcpy using a NULL pointer as source is initiated.
This results in a memory access violation.

This fixes that memory access violation by simply setting the
destination mask to all zeroes when the source mask is NULL.
Seems to work fine at least in this case.

Fixes bug #10643.


bluegr:
Nice! Merging

sev-:
This also needed to be committed to `master`. I just did that

PR #2439, opened 2020-08-31
SCUMM: Fix compilation for DS & add translations for the DS Help Dialog, Update scummvm.pot

aryanrawlani28:


sev-:
Thanks, however, you forgot to update the .po files. I'll do it.

PR #2440, opened 2020-09-01
COMMON: Standardize language names and codes

trembyle:
Please see the additional documentation in language.cpp. The goal is to use more consistent naming conventions based wherever possible on ISO standards.

Some existing entries needed to be changed in common and also in individual engines. I was able to build successfully and detect and run games in these languages. I'm looking for feedback on this. Is there anything else that I need to consider?


sev-:
Let's imagine I already added Chinese DIG to ScummVM (I really did yesterday). It is stored in scummvm.ini with 'language=zh', which was ZH_TWN. Now, that language is parsed to ZH_ANY. But in SCUMM engine code there are no checks against ZH_ANY, and the game will stop working.

Thus, a separate step for converting scummvm.ini is required. We did that in the past when we made some targets obsolete. See engines/obsolete.cpp. In order to avoid multiple conversions, I would do the following: Check `versioninfo` field in the INI file, if it is 2.2.* or less, then iterate over all targets and convert language, then save the file. Saving file will update versioninfo, and you will avoid double conversion.

common/language.h
```diff
@@ -34,11 +34,11 @@ class String;
  */
 enum Language {
 	ZH_ANY,     // Generic Chinese (when only one game version exist)
```
lolbot-iichan, line 36: I guess those could be sorted as well.

common/language.cpp
```diff
@@ -41,26 +63,25 @@ const LanguageDescription g_languages[] = {
...
+	{    "lv", "lv_LV", "Latvian", LV_LVA },
 	{    "nb", "nb_NO", "Norwegian Bokm\xE5l", NB_NOR },
-	{    "fa", "fa_IR", "Persian (Iran)", FA_IRN },
+	{    "fa", "fa_IR", "Persian", FA_IRN },
 	{    "pl", "pl_PL", "Polish", PL_POL },
 	{    "br", "pt_BR", "Portuguese (Brazil)", PT_BRA },
-	{    "pt", "pt_PT", "Portuguese (Portugal)", PT_POR },
+	{    "pt", "pt_PT", "Portuguese (Portugal)", PT_PRT },
 	{    "ru", "ru_RU", "Russian", RU_RUS },
+	{    "sr", "sr_RS", "Serbian", SR_SRB },
```
lolbot-iichan, line 79: Maybe we need "sr_RS@latin" here, since "sr_RS" seems to be cyrillic locale and the only currently supported game in Serbian is using latin script.

common/language.cpp
```diff
@@ -41,26 +63,25 @@ const LanguageDescription g_languages[] = {
 	{    "fi", "fi_FI", "Finnish", FI_FIN },
 	{    "fr", "fr_FR", "French", FR_FRA },
 	{    "de", "de_DE", "German", DE_DEU },
-	{    "gr", "el_GR", "Greek", GR_GRE },
+	{    "el", "el_GR", "Greek", EL_GRC },
 	{    "he", "he_IL", "Hebrew", HE_ISR },
-	{    "hb", "he_IL", "Hebrew", HE_ISR }, // Deprecated
 	{    "hu", "hu_HU", "Hungarian", HU_HUN },
 	{    "it", "it_IT", "Italian", IT_ITA },
-	{    "jp", "ja_JP", "Japanese", JA_JPN },
```
lolbot-iichan, line 49: I just realized that changing 1st column may break some ConfMan and AdvancedDetector related stuff with games played with previous versions of ScummVM. See `getLanguageCode` function.

lolbot-iichan:
> Thus, a separate step for converting scummvm.ini is required. We did that in the past when we made some targets obsolete. See engines/obsolete.cpp. In order to avoid multiple conversions, I would do the following: Check `versioninfo` field in the INI file, if it is 2.2.* or less, then iterate over all targets and convert language, then save the file. Saving file will update versioninfo, and you will avoid double conversion.
Besides ConfMan's language records and AdvancedDetector target names, there is one more scenario: engines/glk/quetzal.cpp is writing language codes to savenames. Not sure if it's fine or not.

common/language.cpp
trembyle: The right way might be to include another parameter for script. We also don't differentiate between Traditional and Simplified Chinese either, but it looks like both are available in Wintermute.

trembyle:
> Besides ConfMan's language records and AdvancedDetector target names, there is one more scenario: engines/glk/quetzal.cpp is writing language codes to savenames. Not sure if it's fine or not.

Since this is the first release to include GLK, that shouldn't affect any older saved games from prior releases.

common/language.h
trembyle: Good point!

trembyle:
> Thus, a separate step for converting scummvm.ini is required. We did that in the past when we made some targets obsolete. See engines/obsolete.cpp. In order to avoid multiple conversions, I would do the following: Check `versioninfo` field in the INI file, if it is 2.2.* or less, then iterate over all targets and convert language, then save the file. Saving file will update versioninfo, and you will avoid double conversion.

I think I got close here, but it's not quite working yet. I'm unable to detect some games with new codes. Can you see what I'm doing wrong?

sev-:
What are "some" games?

sev-:
> > Besides ConfMan's language records and AdvancedDetector target names, there is one more scenario: engines/glk/quetzal.cpp is writing language codes to savenames. Not sure if it's fine or not.
> 
> Since this is the first public release to include GLK, users shouldn't expect saved games from prior development builds to work in stable 2.2.0.

No, this change cannot go to 2.2.0, it is way too risky so close before the release.

trembyle:
> Besides ConfMan's language records and AdvancedDetector target names, there is one more scenario: engines/glk/quetzal.cpp is writing language codes to savenames. Not sure if it's fine or not.

GLK has only one game in an affected language (nissen). This is only available in Danish, and the two character code for that language is not changing.

common/language.h
lolbot-iichan: Thanks!

common/language.cpp
lolbot-iichan: >We also don't differentiate between Traditional and Simplified Chinese either
In WME, "Chinese (China)" and "Chinese (Taiwan)" are used in "Simplified Chinese" and "Traditional Chinese" meaning respectfully. As far as I understand, those things are pretty much the same:
- Simplified Characters were introduced by China and are also used at Malaysia and Singapore, China is much bigger then Malaysia and Singapore in both area and population.
- Traditional Characters are still used at Taiwan, Hong Kong and Macau, Taiwan is much bigger then Hong Kong and Macau in both area and population.
I guess "Chinese (China)" and "Chinese (Taiwan)" can be renamed to "Chinese (Simplified)" and "Chinese (Traditional)".

ZH_ANY was added as a workaround for a WME game Reversion that contained 3 Chinese translation packages in a single game directory after an update: new Simplified and Traditional translations were added and old Chinese package was not deleted (and detected by detection.cpp). To fix this, we decided to mark all old Reversion Chinese games as ZH_ANY ("generic" Chinese) and use ZH_CHA/ZH_TWN for Simplified/Traditional.

sev-:
I think, it is now a good time to merge this, but it got bitrot in the meantime. Could you please rebase?

trembyle:
> I think, it is now a good time to merge this, but it got bitrot in the meantime. Could you please rebase?

Yes, I will work on fixing the config file updates for older language codes first, then rebase.

After the merge, I'll review engine detection entries for any codes that are changing so that they meet the new standard.

trembyle:
I'm getting closer on the functionality, but it's not there yet. I pushed just to show some progress. Only the bare minimum rebase has happened, just to get it running. More cleanup (splitting & squashing) of commit history is needed.

I'll keep working on this. Certainly I'm not expecting a review until it's complete, although if anyone is feeling charitable, pointers will be gladly accepted. In particular, I need to figure out where to put this new code. It's not getting called from `GUI::LauncherDialog::loadGame` as I expected.

trembyle:
@sev- It should be working now.

sev-:
Thank you!

PR #2441, opened 2020-09-02
[RFC] Prefer in-memory SoundFont data over global setting in FluidSynth driver

eriktorbjorn:
Blazing Dragons provides a custom SoundFont to the FluidSynth driver. Unfortunately, I can't find any sensible way to use that because my global SoundFont setting takes precedence over the in-memory data. Even overriding the MIDI settings for Blazing Dragons and leaving the SoundFont setting blank did not help.

What I've tried to do here is to soften the check so that FluidSynth will prefer in-memory data unless a SoundFont has been explicitly configured for the current game.

Does this seem like a sensible solution? If so, should it also go into the 2.2 branch?


sev-:
Yes, it makes sense. Merging

PR #2442, opened 2020-09-02
CINE: OS: Fallback to default error messages

karisal:
If the error messages can not be found (They are in the file
ERRMESS.DAT in Operation Stealth) then use the default values.

There are now the following default error messages included in ScummVM
for Operation Stealth:
 - American English (Protagonist is James Bond)
 - British English (Protagonist is John Glames)
 - French
 - German

For the Italian and Spanish releases of Operation Stealth the error
messages from Future Wars are used as fallback. This is incorrect
but better than nothing.

This makes the error message handling more lenient than before so that
not having an ERRMESS.DAT does not block the starting of the game
anymore.

Fixes bug #11649.

### DIFF
--- a/engines/cine/texte.cpp
+++ b/engines/cine/texte.cpp
@@ -186,6 +186,78 @@ void initLanguage(Common::Language lang) {
 		"A wall of silence ..."
 	};
 
+	/**
+	 * British English error messages for Operation Stealth.
+	 */
+	static const char *const failureMessages_OS_EN[] = {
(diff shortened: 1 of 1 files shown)


karisal:
Please whoever merges this into the master could you
cherry-pick this to the branch-2-2 or tell me how to do
it myself (I have no write access to the master or branch-2-2
repository)? Thank you!

Kari Salminen

sev-:
Thank you

PR #2443, opened 2020-09-03
SCI: Fix #11671 - add `applyCount` to vm_hooks

ZvikaZ:


bluegr:
I've reverted the original script patch. I'm not happy with the way this turned out - we still don't know why this bug occurs in ScummVM, so patching game scripts in this manner, without knowing the actual reason, is wrong. We've discussed this with @sluicebox, and we both agree on this course of action.

Closing this. Feel free to submit the rest of the patch (applyCount)

PR #2444, opened 2020-09-03
CINE: OS: Fix MT-32/AdLib SFX & loading savegames

karisal:
In Operation Stealth DOS and Atari ST versions, but not Amiga or
demo versions, sound effects are loaded in AUTO00.PRC using a
combination of o2_loadAbs and o2_playSample(1, ...) before
o1_freePartRange(0, 200). This is a better heuristic for ending the
script at the correct position than checking whether global variable
255 is compared to 0. In the original game AUTO00.PRC was run when
starting or restarting the game and one could not load a savegame
before passing the copy protection. Thus, we try to emulate that
behaviour by running at least part of AUTO00.PRC before loading a
savegame.


karisal:
After the merge could you please also cherry-pick this
to the branch-2-2 branch? Thank you!

Kari Salminen

sev-:
Thank you, merging

PR #2448, opened 2020-09-06
BASE: WIN32: Exclude scummvm's base directory from plugin path

lotharsm:
Currently, the plugin architecture searches for plugins in ScummVM's
base directory and in the plugins directory itself.

However, for Win32, we need to bundle several DLL files in order to
make ScummVM run at all. Currently, this leads to several warning
messages since ScummVM tries to load those DLLs as plugins.

This patch excludes the ScummVM base directory when running on Win32 and
only accepts plugins from the plugins directory.

### DIFF
--- a/base/plugins.cpp
+++ b/base/plugins.cpp
@@ -159,7 +159,9 @@ PluginList FilePluginProvider::getPlugins() {
 	Common::FSList pluginDirs;
 
 	// Add the default directories
+	#ifndef WIN32
 	pluginDirs.push_back(Common::FSNode("."));
+	#endif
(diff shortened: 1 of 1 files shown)


henke37:
I think this was by design. Bad design. I support this fix.

lotharsm:
@aryanrawlani28 Would this change in any shape or form interfere with your plans regarding the plugins system? Otherwise, I'll merge this to master soon.

aryanrawlani28:
@lotharsm Thanks for letting me know! No, it doesn't really effect my plugins PR. Feel free to merge :)

PR #2454, opened 2020-09-09
CINE: 9 fixes to music, UI, savegame etc

karisal:
Hi!
Here are some fixes to the Cine-engine. I think they should be
quite safe to put into the 2.2.0 release branch too (I used
them on the branch-2-2 while testing locally. I played through
Italian Amiga version of Future Wars while working on them).

Kari Salminen


criezy:
Thank you. Merging.

PR #2457, opened 2020-09-11
DOXYGEN: Add doxygen build files

b-gent:
This commit adds the doxygen build framework to the main
scummvm repo.
This allows to easily build and test API doc when making any
kind of changes in doxygen blocks in header files.

The build can be run from command line in the 'doxygen'
folder with 'doxygen scummvm.doxyfile'.

Output is produced in the 'html' folder and can be viewed
by opening the 'index.html' file.

The 'buildfiles' folder contains files related to style.


b-gent:
Retry DeepCode

doc/doxygen/buildfiles/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Pieter Conradie <https://piconomix.com>
```
Mataniko, line 3: What is this license file for?

doc/doxygen/Doxyfile.doxyfile
```diff
@@ -0,0 +1,2499 @@
...
+
+# If the CREATE_SUBDIRS tag is set to YES then doxygen will create 4096 sub-
+# directories (in 2 levels) under the output directory of each output format and
+# will distribute the generated files over these directories. Enabling this
+# option can be useful when feeding doxygen a huge amount of source files, where
+# putting all generated files in the same directory would otherwise causes
+# performance problems for the file system.
+# The default value is: NO.
+
+CREATE_SUBDIRS         = NO
```
Mataniko, line 71: may be needed TRUE with our amount of files

doc/doxygen/Doxyfile.doxyfile
```diff
@@ -0,0 +1,2499 @@
...
+# the logo to the output directory.
+
+PROJECT_LOGO           = buildfiles/scummvm_logo.png
+
+# The OUTPUT_DIRECTORY tag is used to specify the (relative or absolute) path
+# into which the generated documentation will be written. If a relative path is
+# entered, it will be relative to the location where doxygen was started. If
+# left blank the current directory will be used.
+
+OUTPUT_DIRECTORY       = html/
```
Mataniko, line 61: /var/www/doxygen

doc/doxygen/groups.dox
```diff
@@ -0,0 +1,14 @@
+/**
+
+@defgroup common Common API
+@brief Common functions and structures.
+
+*/
+
+/**
+
+@defgroup lua Lua API
```
Mataniko, line 10: This looks like it belongs from another project

doc/doxygen/links.doxyfile
```diff
@@ -0,0 +1,4 @@
+ALIASES += \
+linkLua="<a href=\"https://www.lua.org\" target=\"_blank\">Lua.org</a>" \
```
Mataniko, line 2: IS this also from another project?

doc/doxygen/scummvm.doxyfile
```diff
@@ -0,0 +1,2484 @@
+# Doxyfile 1.8.14
```
Mataniko, line 1: Why are there two doxyfiles?

doc/doxygen/links.doxyfile
Mataniko: I see where this is used now while reviewing the other PR

doc/doxygen/Doxyfile.doxyfile
b-gent: I agree.
I've run a lot of builds locally so far and I have changed this and some other parameters since this was opened. I will push it soon, still working on potential improvements to the build process and output.

doc/doxygen/Doxyfile.doxyfile
b-gent: Is it needed here? I understand this is the path on the Linux machine that will host this?
Not sure what the build process was so far - doxygen built and hosted on the same server, or built somewhere else and copied to the server?
My idea of providing the config file is that everyone can build doxygen locally to check how their changes are reflected in doc output. Building into ``html`` (relative path) would be preferable in such case in my opinion, as you can quickly jump to the output folder from the folder where you just ran the command.
The build server could have a script that replaces this configuration option, or use its own config file, or copy the output to the right directory afterwards.
Let me know what you think @Mataniko , maybe I am misunderstanding this

doc/doxygen/groups.dox
b-gent: This is a doxygen group definition for all headers in the ``common/lua`` folder. So it's related to this change from the other PR -> https://github.com/scummvm/scummvm/pull/2361/files#diff-30c89c991aba94a200fe79266635efa2

This file is a central dox file to manage groups, for example when there are multiple header files that need a 'parent' group, you would define the group here instead of an arbitrary header file. The file can also be used to manipulate the hierarchy of different groups.

I can move it to PR#2361 if you prefer, but they are really both complementary PRs and none will work properly without the other.

doc/doxygen/scummvm.doxyfile
b-gent: My bad, I overlooked this. I was experimenting with having two different doxyfiles with slightly different configurations (for example, for local builds and for the web server), but only one should have been committed here. I will remove.

This one (scummvm.doxyfile) is actually the corrent one.

doc/doxygen/buildfiles/LICENSE
b-gent: I was looking for a nice doxygen customization (haven't done much of that before, usually used the default) and I ended up borrowing the doxygen theme from this project -> https://github.com/piconomix/doxygen-github-markdown-theme

I'm treating this whole CSS as a starting point really, and will be working on making the doxygen output more consistent with the look-and-feel of the ScummVM website.

I don't have much experience with licenses and I'm not sure if this should be kept, considering that we won't be using this as-is.

b-gent:
I made some significant changes related to style (as this week of GSoD should be devoted to it). Sample output can be viewed here:
https://b-gent.github.io/PR2457/da/dc5/class_common_1_1_archive.html

The idea was to make it look slightly more like the ScummVM website, using the same colors.

I also made some modifications in the doxygen config file, changing configurations like:
SORT_MEMBER_DOCS       = NO
STRIP_FROM_PATH        = ../../
CREATE_SUBDIRS         = YES
,in order to fine-tune the output.

OUTPUT_DIRECTORY is still set to just 'html' so we need to clarify this.

criezy:
What we do at work for our doxygen config files is that we use an environment variable for our OUTPUT_DIRECTORY and we have a `make doc` target to run doxygen that sets this variable (to doc/html/ in our case). So when developer want to test things they run `make doc`. But on the server it sets the environment variable to a different pathand directly runs doxygen (and not through make).

Maybe something similar could be done here?

b-gent:
> What we do at work for our doxygen config files is that we use an environment variable for our OUTPUT_DIRECTORY and we have a `make doc` target to run doxygen that sets this variable (to doc/html/ in our case). So when developer want to test things they run `make doc`. But on the server it sets the environment variable to a different pathand directly runs doxygen (and not through make).
> 
> Maybe something similar could be done here?

Thanks for the comment @criezy , always good to know someone is looking at this :)

Make is definitely an option, I think I also saw this in some project a couple years back. I don't have experience with setting up make though.

I also found this -> https://stackoverflow.com/a/49148112
, which seems very straightforward to implement in a simple bash script. It's based on the same idea of environment variables.

b-gent:
I reverted the doygen skin to the vanilla version, as agreed on Discord.
I am now figuring out the best way to use a variable for output directory, using my limited scripting skills.

criezy:
I merge this manually and also added a commit to
- use a variable for the output directory
- add a make target to run doxygen: `make doxygen`
- and add a convenient script: `doc/doxygen/run_doxygen.sh`

With the script you can also set the `DOXYGEN_OUTPUT_DIRECTORY` environment variable to the path where you want to generate the documentation if the default is not what you want. For example `DOXYGEN_OUTPUT_DIRECTORY=/var/www/doxygen doc/doxygen/run_doxygen.sh`

PR #2458, opened 2020-09-12
ENGINES: Change mac resource fork file detection to use the file cache

bgK:
Common::MacResMan is now able to open files from a specified
Common::Archive. This is a bit hacky as dynamic_cast is used to break
the Archive encapsulation to retreive the underlying FSNode. It should
however be more correct than the previous code that assumed files were
at the root of the currently running game's path.

AdvancedDetector constructs a Common::Archive from its FileMap based
filesystem cache and uses it to detect the mac resource fork files
instead of repeatedly querying the filesystem for file presence.

This cuts the time it takes to run the detection code with all the
engines enabled as dynamic plugins on the 3DS to 30 s from 280 s.

This seems to work properly as I've been able to play Pegasus Prime
on Linux and to detect the Director version of The Journeyman Project
on macOS. However I'd be glad if someone familiar could check it.

### DIFF
--- a/common/macresman.cpp
+++ b/common/macresman.cpp
@@ -116,15 +116,20 @@ String MacResManager::computeResForkMD5AsString(uint32 length) const {
 }
 
 bool MacResManager::open(const String &fileName) {
+	return open(fileName, SearchMan);
+}
+
+bool MacResManager::open(const String &fileName, Archive &archive) {
 	close();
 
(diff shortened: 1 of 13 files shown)


bgK:
I'll merge this tomorrow unless someone needs some more time to review it.

bgK:
Merged  as 0f2c6cd.

PR #2459, opened 2020-09-13
PSP2: Simplify the filesystem code

bgK:
The Vita SDK gained support for dirent at the end of 2017. Our copy
is no longer necessary. 

DrivesPOSIXFilesystemFactory allows to specify the contents of the
pseudo-root file system node. There is no need to hardcode them in
posix-fs.cpp anymore.

I have tested this on a Vita. Seems to work fine.


sev-:
If you want this to be in 2.2.0, then there are only a few hours left before the tagging.

bgK:
> If you want this to be in 2.2.0, then there are only a few hours left before the tagging.

This is just a cleanup, there is no point in including it in the 2.2.0 release.

rsn8887:
I am wondering if this is related to a bug that plagued PSP, 3DS and Switch, where an fseek in The Dig caused crashes:
See
https://bugs.scummvm.org/ticket/11342 
and
https://bugs.scummvm.org/ticket/11384

The fix to these bugs was to use the ScummVM internal filesystem implementation, instead of native SDK-provided one. 

I think it might be good to carefully check that the changes in this PR don't cause problems with fseek similar to the above cited bugs.

bgK:
This PR does not change the file stream implementation, it only changes the listing of files and folders. I have tested the scene that showed the issue on the other portable console platforms in The Dig. It still works fine.

rsn8887:
Ok then I am all for it.

PR #2463, opened 2020-09-14
SCI: PQ4 - Partial fix for #11660 (keep speech and subtitles setting)

ZvikaZ:
Issue #11660 is caused by signTalker overriding global90,
which keeps speech and subtitles settings.
That's fixed in this patch.

Similar problem can also be caused by computerTalker.
Waiting to see it happening, before fixing.

### DIFF
--- a/engines/sci/engine/script_patches.cpp
+++ b/engines/sci/engine/script_patches.cpp
@@ -10774,6 +10774,46 @@ static const uint16 pq4CdSpeechAndSubtitlesPatch[] = {
 	PATCH_END
 };
 
+// Applies to: PC CD, Mac CD
+// Responsible methods: signTalker:init, signTalker:dispose,
+//                      computerTalker:init, computerTalker:dispose
+// Fixes bug: #11660
+static const uint16 pq4CdSpeechAndSubtitlesTalkerSignature1[] = {
+	0x89, 0x5a,                     // lsg 5a
+	0x35, 0x02,                     // ldi 02
+	0x12,                           // and
+	SIG_MAGICDWORD,
+	0x31, 0x04,                     // bnt 04
+	0x35, 0x03,                     // ldi 03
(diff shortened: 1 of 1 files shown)


ZvikaZ:
Note: in this patch I assumed that `global190` isn't used.
I'd be glad if someone can verify me.

ZvikaZ:
I want to improve few things here, do not merge yet.
Thanks.

ZvikaZ:
It took too much time...
But now it's ready for review.

Thanks

engines/sci/engine/script_patches.cpp
```diff
@@ -10774,6 +10774,46 @@ static const uint16 pq4CdSpeechAndSubtitlesPatch[] = {
...
+	0x35, 0x03,                     // ldi 03
+	0xa1, 0x5a,                     // sag 5a
+	SIG_END
+};
+
+static const uint16 pq4CdSpeechAndSubtitlesTalkerPatch1[] = {
+	0x80, PATCH_UINT16(0x005a),     // lag 005a
+	0xa0, PATCH_UINT16(0x00c0),     // sag 00c0 [ save message mode ]
+	0x39, 0x01,                     // pushi 01
+	0x14,                           // or [ enable text ]
```
bluegr, line 10796: I don't understand this bit. Why is the or opcode needed here?

bluegr:
Nice work! Much much cleaner now. Please check the comment I made, and possibly add a relevant comment

engines/sci/engine/script_patches.cpp
bluegr, line 10796:
```suggestion
	0x14,                           // or [ enable text ] - always display text, regardless of the mode
```

bluegr: @sluicebox  explained this, so I've added a comment to this bit

bluegr:
Thanks for this. Merged

PR #2466, opened 2020-09-15
AMIGAOS/MORPHOS: Add non-debug build optimization and fix ordering

raziel-:
- Stole a setting from MorphOS :-)
- Moved MorphOS further down to keep the alphabetic order intact ;-)

Also, indented comments seem to work now in AmigaOS shell


sev-:
Thanks. Should it be backported for the potential 2.2.1?

raziel-:
@sev-

Backporting would be nice, des.

Thank you

PR #2467, opened 2020-09-15
[DNM] DOXYGEN: updates in header files 1

b-gent:
Add and improve doxygen blocks in the header file.

Just opening a PR with a single file to get some feedback from you if this is the way to proceed.

Output is available on https://b-gent.github.io/PR2467/index.html
@sev- @Mataniko

### DIFF
--- a/common/achievements.h
+++ b/common/achievements.h
@@ -51,58 +51,135 @@ enum AchievementsPlatform {
 
 
 /**
- * Per-game achievements information structure item.
+ * Information structure for game-specific achievements.
  */
 struct AchievementDescription {
-	const char *id;            // achievement internal id, e.g. "ACHIEVEMENT_TIMING"
-	bool isHidden;             // achievement is hidden
-	const char *title;         // achievement displayed text, e.g. "Marathon Runner"
-	const char *comment;       // optional achievement hint / comment, e.g. "Finish the game in less than 4 hours"
+	const char *id;            //!< Achievement internal ID, such as "ACHIEVEMENT_TIMING".
+	bool isHidden;             //!< Whether the achievement is hidden.
+	const char *title;         //!< Achievement displayed text, such as "Marathon Runner".
+	const char *comment;       //!< Optional achievement hint or comment, such as "Finish the game in less than 4 hours".
 };
 
 
 /**
- * Per-game achievements information structure item.
+ * Information structure for platform-specific achievements.
  */
 struct AchievementsInfo {
-	Common::AchievementsPlatform platform;              // achievements platform, e.g. STEAM_ACHIEVEMENTS
-	Common::String appId;                               // achievements application ID of given platform
-	Common::Array<AchievementDescription> descriptions; // descriptions of all game achievements
+	Common::AchievementsPlatform platform;              //!< Achievements platform, such as "STEAM_ACHIEVEMENTS".
+	Common::String appId;                               //!< Achievements application ID of the given platform.
+	Common::Array<AchievementDescription> descriptions; //!< Descriptions of all game achievements.
 
 	AchievementsInfo() {platform = Common::UNK_ACHIEVEMENTS;}
 };
 
+/**
+ * Class for manipulating the achievements.
+ *
+ * Use the Achievements Manager class to edit the in-game achievements.
+ */
 
 class AchievementsManager : public Singleton<AchievementsManager> {
 public:
 	AchievementsManager();
 	~AchievementsManager();
-
+    
+    /**
+     * Set a platform and application ID as active domain.
+     *
+     * @param[in] platform Achievements platform.
+     * @param[in] appId    Achievements application ID of the given platform.
+     */
 	bool setActiveDomain(AchievementsPlatform platform, const String &appId);
-	bool unsetActiveDomain();
-	bool isReady() { return _iniFile != nullptr; }
+	bool unsetActiveDomain();                      //!< Unset the current active domain.
+	bool isReady() { return _iniFile != nullptr; } //!< Check whether the domain is ready.
 
-	// Methods to manipulate individual achievements
+	/**
+     * @name Methods for manipulating individual achievements
+     * @{
+     */
+	 
+	/** Set an achievement.
+     *
+     * @param[in] id               Internal ID of the achievement.
+     * @param[in] displayedMessage Message displayed when the achievement is achieved.
+     */
+
 	bool setAchievement(const String &id, const String &displayedMessage);
+	
+    /**
+     * Set an achievement as achieved.
+     *
+     * @param[in] id Internal ID of the achievement.
+     */	
+	
 	bool isAchieved(const String &id);
+	
+    /**
+     * Clear an achieved achievement.
+     *
+     * @param[in] id Internal ID of the achievement.
+     */	
+	
 	bool clearAchievement(const String &id);
+    
+    /** @} */
+	
+    /**
+     * @name Methods for manipulating individual statistics
+     * @{
+     */
 
-	// Methods to manipulate individual statistics
+    /**
+     * Get a statistic (integer).
+     *
+     * @param[in] id Internal ID of the achievement.
+     */	
+	 
 	int getStatInt(const String &id);
+	
+    /**
+     * Set a statistic to an integer number.
+     *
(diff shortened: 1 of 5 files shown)


common/achievements.h
```diff
@@ -42,58 +42,145 @@ enum AchievementsPlatform {
...
 	bool setActiveDomain(AchievementsPlatform platform, const String &appId);
+	
+    /**
+     * Unset the current active domain.
+     */
+	
 	bool unsetActiveDomain();
+	
+     /**
+     * Set the domain as ready.
```
b-gent, line 92: Not really sure what this does.

common/achievements.h
```diff
@@ -42,58 +42,145 @@ enum AchievementsPlatform {
...
 class AchievementsManager : public Singleton<AchievementsManager> {
 public:
 	AchievementsManager();
 	~AchievementsManager();
-
+    
+	/**
+     * Set a platform and application ID as active domain.
+	 *
+	 * @param[in] platform Achievements platform.
```
b-gent, line 80: The big question is whether we should include documentation for all parameters in every method. I think this is a good practice but it definitely adds a lot of doxygen noise to the header files. If parameters were not documented, most of these class members could be documented with ``//!<`` in the same line, instead of adding a preceding doxygen block.

common/achievements.h
```diff
@@ -42,58 +42,135 @@ enum AchievementsPlatform {
...
+	 * @param[in] appId    Achievements application ID of the given platform.
+     */
 	bool setActiveDomain(AchievementsPlatform platform, const String &appId);
-	bool unsetActiveDomain();
-	bool isReady() { return _iniFile != nullptr; }
+	bool unsetActiveDomain();                      //!< Unset the current active domain.
+	bool isReady() { return _iniFile != nullptr; } //!< Set the domain as ready.
```
b-gent, line 85: Not sure if this is a correct description.

sev-: Checks if the domain is ready.

common/archive.h
```diff
@@ -164,58 +176,56 @@ class SearchSet : public Archive {
...
 	 *
 	 *   "game/itedata"
 	 *
-	 * In this example the code would first try to search for all directories matching
+	 * In this example, the code first tries to search for all directories matching
 	 * "game" (case insensitive) in the path "directory" first and search through all
 	 * of the matches for "itedata" (case insensitive too).
 	 *
-	 * Note that it will add *all* matches found!
+	 * Note that it will add all matches found!
```
sev-, line 202: This supposed to stress the word "all", by putting it in *italic*.

common/archive.h
```diff
@@ -164,58 +176,56 @@ class SearchSet : public Archive {
...
 	}
 
 	/**
-	 * Create and add sub directories by pattern.
+	 * Create and add subdirectories by pattern.
 	 *
-	 * It is also possible to add sub directories of sub directories (of any depth) with this function.
-	 * The path seperator for this case is SLASH for *all* systems.
+	 * It is also possible to add subdirectories of subdirectories (of any depth) with this function.
+	 * The path seperator for this case is SLASH for all systems.
```
sev-, line 216: Same thing here.

sev- (approved):
Looks good. Yes, non-trivial public methods and their parameters would preferably need to be documented.

Trivial ones are, well, trivial :)

common/archive.h
b-gent: I can revert this change if you want but honestly, whether we stress it or not, this sentence pretty much doesn't say anything else and it's impossible to miss the 'all'.

b-gent:
I have two more files changed locally on this branch which I will push today so please wait a bit with approving/merging.

Also, as you can see in the output, for example here:
https://b-gent.github.io/group__common__alg.html
there are a few undocumented structures here. Easy to spot as they are in black, plain text instead of blue, hyperlinked.

There are not so many in the 3 first files but I am now working with arrays.h and about half of the functions there is not described:
https://b-gent.github.io/class_common_1_1_array.html

@sev-  Do you think we have the capability for developers to provide me with these descriptions? Even if not for all of them, we need someone to identify which are the more important ones that need documentation and which we can skip (at least for now).

It seems to me providing a description for everything will be barely possible. A file like arrays.h misses documentation for about 30 structures. And there are 80 headers in the common folder alone.

common/bitstream.h
```diff
@@ -33,26 +33,35 @@
...
  *
  * Such a bit stream reads valueBits-wide values from the data stream and
  * gives access to their bits, one at a time.
  *
  * For example, a bit stream with the layout parameters 32, true, false
- * for valueBits, isLE and isMSB2LSB, reads 32bit little-endian values
+ * for valueBits, isLE and isMSB2LSB, reads 32-bit little-endian values
  * from the data stream and hands out the bits in the order of LSB to MSB.
  */
 template<class STREAM, int valueBits, bool isLE, bool MSB2LSB>
```
b-gent, line 55: This class is nicely documented, only misses docs for two small structures in lines 277 and 281.

b-gent:
I pushed two more files, I think I won't add any more to this PR and open another one instead. Splitting this into more PRs will make it easier to manage and review.

The output on github.io is also updated.

@sev-  please re-check and address my comments.

b-gent:
``achievements.h`` is in a very good shape. More work is required on the other header files from this PR.
I need descriptions for the following structures:

algorithm.h:
-------------
- unsigned int distance(T *first, T *last)
- unsigned int distance(T first, T last)
- T *sortChoosePivot(T *first, T *last)
- T sortChoosePivot(T first, T last)
- T sortPartition(T first, T last, T pivot, StrictWeakOrdering &comp)
(these are all in lines 207-236)

archive.h:
-----------
- typedef SharedPtr<ArchiveMember> ArchiveMemberPtr; (line 64) (shared pointer to an archive member?)
- typedef List<ArchiveMemberPtr> ArchiveMemberList; ( line 65) (list archive members?)
- bool operator()(const ArchiveMemberPtr &a, const ArchiveMemberPtr &b) (line 68)
- SeekableReadStream *createReadStream() const (line 90)
- friend class Singleton<SingletonBaseType> (line 282)

array.h
----------
- typedef T *iterator;
  typedef const T *const_iterator;
  typedef T value_type;
  typedef uint size_type;    (lines 56-61)
- Array(const Array<T> &array) (line 89)
- void insert_at(size_type idx, const T &element) (line 172)
- void insert_at(size_type idx, const Array<T> &array) (line 177)
- T remove_at(size_type idx) (line 189)
- T &operator[](size_type idx) (line 201)
- Array<T> &operator=(const Array<T> &array) (line 211)
- size_type size() (line 223)
- void clear() (line 227)
- iterator erase(iterator pos) (line 234)
- bool empty() (line 242)
- bool operator==(const Array<T> &other) const (line 246)
- bool operator!=(const Array<T> &other) const (line 258)
- iterator       begin() (line 262)
- iterator       end() (line 266)
- const_iterator begin() const (line 270) (I guess this is an overload for iterator       begin() )
- const_iterator end() const (line 274) (I guess this is an overload for iterator       end() )
- void reserve(size_type newCapacity) (line 278)
- void resize(size_type newSize) ( line 292)
- void assign(const_iterator first, const_iterator last) (line 299)
- void allocCapacity(size_type capacity) (line 316)
- void freeStorage(T *storage, const size_type elements) (line 327)
- static size_type roundUpCapacity(size_type capacity) (line 307)
- size_type _capacity (line 64)
- size_type _size (line 65)
- T *_storage (line 66)

bitstream.h
-------------
- bool eos() const (line 277)
- bool eos() const (line 320) - probably same as the previous one
- bool err() const (line 324)
- int32 pos() const (line 328)
- int32 size() const (line 332)
- bool seek(uint32 offset) (line 336)
- byte readByte() (line 345)
- uint16 readUint16LE() (line 355)
- uint16 readUint16BE() (line 374)
- uint32 readUint32LE() (line 393)
- uint32 readUint32BE() (line 411) - I guess the difference between the last four is that they are either big endian or little endian but still I need a description of what this really is

sev-:
This now needs to be rebased.

b-gent:
> This now needs to be rebased.

rebased

sev-:
Thanks! Mergning

PR #2473, opened 2020-09-21
BUILD: Add option not to bundle all resources in the executable

lotharsm:
Currently, the Win32 builds (and possibly other ports as well) bundle all resources (engine-data, fonts, wwwroot...) straight into the main executable. This might have been a good idea back in the days, but now that we easily exceed 20MB for single resource files, this could be pretty limiting in the future.

This patch adds an option to the configure script to create a "slim" build by excluding the resources from being built into the executable. This is currently only used for the Win32 build via scummvm.rc, but it might be helpful for other builds as well.

By default, the resources are built in, so devs have everything in one executable, while the release builds for the end users could be created as what I call slim builds.

### DIFF
--- a/configure
+++ b/configure
@@ -210,6 +210,7 @@ _nasm=auto
 _optimization_level=
 _default_optimization_level=-O2
 _nuked_opl=yes
+_builtin_resources=yes
 # Default commands
 _ranlib=ranlib
 _strip=strip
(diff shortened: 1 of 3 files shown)


henke37:
Good start, but there are plenty more data that ends up stuffed in the executable. Data tables, bitmaps, even the entire fallback theme.

lotharsm:
Yep, indeed, this is more or less a first attempt to get the executable as slim as possible. With engines built as plugins and almost everything excluded, I can get a executable size of only 6MB... And it compresses much better for the installers (that's why 2.3.0 will most likely been built as slim executable...)

sev-:
Data tables? Well, it is very well outside of scope of this patch.

@henke37 if you would like to improve some engine by moving its static data to a *.dat file, you're more than welcome!

PR #2475, opened 2020-09-23
AUDIO: Increase number of audio mixer channels to 32

aquadran:

Digital iMuse in Grim engine (GrimE) require more channels in audio mixer available.


sev-:
Thanks

PR #2476, opened 2020-09-23
COMMON: Added PS2 and XBOX platform definitions.

aquadran:

There are PS2 and XBOX game versions of Myst III: Exile.
Change add them to platform definitions which are used by code game detection in Myst engine.


sev-:
Thanks!

PR #2477, opened 2020-09-23
GRAPHICS: Added support for stem darkening in TTF

aquadran:

The change introduce support for stem darkening in TTF.
It's used in The Longest Journey game engine.


sev-:
Thanks!

PR #2478, opened 2020-09-23
VIDEO: Added getter for getFrameRate() for BINK and SMK video streams.

aquadran:

Added public method, getter for SMK and BINK video frame rate.
It's used for override frame rate by video player in The Longest Journey engine.


sev-:
Thanks, merging

PR #2479, opened 2020-09-23
COMMON: Mempool checked build

henke37:
This PR adds validation to the memory pool system to force correct usage.

### DIFF
--- a/common/memorypool.h
+++ b/common/memorypool.h
@@ -73,7 +73,7 @@ public:
 	 * @param chunkSize		the chunk size of this memory pool
 	 */
 	explicit MemoryPool(size_t chunkSize);
-	~MemoryPool();
+	virtual ~MemoryPool();
 
 	/**
 	 * Allocate a new chunk from the memory pool.
--- a/common/memorypool.h
+++ b/common/memorypool.h
@@ -26,6 +26,10 @@
 #include "common/scummsys.h"
 #include "common/array.h"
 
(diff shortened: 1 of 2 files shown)


common/memorypool.h
```diff
@@ -26,6 +26,14 @@
 #include "common/scummsys.h"
 #include "common/array.h"
 
+#ifndef NDEBUG
```
criezy, line 29: We do not want to enable those checks in release builds, and currently we do not define `NDEBUG` in our release builds. I don't see any good reason for it, so I created PR #2486 to add it. But if it turns out there was a good reason not to define `NDEBUG`, this line will need to be changed to check `RELEASE_BUILD` instead.

criezy (approved):
This looks good to me. I don't know how useful it will prove to be, but other than a slight performance decrease in debug builds it should no do any harm.

I would just wait on PR #2486 before merging this in case there was a good reason why we might not want to define `NDEBUG` in release builds.

common/memorypool.h
criezy: It looks like adding NDEBUG for release builds might not happen anytime soon. Can you please replace `#ifndef NDEBUG` with `#ifndef RELEASE_BUILD` and then I think we can merge this?

sev-:
Closing this with no reaction. Moreover, error() on this kind of problem could lead to major regressions.

PR #2480, opened 2020-09-23
VIDEO: Added support for seeking BINK streams

aquadran:

This change introduce support for seeking BINK streams.
It's used by Grim engine for "Escape from Monkey Island".
Video player needs support for looping and restore movie state from save game.


sev-:
Thanks!

PR #2481, opened 2020-09-24
VIDEO: Added support for alpha component support for BINK videos

aquadran:

This adds support for BINK videos with alpha component used in Myst3 engine.
It's also integrate support alpha channel in yuv to rgb conversion code.


sev-:
Thanks!

PR #2482, opened 2020-09-24
VIDEO: Added support for PS2 audio codec in MPEG stream

aquadran:

This change introduce support PS2 audio codec in MPEG stream.
It's used by Grim engine for "Escape from Monkey Island" in video player.


sev-:
Thanks, great!

PR #2484, opened 2020-09-25
COMPOSER: fix regression macintosh darby & gregory

BLooperZ:
Fix regression in macintosh version of "Gregory and the Hot Air Balloon" and "Darby the Dragon" caused by commit #1948


digitall:
Since this fix is to work done on PR 1948 which was @BLooperZ himself anyway and the change is limited to COMPOSER engine, I think this is safe to commit for testing...

PR #2485, opened 2020-09-26
ANDROID: Add more icon variants

mduggan:
This makes the icon render much better on Android 11 and Chromebook devices.  I
think it should also be backwards compatible - although the oldest device I had
to test was Android 7, it works fine there.


mduggan:
@antoniou79 Did this as a PR so you could take a quick look if you want.

mduggan:
Updated to use the orange background for the vector icon. I think it's good to merge now, and we should consider cherry-picking for the 2.2.1 release as it makes the app look much more appealing :)

antoniou79:
Excellent, thank you for your work I'll test on older devices too; at least an old Huawei Y500 (Android 4.4.2). I have a few others but they are more... stubborn these days due to several hardware faults that come with age and usage. I've sadly had to give away my Moto E (Android 5.1 Lollipop) device. 

And I agree about cherrypicking. I'll do that after the quick testing.

Mataniko:
Sorry I'm a little late on this one - but please make sure you add these new variants to the make file in scummvm-media so we can easily regenerate

mduggan:
Sure, I'll update that repo too.  For the vector version I don't think there's an easy way to update unfortunately, but the pngs are no problem.

PR #2486, opened 2020-09-26
BUILD: Add NDEBUG define for release builds

criezy:
Currently we do not define `NDEBUG` in our release builds. However this define is checked in a few places (`common/hashmap.h`, `audio/mods/maxtrax.cpp`, and `engines/sci/graphics/celobj32.cpp`). This define also affects the behaviour of `assert()`, which we use in many more places.

It seems to me that we likely want to define `NDEBUG` in our release builds, but I am creating a pull request for this in case I overlooked something and there is a good reason for not defining it.

Edit: I added another commit to disable debug builds by default when doing a release build. On Discord several porter mentioned using both `--enable-release` and `--disable-debug` for their release build, and it seems to me that disabling debug symbols should be the default when we use `--enable-release`.

This also raise the question: should `NDEBUG` be defined when `_release_build` =  `yes` or when `_debug_build` = `no`? In other word, if we use `configure --enable-release --enable-debug`, do we want `NDEBUG` to be defined or not?

### DIFF
--- a/common/textconsole.h
+++ b/common/textconsole.h
@@ -77,5 +77,16 @@ void warning(const char *s, ...) GCC_PRINTF(1, 2);
 
 #endif
 
+/**
+ * Test the given expression and call error() with the file, line, and expression
+ * if it is false.
+ *
+ * Unlike assert(), scumm_assert() is not disabled when NDEBUG is defined. Also
+ * the message is visible to the user (or at least logged) on all platforms.
+ */
+#define scumm_assert(expr) \
+	if (expr) {} \
+	else \
+		error("Assertion \"%s\" failed: file \"%s\", line %d", #expr, __FILE__, __LINE__)
 
 #endif
--- a/configure
+++ b/configure
@@ -2296,6 +2296,9 @@ echo $_use_cxx11
 #
 # Determine extra build flags for debug and/or release builds
 #
+if test "$_debug_build" = auto && test "$_release_build" = yes; then
+	_debug_build=no
+fi
 if test "$_debug_build" != no; then
 	# debug mode not explicitly disabled -> compile with debug information
 	echo_n "Checking best debug mode... "
--- a/configure
+++ b/configure
@@ -2329,6 +2329,7 @@ if test "$_release_build" = yes; then
 	# Add a define, which indicates we are doing
 	# an build for a release version.
 	append_var DEFINES "-DRELEASE_BUILD"
+	append_var DEFINES "-DNDEBUG"
 fi
 
 # By default, we add -pedantic to the CXXFLAGS to catch some potentially
--- a/devtools/create_project/visualstudio.cpp
+++ b/devtools/create_project/visualstudio.cpp
@@ -272,7 +272,7 @@ void VisualStudioProvider::createBuildProp(const BuildSetup &setup, bool isRelea
 	if (isRelease) {
 		properties << "\t\tEnableIntrinsicFunctions=\"true\"\n"
 		           << "\t\tWholeProgramOptimization=\"true\"\n"
-		           << "\t\tPreprocessorDefinitions=\"WIN32;RELEASE_BUILD\"\n"
+		           << "\t\tPreprocessorDefinitions=\"WIN32;RELEASE_BUILD;NDEBUG\"\n"
 		           << "\t\tStringPooling=\"true\"\n"
 		           << "\t\tBufferSecurityCheck=\"false\"\n"
 		           << "\t\tDebugInformationFormat=\"0\"\n"
--- a/devtools/create_project/xcode.cpp
+++ b/devtools/create_project/xcode.cpp
@@ -985,6 +985,7 @@ void XcodeProvider::setupBuildConfiguration(const BuildSetup &setup) {
 	REMOVE_SETTING(scummvm_Release, "GCC_PREPROCESSOR_DEFINITIONS");
 	ValueList scummvm_Release_defines(scummvm_defines);
 	ADD_DEFINE(scummvm_Release_defines, "RELEASE_BUILD");
+	ADD_DEFINE(scummvm_Release_defines, "NDEBUG");
 	ADD_SETTING_LIST(scummvm_Release, "GCC_PREPROCESSOR_DEFINITIONS", scummvm_Release_defines, kSettingsNoQuote | kSettingsAsList, 5);
 
 	scummvm_Release_Object->addProperty("name", "Release", "", kSettingsNoValue);


antoniou79:
Maybe we shouldn't allow both --enable-release and --enable-debug. One should supersede the other. Otherwise it is counter-intuitive and confusing.

criezy:
There might be a reason why we might want to make a release build with debug symbols (e.g. for profiling since profiling a non-optimised build is usually useless).

What I find counter-intuitive is that when we use `--enable-release`, doing a debug build is the default currently, which is why I am suggesting of changing the default here. But I don't think we need to forbid doing an explicit release build with debug symbols.

Although we could do profiling with `--enable-optimizations` instead of `--enable-release`. So I am not completely opposed to forbidding `--enable-debug` when `--enable-release` is used.

criezy:
I checked that cmake defines `NDEBUG` for the `RELWITHDEBINFO` build type (which is similar to our `--enable-release --enable-debug` config). And what I  currently have in this PR is consistent with that.

sluicebox:
If this removes asserts from release builds then that would be a huge mistake. It would be like removing error(). The bug tracker has 474 search results for "assert": https://bugs.scummvm.org/search?q=assert&ticket=on

I know that NDEBUG removes asserts in Visual Studio / msbuild but I don't know the story in MinGW and others.

We fail asserts all over the place in release builds and those immediate failures + messages have fueled many bug reports that I've fixed. Most had been in place for years. Some earned their own bullet points in NEWS.md, including in this latest release. Without asserts in release builds they'd still be active today. I have a SCI bug fix in progress right now due to a screenshot someone posted to a GOG forum, using the release build that shipped with a GOG version, and it includes an assert message box that revealed the bug and its cause. Without that they would have buffer overflowed and maybe (maybe!) crashed later. Either way I'd have no idea. I can't imagine the rest of you aren't constantly fielding assertion bug reports.

I'm kind of terrified of that SCI32 drawing code. This introduces an untested configuration into critical code that only the former team member who wrote it understood, and only in release builds. That sounds like the start of a war story with a predictable ending. I'll take a look in the SCI code soon and neutralize it one way or another.

lotharsm:
By default and without any further arguments, `--enable-release` should implicitly set `--disable-debug`. However, `--enable-release --enable-debug` should be allowed to replicate CMake's `RelWithDebugInfo` behavior.

I'm not familiar with the code/different defines enough to further comment on the proper way to handle `NDEBUG`.

antoniou79:
Is there a performance impact from keeping asserts in release builds? I suppose if there is, it might be minimal at least on modern high hw spec platforms. 

But sluicebox is right. We may be needing the asserts because quite a few of the bug reports are created based on them, and (from my experience) people tend to test the release builds rather than the development ones. It is unfortunate, but it is how it is. 

I am in favor of removing them in release builds (for performance gains) but we don't want to lose the bug reports!

Maybe it would make sense to have define checks so that the asserts (in debug builds) are swapped with error() code (in release builds) and if those are triggered for a user, then we follow up on the bug report by telling them to try the debug build (or we, ourselves, test with the debug build for the trace info). But that would mean we have to change this everywhere in our code base. 

Also, maybe --enable-debug should be split into --enable-debug (meaning step-by-step debug, no optimizations, special macros enabled or disabled (ie NDEBUG)  etc) and --enable-debug-symbols which is specific for debug symbols and its purpose is self-evident.

criezy:
> Is there a performance impact from keeping asserts in release builds?

The performance impact of keeping assert in release builds would be minimal.

> If this removes asserts from release builds then that would be a huge mistake.

Yes it does remove asserts from release builds.

> We fail asserts all over the place in release builds and those immediate failures + messages have fueled many bug reports that I've fixed.

The main issue I see is that we are abusing `assert()` in the ScummVM code. This macro is designed to capture programming errors, not user or run-time errors, since it is generally disabled after a program exits its debugging phase. Also on non-desktop ports this just causes the application to crash with no message to the user, which looks bad (at least that is what happens on iOS when you are not running the application through the debugger).

This also causes inconsistencies and confusion.  Some build systems like `CMake` automatically define `NDEBUG` for release builds. And some of the ScummVM ports also define it (since it is defined in `backends/platform/openpandora/build/build.sh`, `backends/platform/gph/build/gp2x-config.s`, and `backends/platform/ds/arm9/makefile`). Fortunately for us if we want to keep those in release builds, neither Xcode nor Visual Studio define `NDEBUG` automatically for release builds, and as far as I know no porter use the cmake file generated by`create_project` to build releases.

But for all those reason it seems to me that if we have sanity check that we want to keep in release builds, those should be using `error()` and not `assert()`.

However given the number of `assert()` we have in our source code, a more pragmatic approach might indeed be to reject this PR and not define `NDEBUG` in our release build (or at least not for the desktop ports).

> maybe --enable-debug should be split into --enable-debug (meaning step-by-step debug, no optimizations, special macros enabled or disabled (ie NDEBUG) etc) and --enable-debug-symbols which is specific for debug symbols and its purpose is self-evident.

I don't think there is anything to split. The only thing that `--enable-debug` does is to enable debug symbols. All the other things you mention seem to imply a definition of `debug` as `not release` and this is controled by `--enable-release` / `--disable-release`. We don't need another option to control the same thing.

sluicebox:
I completely agree that assert() is being inappropriately relied on, that's been my reaction several times. If that were changed then defining NDEBUG is a good idea.

When to use error() vs assert() sounds like a good candidate for the coding style/guidelines doc. It's subtle and there's a lot of bad assert examples out there in the world that get learned from.

The scummvm asserts I encounter most often are the basic bounds checks in the stream and collection classes. The binary-format parsing code in SCI and common that I've dealt with doesn't do much (or any!) error checking and those asserts are often the only things keeping it from going off the rails when given unexpected input. Of course, the only reason I'm dealing with those ones in the first place is that something went wrong. Unexpected input is the classic example of something that shouldn't be handled by asserts. (But I'll still take it over an out of bounds read!)

I took a look at the SCI32 drawing code and it's no big deal, phew.

criezy:
I have added a commit that defines a `scummvm_assert` macro that is similar to `assert` except that it is not disabled by the `NDEBUG` define and that it calls `error()` to handle the failed assert.

The idea is that it should be used instead of `assert` in cases such as the one highlighted by @sluicebox (unexpected inputs). If this addition is accepted then the [coding guidelines](https://wiki.scummvm.org/index.php?title=Coding_Conventions) can be updated to indicate when it should be used in place of asserts.

This also raise the question of what we should do with all the existing asserts. Looking at all of them individually would take a lot of times. I could do a bulk search and replace for all of them, or only in some subdirectories (e.g. engines).

sluicebox:
It's a tricky combination:

- Large codebase full of input parsing without explicit validation. (Most parsers use streams and collections)
- Streams and collections with asserts on their operations. (De-facto input validations so we want them in release builds)
- Large codebase with a lot of misc asserts (Some are catching bugs in release builds)

I think we'd want to avoid adding another mechanism for code to assert, that sounds like it would make things more confusing. Turning those kinds of asserts into errors in all builds seems like a good long term goal, but it might take a while. Starting with collections and streams seems easiest. I'm leaning towards leaving things as-is for now until more of this is figured out. I'd like to see more dev input on this, "when do we assert?" and "what do we do with all these asserts?" are fun conversations. This PR has become much more interesting than the title! The status quo isn't consistent but it's still pretty good and I just don't want to lose the benefits before it's all sorted out.

criezy:
I will merge commit 1d69e7d manually since that one is unrelated to the `assert` and `NDEBUG` discussion and doesn't seem controversial.

And I will keep this PR open for now so that the discussion on the use of `assert` can continue here.

mgerhardy:
I would also vote for a `scummvm_assert`. Not sure if anyone had a look at the `SDL_assert` feature. It's really great. It provides a popup windows on assert with several options (like, break-into-debugger, retry, abort, ...).
I've even extended this in my own engine with a stacktrace: https://github.com/mgerhardy/engine/blob/master/src/modules/core/Assert.h
That way the assert provides a lot more details and information.
I would also support defining `NDEBUG` for release builds. It might disable slow paths in system headers (we are not really using the STL much - the benefits would be even greater there).

fedor4ever:
Symbian port use implicit NDEBUG for releases and _DEBUG for debug.

lotharsm:
I think we should keep the assertions even in release builds with debug (mostly) disabled.

The release builds are by far the most wide-spread versions of ScummVM out in the wild, so we have the highest chance of getting some feedback and bug reports. There are so many different system configurations out there I consider it nearly impossible to really test _every possible combination_ during release testing, so I think we should be able to do (basic) debugging on release builds as well.

Keeping the asserts in might be very helpful during debugging sessions. And regarding debugging: Is it possible to get a "weaker" version of  `NDEBUG` that mimics CMake's `RelWithDebugInfo`? Having _all_ the debug information removed might be too much even on release builds.

somaen:
IMHO assert() shouldn't be part of the release, but something like scummvm_assert() should. But that also assumes that we generally have moved all runtime-validation to use scummvm_assert(), and only use assert() where actually guarding against programmer errors.

mikrosk:
>> Is there a performance impact from keeping asserts in release builds?

> The performance impact of keeping assert in release builds would be minimal.

Just my two cents: it depends on the platform. When running on hardware with tens of megahertz (instead of thousands) you can definitely see them contributing to CPU load significantly (not to speak about file size...).

So I'm happy for the option to disable them completely. For modern machines I'd be inclined to keep them, though.

PR #2488, opened 2020-09-27
[DNM] DOXYGEN: doxygen changes in header files 2

b-gent:
Edited files in the common folder:

- bufferedstream.h
- callback.h
- config-manager.h
- coroutines.h
- cosinetables.h
- dcl.h
- debug.h
- debug-channels.h

Output is available on https://b-gent.github.io/PR2488/group__common__buffstream.html

### DIFF
--- a/common/callback.h
+++ b/common/callback.h
@@ -29,37 +29,36 @@ namespace Common {
  * @defgroup common_callback Callbacks
  * @ingroup common
  *
- * @brief  Callback templates.
- *
+ * @brief  Callback class templates.
  * @{
  */
 
 /**
  * BaseCallback<S> is a simple base class for object-oriented callbacks.
  *
- * Object-oriented callbacks are such callbacks that know exact instance
- * which method must be called.
+ * Object-oriented callbacks are callbacks that know the exact instance
+ * of the method that must be called.
  *
- * For backwards compatibility purposes, there is a GlobalFunctionCallback,
+ * For backward compatibility purposes, GlobalFunctionCallback is available,
  * which is BaseCallback<void *>, so it can be used with global C-like
  * functions too.
  *
- * <S> is the type, which is passed to operator() of this callback.
- * This allows you to specify that you accept a callback, which wants
- * to receive an <S> object.
+ * \<S\> is the type that is passed to operator() of this callback.
+ * This allows you to specify that you accept a callback that wants
+ * to receive an \<S\> object.
  */
 template<typename S = void *> class BaseCallback {
 public:
 	BaseCallback() {}
 	virtual ~BaseCallback() {}
-	virtual void operator()(S data) = 0;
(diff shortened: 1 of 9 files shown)


b-gent:
- bufferedstream.h looking very good

Descriptions required for the following:

callback.h
-----------
I don't fully get these callbacks, I added some descriptions (please check() but the following are still missing:
- GlobalFunctionCallback(GlobalFunction cb) (line 70)
- Callback(T *object, TMethod method) (line 99)
- typedef void(T::*TMethod) (line 95)
- T *_object (line 96) 
- TMethod _method (line 97)
- CallbackBridge(T *object, TCallbackMethod method, BaseCallback<OS> *outerCallback) (line 141)

config-manager.h
--------------------
- typedef HashMap<String, Domain, IgnoreCase_Hash, IgnoreCase_EqualTo> DomainMap (line 94)
- 	bool				hasKey(const String &key) const; -> not sure how to define this 'key' and thus don't know how to document the structures that use it as an argument
	const String &		get(const String &key) const;
	void				set(const String &key, const String &value);  (lines 129-131)
- 	bool				hasKey(const String &key, const String &domName) const;
	const String &		get(const String &key, const String &domName) const;
	void				set(const String &key, const String &value, const String &domName);
	void				removeKey(const String &key, const String &domName);    (lines 151-155)
- friend class Singleton<SingletonBaseType> (line 201)
- 	void				registerDefault(const String &key, const String &value);
	void				registerDefault(const String &key, const char *value);
	void				registerDefault(const String &key, int value);
	void				registerDefault(const String &key, bool value);   (lines 170-173)
- 	bool				hasGameDomain(const String &domName) const;
	bool				hasMiscDomain(const String &domName) const; (lines 190-191)
- 	DomainMap::iterator beginGameDomains() { return _gameDomains.begin(); 
	DomainMap::iterator endGameDomains() { return _gameDomains.end();  (lines 194-195)
- void copyFrom(ConfigManager &source); (line 198)
- ALL members of the Domain class are undocumented (lines 66-91)

sev-:
This needs rebase now.

b-gent:
> This needs rebase now.

Rebased.

sev-:
Good progress, merging.

common/callback.h
```diff
@@ -69,43 +68,45 @@ template<typename T> class GlobalFunctionCallback: public BaseCallback<T> {
...
  *         &MyClass::myMethod
  *     )
+ * @endcode
  */
 template<class T, typename S = void *> class Callback: public BaseCallback<S> {
 protected:
 	typedef void(T::*TMethod)(S);
 	T *_object;
-	TMethod _method;
+		;
```
sev-, line 96: That broke the compilation horribly, for obvious reasons.

b-gent: To be honest, I was assuming we would still be working on this PR, at least adding missing doc in the high-priority headers. Hence the 'Do Not Merge' label in its title.

PR #2489, opened 2020-09-30
GUI: Small Improvements

aryanrawlani28:
- ~Remove redundant constructor/function since shifted to C++11 from the U32 code~
- Hindi language proper name
- improper GUI fixed when changing to Hebrew in the classic theme. (see pictures below.)

**Before**

before

**After**

after

### DIFF
--- a/gui/options.cpp
+++ b/gui/options.cpp
@@ -2297,7 +2297,6 @@ void GlobalOptionsDialog::apply() {
 		ConfMan.set("gui_language", newLang);
 		newCharset = TransMan.getCurrentCharset();
 		isRebuildNeeded = true;
-		g_gui.setLanguageRTL();
 	}
 
 	bool guiUseGameLanguage = _guiLanguageUseGameLanguageCheckbox->getState();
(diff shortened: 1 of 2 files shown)


criezy:
Is there a clear benefit in using delegating constructors in this case?

While c++11 is now enabled by default, there is still an option to disable it in configure, and several of our ports do not yet use a compiler recent enough to use c++11. Until the new build server with updated toolchains is deployed, I would prefer if we avoid breaking compilation on the current one, unless there is a very good reason for it.

Using c++11 in engines is fine I think (we can disable those engines on buildbots for ports with an old compiler), but I think we should avoid using c++11 in common codes for now (unless it is optional and can be toggled with a `#ifdef USE_CPP11` such as move constructors).

aryanrawlani28:
> While c++11 is now enabled by default, there is still an option to disable it in configure, and several of our ports do not yet use a compiler recent enough to use c++11. Until the new build server with updated toolchains is deployed, I would prefer if we avoid breaking compilation on the current one, unless there is a very good reason for it.

Ah, sorry! I guess I didn't think about it from that point of view. You are correct then, it doesn't make a good case to do this - it doesn't really do anything. I will drop that commit.

Thanks!

sev-:
Thanks!

PR #2490, opened 2020-10-03
Partial c++11 support

criezy:
What I am proposing in with this pull request is to:
- Allow any engine to use c++11
- Allow any backend to use c++11
- Restrict c++11 in common code to features that can be disabled

The idea is that I would like to avoid breaking compilation for ports that do not yet support c++11, while allowing more c++11 use in the project.

The way I went about this is:
1. Add a cxx11 feature that can be used as a dependency by engines that use c++11. This will automatically disable those engines when c++11 is not enabled. This is added for the petka engine, and should fix the compilation on buildbot for the canoo, dingux, gp2xwiz, openpandora, and osx_intel builders.
2. Add a USE_CXX11 define when C++11 is enabled. This can be used in common code to add code such as constructors with the move semantics or intializer list. I added those to the Array class as an example.
3. Add check in configure, and replacement in common for `std::initializer_list` when not supported by the standard library (as is the case when compiling for Mac OS X 10.6 on a more recent system (where the compiler supports c++11, but the libstdc++ used when targeting those system does not - Apple switched to libc++ in Mac OS X 10.9, and only that one fully support c++11).

Feedback on both the proposal and the actual commits in this PR are welcome.
Would you for example prefer is we stopped supporting non-c++11 builds and just always enabled c++11 everywhere in the project? And do you see an issue with the commits in this PR?

### DIFF
--- a/common/array.h
+++ b/common/array.h
@@ -28,6 +28,10 @@
 #include "common/textconsole.h" // For error()
 #include "common/memory.h"
 
+#ifdef USE_CXX11
+#include <initializer_list>
+#endif
+
 namespace Common {
 
 /**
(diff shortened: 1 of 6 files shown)


henke37:
So much for my plan to add move semantics to the custom containers.

criezy:
> So much for my plan to add move semantics to the custom containers.

You are welcome to add more. I only did it for the `Common::Array` ;-)

henke37:
Been there, done that. It's a branch on my fork.

sev-:
Yes, this is a great improvement, thank you!

Of course, now it cannot apply cleanly.

I would like to make c++11 as a must-have feature, and allow it in the common code as well. Currently, this will affect only intel port, the other Dingux-based ports are dead anyway.

criezy:
I just rebased it and it can be merged again.

criezy:
> I would like to make c++11 as a must-have feature, and allow it in the common code as well. Currently, this will affect only intel port, the other Dingux-based ports are dead anyway.

Yes, that should be the end goal. But I though this intermediate step where we restrict how much c++11 we can use in common code gives us more time to deploy the new buildbot and gives porter more time to update their toolchains.

sev-:
Thanks!

PR #2492, opened 2020-10-04
CREATE_PROJECT: Added MacPorts paths to Xcode

aquadran:
Added MacPorts paths to Xcode project.
Current paths only support /usr/local.


sev-:
Thanks!

PR #2494, opened 2020-10-04
SDL: Prefer OpenGL over Metal rendering on macOS

dwatteau:
SDL 2.0.10 and above favors Metal over OpenGL rendering on macOS, but
this currently causes blurriness issues on Retina displays.

Going back to OpenGL rendering works around this bug for now.

Users can override this with the SDL_RENDER_DRIVER environment variable.

c59bf95ba85b25b1 and 70f79d3df8b4a9d4 are still a requirement to
work around this bug, too.

Bug #11430.

(This should probably backported to any 2.2.1 release)


criezy:
Thank you. That looks good. Merging...

PR #2495, opened 2020-10-05
Add division and multiplication operators to Common::Point

phcoder:
They're convenient for performing computations and I use
them in Hades Challenge engine


sev-:
Thanks!

henke37:
A bit inconsistent to not do the assignment operators too. Want to add those operators too?

PR #2496, opened 2020-10-05
common/array: Fix comment

phcoder:
Comment speaks about double-linked list but we actually have an array


sev-:
Good catch, thanks!

PR #2497, opened 2020-10-06
PETKA: Fix memory leaks

Rampopula:
I checked engine in valgrind and have fixed some memory leaks


whiterandrek:
Nice work. Thanks!

PR #2512, opened 2020-10-11
GUI: Update "edit game" for new graphics options (please test before merge!)

eriktorbjorn:
The new graphics settings (I assume they were added in the ResidualVM merge) are always active in "edit game", even if we're not overriding the global settings. This attempts to fix that.

I'm not sure if I got it all right, and the "V-Sync" checkbox is a bit unusual in that this setting is true by default. But it's a start.


aquadran:
tested, looks ok

PR #2513, opened 2020-10-12
CONFIGURE: Also install shaders if gles2 enabled

lmerckx:
Since the commit "CONFIGURE: Do not enable opengl_shaders feature flag for gles2. " (b3ffd89c6824b82b958784305017d5c9e4f207d9), the shaders are no more installed for opengles2 systems.
But they need them as they are used in the code.

I don't know if my way to test "USE_OPENGL_SHADERS or USE_GLES2" in Makefile is the standard of ScummVM but it works.
I'm open to any suggestions ...

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -442,7 +442,7 @@ endif
 
 # Shaders
 DIST_FILES_SHADERS=
-ifdef USE_OPENGL_SHADERS
+ifneq ($(USE_OPENGL_SHADERS)$(USE_GLES2),)
 ifdef ENABLE_GRIM
 DIST_FILES_SHADERS+=$(wildcard $(srcdir)/engines/grim/shaders/*)
 endif
--- a/Makefile.common
+++ b/Makefile.common
@@ -440,7 +440,7 @@ ifdef ENABLE_AGI
 DIST_FILES_ENGINEDATA+=$(srcdir)/dists/pred.dic
 endif
 
-# Shaders
+# Shaders: install if USE_OPENGL_SHADERS or USE_GLES2 defined
 DIST_FILES_SHADERS=
 ifneq ($(USE_OPENGL_SHADERS)$(USE_GLES2),)
 ifdef ENABLE_GRIM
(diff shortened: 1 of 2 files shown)


aquadran:
you are right, shaders should be installed on 'OR" conditions now. I'm not good enough with makefile syntax to comment :)
But in gerenal 👍

Makefile.common
```diff
@@ -442,7 +442,7 @@ endif
 
 # Shaders
 DIST_FILES_SHADERS=
-ifdef USE_OPENGL_SHADERS
+ifneq ($(USE_OPENGL_SHADERS)$(USE_GLES2),)
```
tsoliman, line 445: Interesting. I didn't know this but apparently it is _the way_ according to [the GNU make docs](https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html)

The docs also show usage of `strip` to remove white space, as otherwise whitespace values are interpreted as non-empty. Not sure if it is needed here.

I'm not great with make and so the intention wasn't obvious to me at all. Perhaps a comment might be helpful?

lmerckx: Thanks for the proposition.
I've just added a comment.

tsoliman:
Looking at this more, wouldn't this make more sense to be handled in `configure`?

Sort of keeping USE_OPENGL_SHADERS as the condition in the makefile but setting `_opengl_shaders` to `yes` if it was `auto` if `_opengles2` is set to `yes`?

Basically revert the changes in the `Makefile.common` and `ports.mk` but make the changes in `configure` instead?

something like this:

```
if test "$_opengl_shaders" = auto && test "$_opengles2" = yes; then
        _opengl_shaders=yes
fi
```

This is a lot more clear and follows the intention better.

(Note: This will obviously break if someone configures with --enable-opengles2 and --disable-opengl-shaders but I think that's ok and a good thing. It's certainly better than silently enabling shaders when explicitly requesting disabling shaders)

lmerckx:
It is almost what was did before b3ffd89c6824b82b958784305017d5c9e4f207d9:
When opengles2 was detected by the `configure`, opengl_shaders was also enabled.

But perhaps it implied non wanted side effects ?

aquadran:
opengles2 and opengl_shaders are similar but needs to be separated, opengl_shaders stands for opengl (non-gles) but capable shaders. generally they cover most same code, but you can exclude each other if needed.

aquadran:
I'm pushing this to unbreak builds. if needed we can continue discussion

PR #2518, opened 2020-10-13
COMMON: Try get functionality for HashMap and ConfMan

henke37:
Makes for simpler/more efficient code when keys are optional.

Saves having to check for the key existing and then fetching the element at that key, which is two searches in the hashmap and two function calls in the code using the map.


sev-:
Thanks!

PR #2525, opened 2020-10-14
CREATE_PROJECT: export compile_commands.json via cmake

mgerhardy:
This is useful for e.g. clangd or other clang based tools like the analyser.

https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

I'm using the generated `compile_commands.json` with clangd and vscode (the file must be located in the project root to be found automatically by clangd)

Install the clangd plugin for vscode, execute `cmake`, copy the `your_build_dir/compile_commands.json` to `scummvmcheckout/compile_commands.json` and execute `Restart clangd` in vscode.

```bash
cd devtools/create_project/cmake
cmake .
cmake --build .
cd ../../..
./devtools/create_project/cmake/create_project . --cmake
mkdir build
cd build
cmake ..
cd ..
ln -sf build/compile_commands.json compile_commands.json
```
Bildschirmfoto von 2020-10-14 16-49-11

Now install clangd and open the project. It might be problematic if you have the project already configured with intellisense. So doing a fresh vscode workspace might be better and answering the intellisense questions with **no**

Another nice option is now to add a [.clang-tidy](https://github.com/mgerhardy/engine/blob/master/src/.clang-tidy) and .clang-format - they are automatically picked up for static code analysis and code formatting in the ide.

Bildschirmfoto von 2020-10-14 16-57-24

### DIFF
--- a/devtools/create_project/cmake.cpp
+++ b/devtools/create_project/cmake.cpp
@@ -75,7 +75,7 @@ const CMakeProvider::Library *CMakeProvider::getLibraryFromFeature(const char *f
 void CMakeProvider::createWorkspace(const BuildSetup &setup) {
 	std::string filename = setup.outputDir + "/CMakeLists.txt";
 	std::ofstream workspace(filename.c_str());
-	if (!workspace)
+	if (!workspace || !workspace.is_open())
 		error("Could not open \"" + filename + "\" for writing");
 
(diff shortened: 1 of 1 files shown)


sev-:
Thanks!

PR #2526, opened 2020-10-14
CREATE_PROJECT: Put detection in a separate project

henke37:
Keeps the main project clean and is a prerequisite for building detection as a plugin.

### DIFF
--- a/devtools/create_project/create_project.cpp
+++ b/devtools/create_project/create_project.cpp
@@ -1512,7 +1512,11 @@ void ProjectProvider::createProject(BuildSetup &setup) {
 	// We also need to add the UUID of the main project file.
 	const std::string svmUUID = _allProjUuidMap[setup.projectName] = createUUID(setup.projectName);
 	// Add the uuid of the detection project
-	const std::string detUUID = _allProjUuidMap["scummvm-detection"] = createUUID("scummvm-detection");
+	const std::string detProject = setup.projectName + "-detection";
+	const std::string detUUID = createUUID(detProject);
+	if (setup.useStaticDetection) {
+		_allProjUuidMap[detProject] = _engineUuidMap[detProject] = detUUID;
+	}
 
 	createWorkspace(setup);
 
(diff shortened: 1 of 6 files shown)


devtools/create_project/create_project.cpp
```diff
@@ -1499,22 +1499,26 @@ void ProjectProvider::createProject(BuildSetup &setup) {
...
 		targetFolder = "/engines/";
 	}
 
+	_allProjUuidMap = _engineUuidMap;
+
 	// We also need to add the UUID of the main project file.
-	const std::string svmUUID = _uuidMap[setup.projectName] = createUUID(setup.projectName);
+	const std::string svmUUID = _allProjUuidMap[setup.projectName] = createUUID(setup.projectName);
+	// Add the uuid of the detection project
+	const std::string detUUID = _allProjUuidMap["Detection"] = createUUID("Detection");
```
SupSuper, line 1514: Since VS sorts projects alphabetically, it might be more convenient to call this something like `scummvm-detection` (so it's next to it).

SupSuper:
Does this work with every IDE?

henke37:
As far as I know, I updated the code for all of the project generators. So, assuming I didn't screw up, it should work will all IDEs.

devtools/create_project/create_project.cpp
henke37: Good idea! I will incorporate this change.

Arvi3d:
Deepcode retry

SupSuper (approved):
I see no further issues with this so I'm gonna go ahead and merge it.

henke37:
Not sure why you put back the condition for when to generate the project. Shouldn't it be used both with static and dynamic plugins?

SupSuper:
When I tried it, it broke compilation, so I decided to leave it in there just to be safe.
It's not clear what the flag is supposed to do anyways, since create_project doesn't support plugins atm, so it'll probably need to be revisited later.

PR #2527, opened 2020-10-14
ALL: add .clang-tidy

mgerhardy:
This adds a basic clang-tidy configuration that is e.g. picked by by the clang-analyser or clangd for IDE integration.

Bildschirmfoto von 2020-10-14 16-57-24


sev-:
Let's give it a try...

PR #2530, opened 2020-10-15
GUI: Fix #11767 - STARK: The Longest Journey - random text replaced with ...

GeorgeSapkin:
This is an attempt to fix ellipsifying of text in STARK by changing width and bounding box calculation discrepancy in recently introduced `handleEllipsis()` and `getBoundingBoxImpl()`.

I haven't dug too deep so this probably makes no sense. Only tested in STARK so far.

### DIFF
--- a/graphics/font.cpp
+++ b/graphics/font.cpp
@@ -52,8 +52,7 @@ Common::Rect getBoundingBoxImpl(const Font &font, const StringType &str, int x, 
 		x = x + w - width;
 	x += deltax;
 
-	bool first = true;
-	Common::Rect bbox;
+	Common::Rect bbox(x, y, x + width, y);
 
 	typename StringType::unsigned_type last = 0;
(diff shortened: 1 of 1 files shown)


chrisis123:
Just tried it, it definitely fixes the bug in TLJ, good work!

No idea if it introduces any problems in other games (only quickly started a random SCUMM game and didn't notice anything wrong) but so far this looks good to me.

aquadran:
Fixed by eef527e22706a12be42eaeb1d17112753879de59

GeorgeSapkin:
@aquadran why is your solution preferable?

sev-:
@GeorgeSapkin because your changed behaviour of text wrapping for other 9 engines and our GUI. :)

GeorgeSapkin:
@sev- that I fully understand. But did it actually cause any regressions?

PR #2533, opened 2020-10-16
MIDI: Add MT-32/GM MidiDriver

NMIError:
MIDI: Add MT-32/GM MidiDriver

I've added a MidiDriver subclass specifically for MT-32 and GM devices. This 
pulls together some functionality which I saw was duplicated in a number of 
engines. I've moved some existing MT-32/GM specific code over from the 
MidiDriver base class and the Miles driver.
This driver can be used as a basis for implementing support for these devices 
in new engines, or old engines can be reworked to use this driver if they want 
to use some of the features of this driver.

I've used this driver to complete MT-32 support in Lure of the Temptress. I've 
also updated the SMF parser to be able to use the multiple source feature for 
use with this game.

The driver has the following features (see mt32gm.h for more details):
- MIDI device initialization
- MT-32 <> GM conversion
- User volume settings
- Reverse stereo
- Correct Roland GS bank and drumkit selects
- SysEx queue
- Multiple MIDI sources, with:
  - Channel mapping
  - Music/SFX volume
  - Source volume
  - Volume fading

### DIFF
--- a/audio/mididrv.cpp
+++ b/audio/mididrv.cpp
@@ -57,45 +57,6 @@ const byte MidiDriver::_gmToMt32[128] = {
 	101, 103, 100, 120, 117, 113,  99, 128, 128, 128, 128, 124, 123, 128, 128, 128, // 7x
 };
 
-// These are the power-on default instruments of the Roland MT-32 family.
-const byte MidiDriver::_mt32DefaultInstruments[8] = {
-	0x44, 0x30, 0x5F, 0x4E, 0x29, 0x03, 0x6E, 0x7A
-};
-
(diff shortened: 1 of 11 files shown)


criezy:
Looking at this pull request reminded me that I understand very little about MIDI. I can't really comment on the pull request, but I hope that somebody who understand this aspect of ScummVM will have a look, because this looks nice, and the prospect of MT-32 support for Lure or the Temptress is exciting.

NMIError:
I hope so too...
Thanks anyway for having a look!

criezy:
I had another quick look to make sure that there is not obvious issue, and also tested a few games.
The code quality looks good, and since it doesn't look like anybody else will be reviewing these changes, I will merge this now.

PR #2535, opened 2020-10-16
DOXYGEN: Further doc updates in high priority files

b-gent:
Adding, editing, updating doxygen comments in header files.

HTML output available at: https://b-gent.github.io/PR2535/d6/d8f/class_common_1_1_config_manager.html

### DIFF
--- a/common/array.h
+++ b/common/array.h
@@ -47,7 +47,7 @@ namespace Common {
  * can be accessed similarly to a regular C++ array. Accessing
  * elements is performed in constant time (like with plain arrays).
  * In addition, you can append, insert, and remove entries (this
- * is the 'dynamic' part). Doing that in general takes time
+ * is the 'dynamic' part). In general, doing that takes time
  * proportional to the number of elements in the array.
  *
  * The container class closest to this in the C++ standard library is
@@ -56,23 +56,23 @@ namespace Common {
 template<class T>
 class Array {
 public:
-	typedef T *iterator;
-	typedef const T *const_iterator;
+	typedef T *iterator; /*!< Array iterator. */
+	typedef const T *const_iterator; /*!< Const-qualified array iterator. */
 
-	typedef T value_type;
+	typedef T value_type; /*!< Value type of the array. */
 
-	typedef uint size_type;
+	typedef uint size_type; /*!< Size type of the array. */
 
 protected:
-	size_type _capacity;
-	size_type _size;
-	T *_storage;
+	size_type _capacity; /*!< Maximum number of elements the array can hold. */
+	size_type _size; /*!< How many elements the array holds. */
+	T *_storage;  /*!< Memory used for element storage. */
 
 public:
 	Array() : _capacity(0), _size(0), _storage(nullptr) {}
 
 	/**
-	 * Construct an array with `count` default-inserted instances of @p T. No
+	 * Construct an array with @p count default-inserted instances of @p T. No
 	 * copies are made.
 	 */
 	explicit Array(size_type count) : _size(count) {
@@ -82,13 +82,16 @@ public:
 	}
 
 	/**
-	 * Construct an array with `count` copies of elements with value `value`.
+	 * Construct an array with @p count copies of elements with value @p value.
 	 */
 	Array(size_type count, const T &value) : _size(count) {
 		allocCapacity(count);
 		uninitialized_fill_n(_storage, count, value);
 	}
 
+	/**
+	 * Construct an array as a copy of the given @p array.
+	 */
 	Array(const Array<T> &array) : _capacity(array._size), _size(array._size), _storage(nullptr) {
 		if (array._storage) {
 			allocCapacity(_size);
@@ -98,7 +101,7 @@ public:
 
 #ifdef USE_CXX11
 	/**
-	 * Constructs an array as a copy of the given array using the c++11 move semantic.
+	 * Construct an array as a copy of the given array using the C++11 move semantic.
 	 */
 	Array(Array<T> &&old) : _capacity(old._capacity), _size(old._size), _storage(old._storage) {
 		old._storage = nullptr;
@@ -107,7 +110,14 @@ public:
 	}
 
 	/**
-	 * Constructs an array using list initialization.
+	 * Construct an array using list initialization.
+	 * For example:
+	 * @code
+	 * Common::Array<int> myArray = {1, 7, 42};
+	 * @endcode
+	 * constructs an array with 3 elements whose values are 1, 7, and 42 respectively.
+	 * @note
+	 * This constructor is only available when C++11 support is enabled.
 	 */
 	Array(std::initializer_list<T> list) : _size(list.size()) {
 		allocCapacity(list.size());
@@ -191,12 +201,13 @@ public:
 		return _storage[_size-1];
 	}
 
-
+    /** Insert an element into the array at the given position. */
 	void insert_at(size_type idx, const T &element) {
 		assert(idx <= _size);
 		insert_aux(_storage + idx, &element, &element + 1);
 	}
-
+    
+	/** Insert copies of all the elements from the given array into this array at the given position. */
 	void insert_at(size_type idx, const Array<T> &array) {
 		assert(idx <= _size);
 		insert_aux(_storage + idx, array.begin(), array.end());
@@ -208,7 +219,8 @@ public:
 	void insert(iterator pos, const T &element) {
 		insert_aux(pos, &element, &element + 1);
 	}
-
+    
+	/** Remove an element at the given position from the array and return the value of that element. */
 	T remove_at(size_type idx) {
 		assert(idx < _size);
 		T tmp = _storage[idx];
@@ -221,16 +233,19 @@ public:
 
 	// TODO: insert, remove, ...
 
+	/** Return a reference to the element at the given position in the array. */
 	T &operator[](size_type idx) {
 		assert(idx < _size);
 		return _storage[idx];
 	}
 
+    /** Return a const reference to the element at the given position in the array. */
 	const T &operator[](size_type idx) const {
 		assert(idx < _size);
 		return _storage[idx];
 	}
 
+    /** Assign the given @p array to this array. */
 	Array<T> &operator=(const Array<T> &array) {
 		if (this == &array)
 			return *this;
@@ -244,6 +259,7 @@ public:
 	}
 
 #ifdef USE_CXX11
+    /** Assign the given array to this array using the C++11 move semantic. */
 	Array &operator=(Array<T> &&old) {
 		if (this == &old)
 			return *this;
@@ -261,10 +277,12 @@ public:
 	}
 #endif
 
+    /** Return the size of the array. */
 	size_type size() const {
 		return _size;
 	}
 
+    /** Clear the array of all its elements. */
 	void clear() {
 		freeStorage(_storage, _size);
 		_storage = nullptr;
@@ -272,6 +290,7 @@ public:
 		_capacity = 0;
 	}
 
+    /** Erase the element at @p pos position and return an iterator pointing to the next element in the array. */
 	iterator erase(iterator pos) {
 		copy(pos + 1, _storage + _size, pos);
 		_size--;
@@ -279,11 +298,13 @@ public:
 		_storage[_size].~T();
 		return pos;
 	}
-
+		
+    /** Check whether the array is empty. */
 	bool empty() const {
 		return (_size == 0);
 	}
 
+    /** Check whether two arrays are identical. */
 	bool operator==(const Array<T> &other) const {
 		if (this == &other)
 			return true;
@@ -296,26 +317,34 @@ public:
 		return true;
 	}
 
+    /** Check if two arrays are different. */
 	bool operator!=(const Array<T> &other) const {
 		return !(*this == other);
 	}
 
+    /** Return an iterator pointing to the first element in the array. */
 	iterator       begin() {
 		return _storage;
 	}
 
+    /** Return an iterator pointing past the last element in the array. */
 	iterator       end() {
 		return _storage + _size;
 	}
 
+    /** Return a const iterator pointing to the first element in the array. */
 	const_iterator begin() const {
 		return _storage;
 	}
 
+    /** Return a const iterator pointing past the last element in the array. */
 	const_iterator end() const {
 		return _storage + _size;
 	}
 
+    /** Reserve enough memory in the array so that it can store at least the given number of elements. 
+	 *  The current content of the array is not modified.
+	 */
 	void reserve(size_type newCapacity) {
 		if (newCapacity <= _capacity)
 			return;
@@ -330,6 +359,7 @@ public:
 		}
 	}
 
+    /** Change the size of the array. */
 	void resize(size_type newSize) {
 		reserve(newSize);
 		for (size_type i = _size; i < newSize; ++i)
@@ -337,6 +367,9 @@ public:
 		_size = newSize;
 	}
 
+    /** Assign to this array the elements between the given iterators from another array,
+	 *  from @p first included to @p last excluded.
+	 */
 	void assign(const_iterator first, const_iterator last) {
 		resize(distance(first, last)); // FIXME: ineffective?
 		T *dst = _storage;
@@ -345,15 +378,17 @@ public:
 	}
 
 protected:
+    /** Round up capacity to the next power of 2.
+	  * A minimal capacity of 8 is used.
+	  */
 	static size_type roundUpCapacity(size_type capacity) {
-		// Round up capacity to the next power of 2;
-		// we use a minimal capacity of 8.
 		size_type capa = 8;
 		while (capa < capacity)
 			capa <<= 1;
 		return capa;
 	}
 
+    /** Allocate a specific capacity for the array. */
 	void allocCapacity(size_type capacity) {
 		_capacity = capacity;
 		if (capacity) {
@@ -365,6 +400,7 @@ protected:
 		}
 	}
 
+    /** Free the storage used by the array. */
 	void freeStorage(T *storage, const size_type elements) {
 		for (size_type i = 0; i < elements; ++i)
 			storage[i].~T();
--- a/common/config-manager.h
+++ b/common/config-manager.h
@@ -35,7 +35,7 @@ namespace Common {
  * @defgroup common_config Configuration manager
  * @ingroup common
  *
- * @brief  The (singleton) configuration manager, used to query & set configuration
+ * @brief  The (singleton) configuration manager, used to query and set configuration
  *         values using string keys.
  *
  * @{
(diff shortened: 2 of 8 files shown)


b-gent:
The last commit added the rest of missing doxygen in str.h. I won't add any more files to this PR and will open a new one instead, it's already a lot to review.
HTML output is updated here: https://b-gent.github.io/PR2535/d8/dbd/group__common__str.html

Please use the suggestion option to correct me in the given lines, I'm sure there will be quite a few. My approach was to add as much as possible, even where I really wasn't sure.

I will rebase tomorrow, the merge conflict in config-manager looks pretty bad.
@sev-

common/array.h
```diff
@@ -221,16 +226,19 @@ class Array {
 
 	// TODO: insert, remove, ...
 
+	/** Return a pointer to the underlying memory serving as element storage. */
 	T &operator[](size_type idx) {
 		assert(idx < _size);
 		return _storage[idx];
 	}
 
+    /** Return a pointer to the underlying memory serving as element storage. */
```
mgerhardy, line 235: These an a few of the following look like whitespace errors

b-gent: thanks, I'll make sure to align with function declaration

b-gent:
Rebased. It wasn't as bad as it seemed - [this commit](https://github.com/scummvm/scummvm/commit/465807f165a2d0a4f3f8fb73686ce52db677bbaa) introduced some serious changes in line spacing so git showed a lot of changed lines but content was not the issue. So the only addition coming from master was [this](https://github.com/scummvm/scummvm/commit/ee4799c51aee8096968f9088f85b915dfac9c109).

common/config-manager.h
```diff
@@ -60,38 +60,44 @@ class ConfigManager : public Singleton<ConfigManager> {
...
 
-		void setVal(const String &key, const String &value) { _entries.setVal(key, value); }
+		void           setVal(const String &key, const String &value) { _entries.setVal(key, value); } /*!< Assign a @p value to a @p key. */
 
-		String &getVal(const String &key) { return _entries.getVal(key); }
-		const String &getVal(const String &key) const { return _entries.getVal(key); }
-		bool tryGetVal(const String &key, String &out) const { return _entries.tryGetVal(key, out); }
+		String        &getVal(const String &key) { return _entries.getVal(key); } /*!< Retrieve the value of a @p key. */
+		const String  &getVal(const String &key) const { return _entries.getVal(key); } /*!< @overload */
+         /**
```
b-gent, line 81: @henke37  you added this recently, can you let me know what is the right description?

common/array.h
b-gent: They all seem well-aligned actually, you can check it for yourself in the raw source view of the file -> https://github.com/scummvm/scummvm/blob/3bd1b124f233d2a66cecf532de1f60e6f1d6140f/common/array.h

Unless you meant something else by 'whitespace error'? The whitespace between the full stop and doxygen comment closing is intentional and used consistently everywhere as part of doxygen style, for better readability.

common/config-manager.h
henke37: It retrieves the value if the key exists and leaves the referenced variable unchanged if the key does not exist.

b-gent: thanks

common/array.h
```diff
@@ -82,13 +82,16 @@ class Array {
...
+	 * Construct an array with @p count copies of elements with value @p value.
 	 */
 	Array(size_type count, const T &value) : _size(count) {
 		allocCapacity(count);
 		uninitialized_fill_n(_storage, count, value);
 	}
 
+	/**
+	 * @todo Doc required
+	 */
```
criezy, line 94: Construct an array as a copy of the given @p array.

I think this should be explicit enough for developers reading the doc. But if we wanted more details we could say:
Construct an array with the same size as the given @p array and whose elements are copies of the elements in @p array.

common/array.h
```diff
@@ -107,7 +110,7 @@ class Array {
 	}
 
 	/**
-	 * Constructs an array using list initialization.
+	 * Construct an array using list initialization.
```
criezy, line 113: Maybe we should add an example to clarify what list initialisation is?
For example
`Common::Array<int> myArray = {1, 7, 42};`
constructs an array with 3 elements whose values are 1, 7, and 42 respectively.

Also since currently C++11 support can be disabled when running configure, we want to indicate that this constructor is only available when C++11 support is enabled.

common/array.h
```diff
@@ -221,16 +226,19 @@ class Array {
...
 		return _storage[idx];
 	}
 
+    /** Return a pointer to the underlying memory serving as element storage. */
 	const T &operator[](size_type idx) const {
 		assert(idx < _size);
 		return _storage[idx];
 	}
 
+    /** @todo Doc required. */
```
criezy, line 241: Assign the given @p array to this array.

common/array.h
```diff
@@ -244,6 +252,7 @@ class Array {
 	}
 
 #ifdef USE_CXX11
+    /** @todo Doc required. */
```
criezy, line 255: Assign the given array to this array using the C++11 move semantic.

common/array.h
```diff
@@ -261,29 +270,34 @@ class Array {
...
 
+    /** Clear the array of all its elements. */
 	void clear() {
 		freeStorage(_storage, _size);
 		_storage = nullptr;
 		_size = 0;
 		_capacity = 0;
 	}
 
+    /** Erase the element at @p pos position. */
```
criezy, line 286: ...and return an iterator pointing to the next element in the array.

common/array.h
```diff
@@ -261,29 +270,34 @@ class Array {
...
 		return pos;
 	}
-
+		
+    /** Check whether the array is empty. */
 	bool empty() const {
 		return (_size == 0);
 	}
 
+    /** Check if two arrays are of the same size. */
```
criezy, line 300: It does not just check that the arrays have the same size, but also that they have the same content.
So in short: `Check if two arrays are identical`.

common/array.h
```diff
@@ -296,26 +310,32 @@ class Array {
 		return true;
 	}
 
+    /** Check if two arrays are different. */
 	bool operator!=(const Array<T> &other) const {
 		return !(*this == other);
 	}
 
+    /** Return a pointer to the underlying memory where the array begins. */
```
criezy, line 318: Again, I would not mention `underlying memory` here.
I would suggest: `Return an iterator pointing to the first element in the array.`

common/array.h
```diff
@@ -296,26 +310,32 @@ class Array {
...
 	iterator       begin() {
 		return _storage;
 	}
 
+    /** Return a pointer to the underlying memory where the array ends. */
```
criezy, line 323: Return an iterator pointing past the last element in the array.

common/array.h
```diff
@@ -296,26 +310,32 @@ class Array {
...
 	iterator       end() {
 		return _storage + _size;
 	}
 
+    /** Return a pointer to the underlying memory where the array begins. */
```
criezy, line 328: Return a const iterator pointing to the first element in the array.

common/array.h
```diff
@@ -296,26 +310,32 @@ class Array {
...
 	const_iterator begin() const {
 		return _storage;
 	}
 
+    /** Return a pointer to the underlying memory where the array ends. */
```
criezy, line 333: Return a const iterator pointing past the last element in the array.

common/array.h
```diff
@@ -296,26 +310,32 @@ class Array {
...
 	const_iterator end() const {
 		return _storage + _size;
 	}
 
+    /** Copy array data to a new place in memory. */
```
criezy, line 338: It might copy the array data to a new place in memory, but this is a side effect, and may not happen.

I would suggest: `Reserve enough memory in the array so that it can store at least the given number of elements. The current content of the array is not modified.`

common/array.h
```diff
@@ -330,13 +350,15 @@ class Array {
...
 
+    /** Change the size of the array. */
 	void resize(size_type newSize) {
 		reserve(newSize);
 		for (size_type i = _size; i < newSize; ++i)
 			new ((void *)&_storage[i]) T();
 		_size = newSize;
 	}
 
+    /** @todo Doc required. */
```
criezy, line 361: Assign to this array the elements between the given iterators (from @p first included to @p last excluded) from another array.

criezy (changes requested):
I made a partial review of this pull request and it needs some changes.
I will continue the review tomorrow.

common/rect.h
```diff
@@ -109,91 +143,104 @@ static inline Point operator*(double multiplier, const Point &p) { return Point(
...
+	int16 bottom, right;	/*!< The point at the bottom right of the rectangle (not part of the Rect). */
 
 	Rect() : top(0), left(0), bottom(0), right(0) {}
+	/**
+     * Create a rectangle with width @p w and height @p h.
+     */
 	Rect(int16 w, int16 h) : top(0), left(0), bottom(h), right(w) {}
+	/**
+     * Create a rectangle between the points defined with @p x1, @p y1, @p x2, and @p y2.
+	 * Additionally, check whether it is a valid rectangle.
```
criezy, line 156: The `Additionally, check whether it is a valid rectangle.` is not really correct. It gives the impression that it is safe to pass invalid coordinates. But such a check in a `assert` actually causes a crash if the check fails. It is used to detect incorrect code.

So it would be more correct to say that it create a rectangle with top-left corner at the given x1 and y1 positions, and the bottom-right corner at the given x2 and y2 positions. The x2 value has to be equal or bigger than the x1 value and the y2 value has to be equal or bigger than the y1 value.

common/rect.h
```diff
@@ -109,91 +143,104 @@ static inline Point operator*(double multiplier, const Point &p) { return Point(
...
 */
 struct Rect {
-	int16 top, left;		///< The point at the top left of the rectangle (part of the rect).
-	int16 bottom, right;	///< The point at the bottom right of the rectangle (not part of the rect).
+	int16 top, left;		/*!< The point at the top left of the rectangle (part of the Rect). */
+	int16 bottom, right;	/*!< The point at the bottom right of the rectangle (not part of the Rect). */
 
 	Rect() : top(0), left(0), bottom(0), right(0) {}
+	/**
+     * Create a rectangle with width @p w and height @p h.
```
criezy, line 151: Maybe mention `a rectangle with the top-left corner at position (0, 0) and the given width @p w and height @p h`?

common/rect.h
```diff
@@ -109,91 +143,104 @@ static inline Point operator*(double multiplier, const Point &p) { return Point(
...
 	Rect(int16 w, int16 h) : top(0), left(0), bottom(h), right(w) {}
+	/**
+     * Create a rectangle between the points defined with @p x1, @p y1, @p x2, and @p y2.
+	 * Additionally, check whether it is a valid rectangle.
+     */
 	Rect(int16 x1, int16 y1, int16 x2, int16 y2) : top(y1), left(x1), bottom(y2), right(x2) {
 		assert(isValidRect());
 	}
+	/**
+     * Create a rectangle with width @p w and height @p h.
```
criezy, line 162: Where does that come from?
This is a comparison operator that checks if two rectangles are identical. It returns true if the two rectangles are the same and false otherwise.

common/rect.h
```diff
@@ -109,91 +143,104 @@ static inline Point operator*(double multiplier, const Point &p) { return Point(
...
+     */
 	bool operator==(const Rect &rhs) const { return equals(rhs); }
+    /**
+     * Create a rectangle with width @p w and height @p h.
```
criezy, line 166: And another comparison operator that checks if two rectangles are different. It returns true if the two rectangles are different, and false if they are the same.

common/rect.h
```diff
@@ -109,91 +143,104 @@ static inline Point operator*(double multiplier, const Point &p) { return Point(
...
+     */
 	bool operator!=(const Rect &rhs) const { return !equals(rhs); }
 
-	int16 width() const { return right - left; }
-	int16 height() const { return bottom - top; }
+	int16 width() const { return right - left; }  /*!< Compute the width of a rectangle. */
```
criezy, line 170: I would not say `compute` here, but just `return`.

It's an implementation detail that the rectangle stores internally the left and right position rather than the left position and the width. The two would be equivalent for the developer using a rectangle.

common/rect.h
```diff
@@ -109,91 +143,104 @@ static inline Point operator*(double multiplier, const Point &p) { return Point(
...
+	int16 height() const { return bottom - top; } /*!< Compute the height of a rectangle. */
```
criezy, line 171: And same comment as above for the width.

common/rect.h
```diff
@@ -226,6 +273,9 @@ struct Rect {
 		right += offset;
 	}
 
+	/**
+	 * Clip this reactangle with another rectangle @p r.
```
criezy, line 277: Typo in `reactangle` ;-)

common/rect.h
```diff
@@ -243,41 +293,62 @@ struct Rect {
...
 
+   	/**
+	 * Reduce the dimensions of this rectangle by setting max width and max heigth.
+	 */
 	void clip(int16 maxw, int16 maxh) {
 		clip(Rect(0, 0, maxw, maxh));
 	}
 
+   	/**
+	 * @todo Doc required.
```
criezy, line 304: Return true if the rectangle is empty (the width or the height is 0) or is invalid (negative width or height). Return false if this is a non-empty valid rectangle.

common/rect.h
```diff
@@ -243,41 +293,62 @@ struct Rect {
...
+	 */
 	void moveTo(int16 x, int16 y) {
 		bottom += y - top;
 		right += x - left;
 		top = y;
 		left = x;
 	}
 
+	/**
+	 * @todo Doc required.
```
criezy, line 328: Move the rectangle by the given delta x and y values.

common/savefile.h
```diff
@@ -54,48 +54,89 @@ typedef SeekableReadStream InSaveFile;
  */
 class OutSaveFile: public WriteStream {
 protected:
-	WriteStream *_wrapped;
+	WriteStream *_wrapped; /*!< @todo Doc required. */
 
 public:
-	OutSaveFile(WriteStream *w);
+	OutSaveFile(WriteStream *w); /*!< Create a WriteStream instance @p w. */
```
criezy, line 60: It doesn't create a WriteStream instance. It creates an `OutSaveFile` that uses the given `WriteStream` to actually write the data.

Technically, since `OutSaveFile` is itself a `WriteStream`, it does create a `WriteStream`, but not @p w. The `w` WriteStream already exists.

criezy (commented):
I reviewed some more changes.
I have not checked the changes to `str.h` yet. I will do that tomorrow.

b-gent:
@criezy  thanks so much ! I will start implementing the comments tomorrow and let you know when done.

criezy:
There were significant changes to `common/str.h` this morning, so I will wait until that is resolved to review the changes in that file. One possibility would be to drop the changes to `common/str.h` from this PR and document it, as well as `common/base-str.h`, in a new PR.

b-gent:
> There were significant changes to `common/str.h` this morning, so I will wait until that is resolved to review the changes in that file. One possibility would be to drop the changes to `common/str.h` from this PR and document it, as well as `common/base-str.h`, in a new PR.

Yes, I can see there are quite a few changes there and resolving the conflict will be very error-prone because in most cases I modify the same lines. I guess we will go for the safer solution of moving this file to a separate PR.

common/array.h
```diff
@@ -221,16 +226,19 @@ class Array {
 
 	// TODO: insert, remove, ...
 
+	/** Return a pointer to the underlying memory serving as element storage. */
```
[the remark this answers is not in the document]
b-gent: Thanks for the explanation, these little bits of information really help to understand the code.

common/list.h
```diff
@@ -131,48 +131,49 @@ class List {
...
 
-	/** Removes the last element of the list. */
+	/** Remove the last element of the list. */
 	void pop_back() {
 		assert(!empty());
 		erase(_anchor._prev);
 	}
 
-	/** Returns a reference to the first element of the list. */
+	/** Return a reference to the first element of the list. */
```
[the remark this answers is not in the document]
b-gent: If 'returning' is the only thing that the function does, I think it's better to keep it as a ``@brief`` (we are using autobrief so all these descriptions are essentially a ``@brief`` element). Honestly, I'm not even sure if it would be possible to have a ``@return`` without a brief first. And duplicating the same thing in both brief and return does not make much sense.

That said, ``@return`` or ``@retval`` is of course a great thing to have in more complex functions that can return various values.

common/list.h
```diff
@@ -131,48 +131,49 @@ class List {
...
 		return static_cast<Node *>(_anchor._prev)->_data;
 	}
 
-	/** Returns a reference to the last element of the list. */
+	/** Return a reference to the last element of the list. */
 	const t_T &back() const {
 		return static_cast<Node *>(_anchor._prev)->_data;
 	}
 
+    /** @todo Doc required. */
```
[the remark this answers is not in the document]
b-gent: So ``Assign a given @p list to this list`` would be correct?

common/list.h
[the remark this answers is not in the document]
criezy: I am not familiar with `@brief` (we decided to turn it off at work) and forgot it exists. But I think that what you wrote makes sense.

common/rect.h
```diff
@@ -44,35 +44,69 @@ namespace Common {
...
+	/**
+     * Create a point by adding the @p delta value to the x and y arguments.
+     */
 	Point operator+(const Point &delta) const { return Point(x + delta.x, y + delta.y); }
+	/**
+     * Create a point by subtracting the @p delta value from the x and y arguments.
+     */
 	Point operator-(const Point &delta) const { return Point(x - delta.x, y - delta.y); }
+	/**
+     * Create a point by dividing the x and y arguments by the @p delta value.
```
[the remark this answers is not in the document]
b-gent: yes, I overlooked a copy-paste issue, thanks

common/rect.h
b-gent: Changed to 
```
	/**
     * Create a rectangle with top-left corner at the given (x1,y1) position
	 * and the bottom-right corner at the (x2,y2) position.
	 * 
	 * The @p x2 value must be greater or equal @p x1 and @p y2 must be greater or equal @p y1.
     */
```

common/rect.h
```diff
@@ -44,35 +44,69 @@ namespace Common {
...
+	/**
+     * Determine whether the position of two points is the same.
+     */
 	bool  operator==(const Point &p)    const { return x == p.x && y == p.y; }
+	/**
+     * Determine whether the position of two points is not the same.
+     */
 	bool  operator!=(const Point &p)    const { return x != p.x || y != p.y; }
+	/**
+     * Create a point by adding the @p delta value to the x and y arguments.
```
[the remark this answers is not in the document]
b-gent: Changed to
`Create a point by adding the @p delta value to a point.`
(and similar in all of these)

b-gent:
@criezy  all your comments are addressed now

b-gent:
I have reverted ``str.h`` to the state of master but Github is still showing that I changed things in that file. But that's not right, for example it is showing that I added `#include "common/base-str.h"` , while this has been recently introduced to master through another PR.

criezy (approved):

criezy:
I merged the commit manually to double check what was happening with the `str.h` file.

PR #2536, opened 2020-10-17
GRAPHICS: Problem with virtual mouse position in an overlay when full…

lmerckx:
There is a problem with the virtual mouse position when game is in an overlay (menu, warning message) in fullscreen mode (for OPENGL, OPENGLES):
The cursor moves far too fast and stays stuck at the bottom-right corner.

It is due to the fact that coordinates are converted as it was still in game.
So a test "!overlayVisible" must be added, as it is done in the reverse method (transformMouseCoordinates).


aquadran:
what case do you refer? I don't see any problem on my side

lmerckx:
Occurs on my side:
- When emulating mouse with joystick
- In fullscreen mode, with OPENGL and OPENGLES
- On x64 (Ubuntu) and Raspberry
- When overlay is open (ScummVM menu, warning message)
- on ResidualVM games (GRIM, EMI)

aquadran:
ok, I got repro too. fix helps, merging

PR #2537, opened 2020-10-17
ALL: Run optipng to reduce png file size

mduggan:
The `optipng` tool reduces png file size with no difference in output.  This helps create smaller deployables, and in my experience it's totally safe.

Doing as a PR anyway so people can have a look and check it doesn't break anything.


henke37:
Not a bad idea, but I'd like to see it automated.

dreammaster:
Good idea. Though at least for the Ultima related files, the .dat file would need to be regenerated to take advantage of the smaller sizes, and I'm a bit leery of doing that, and adding to the size of the Git repository, unless we have some other bugfixes or new content being added to warrant it.

mduggan:
> Not a bad idea, but I'd like to see it automated.

Do you mean in a script? I don't think we should go as far as adding a git hook, so here's the script..
```
for x in  `find . -name \*.png`; do optipng $x; done
```
I guess I could add that to the Makefile, but it couldn't be fully automatic, you basically have to run the program to see if it does anything, so you'd just have a make target you'd have to know about, at which point you can almost as easily just write that command above :)

mduggan:
> .. I'm a bit leery of doing that, and adding to the size of the Git repository, unless we have some other bugfixes or new content being added to warrant it.

Agreed - it's probably not worth generating the `ultima.dat` just for this, but we can put this there for the next time it gets regenerated for some other reason - I'm sure it will come up at some point.

mgerhardy:
Does `optipng` use quantization? I've used [pngquant](https://pngquant.org/) to reduce the size for my other games a lot.

mduggan:
optipng should leave the pixel data identical, it just recompresses using more efficient tables etc.  There's some technical details on what it does here:
http://optipng.sourceforge.net/pngtech/optipng.html

mduggan:
I'll close this because it's got conflicts now, but since the overall response is positive I'll push a similar change to master.

PR #2539, opened 2020-10-17
PRINCE: fix bug in the conversation with the priest

dtelyukh:
BUG: https://bugs.scummvm.org/ticket/11771
This conversation have no audio files, because it is performed using gestures. So it shouldn't try to load audio.


sev-:
Thanks, makes sense.

PR #2540, opened 2020-10-17
GRIM: Fix mapping of joystick action buttons in EMI

lmerckx:
Joystick buttons are not working in EMI for actions:
It seems that key mapping doesn't map joystick buttons to correct key events.

According to manual: https://mocagh.org/lucasfilm/mi4-manual.pdf

- Examine/Look is E or L
- Use/Talk is U
- Pick up/Put in inventory is P or + (keypad)
- Inventory is I or Insert

With these modifications, the joystick buttons are now working.
Tested on x64 (Ubuntu) and Raspberry 2 and 3 with the French Windows and PS2 versions of the game.


aquadran:
looks like I forgot about key codes, thx

PR #2543, opened 2020-10-17
WINTERMUTE: Opengl es 2 conformance

GunnarBirke:
This should remove all non-conforming OpenGL calls from the shader renderer.


aquadran:
looks great, thx

PR #2544, opened 2020-10-18
KYRA: Support Russian LoK translation by Siberian GRemlin

dgladkov:
Siberian GRemlin implemented a modern and fairly recent (last update 2017) translation of the LoK CD edition , which includes full translation of the image assets.

* Verified that both translated subtitles and original English audio work as expected.
* Verified that all menu items are translated correctly and work as expected.
* Tested playthrough of roughly 1/3 of the game without any issues.
* Translation is marked as fan to force subtitles by default.


athrxx:
Okay, this one is a bit easier than the other one :-)

Normally, the talkie versions don't show the story screen if only speech is enabled and subtitles are disabled. For the Italian fan translation it seems that the story screen is always displayed (probably because the spoken language is still English). See sequences_lok.cpp, lines 250-251. Do you want it like that for Russian? I leave it up to you...

dgladkov:
This is a very good suggestion, but I feel it's not enough as the game still starts with subtitles off by default and first time players are going to miss the subtitles in subsequent into scenes. However looks like there's already the code that fixes both problems perfectly:

https://github.com/scummvm/scummvm/blob/6f8aa0b21b90b3940f929708ac65154e8a131f66/engines/kyra/engine/kyra_v1.cpp#L564-L571

I ended up marking Russian translation as fan translation to force subtitles and left story screen logic as it was since players will leave subtitles on anyway.

dgladkov:
At this point, I'm OK with merging this PR unless there are comments or objections. Thanks for reviews and helpful tips, @athrxx, much appreciated:)

athrxx:
Thanks for your work on this :-)

sev-:
In the end, we had to mark this version as unsupported by the author's request.

PR #2545, opened 2020-10-18
KYRA: Add localized strings for the Russian CD Release of HoF

dgladkov:
Add localized strings from the Fan translation of the HoF. The same edition is already detected correctly as Russian CD release.


athrxx:
Thanks!

PR #2547, opened 2020-10-19
IMAGE: Add decoder for CEL 3DO format

phcoder:
This format is used by 3DO version of Plumbers Don't Wear Ties
but it's a generic format on 3DO similar to BMP on windows.


bluegr:
Closing in favor of #2551

PR #2548, opened 2020-10-19
ALL: HashMap & ConfMan getVal refactoring

henke37:
Refactoring the getVal function as seen in HashMap and ConfMan. It is now an error to try to get the value of an unknown key, with the variant for getting a default being called out by name. There is a separate `getValOrDefault` function for when a default is actually requested.

### DIFF
--- a/backends/keymapper/keymap.cpp
+++ b/backends/keymapper/keymap.cpp
@@ -68,7 +68,7 @@ void Keymap::addAction(Action *action) {
 }
 
 void Keymap::registerMapping(Action *action, const HardwareInput &hwInput) {
-	ActionArray &actionArray = _hwActionMap.getVal(hwInput);
+	ActionArray &actionArray = _hwActionMap.getOrCreateVal(hwInput);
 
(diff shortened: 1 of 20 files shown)


criezy:
I didn't look at the details of the changes, but the idea seems good to me. This clarifies the intend of each function and may help to catch errors in the code using those.

criezy:
And I have now tried to use that code, and it crashes immediately when I start ScummVM.
The first crash is in LauncherDialog::updateListing():
```
		String description(iter->_value.getVal("description"));

		if (description.empty()) {
			QualifiedGameDescriptor g = EngineMan.findTarget(iter->_key);
			if (!g.description.empty())
				description = g.description;
		}
```

henke37:
I'm going to try one more tweak to handle `operator []` gracefully.

henke37:
I will be continuing my work on that idea on a separate branch, this PR has been stalled for long enough.

sev-:
okay, merging this

PR #2549, opened 2020-10-20
CREATE_PROJECT: fixed cmake generator

mgerhardy:
With these patches I'm now able to configure, compile and link with cmake

* Support for pkg-config was added
* Shortened the CMakeLists.txt by more than 4000 lines
* Added macro for feature detection
* Configure stage is reporting missing libs as error now (they were reported as compile errors during the build before)
* Link the scummvm-detection lib (this was missing before)
* Added missing defines like `DETECTION_STATIC` and `USE_SDL2`
* Added the generated `CMakeLists.txt` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ lib*.a
 /scummvm-conf.cpp
 /tmp_*.cpp
 /*.apk
+/CMakeLists.txt
 
 /README
 /README.html*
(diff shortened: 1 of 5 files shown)


lotharsm:
I ran into similar issues when I played around with create_project. +1 from me.

mgerhardy:
With the latest commit I'm now able to not only configure the build system - but also compile and link the binary. But

```
 ~/dev/scummvm/build   remotes/mgerhardy/pr/create-project-cmake-pkgconfig ? ⍟53  ./scummvm --list-engines                                                                                                                                                               ✔  17:31:09
Engine ID       Engine Name                                           
--------------- ------------------------------------------------------
```
There are also no shared objects created. So I suppose there is some define missing here?

mgerhardy:
This also fixes some missing defines since the detection plugin rework.

mgerhardy:
[CMakeLists.txt](https://github.com/scummvm/scummvm/files/5506068/CMakeLists.txt)

Attached a generated cmake script

PR #2552, opened 2020-10-20
VIDEO: uplift 3do movie decoder from sherlock engine

phcoder:
The format is generic to 3DO and is also used by plumbers.
I think it's also accelerated on 3DO so probably is used by
a lot of 3DO titles


bluegr:
Closing in favor of #2553

bluegr:
Actually, #2551 is the full PR

PR #2553, opened 2020-10-20
VIDEO: support multi-track audios in 3DO videos

phcoder:
It's used in plumbers with first track being speech and second track
being music


bluegr:
Closing in favor of #2551

PR #2555, opened 2020-10-22
GUI: Fix keymap of virtual keyboard if no GUI opened before

lmerckx:
When starting a game from the command line (without using the menu) and opening the virtual keyboard, the keymap `kKeymapTypeGui` is not defined and it is not possible to click on virtual keyboard with a joystick/controller.

So, this patch initializes the `kKeymapTypeGui` keymap if it was not already done before (using the standard GuiManager mapping).

Perhaps, there are other solutions to solve this issue: For instance, make `GuiManager::initKeymap()` public and directly call it.
I'm open to any suggestion.

### DIFF
--- a/backends/vkeybd/virtual-keyboard.cpp
+++ b/backends/vkeybd/virtual-keyboard.cpp
@@ -24,6 +24,7 @@
 
 #ifdef ENABLE_VKEYBD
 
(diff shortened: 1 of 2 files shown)


aquadran:
I think we can merge it as it is

PR #2556, opened 2020-10-24
I18N: Update Korean translation

wonst719:
I was making a bunch of suggestions on https://translations.scummvm.org, then I realized that it would be easier to edit the translation file manually.
So this is (hugely) updated Korean translation.
Almost all GUI strings are done, port-specific and engine-specific strings are in progress.
There are inconsistencies here and there, so I might make additional PRs.


sev-:
ah, thanks! I think, it also could be easy if I add you as a translator. I'm doing it

sev-:
And while I have your attention. Shall we work on merging in your scummvm-ko changes? Please, drop me a line at sev@scummvm.org

wonst719:
Thanks. Regarding to scummvm-kor, I will write to you shortly.

PR #2557, opened 2020-10-24
TWINE: Add detection for UK version of Little Big Adventure

eriktorbjorn:
This adds detection for the UK release of Little Big Adventure. This is what my CD looks like:

https://www.mobygames.com/game/dos/relentless-twinsens-adventure/cover-art/gameCoverId,198155/

And this is what the box looks like:

https://www.mobygames.com/game/dos/relentless-twinsens-adventure/cover-art/gameCoverId,198153/

It may be misleading to call it an "UK release", because the "vox" directory (which I added to the search manager, even though I'm not sure if it's used) contains files prefixed "en_", "de_" and "fr_". But the CD says "Made in the U.K.", and the manual - which I **clearly** need to read, because right now I'm just hitting buttons blindly - is in English, so... But feel free to edit the description in any way.

The "fla" directory is probably animated cutscenes, or something like that? I definitely had to add that to the search manager.

Anyway, the game starts and runs for a bit. So that's promising.

### DIFF
--- a/engines/twine/detection.cpp
+++ b/engines/twine/detection.cpp
@@ -59,6 +59,15 @@ static const ADGameDescription twineGameDescriptions[] = {
 		ADGF_UNSTABLE,
 		GUIO1(GUIO_NONE)
 	},
+	{
+		"twine",
+		"",
+		AD_ENTRY1s("text.hqr", "31d880f658cc6cc6d6cf70df732aec4f", 248829),
+		Common::EN_GRB,
+		Common::kPlatformDOS,
+		ADGF_UNSTABLE,
+		GUIO1(GUIO_NONE)
+	},
 	AD_TABLE_END_MARKER
 };
 
--- a/engines/twine/twine.cpp
+++ b/engines/twine/twine.cpp
@@ -68,6 +68,11 @@ namespace TwinE {
 
 TwinEEngine::TwinEEngine(OSystem *system, Common::Language language, uint32 flags)
     : Engine(system), _gameLang(language), _gameFlags(flags), _rnd("twine") {
+	// Add default file directories
+	const Common::FSNode gameDataDir(ConfMan.get("path"));
+	SearchMan.addSubDirectoryMatching(gameDataDir, "fla");
+	SearchMan.addSubDirectoryMatching(gameDataDir, "vox");
+
 	setDebugger(new GUI::Debugger());
 	_actor = new Actor(this);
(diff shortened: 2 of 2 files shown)


mgerhardy:
Yes, hard to say - the GOG version is afair sold as english - but german, english and french are also included. I would say I will add those languages to your release, too.
Thanks a lot.

eriktorbjorn:
This detection entry was added in https://github.com/scummvm/scummvm/commit/f2ae4b1b87cb7fd3b6c383c1385e4db5e6174893

It was further extended in https://github.com/scummvm/scummvm/commit/1cf46c4ee0232dd3b2aee5b5d4728697e7282fb3

Then it was removed in https://github.com/scummvm/scummvm/commit/98df1b2020e309db636174bd22f344614ac6112e and now my version of the game is no longer recognized.

mgerhardy:
Weird - that was the GOG version - where is your version from? The GOG version is still detected for me.

eriktorbjorn:
It's a "Made in the U.K." CD version. There are links to what the CD and box look like in the pull request description. I'm pretty sure there were detection entries for the GOG version before the one I submitted was added.

It appears to be multi-lingual, since the installation program gives me a choice between English, Francais, Deutsch, Español and Italiano. I only see "de_", "en_" and "fr_" prefixes for the .vox files, though. It's one of those games that appears to require some fiddling around to get it to install in DOSBox, so I haven't checked.

eriktorbjorn:
This was the signature I submitted:

`AD_ENTRY1s("text.hqr", "31d880f658cc6cc6d6cf70df732aec4f", 248829),`

eriktorbjorn:
I finally managed to get the game running in DOSBox. It _is_ a multi-lingual version, even though the printed manual is only in English. (I think. I bought it second-hand.):

- English (speech and text)
- French (speech and text)
- German (speech and text)
- Italian (text only)
- Spanish (text only)

I can pick any combination of text and speech (or no speech), though I have to use the installation/setup program to do so. I couldn't find any in-game menu for it.

mgerhardy:
I'm going to re-add your entry in the evening. Sorry for this.

Most lba distributions (or better all I know) have at least these languages included. In ScummVM you will be able to select them at runtime - as the installer ist just adding a `Language: Value` to the ini file of the installation

mgerhardy:
I've re-added them in  b13c10f0b4818686c1c6976a24a96848944f7e92

PR #2558, opened 2020-10-24
SWITCH: Add support for 3d games

usineur:
This commit add support for 3d games on Nintendo Switch.

Note that, event if devkitpro toolchain provides a switch-glad package, it's designed for OpenGL 4.3.
That's why i had to add some customized glad stuff in "graphics".
If this is a problem, it could be considered to outsource it to a library for example.


aquadran:
From what I understand this glad is generated for specific platform and customised. I'm afraid this can not be in common code.
Also Importing library into source code is not best practice. Like for example we don't import GLEW into source tree.
What is customised in glad for Switch? Maybe changes could be sent upstream into glad Switch toolkit?
What about GLEW support on switch, exists?

As we need to able compile port by others, the glad sources with changes must be placed somewhere.
Maybe put it into backend specific for switch? Still I would go this direction as bottom line.

usineur:
No there is no GLEW support on Switch.

Here, I provided one generated for 3.0 compatibility, which adds "OpenGL" as a game 3d renderer and with GL_ARB_framebuffer_object and GL_EXT_framebuffer_object extensions.

Exvat:
> No there is no GLEW support on Switch.

Just because GLEW is not provided by the package manager doesn't mean it can't be compiled for the Switch. As far as I know glad is only provided in pacman for convenience, not because it's the only option for binding OpenGL functions. Under the hood glad is using EGL to bind the functions on the Switch, which is also supported by GLEW.

usineur:
> Just because GLEW is not provided by the package manager doesn't mean it can't be compiled for the Switch.

You are right certainly, i will try to see if I can port it.

Exvat:
Now that I think about it there is one issue you'll run into: GLEW actually assumes the system already provides a way to link to OpenGL 1.1 and doesn't attempt to bind any of those functions. This is not true on the Switch where devkitpro doesn't provide any GL functions to link to without binding them.

aquadran:
"Here, I provided one generated for 3.0 compatibility, which adds "OpenGL" as a game 3d renderer" do not assume it's compatible with others. this is why its more Switch specific.

"but shaders for EMI / Grim are incomplete" what you mean by that? "OpenGL with shaders" renderer is working for long time

aquadran:
when I think more about it. I think glad should be as external lib only. not included into our code base. it's regenerated to specific opengl(es) setup. this could be extended to similar as glew support. but I think we should explore glew port option first.

sev-:
> when I think more about it. I think glad should be as external lib only. not included into our code base. it's regenerated to specific opengl(es) setup. this could be extended to similar as glew support. but I think we should explore glew port option first.

Well, another viable alternative is to put it to backends/platforms/switch/

Exvat:
I've created a patch file that shows the changes necessary to glew to add devkitpro Switch support: https://gist.github.com/Exvat/c6b799a13a4d68e6b2097f0f849ce04f

aquadran:
> > when I think more about it. I think glad should be as external lib only. not included into our code base. it's regenerated to specific opengl(es) setup. this could be extended to similar as glew support. but I think we should explore glew port option first.
> 
> Well, another viable alternative is to put it to backends/platforms/switch/

I'm rather against to include glad currently, it's like include glew into our repo. btw. both has very similar goal.
glew was not added into residualvm before as we didn't want to maintenance it. do we want add third party libs?
I think it's better to include glew than glad, as last one it's just output from code generator for specific setup.

usineur:
@Exvat thx for your help, I can confirm that it's working well and we can now avoid to use glad

@aquadran: When I said "incomplete", it's not the right wording I guess. For example, in EMI, when using "OpenGL with shaders" renderer, these kind of glitches are present:

<details>
<summary>Screenshots</summary>

emi1]
emi2]
</details>

I also made changes which now required Exvat's [pull](https://github.com/devkitPro/pacman-packages/pull/204) to be merged

engines/grim/gfx_opengl.h
```diff
@@ -29,7 +29,7 @@
 
 #ifdef USE_GLEW
 #include <GL/glew.h>
-#elif defined (SDL_BACKEND) && !defined(__amigaos4__)
```
aquadran, line 32: what happen here?

backends/platform/sdl/sdl.cpp
```diff
@@ -218,7 +218,9 @@ void OSystem_SDL::initBackend() {
 
 #if defined(USE_OPENGL_GAME) || defined(USE_OPENGL_SHADERS) || defined(USE_GLES2)
 	detectFramebufferSupport();
+#ifndef NINTENDO_SWITCH
```
aquadran, line 221: what is the reason not supporting AA detection?

backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
```diff
@@ -262,7 +262,7 @@ void OpenGLSdlGraphics3dManager::createOrUpdateScreen() {
 		g_system->quit();
 	}
 
-#ifdef USE_GLEW
+#if defined(USE_GLEW) && !defined(NINTENDO_SWITCH)
```
aquadran, line 265: why glewinit is moved?

rsn8887:
I would Bengali to compile a customized glad for Switch and add it to the buildbot toolchain, that wouldn’t be any problem. All I need is a link to a verified working source repo and build instructions.

engines/grim/gfx_opengl.h
usineur: oops, wrong line removed, fixed

backends/platform/sdl/sdl.cpp
usineur: detectAntiAliasingSupport create multiple SDL window, which is currenlty not supported by Switch SDL2 implementation

backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
usineur: Directly after "_glContext = SDL_GL_CreateContext(_window->getSDLWindow());", there is a call to glClear, which is failed until glewInit si called.
It is the glew library that supports links to OpenGL 1.1 functions on Switch port.

aquadran: can you put some comment on this, here and below?

backends/platform/sdl/sdl.cpp
aquadran: ok, does it support AA at all?

Exvat: It does, but you have to create an FBO for it and then resolve it into the backbuffer. The devkitpro OpenGL implementation does not support a multisample backbuffer.

backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
usineur: Done

backends/platform/sdl/sdl.cpp
usineur: Tbh, I don't know. I'm currently asking to devkitpro devs.

aquadran: ok, then. please move "#ifndef NINTENDO_SWITCH" into "detectAntiAliasingSupport" function

Exvat: @usineur I'm pretty sure it does support it, you can query `GL_MAX_SAMPLES` to check. It just doesn't support a multisampled context which ScummVM expects, you can only do it by rendering to an FBO.

Exvat: Looking at the code we already have the necessary render-to-FBO code and ScummVM won't attempt to use a multisample context in that case. The only problem is it still uses a multisample context for detection.

We can very simply solve this however by adding an alternative implementation to the guard. We should simply query `GL_MAX_SAMPLES` and then run a loop `for (int i = 1; i <= maxSamples; i <<= 1)` which adds i to the list of sample sizes supported.

aquadran: indeed, it might work

usineur: I don't see how to do that in "detectAntiAliasingSupport" function as glew hasn't be initialized yet, and a call to 
`glGetIntegerv(GL_MAX_SAMPLES, (GLint *)&maxSamples);` will result in a crash.

Exvat: In that case it would be necessary to move the call `detectAntiAliasingSupport`, which should be fine since the result of it is only used in the UI and nowhere else.

usineur:
https://github.com/devkitPro/mesa/pull/3

This basically means that "OpenGL" renderer can't be supported on Switch as is, due to specific changes in Switch's Mesa 20.1.0.

If no one sees any objection, I will adapt my changes to support "USE_GLES2" mode only. I think we can live with that.

Exvat:
@usineur If you're using GLES2, there's no need for GLEW or glad. GLES2 can be directly linked to on switch.

Are there any benefits we give up with OpenGL? There's always the option of simply forking since mesa is statically linked with each homebrew application.

aquadran:
yes, if platform is capable gles2 you can use it without glew.
if platform provide both OpenGL and gles2, configure script pickup OpenGL as default. you have to use '--force-opengl-game-es2' option for configure

usineur:
@Exvat There is no real benefit, but it simplifies integration. I don't believe a fork in a viable way.

@aquadran Changes pushed.

I still have to look for AA, but it might be done in a second time.

@rsn8887 Switch port must be built without "wme3d" engine, it fails at build for now.

aquadran:
@usineur what fail on wme3d? other ports with gles2 compiles fine, so my be something new

usineur:
@aquadran https://pastebin.com/wc1NWn9X

aquadran:
@usineur I think I fixed, you can try it

usineur:
@aquadran It doesn't launch unfortunately

```
Could not compile shader wme_modelx.vertex: 0:44(62): warning: Float suffixes are invalid in GLSL 1.10
0:54(26): warning: Float suffixes are invalid in GLSL 1.10
0:64(68): warning: Float suffixes are invalid in GLSL 1.10
0:63(6): error: could not implicitly convert operands to relational operator
0:63(6): error: if-statement condition must be scalar boolean
0:66(24): error: no matching function for call to `normalize()'; candidates are:
0:66(24): error:    float normalize(float)
0:66(24): error:    vec2 normalize(vec2)
0:66(24): error:    vec3 norma!
```

aquadran:
@usineur I think error is fixed now

usineur:
@aquadran I can confirm. It works now. Tested with Arts of murder

PR #2561, opened 2020-10-24
TWINE: tentative fix for amigaos4

phcoder:


sev-:
I've took care of it :)

raziel-:
Thank you :-)

PR #2562, opened 2020-10-24
HADESCH: Warning pass

phcoder:
Fixes all MSVC warnings and many on minor platforms


sev-:
Thanks! Please commit directly now :)

PR #2565, opened 2020-10-25
COMMON: Fix game "Just Grandma and Me" german and french version crash

besentv:
The german and french version of "Just Grandma and Me" (Mohawk engine) have the special characters "ç" and "ö" in their config files. This causes ScummVM to crash when trying to start the game in german or french. Adding special checks for these chars fixes this issue.


SupSuper:
It would make more sense to enable `INIFile::allowNonEnglishCharacters` for these games, rather than change the common INI behavior across all game engines.

besentv:
Okay that seems to be a better solution. Should I look into it or is someone else going to do it?

PR #2566, opened 2020-10-26
STARK: Play sound when hovering active cursor

dgladkov:
* Play sound when hovering over action menu items. Fixes https://bugs.scummvm.org/ticket/11875
* Play sound when hovering useable items. Fixes https://bugs.scummvm.org/ticket/11874
* Rename `_fading` cursor property to `_itemActive` to reflect that this property now controls more than just fading.

### DIFF
--- a/engines/stark/ui/cursor.h
+++ b/engines/stark/ui/cursor.h
@@ -35,6 +35,10 @@ namespace Gfx {
 class Driver;
 }
 
+namespace Resources {
+class Sound;
+}
+
 /**
  * Manager for the current game Cursor
  */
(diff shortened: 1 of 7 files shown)


engines/stark/ui/world/gamewindow.cpp
```diff
@@ -117,7 +117,6 @@ void GameWindow::onRender() {
 
 void GameWindow::onMouseMove(const Common::Point &pos) {
 	_renderEntries = StarkGlobal->getCurrent()->getLocation()->listRenderEntries();
-	_cursor->setFading(false);
```
dgladkov, line 120: Since setting the flag has a side effect of controlling sound playback now, we should avoid setting it back and forth to prevent playback artifacts.

engines/stark/ui/cursor.h
```diff
@@ -87,7 +91,9 @@ class Cursor {
 	VisualText *_mouseText;
 	CursorType _currentCursorType;
 
-	bool _fading;
+	Resources::Sound* _actionHoverSound;
```
somaen, line 94: Formatting: (Asterisk aligns to variable, not type)

Resources::Sound *_actionHoverSound;

somaen (commented):
The second commit should also have a "STARK:" prefix.

aquadran:
thx for fixes!

PR #2567, opened 2020-10-26
STARK: Workaround to support Emma's lipsync

dgladkov:
Fixes https://bugs.scummvm.org/ticket/11872

Alternatively, we can fix material name during initialization.


aquadran:
thanks for the patch!

PR #2570, opened 2020-10-26
SCI: Add RTL support in GfxControls16::drawListControl

ZvikaZ:
When using original save/load screens, the games list wasn't properly
aligned, if contained Hebrew text. Fixed by calling the BiDi algorithm
in `drawListControl`.


bluegr:
Thanks, merging

PR #2571, opened 2020-10-26
SCI: Hebrew - fix save games encoding

ZvikaZ:
Make sure that Hebrew save games are using UTF8 (otherwise, when saving
in Hebrew it works well from original Sierra save/restore dialogs, but
when loading from ScummVM's dialog, it looks garbled)


bluegr:
Thanks, merging

PR #2572, opened 2020-10-27
I18N: Always call BiDi algorithm for U32String

ZvikaZ:
U32String can contain characters from all languages, without relation
to the GUI's language. Therefore, we need to always call it.

Use case:
In Hebrew SQ3, when using the game's original save/restore dialogs, it's
possible to give save games Hebrew names.
What happens when we then use ScummVM's load dialog?
- If ScummVM's GUI language is Hebrew, it's calling BiDi algo, and
  the text direction is OK.
- However, if ScummVM's GUI language is English, before this commit it
  didn't call the BiDi algo, and the text was reversed.

### DIFF
--- a/common/translation.cpp
+++ b/common/translation.cpp
@@ -412,9 +412,6 @@ bool TranslationManager::checkHeader(File &in) {
 }
 
 U32String TranslationManager::convertBiDiString(const U32String &input) {
-	if (getCurrentLanguage() != "he")		//TODO: modify when we'll support other RTL languages, such as Arabic and Farsi
-		return input;
-
 	return Common::convertBiDiU32String(input).visual;
 }
 
--- a/common/translation.cpp
+++ b/common/translation.cpp
@@ -411,10 +411,6 @@ bool TranslationManager::checkHeader(File &in) {
 	return true;
 }
 
-U32String TranslationManager::convertBiDiString(const U32String &input) {
-	return Common::convertBiDiU32String(input).visual;
-}
-
 } // End of namespace Common
 
 #endif // USE_TRANSLATION
--- a/common/translation.h
+++ b/common/translation.h
@@ -169,13 +169,6 @@ public:
 	 */
 	String getCurrentLanguage() const;
 
-	/*
-	 * Wrapper for GNU FriBidi implementation of the Unicode Bidirectional Algorithm
-	 * For LTR (Left To Right) languages, returns the original input
-	 * For RTL (Right To Left) languages, returns visual representation of a logical single-line input
-	 */
-	U32String convertBiDiString(const U32String &input);
-
 private:
 	/**
 	 * Tries to find the given language or a derivate of it.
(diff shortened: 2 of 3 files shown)


criezy:
The change seems simple and logical.
But maybe we should just remove `convertBiDiString` from `TranslationManager` altogether and always use `Common:: convertBiDiString`?

I am wondering though, what is the performance impact of always doing the conversion?
I guess if this is an issue on some low end platforms, we should just compile ScummVM without fribidi for those anyway as performance would be an issue when using hebrew GUI otherwise anyway.

ZvikaZ:
1.
Removing `convertBiDiString` from `TranslationManager`  and always using `Common:: convertBiDiString` makes sense; however, note that `TranslationManager::convertBiDiString` returns `Common::convertBiDiU32String(input).visual`. I think that the `.visual` suffix isn't very intuitive (honestly, I don't even remember why is it there :-) ), and it's an opening of weird bugs because of "clients" of this method forgetting it...

2.
Agree, I don't think that the performance penalty is of issue here, as you described it.

phcoder:
> I am wondering though, what is the performance impact of always doing the conversion?
> I guess if this is an issue on some low end platforms, we should just compile ScummVM without fribidi for those anyway as performance would be an issue when using hebrew GUI otherwise anyway.

If performance is a concern then the code should first check if the string contains any bidi-relevant characters like strong RTL, weak RTL and so on and if it doesn't, then just pass the string through, ideally in a zero-copy way

ZvikaZ:
So, let me summarize.
We have 2 points here:

1. Maybe we should remove `convertBiDiString` from `TranslationManager`.
@criezy - what do you think? should we do that? or maybe it's better not because of the `.visual` suffix I've mentioned above?

2. Possible performance problem on slower device.
We have here @phcoder 's suggestion - but I don't think we want to to do it:
- if it's a "strong" machine, there isn't performance issue to solve
- if it's a "weak" machine, it'll probably be compiled without FriBiDi support anyway, so again, there isn't performance issue to solve

criezy:
For your first point I would say yes. And from what I can see it is only used in one place anyway (in ThemeEngine.cpp).
If we don't like the .visual suffix we can always add a `Common::convertBidiString` overload that takes a U32String and returns a U32String in common/unicode-bidi.h. But I don't think the TranslationManager is the right place for such a function.

ZvikaZ:
As you suggested, I've removed it from `TranslationManager`.
And regarding the `.visual` - I tried without it, and the Hebrew GUI looks great.
It's indeed not needed.

criezy:
> And regarding the `.visual` - I tried without it, and the Hebrew GUI looks great.
> It's indeed not needed.

I am actually surprised about that. But after double checking, there is indeed a `U32String` constructor that takes a `UnicodeBiDiText`, so it indeed works, but at the cost of making an additional copy of the string.

criezy:
I will merge this now, but I may change the way we convert from UnicodeBiDiText to U32String afterward.

PR #2573, opened 2020-10-27
COMMON: Several improvements for U32String

phcoder:


sev-:
This now needs to be reworked or dropped.

PR #2575, opened 2020-10-27
MOHAWK: Fix game "Just Grandma and Me" german and french version crash

besentv:
The german and french version of "Just Grandma and Me" (Mohawk engine) have the special characters "ç" and "ö" in their config files. This causes ScummVM to crash when trying to start the game in german or french. Allowing non english chars in config files for Living Books games fixes this issue.


besentv:
Based on the suggestion from: https://github.com/scummvm/scummvm/pull/2565

sev-:
Makes sense, thanks!

PR #2576, opened 2020-10-28
SCI: Fix font rendering with sq3-mac

phcoder:
Without this change all fonts result in empty/fully transparent glyphs.
I'm not 100% sure that it's the right solution but it allows to progress the
game to at least 15 points


bluegr:
The change itself isn't correct, as the Mac version of SQ3 ran at a 480x300 resolution. Having said that, the font rendering is broken at this upscaled resolution, so the upscaling code itself should be fixed.

bluegr:
Fixed in c5df23452150bdfcb959bb34efe1e519c383843a (although text looks awful at the moment). Closing

PR #2577, opened 2020-10-28
SCI: Fix crash when running sq6-mac with macbinary

phcoder:
The resource validation fails big time because no file Data9 exists.
It's expected as its in Data9.bin instead. Since we check resource fork
explicitly in the case of resource-fork resources this check is redundant

### DIFF
--- a/engines/sci/resource.cpp
+++ b/engines/sci/resource.cpp
@@ -2187,16 +2187,22 @@ Resource *ResourceManager::updateResource(ResourceId resId, ResourceSource *src,
 	// Update a patched resource, whether it exists or not
 	Resource *res = _resMap.getVal(resId, nullptr);
 
-	Common::SeekableReadStream *volumeFile = getVolumeFile(src);
-	if (volumeFile == nullptr) {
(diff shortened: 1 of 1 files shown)


sluicebox:
I'm confused, what is it about this situation that is specific to SQ6 or its Data9 file? All the SCI1.1/SCI32 Mac games have Data# files, and SQ6 comes with Data1 through Data12 files, including Data9.

I haven't used the macbinary format for storing SCI Mac files, but I know someone who does and they've been able to run SCI32 Mac games, so I don't understand what issue this fixes that wouldn't be a problem for every game / file.

trembyle:
Both KQ7 and SQ6 did not start for me in MacBinary format before this patch. They work now (with known sound issues in SQ6).

phcoder:
> To clarify, this is an issue with all SCI Mac "Data#" files, right? It's not specific to SQ6 or Data9?

Right. It's about all Data# on mac versions of SCI32 games

bluegr:
Nice work, thanks!

Merging

PR #2578, opened 2020-10-28
COMMON: Merge code for str.cpp and ustr.cpp

phcoder:
Most of the code in str.cpp and ustr.cpp is actually the same. Do some
template magic to merge them.


bluegr:
This has also been discussed in #2452

phcoder:
Updated to remove second parameter as per feedback

common/serializer.h
```diff
@@ -264,6 +264,29 @@ class Serializer {
...
+	void syncString32(U32String &str, Version minVersion = 0, Version maxVersion = kLastVersion) {
+		if (_version < minVersion || _version > maxVersion)
+			return; // Ignore anything which is not supposed to be present in this save game version
+
+		uint32 len = str.size();
+
+		syncAsUint32LE(len);
+
+		if (isLoading()) {
+			uint32 *sl = new uint32[len];
```
criezy, line 279: Maybe use `U32String::value_type` rather than `uint32` in case we decide to use something else (such as `char32_t`).

common/ustr.h
```diff
@@ -52,66 +53,24 @@ class UnicodeBiDiText;
  * The presence of \0 characters in the string will cause undefined
  * behavior in some operations.
  */
-class U32String {
+class U32String : public BaseString<uint32> {
```
criezy, line 56: Maybe use `char32_t` instead of `uint32` when `USE_CXX11` is defined to be consistent with UTF-32 string literal in C++11. That would also allow creating a `U32String` from a `U"C++11 UTF-32 string literal"`.

An alternative to this would be to keep using `uint32` but at least add a constructor that takes a `const char32_t *` so that we can create a `U32String` from a UTF-32 string literal.

criezy (commented):
Good work!
My only comments are related to the char type to use for U32String as C++11 introduce the `char32_t` for UTF-32 strings, as well as UTF-32 literal strings (of type `const char32_t *`).

common/serializer.h
phcoder: done

common/ustr.h
phcoder: Done

sev-:
Thanks, merging.

PR #2582, opened 2020-10-29
COMMON: added MemorySeekableReadWriteStream

mgerhardy:
A seekable read and writeable memory stream that operates on an already existing memory buffer

I need this for TwinE - the others are copying the memory - this one operates on the given memory buffer. If it's not worth it to add to `Common` it will just live in `TwinE` namespace.

Opinions?

### DIFF
--- a/common/memstream.h
+++ b/common/memstream.h
@@ -342,6 +342,96 @@ public:
 	byte *getData() { return _data; }
 };
 
+/**
+ * A seekable read and writeable memory stream that operates on an already existing memory buffer
+ */
+class MemorySeekableReadWriteStream : public SeekableReadStream, public SeekableWriteStream {
+private:
+	const uint32 _bufSize;
+	byte *_ptrOrig;
+	byte *_ptr;
+	uint32 _pos;
+	bool _err;
+	bool _eos;
+
+public:
+	MemorySeekableReadWriteStream(byte *buf, uint32 len) : _ptrOrig(buf), _ptr(buf), _bufSize(len), _pos(0), _err(false), _eos(false) {}
+
+	uint32 write(const void *dataPtr, uint32 dataSize) override {
+		// Write at most as many bytes as are still available...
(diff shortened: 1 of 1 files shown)


common/memstream.h
```diff
@@ -342,6 +342,96 @@ class MemoryReadWriteStream : public SeekableReadStream, public SeekableWriteStr
...
+			_pos += offset;
+			break;
+		}
+		// Post-Condition
+		if ((int32)_pos > size()) {
+			_pos = size();
+			_ptr = _ptrOrig + _pos;
+		}
+
+		return true;
```
criezy, line 431: Since a successful `seek` is supposed to reset the end-of-stream flag, I think you need to set `_eos = false;` before returning here.

common/memstream.h
```diff
@@ -342,6 +342,96 @@ class MemoryReadWriteStream : public SeekableReadStream, public SeekableWriteStr
...
+private:
+	const uint32 _bufSize;
+	byte *_ptrOrig;
+	byte *_ptr;
+	uint32 _pos;
+	bool _err;
+	bool _eos;
+
+public:
+	MemorySeekableReadWriteStream(byte *buf, uint32 len) : _ptrOrig(buf), _ptr(buf), _bufSize(len), _pos(0), _err(false), _eos(false) {}
```
criezy, line 358: Maybe for consistency you should have a `DisposeAfterUse::Flag` here as well as we do for other steam classes (and if it is `YES` free the `_ptrOrig` in the destructor)?

criezy (commented):
Adding this to common seems OK to me, but I made a couple of comments on the actual implementation.

common/memstream.h
mgerhardy: Nice catch - thanks a lot. Fixed now.

common/memstream.h
mgerhardy: Added the flag. Thanks

criezy (approved):

PR #2584, opened 2020-10-30
IPHONE: Fix building with recent cross-compiled SDK

lephilousophe:
This PR contains several little fixes to the Makefile based build system which are needed for my iPhone toolchain.
- defines are updated to match what's done in create_project for Xcode in ios7 platform
- When configure script don't find SDL but SDL_Net (when it is patched to not depend on SDL), don't try to link with SDL.
- Link against Security framework when linking with curl statically, in the case curl uses DarwinSSL.

These changes shouldn't affect existing buildbot toolchain except maybe the first one.

### DIFF
--- a/configure
+++ b/configure
@@ -3909,14 +3909,18 @@ EOF
 	if test "$_sdlnet" = yes ; then
 		set_var SDL_NET_LIBS   "$SDL2_NET_LIBS"
 		set_var SDL_NET_CFLAGS "$SDL2_NET_CFLAGS"
+		add_line_to_config_mk "SDL_NET_MAJOR = 2"
 	else
 		cat > $TMPC << EOF
 #include "SDL_net.h"
 int main(int argc, char *argv[]) { SDLNet_Init(); return 0; }
 EOF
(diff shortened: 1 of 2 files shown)


criezy (approved):
I always use Xcode to compile the iOS bundle and never tried to do it with the command line. But the changes here seem to make sense and are consistent with changes I needed to make to create_project.

The only change that could maybe be problematic is that this PR seems to also add `-framework Security` to the macOS build (and not just to the iOS build) when building with libcurl. However I don't think this is an issue. Currently we don't explicitly link with the Security framework for macOS, and it works, but maybe we actually should as libcurl has a dependency to it on macOS as well.

lephilousophe:
I suppose `-framework Security` isn't needed on MacOS today as we link against the dynamic libcurl present in SDK.
As it's dynamic, linking information to Security is in the library and we don't need to provide it.
On iPhone, SDK doesn't provide libcurl and I built a static version (current buildbot doesn't have any library).
I checked that Security framework is present since MacOS 10 and iOS 3 so it should be safe to link with it even if we don't need it. Linker will decide but shouldn't error because it's missing.

lephilousophe:
Merging now

PR #2585, opened 2020-10-31
COMMON: introduce substr for U32String

phcoder:
It's safer than direct array manipulation. Replace current uses
of subsetting with substr.


bluegr:
Quite clean and straightforward. Thumbs up from me! :)

PR #2587, opened 2020-11-01
EMI: Fix screenshot when saving on ES2

usineur:
Currently when saving on EMI (gles2 only), screenshot is not correctly
stored. When going in "Load game", screenshot is a black blank image.
This commit fixes this problem.

<details>
Before

emi1

After

emi2
</details>


engines/grim/gfx_opengl_shaders.cpp
```diff
@@ -2116,8 +2116,18 @@ static void readPixels(int x, int y, int width, int height, byte *buffer) {
 
 Bitmap *GfxOpenGLS::getScreenshot(int w, int h, bool useStored) {
 	Graphics::PixelBuffer src(Graphics::PixelFormat(4, 8, 8, 8, 8, 0, 8, 16, 24), _screenWidth * _screenHeight, DisposeAfterUse::YES);
-#ifndef USE_GLES2
 	if (useStored) {
+#ifdef USE_GLES2
```
aquadran, line 2120: this code does not use _storedDisplay, should not be placed here

aquadran: previous code should work on gles2, bugy driver?

aquadran:
this is rather w/a for some gles2 issue, can not be merged in this form

PR #2592, opened 2020-11-02
SLUDGE: Remove UTF8 code

phcoder:
This is basically a duplicate of UTF8 from common.
So just use common code instead

Tested with German Cubert


aquadran:
👍

PR #2595, opened 2020-11-03
ANDROID: Merge few files between android and android3d

phcoder:
While unfortunately it doesn't help with existing differences it help
to prevent their increase


antoniou79:
looks good to me.

PR #2596, opened 2020-11-04
SCUMM: COMI: Fix #4635/#10473 and #4391 (iMUSE Digital)

AndywinXp:
`SCUMM: COMI: Fix iMUSE Digital #4635/#10473 bugs`

This commit fixes the early stopping of ArriveBarber sequence, during the first arrival in Barbery Coast.
If this particular cutscene is skipped while this sequence is playing (e.g.: via ESCAPE key), the sequence is faded out
and the main Barbery Coast theme is played, as per correct behavior.
That's also the only COMI sequence using a transitionType with value 4, so the fix should not be able to break anything else, and in fact I haven't been able to break anything during my test run.

`SCUMM: COMI: Fix #4391 iMUSE Digital fades`

This commit implements fade-ins during song transitions (e.g. statePreVooOut --> statePreVooIn --> statePreVooLady) and also between regions of the same song (e.g. any JUMP instruction used for looping songs).

This results in a linear cross-fade. This is not ideal since the correct thing would be to implement a logarithmic crossfade (or "audio taper"), but for now this will do without changing too much of the current codebase. I'm still going to investigate the possibility of implementing this in a correct way.

I know, _I know_ the implementation looks terrible, a "real" fix would supposedly use fade tracks also for fade-ins, and require some kind of cloneToFadeInTrack(track, fadeDelay) to be used together with cloneToFadeOutTrack.

I didn't do this for two reasons:
- I think this would create some desync problems between fade tracks and proper tracks (for the life of me I just can't get fade out tracks to **always** be in perfect sync with the "parent" track).
- I would have to delaying putting the new song in the parent track until the correspondent fade in has occurred, and I don't know how to do it.

Having said that, this fix unfortunately does not account for the _**sync issue between fade-out and parent track**_, but it already sounds much better than the current implementation (even in the 30% of times in which the fade-outs happen not to be in sync). And besides, these bugs have remained unfixed for more than 10 years, so at this point I doubt anyone else has a plan to have a proper attempt at this anytime soon.
	
There might be another solution, for sure, but it's currently beyond my knowledge, I'm sorry 😢

### DIFF
--- a/engines/scumm/imuse_digi/dimuse.cpp
+++ b/engines/scumm/imuse_digi/dimuse.cpp
@@ -218,32 +218,67 @@ void IMuseDigital::callback() {
 				return;
 
 			if (track->volFadeUsed) {
-				if (track->vol == track->volFadeDest) // Sanity check
-					track->volFadeUsed = false;
+				if (_vm->_game.id == GID_CMI) {
+					if (track->vol == track->volFadeDest) // Sanity check
+						track->volFadeUsed = false;
 
-				if (track->volFadeStep < 0) { // Fade out
-					if (track->vol > track->volFadeDest) {
-						track->vol += track->volFadeStep;
-						if (track->vol <= track->volFadeDest) {
-							track->vol = track->volFadeDest;
-							track->volFadeUsed = false;
-						}
-						if (track->vol == 0) {
-							// Fade out complete -> remove this track
-							flushTrack(track);
-							continue;
+					if (track->volFadeStep < 0) { // Fade out
+						if (track->vol > track->volFadeDest) {
+							int tempVolume = transformVolumeEqualPowToLinear(track->vol, 1); // Equal power to linear...
+							tempVolume += track->volFadeStep; // Remove step...
+							track->vol = transformVolumeLinearToEqualPow(tempVolume, 1); // Linear to equal power...
+
+							if (track->vol <= track->volFadeDest) {
+								track->vol = track->volFadeDest;
+								track->volFadeUsed = false;
+								flushTrack(track);
+								continue;
+							}
+							if (track->vol == 0) {
+								// Fade out complete -> remove this track
+								flushTrack(track);
+								continue;
+							}
 						}
 					}
-				} else if (track->volFadeStep > 0) { // Fade in
-					if (track->vol < track->volFadeDest) {
-						track->vol += track->volFadeStep;
-						if (track->vol >= track->volFadeDest) {
-							track->vol = track->volFadeDest;
-							track->volFadeUsed = false;
+					else if (track->volFadeStep > 0) { // Fade in
+						if (track->vol < track->volFadeDest) {
+							int tempVolume = transformVolumeEqualPowToLinear(track->vol, 1); // Equal power to linear...
+							tempVolume += track->volFadeStep; // Add step...
+							track->vol = transformVolumeLinearToEqualPow(tempVolume, 1); // Linear to equal power...
+							if (track->vol >= track->volFadeDest) {
+								track->vol = track->volFadeDest;
+								track->volFadeUsed = false;
+							}
 						}
 					}
 				}
- 				debug(5, "Fade: sound(%d), Vol(%d)", track->soundId, track->vol / 1000);
+				else {
+					if (track->volFadeStep < 0) {
+						if (track->vol > track->volFadeDest) {
+							track->vol += track->volFadeStep; 
+							if (track->vol < track->volFadeDest) {
+								track->vol = track->volFadeDest;
+								track->volFadeUsed = false;
+							}
+							if (track->vol == 0) {
+								// Fade out complete -> remove this track
+								flushTrack(track);
+								continue;
+							}
+						}
+					}
+					else if (track->volFadeStep > 0) {
+						if (track->vol < track->volFadeDest) {
+							track->vol += track->volFadeStep;
+							if (track->vol > track->volFadeDest) {
+								track->vol = track->volFadeDest;
+								track->volFadeUsed = false;
+							}
+						}
+					}
+				}
+ 				debug(5, "Fade: sound(%d), Vol(%d) in track(%d)", track->soundId, track->vol / 1000, track->trackId);
 			}
 
 			if (!track->souStreamUsed) {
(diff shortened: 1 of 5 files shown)


aquadran:
some changes are risky, and may also break The Dig

AndywinXp:
> some changes are risky, and may also break The Dig

Thanks for the answer, I would love to improve the solution (also, I'm working on an alternate version of cloneToFadeOutTrack right know, I hope my idea works). 
Could you please elaborate on why these are risky, and what could break The Dig? Expecially about commit 769971d.

aquadran:
you are changing logic in fades, that may break The Dig

AndywinXp:
> you are changing logic in fades, that may break The Dig. I suggest to separate flow only for COMI

Thanks for the suggestion, I will correct my code accordingly here and also in the changes I haven't commited yet.

AndywinXp:
Changes introduced in the commit above (all exclusive to COMI):

- Implemented audio crossfading for transitions;
- Fixed music timing issues for both crossfades and JUMP instructions;
- Removed most of cloneToFadeOutTracks calls for the game, since they're not used in transition or in region switches (I couldn't get them to play on sync no matter what): as a result the loops in the music tracks are seamless now.

What's left to do?
- The loops and the region jumps are seamless and in sync, but in the original interpreter there is crossfading between the previous region and the new region. I couldn't get the crossfade to work properly by fading out the current track, flushing it afterwards, and starting a new track with the new region fading in: it kept spitting out exceptions about the region offset being either negative or bigger than the region itself;
- There is a strange music bug involving the end of the cannon minigame in the first part of the game: after destroying the last boat, the current track (stateCannon) jumps to a particular region of the song (by calling seqLastBoat) and then SEQ_NULL is immediately sent, effectively fading out and ending the song. After that, the game calls for stateHold2 with a hook for the JUMP with id 1. The intended effect is to start the song from the beginning (region 1 I guess?) and use the JUMP to skip to a farther region of the song. This almost works: the jump is done correctly but the song is started from the very beginning of the file (region 0, which is silent!), leaving everything silent until region 1 starts. This bug has been here for ages, if I recall correctly.

I'll be sure to send a ticket in the issue tracker about the last point, and I won't work both of these points for now, I'd just love to get these current fixes in the master, for now.

engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -295,8 +319,14 @@ void IMuseDigital::fadeOutMusic(int fadeDelay) {
 		Track *track = _track[l];
 		if (track->used && !track->toBeRemoved && (track->volGroupId == IMUSE_VOLGRP_MUSIC)) {
 			debug(5, "IMuseDigital::fadeOutMusic(fade:%d, sound:%d)", fadeDelay, track->soundId);
-			cloneToFadeOutTrack(track, fadeDelay);
-			flushTrack(track);
+			if (_vm->_game.id == GID_CMI) {
+				handleComiFadeOut(track, fadeDelay);
+				
+			}
+			else {
```
aquadran, line 326: formatting

engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -368,10 +407,100 @@ Track *IMuseDigital::cloneToFadeOutTrack(Track *track, int fadeDelay) {
...
+			linearValue = mappedValue * mappedValue;
+			break;
+		case 1:  // Quarter sine curve
+			linearValue = 0.63662 * asin(mappedValue);
+			break;
+		case 2:  // Parabola
+			linearValue = 1 - sqrt(1 - mappedValue);
+			break;
+		case 3:  // Logarithmic 1
+			linearValue = 0.00001*pow(M_E, 11.5129*mappedValue);
```
aquadran, line 482: formatting

engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -368,10 +407,100 @@ Track *IMuseDigital::cloneToFadeOutTrack(Track *track, int fadeDelay) {
...
+			break;
+		case 6:  // Half sine curve
+			eqPowValue = (1.0 - cos(mappedValue * M_PI)) / 2.0;
+		default: // Fallback to linear
+			eqPowValue = mappedValue;
+			break;
+		}
+
+		// Change range again, this time round the result
+		result = (((eqPowValue - 0.0)*(127000 - 0)) / (1.0 - 0.0)) + 0.0;
```
aquadran, line 453: formatting

engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -356,7 +395,7 @@ Track *IMuseDigital::cloneToFadeOutTrack(Track *track, int fadeDelay) {
 		// so gave up
 		error("Game not supported while playing on 2 different CDs");
 	}
-	track->soundDesc = soundDesc;
+	fadeTrack->soundDesc = soundDesc;
```
aquadran, line 398: looks like changing logic here

AndywinXp: Explaination for the change: 
in a previous row:
`ImuseDigiSndMgr::SoundDesc *soundDesc = _sound->cloneSound(track->soundDesc);`
we are cloning the soundDesc from the old track and storing it in *soundDesc.
I suppose this was done with the intention of assigning it to the fadeTrack (which is the clone currently being constructed).
Was the assignment to track->soundDesc in the old code intentional?

engines/scumm/imuse_digi/dimuse_track.cpp
AndywinXp: Whoops, will fix, thanks.

engines/scumm/imuse_digi/dimuse_track.cpp
AndywinXp: Whoops, will fix, thanks.

engines/scumm/imuse_digi/dimuse.cpp
```diff
@@ -218,29 +218,67 @@ void IMuseDigital::callback() {
...
+								track->volFadeUsed = false;
+							}
+							if (track->vol == 0) {
+								// Fade out complete -> remove this track
+								flushTrack(track);
+								continue;
+							}
+						}
+					}
+					else if (track->volFadeStep > 0) {
```
aquadran, line 271: formatting

engines/scumm/imuse_digi/dimuse_track.cpp
aquadran: yes, I think it was intentional

AndywinXp: Very well, I'll revert it then.

engines/scumm/imuse_digi/dimuse.cpp
```diff
@@ -218,29 +218,67 @@ void IMuseDigital::callback() {
...
+								flushTrack(track);
+								continue;
+							}
 						}
-						if (track->vol == 0) {
-							// Fade out complete -> remove this track
-							flushTrack(track);
-							continue;
+					}
+					else if (track->volFadeStep > 0) { // Fade in
```
aquadran, line 244: formatting

engines/scumm/imuse_digi/dimuse.cpp
AndywinXp: Corrected, sorry

engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -368,10 +406,100 @@ Track *IMuseDigital::cloneToFadeOutTrack(Track *track, int fadeDelay) {
...
+		return volume;
+
+	int result = volume;
+	if (!(volume < 0 || volume > 127000)) {
+		// Change range of values from 0-127*1000 to 0.0-1.0
+		double mappedValue = (((volume - 0) * (1.0 - 0.0)) / (127000 - 0)) + 0;
+		double linearValue;
+
+		switch (mode)
+		{
```
aquadran, line 470: formatting

engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -368,10 +406,100 @@ Track *IMuseDigital::cloneToFadeOutTrack(Track *track, int fadeDelay) {
...
+		return volume;
+
+	int result = volume;
+	if (!(volume < 0 || volume > 127000)) {
+		// Change range of values from 0-127*1000 to 0.0-1.0
+		double mappedValue = (((volume - 0) * (1.0 - 0.0)) / (127000 - 0)) + 0;
+		double eqPowValue;
+
+		switch (mode)
+		{
```
aquadran, line 425: formatting

PR #2598, opened 2020-11-04
GRIM: Fix alloc-dealloc mismatch in grim engine

mparnaudeau:
A previous fix in ResidualVM changed 2 calls to free() into delete
operators because scumm_strdup() was using the new operator.
Later scumm_strdup() switched to malloc() to be aligned with strdup()
function. So move back to free().


criezy:
It makes perfect sense. Thank you!

PR #2599, opened 2020-11-04
COMMON: Add conversion operator from UnicodeBiDiText to U32String

criezy:
Replace the `U32String` constructor that takes a `UnicodeBiDiText` by an implicit conversion operator in `UnicodeBiDiText`. This serves the same purpose of allowing to use a `UnicodeBiDiText` in places where a `U32String` is expected (such as in [ThemeEngine::drawDDText()](https://github.com/scummvm/scummvm/blob/606eeae46f9742e4947d5bea5e198ba3f822a1fe/gui/ThemeEngine.cpp#L1022) , but avoids making a copy (although we can still make a copy using the `U32String` copy constructor if that is what we want to do).

I am suggesting this change in a pull request in case C++ experts want to have a look and comment before this is merged, as I may have missed something.

### DIFF
--- a/common/unicode-bidi.h
+++ b/common/unicode-bidi.h
@@ -42,6 +42,11 @@ public:
 	UnicodeBiDiText(const Common::String &str, const Common::CodePage page);
 	~UnicodeBiDiText();
 
+	/**
+	 * Implicit conversion to U32String to get the visual representation.
+	 */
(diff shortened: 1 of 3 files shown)


criezy:
I merged the change manually after resolving the conflict.

PR #2603, opened 2020-11-05
KYRA: replace const char * with Common::String around getTableString.

phcoder:
Currently const char * points into some random buffer with unclear
lifetime.

Instead use Common::String which ensures lifetime as long as a reference is
held.

This also removes the need of copying into existing buffers in HoF and MR.

UnkBuf* are gone as well.


bluegr:
At a first glance, the changes look straightforward and the code is much cleaner. Nice work!

athrxx:
Yes, looks good. It seems I somehow forgot about this PR. I'll do some testing and then merge it (if I don't find any issues).

athrxx:
Currently, this commit breaks LOL, HOF and MR. The games crash immediately due to invalid memory reads. I haven't investigated it yet, probably some string object lifetime issue...

Also, there is a minor merge conflict from the latest Hebrew update. But that's easy to fix.

athrxx:
Okay, I think I fixed the merge conflict. I also spotted a minor regression in HOF which I can fix myself (the articles before the nouns dont't get properly cut off any more in updateCommandLineEx()). So I'll merge this now, since it is definitely a nice improvement. Thanks for your work!

PR #2606, opened 2020-11-06
SCUMM: Multi-font for Korean fan translated games

wonst719:

It adds the ability to load and use multiple external Korean fonts to match the original look.

On a separate note, I intend to refactor CJK text rendering to use Graphics::Font*.

Tested with:
- Loom (DOS/CD/Korean patched)
- Monkey Island 2 (DOS/Korean patched)
- Indy 4 (DOS/CD/Korean patched)
- DOTT (DOS/CD/Korean patched)
scummvm-monkey2-kr-00002
scummvm-monkey2-kr-00004

### DIFF
--- a/engines/scumm/charset.cpp
+++ b/engines/scumm/charset.cpp
@@ -42,12 +42,81 @@ The advantage will be cleaner coder (easier to debug, in particular), and a
 better separation of the various modules.
 */
 
+bool CharsetRenderer::isScummvmKorTarget() {
+	if (_vm->_language == Common::KO_KOR && (_vm->_game.version < 7 || _vm->_game.id == GID_FT)) {
+		return true;
+	}
+	return false;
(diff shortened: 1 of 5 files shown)


athrxx:
The new code seems to be well separated from the existing code. I don't see any issues.

Rendering through the CharsetRendererTownsV3 is a bit ugly. And half of the methods get redirected to CharsetRendererV3 anyway. And the actual MI2 FM-Towns version doesn't even use CharsetRendererTownsV3 (since it isn't Scumm3; it uses CharsetRendererClassic). But that's just nitpicking. We can still check after the merge whether it makes sense to come up with a new CharsetRenderer subclass...

sev-:
Thanks, all is nice and clean. I'll do the occasional formatting fixes in-tree.

PR #2607, opened 2020-11-06
SCUMM: Support variation of Korean postpositions

wonst719:

Korean text postpositions varies with the final consonant of the previous character.
e.g.
Book = 책
Ladder = 사다리
Use book = 책'을' 사용하기
Use ladder = 사다리'를' 사용하기

This patch introduces special "verb parser" to support this variation.
Only used in Monkey1 and Monkey2.

scummvm-monkey2-kr-00006

scummvm-monkey2-kr-00007

(망치'를' 향해 걸어가기 vs 달력'을' 향해 걸어가기)


athrxx:
And this one doesn't seem to be guarded against our existing code, e. g. you just check _language == Common::KO_KOR instead of using the new isScummvmKorTarget() method.
So, is there any way this could affect DIG or COMI? Or is it only used in code parts that these games will never reach (e. g. verb interface)?

We just recently had some fixes to make DIG and COMI games almost pixel-exact. That's why I am asking...

sev-:
Thank you!

wonst719:
Thanks.

> And this one doesn't seem to be guarded against our existing code, e. g. you just check _language == Common::KO_KOR instead of using the new isScummvmKorTarget() method.
> So, is there any way this could affect DIG or COMI? Or is it only used in code parts that these games will never reach (e. g. verb interface)?
> 
> We just recently had some fixes to make DIG and COMI games almost pixel-exact. That's why I am asking...

 I don't think this will affect DIG or COMI. It's only used in verb interface for MI1 and MI2. You're right that it should be guarded against existing code though. I will add the check soon.

PR #2608, opened 2020-11-06
SCUMM: Hacks for alternative line-break

wonst719:

This patch introduces alternative line-break style (similar to CSS word-break: break-all), and moves the hanging punctuation mark to the next line for cleaner text display.
This is used in Monkey2 (in the Phatt Island Library.)

scummvm-monkey2-kr-00009


athrxx:
It seems to be well guarded against the existing code. I don't exactly understand the purpose of the code (what is a 'hanging punctuation mark'?), but if this just about the layout of the Korean library index texts it shouldn't matter...

sev-:
Thanks, merging.

PR #2610, opened 2020-11-07
GRAPHICS: Fix no shader error on Raspberry PI

lmerckx:
On Raspberry PI, an error in OpenGLES shaders is not reported during the compilation step but during the link step.
See https://www.raspberrypi.org/forums/viewtopic.php?t=26205 for more information (last post).

This commit simply adds reporting of errors which occurred during the link step.


lmerckx:
Useful to debug some shaders which are not compatible with Raspberry.

PR #2615, opened 2020-11-10
COMMON: Remove resetGraphicsScale from OSystem

criezy:
This function used to be called from the Engine class, but was not longer used since commit 432fd52. The single remaining use was an internal use in the SurfaceSdlGraphicsManager, and I kept the behaviour there while removing the function.


aquadran:
👍

PR #2616, opened 2020-11-10
VITA: add support for booting game directly via bubble

cy33hc:
Add support for booting a pre configured game directly via a custom bubble similar to RetroArch on the PSVita.

@rsn8887 Here's the new PR with 1 single commit.


rsn8887:
Thank you, merged.

PR #2618, opened 2020-11-11
GRIM: Simplify shaders for Raspberry PI

lmerckx:
The shaders used by the Grim engine are not compatible with old or weak GPU (ex: Raspberry PI).

Here are the list of modifications done to support limitations:

- glDrawElements doesn't support GL_UNSIGNED_INT on GLES2. So, I replaced it by GL_UNSIGNED_SHORT and updated associated structures (Vector3int, _indexes, ...). This modification has been separated in another pull request: https://github.com/scummvm/scummvm/pull/2617
- Another limitation is the non-support of array of structures in uniforms. Raspberry's GPU supports arrays, structures but not array of structs. In that case, only the first element of the uniform is initialized. <br>It is the explanation of why Manny is so dark in Grim Fandango on the PI. To solve the problem, I splitted the array of Light in 4 distinct arrays in the shader.
- Finally, the last limitation, which causes an OutOfMemory and only on the PI (I guess) is the complexity of the shaders: conditional branching in shaders seem very badly implemented by the GPU.<br>So, the advice is to split a complex shader in multiple simpler ones (ex: one with light, one without, ...).<br>Following this recommendation, I split emi_actor.vertex in 3: one for actor with lights enabled, another for actors with lights disabled and the last one is for sprites (on billboard).<br>For GRIM: the two first shaders are enough.

Here are my sources for RPI limitations:
- For array of structures, the chapter "Why not an Array of Structures?" of this article made me doubt: http://pyopengl.sourceforge.net/context/tutorials/shader_7.html<br>It is for Python but when I debugged the call of glGetUniformLocation, it returned -1 for all uniforms of type "lights[%u]._whatever" except for index 0.
- And here is the post about the OutOfMemory and branching problem on PI (last post is the most important): https://www.raspberrypi.org/forums/viewtopic.php?t=19616

For information, I've tested these modifications on Raspberry PI 2, Raspberry PI 3 (both OpenGLES2), Raspberry PI 4, Odroid XU 4 (OpenGLES3), X86-64 (OpenGLES3, OpenGL without shaders and TinyGL) and it works for any platforms.

The same modification must be done for Stark engine but it's more complex because the shaders also reach the number of uniforms limitation. Perhaps also for Wintermute for light structure. There is no problem for Myst3.

This merge has initially been proposed on ResidualVM project.
See residualvm/residualvm#1519 for the corresponding issue and details.


engines/grim/gfx_opengl_shaders.cpp
```diff
@@ -676,48 +682,25 @@ void GfxOpenGLS::startActorDraw(const Actor *actor) {
...
-		}
-		_spriteProgram->setUniform("normalMatrix", normalMatrix);
-
-		_spriteProgram->setUniform("actorPos", pos);
-		_spriteProgram->setUniform("isBillboard", GL_FALSE);
-		_spriteProgram->setUniform("useVertexAlpha", GL_FALSE);
-		_spriteProgram->setUniform("uniformColor", color);
-		_spriteProgram->setUniform1f("alphaRef", 0.0f);
-		_spriteProgram->setUniform1f("meshAlpha", 1.0f);
+		for (int i=0; i < 3; i++) {
```
aquadran, line 685: formatting

engines/grim/gfx_opengl_shaders.cpp
```diff
@@ -660,6 +664,8 @@ void GfxOpenGLS::startActorDraw(const Actor *actor) {
 	Math::Matrix4 viewMatrix = _viewMatrix;
 	viewMatrix.transpose();
 
+	OpenGL::ShaderGL *shaders[] = {_spriteProgram, _actorProgram, _actorLightsProgram};
```
aquadran, line 667: formatting

engines/grim/gfx_opengl_shaders.cpp
```diff
@@ -728,16 +711,19 @@ void GfxOpenGLS::startActorDraw(const Actor *actor) {
...
-		_actorProgram->setUniform("viewMatrix", _viewMatrix);
-		_actorProgram->setUniform("projMatrix", _projMatrix);
-		_actorProgram->setUniform("extraMatrix", extraMatrix);
-		_actorProgram->setUniform("tex", 0);
-		_actorProgram->setUniform("texZBuf", 1);
-		_actorProgram->setUniform("hasZBuffer", hasZBuffer);
-		_actorProgram->setUniform("texcropZBuf", _zBufTexCrop);
-		_actorProgram->setUniform("screenSize", Math::Vector2d(_screenWidth, _screenHeight));
-		_actorProgram->setUniform1f("alphaRef", 0.5f);
+		for (int i=0; i < 3; i++) {
```
aquadran, line 714: formatting

engines/grim/gfx_opengl_shaders.cpp
```diff
@@ -752,43 +738,48 @@ void GfxOpenGLS::startActorDraw(const Actor *actor) {
 		if (!_currentShadowArray->dontNegate)
 			normal = -normal;
 
-		_actorProgram->setUniform("shadow._active", true);
-		_actorProgram->setUniform("shadow._color", color);
-		_actorProgram->setUniform("shadow._light", _currentShadowArray->pos);
-		_actorProgram->setUniform("shadow._point", shadowSector->getVertices()[0]);
-		_actorProgram->setUniform("shadow._normal", normal);
+		for (int i=0; i < 3; i++) {
```
aquadran, line 741: formatting

engines/grim/gfx_opengl_shaders.cpp
```diff
@@ -752,43 +738,48 @@ void GfxOpenGLS::startActorDraw(const Actor *actor) {
...
+			shaders[i]->setUniform("shadow._normal", normal);
+		}
 
 		glDepthMask(GL_FALSE);
 		glDisable(GL_BLEND);
 		glEnable(GL_POLYGON_OFFSET_FILL);
 	}
 	else {
-		_actorProgram->setUniform("shadow._active", false);
+		for (int i=0; i < 3; i++) {
```
aquadran, line 755: formatting

engines/grim/gfx_opengl_shaders.cpp
```diff
@@ -676,48 +682,25 @@ void GfxOpenGLS::startActorDraw(const Actor *actor) {
...
-			_spriteProgram->setUniform("projMatrix", _overworldProjMatrix);
-			_spriteProgram->setUniform("cameraPos", Math::Vector3d(0,0,0));
-		} else {
-			_spriteProgram->setUniform("viewMatrix", viewRot);
-			_spriteProgram->setUniform("projMatrix", _projMatrix);
-			_spriteProgram->setUniform("cameraPos", _currentPos);
-		}
-		_spriteProgram->setUniform("normalMatrix", normalMatrix);
-
-		_spriteProgram->setUniform("actorPos", pos);
```
aquadran, line 713: I see "actorPos" is not used

aquadran (changes requested):

aquadran:
please update dists/scummvm.rc

engines/grim/gfx_opengl_shaders.cpp
lmerckx: Was it blanks missing near the braces ?

engines/grim/gfx_opengl_shaders.cpp
lmerckx: Indeed, this uniform was set but never used.

PR #2619, opened 2020-11-14
CONFIGURE: Fix TTS detection with Mingw

lephilousophe:
This matches commit da77bfd7928f47438063e5e2d7ffffdc40e0c7d2 in real code.


sluicebox (approved):

sluicebox:
Good catch!

PR #2620, opened 2020-11-14
SCUMM: Implement language bundle for Korean fan translation

wonst719:
This adds support for language bundle to the SCUMM v1~v7 games.
It enables fan translation without the need for patching game resource.
While it's kind of similar to the DIG/COMI language bundle, the file format and the logic is different.

Tested with:
- Maniac Mansion (V1 DOS)
- Indy 3, 4 (Steam)
- Loom (EGA)
- Day of the Tentacle (CD)
- Sam & Max (CD)
- Full Throttle

### DIFF
--- a/engines/scumm/resource.cpp
+++ b/engines/scumm/resource.cpp
@@ -1116,8 +1116,7 @@ void ScummEngine::loadPtrToResource(ResType type, ResId idx, const byte *source)
 
 	// Translate resource text
 	byte translateBuffer[512];
-	if (isScummvmKorTarget())
-	{
+	if (isScummvmKorTarget()) {
 		if (!source) {
 			refreshScriptPointer();
 			source = _scriptPointer;
--- a/engines/scumm/resource.cpp
+++ b/engines/scumm/resource.cpp
@@ -1103,11 +1103,30 @@ void ScummEngine::loadPtrToResource(ResType type, ResId idx, const byte *source)
 	byte *alloced;
 	int len;
 
+	bool sourceWasNull = !source;
+	int originalLen;
+
 	_res->nukeResource(type, idx);
 
 	len = resStrLen(source) + 1;
 	if (len <= 0)
 		return;
+
+	originalLen = len;
+
+	// Translate resource text
+	byte translateBuffer[512];
+	if (isScummvmKorTarget())
+	{
+		if (!source) {
+			refreshScriptPointer();
+			source = _scriptPointer;
+		}
+		translateText(source, translateBuffer);
+
+		source = translateBuffer;
+		len = resStrLen(source) + 1;
+	}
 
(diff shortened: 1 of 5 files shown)


engines/scumm/resource.cpp
```diff
@@ -1103,20 +1103,43 @@ void ScummEngine::loadPtrToResource(ResType type, ResId idx, const byte *source)
...
 
 	len = resStrLen(source) + 1;
 	if (len <= 0)
 		return;
 
+	originalLen = len;
+
+	// Translate resource text
+	byte translateBuffer[512];
+	if (isScummvmKorTarget())
```
sev-, line 1119: Formatting

engines/scumm/string.cpp
```diff
@@ -1620,6 +1620,11 @@ static int indexCompare(const void *p1, const void *p2) {
 
 // Create an index of the language file.
 void ScummEngine_v7::loadLanguageBundle() {
+	if (isScummvmKorTarget()) {
+		// HACK to support language bundle for FT
```
sev-, line 1624: Why is it a hack?

engines/scumm/string.cpp
```diff
@@ -1901,7 +1911,204 @@ void ScummEngine_v7::translateText(const byte *text, byte *trans_buff) {
...
+	}
+	_languageBuffer = new byte[size - bodyPos];
+	file.read(_languageBuffer, size - bodyPos);
+	file.close();
+
+	debug(2, "loadLanguageBundle: Loaded %d entries", _numTranslatedLines);
+}
+
+// strcmp equivalent for Scumm string
+static int compareResString(const byte* a, int aLen, const byte* b, int bLen) {
```
sev-, line 1995: Formatting. Should be `const byte *a`, same for `b`, but in general, why not use 

```c++
strncmp(a, b, MIN(aLe, bLen))
```

Which looks like the full equivalent.

sev- (approved):
Besides a few minor formatting fixes and one potential improvement, looks fine to me.

engines/scumm/string.cpp
wonst719: I don't know what I was thinking. :-D
strncmp won't work, because SCUMM resource strings contain \0 middle of the string.
But memcmp(a, b, MIN(aLen, bLen)) will do.

engines/scumm/string.cpp
wonst719: I didn't feel good about making a exception path for just one game.
I can't think of a better way, though.

Leemann07:
Hi I'm Leemann. I find an English script for Beneath a Steel Sky because I want to do a fan translation for Koreans. Could someone please give me some advice? I want to output Korean in scummvm. :)

PR #2623, opened 2020-11-15
SCUMM: Add MD5 for Zak FM-TOWNS Japanese

BenCastricum:
SCUMM: Add MD5 for Zak FM-TOWNS Japanese

implements feature request #11485


raziel-:
Uuuh, neat, thank you BenCastricum

PR #2628, opened 2020-11-17
GUI: Retain scroll position when deleting a save

sluicebox:
Currently, deleting a save game resets the scroll position, scrolling the list to the very top. This isn't noticeable if there's not many saves, but once there's a lot it's a pain to delete save 72, get kicked back to the beginning, and then try to find where you were again. This happens to me frequently. It doesn't look like the current behavior is intentional and I think it's reasonable to expect the list to be stable when deleting.


criezy:
I agree that this is a better behaviour. Thanks for the PR!

PR #2633, opened 2020-11-21
DOXYGEN: Review of high prio headers from engines

b-gent:
This is a doxygen review for 4 header files that have been
identified as high priority for GSoD:

- advancedDetector.h
- metaengine.h
- engine.h
- util.h

It also adds doxy groups to other headers from this folder to
make them appear properly in the structure.

HTML output available here:
https://b-gent.github.io/2633/de/d3d/group__engines__advdetector.html

@sev-  @criezy

### DIFF
--- a/engines/advancedDetector.h
+++ b/engines/advancedDetector.h
@@ -34,14 +34,22 @@ namespace Common {
 class Error;
 class FSList;
 }
-
+/**
+ * @defgroup engines_advdetector Advanced Detector
+ * @ingroup engines
+ *
+ * @brief The Advanced Detector (AD) provides a standard framework for filename and MD5-based game detection.
+ *
+ * @{
+ */
+ 
 /**
  * A record describing a file to be matched for detecting a specific game
  * variant. A list of such records is used inside every ADGameDescription to
  * enable detection.
  */
 struct ADGameFileDescription {
-	const char *fileName; ///< Name of described file.
+	const char *fileName; ///< Name of the described file.
 	uint16 fileType;      ///< Optional. Not used during detection, only by engines.
 	const char *md5;      ///< MD5 of (the beginning of) the described file. Optional. Set to NULL to ignore.
 	int32 fileSize;       ///< Size of the described file. Set to -1 to ignore.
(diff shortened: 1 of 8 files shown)


b-gent:
I wasn't sure what to do with file obsolete.h , should I add doxygen groups to it? Its name suggests maybe we should exclude it from the doc build?

criezy:
> I wasn't sure what to do with file obsolete.h , should I add doxygen groups to it? Its name suggests maybe we should exclude it from the doc build?

I didn't even know that this file existed. After more than 10 years I am still discovering some new aspect of the code base :-)

A quick look indicate that the code in this file is not itself obsolete. It is to deal with obsolete game IDs (when we change the ID of a game between ScummVM versions) and handles update of the config file. So it should not be excluded from the doc.

engines/advancedDetector.h
```diff
@@ -291,45 +379,49 @@ class AdvancedMetaEngineDetection : public MetaEngineDetection {
...
 
 	/** Get the properties (size and MD5) of this file. */
 	bool getFileProperties(const FileMap &allFiles, const ADGameDescription &game, const Common::String fname, FileProperties &fileProps) const;
 
-	/** Convert an AD game description into the shared game description format */
+	/** Convert an AD game description into the shared game description format. */
 	virtual DetectedGame toDetectedGame(const ADDetectedGame &adGame) const;
 
-	friend class FileMapArchive; // for FileMap
+	/** @todo Doc required */
```
criezy, line 423: As discussed in the other PR, I don't think there is a need to document friends.

engines/metaengine.h
```diff
@@ -54,33 +54,51 @@ class GuiObject;
...
-	uint8 version;
-	Common::String saveName;
-	Common::String description;
-	uint32 date;
-	uint16 time;
-	uint32 playtime;
-	Graphics::Surface *thumbnail;
-	bool isAutosave;
+	char id[6];                   /*!< ID of the savegame file. */
+	uint8 version;                /*!< Version of the game. */
```
criezy, line 94: This is actually the version for the save game header and not the game version.

The idea here is that we may change the content of this header from time to time. For example the initial header we were using in old ScummVM versions did not have the `description` and `isAutoSave` information. When we read back those savegame files we need to know what header information is in the file so that we can read it properly. So the first thing that is read is the version, and then we know what else is there and read the rest of the header.

engines/metaengine.h
```diff
@@ -313,200 +343,226 @@ class MetaEngine : public PluginObject {
...
 	/**
 	 * Determine whether the engine supports the specified MetaEngine feature.
-	 * Used by e.g. the launcher to determine whether to enable the "Load" button.
+	 *
+	 * Used by e.g. the launcher to determine whether to enable the Load button.
 	 */
 	virtual bool hasFeature(MetaEngineFeature f) const;
 
+	/**
+	 * @todo Append extended save (?).
```
criezy, line 491: This writes the extended savegame header (see the `ExtendedSavegameHeader` struct) to the given savegame file.

engines/metaengine.h
```diff
@@ -313,200 +343,226 @@ class MetaEngine : public PluginObject {
...
+	/**
+	 * @todo Parse savegame header to retrieve its description.
+	 */
 	static void parseSavegameHeader(ExtendedSavegameHeader *header, SaveStateDescriptor *desc);
+	/**
+	 * @todo Create a dummy savegame header.
+	 */
 	static void fillDummyHeader(ExtendedSavegameHeader *header);
+	/**
+	 * @todo Doc required.
```
criezy, line 503: Read the extended savegame header from the given savegame file.

engines/metaengine.h
```diff
@@ -313,200 +343,226 @@ class MetaEngine : public PluginObject {
...
+	/**
+	 * @todo Append extended save (?).
+	 */
 	static void appendExtendedSave(Common::OutSaveFile *saveFile, uint32 playtime, Common::String desc, bool isAutosave);
+	/**
+	 * @todo Parse savegame header to retrieve its description.
+	 */
 	static void parseSavegameHeader(ExtendedSavegameHeader *header, SaveStateDescriptor *desc);
+	/**
+	 * @todo Create a dummy savegame header.
```
criezy, line 499: Populate the given extended savegame header with dummy values. This is used when failing to read the header from a savegame file.

engines/metaengine.h
```diff
@@ -313,200 +343,226 @@ class MetaEngine : public PluginObject {
...
+	 * Used by e.g. the launcher to determine whether to enable the Load button.
 	 */
 	virtual bool hasFeature(MetaEngineFeature f) const;
 
+	/**
+	 * @todo Append extended save (?).
+	 */
 	static void appendExtendedSave(Common::OutSaveFile *saveFile, uint32 playtime, Common::String desc, bool isAutosave);
+	/**
+	 * @todo Parse savegame header to retrieve its description.
```
criezy, line 495: Indeed. It parse the extended savegame header to retrieve the SaveStateDescriptor information (save date, save time, and total play time).

criezy (changes requested):
I have reviewed all the files except `engine.h` (this is a big one, and I am keeping it for tomorrow).

Overall this looks good, but there are a few mistakes.

engines/engine.h
```diff
@@ -52,42 +52,80 @@ class Dialog;
...
+ * @ingroup engines
+ *
+ * @brief API for managing various functions of the engine.
+ *
+ * @{
+ */
+
+/**
+ * Initialize graphics and show an error message.
+ * Redirects to GUIErrorMessage with U32Strings.
```
criezy, line 65: You can drop the `Redirects to GUIErrorMessage with U32Strings.` part.
This was a comment for the one that takes a `Common::String` only, to indicates that it converts that string to a `Common::U32String` and then calls this overload. So having this comment for the overload that takes a `Common::U32String` does not make sense. And even for the other ones for which it could make sense, I don't think we need to have it in the documentation. This is more a comment about the way it is implemented.

engines/engine.h
```diff
@@ -96,31 +134,58 @@ class PauseToken {
...
+	 * The SaveFileMAnager used by the engine.
+	 */
 	Common::SaveFileManager *_saveFileMan;
 
+	/**
+	 * The Dialog instance used by the engine.
+	 */
 	GUI::Dialog *_mainMenuDialog;
+	/**
+	 * Run the main menu dialog.
```
criezy, line 173: Run a GUI dialog. This can be the main menu dialog, but this can also be a different dialog.
To show the main menu dialog, you would not use this function though. You would call `openMainMenuDialog()` instead.

engines/engine.h
```diff
@@ -215,30 +291,37 @@ class Engine {
 	/** @name Overloadable methods
 	 *
 	 *  All Engine subclasses should consider overloading some or all of the following methods.
+	 *  @{
 	 */
-	//@{
 
+	/**
+	 * Initialize the OSystem for this engine.
```
criezy, line 298: No.
It creates an instance of the Engine. The engine will be using the given OSystem instance.

engines/engine.h
```diff
@@ -215,30 +291,37 @@ class Engine {
...
+	 */
 	Engine(OSystem *syst);
+	/**
+	 * @todo doc required
```
criezy, line 302: Destroy the Engine instance.

engines/engine.h
```diff
@@ -285,128 +373,152 @@ class Engine {
...
+	 *
+	 * Sends an EVENT_QUIT event to the Event Manager.
 	 */
 	static void quitGame();
 
 	/**
-	 * Return whether the ENGINE should quit respectively should return to the
-	 * launcher.
+	 * Return whether the engine should quit respectively should return to the
+	 * launcher. @todo Doc review, not sure what this is supposed to mean
```
criezy, line 486: Return whether the engine should quit (which means either quitting ScummVM altogether, or returning to the launcher). This should be used in games loops after pulling event to know if we need to exit the game loops.

engines/engine.h
```diff
@@ -446,61 +558,80 @@ class Engine {
...
 	 */
 	virtual int getAutosaveSlot() const {
 		return 0;
 	}
-
+	
+	/**
+	 * Check whether it is time to autosave based on the
+	 * provided @p lastSaveTime. @todo Doc check please
+	 */
```
criezy, line 616: This is correct but we should indicate that this function is deprecated.
Now autosaves are handled directly by the `Engine` class and derived classes do not need to worry about it other than to implement `canSaveAutosaveCurrently()` and `getAutosaveSlot()` if the default implementations are not sufficient.

criezy (changes requested):
I finished reviewing this pull request, and there are a few more changes that needs to be made before it can be merged.

engines/engine.h
b-gent: Just removing everywhere.

b-gent:
Comments fixed, rebased, and conflicts resolved.

Should be good to merge @criezy

b-gent:
Rebased.

b-gent:
One more amend, I realized I didn't add a doxy group to obsolete.h.

criezy (approved):
Indeed, it looks good now.
Thank you!

PR #2635, opened 2020-11-21
SCUMM: Fix pause message in Russian version of Atlantis and other games

phcoder:


criezy (approved):
This change seems simple enough and makes sense to me.

sev-:
Thanks, adding.

PR #2638, opened 2020-11-22
MORPHOS: update NEWS.md

BeWorld2018:
Add news fetures to MorphOS port

And remove include "common/encoding.h" in morphos-dialogs.cpp


criezy:
Thank you.

PR #2640, opened 2020-11-22
SCUMM: COMI: improve handling of iMUSE markers and triggers

AndywinXp:
These commits fix the root cause for bug [#11956](https://bugs.scummvm.org/ticket/11956) (even though that song now starts way too early, I need some help on figuring out why that happens, but at least the JUMP behavior is now correct), and implements handling for "_end" triggers found in transitions type 4 and 9.
As a result, this fixes the ending point for a couple of music tracks (seqArriveBarber and seqAttack), which played until the whole audio file had finished, instead of ending on the "_end" marker.

### DIFF
--- a/engines/scumm/imuse_digi/dimuse.cpp
+++ b/engines/scumm/imuse_digi/dimuse.cpp
@@ -414,20 +414,36 @@ void IMuseDigital::switchToNextRegion(Track *track) {
 	ImuseDigiSndMgr::SoundDesc *soundDesc = track->soundDesc;
 	if (_triggerUsed && track->soundDesc->numMarkers) {
 		if (_sound->checkForTriggerByRegionAndMarker(soundDesc, track->curRegion, _triggerParams.marker)) {
-			debug(5, "SwToNeReg(trackId:%d) - trigger %s reached", track->trackId, _triggerParams.marker);
-			debug(5, "SwToNeReg(trackId:%d) - exit current region %d", track->trackId, track->curRegion);
-			debug(5, "SwToNeReg(trackId:%d) - call cloneToFadeOutTrack(delay:%d)", track->trackId, _triggerParams.fadeOutDelay);
-			Track *fadeTrack = cloneToFadeOutTrack(track, _triggerParams.fadeOutDelay);
-			if (fadeTrack) {
-				fadeTrack->dataOffset = _sound->getRegionOffset(fadeTrack->soundDesc, fadeTrack->curRegion);
-				fadeTrack->regionOffset = 0;
-				debug(5, "SwToNeReg(trackId:%d)-sound(%d) select region %d, curHookId: %d", fadeTrack->trackId, fadeTrack->soundId, fadeTrack->curRegion, fadeTrack->curHookId);
-				fadeTrack->curHookId = 0;
+			if (_vm->_game.id != GID_CMI) {
+				debug(5, "SwToNeReg(trackId:%d) - trigger %s reached", track->trackId, _triggerParams.marker);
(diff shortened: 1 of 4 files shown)


engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -171,7 +171,7 @@ void IMuseDigital::startSound(int soundId, const char *soundName, int soundType,
 				track->regionOffset -= track->regionOffset >= (track->feedSize / _callbackFps) ? (track->feedSize / _callbackFps) : 0;
 			}
 		}
-		if (_vm->_game.id == GID_CMI) {
+		if (_vm->_game.id == GID_CMI && (track->volGroupId == IMUSE_VOLGRP_MUSIC)) {
```
AndywinXp, line 174: I don't know why this has been duplicated, it should't have. This is the same fix as commit 3d9a8862046acb84feb9bf6256c21b492cd67ff0

engines/scumm/imuse_digi/dimuse_track.cpp
```diff
@@ -441,6 +441,7 @@ int IMuseDigital::transformVolumeLinearToEqualPow(int volume, int mode) {
 			break;
 		case 6:  // Half sine curve
 			eqPowValue = (1.0 - cos(mappedValue * M_PI)) / 2.0;
+			break;
```
AndywinXp, line 444: I don't know why this has been duplicated, it should't have. This is the same fix as commit 3d9a8862046acb84feb9bf6256c21b492cd67ff0

AndywinXp:
I don't know why DeepCode is whining, since I've edited dimuse_music.cpp multiple times before and I don't think I ever got a strcmp warning on line 380, which has always been there.

engines/scumm/imuse_digi/dimuse_track.cpp
aquadran: you need work on separated branch

engines/scumm/imuse_digi/dimuse.cpp
```diff
@@ -414,20 +414,36 @@ void IMuseDigital::switchToNextRegion(Track *track) {
...
-			debug(5, "SwToNeReg(trackId:%d) - trigger %s reached", track->trackId, _triggerParams.marker);
-			debug(5, "SwToNeReg(trackId:%d) - exit current region %d", track->trackId, track->curRegion);
-			debug(5, "SwToNeReg(trackId:%d) - call cloneToFadeOutTrack(delay:%d)", track->trackId, _triggerParams.fadeOutDelay);
-			Track *fadeTrack = cloneToFadeOutTrack(track, _triggerParams.fadeOutDelay);
-			if (fadeTrack) {
-				fadeTrack->dataOffset = _sound->getRegionOffset(fadeTrack->soundDesc, fadeTrack->curRegion);
-				fadeTrack->regionOffset = 0;
-				debug(5, "SwToNeReg(trackId:%d)-sound(%d) select region %d, curHookId: %d", fadeTrack->trackId, fadeTrack->soundId, fadeTrack->curRegion, fadeTrack->curHookId);
-				fadeTrack->curHookId = 0;
+			if (_vm->_game.id != GID_CMI) {
```
aquadran, line 417: don't change non COMI flow

engines/scumm/imuse_digi/dimuse_track.cpp
AndywinXp: Whoops, sorry, you were right. I probably opened this branch concurrently to the other one you merged an hour ago.

PR #2642, opened 2020-11-23
GRIM: fix localization of dialogue transcript

BLooperZ:
Fixes bug where cutscenes lines were always written to dialogue transcript in english regardless of game language
as described in https://bugs.scummvm.org/ticket/11878

The change removes the second condition of checking the string to localize ends with `/`
sayLine do receive string formatted as `/{msgId}/`
but in cutscene they received as `/{msgId}/{original}`

as `parseMsgText` / `localize` only care about the first part `/{msgId}/` (rest is ignored)
it seems to work on both cases


aquadran:
thx!

PR #2643, opened 2020-11-24
AMIGAOS: Keep platform name continuity

raziel-:
This was bugging me for a long time now.

There are some files called amigaos4... and some called amigaos...
There is even one subdirectory simply called amiga.

To add and to keep some continuity to the platform name i have changed all instances of the platform to "amigaos".
Reason is for future development.
The only instance of amigaos4 remaining is with #ifdef (__ amigaos4 __), since that is how my platform is defined.
But that may change in the future, once the OS raise it´s version level it might also get a new define (or it may not, but anyway).

If there ever will be another amigaos(-like) platform (f.e. the classic 68k line) supported, then they can adapt their stuff just like MorphOS did.

- subdirectory renames
- file renames of physical files and path corrections inside the files
- some comments fixed/adapted

I hope i did't break anything else.

I had to do the renaming and all the other stuff from the github website, so it may look somewhat complicated (that's also why all the file changes have the same comment), please excuse


apolkosnik:
I apologize, but this patch seems to make a few things more confusing than they need to be.

AmigaOS is pretty much just running on Motorola 680x0 CPUs.

AmigaOS4 runs only on PowerPC machines that were created after the Commodore went under.

AmigaGuide is a hypertext format used on Amiga platform, and not "Amiga guide".

raziel-:
Reverted back to using AmigaGuide (though it was only in comments anyway)

Since ScummVM can't be built on a 68k AmigaOS(3) target (at least not with a lot of work, see MorphOS), i figured it would be easiest to just (re)use the (classic?) target name.

Of course, i could redo the patch and rename everything to amigaos4 instead, but what if the target name changes in the future?

What would be your take on this patch? amigaosppc, perhaps?
Since i don´t really know what parts of the code could be shared by a potential 68k target (and i don´t have the skills or hardware to try a port), i don´t know what i´d need to leave untouched.
And as such i´d propose to merge everything under the same target.
If ever some skilled developers comes around and tries to do a 68k port, then it should be possible to #ifdef in everything that´s needed (just with MorphOS).

It´s a mess right now and i don´t want to leave it at that

aquadran:
what is specific for amigaos4 should be called that. if it's common let just amigaos

engines/grim/gfx_opengl.cpp
```diff
@@ -1742,7 +1742,7 @@ void GfxOpenGL::prepareMovieFrame(Graphics::Surface *frame) {
 		format = GL_BGRA;
 		dataType = GL_UNSIGNED_INT_8_8_8_8;
 #else
-		// AmigaOS' MiniGL does not understand GL_UNSIGNED_INT_8_8_8_8 yet.
+		// MiniGL on AmigaOS doesn't understand GL_UNSIGNED_INT_8_8_8_8 yet.
```
aquadran, line 1745: this should mention amigos4

engines/grim/gfx_opengl.cpp
```diff
@@ -1752,7 +1752,7 @@ void GfxOpenGL::prepareMovieFrame(Graphics::Surface *frame) {
...
```
aquadran, line 1755: this should mention amigos4

backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
```diff
@@ -421,10 +421,10 @@ bool OpenGLSdlGraphics3dManager::createOrUpdateGLContext(uint gameWidth, uint ga
 		SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
 		SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
 #else
-		// AmigaOS4's OpenGL implementation is close to 1.3. Until that changes we need
+		// AmigaOS OpenGL implementation is close to 1.3. Until that changes we need
```
aquadran, line 424: this should mention amigos4

backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
```diff
@@ -421,10 +421,10 @@ bool OpenGLSdlGraphics3dManager::createOrUpdateGLContext(uint gameWidth, uint ga
...
 		// to use 1.3 as version or ScummVM will cease working at all on that platform.
 		// Profile Mask has to be 0 as well.
-		// This will be revised and removed once AmigaOS4 supports 2.x or OpenGLES2.
+		// This will be revised and removed once AmigaOS supports OpenGL 2.x or OpenGLES2.
```
aquadran, line 427: this should mention amigos4

configure
```diff
@@ -5434,7 +5434,7 @@ EOF
 	# Try different library names
 	if test "$_opengl_game_es2" = yes ; then
 		# 1) GLESv2    This is generally used by nearly all platforms for OpenGL ES 2
-		# 2) ogles2    This is used by AmigaOS4 (and probably others)
+		# 2) ogles2    This is used by AmigaOS (and probably others)
```
aquadran, line 5437: this should mention amigos4

sev-:
Yes, makes sense in our context.

PR #2645, opened 2020-11-25
DOXYGEN: Review of high prio files from graphics

b-gent:
Editing doxygen comments in:

- cursorman.h
- font.h
- managed_surface.h
- pixelformat.h
- surface.h

Still working on other files but pushing what I have now.

### DIFF
--- a/graphics/cursorman.h
+++ b/graphics/cursorman.h
@@ -31,19 +31,28 @@
 
 namespace Graphics {
 
+/**
+ * @defgroup graphics_cursorman Cursor manager
+ * @ingroup graphics
+ *
+ * @brief  The CursorManager class for managing the behavior of the mouse cursor.
+ *
+ * @{
+ */
+
 class CursorManager : public Common::Singleton<CursorManager> {
 public:
-	/** Query whether the mouse cursor is visible. */
+	/** Return whether the mouse cursor is visible. */
 	bool isVisible();
 
 	/**
 	 * Show or hide the mouse cursor.
 	 *
 	 * This function does not call OSystem::updateScreen, when visible is true.
-	 * This fact might result in a non visible mouse cursor if the caller does
-	 * not call OSystem::updateScreen itself after a showMouse(true) call.
+	 * You may need to call OSystem::updateScreen after a call to showMouse(true)
+	 * to ensure that the mouse cursor becomes visible.
 	 *
-	 * TODO: We might want to reconsider this behavior, it might be confusing
+	 * @todo We might want to reconsider this behavior, it might be confusing
 	 * for the user to call OSystem::updateScreen separately, on the other
 	 * hand OSystem::updateScreen might as well display unwanted changes on
 	 * the screen. Another alternative would be to let the backend worry
@@ -54,22 +63,24 @@ public:
 	bool showMouse(bool visible);
 
 	/**
-	 * Push a new cursor onto the stack, and set it in the backend. A local
-	 * copy will be made of the cursor data, so the original buffer can be
-	 * safely freed afterwards.
+	 * Push a new cursor onto the stack, and set it in the backend.
 	 *
-	 * @param buf		the new cursor data
-	 * @param w			the width
-	 * @param h			the height
-	 * @param hotspotX	the hotspot X coordinate
-	 * @param hotspotY	the hotspot Y coordinate
-	 * @param keycolor	the color value for the transparent color. This may not exceed
+	 * A local copy of the cursor data will be made, so the original buffer
+	 * can be safely freed afterwards.
+	 *
+	 * @param buf		New cursor data.
+	 * @param w			Width.
+	 * @param h			Height.
+	 * @param hotspotX	Hotspot X coordinate.
+	 * @param hotspotY	Hotspot Y coordinate.
+	 * @param keycolor	Color value for the transparent color. This cannot exceed
 	 *                  the maximum color value as defined by format.
-	 * @param dontScale	Whether the cursor should never be scaled. An exception are high ppi displays, where the cursor
-	 *                  would be too small to notice otherwise, these are allowed to scale the cursor anyway.
-	 * @param format	a pointer to the pixel format which the cursor graphic uses,
-	 *					CLUT8 will be used if this is NULL or not specified.
-	 * @note It is ok for the buffer to be a NULL pointer. It is sometimes
+	 * @param dontScale	Whether the cursor should never be scaled. An exception are high PPI displays, where the cursor
+	 *                  would be too small to notice otherwise. These are allowed to scale the cursor anyway.
+	 * @param format	Pointer to the pixel format that the cursor graphic uses.
+	 *					CLUT8 will be used if this is null or not specified.
+	 *
+	 * @note It is acceptable for the buffer to be a null pointer. It is sometimes
 	 *       useful to push a "dummy" cursor and modify it later. The
 	 *       cursor will be added to the stack, but not to the backend.
 	 */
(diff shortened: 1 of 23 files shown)


graphics/cursorman.h
```diff
@@ -31,19 +31,28 @@
...
 	bool isVisible();
 
 	/**
 	 * Show or hide the mouse cursor.
 	 *
 	 * This function does not call OSystem::updateScreen, when visible is true.
-	 * This fact might result in a non visible mouse cursor if the caller does
-	 * not call OSystem::updateScreen itself after a showMouse(true) call.
+	 * This might result in a non-visible mouse cursor if the caller does
+	 * not call OSystem::updateScreen after a showMouse(true) call.
```
criezy, line 53: Maybe rewrite that sentence to improve the flow.
I would for example suggest something like: `You need to call OSystem::updateScreen after a call to showMouse(true) to ensure that the mouse cursor becomes visible.`

graphics/font.h
```diff
@@ -112,91 +126,103 @@ class Font {
...
+	 * @param y     The y coordinate where to draw the character.
 	 * @param color The color of the character.
 	 */
 	virtual void drawChar(Surface *dst, uint32 chr, int x, int y, uint32 color) const = 0;
+	/** @overload */
 	void drawChar(ManagedSurface *dst, uint32 chr, int x, int y, uint32 color) const;
 
-	// TODO: Add doxygen comments to this
+	/**
+	 * @todo Doc required!
```
criezy, line 183: All the `drawString` functions below, as their name suggest, draw the given `str` string to the given `dst` surface.
The arguments are the same ones as in the `getBoundingBox` above, with a few additional ones:
- `dst` The surface on which to draw the string.
- `str`  The string to draw.
- `x`    The x position where to start drawing.
- `y`    The y position where to start drawing.
- `w`   The width of the text area.
- `color` The colour with which to draw the string.
- `align` The text alignment. This can be used to center the string in the given area or to align it to the right.
- `deltax` Offset to the x starting position of the string.
- `useEllipsis` Use ellipsis if needed to fit the string in the area.

criezy (changes requested):
There are a few minor corrections to make, and a few undocumented functions to document.
But otherwise this looks quite good already.

graphics/cursorman.h
```diff
@@ -120,59 +136,68 @@ class CursorManager : public Common::Singleton<CursorManager> {
...
 	void replaceCursorPalette(const byte *colors, uint start, uint num);
 
+	/**
+	 * Lock or unlock the mouse cursor. @todo Please verify
+	 */
 	void lock(bool locked);
 private:
+	/**
+	* Generic class for implementing the singleton design pattern.
+	*/
```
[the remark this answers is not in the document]
b-gent: Yes, and the description is taken from that file.

With the current config, doxygen shows friend classes in the output and I just want everything in the output to be documented.

image

 I'm not trying to say it's declared here, just providing the description of what the class is. I could also replace with just a reference to the doc of this class - "See @ref (...)"

Another question is, should we show friend classes in the doc output. Maybe it's completely unnecessary.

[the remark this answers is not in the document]
criezy: I think it is unnecessary to show it. This is an internal implementation details for the class. Developers using the CursorManager do not need to know that.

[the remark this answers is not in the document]
b-gent: Ok, I will make sure to disable showing friend classes in an upcoming PR.

b-gent:
Hi @criezy  , here are today's updates:

1. Added managed_surface.h, pixelformat.h, and surface.h to the initial commit. Pixelformat gave me a lot of headache and I had to do a ton of reading to understand the whole concept, but I was still very hesitant when adding comments there. Please review.
2. Added a second commit that adds doxy groups to the other headers in the graphics folder. I only did it for the ones that contain some doxygen. The descriptions for the groups do not matter much right now, I will be reviewing these files later anyway, so unless they are very wrong, you do not need to review them.
3. I uploaded HTML output to here:
https://b-gent.github.io/2645/d8/d40/group__graphics__conversion.html
The navi looks like this right now, which I think is pretty neat:
image

graphics/pixelformat.h
```diff
@@ -28,18 +28,27 @@
...
+ *
+ * @{
+ */
+
+/** Template to expand from an n-bit component to an 8-bit component. */
 template<int depth>
 struct ColorComponent {
 };
-
+/** @todo Doc required (0-bit element?) */
```
criezy, line 44: I am not quite sure why we have a specialisation for 0.  Maybe it is here for completeness or it might be needed if we have formats without alpha channel for example (although the code in `colorToARGB` seems to be checking for that explicitly to use 0xFF, which makes more sense than 0).

graphics/pixelformat.h
```diff
@@ -111,7 +120,7 @@ struct ColorComponent<7> {
 		       (value >> 6);
 	}
 };
-
+/** Return.... @todo not sure what that does if it's already 8-bit? */
```
criezy, line 123: It's just returning the given value (and making it doesn't use more than 8 bits).
I think it is mostly here so that 8 bits component do not have to be treated in a specific way and we can use the same code for all the cases.

graphics/pixelformat.h
```diff
@@ -139,12 +148,14 @@ struct PixelFormat {
 	byte rLoss, gLoss, bLoss, aLoss; /**< Precision loss of each color component. */
 	byte rShift, gShift, bShift, aShift; /**< Binary left shift of each color component in the pixel value. */
 
+	/** Define a pixel format ... @todo Doc required */
```
criezy, line 151: Default constructor that creates a NULL pixel format.

graphics/pixelformat.h
```diff
@@ -139,12 +148,14 @@ struct PixelFormat {
...
 	inline PixelFormat() {
 		bytesPerPixel =
 		rLoss = gLoss = bLoss = aLoss =
 		rShift = gShift = bShift = aShift = 0;
 	}
 
+	/** Define a pixel format ... @todo Is this RGB888? */
```
criezy, line 158: It is not necessarily RGB888. It depends on the value you pass to the constructor.
RGB888 would be `BytesPerPixel` = 3, `RBits` = `GBits` = `BBits` = 8, `ABits` = 0, `RShift` = 16, `GShift` = 8, `BShift` = 0, `AShift = 0

So basically `BytesPerPixel` is the total number of bytes used to define a colour, `RBits`, `GBits`, `BBits`, and `ABits` are the number of bits for each component, and `RShift`, `GShift`, `BShift`, and `AShift` is the amount of bit shift for the component in the colour value.

Some other examples:
- RGBA8888 is `BytesPerPixel` = 4, `RBits` = `GBits` = `BBits` = `ABits` = 8, `RShift` = 24, `GShift` = 16, `BShift` = 8, `AShift = 0
- ABGR8888 is `BytesPerPixel` = 4, `RBits` = `GBits` = `BBits` = `ABits` = 8, `RShift` = 0, `GShift` = 8, `BShift` = 16, `AShift = 24
- RGB565 is `BytesPerPixel` = 2, `RBits` = 5, `GBits` = 6, `BBits` = 5, `ABits` = 0, `RShift` = 11, `GShift` = 5, `BShift` = 0, `AShift = 0

graphics/pixelformat.h
```diff
@@ -196,64 +212,96 @@ struct PixelFormat {
...
 
+	/**
+	 * Return the number of alpha component bits.
+	 */
 	inline byte aBits() const {
 		return (8 - aLoss);
 	}
 
+	/**
+	 * Return the number of component bits from the specified bits per pixel (bpp) value.
```
criezy, line 264: Which specified `bpp` value? The function does not take any argument.
It returns the total number of bits for the pixel format.

graphics/surface.h
```diff
@@ -134,223 +145,234 @@ struct Surface {
...
+	 * @param color  The color of the frame.
 	 */
 	void frameRect(const Common::Rect &r, uint32 color);
 
-	// See comment in graphics/surface.cpp about it
+	/**
+	 * Move the surface horizontally or vertically.
+	 *
+	 * See @c surface.cpp for the implementation.
+	 * @todo More explanation required?
```
criezy, line 349: It would be more accurate to say that it moves the surface content horizontally and/or vertically by the given number of pixels. And I don't think you need more explanation than that.

graphics/managed_surface.h
```diff
@@ -30,143 +30,159 @@
...
-	 * Inner method for blitting
+	 * Inner method for blitting.
 	 */
 	void blitFromInner(const Surface &src, const Common::Rect &srcRect,
 		const Common::Point &destPos, const uint32 *palette);
 
 	/**
-	 * Inner method for copying another surface into this one at a given destination position
+	 * Inner method for copying another surface into this one at a given destination position.
+	 * @todo there are a lot of parameters here, are they all fairly clear or should we document them?
```
criezy, line 102: I don't think you need to document them. As the description indicates, it is an internal function for the class. Code using a `ManagedSurface` will not use this function directly.

graphics/managed_surface.h
```diff
@@ -30,143 +30,159 @@
...
-	 * because we don't want changes being done directly to it,
-	 * since it would bypass dirty rect handling
+	 * Automatically convert to a Graphics::Surface by
+	 * simply returning the inner surface.
+	 *
+	 * This must be const, because changes are not supposed to be done
+	 * directly to it, since it would bypass dirty rect handling.
 	 */
 	operator const Surface &() const { return _innerSurface; }
+	/** @todo not sure what that rawSurface is */
```
criezy, line 160: The `ManagedSurface` class is a wrapper around a `Surface` object. The `rawSurface` is simply this `Surface` object that is being wrapped.

criezy (changes requested):
There are only a few changes needed.
Overall it looks close to a state in which it can be merged.

graphics/cursorman.h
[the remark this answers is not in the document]
b-gent: Commit added here now.

graphics/pixelformat.h
b-gent: I coud use 'Return 0 for an empty color component.' ? Does that make sense?

graphics/managed_surface.h
b-gent: I copied the comment from the method above.

Question - is this technically an ``@overload`` if this method has a different name but does the same thing as the other method?

b-gent:
Comments are fixed now, branch is rebased. I left two small questions in the unresolved comments, please anwer whenever you have time @criezy

graphics/pixelformat.h
criezy: Yes, I think it does.

graphics/managed_surface.h
criezy: Technically an overloaded functions are functions with the same name but different arguments. So no, it is not an overload.
Also, as you can see, the method above doesn't actually have a name. This is something called  a `conversion operator` in C++.

criezy (approved):
It looks good to merge now. Thank you!

PR #2651, opened 2020-11-29
LURE: Fix MT-32 support

NMIError:
Lure of the Temptress has partial support for the Roland MT-32 in ScummVM, but 
it is unfinished and does not work properly. This PR finishes implementation of 
the MT-32 support.

This implements the following:
- SysExes for custom sounds
- Correct channel allocation
- Looping sound effects / music
- Correct sound effect / music volume
- Various small fixes to bring MIDI output in line with the original interpreter

See the commit messages for more details.

Notes:
- A new version of lure.dat must be generated with the updated create_lure tool 
before the game can be played.
- Lure of the Temptress makes extensive use of the extra sound effects 
introduced with the Roland CM-32L module, so you should use a CM-32L compatible 
module or ROM set when playing the game. It does not sound very good with an 
actual MT-32.
- AdLib support is also unfinished. Some of the changes made for MT-32 might 
also apply to AdLib, but I left this as-is for now. I do plan to fix this at a 
later time.


NMIError:
Side note: I have two more PRs ready for Lure of the Temptress. One fixes 
issues with the animation sequences in the game (mostly synchronization of the 
visuals with the music and sound effects), and the other fixes generic sound 
bugs (such as the missing door sounds and the shop clock). So keep in mind that 
there are still some audio issues with this PR.

ZvikaZ:
> Lure of the Temptress makes extensive use of the extra sound effects
introduced with the Roland CM-32L module, so you should use a CM-32L compatible
module or ROM set when playing the game. It does not sound very good with an
actual MT-32.

After this PR will be accepted, users won't be aware of this comment...
Is there a way to check what ROM is used, and warn if it's the wrong one?

If not, maybe it's worthy to issue such warning once per game, or something like that, so the user will do whatever necessary?

And anyway, it's great to see MT32 support being improved :-)

NMIError:
I think there was code in ScummVM at some point that would check the MT-32 ROM hash, but it could not be used for some reason... Also if a user is using a MIDI module, there's no way to check what's connected to the MIDI port (unless the user has connected an otherwise unnecessary second MIDI cable for messages from the module to the computer).

This issue is also not unique to Lure of the Temptress. There are other games which use the CM-32L sound effects (Beneath a Steel Sky, Leisure Suit Larry 5, several LucasArts games), though probably not as extensively as Lure. The reverse is also true, as there are several Sierra games which use sounds that are only correct on MT-32 v1.

So I think if we want to address this in ScummVM, a more generic solution for this would be good. We could specify in the game definition if it prefers a certain type of device, and the user should be able to specify the available sound options in more detail (what kind of device is hooked up to which MIDI port, or which ROM sets are available for use with the emulator). ScummVM can then pick the best device to use for a game.

criezy:
> > Lure of the Temptress makes extensive use of the extra sound effects
> > introduced with the Roland CM-32L module, so you should use a CM-32L compatible
> > module or ROM set when playing the game. It does not sound very good with an
> > actual MT-32.
> 
> After this PR will be accepted, users won't be aware of this comment...

We will at least want to document this somewhere. I think adding this info when we update https://www.scummvm.org/compatibility/DEV/lure/ would make sense, and we might want to add it to the README.md file as well (in the Supported Games section).

NMIError:
This is mentioned in the new documentation ( https://scummvm-documentation.readthedocs.io/en/latest/advanced_topics/understand_audio.html ). This also refers to a Wikipedia page with a list of MT-32 games and preferred devices ( https://en.wikipedia.org/wiki/List_of_MT-32-compatible_computer_games ). Maybe we could add a similar condensed list to the documentation that focuses on the games and issues relevant to ScummVM?

criezy (approved):
That's a great improvement!

I will leave this open a bit longer in case others want to review it, but if there is no objections my plan is to merge this next Sunday. We should also not forget to regenerate the `lure.dat` file (by the way, is there any reason why you did not include the updated file in this PR?).

NMIError:
I was under the impression that the CI build would generate the updated lure.dat. I've added it to the PR.

criezy:
Thank you. The engine data file are not regenerated automatically, so this is indeed something we have to do manually when needed.

Also would you like to join the team? This will give you the ability to commit directly some of your future work without going through pull requests.

NMIError:
Sure I'd like to join! I've got some plans for work on the sound support of various games and I'd like to get into some more general reverse engineering later as well. It's been fun and educational so far!

criezy:
We can ignore the travis failure. It is not related to this PR (and was probably fixed in dcdd9da5).
I will merge this now.

PR #2652, opened 2020-11-29
MOHAWK: Fix Myst regression (bug #11954)

eriktorbjorn:
This is my attempt at fixing https://bugs.scummvm.org/ticket/11954

Apparently, when reading a movie file name the engine needs to read an even number of bytes (probably to keep the rest of the data aligned), so if the length of the file name is even (not counting the terminating 0) it has to read an extra byte. This was already done, but apparently the String class used to allow you to add terminators to a string, and these were then counted into
the length of it. Not adding the terminator avoids this disambiguity, in case the behavior changes again. (I know it was changed to its current behavior to fix regressions in other engines.)

I've played through the non-Masterpiece version of Myst, getting both the red and blue pages and watching all the bad endings, without any crashes. I figured that this would be the less easily available version even though from what I remember they used to hand it out like candy. (I think my parents got one for free, and since they never played it they had no objections to me taking it.) But there is a limit to how much Myst I can take in one evening, so I would be grateful if someone else would test the Masterpiece version.

I had hoped to be able to find someone familiar with the engine to discuss this change with, but I failed. So if anyone has a better fix, please don't hesitate to chime in.

### DIFF
--- a/engines/mohawk/myst_areas.cpp
+++ b/engines/mohawk/myst_areas.cpp
@@ -192,10 +192,6 @@ MystAreaVideo::MystAreaVideo(MohawkEngine_Myst *vm, ResourceType type, Common::S
 		rlstStream->skip(1);
 	}
 
-	// Trim method does not remove extra trailing nulls
-	while (_videoFile.size() != 0 && _videoFile.lastChar() == 0)
-		_videoFile.deleteLastChar();
-
 	_videoFile = convertMystVideoName(_videoFile);
 	_videoFile = _vm->selectLocalizedMovieFilename(_videoFile);
 
--- a/engines/mohawk/myst_areas.cpp
+++ b/engines/mohawk/myst_areas.cpp
@@ -184,12 +184,13 @@ MystAreaVideo::MystAreaVideo(MohawkEngine_Myst *vm, ResourceType type, Common::S
 		MystAreaAction(vm, type, rlstStream, parent) {
 	char c = 0;
 
-	do {
-		c = rlstStream->readByte();
+	while ((c = rlstStream->readByte()) != 0) {
 		_videoFile += c;
-	} while (c);
+	}
 
-	rlstStream->skip(_videoFile.size() & 1);
+	if ((_videoFile.size() & 1) == 0) {
+		rlstStream->skip(1);
+	}
 
 	// Trim method does not remove extra trailing nulls
 	while (_videoFile.size() != 0 && _videoFile.lastChar() == 0)


eriktorbjorn:
I noticed there was some code to trim terminators from the string afterwards, which should no longer be necessary. I've played the game far enough to see it load both file names with odd and even lengths, but I have not played through the entire game again.

eriktorbjorn:
By the way, I vaguely remember hearing that the commit that caused this regression was meant to fix regressions in two game engines. Unfortunately, I only know that one of them was Myst III, where it would fail to find the subtitle font face because it was comparing "Arial Narrow" to "Arial Narrow\0".

It should be possible to keep Myst III from adding terminators to the strings. Perhaps something like this?

```
diff --git a/engines/myst3/archive.cpp b/engines/myst3/archive.cpp
index cbaaeb59d1..3eed080aac 100644
--- a/engines/myst3/archive.cpp
+++ b/engines/myst3/archive.cpp
@@ -286,17 +286,13 @@ Common::String ResourceDescription::getTextData(uint index) const {
        // extract the wanted one
        cnt = 0;
        int i = 0;
-       Common::String text;
-       while (cnt <= index && i < 89) {
-               if (cnt == index)
-                       text += decrypted[i];
-
+       while (cnt < index) {
                if (!decrypted[i])
                        cnt++;
-
                i++;
        }
 
+       Common::String text((const char *)&decrypted[i]);
        return text;
 }
```

But I don't know if that's correct. I haven't played Myst III in ages (no pun intended), and I've only checked that subtitles seem to work after that change, regardless of whether you can add extra string terminators or not.

dafioram:
This fix prevents the crash from happening in the beginning for myst me and myst 25. I haven't gone farther than that or tested myst3.

eriktorbjorn:
I've played through the Masterpiece version of Myst without any crashes. I don't own the 25th Anniversary version, so I can't test that one. Is it different from the older one, or just repackaged? The detection entry suggests it's multi-lingual. I've only played the English versions.

dafioram:
I believe myst 25 english is just a repackage of myst me with some menus added. The other myst 25 languages use different videos. I can test the intro at least for those. bgK would know for sure.

dafioram:
I tested myst 25 with the different languages on the intro and none of them crashes. I think this fix should be good.

eriktorbjorn:
Of course, this side-steps the issue whether or not it should be possible to add extra terminators to a String. Though in my opinion, well-behaved code shouldn't assume either behavior. (Maybe pick one behavior, and add a warning?)

engines/mohawk/myst_areas.cpp
```diff
@@ -184,16 +184,13 @@ MystAreaVideo::MystAreaVideo(MohawkEngine_Myst *vm, ResourceType type, Common::S
...
 		_videoFile += c;
-	} while (c);
-
-	rlstStream->skip(_videoFile.size() & 1);
+	}
 
-	// Trim method does not remove extra trailing nulls
-	while (_videoFile.size() != 0 && _videoFile.lastChar() == 0)
-		_videoFile.deleteLastChar();
+	if ((_videoFile.size() & 1) == 0) {
```
phcoder, line 191: This looks wrong. Shouldnt it be != 0 ?

criezy: That looks correct to me.

The way I understand that code, it looks like the stream always uses even number of bytes to store the string, so if we read an odd number of bytes, we need to read one my byte.

That's at least what the old code was doing since with the old `Common::String` code, adding the 0 would also increment the string size. So `_videoSize.size()` was giving us the number of bytes read.

With the new `Common::String` code, trying to add a 0 to the string does nothing (and in particular the size is not incremented). This is why this code broke. Now instead of checking is the size is odd, we need to check if it is even, because it does not count the additional terminating 0 (whether we try to add the 0 to the string or not). And this is what this new code is doing.

criezy:
The change looks good to me. At least it makes sense. And I think the change so that we are no longer relying on a specific behaviour for adding a `\0` to a string is a good move.

engines/mohawk/myst_areas.cpp
eriktorbjorn: That's the way I understood it too. My idea was to still read the same bytes from the file, but not to store the superfluous terminator in the resulting String. That way it doesn't matter whether or not String allows you to add extra terminators. But since that means _videoFile.size() is one less than it used to be, I had to invert the logic for when to skip a byte.

The idea behind the very untested Myst III patch above is also to not store the extra terminator. Myst III does work fine at the moment, but will break if the change to the String class is ever reverted.

I don't know where the Stark engine is impacted. I did manage to trigger the case where it adds 0 to the string by just waiting around in the dream at the beginning of the game. But I wasn't able to reproduce it when I tried running in a debugger.

phcoder: Then LGTM from me, feel free to merge

PR #2653, opened 2020-11-29
SCUMM: COMI: rebalance audio mix and implement music ducking during speech

AndywinXp:
This PR implements the volume reduction (about 4 dB, like in the original exe) of music tracks while speech audio is playing.
When a speech file starts playing, the music fades to a lower volume, and when the speech ends, the music immediately gets back to its original volume, just like in the original interpreter).
Additionally, the audio mix has been rebalanced to match with the one in original exe.


AndywinXp:
Even though this wasn't my original plan (I originally wanted to tamper with the SCUMM startup routine, but I managed to do what I wanted entirely in the iMUSE scope), I've decided to use the same PR for these two commits since these changes are complementary to each other.

AndywinXp:
_**Please do not merge yet!**_
I have just noticed panning is handled somewhat in the wrong way (we assume it goes from 0 (Left) to 127 (Right), but I'm fairly sure the numbers which come from the scumm scripts have a different range (I have encountered a setPan(150), which in the original intepreter sounds like dead center, but I have to check)

AndywinXp:
@aquadran The bug above is fixed (also, actor speech panning is implemented now), so the PR is mergeable if it looks good to you.

engines/scumm/imuse_digi/dimuse.cpp
```diff
@@ -203,6 +204,17 @@ void IMuseDigital::saveLoadEarly(Common::Serializer &s) {
 void IMuseDigital::callback() {
 	Common::StackLock lock(_mutex, "IMuseDigital::callback()");
 
+	_speechIsPlaying = false;
+	// Check for any track playing a speech line
+	if (_vm->_game.id == GID_CMI)
```
aquadran, line 209: missing bracket

engines/scumm/imuse_digi/dimuse.cpp
```diff
@@ -278,6 +290,29 @@ void IMuseDigital::callback() {
  				debug(5, "Fade: sound(%d), Vol(%d) in track(%d)", track->soundId, track->vol / 1000, track->trackId);
 			}
 
+			// Music gain reduction during speech
+			if (_vm->_game.id == GID_CMI && track->volGroupId == IMUSE_VOLGRP_MUSIC)
```
aquadran, line 294: missing bracket

engines/scumm/imuse_digi/dimuse_track.h
```diff
@@ -45,47 +45,53 @@ enum {
...
 	Audio::QueuingAudioStream *stream;		// sound mixer's audio stream handle for *.la1 and *.bun
+	Actor *speakingActor;					// actor reference for CMI speech
 
 	Track() : soundId(-1), used(false), stream(nullptr) {
 	}
 
 	void reset() {
 		trackId = 0;
-		pan = 0;
+		pan = 64;
```
aquadran, line 86: why changed to 64? it should be 0 as center

engines/scumm/imuse_digi/dimuse.cpp
AndywinXp: Thanks, solved

engines/scumm/imuse_digi/dimuse.cpp
AndywinXp: Thanks, solved

PR #2655, opened 2020-11-30
DOXYGEN: Review headers from 'audio'

b-gent:
This PR has 3 headers in scope of GSoD.
The files are:
- audiostream.h
- mixer.h
- timestamp.h

The rest of the files are only receiving a doxy group.

### DIFF
--- a/audio/audiostream.h
+++ b/audio/audiostream.h
@@ -37,6 +37,14 @@ class SeekableReadStream;
 namespace Audio {
 
 /**
+ * @defgroup audio_audiostream Audio streams
+ * @ingroup audio
+ *
+ * @brief API for managing audio input streams.
+ * @{
+ */
+
+/**
  * Generic audio input stream. Subclasses of this are used to feed arbitrary
  * sampled audio data into ScummVM's audio mixer.
  */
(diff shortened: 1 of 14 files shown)


b-gent:
@criezy 
Output is here
https://b-gent.github.io/2655/db/d71/group__audio__audiostream.html

Overall not a huge amount to check here, I think.

audio/audiostream.h
```diff
@@ -305,37 +331,45 @@ class SubSeekableAudioStream : public SeekableAudioStream {
 	Timestamp _pos;
 };
 
+/**
+ * A QueuingAudioStream class that allows for queuing audio streams for playback.
+ * @todo A bit more might be needed here
```
criezy, line 336: I think it is fine like this. Although maybe we can highlight its specificity a bit more by adding `multiple`:
`A QueuingAudioStream class that allows for queuing multiple audio streams for playback.`

audio/timestamp.h
```diff
@@ -28,128 +28,162 @@
...
+	 * This means that two timestamps can compare as equal
+	 * even if they use different frame rates.
 	 */
 	bool operator==(const Timestamp &ts) const;
+	/**
+	 * Check whether two timestamps describe a different moment in time.
+	 */
 	bool operator!=(const Timestamp &ts) const;
+	/**
+	 * Check whether this timestamp describes an earlier moment in time than another timestamp. @todo very unsure if I understand this correctly
```
criezy, line 123: Yes, that is correct.

criezy (changes requested):
This looks very good.
It is almost ready to merge.

b-gent:
Comments fixed.

criezy (approved):
That was quick! And it is now good to merge.

PR #2656, opened 2020-12-01
AMIGAOS: Platform updates

raziel-:
- Readability
- Changed and added comments
- Changed one variable name for comprehensibility
- Added no builtin resources as default to save as much binary size as possible
- Tried to apply to the coding standards (and probably failed miserably)
- Re-activate the formerly locked-out nuked OPL Adlib driver
- Reflect changes in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -80,6 +80,10 @@ For a more comprehensive changelog of the latest experimental code, see:
    - Fixed occasional border corruption during fights.
    - Improvements to cutscenes to better match the original games.
 
+ AmigaOS port:
+   - Add native system file browser feature
+   - Re-activate nuked OPL Adlib driver
+
  MorphOS port:
    - Add native system file browser feature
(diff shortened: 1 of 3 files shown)


raziel-:
Forgot to mention:
Reactivate the nuked OPL driver (was locked out from AmigaOS for unknown reasons back in the day.
After returning it to the compilation, and testing with some SCI MIDI games, it shows no bad behaviour...actually, it sounds quite good [It doesn't beat real hardware, of course, but it's still better than most other MIDI alternatives on my platform] and it's fast too. It beats the MT-32 emulator by far speed-wise.)

lotharsm:
Thanks a lot! The changes in `configure` look very plausible to me. I can only vaguely comment on the changes in `amigaos-main.cpp` since it's quite platform specific, but at least I couldn't spot anything obvious :-)

raziel-:
Thank you

PR #2657, opened 2020-12-01
ENGINES: Add detection for international releases in many engines

trembyle:
I have several more to add, but I wanted to submit these in case detection code changes and they become obsolete.

Korean Return to Zork is only dubbed. There are no fonts to render, even in the menus. Unlike the Japanese version, this version should work without further changes.

KYRA.DAT still needs to be updated.

### DIFF
--- a/engines/made/detection_tables.h
+++ b/engines/made/detection_tables.h
@@ -143,7 +143,11 @@ static const MadeGameDescription gameDescriptions[] = {
 		{
 			"rtz",
 			"V1.2, 9/29/94, CD",
-			AD_ENTRY1s("rtzcd.red", "946997d8b0aa6cb4e848bad02a1fc3d2", 276584),
+			{
+				{ "rtzcd.red", 0, "946997d8b0aa6cb4e848bad02a1fc3d2", 276584 },
+				{ "rtzcd.prj", 0, "974d74410c3c29d50e857863e8bf40e2", 43016792 },
+				AD_LISTEND
+			},
 			Common::EN_ANY,
 			Common::kPlatformDOS,
 			ADGF_CD,
@@ -253,6 +257,29 @@ static const MadeGameDescription gameDescriptions[] = {
 			"V1.2, 3/31/95, CD",
 			AD_ENTRY1s("rtzcd.red", "946997d8b0aa6cb4e848bad02a1fc3d2", 354614),
 			Common::FR_FRA,
+			Common::kPlatformDOS,
+			ADGF_CD,
+			GUIO0()
+		},
+		GID_RTZ,
+		0,
+		GF_CD_COMPRESSED,
+		3,
+	},
+
+	{
+		// Return to Zork - Korean CD version 1.2 9/29/94
+		// Dub only. No text was translated, even in menus, so there are no font issues.
+		// submitted by trembyle
+		{
+			"rtz",
+			"V1.2, 9/29/94, CD",
+			{
+				{ "rtzcd.red", 0, "946997d8b0aa6cb4e848bad02a1fc3d2", 276584 },
+				{ "rtzcd.prj", 0, "3c8644f7ce77b74968637c035c3532d8", 48083511 },
+				AD_LISTEND
+			},
+			Common::KO_KOR,
 			Common::kPlatformDOS,
 			ADGF_CD,
 			GUIO0()
--- a/engines/zvision/detection_tables.h
+++ b/engines/zvision/detection_tables.h
@@ -104,7 +104,11 @@ static const ZVisionGameDescription gameDescriptions[] = {
 		{
 			"znemesis",
 			0,
-			AD_ENTRY1s("CSCR.ZFS", "88226e51a205d2e50c67a5237f3bd5f2", 2397741),
+			{
+				{ "CSCR.ZFS", 0, "88226e51a205d2e50c67a5237f3bd5f2", 2397741 },
+				{ "ASCR.ZFS", 0, "9a1e1a48a56cf12a22bad2d2e47f6c92", 917452 },
+				AD_LISTEND
+			},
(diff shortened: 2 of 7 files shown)


athrxx:
The "KYRA: Detect Chinese and Korean variants" seems not to do what it advertises. It doesn't actually include detection entries. If you still want to submit the detection entries please mark them with an ADGF_UNSUPPORTED flag. Otherwise users would only run into the "You are missing the kyra.dat(...)" error.

Regarding your changes to the kyra.dat tool I see no advantage in having such code fragments added, unless they actually do something. So, to give you a rough idea what the necessary steps are for adding a new language to the kyradat tool I dug up some commits from history:

Spanish for HOF:
https://github.com/scummvm/scummvm/commit/3bc54abd79b5278b3d9bb90467c7810ec52acd10
Spanish for LOK:
https://github.com/scummvm/scummvm/commit/5b660fe8aece110843ba8b170fd1e1cb38c5fa5e
Japanese for EOB1/SegaCD:
https://github.com/scummvm/scummvm/commit/6bcad5fdf387474905a2a08eb66d9dbcd9208790

HOF would probably be the easiest to do, so maybe practice on that... Or just limit this commit to the actual detection until someone else adds the support.

trembyle:
@athrxx I pushed the wrong commit for KYRA. I will fix this.

Also, in general, I will check whether all of these games run and add the ADGF_UNSUPPORTED flag as necessary.

trembyle:
Rebased and added detection entries for the KYRA games, along with unsupported flags and messages. I added flags and messages for the other engines where necessary.

Portuguese Riven and Korean Return to Zork both appear to be stable. Korean Zork Nemesis may be close to stable, but is currently flagged unsupported.

I will raise bug reports for specific issues where appropriate. And I may continue to work on KYRA.DAT; otherwise, if I can't figure that out, I'll raise a bug for that too.

engines/zvision/detection_tables.h
```diff
@@ -104,7 +104,11 @@ static const ZVisionGameDescription gameDescriptions[] = {
 		{
 			"znemesis",
 			0,
-			AD_ENTRY1s("CSCR.ZFS", "88226e51a205d2e50c67a5237f3bd5f2", 2397741),
+			{
+				{ "CSCR.ZFS", 0, "88226e51a205d2e50c67a5237f3bd5f2", 2397741 },
+				{ "ASCR.ZFS", 0, "9a1e1a48a56cf12a22bad2d2e47f6c92", 917452 },
```
trembyle, line 109: `CSCR.ZFS` is identical between English and Korean.

Added `ASCR.ZFS` to differentiate.

engines/made/detection_tables.h
```diff
@@ -143,7 +143,11 @@ static const MadeGameDescription gameDescriptions[] = {
 		{
 			"rtz",
 			"V1.2, 9/29/94, CD",
-			AD_ENTRY1s("rtzcd.red", "946997d8b0aa6cb4e848bad02a1fc3d2", 276584),
+			{
+				{ "rtzcd.red", 0, "946997d8b0aa6cb4e848bad02a1fc3d2", 276584 },
+				{ "rtzcd.prj", 0, "974d74410c3c29d50e857863e8bf40e2", 43016792 },
```
trembyle, line 148: `rtzcd.red` is identical between English and Korean. This is likely because the text was untranslated in this release. I added `rtzcd.prj` to differentiate.

trembyle:
After testing, I reverted Zork Nemesis back to stable.

I found only one minor bug that I raised as [#11992](https://bugs.scummvm.org/ticket/11992).

athrxx:
For the KYRA part, this now looks okay

athrxx:
I have taken a look at that Chinese EOB II version. This isn't an "international release". It is a heavily modified fan project (even including Chinese voice acting). They have hacked up the character generator and the ingame interface in a rather crude way to accomodate the Chinese font, also using weird gui colors.
While I am willing to support "normal" fan translations with minor modifications this is a bit too much. So I still don't mind the detection entry with the ADGF_UNSUPPORTED flag. But IMHO that should be the final stage of this version in ScummVM.
I'll check the LOK and HOF versions whether these are more eligible...

dreammaster:
Xeen and Sherlock detections look fine

trembyle:
Discussed this with @athrxx offline. The game was definitely released in Chinese.

There is a boxed version for sale [here](http://book.kongfz.com/276039/2378721958/) (only US$865!). On the front of the box, top left, the characters say 中文版 or "Chinese Version". This is also on the green sticker on the side of the box.

I've attached the images below in case this listing disappears.

Now the remaining question is whether this official version is indeed the version I'm adding, or if I've found some cheap fan-made hack instead. I'll do some more research to see if I can track down different disks.

eob2-zh-tw-front
eob2-zh-tw-side
eob2-zh-tw-disks
eob2-zh-tw-back

trembyle:
Looking through a stack of disks, every version I can find contains the text `MODIFY BY SHELL STONE V 1.0` below the main menu when starting the game.

It's also present on any gameplay video I've seen, for example:
https://www.youtube.com/watch?v=LsyjUhfhK1w
https://www.bilibili.com/video/av2060146/

I realize that this isn't definitive. The people recording the videos might have obtained copies from the same ultimate source.

As a counterpoint to the fan translation explanation, two other possibilities come to mind.

1. The officially licensed version does indeed contain this text. Hopefully this line wasn't written by the translator, due to the poor grammar.
2. This is more of a copy protection hack than a translation. If this was commonly available, then it could be on a lot of users' disks as well, and therefore should be supported.

sev-:
Thank you!

sev-:
But it needs to be rebased....

sev-:
Thanks, merging!

PR #2658, opened 2020-12-01
GUI: Don't leak theme and TextDrawData when loading fails

lephilousophe:
This was spotted with a FreeType disabled build and with Address Sanitizer

### DIFF
--- a/gui/ThemeEngine.cpp
+++ b/gui/ThemeEngine.cpp
@@ -577,6 +577,11 @@ bool ThemeEngine::addFont(TextData textId, const Common::String &language, const
 					ttsMan->setLanguage("en");
 #endif // USE_TTS
 #endif // USE_TRANSLATION
+
+				// No font, cleanup TextDrawData
+				delete _texts[textId];
+				_texts[textId] = nullptr;
+
 				return false; // fall-back attempt failed
 			}
 			// Success in fall-back attempt to standard (non-localized) font.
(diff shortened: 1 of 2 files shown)


gui/ThemeEngine.cpp
```diff
@@ -592,6 +597,10 @@ bool ThemeEngine::addFont(TextData textId, const Common::String &language, const
 #endif // USE_TRANSLATION
 			// Returning true here, would allow falling back to standard fonts for the missing ones,
 			// but that leads to "garbage" glyphs being displayed on screen for non-Latin languages
+			// No font, cleanup TextDrawData
```
criezy, line 600: It did manage to load a font in this case. So I am not sure we want to cleanup the TextDrawData.
Although as the comment above says, `If we return false anyway why would we attempt the fall-back in the first place?` I am asking myself the same question, but I have no answer for that.

lephilousophe: In fact, when we return false here, `ThemeParser` will return an error and `loadTheme` will consider theme as not loaded (`themeOk` is false).
This means that when the ThemeEngine will get destroyed, it won't delete `_texts` items.
Does this mean, we should relax the constraint when unloading theme and cleanup every arrays even if theme is not loaded?
That sounds like more logical than my fix but I didn't know the implications.

criezy: In any case the change here should do no harm. We can do some cleanup at a later time.

criezy:
Merging this now.

PR #2660, opened 2020-12-02
DOXYGEN: Fix the build output and structure

b-gent:
This commit introduces a couple tweaks to the
doxygen build configuration to assure the
overall structure looks correct.

@criezy @Mataniko @sev-


criezy (approved):
Thank you.
All the changes make sense here.

PR #2663, opened 2020-12-03
MorphOS: active nuked OPL Adlib driver

BeWorld2018:
Thanks Raziel to see that on AmigaOS platform :-)


raziel-:
👍  :-)

criezy:
This is a very localised change that only impact the MorphOS port, and I assume you have tested it.
So I will merge this now.

PR #2668, opened 2020-12-06
SCUMM: Support for uncompressed bundles like used in some fan transla…

phcoder:
…tions

Apparently it's a beta variant of storing files in the bundle without
compression and in little-endian. I tried to find an indicator for endianness
in uncompressed iMuse file but apparently there isn't any. Hence we need to
propagate a bool from bundle to imuse layer.

This is used by Akella Russification of COMI

### DIFF
--- a/engines/scumm/imuse_digi/dimuse_bndmgr.cpp
+++ b/engines/scumm/imuse_digi/dimuse_bndmgr.cpp
@@ -193,6 +193,7 @@ bool BundleMgr::open(const char *filename, bool &compressed, bool errorFlag) {
 	_indexTable = _cache->getIndexTable(slot);
 	assert(_bundleTable);
 	_compTableLoaded = false;
+	_isUncompressed = false;
 	_outputSize = 0;
 	_lastBlock = -1;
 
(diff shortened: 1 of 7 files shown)


engines/scumm/imuse_digi/dimuse_bndmgr.cpp
```diff
@@ -250,7 +258,8 @@ int32 BundleMgr::decompressSampleByCurIndex(int32 offset, int32 size, byte **com
 	return decompressSampleByIndex(_curSampleId, offset, size, compFinal, headerSize, headerOutside);
 }
 
-int32 BundleMgr::decompressSampleByIndex(int32 index, int32 offset, int32 size, byte **compFinal, int headerSize, bool headerOutside) {
+int32 BundleMgr::decompressSampleByIndex(int32 index, int32 offset, int32 size, byte **compFinal, int headerSize, bool headerOutside,
+					 bool *uncompressedBundle) {
```
aquadran, line 262: use reference not a pointer

sev-: There is a default value, `nullptr`

phcoder: Changed

PR #2670, opened 2020-12-07
SCUMM: Support Lokalizator plugin.

phcoder:
This was used by Russian translation of Pajama Sam 2 and Spy Fox 3

### DIFF
--- a/engines/scumm/he/localizer.cpp
+++ b/engines/scumm/he/localizer.cpp
@@ -1,0 +1,122 @@
+/* ScummVM - Graphic Adventure Engine
+ *
+ * ScummVM is the legal property of its developers, whose names
+ * are too numerous to list here. Please refer to the COPYRIGHT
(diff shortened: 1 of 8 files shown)


engines/scumm/localizer.cpp
```diff
@@ -0,0 +1,122 @@
...
+				break;
+			case 1:
+				x = st + 35;
+				break;
+			case 2:
+				x = st + 70;
+				break;
+			case 3:
+				x = st + 105;
+				st += 45707404;
```
henke37, line 83: This looks like a subtraction being incorrectly rendered as an add.

sev-: Not at all. this is 45mln, not 4 bln. we're talking here about 1/100th of that, 0x2b9708c doesn't look like a negative 32-bit number to me.

sev-:
Thanks!

PR #2672, opened 2020-12-08
SCUMM: COMI: implement iMUSE crossfades between regions

AndywinXp:
This was a tough one, I also lost hours of work after a wrong file overwrite, but I managed to finish it and test it.
Apart from some bug fixes to track volume ranges, I have implemented the last missing feature for iMUSE Digital (for COMI): crossfades between regions triggered by JUMP instructions.

As a result, we have seamless loops, and graceful jumps to other sections of the same track.
This applies both to music and SFX.

Every known jump or music loop has been tested, one by one, by hand, at least 3 times each, and these modifications appears to be stable (worst case scenario, when stressing ScummVM, a crossfade gets played a little bit late).
It is curious that some music tracks have the loop point shifted behind by -8820 (samples? units? still don't understand what's that number), so this is addressed via a modification of the COMI music table.

There appear to be a special case for the song in part 3 (the crossfades stay out of sync no matter what), I'll try to tackle that in the next hours so please _do not merge yet_, but I still want to get the PR here since I'm scared to lose work... again.

EDIT: it's now mergeable


AndywinXp:
Very well, I'm done fixing the song (even though two of those crossfades are still best guesses, but it's the best I could do without giving uneven offset shifts to the track), so the commit is now mergeable

aquadran:
I'm not sure about this PR. we have too many hacks right now

AndywinXp:
I understand. 
The only thing I can say is that this would give a 1:1 match to the original exe, in terms of result produced, concluding the restoration work I started a month ago (therefore this would be my final PR).
I also am convinced that there are "on the fly" corrections to loop points in the original, since removing crossfades (and just using the current jump statement) still result in desynchronized loops for the tracks which I have shifted.

I understand hacks are bad, and I managed removed a couple of those very old ones myself, but in my defense, these modifications manage to solve some very noticeable (11 years long) bugs which nobody else had time (or will) to solve. And that's a pity, for a game which is currently being sold in bundle with ScummVM.

Please think about it some more time and let me know which your final decision is.

AndywinXp:
I made a little bit better, now there's no prior game specific knowledge embedded in runSchduledCrossfades

PR #2673, opened 2020-12-08
SCI: Add City's fan Russian translation support

phcoder:
This PR has 2 commits. One to tolerate a slightly invalid image and another adds the hashes

### DIFF
--- a/engines/sci/detection_tables.h
+++ b/engines/sci/detection_tables.h
@@ -5699,6 +5699,15 @@ static const struct ADGameDescription SciGameDescriptions[] = {
 	AD_LISTEND },
 	Common::RU_RUS, Common::kPlatformWindows, ADGF_DROPPLATFORM, GUIO_TORIN },
 
+	// Torin's Passage - Russian Windows CD (Dyadushka Research + City)
+	// VERSION file "1.0"
+	{ "torin", "",{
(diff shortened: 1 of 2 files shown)


sluicebox:
Which specific resources have problems? Where can I find this fan translation?

I'd prefer that we use the resource patcher (resource_patcher.cpp) on the broken images instead of allowing the uncompressed version of getRow() to ignore invalid requests. That keeps game-specific workarounds out of the engine code, and I think that is even more important for fan stuff. That way we also have built-in documentation of which image/game/version is the problem. Right now this begs the question "why does the uncompressed reader allow this this but the compressed one doesn't?"

Separately, this is a good example of code that shouldn't be an assertion, since it's validating external data. I think it should be an error() if not a warning(), and should be the same in all the getRow() methods, but that can happen later.

bluegr:
I agree with @sluicebox here. Since there are broken resources in this version, we should either try and patch them, or ignore specific errors for specific resources, not all of them.

bluegr (changes requested):
Overall, the corrupt graphics should be identified and corrected, instead of silently dropping the errors. Perhaps it is possible to identify all the offending resources with external tools, e.g. SCI Companion, or SV

bluegr:
Excellent, thanks for addressing the issues! Merging

bluegr (approved):

PR #2675, opened 2020-12-09
HADESCH: Add subtitles for AIF files

phcoder:

### DIFF
--- a/engines/hadesch/gfx_context.cpp
+++ b/engines/hadesch/gfx_context.cpp
@@ -26,6 +26,9 @@
 #include "hadesch/baptr.h"
 #include "common/system.h"
 #include "graphics/palette.h"
+#include "graphics/font.h"
+#include "graphics/fontman.h"
+#include "hadesch/hadesch.h"
 
 namespace Hadesch {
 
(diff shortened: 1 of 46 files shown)


engines/hadesch/gfx_context.cpp
```diff
@@ -81,27 +84,63 @@ void GfxContext8Bit::fade(int val) {
...
 	if (_palette) {
 		g_system->getPaletteManager()->setPalette(_palette, 0, 256);
 	}
 
-	g_system->copyRectToScreen(_pixels.get() + viewPoint.x + _pitch * viewPoint.y, _w, 0, 0,
+	g_system->copyRectToScreen(surf.getBasePtr(viewPoint.x, viewPoint.y), surf.w, 0, 0,
 				   kVideoWidth, kVideoHeight);
 }
 
+byte GfxContext8Bit::findColor(byte r, byte g, byte b) {
```
sev-, line 106: I think we could use this opportunity to introduce something like `Graphics::Palettemanager` which will do this palette search.

I have a very similar code here:
  https://github.com/scummvm/scummvm/blob/master/graphics/macgui/macwindowmanager.cpp#L968

It is out of scope of this PR, though.

sev- (approved):
Very cool feature. Merging.

PR #2679, opened 2020-12-12
KYRA: adlib driver fixes

miller-alex:
I'm fixing some issues with the Westwood ADL driver in AdPlug, which
is based on ScummVM's Kyra AdLib driver. I think these fixes should
go upstream, so I'm submitting them here first.

Most of the work will be about adding safety checks such that even
manipulated sound files can't crash the driver. I'm fixing other issues
along the way when I encounter them, as in the first few commits.
Later, I may also try to name some of the remaining unknowns and
simplify some code.

Don't hesitate to comment if you have suggestions.

### DIFF
--- a/engines/kyra/sound/drivers/adlib.cpp
+++ b/engines/kyra/sound/drivers/adlib.cpp
@@ -520,6 +520,10 @@ void AdLibDriver::stopAllChannels() {
 			noteOff(chan);
 	}
 	_retrySounds = false;
+
+	_programQueueStart = _programQueueEnd = 0;
+	_programQueue[0] = QueueEntry();
+	_programStartTimeout = 0;
 }
 
 // timer callback
--- a/engines/kyra/sound/drivers/adlib.cpp
+++ b/engines/kyra/sound/drivers/adlib.cpp
@@ -574,12 +574,15 @@ void AdLibDriver::setupPrograms() {
 	++_programQueueStart &= 15;
 
 	const int chan = *ptr++;
+	// Safety check: ignore request for invalid channel
+	if (chan > 9)
+		return;
+	Channel &channel = _channels[chan];
+
 	const int priority = *ptr++;
 
 	// Only start this sound if its priority is higher than the one
 	// already playing.
-
-	Channel &channel = _channels[chan];
 
 	if (priority >= channel.priority) {
 		initChannel(channel);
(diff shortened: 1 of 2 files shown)


athrxx:
I admit that I am reluctant to admit pull requests on this code from outside the dev team,  since debugging this isn't too much fun and I really don't want regressions here. But it turns out that your code changes are straight-forward enough, even though they seem to address bugs not in ScummVM but in some other project(s).

Currently, I can't spot any regressions either. From my tests I thought you might have caused a regression from the queue handling changes in the Brandon Zanthia scene where she is stirring around in her cauldron. The queue will be overflowing and not recovering  when you use the flute repeatedly, so sound effects are basically blocked for that scene (haven't really investigated it yet). But it turns out I have the exact same issue without your changes :-/ So that would not be a blocker for this PR. Just something that I'll keep in mind and try to fix later. Unfortunately simply sticking to the original drivers (which aren't too complicated) will not probably not help, since these are quite buggy and prone to race conditon effects.

miller-alex:
Well, I try not to break things.

The queue handling changes are actually irrelevant for AdPlug, but since there was an obvious bug I fixed it anyway. It may not make a big difference for ScummVM either since there was a workaround, but it's cleaner with the fix. If it turns out it causes regressions, you can drop it.

Adding security checks is primarily for AdPlug. Even though that may not be necessary for ScummVM (what's your threat model?) it should improve robustness there, too. (And of course keeping the code bases in sync makes it easier to share future improvements.)

I think I've fixed the obvious security issues, so I could stop here (unless I find more issues, testing a bit now). If you don't want me to mess around with that code, I'll leave it alone. Or I could try to improve naming and simplify code and add the commits here or in a new PR. It's your choice, so what do you think?

athrxx:
I wouldn't mind having wrongly named or unknown variables renamed and some of the more useless commments removed. I just kept most of this in the state the original author left it. "Simplifying" the code could be a good thing depending on what you mean by that. I don't want the overall design diverge too much from that of the original game drivers, since it would make debugging even less enjoyable. Currently our functions more or less match the original. Keeping that in mind, there are certainly all kinds of possible improvements...
I don't have any preference regarding continueing this PR or making another. This one here is still in draft mode. If you want it merged, please just change the status and tell me that it is ready.

miller-alex:
I haven't spotted any more problematic uses of unchecked data, so let's say it's ready.

These fixes don't need to wait for other improvements, so I'll open a new PR when I have more changes.

athrxx:
Thanks! Merging.

PR #2680, opened 2020-12-12
BUILD: Fix parallel "clean all" builds

FrostyHitoshura:
I'm trying to have successful builds with the "make -j 6 clean all" command and before this change, the failure rate was 100%.

This change fixes the problem that the clean target was run partially at the start of the build and the rest just before the linking of the final executable, which was the reason of the 100% failure rate. The change prevents the redefinition of the "clean" target by hooking on the existing one instead.

Unfortunately, it's not good enough to entirely fix the problem, the failure rate is now about 50%. While the "clean" target runs entirely at the start, it seems that "all" can still start before "clean" is fully completed. I attempted to serialize the "clean" and "all" targets using the ".ORDER" directive and also order only prerequisites with no success.

I'm looking for help to resolve this part.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -133,7 +133,8 @@ endif
 distclean: clean clean-devtools
 	$(RM) config.h config.mk config.log engines/engines.mk engines/plugins_table.h
 
-clean:
+clean: clean-toplevel
+clean-toplevel:
 	$(RM_REC) $(DEPDIRS)
 	$(RM) $(OBJS) $(DETECT_OBJS) $(EXECUTABLE)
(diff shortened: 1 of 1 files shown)


sev-:
Looks like a good improvement to me.

FrostyHitoshura:
I didn't have the time to spend more time on the problem but if you run:

```
make clean
make -j 6 all
```
it works and is more than acceptable for my work flow. Even if my commit doesn't fully resolve the problem, I agree it's a good improvement so I'm going to put it in review.

sev-:
Thanks

PR #2687, opened 2020-12-16
SCI: add a shortcut alias for the registers debug console command

deckarep:
When debugging, I'm constantly inspecting the SCI VM registers and thought this would be a great command to have an alias for.

I am proposing the following alias: `reg` for the command `registers` and also updated the help docs.


bluegr:
Sounds great, thanks!
Merging

PR #2688, opened 2020-12-17
GUI: added debug output for console prints

mgerhardy:
This is useful for situations where the game crashes or is quit, but you still need to
output of the debug commands for investigating the issue


criezy:
This makes sense to me and is simple enough. Merging.

PR #2689, opened 2020-12-17
MAKE: allow to override the quite flags without re-executing configure

mgerhardy:
this allows you to activate the flags for a single build without re-building everything
(due to regenerated config.h)


lephilousophe:
Did you try to pass `VERBOSE_BUILD=yes` to make command line?
I believe it will do what you want.

mgerhardy:
Thanks - indeed - I was just blind.

PR #2690, opened 2020-12-17
GUI: Reflect default Gfx/Stretch Mode immediately

av-dx:
Fix for issue #11744, with similar fix for updating graphics mode as well.

Selecting "\<default\>" for "Graphics Mode" and "Stretch Mode" and clicking "Apply" under Options...->Graphics now immediately reflects the changes in ScummVM window.

### DIFF
--- a/gui/options.cpp
+++ b/gui/options.cpp
@@ -549,8 +549,11 @@ void OptionsDialog::apply() {
 					gm++;
 				}
 			}
-			if (!isSet)
+			if (!isSet) {
 				ConfMan.removeKey("gfx_mode", _domain);
+				if (g_system->getGraphicsMode() != g_system->getDefaultGraphicsMode())
+					graphicsModeChanged = true;
+			}
 
 			if ((int32)_renderModePopUp->getSelectedTag() >= 0)
 				ConfMan.set("render_mode", Common::getRenderModeCode((Common::RenderMode)_renderModePopUp->getSelectedTag()), _domain);
@@ -569,8 +572,11 @@ void OptionsDialog::apply() {
 					sm++;
 				}
 			}
-			if (!isSet)
+			if (!isSet) {
 				ConfMan.removeKey("stretch_mode", _domain);
+				if (g_system->getStretchMode() != g_system->getDefaultStretchMode())
+					graphicsModeChanged = true;
+			}				
 
 			if (_rendererTypePopUp->getSelectedTag() > 0) {
 				Graphics::RendererType selected = (Graphics::RendererType) _rendererTypePopUp->getSelectedTag();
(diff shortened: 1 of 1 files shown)


lotharsm:
Could you please fix the commit message according to our standards, so it reads `GUI: Refresh Gfx/Strech mode when...`?

Besides that, looking good to me.

av-dx:
Yes, I have fixed the commit message now.

gui/options.cpp
```diff
@@ -628,9 +634,8 @@ void OptionsDialog::apply() {
 	if (_domain == Common::ConfigManager::kApplicationDomain && graphicsModeChanged) {
 		g_system->beginGFXTransaction();
 		g_system->setGraphicsMode(ConfMan.get("gfx_mode", _domain).c_str());
-
-		if (ConfMan.hasKey("stretch_mode"))
-			g_system->setStretchMode(ConfMan.get("stretch_mode", _domain).c_str());
+		g_system->setStretchMode(ConfMan.get("stretch_mode", _domain).c_str());
```
bluegr, line 637: At this point, it's possible that the stretch_mode setting is removed in line 576, and graphicsModeChanged is true, so this if block gets executed. So, removing the hasKey() check is not correct

bluegr (changes requested):

gui/options.cpp
av-dx: When the option value is set to "default", the key gets deleted from file. If we are only refreshing when hasKey() is true, then a refresh will not happen when "default" is selected, which is the issue. 
As of now, ConfMan.get() returns "default" string in case the key doesn't exist, which is a case accounted for by both setGraphicsMode() and setStretchMode() internally (I noticed it is not handled by setFeatureState(), so hasKey() is required there). 
Should it be changed so that : if hasKey() is false then "default" string is explicitly passed to setGraphicsMode() and setStretchMode()? Or are there other solutions?

criezy: Removing the check indeed should work as:
1. We register a default value when starting ScummVM (https://github.com/scummvm/scummvm/blob/master/base/commandLine.cpp#L239)
2. `hasKey()` doesn't check the default, so indeed returns false if the setting was removed and only the default remains. But `get()` does check the default and will return that value.

And you may notice that the `hasKey()` check is not present for the "gfx_mode", and this is fine for the same reason outlined above.

I will however take a bit more time later today to check properly the changes in this pull request. I suspect for example that the same issue exists for the shaders setting (which we currently only have on PS Vita, so it is more difficult to test).

bluegr: I've checked hasKey() and get(). Indeed, the code above should work fine. Resolving this, thanks!

bluegr (approved):

criezy (approved):
The changes are straightforward enough and make sense.
Thank you for this pull request, I will merge this now.

PR #2691, opened 2020-12-18
KYRA: Name most unknowns in AdLib driver

miller-alex:
Name all unknown variables and callback methods except a few that don't do anything useful.

I'm not very good at naming things, so I don't mind if somebody comes up with further improvements, but the code should be easier to understand than before.

Also includes two small checks I missed for the last PR.


athrxx:
Thanks! It's basically the same as with the last PR. It doesn't fix anything, but certainly makes the code more enjoyable. I haven't spotted anything that looks like it might cause a regression. And I think you even fixed some opcodes that are never used, at least not by any game we support.

PR #2692, opened 2020-12-19
STARTREK: Specify that action is signed char

phcoder:
x86 uses signed chars but this is architecture-dependent and on arm it's
unsigned and leads to compilation failure


bluegr:
Thanks! Merging

PR #2694, opened 2020-12-20
STARTREK: Fix compilation for all platforms

lephilousophe:
Use an int8 everywhere for the Action type

This completes PR #2692 and take @sev- suggestion into account.


bluegr:
Thanks, I've already made this change in e46e7e022b1fc30f2ee44b433240b48c6f355b8e. Closing this

PR #2695, opened 2020-12-20
HADESCH: Mark the engine as C++11 dependent

lephilousophe:
Aggregate initialization for classes with constructor requires C++11


lephilousophe:
This does fix build on my test toolchains.
Applying.

PR #2698, opened 2020-12-25
SCUMM: FT: improve and fix iMUSE Digital

AndywinXp:
This PR consists in accuracy fixes to iMUSE Digital for Full Throttle with respect to the original DOS interpreter.
We have:

- A fix in `switchToNextRegion(Track *track)` (my fault, it was caused by a regression introduced by COMI fixes);
- A substantial refactoring in dimuse.cpp (which removes a lot of code duplication, since COMI and FT share lots of traits for music handling);
- Added fade-ins (they occur when there was another music track already running, otherwise the new music track is played without a fade-in); this implementation also prevents a bug found in the original EXE (i.e.: try getting in and out of the bar at the beginning of the game, the music will not fade properly and instead is replaced by silence)
- Fixed bug [#6857](https://bugs.scummvm.org/ticket/6857) by correctly implementing an edge case behavior
- Fine tuned the overall audio mix to be consistent with the original EXE;
- (Even though I've tested the whole game flawlessly, I want to ask you, aquadran, if this can be done, or if there's a better way to do it) **Changed _callbackFps from default 10 to 20 (doubled it) for this game only**; this is needed for two reasons:

1. The original game iMUSE callbacks seems to operate at the same rate (and in fact, a hearing comparison between the EXE and ScummVM shows that the crossfades are played in the same way, with this change)
2. With only 10 calls to the callback every second, the crossfades are not smooth, since the volume steps are too far apart from each other

Of course, every common change has been tested with all three games involved (FT and COMI for the improvements, and DIG for the "please don't crash or change your original behavior").


engines/scumm/imuse_digi/dimuse.cpp
```diff
@@ -52,6 +52,8 @@ IMuseDigital::IMuseDigital(ScummEngine_v7 *scumm, Audio::Mixer *mixer, int fps)
 	_sound = new ImuseDigiSndMgr(_vm);
 	assert(_sound);
 	_callbackFps = fps;
+	if (_vm->_game.id == GID_FT)
+		_callbackFps *= 2;
```
AndywinXp, line 56: This is the change I'm referring to in the PR description

mgerhardy:
Could you please rebase this to remove the merge commits? If you pull the remote repository, please do `git pull --rebase` or set `pull.rebase` to `true` with `set git config --global pull.rebase true`.

AndywinXp:
> Could you please rebase this to remove the merge commits? If you pull the remote repository, please do `git pull --rebase` or set `pull.rebase` to `true` with `git config --global pull.rebase true`.

Sure, sorry for any confusion and thanks for the heads up, I will rebase as soon as I get home.

AndywinXp:
Uhh.. Yeah I'll just... start over again with this one

PR #2701, opened 2020-12-25
SCUMM: FT: improve and fix iMUSE Digital

AndywinXp:
...second time's the charm, I guess 😓 

This PR consists in accuracy fixes to iMUSE Digital for Full Throttle with respect to the original DOS interpreter.
We have:

- A fix in `switchToNextRegion(Track *track)` (my fault, it was caused by a regression introduced by COMI fixes);
- A substantial refactoring in dimuse.cpp (which removes a lot of code duplication, since COMI and FT share lots of traits for music handling);
- Added fade-ins (they occur when there was another music track already running, otherwise the new music track is played without a fade-in); this implementation also prevents a bug found in the original EXE (i.e.: try getting in and out of the bar at the beginning of the game, the music will not fade properly and instead is replaced by silence)
- Fixed bug [#6857](https://bugs.scummvm.org/ticket/6857) by correctly implementing an edge case behavior
- Fine tuned the overall audio mix to be consistent with the original EXE;
- **Changed _callbackFps from default 10 to 20 (doubled it) for this game only**;

Of course, every common change has been tested with all three games involved (FT and COMI for the improvements, and DIG for the "please don't crash or change your original behavior").

**EDIT, Feb 17 2021**: after some weeks of decompiling and researching here's what the original interpreter does in the main callback:
- The timer is operated at a base rate of 50Hz, which means it's being called every 20000 microseconds;
- This 20000 value is manually added to a counter in every iteration of the callback;
- This counter is being used to have some of the things operating at two different rates (10Hz and curiously, 60Hz);
- Usual audio stuff, like playing sound from tracks and switching streams (which also means: crossfades), happens at 50Hz rate, which is the base rate;
- Triggers and fades handling happens at a (somewhat hacky) 60Hz rate;
- Music gain reduction caused by speech happens at a 10Hz rate.

Of course I'm working towards doing a proper accurate overhaul of the current system, but for now this produces a better and acceptable result than before.

### DIFF
--- a/engines/scumm/imuse_digi/dimuse.cpp
+++ b/engines/scumm/imuse_digi/dimuse.cpp
@@ -52,6 +52,8 @@ IMuseDigital::IMuseDigital(ScummEngine_v7 *scumm, Audio::Mixer *mixer, int fps)
 	_sound = new ImuseDigiSndMgr(_vm);
 	assert(_sound);
 	_callbackFps = fps;
+	if (_vm->_game.id == GID_FT)
+		_callbackFps *= 2;
 	resetState();
 	for (int l = 0; l < MAX_DIGITAL_TRACKS + MAX_DIGITAL_FADETRACKS; l++) {
 		_track[l] = new Track;
@@ -183,6 +185,8 @@ void IMuseDigital::saveLoadEarly(Common::Serializer &s) {
 			}
 
 			if (_vm->_game.id == GID_CMI) {
+				track->gainRedFadeDest = 127 * 290;
+
 				if (track->soundId / 1000 == 1) { // State
 					for (l = 0; _comiStateMusicTable[l].soundId != -1; l++) {
 						if ((_comiStateMusicTable[l].soundId == track->soundId)) {
@@ -198,6 +202,10 @@ void IMuseDigital::saveLoadEarly(Common::Serializer &s) {
 						}
 					}
 				}
+			}
+
+			if (_vm->_game.id == GID_FT) {
+				track->gainRedFadeDest = 127 * 180;
 			}
 
 			track->sndDataExtComp = _sound->isSndDataExtComp(track->soundDesc);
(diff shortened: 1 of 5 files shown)


engines/scumm/imuse_digi/dimuse.cpp
```diff
@@ -52,6 +52,8 @@ IMuseDigital::IMuseDigital(ScummEngine_v7 *scumm, Audio::Mixer *mixer, int fps)
 	_sound = new ImuseDigiSndMgr(_vm);
 	assert(_sound);
 	_callbackFps = fps;
+	if (_vm->_game.id == GID_FT)
+		_callbackFps *= 2;
```
AndywinXp, line 56: This is the change I'm referring to in the PR description.

AndywinXp: I now have a proper explaination to support why this change above is technically closer to the real thing (while I'm doing the whole decomp thing). @aquadran 

So it appears, from the decomp, that there is a 100 Hz fine grained timer, under which, **fades** and **triggers** are handled at 60 Hz, and simple volume/panning changes (and sound switching I think) are handled at 10 Hz.
With the change above, **everything in the iMUSE engine is handled at 20 Hz** (10*2), which is a good compromise for now, since at the moment the engine is not able to handle the three timers (and the three correspondent behaviors) separatedly with the current code structure.

AndywinXp: After further study I must make some corrections about what I said above: 
while fades and triggers are handled at 60 Hz, it appears that the only thing operated at 10 Hz is the gain reduction on music while speech is playing.

The rest of it (sounds, volume changes, and crossfades generated by a JUMP instruction) are handled at 100 Hz.

I don't think currently the engine can handle filling the sound buffer at 100 fps speed (it can do it, but there are going to be some hiccups here and there). Still, the system needs to be speedier than it currently is, at least 60Hz for fades and triggers detection.

Gain reduction will have to stay at 10Hz, like the game code says.

I'll investigate further.

sev-:
Thank you

PR #2704, opened 2020-12-27
SCUMM: COMI: fix missing soundKludge call

AndywinXp:
This PR patches a script bug (or oversight?) which was of course also present in the original EXE: 
after greasing the cannonballs in the Plunder Town Theater, during the juggling show, the game cuts from room 18 (backstage) to room 19 (stage).

Usually, when loading a room, script 29 is in charge of switching the background music, based on which room we've just loaded.
Unfortunately, during this cutscene, script 29 is not executing, therefore the music is unchanged from room 18 to 19.
This creates a coherency issue between the current music (state 1277, which is a muffled version of the juggling theme, used in the backstage) and the drums fill played afterwards (sequence 2225, which is a continuation of the unmuffled juggling theme played on the stage).

This fix checks for this exact situation happening, and makes a call to the appropriate soundKludge operation, like script 29 would have done.

### DIFF
--- a/engines/scumm/script_v6.cpp
+++ b/engines/scumm/script_v6.cpp
@@ -736,6 +736,25 @@ void ScummEngine_v6::o6_startScript() {
 	script = pop();
 	flags = pop();
 
+	// WORKAROUND for a bug also present in the original EXE: After greasing (or oiling?)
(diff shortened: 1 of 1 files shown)


bluegr:
Thanks, merging

PR #2705, opened 2020-12-27
XEEN: Fix many codacy issues

FrostyHitoshura:
The goal of these changes is to fix all files that are in the "B" quality level or worse for the Xeen engine & creation tool. I noticed other small problems while working on that I also fixed.


dreammaster:
All looks good. Some of it seems somewhat redundant, but I'll assume that there's reasons for the warnings being fixed in the first place

PR #2706, opened 2020-12-29
SCUMM: FT: update cutscene fix to account for Version B

AndywinXp:
Updated the workaround for bug #12022, accounting for the existence of Version B.
Luckily this didn't require comparing md5 values, as it appears there are no edge cases which can take the wrong if-"route" on the non correspondent version.


bluegr:
Quite straightforward. Thanks for your work! Merging

PR #2707, opened 2020-12-30
ENGINES: fix wrong parameter in constructor.

fedor4ever:
Cause compilation error.


bluegr:
Nice catch! It's like this in other use cases of ButtonWidget in this file. Merging

fedor4ever:
ButtonWidget ctor has too many default parameters which lead to errors. I think add overload ctror good idea.

PR #2708, opened 2020-12-31
GUI: Fix RTL layout in engines with custom options widget

SupSuper:
This addresses bug [GUI: I18N - RTL: "Edit game" dialog looks distorted for SCI games](https://bugs.scummvm.org/ticket/12028#no1).

Fixed version

The issue affects all engines with a custom options widget, as currently the RTL layout doesn't apply to widgets created via code. I made the following changes:

- Set `useRTL` to `true` in the `addWidget()` calls.

Should this be the default instead? There's no other uses of `addWidget()`.

- Renamed the dialogs to `"GameOptions"` so the RTL rule would apply correctly: https://github.com/scummvm/scummvm/blob/master/gui/widget.cpp#L119

Should this be a Widget property instead? Relying on a magic name isn't obvious.

### DIFF
--- a/engines/mohawk/dialogs.cpp
+++ b/engines/mohawk/dialogs.cpp
@@ -95,7 +95,7 @@ enum {
 #ifdef ENABLE_MYST
 
 MystOptionsWidget::MystOptionsWidget(GuiObject *boss, const Common::String &name, const Common::String &domain) :
-		OptionsContainerWidget(boss, name, "MystOptionsDialog", false, domain),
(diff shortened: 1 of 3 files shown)


sev-:
I think a better solution would be to change `useRTL` default to `addWidget` to `true`. As we by default making widgets RTL except a few carefully selected cases.

Could you please try that? It will shorten this PR like 10x times.

criezy (approved):
This dialog name change feels like black magic. But if it works that's good.

PR #2709, opened 2021-01-01
SYMBIAN: Global Main Menu called with asterisk.

fedor4ever:
SYMBIAN: call main menu by key asterisk.

Replace hotkeys for calling main menu by asterisk because that used long ago. Asterisk used to call main menu because it presen on phone keyboard. On devices with touchscreen symbian sdl backend provides screen keyboard with several layouts with that key.

### DIFF
--- a/backends/platform/symbian/src/SymbianOS.cpp
+++ b/backends/platform/symbian/src/SymbianOS.cpp
@@ -48,6 +48,9 @@
 #include "backends/saves/default/default-saves.h"
 #include "backends/mixer/symbiansdl/symbiansdl-mixer.h"
 
+#include "backends/keymapper/keymapper.h"
+#include "backends/keymapper/keymapper-defaults.h"
+
 #ifdef GUI_ENABLE_KEYSDIALOG
 #include "backends/platform/symbian/src/SymbianActions.h"
 #include "backends/events/symbiansdl/symbiansdl-events.h"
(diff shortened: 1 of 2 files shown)


ccawley2011:
A better way of doing this would be to override the `getKeymapperDefaultBindings` method in `OSystem_SDL_Symbian`, and setting the default binding there. The Android backend has an example of how to do this [here](https://github.com/scummvm/scummvm/blob/master/backends/platform/android/android.cpp#L494).

sev-:
Indeed. This PR needs to be reworked.

fedor4ever:
Ok
--
Отправлено из myMail для Android пятница, 01 января 2021г., 21:14 +03:00 от Eugene Sandulenko  notifications@github.com :

fedor4ever:
delay due compilation error

fedor4ever:
Is scummvm itself free KeymapperDefaultBindings object? If no why other ports doesnt free?

backends/platform/symbian/src/SymbianOS.cpp
```diff
@@ -210,6 +212,13 @@ RFs& OSystem_SDL_Symbian::FsSession() {
 	return *_RFs;
 }
 
+
+Common::KeymapperDefaultBindings *OSystem_SDL_Symbian::getKeymapperDefaultBindings(){
+	Common::KeymapperDefaultBindings *keymapperDefaultBindings = new Common::KeymapperDefaultBindings();
+	keymapperDefaultBindings->setDefaultBinding(Common::kGlobalKeymapName, "MENU", "ASTERISK");
+}
```
ccawley2011, line 219:
```suggestion
	keymapperDefaultBindings->setDefaultBinding(Common::kGlobalKeymapName, "MENU", "ASTERISK");
	return keymapperDefaultBindings;
}
```

sev-:
Thanks!

PR #2714, opened 2021-01-03
MACOSX: simplify Retina blurriness workarounds and restore Dark Mode

dwatteau:
This is related to bugs #11430 and #11305.

It appears that commit a1d930209a6452649ab704dbbcf2084c9177936e is actually enough to avoid blurriness display issues on Retina screens (i.e. favour OpenGL rendering over Metal rendering inside SDL2, because of interaction problems between macOS on Retina screens and SDL2).

Previous attempts at working around this included disabling Dark Mode (c59bf95ba85b25b134110eb0ff9c3d0e17750630) and passing a particular "invalid" argument to the `-platform_version` linker flag (70f79d3df8b4a9d4c6d1b69838ff5a722c51b61e), but it appears that `SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl")` is sufficient to avoid this bug (until a future version of SDL2 possibly fixes this for good and makes any workaround unnecessary).

(Note when manually testing changes to `Info.plist`: temporary rename your `.app` before launching it again, otherwise macOS may cache some previous values ― that's probably why I thought the other commits were still necessary when submitting my `SDL_HINT_RENDER_DRIVER` PR.)

This works for me with a local Xcode 12.2 build on macOS 11.1, with a quick test of the old builtin theme, which isn't blurry. (I can also test this on my other Retina machine with Mojave if you like, but only in a few weeks.) However, I use newer SDKs than what's used for official builds, so maybe the best way to validate this PR would be to provide an official test build containing it?

And if this PR is accepted, a backport to the `2.2` branch might be wanted.

### DIFF
--- a/configure
+++ b/configure
@@ -2902,9 +2902,6 @@ case $_host_os in
 			fi
 		fi
 
-		# Building with SDK 10.14+ causes blurry display on Retina screens.
-		# A workaround is to set the LC_VERSION_MIN_MACOSX load command's sdk value
-		# to n/a (i.e. 0.0). See bug #11430 for details.
 		echocheck "macOS deployment target"
 		cat > $TMPC << EOF
 #include "AvailabilityMacros.h"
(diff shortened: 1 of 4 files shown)


criezy:
Thank you for this!
I have checked that when I build natively on big sur on my new retina laptop is indeed works. But before I merge this I want to make a build with the release toolchain (which uses an older SDK) to check that it still works properly. I am quite hopeful and will be happy to see the SDK version hack go. And having access to the dark theme will be a nice bonus.

criezy:
I have now verified this works well when building with an older SDK as well.

PR #2715, opened 2021-01-04
CLOUD: Update Dropbox to use refresh_token

Tkachov:
Dropbox [is switching](https://www.dropbox.com/lp/developers/reference/oauth-guide) to short-lived `access_token` OAuth mid 2021. This commit adapts `Cloud::DropboxStorage` to use `refresh_token` similarly to how other Storages do: by introducing a `DropboxTokenRefresher`.

@Mataniko already added a new endpoint on cloud.scummvm.org, which will be used for Dropbox from now on. Older versions of ScummVM, which do not have this change, will still work with "Legacy" endpoint until Dropbox deprecates it.

I've checked all the `Dropbox*Request` classes with Testbed for handling the "expired_access_token" error (handled via the `DropboxTokenRefresher`) and all of them work.

### DIFF
--- a/backends/cloud/dropbox/dropboxcreatedirectoryrequest.cpp
+++ b/backends/cloud/dropbox/dropboxcreatedirectoryrequest.cpp
@@ -21,6 +21,8 @@
  */
 
 #include "backends/cloud/dropbox/dropboxcreatedirectoryrequest.h"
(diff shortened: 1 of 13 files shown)


Mataniko (approved):

sev-:
Thanks!

PR #2718, opened 2021-01-05
GUI: Refresh stretch mode upon returning to launcher

av-dx:
When the Options...->Misc->"Always return to the launcher on leaving a game" option is checked, the "relaunched" launcher GUI will not launch with correct stretch mode, if "Edit Game...->Graphics->Stretch Mode" was set to <default> from some non-default value. This also results in the game re-launching with the same non-default setting as applied above, as described in Issue #12041.

This requires the same fix as was done in PR #2690, which was to remove the hasKey() constraint, because setStretchMode() and ConfMan.get() methods together were able to set the <default> setting properly in absence of a key.

### DIFF
--- a/base/main.cpp
+++ b/base/main.cpp
@@ -340,6 +340,8 @@ static void setupGraphics(OSystem &system) {
 	system.beginGFXTransaction();
 		// Set the user specified graphics mode (if any).
 		system.setGraphicsMode(ConfMan.get("gfx_mode").c_str());
+		system.setStretchMode(ConfMan.get("stretch_mode").c_str());
+		system.setShader(ConfMan.get("shader").c_str());
 
 		system.initSize(320, 200);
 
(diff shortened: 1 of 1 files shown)


bluegr:
So why are the other options handled differently? (e.g. "shading", "fullscreen" etc)

criezy:
I have not tested, but the change makes sense.

> So why are the other options handled differently? (e.g. "shading", "fullscreen" etc)

This is related to whether the option has a `<default>` value in the GUI or not. For those that do, such as `gfx_mode` and `stretch_mode`, if the user select this item in the popup widget this removes the key from the config. So hasKey() returns false, but we still need to call setGraphicsMode()/setStretchMode() (with the default value) so that the default is respected.

Options that are checkboxes, such as `fullscreen` or `filtering`, do not have a `<default>` setting.

But you are right about `shader`. This one does have a `<default>` value and we need to remove the `hasKey()` check. For the options dialog, I did it in commit f88381f at the same time I merged PR #2690. @av-dx can you please add this to this PR for the Return to Launcher case? And please squash it with your existing commit. There is no need to have two separate commits for this.

av-dx:
Thanks for clearing this up. I have removed the ```hasKey()``` check for the ```shader``` setting.

bluegr:
Thanks for your work @av-dx! And thanks for clearing things up @criezy. This change is quite straightforward, and can be merged. Merging

PR #2719, opened 2021-01-06
GRAPHICS: allow scaling for opaque surfaces

mgerhardy:
This adds scaling support for non transparent blit functions of the ManagedSurface class

Related issue is https://bugs.scummvm.org/ticket/12050


sev-:
Thanks!

PR #2726, opened 2021-01-10
SCI: Hebrew in parentheses isn't stage direction

ZvikaZ:
Before this PR, Hebrew text in parentheses was treated as stage direction (because it doesn't contain English lower case letters, and usually doesn't contain digits), and wasn't printed.

This PR makes Hebrew letters behave similar to English lower case letters - if they exist - print the message.
(There is no point in translating the stage direction to Hebrew, if ever there will be a Hebrew dubbing project, the actors will use the English directions...)

This PR is coordinated with the group that translated previous Sierra titles to Hebrew (Torin, GK2, SQ3), and working now on QFG1VGA - that's were we understood that problem. In previous titles, the text in parentheses wasn't translated at all, thus, this fix won't make any unwanted text to start appearing.

### DIFF
--- a/engines/sci/engine/message.cpp
+++ b/engines/sci/engine/message.cpp
@@ -444,6 +444,10 @@ bool MessageState::stringStage(Common::String &outstr, const Common::String &inS
 		// SCI32 seems to support having digits in stage directions
 		if (((inStr[i] >= 'a') && (inStr[i] <= 'z')) || ((inStr[i] >= '0') && (inStr[i] <= '9') && (getSciVersion() < SCI_VERSION_2)))
 			return false;
+
+		// If it contains Hebrew letters, it's not a stage direction
+		if (g_sci->getLanguage() == Common::HE_ISR && (byte)inStr[i] >= 128 && (byte)inStr[i] <= 255)
+			return false;
 	}
 
 	// We ran into the end of the string without finding a closing bracket


noomorph:
Just a few pics to illustrate the point:

Prior the fix:

image

After the fix:

image

Original:

image

engines/sci/engine/message.cpp
```diff
@@ -444,6 +444,14 @@ bool MessageState::stringStage(Common::String &outstr, const Common::String &inS
 		// SCI32 seems to support having digits in stage directions
 		if (((inStr[i] >= 'a') && (inStr[i] <= 'z')) || ((inStr[i] >= '0') && (inStr[i] <= '9') && (getSciVersion() < SCI_VERSION_2)))
 			return false;
+
+		// If it contains Hebrew letters, it's not a stage direction
+		// besides that point, keep the English logic
+		if (g_sci->getLanguage() == Common::HE_ISR)
```
bluegr, line 450: Please, enclose the code in this if statement in brackets - { }, in order to improve readability

ZvikaZ: Done.

engines/sci/engine/message.cpp
```diff
@@ -444,6 +444,15 @@ bool MessageState::stringStage(Common::String &outstr, const Common::String &inS
 		// SCI32 seems to support having digits in stage directions
 		if (((inStr[i] >= 'a') && (inStr[i] <= 'z')) || ((inStr[i] >= '0') && (inStr[i] <= '9') && (getSciVersion() < SCI_VERSION_2)))
 			return false;
+
+		// If it contains Hebrew letters, it's not a stage direction
+		// besides that point, keep the English logic
+		if (g_sci->getLanguage() == Common::HE_ISR) {
+			if (((inStr[i] >= 'a') && (inStr[i] <= 'z')) ||
+				(((byte)inStr[i]) >= 128 && ((byte)inStr[i] <= 255)) ||
+				((inStr[i] >= '0') && (inStr[i] <= '9') && (getSciVersion() < SCI_VERSION_2)))
```
bluegr, line 453: You don't need to repeat the same checks as the ones in the check before this one.
This can be simplified as:

    if (g_sci->getLanguage() == Common::HE_ISR && getSciVersion() < SCI_VERSION_2 && (byte)inStr[i] >= 128 && (byte)inStr[i] <= 255)
         return false;

ZvikaZ: Makes sense.
But I'm not sure regarding the `getSciVersion() < SCI_VERSION_2`.

If we examine the original English check (with added `newline` ):
```
		// If we find a lowercase character or a digit, it's not a stage direction
		// SCI32 seems to support having digits in stage directions
		if (((inStr[i] >= 'a') && (inStr[i] <= 'z')) || 
		    ((inStr[i] >= '0') && (inStr[i] <= '9') && (getSciVersion() < SCI_VERSION_2)))
			return false;
```
it's of the kind `if A || B`, and the version detection is only needed for `B`, while the letter comparison is done regardless of the version. (as also explained in the comment)

Therefore, I think my PR should be further simiplified to:
```
		if (g_sci->getLanguage() == Common::HE_ISR && (byte)inStr[i] >= 128 && (byte)inStr[i] <= 255)
			return false;
```

bluegr: Indeed, it can be simplified to just that. Please do so, and we'll get this merged :)

ZvikaZ: Great.
I've pushed it.

bluegr: Thanks, merged

PR #2727, opened 2021-01-10
ALL and VITA: Link with GNU Make recursive flag

lephilousophe:
There are three changes in this PR which are dependent, thus the grouping.

First one is to use $(LD) instead of $(CXX) when linking.
I think it's more coherent and makes it easier to grep linking places in makefiles.
This change shouldn't have any impact as we set LD to ${CXX} in configure script

Second one is to make all link commands Make recursive by prepending them with a +.
We can then use -flto=jobserver to let GCC handle LTO threading properly.
This basically tells to make to open 2 more FDs for the child process so it shouldn't have any impact when linker doesn't use it.
(That's what GNU make documentation implies)

Third one is to make use of -flto=jobserver on Vita builds instead of hardcoding a 20 cores value which can have negative impacts on some servers.
Specifying number of cores for compilation is useless and is removed.


henke37:
I presume that this has been tested as not to confuse `create_project`?

lephilousophe:
I don't see why `create_project` would be impacted by this change, I don't think it parses Makefiles.
Do you have a precise idea about what would fail?

henke37:
As a matter of fact, it _does_ parse the make files. It happens in `ProjectProvider::createModuleList` in the `create_project.cpp` file.

lephilousophe:
It seems it reads variables definitions (and ifeq, ...) and not recipes.

lotharsm:
Looking good. +1 from me.

sev-:
Thanks, merging

PR #2728, opened 2021-01-12
KYRA: More adlib driver code cleanup

miller-alex:
Again some improvements to the code that shouldn't change behaviour in any way.

If you don't like that I'm changing the way the callback functions work, you may drop the last commit, but I think it's cleaner like that.

### DIFF
--- a/engines/kyra/sound/drivers/adlib.cpp
+++ b/engines/kyra/sound/drivers/adlib.cpp
@@ -659,21 +659,17 @@ void AdLibDriver::executePrograms() {
 	// each AdLib channel (0-8), plus one "control channel" (9) which is
 	// the one that tells the other channels what to do.
 
-	// This is where we ensure that channels that are made to jump "in
-	// sync" do so.
-
 	if (_syncJumpMask) {
-		bool forceUnlock = true;
+		// This is where we ensure that channels that are made to jump
+		// "in sync" do so.
 
 		for (_curChannel = 9; _curChannel >= 0; --_curChannel) {
-			if ((_syncJumpMask & (1 << _curChannel)) == 0)
-				continue;
-
-			if (_channels[_curChannel].dataptr && !_channels[_curChannel].lock)
-				forceUnlock = false;
+			if ((_syncJumpMask & (1 << _curChannel)) && _channels[_curChannel].dataptr && !_channels[_curChannel].lock)
+				break; // don't unlock
 		}
 
-		if (forceUnlock) {
+		if (_curChannel < 0) {
+			// force unlock
 			for (_curChannel = 9; _curChannel >= 0; --_curChannel)
 				if (_syncJumpMask & (1 << _curChannel))
 					_channels[_curChannel].lock = false;
@@ -682,7 +678,6 @@ void AdLibDriver::executePrograms() {
 
 	for (_curChannel = 9; _curChannel >= 0; --_curChannel) {
 		Channel &channel = _channels[_curChannel];
-		int result = 1;
 
 		if (!channel.dataptr)
 			continue;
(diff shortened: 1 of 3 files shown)


miller-alex:
Additional note/question:

I've noticed an inconsistency in the encoding of addresses in the sound programs. There are two diffent encodings: in EoB addresses have an offset of 191, newer games use relative addresses. But the four callbacks with addresses as parameters use different conditions: update_checkRepeat() uses always relative addresses, update_jump() uses Eob style for version < 2, update_jumpToSubroutine() uses EoB style for version < 3, and update_setupSecondaryEffect1() always uses EoB style addresses. I don't know whether this is an inconsistency in the original drivers, a bug, or doesn't matter since the features aren't used by the games where it would make a difference. But unless it needs to be the way it is, the same code should be used everywhere to avoid confusion.

Do you know whether this can/should be changed?

I also considered reordering struct Channel to reduce padding, but didn't. What do you think?

miller-alex:
DeepCode seems to be deeply confused. The description of the issue is nonsense. Maybe it doesn't like line 755?

athrxx:
Heya, thanks for your work on this. Please ignore the deepcode and codacy output, unless there is actually something valid in it. No need to make code changes just to silence the tools. The only thing you really want to check is the travis output for clang and gcc (there are no warnings from your code, so all is good).

From first glance I don't see anything I wouldn't be willing to merge. Maybe change the long type to int or int32, since we don't seem to use that type in Kyra, but that's not a real point of critique...

Regarding your question about the offsets: I would like to think that we got it right the way it is, since we went over the original driver disasms often enough. But honestly, I'll just have to take another look to be sure. I'll get back to you when I have your answer.

Changing the channel struct layout isn't really necessary, unless you really want to do it and the code layout doesn't suffer from it.

athrxx:
So, about the offsets: the absolute offsets in the original drivers are inconsistent, as they are not in match with the data provided by the sequencer (except maybe EOB1? - I remember fixing the teleporter sounds in EOB2, but EOB1 might have been correct).
From the data files (or from EOB1? can't remember) I managed to figure out that the sequencer must have based the offsets on 191.

This is what I found in the disasms:

- secondaryEffect1(): all driver versions resolve the address via xlat so it will be an absolute offset from the ds (=current) segment.
			--> so this one seems to be okay (or at least can't reasonably be fixed).

- update_checkRepeat()/update_jump(): only the EOB 1 version driver reads an absolute offset (in current segment), all other versions read a relative offset (EOB has 'mov si, ax', the others have 'add si, ax'
                        --> so this is correct (this would definitely have caused issues if it weren't)

- update_jumpToSubroutine(): EOB 1 and 2 version drivers use absolute offsets (in current segment), the later versions read a relative offset (EOB 1 + 2 have 'mov si, ax', the others have 'add si, ax'
                        --> so this is correct (this would also have caused issues if it weren't)


Apparently Westwood changed these offsets to relative over time whenever they spotted a particular bug.
I believe that the xlat in secondaryEffect1() for the later versions is an oversight. I have run all Kyra 1 and 2 sound effects to see whether any one of these triggers the secondaryEffect1. Just played through all possible track numbers from all files. Not one single secondaryEffect1(). 

Still, I see no way to fix it, unless it becomes evident that there is actual corresponding data which is also correspondingly fixed.

miller-alex:
Thank you for taking the time to check the situation in the disasms. Well, since the code can't be made more consistent, I won't change it, of course.

athrxx:
Is this ready to be merged? I have done a few quick tests with the relevant games. Seems like everything still works... ;-)

miller-alex:
Yes, I think it's ready.

PR #2730, opened 2021-01-13
BACKENDS: search also for track_x and track_xx audio cd tracks

mgerhardy:
this allows us to use the unmodified version of dotemu lba1 enhanced music track paths

see 1b67a288abbed4e130fac8507d8f66940fcc0530

### DIFF
--- a/backends/audiocd/default/default-audiocd.cpp
+++ b/backends/audiocd/default/default-audiocd.cpp
@@ -24,6 +24,7 @@
 #include "audio/audiostream.h"
 #include "common/config-manager.h"
 #include "common/system.h"
+#include "common/util.h"
 
 DefaultAudioCDManager::DefaultAudioCDManager() {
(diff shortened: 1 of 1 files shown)


mgerhardy:
Maybe the buffer should get increased or snprintf should be used. `track_2147483648` would overflow the buffer by 1 byte

mgerhardy:
Using `Common::String::format` now - still fits into the internal buffer, so no mem allocations - `openStreamFile` was converting it into a `Common::String` anyway. Also replaced the hardcoded arraysize.

bluegr:
This looks good, we don't need to add any extra logic for these tracks, other than their naming template, so there is no impact, and the dotemu version will work out of the box.

I propose to merge this this week, unless there are any objections.

bluegr:
Merging as mentioned. Thanks for your work!

PR #2731, opened 2021-01-14
ULTIMA4: Change damage to match DOS version

Fenyx4:
Through an interesting series of events I noticed that the xu4 damage was different from the DOS version of the game.
Disassembled DOS version: https://github.com/ergonomy-joe/u4-decompiled/blob/master/SRC/U4_AI.C#L127

Note: D_23D2[C_7C25((unsigned char)Fighters._tile[bp04])] is equivalent to creature.basehp

If I'm doing the math right then they are equal when the monster base health is below 64. But above that they diverge with xu4 creatures doing significantly less damage.

(It is possible that the xu4 version is getting the damage calculation from a different port of the game. I haven't been able to check that.)

### DIFF
--- a/engines/ultima/ultima4/game/creature.cpp
+++ b/engines/ultima/ultima4/game/creature.cpp
@@ -253,7 +253,7 @@ int  Creature::getDamage() const {
 	int damage, val, x;
 	val = _baseHp;
 	x = xu4_random(val >> 2);
-	damage = (x >> 4) + ((x >> 2) & 0xfc);
+	damage = (x >> 4) * 10;
 	damage += x % 10;
 	return damage;
 }


Thunderforge:
My understanding is that this new damage calculation will apply to all versions of Ultima IV loaded in SummVM (Apple II, Commodore 64, DOS, Amiga, etc), right?

If possible, we should check the Apple II version of the game before merging, since according to Wikipedia, that was the initial version programmed and others, including DOS, are technically ports.

Fenyx4:
Sadly I don't know assembly so I can't figure out the apple II damage calculation.
If someone wants to give it a go here is the disassembled code; https://github.com/jtauber/u4a2/blob/master/DISK_01/ULT4/full_listing.md

I did just load up the Apple II version and let myself get pummeled by an orc. I got hit for these values [0, 3, 4, 5, 8, 10, 13, 14, 15, 16, 17]. Which is more than an orc can hit you with the current xu4 calculation.
(It also hit me for 10, 13, and 14 which the DOS version of an orc can't hit you for. So we know they are different calculations between DOS and Apple II.)

Which version of U4 is ScummVM trying to match?

dreammaster:
Looks good to me, I'm going to accept it. As far as I'm aware, xu4 is based on the PC version.

dreammaster:
And kudos for noticing the problem, and taking the time to investigate the matter

Fenyx4:
Thanks @dreammaster !

PR #2736, opened 2021-01-17
README:  Add ScummVm logo in README

ayushbisht2001:
@sev- I have added the ScummVm logo in README.md.


sev-:
Why do you think we need to add our logo like this? There is no point in it.

Besides, your commit log messages still do not conform to our standards, and in general, having 3 commits for such simple thing looks odd.

PR #2739, opened 2021-01-22
DOCS: Adds the doc portal to the 2.2.0 branch

cadihowley:
I’ve added the complete doc portal to the 2.2.0 branch, so that Read the Docs can build from that branch for the current 2.2.0 documentation. Future docs can then be pushed to the master branch.


criezy:
Committing to the `branch-2-2-0` now feels unorthodox, although since there is no change to the source code we could maybe do an exception.
But my suggestion would be to commit to `branch-2-2` and then maybe create a new `v2.0.0-doc` tag, or something similar. My understanding is that Read The Docs uses such tags for versions (see https://docs.readthedocs.io/en/stable/versions.html).

PR #2741, opened 2021-01-23
SCUMM: Digital iMUSE: fix FT crash and music breaking bug

AndywinXp:
This is an urgent fix which is contained in PR [#2701](https://github.com/scummvm/scummvm/pull/2701). Since that PR is quite big and it's taking some time to be approved, and since people are rightfully telling me that FT is currently broken in the current dev build, I'm sending these _urgent_ fixes as a separate PR.

_Please_ merge this as soon as you can, since this fixes Full Throttle crashing and another bug in COMI which allowed zero volume (toBeRemoved) tracks to be able to make JUMPs and loops before being flushed.

### DIFF
--- a/engines/scumm/imuse_digi/dimuse.cpp
+++ b/engines/scumm/imuse_digi/dimuse.cpp
@@ -589,22 +589,25 @@ void IMuseDigital::switchToNextRegion(Track *track) {
 	}
 
 	int jumpId = _sound->getJumpIdByRegionAndHookId(soundDesc, track->curRegion, track->curHookId);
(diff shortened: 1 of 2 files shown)


criezy:
I missed that one. I am guessing it fixes [bug #12156](https://bugs.scummvm.org/ticket/12156).
Let's merge.

criezy:
And indeed it fixes the aforementioned bug. Thank you!

AndywinXp:
Thanks @criezy ! 🙏

lgRecip:
Thanks.

PR #2744, opened 2021-01-24
SDL: Fix trying to set unsupported pixel format when changing  gfx mode

criezy:
When switching between the SDL and OpenGL graphics managers, trying to restore the state could fail as the two managers do not have the same list of supported formats, so we may not be able to transfer the pixel format from one to the other. This then resulted in an assert.

This fixes [bug #12079](https://bugs.scummvm.org/ticket/12079).

I am not sure the way I fixed it is the best way to do it. So I am opening that pull request for review before merging.


criezy:
Merging this now. I am not particularly happy with the solution, but after two weeks I still don't have a better idea.

PR #2746, opened 2021-01-25
SCUMM: COMI: fix hebrew multiline dialogue

BLooperZ:
Fixes bug in dialogue text in COMI hebrew fan translation,
bug was: when the dialogue option should have span on multiple lines, it was instead cropped out of the screen.

this is solved by aligning the text to the right after lines are splitted.

Thanks.

### DIFF
--- a/engines/scumm/verbs.cpp
+++ b/engines/scumm/verbs.cpp
@@ -1013,9 +1013,6 @@ void ScummEngine_v7::drawVerb(int verb, int mode) {
 		_charset->setCurID(vs->charset_nr);
 
 		// Compute the text rect
-		if (_language == Common::HE_ISR) {
-			vs->curRect.left = _screenWidth - _charset->getStringWidth(0, buf);
-		}
 		vs->curRect.right = 0;
 		vs->curRect.bottom = 0;
 		const byte *msg2 = msg;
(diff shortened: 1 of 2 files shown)


BLooperZ:
Also, added fix for regression in hebrew text in smush videos introduced in https://github.com/scummvm/scummvm/commit/9ff197ef8bb95998d831a51a7d586fd628613925
Where first character in sentence didn't appear, and multiline text was just repeated all over.

engines/scumm/verbs.cpp
```diff
@@ -1046,12 +1043,26 @@ void ScummEngine_v7::drawVerb(int verb, int mode) {
 				}
 				--len;
 			}
+			if (_language == Common::HE_ISR) {
+				vs->curRect.right -= vs->curRect.left;
+				vs->curRect.left = _screenWidth - _charset->getStringWidth(0, tmpBuf);
+				vs->curRect.right += vs->curRect.left;
+			}
```
bluegr, line 1050: Can you explain what you're doing here?

BLooperZ: Aligning the text to the right.
The first and third lines is replacing the old left position parameter with the new left position. (substract old, add new)

bluegr: Yes, but you're trying to right-align the text relative to the screen Is that intended? Does the original align the text to the left of the screen?

BLooperZ: Yes, this is the dialogue on SCUMM-7-8, which is aligned to the left.
altough I have checked this now again, the original version of COMI is indeed indented 5 pixels to the left (value of `vs->curRect.left`). (not 0 as I assumed)
ideally we would use this value to calculate
`drawLeft = _screenWidth - originalLeft - _widthOfString`
problem is the updated value is used for the clickable/hoverable mouse area, and it causes the new value to be received when drawing the next time, so using it make the text to jump from side to side.
the difference is (IMO) somewhat neglectable visually, though.
could be solved by making 2 `_verbs` arrays, but that is a change that may affect much more games.

bluegr:
Thanks, merging

PR #2747, opened 2021-01-28
ULTIMA8: add new engine options

OMGPizzaGuy:
ULTIMA8: Add engine option for replacing the game font with ttf rendering

This option was previously controlled by pentagram.ini "ttf" setting.
I'd prefer a review as this is the first option added for this engine.

ULTIMA8: Add original save/load engine option

Original save / load menus were already implemented.
Creating an option gives us the ability to use the ScummVM dialogs from the game menu.


mduggan:
All looks good to me.  Thanks for implementing!

PR #2748, opened 2021-01-29
GRAPHICS: Improve TTF rendering on transparent surfaces.

OMGPizzaGuy:
Update color blending to match blending found in ManagedSurface. TTF rendering assumed the alpha channel of the surface was completely opaque. ULTIMA8 TTF rendering attempts to use a transparent surface and became near-illegible for text on scrolls. This change will affect TTF rendering on surfaces that are not already painted.

### DIFF
--- a/graphics/fonts/ttf.cpp
+++ b/graphics/fonts/ttf.cpp
@@ -599,20 +599,16 @@ static void renderGlyph(uint8 *dstPos, const int dstPitch, const uint8 *srcPos, 
 				uint8 dA, dR, dG, dB;
 				dstFormat.colorToARGB(*rDst, dA, dR, dG, dB);
 
-				if (dA == 0) {
-					*rDst = dstFormat.ARGBToColor(sA, sR, sG, sB);
-				} else {
-					double alpha = (double)sA / 255.0;
-					dR = static_cast<uint8>((sR * alpha) + (dR * (1.0 - alpha)));
-					dG = static_cast<uint8>((sG * alpha) + (dG * (1.0 - alpha)));
-					dB = static_cast<uint8>((sB * alpha) + (dB * (1.0 - alpha)));
+				double sAn = (double)sA / 255.0;
+				double dAn = (double)dA / 255.0;
+				double oAn = sAn + dAn * (1.0 - sAn);
 
-					if (sA > dA) {
-						dA = static_cast<uint8>((sA * alpha) + (dA * (1.0 - alpha)));
-					}
+				dR = static_cast<uint8>(sR * sAn + dR * dAn * (1.0 - sAn) / oAn);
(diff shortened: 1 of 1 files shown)


mduggan:
This looks good to me. I tried to find other uses of this code to see if there's any potential side-effects but I didn't find any place I thought it could cause any change other than U8.  I may have missed something though so if anyone else wants to comment please feel free.

graphics/fonts/ttf.cpp
```diff
@@ -594,16 +594,25 @@ static void renderGlyph(uint8 *dstPos, const int dstPitch, const uint8 *srcPos,
...
+					double alpha = (double)sA / 255.0;
+					dR = static_cast<uint8>((sR * alpha) + (dR * (1.0 - alpha)));
+					dG = static_cast<uint8>((sG * alpha) + (dG * (1.0 - alpha)));
+					dB = static_cast<uint8>((sB * alpha) + (dB * (1.0 - alpha)));
 
-				*rDst = dstFormat.RGBToColor(dR, dG, dB);
+					if (sA > dA) {
+						dA = static_cast<uint8>((sA * alpha) + (dA * (1.0 - alpha)));
+					}
+
```
criezy, line 613: Where do those formula come from?
My understanding of color blending is that it should be:
```c++
double sAn = (double)sA / 255.0;
double dAn = (double)dA / 255.0;
dR = static_cast<uint8>(sR * sAn + dR * dAn * (1.0 - sAn) / (sAn + dAn * (1.0 - sAn)));
dG = static_cast<uint8>(sG * sAn + dG * dAn * (1.0 - sAn) / (sAn + dAn * (1.0 - sAn)));
dB = static_cast<uint8>(sB * sAn + dB * dAn * (1.0 - sAn) / (sAn + dAn * (1.0 - sAn)));
dA = static_cast<uint8>(sA + dA * (1.0 - sAn));
```
This also means there is no need for a special case if `dA` is 0.
And you can of course optimise that code a bit by computing `dAn * (1.0 - sAn)` only once instead of 6 times.

OMGPizzaGuy: Mostly followed suit with managed_surface.cpp:487-490
Figured if case for dA is 0 might be better to avoid calculations, but perhaps not.
Destination alpha... I just wasn't sure the best call.

OMGPizzaGuy: OK, looking at https://en.wikipedia.org/wiki/Alpha_compositing#Alpha_blending
If I'm understanding it correctly, what I used would be correct for dA being 255, but not complete enough for any other value.
I'll take another swing at this.

OMGPizzaGuy: Yep, the suggested code was great. 
I went with calculating a normalized output alpha and converting that a byte at the end.
I thought it would be a little more clear to anyone reading the code.

criezy (approved):
Thank you. It looks good to me now.

PR #2750, opened 2021-02-02
DIRECTOR: Adding function and game

CustomLenny:
Add the Windows version of "The Simpsons: Cartoon Studio" and its player to the detection in the _Director_ engine.
Implement the kTheMaxInteger function in the lingo switch statement.


sev-:
Thanks!

PR #2751, opened 2021-02-03
CREATE-PROJECT: Various portability fixes for Mac OS X Leopard and other older systems

dwatteau:
Here's a set of changes I had to make so that `create_project` could be built on Mac OS X Leopard.

In summary:

* pick the `override` compat macro from `c++11-compat.h` to `msbuild.h`, because this file makes use of the `override` keyword, and older compilers don't have it.
* use `<CommonCrypto/CommonDigest.h>`, not `<CommonCrypto/CommonCrypto.h>` for `CC_MD5()`. This is what the manpage documents, and Leopard will not find it without the proper header.
* use the older POSIX.1-2001 syntax for `realpath()`, not the POSIX.1-2008 one, because it will cause a crash on older systems. Don't `free(rp);` anymore, since we're no longer using the heap for this.

This is maybe useless (although the one for `CommonDigest.h` still looks good to me), or you may want to do this a bit differently, but here's my proposal :)

Tested on Leopard and Mojave.

### DIFF
--- a/devtools/create_project/create_project.h
+++ b/devtools/create_project/create_project.h
@@ -27,6 +27,10 @@
 #define __has_feature(x) 0 // Compatibility with non-clang compilers.
 #endif
 
+#if __cplusplus < 201103L && (!defined(_MSC_VER) || _MSC_VER < 1700)
+#define override           // Compatibility with non-C++11 compilers.
+#endif
+
 #include <list>
 #include <map>
 #include <map>
--- a/devtools/create_project/xcode.cpp
+++ b/devtools/create_project/xcode.cpp
@@ -29,7 +29,7 @@
 #ifdef MACOSX
 #include <sstream>
 #include <iomanip>
-#include <CommonCrypto/CommonCrypto.h>
+#include <CommonCrypto/CommonDigest.h>
 #endif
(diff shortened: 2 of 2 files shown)


sluicebox:
Maybe move the override workaround to create_project.h so that it's applied everywhere, otherwise using override in the future in a different file will break this again.

These changes look good to me, thanks! I appreciate create_project building in old environments.

dwatteau:
> Maybe move the override workaround to create_project.h so that it's applied everywhere, otherwise using override in the future in a different file will break this again.
> 
> These changes look good to me, thanks! I appreciate create_project building in old environments.

Thank you! I've edited the PR with your suggestion.

sluicebox:
xcode.cpp was just updated in master and now there's a conflict, can you please rebase this against master?

dwatteau:
> xcode.cpp was just updated in master and now there's a conflict, can you please rebase this against master?

Ah yeah, I had a conflict with one of my own commits there :)

It should be OK now.

sluicebox:
I've tested this on Big Sur (M1) and Catalina (Intel) and each change makes sense. Nice job!

PR #2753, opened 2021-02-05
AUDIO: Combine WaveFormat enums under one header

SupSuper:
I noticed the A/V code duplicates the WAV Format Tags multiple times, so I combined all of them into the common `wave_types.h` header, as this should make it easier to add new formats in the future and avoid conflicts.

Dunno if decimal or hexadecimal is better, but since the format specification uses hexadecimal I went with that.


bluegr:
Quite a straightforward change, and makes perfect sense. Thanks for your work on this! Merging

PR #2755, opened 2021-02-07
CONFIGURE: Update config.* to latest version

lephilousophe:
New version handles Darwin arm64

### DIFF
--- a/config.guess
+++ b/config.guess
@@ -1,8 +1,8 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2020 Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


criezy:
Building works fine on a mac M1 both with and without this change. So it's fine with me.
Can you clarify what are the benefit? The only difference I have noticed is that the host CPU architecture is recognised as `aarch64` instead of `arm`, but this doesn't seem to make much of a difference in the end.

lephilousophe:
It's just this.
As @sev- already updated this code on a regular basis, I thought it was fine to do it but buildbot can live without it.
It currently uses aarch64 prefix.

criezy:
Yes, it's fine to update that code. Merging.

PR #2756, opened 2021-02-07
AGS: Add detection entries for 'Unavowed (English/Windows)'

lotharsm:
This patch adds an detection entry for the Windows release of "Unavowed" available at Steam and GoG.

The came is currently unable to launch and fails with the following stacktrace:
https://paste.ubuntu.com/p/zT8KCCH9wk/


raziel-:
I didn't know two files with different sizes can have the exact same md5sum...nice

lephilousophe:
md5 checksum is computed only on first 5000 bytes

raziel-:
@lephilousophe 

D'oh...I mean, I knew that :-D

Just ignore me... :-/

PR #2761, opened 2021-02-08
DOCS: Adds doc portal for v2.2 docs

cadihowley:
As discussed, I am opening a PR to add the existing v2.2.x documentation to the branch-2-2, so that we can work on new documentation on the master branch.


Mataniko:
I think what we'd want to do here is to rebase the individual commits onto the branch. The easiest way to do it is probably just open a PR from the DOCS branch, same way we did originally, and instead of master, apply it to 2.2

PR #2766, opened 2021-02-10
AGI: support new WinAGI *.wag file format

ZvikaZ:

### DIFF
--- a/common/ini-file.cpp
+++ b/common/ini-file.cpp
@@ -32,7 +32,7 @@ bool INIFile::isValidName(const String &name) const {
 	if (_allowNonEnglishCharacters)
 		return true;
 	const char *p = name.c_str();
-	while (*p && (isAlnum(*p) || *p == '-' || *p == '_' || *p == '.' || *p == ' '))
+	while (*p && (isAlnum(*p) || *p == '-' || *p == '_' || *p == '.' || *p == ' ' || *p == ':'))
 		p++;
 	return *p == 0;
 }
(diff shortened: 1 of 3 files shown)


bluegr:
Which games use this newer format? Is there any documentation on it, which was used to deduce the newer syntax?

ZvikaZ:
I'm not aware of any games that are using the new format; but I'm also not aware of games using the old format.
I've stumbled upon this while working on translation of PQ1. I noticed that it's trying to parse the .wag file (because the game wasn't detected due to my changes), but failed.

Any of the games that used the old .wag file might use the newer file, if they will have some revision, using newer WinAGI.
I assumed that there are considerations that led to adding support for .wag file in the first place, and they might hold for future games. Therefore, I wanted this feature to work. 

Regarding the format - well, I haven't seen official documentation of the file. But looking at it, it's seems to be just a simple text INI file.

Here's the beginning of the file:

	#
	# WinAGI Game Property File
	#

	[General]
	   WinAGIVersion = 2.1.1
	   GameID = pq1
	   Interpreter = 2.917

AGKorson:
I am the author of WinAGI (which uses *.wag files). The format was updated in version 1.2.3 (current version is 2.1.5). The new format is text based as ZvikaZ describes. The format specifications enforced by WinAGI are:

  'elements of a settings file:
  '
  '  #comments begin with hashtag; all characters on line after hashtag are ignored
  '  'comments can be added to end of valid section or key/value line
  '  blank lines are ignored
  '  [::BEGIN group::] marker to indicate a group of sections
  '  [::END group::] marker to indicate end of a group
  '  [section] sections indicated by square brackets; anything else on the line gets ignored
  '  key=value  key/value pairs separated by an equal sign; no quotes around values means only
  '    single word; use quotes for multiword strings
  '  if string is multline, use '\n' control code to represent line breaks  (and use multiword option)
(The group markers are used to keep resource sections (i.e. Logic1, Logic2, etc) together, and can safely be ignored when reading a wag file.)

If you need a detailed list of recognized properties with descriptions or any other information, let me know. You can reach me on the http://sciprogramming.com forums.

bluegr:
Good to have feedback by the WinAGI author :) Thanks for the information!
Nice work, and a good addition to the AGI detector. Merging

PR #2767, opened 2021-02-12
AGS: Plugin stubs for Nickenstien's NGFX Graphics Library

kyandora:
This pull request adds stubs for Nickenstien's NGFX Graphics Library used by Size Five Games' Time Gentlemen, Please! (timegentlemenplease). ScummVM now proceeds in-game instead of immediately crashing.

### DIFF
--- a/engines/ags/module.mk
+++ b/engines/ags/module.mk
@@ -294,6 +294,7 @@ MODULE_OBJS = \
 	plugins/ags_flashlight/ags_flashlight.o \
 	plugins/ags_galaxy_steam/ags_blackwell_steam.o \
 	plugins/ags_galaxy_steam/ags_galaxy_steam.o \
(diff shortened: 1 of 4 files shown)


dreammaster (changes requested):
I would actually prefer any such plugins added follow the new style I'm promoting, where each plugin derives from the DLL class, and implements all the methods separately, to make it easier if we later decide to implement the contents. Particularly in cases like this.. from the names of methods, the library provides various effects that we may have to eventually reverse engineer and re-implement, so it'll simplify matters if there are already methods in place we can just put the contents into.

Just look at some of the other plugins for the example of how the methods are defined, and the parameters are extracted.

dreammaster:
Whoops. Apologies for overlooking that you'd done the corrections I asked. Merging it in now

dreammaster:
I manually cherry-picked your commit, and fixed compilation issues

PR #2770, opened 2021-02-13
ENGINES: Make creation of savegame thumbnails an overrideable method

dreammaster:
I think this is a straightforward minor change, but since it touches the engine base, I thought I'd do a pull request for a quick review.

This change changes the methods for appending the new style extra savegame info to instance methods, and the method that generates the thumbnail a virtual method. This will allow engines to override the method in their own meta engine subclass, and do things like create a thumbnail without any on-screen dialogs showing.


mgerhardy:
Nice change - I could use this for twine, too.

PR #2774, opened 2021-02-14
AGS: Fix Android crash to OS on game launch

lowtraxx:
The Android NDK uses unsigned char by default. Fixed the decompression algorithm by replacing them with int8. Also not using the fixed PixelFormat with OpenGL as it is not supported there. These changes allow the Games to run, albeit with messed up colors.

### DIFF
--- a/engines/ags/shared/util/compress.cpp
+++ b/engines/ags/shared/util/compress.cpp
@@ -177,7 +177,7 @@ int cunpackbitl(uint8 *line, int size, Stream *in) {
 		if (in->HasErrors())
 			break;
 
-		char cx = ix;
+		int8 cx = ix;
 		if (cx == -128)
 			cx = 0;
 
@@ -214,7 +214,7 @@ int cunpackbitl16(uint16 *line, int size, Stream *in) {
 		if (in->HasErrors())
 			break;
 
-		char cx = ix;
+		int8 cx = ix;
 		if (cx == -128)
 			cx = 0;
 
@@ -251,7 +251,7 @@ int cunpackbitl32(uint32 *line, int size, Stream *in) {
 		if (in->HasErrors())
 			break;
 
-		char cx = ix;
+		int8 cx = ix;
 		if (cx == -128)
 			cx = 0;
 


engines/ags/ags.cpp
```diff
@@ -386,7 +386,11 @@ SaveStateList AGSEngine::listSaves() const {
 }
 
 void AGSEngine::setGraphicsMode(size_t w, size_t h) {
+#ifdef SCUMM_LITTLE_ENDIAN
```
lephilousophe, line 389: Why this ifdef? It also applies on x86 targets.
It implies that this piece of code didn't work on x86 either.
Your previous PR used OPENGL here.

lowtraxx: Thank you for noticing. Applied the wrong (older) patch by accident. Amended now.

antoniou79:
Looks good to me.

lephilousophe:
LGTM but I am not the most relevant on this PR.

engines/ags/ags.cpp
```diff
@@ -386,7 +386,11 @@ SaveStateList AGSEngine::listSaves() const {
 }
 
 void AGSEngine::setGraphicsMode(size_t w, size_t h) {
+#ifdef USE_OPENGL
+	Graphics::PixelFormat FORMAT(4, 8, 8, 8, 8, 0, 8, 16, 24);
```
lephilousophe, line 390: Maybe add a comment here that the format isn't correct and colors don't match? At least it's what I understood from your previous PR.

lephilousophe (approved):

engines/ags/ags.cpp
criezy: I am not sure about including the pixel format change here since this is not a proper fix but rather a quick hack to allow the games to start (with wrong colours) if I understand properly? Ideally we would use `System::getSupportedFormats()` and do format conversions where needed. Maybe include the change as a separate commit if you want to include it since this is unrelated to the `char` -> `int8` change?

engines/ags/lib/allegro/gfx.cpp
```diff
@@ -147,7 +147,11 @@ BITMAP *create_bitmap_ex(int color_depth, int width, int height) {
 		format = Graphics::PixelFormat(2, 5, 6, 5, 0, 11, 5, 0, 0);
 		break;
 	case 32:
+#ifdef USE_OPENGL
+		format = Graphics::PixelFormat(4, 8, 8, 8, 8, 0, 8, 16, 24);
+#else
```
criezy, line 152: Does this change really make sense? We could either keep bitmaps in their original format and do the pixel conversion when blitting to the game surface, or we could do the conversion when loading the data, but this does not seem to be included in your changes.

criezy (commented):
The `char` to `int8` change makes perfect sense.

I am not sure about the changes to the pixel formats though, and I have added relevant comments on the changes.

engines/ags/ags.cpp
criezy: Also this change uses a compile time define, which means it also changes the format when using the surface sdl graphics manager on SDL ports (and I guess break the coolers there). Ideally if you want to include such a change it should only be done when the OpenGL graphics manager is used (so as a runtime check rather than `#define`).

engines/ags/lib/allegro/gfx.cpp
lowtraxx: I am working on a replacement FakeTexture anyways, that allows to render with the right colors. That makes this osolete. I just inserted it, because otherwise AGS on Android would not run at all because it would not detect the right format.

engines/ags/ags.cpp
lowtraxx: Will open a new PR for fixing the Color Formats.

engines/ags/lib/allegro/gfx.cpp
lowtraxx: Will open new PR with PixelFormat fixes.

criezy:
Thank you for moving the pixel format change to a separate pull request.
This one can be merged now.

PR #2776, opened 2021-02-14
Add engine option widget for language selection

criezy:
AGS games can have multiple languages by providing additional translations as .tra files. The language selection is done by setting the config "translation" key to the tra file name. This means that we cannot easily use the standard ScummVM language selection for this purpose as the tra file name may not map to language names easily.

Also the advanced detector does not allow specifying multiple languages for a game, and adding separate detections for all the languages by using the tra files for the detection would be cumbersome.

Instead this pull request  does something similar to the original games by detecting at runtime the available .tra files, but replaces the original setup utility by a game option widget.

image

It's not ideal that we have an ineffective Language selector in the Game tab, and another one in the Engine tab. Maybe to avoid confusion we could add a GUIO_NOLANGUAGE  option that would hide the Language selector in the Game tab for the games that define it.

### DIFF
--- a/engines/ags/detection.cpp
+++ b/engines/ags/detection.cpp
@@ -24,8 +24,14 @@
 #include "common/config-manager.h"
 #include "common/file.h"
 #include "common/md5.h"
+#include "common/str-array.h"
+#include "common/translation.h"
 #include "ags/detection.h"
 #include "ags/detection_tables.h"
+
+#include "gui/ThemeEval.h"
+#include "gui/widget.h"
+#include "gui/widgets/popup.h"
 
 namespace AGS3 {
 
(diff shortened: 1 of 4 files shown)


henke37:
I don't think this should be an engine specific feature, since other engines are in the same situation, such as `supernova`.

criezy:
And for reference here is the original setup utility, courtesy of @SupSuper:
image

criezy:
> I don't think this should be an engine specific feature, since other engines are in the same situation, such as `supernova`.

Supernova is in a different situation because we do not need to detect at run time which languages are available. So this can be done in the detection. And if we wanted to remove the multiple detection entries, the solution would be to extend the advanced game detector to support having multiple languages for a game. Then the Language popup in the Game tab would let the user select the language.

criezy:
I will merge this now. Further work on the config can then be done directly in master.

PR #2778, opened 2021-02-15
DOCS: Adds V2.2 documentation portal to branch-2-2

cadihowley:


lephilousophe:
I'm not sure last one is OK as @ccawley2011 did some changes on the DS port before pushing this documentation.

cadihowley:
> I'm not sure last one is OK as @ccawley2011 did some changes on the DS port before pushing this documentation.

Possibly - however the document added with that commit is listed as a first draft, and as it’s not accounted for on the index page it’s not rendering in the documentation. Practically speaking it makes no difference, however for the sake of accuracy I could revert that commit.

sev-:
Thank you, merging.

PR #2783, opened 2021-02-18
GUI: Increase allowed number of ExtraGuiOptions

OMGPizzaGuy:
Warn and limit options when adding more ExtraGuiOptions than supported. OptionsContainerWidget passes layout to the scroll container if used.


criezy:
With the changes to the themes you might need to increment the theme version (in `gui/ThemeEngine.h` and the `THEMERC` for each theme) so that it does not try to use an old theme if you happen to have one in your extra path.

SupSuper:
I wonder if it would make more sense to take these out of the theme so they could just be procedurally added like other GameOptionsDialogs, eg. https://github.com/scummvm/scummvm/blob/master/engines/sci/detection.cpp#L384

OMGPizzaGuy:
@SupSuper - I'll check that out. I originally attempted something like the RemapWidget but had trouble making good layout. That example looks better.

OMGPizzaGuy:
Procedurally generated layout works great!
I could see adding theme variables for the layout padding if desired - there's also the existing global padding variables, but they did not consistently match the dialog's padding

bluegr:
Very nice! Thanks for following @SupSuper's advice. The end result looks superb, and ticks all the right boxes:
- There is now no limit of engine-specific options
- The options are now in a scrollable container
- The hardcoded engine-specific options are now rightly out of the theme files

I had added the original comment in dialogs.cpp, and the advice given then about adding all the options in the theme files is now obsolete, thanks to procedurally generated options.

The end result is greatly simplified code, and a much more extendable custom options interface. I only see positive changes here, without any loss of functionality, so this is good to be merged now.

Many thanks for your work! Merging

bluegr:
I'll squash the commits, to keep the latest version of the code, cause the initial commits followed a different type of implementation.

PR #2784, opened 2021-02-18
DRAGONS: Incomplete attempt at improving big-endian compatibility

dwatteau:
Hi,

This is related to https://bugs.scummvm.org/ticket/11710.

DRAGONS has some important rendering issues with big-endian systems. I have a PowerBook G4 running Debian unstable on ppc32 (big-endian), so I tried having a look at this. (I'm doing my tests with `Blazing Dragons (Sony Playstation/French)`, for reference.)

This first commit is incomplete, and has been mostly done through naive trial-and-error, so see this more as an attempt at showing where the problem lies. Feel free to improve on it or to rewrite it completely in a better way :)

It seems that, in this engine, screen content must be written in native endianness here, not in little-endian. So, I've just introduced a naive `WRITE_SCREEN` macro that choses `WRITE_LE_UINT16` or `WRITE_BE_UINT16`, depending on your platform. Probably not the best way to handle this, but once again, let's suggest something and improve it.

The first commit I'm including in this PR appears to fix most scenes, cutscenes, and the bag.

Shadows, menus, and text are still wrong (I think I'm just missing something for the menus, and for the fonts/shadows it looks like some possible alpha blending problem). Audio hasn't been tested yet (Fluidsynth gives me some really strange initialisation errors, but it's maybe a problem with Debian unstable -- I'll look at this a bit later).

By the way, I also see these `(char)` casts in this engine:

```
sound.cpp:      second = (char)iVar3 + (char)(iVar3 / 10) * 6;
sound.cpp:      sector = (char)iVar2 + (char)(iVar2 / 10) * 6;
sound.cpp:      minute = (char)iVar1 + (char)(iVar1 / 10) * 6;
talk.cpp:                               talkDialogEntry->yPosMaybe = talkDialogEntry->yPosMaybe + (char)sVar2;
```

on PowerPC, `char` is unsigned by default, and I'm not sure the `char` here are used in a "pure character" way? Otherwise, we might want to use `signed char`, `int8` or `uint8` instead. But I don't really understand this part yet.

We can arrange for some more tests through Discord or whatever, if you like.

### DIFF
--- a/engines/dragons/bag.cpp
+++ b/engines/dragons/bag.cpp
@@ -71,7 +71,7 @@ void Bag::load(BigfileArchive *bigFileArchive) {
 		} else {
 			//c = (uint16)(((uint)c & 0x1f) << 10) | (uint16)(((uint)c & 0x7c00) >> 10) | c & 0x3e0;
 		}
-		WRITE_LE_UINT16(&pal[i * 2], c);
+		WRITE_SCREEN(&pal[i * 2], c);
 	}
 
 	stream.seek(0x308);
--- a/engines/dragons/font.cpp
+++ b/engines/dragons/font.cpp
@@ -160,7 +160,7 @@ Font *FontManager::loadFont(uint16 index, Common::SeekableReadStream &stream) {
 void updatePalEntry(uint16 *pal, uint16 index, uint16 newValue) {
 	newValue = (uint16)(((uint16)newValue & 0x1f) << 10) | (uint16)(((uint16)newValue & 0x7c00) >> 10) |
 			(newValue & 0x3e0) | (newValue & 0x8000);
-	WRITE_LE_INT16(pal + index, newValue);
+	WRITE_SCREEN(pal + index, newValue);
 }
 
 void FontManager::updatePalette() {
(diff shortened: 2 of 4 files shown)


bluegr:
So this definitely improves endianess issues without introducing any regressions, by the looks of it.
@dwatteau why is it still a draft after all this time? Can it be merged?

dwatteau:
@bluegr Because it's still incomplete (although still an improvement I believe), and not necessarily very clean (done through naive trial-and-error without understanding much about the engine itself or the general practices).

So I thought that it'd be better if a more experienced developper had a look at it, as a reference.

But I don't have any problem with removing the Draft statut and letting you merge it, if you prefer doing it this way.

bluegr:
The approach followed is correct. We've done similar endianess related defines in the tinsel engine.
So, this is a good way to go, and my proposal would be to remove the draft status and merge this

dwatteau:
OK, thank you @bluegr, I've removed the draft status, then 👍

bluegr:
Thanks! This does not introduce any regressions so it can be safely merged. Any further work on this can be done in-tree

PR #2786, opened 2021-02-19
AUDIO: Use SeekableSubReadStream instead of reallocating streams

SupSuper:
This cleans up audio code that mallocs raw audio data just to discard format headers, instead of just wrapping the stream. This reduces redundant reallocations since a lot of engines/decoders already do this anyways, and makes it easier to add WAV decoders since the extra headers are always discarded now.

One downside of this change is some engines have edge cases freeing sounds that are still being played (eg. closing the game mid-sound), which now crashes. I tried to fix the engines I ran across, but obviously I can't test every engine using raw WAVs, so I'd appreciate extra testing (and engine fixes) for this.

### DIFF
--- a/audio/decoders/mac_snd.cpp
+++ b/audio/decoders/mac_snd.cpp
@@ -29,6 +29,7 @@
 
 #include "common/textconsole.h"
 #include "common/stream.h"
+#include "common/substream.h"
 
 #include "audio/decoders/mac_snd.h"
 #include "audio/decoders/raw.h"
@@ -98,15 +99,10 @@ SeekableAudioStream *makeMacSndStream(Common::SeekableReadStream *stream,
 
 	stream->skip(soundDataOffset);
 
-	byte *data = (byte *)malloc(size);
-	assert(data);
-	stream->read(data, size);
-
-	if (disposeAfterUse == DisposeAfterUse::YES)
-		delete stream;
+	Common::SeekableReadStream *dataStream = new Common::SeekableSubReadStream(stream, stream->pos(), stream->pos() + size, disposeAfterUse);
 
 	// Since we allocated our own buffer for the data, we must specify DisposeAfterUse::YES.
-	return makeRawStream(data, size, rate, Audio::FLAG_UNSIGNED);
+	return makeRawStream(dataStream, rate, Audio::FLAG_UNSIGNED);
 }
 
 } // End of namespace Audio
--- a/audio/decoders/quicktime.cpp
+++ b/audio/decoders/quicktime.cpp
@@ -638,11 +638,7 @@ AudioStream *QuickTimeAudioDecoder::AudioSampleDesc::createAudioStream(Common::S
 			flags |= FLAG_STEREO;
 		if (_bitsPerSample == 16)
 			flags |= FLAG_16BITS;
-		uint32 dataSize = stream->size();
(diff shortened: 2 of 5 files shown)


mgerhardy:
It would maybe be a good idea to stop all sounds and music when the return-to-launcher-event is added to the queue - or is that too late already?

SupSuper:
@mgerhardy The base Engine destructor already stops the mixer, but by then it's too late. You could change it to happen before the destruction instead, or just let the mixer free the resources.

Since this is a rare edge case (most audio decoders assume the resources won't be freed while in use), I think it's better to fix any problematic engines than try to cover it up in the common code.

bluegr:
Thanks for your work on this! This is definitely much better than the previous codebase.
The edge cases that this change uncovered should be examined in-tree. Accessing freed memory on engine destruction is definitely a bad practice, and the solution is straightforward, as mentioned.

Merging, with the note that this should be tested for bad behavior on engine destruction. Any crashes upon engine exit should be fixed as mentioned.

sev-:
This caused regressions in mohawk: https://bugs.scummvm.org/ticket/12513 and in tucker engine.

Please revert.

eriktorbjorn:
The Bud Tucker regression seems to be simply that the stream it uses to load the WAV sound has been released by the time it tries to stream data from it. (It's stack-allocated, rather than heap-allocated.)

SupSuper:
Thanks for the testing @eriktorbjorn, I thought every engine was allocating their own audio, but looks like I missed a few. Just in case, I reviewed the code of all engines using makeWAVStream and fixed the following: MOHAWK, PRINCE and TUCKER. There should be no regressions, but I'd appreciate confirmation.

Or if you're feeling gutsy, here's every engine using WAVs: ACCESS, AGOS, AGS, ASYLUM, BURIED, CGE, CGE2, CRYOMNI3D, DIRECTOR, GLK, GNAP, HDB, HOPKINS, ICB, ILLUSIONS, MOHAWK, MYST3, NGI, PETKA, PLUMBERS, PINK, PRINCE, PRIVATE, SCI (32), SCUMM (HE), SHERLOCK, SLUDGE, SWORD1, SWORD2, TITANIC, TONY, TUCKER, NUVIE, ULTIMA4, ZVISION.

eriktorbjorn:
I can't reproduce the problems in Myst (Masterpiece Edition) and Bud Tucker any more, so that's promising.

Admittedly, my test case for Myst was simply "Start game, turn left, walk down the stairs, then walk back up." I've never actually played Bud Tucker (and I'm still mystified my copy calls itself "Bud Tucker in Real Trouble" instead of "Double Trouble"), so there it was just "start game and check that the intro doesn't crash".

eriktorbjorn:
However, I just noticed that the Prologue in the CD version of Freddy Pharkas (a.k.a. The Ballad of Freddy Pharkas) crashes after a few seconds now. I wonder if that's related. It _does_ call makeWAVStream() ...

SupSuper:
@eriktorbjorn Thanks, should be fixed now.

eriktorbjorn:
@SupSuper Thanks, it seems to be all good now. Even Valgrind didn't pick up any errors. (Well, not there at least. It did point to a small memory leak, but that was in a different part of the engine.)

PR #2787, opened 2021-02-19
SCUMM: Fix bad offset on game loading

orgads:
The iterator `l` was used inside the loop, which caused it to bail out
after 12 iterations instead of 16.

Amends 434d1f37f9398b6a7a1b08cd5464a1d5ecccdc41.

### DIFF
--- a/engines/scumm/imuse_digi/dimuse.cpp
+++ b/engines/scumm/imuse_digi/dimuse.cpp
@@ -184,16 +184,16 @@ void IMuseDigital::saveLoadEarly(Common::Serializer &s) {
 
 			if (_vm->_game.id == GID_CMI) {
 				if (track->soundId / 1000 == 1) { // State
-					for (l = 0; _comiStateMusicTable[l].soundId != -1; l++) {
-						if ((_comiStateMusicTable[l].soundId == track->soundId)) {
-							track->loopShiftType = _comiStateMusicTable[l].shiftLoop;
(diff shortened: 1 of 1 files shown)


orgads:
@AndywinXp @BLooperZ

AndywinXp:
Thanks! Luckily the bug didn't affect much the outcome since 0-7 index are the ones which are effectively used (unless some edge cases happen), but if not fixed, this bug would have created some headscratching problems in the long way; kudos for finding it! I'll update my FT pull request with this fix, as soon as this gets merged

orgads:
It actually did cause a strange problem. `_verbLineSpacing` got 0 on load, and dialogs with multiple lines had the second line overwrite the first one in COMI (at least on Hebrew).

AndywinXp:
> It actually did cause a strange problem. `_verbLineSpacing` got 0 on load, and dialogs with multiple lines had the second line overwrite the first one in COMI (at least on Hebrew).

Uh, that's odd, it didn't happen in both english and italian versions (which are the only copies I own), but I'm glad you found the issue, it could have potentially broken other versions too, thanks again!

bluegr:
Nice catch, thanks! Merging

orgads:
@AndywinXp Can you please have a look at https://bugs.scummvm.org/ticket/12188?

AndywinXp:
Sure, but let's move this discussion to discord

PR #2788, opened 2021-02-21
COMMON: Specialize unknownKeyError for more types

henke37:
Gives better error messages if someone is using `const char*` as the hashmap Key.


criezy (approved):
Thank you. Looks good.

PR #2789, opened 2021-02-21
GRAPHICS: Support having transparent ManagedSurface

criezy:
A change made yesterday forces `ManagedSurface` to be fully opaque. This may be an issue for some engines.
This commit revert this and also adds proper handling of the surface alpha when blitting pixels.

### DIFF
--- a/graphics/managed_surface.cpp
+++ b/graphics/managed_surface.cpp
@@ -251,8 +251,7 @@ void ManagedSurface::blitFromInner(const Surface &src, const Common::Rect &srcRe
 
 	uint destPixel;
 	byte rSrc, gSrc, bSrc, aSrc;
-	byte rDest = 0, gDest = 0, bDest = 0;
-	double alpha;
+	byte aDest = 0, rDest = 0, gDest = 0, bDest = 0;
 
 	if (!srcRect.isValidRect())
 		return;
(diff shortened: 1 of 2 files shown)


graphics/managed_surface.cpp
```diff
@@ -486,15 +484,25 @@ void transBlitPixel(TSRC srcVal, TDEST &destVal, const Graphics::PixelFormat &sr
...
 		rDest = static_cast<byte>((rSrc * alpha) + (rDest * (1.0 - alpha)));
 		gDest = static_cast<byte>((gSrc * alpha) + (gDest * (1.0 - alpha)));
 		bDest = static_cast<byte>((bSrc * alpha) + (bDest * (1.0 - alpha)));
+		
+		double sAlpha = (double)aSrc / 255.0;
+		double dAlpha = (double)aDest / 255.0;
+		double dAlphaRatio = dAlpha * (1.0 - sAlpha) / (sAlpha + dAlpha * (1.0 - sAlpha));
+		rDest = static_cast<uint8>(rSrc * sAlpha + rDest * dAlphaRatio);
+		gDest = static_cast<uint8>(gSrc * sAlpha + gDest * dAlphaRatio);
+		bDest = static_cast<uint8>(bSrc * sAlpha + bDest * dAlphaRatio);
```
OMGPizzaGuy, line 501: I think this is off due to parenthesis. This acts as:
bSrc * sAlpha + bDest * (dAlpha * (1.0 - sAlpha) / (sAlpha + dAlpha * (1.0 - sAlpha)))

instead of:
(bSrc * sAlpha + bDest * dAlpha * (1.0 - sAlpha)) / (sAlpha + dAlpha * (1.0 - sAlpha))

graphics/managed_surface.cpp
```diff
@@ -486,15 +484,25 @@ void transBlitPixel(TSRC srcVal, TDEST &destVal, const Graphics::PixelFormat &sr
...
 		gDest = gSrc;
 		bDest = bSrc;
+		aDest = 0xff;
 	} else {
 		// Partially transparent, so calculate new pixel colors
+		destFormat.colorToARGB(destVal, aDest, rDest, gDest, bDest);
 		double alpha = (double)aSrc / 255.0;
 		rDest = static_cast<byte>((rSrc * alpha) + (rDest * (1.0 - alpha)));
 		gDest = static_cast<byte>((gSrc * alpha) + (gDest * (1.0 - alpha)));
 		bDest = static_cast<byte>((bSrc * alpha) + (bDest * (1.0 - alpha)));
```
OMGPizzaGuy, line 494: Looks like these 4 lines are being replaced by what's below

criezy: Yes, totally, I kept then in while writing the new ones, and then forgot to remove them.

OMGPizzaGuy:
ManagedSurface::blitFromInner also blends, but it I think it's safe to assume it should always end with 0xFF as alpha there.

graphics/managed_surface.cpp
criezy: And this is now fixed.

criezy:
> ManagedSurface::blitFromInner also blends, but it I think it's safe to assume it should always end with 0xFF as alpha there.

I think it might be better to also handle the destination transparency there. I will have a look.

graphics/managed_surface.cpp
criezy: Yes, you are right. I got my equations wrong here. I push a change, and hopefully they are correct now.

criezy:
I had a quick look at `ManagedSurface::blitFromInner` and at the very least it is not consistent between the case where the source and destination formats are the same and the generic case. In the former case it just copies the source onto the destination with no blending, which is probably incorrect if the source is partially transparent. So I think I will need to do some changes here anyway.

OMGPizzaGuy (approved):

criezy:
I have now fixed the obviously wrong alpha blending in `ManagedSurface::blitFromInner` and decided to add support for transparent destination there.

I have played through a few games in three different engines that use `ManagedSurface` (AGS, Glk, and Cryomni3d) and did not see any regression.

graphics/managed_surface.cpp
```diff
@@ -272,68 +271,82 @@ void ManagedSurface::blitFromInner(const Surface &src, const Common::Rect &srcRe
...
+					uint32 destColor = *(uint32 *)destVal;
+					format.colorToARGB(destColor, aDest, rDest, gDest, bDest);
 				} else {
-					// Partially transparent, so calculate new pixel colors
-					alpha = (double)aSrc / 255.0;
-					rDest = static_cast<byte>((rSrc * alpha) + (rDest * (1.0 - alpha)));
-					gDest = static_cast<byte>((gSrc * alpha) + (gDest * (1.0 - alpha)));
-					bDest = static_cast<byte>((bSrc * alpha) + (bDest * (1.0 - alpha)));
+					aDest = 0xFF;
+					rDest = destVal[0] = rDest;
```
OMGPizzaGuy, line 327: This line seems strange - intentional?

criezy: No. A copy-paste gone wrong...

criezy:
Thank you for the review and for pointing out my stupid mistakes.
I will merge this now and cross my fingers that it does not introduce regressions..

PR #2790, opened 2021-02-21
GRIM: Fix unknown key error when loading savegame

criezy:
A user reported a crash with `Unknown key!` error when trying to load a save game in MI4 today on IRC/Discord.
@henke37 suggested that the likely culprit was the code changed in this pull request.

I don't have the game, and the demo doesn't let me save games. So I am unable to test this. This is why I am proposing this as a pull request so that developers who know the engine or can at least test it can validate the change.

### DIFF
--- a/engines/grim/pool.h
+++ b/engines/grim/pool.h
@@ -314,9 +314,10 @@ void PoolObject<T>::Pool::restoreObjects(SaveGame *state) {
 	Common::HashMap<int32, T *> tempMap;
 	for (int32 i = 0; i < size; ++i) {
 		int32 id = state->readLESint32();
-		T *t = _map.getVal(id);
-		_map.erase(id);
-		if (!t) {
+		T *t = nullptr;
+		if (_map.tryGetVal(id, t))
+			_map.erase(id);
+		else {
 			t = new T();
 			t->setId(id);
 		}


engines/grim/pool.h
```diff
@@ -314,8 +314,9 @@ void PoolObject<T>::Pool::restoreObjects(SaveGame *state) {
 	Common::HashMap<int32, T *> tempMap;
 	for (int32 i = 0; i < size; ++i) {
 		int32 id = state->readLESint32();
-		T *t = _map.getVal(id);
-		_map.erase(id);
+		T *t = nullptr;
+		if (_map.tryGetVal(id, t))
+			_map.erase(id);
 		if (!t) {
```
henke37, line 320: I would've probably reused the existing if statement. It's clear from context that it's the same condition.

criezy: I didn't know if I could assume that `_map` does not contain null pointers. But if that is a safe assumption we could indeed use an `else` statement here.

henke37: It calls a member function only a few lines bellow, no null pointers could've been stored legitimately.

engines/grim/pool.h
```diff
@@ -314,8 +314,9 @@ void PoolObject<T>::Pool::restoreObjects(SaveGame *state) {
 	Common::HashMap<int32, T *> tempMap;
 	for (int32 i = 0; i < size; ++i) {
 		int32 id = state->readLESint32();
-		T *t = _map.getVal(id);
-		_map.erase(id);
+		T *t = nullptr;
```
aquadran, line 317: maybe?
T *t = _map.getValOrDefault(id, nullptr);

engines/grim/pool.h
aquadran: not sure about save game and I'm not sure about this logic here. I didn't add this pool template.
I agree with henke37 assumption that map does not have null pointers. Also I agree with use 'else'.

criezy: I think the map indeed never contains null pointers. Thus I have changed the code to use `else`, which is a bit clearer.

criezy:
I forgot about this pull request and I will merge it now. I think the change is pretty safe.

PR #2791, opened 2021-02-21
SCUMM: SMUSH: always enable line wrap on hebrew fan translation

BLooperZ:
In Hebrew there are no distinction between capital/small letters,
which results in wider text that although originally was able to fit in screen width, usually go beyond the width of the screen in translation.
This may be the easier way to solve this and it doesn't seem to have negative side effect. 
(the other way is to change the text flag in all animation files)

Thanks

### DIFF
--- a/engines/scumm/smush/smush_player.cpp
+++ b/engines/scumm/smush/smush_player.cpp
@@ -641,7 +641,7 @@ void SmushPlayer::handleTextResource(uint32 subType, int32 subSize, Common::Seek
 	// bit 7 - skip ^ codes (COMI)     0x80        (should be irrelevant for Smush, we strip these commands anyway)
 	// bit 8 - no vertical fix (COMI)  0x100       (COMI handles this in the printing method, but I haven't seen a case where it is used)
 
-	if (flags & 4) {
+	if ((flags & 4) || _vm->_language == Common::HE_ISR) {
 		// COMI has to do it all a bit different, of course. SCUMM7 games immediately render the text from here and actually use the clipping data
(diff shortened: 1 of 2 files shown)


engines/scumm/smush/smush_player.cpp
```diff
@@ -641,7 +641,7 @@ void SmushPlayer::handleTextResource(uint32 subType, int32 subSize, Common::Seek
 	// bit 7 - skip ^ codes (COMI)     0x80        (should be irrelevant for Smush, we strip these commands anyway)
 	// bit 8 - no vertical fix (COMI)  0x100       (COMI handles this in the printing method, but I haven't seen a case where it is used)
 
-	if (flags & 4) {
+	if (flags & 4 || _vm->_language == Common::HE_ISR) {
```
bluegr, line 644: Please use parentheses for brevity. Yes, the bitwise operator takes precedence over the logical operator, but parentheses make this clearer

```suggestion
	if ((flags & 4) || _vm->_language == Common::HE_ISR) {
```

BLooperZ: Thank you, fixed

athrxx:
I am not familiar with the Hebrew fan translation, but maybe you need the same modification in insane.cpp, line 1305, too. Please check...

BLooperZ:
Thanks, will check that.

sev-:
@BLooperZ any news?

BLooperZ:
@sev- sorry, haven't had time to check it yet, I hope to check during this week.

BLooperZ:
As suggested, added to `insane.cpp` as well, thanks.

sev-:
Thanks! merging

PR #2793, opened 2021-02-22
SYMBIAN: automatically build sources from main tree dirs: audio, base…

fedor4ever:
…, common, etc except engines.

Enhance build logging - they differs by datetime now.
Build test exe which can coexist with release.
Correct parsing module.mk for DETECT_OBJS.
Appends to existed engines.mmh and macros.mmh, doesn't replace it.


digitall:
Since this PR only affects the Symbian port which @fedor4ever is maintaining, I think this can be merged as-is as no objections after a week.

PR #2794, opened 2021-02-22
PEGASUS: fix bug 11758.

fedor4ever:
See - https://bugs.scummvm.org/ticket/11758
PEGASUS: colossal memory leak, wan't start.

### DIFF
--- a/common/macresman.cpp
+++ b/common/macresman.cpp
@@ -115,10 +115,6 @@ String MacResManager::computeResForkMD5AsString(uint32 length) const {
 	return computeStreamMD5AsString(resForkStream, MIN<uint32>(length, _resForkSize));
 }
 
-bool MacResManager::open(const String &fileName) {
-	return open(fileName, SearchMan);
-}
-
 bool MacResManager::open(const String &fileName, Archive &archive) {
 	close();
 
@@ -303,7 +299,6 @@ bool MacResManager::loadFromAppleDouble(SeekableReadStream &stream) {
 
 bool MacResManager::isMacBinary(SeekableReadStream &stream) {
 	byte infoHeader[MBI_INFOHDR];
-	int resForkOffset = -1;
 
 	if (stream.read(infoHeader, MBI_INFOHDR) != MBI_INFOHDR)
 		return false;
@@ -320,15 +315,10 @@ bool MacResManager::isMacBinary(SeekableReadStream &stream) {
 		//uint32 rsrcSizePad = (((rsrcSize + 127) >> 7) << 7);
 
 		// Length check
-		if (MBI_INFOHDR + dataSizePad + rsrcSize <= (uint32)stream.size()) {
-			resForkOffset = MBI_INFOHDR + dataSizePad;
-		}
+		if (MBI_INFOHDR + dataSizePad + rsrcSize == (uint32)stream.size())
+			return true;
 	}
-
-	if (resForkOffset < 0)
-		return false;
-
-	return true;
+	return false;
 }
 
 bool MacResManager::isRawFork(SeekableReadStream &stream) {
@@ -357,10 +347,10 @@ bool MacResManager::loadFromMacBinary(SeekableReadStream &stream) {
 
 		uint32 dataSizePad = (((dataSize + 127) >> 7) << 7);
 		// Files produced by ISOBuster are not padded, thus, compare with the actual size
-		//uint32 rsrcSizePad = (((rsrcSize + 127) >> 7) << 7);
+		uint32 rsrcSizePad = (((rsrcSize + 127) >> 7) << 7);
 
 		// Length check
-		if (MBI_INFOHDR + dataSizePad + rsrcSize <= (uint32)stream.size()) {
+		if (MBI_INFOHDR + dataSizePad + rsrcSizePad <= (uint32)stream.size()) {
 			_resForkOffset = MBI_INFOHDR + dataSizePad;
 			_resForkSize = rsrcSize;
 		}
--- a/common/macresman.h
+++ b/common/macresman.h
@@ -72,20 +72,11 @@ public:
 	 * from inside an engine.
 	 *
 	 * @param fileName The base file name of the file
+	 * @param archive The path that holds the forks
 	 * @note This will check for the raw resource fork, MacBinary, and AppleDouble formats.
 	 * @return True on success
 	 */
-	bool open(const String &fileName);
-
-	/**
-	 * Open a Mac data/resource fork pair from within the given archive.
-	 *
-	 * @param path The path that holds the forks
-	 * @param fileName The base file name of the file
-	 * @note This will check for the raw resource fork, MacBinary, and AppleDouble formats.
-	 * @return True on success
-	 */
-	bool open(const String &fileName, Archive &archive);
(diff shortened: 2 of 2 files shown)


sluicebox:
Could you rename this commit to conform to the guidelines? https://wiki.scummvm.org/index.php/Commit_Guidelines

The prefix should be COMMON since this doesn't touch the Pegasus engine. The summary should describe the change as opposed to just a bug number.

I'm confused if this reverts 1bb73862a10ab93ad884c9dec2908dd092cbc21e or not. Do ISOBuster files still work? That commit commented out a line and added the comment "Files produced by ISOBuster are not padded, thus, compare with the actual size", this PR restores the line but leaves that comment, and that doesn't sound right.

My interest is so that our tester who uses ISOBuster can still test my Mac stuff =)

common/macresman.cpp
```diff
@@ -320,15 +315,10 @@ bool MacResManager::isMacBinary(SeekableReadStream &stream) {
 		//uint32 rsrcSizePad = (((rsrcSize + 127) >> 7) << 7);
 
 		// Length check
-		if (MBI_INFOHDR + dataSizePad + rsrcSize <= (uint32)stream.size()) {
-			resForkOffset = MBI_INFOHDR + dataSizePad;
-		}
+		if (MBI_INFOHDR + dataSizePad + rsrcSize == (uint32)stream.size())
```
criezy, line 318: That test is incorrect because with that change only macbinary files created with ISOBuster will (or at least might) work.

A test that might work for both ISOBuster and standard macbinary files would be:
```c++
if (MBI_INFOHDR + dataSizePad + rsrcSize == (uint32)stream.size() || MBI_INFOHDR + dataSizePad + rsrcSizePad == (uint32)stream.size())
```

common/macresman.cpp
```diff
@@ -357,10 +347,10 @@ bool MacResManager::loadFromMacBinary(SeekableReadStream &stream) {
 
 		uint32 dataSizePad = (((dataSize + 127) >> 7) << 7);
 		// Files produced by ISOBuster are not padded, thus, compare with the actual size
-		//uint32 rsrcSizePad = (((rsrcSize + 127) >> 7) << 7);
+		uint32 rsrcSizePad = (((rsrcSize + 127) >> 7) << 7);
 
 		// Length check
-		if (MBI_INFOHDR + dataSizePad + rsrcSize <= (uint32)stream.size()) {
+		if (MBI_INFOHDR + dataSizePad + rsrcSizePad <= (uint32)stream.size()) {
```
criezy, line 353: And that change is also incorrect because it means that macbinary created by ISOBuster will fail to load.
Combined with the previous issue I pointed above it means that the commit as it currently is will likely break all macbinary files. The check here should be the same check as in `isMacBinary()`.

criezy (changes requested):
It seems to me that your change will completely break reading macbinary files (created by ISOBuster or not).

The cleaning in the code is good. But the logic for the two checks in `isMacBinary()` and `loadMacBinary()` needs to be changed (and also they are currently not consistent).

criezy:
I have also requested a review from @sev- 
Since I am not familiar with ISOBuster and he  wrote the code, it might be good for him to confirm that:
`MBI_INFOHDR + dataSizePad + rsrcSize == (uint32)stream.size()` is true for ISOBuster macbinary and thus that the following condition should work in all macbinary cases.
```c++
if (MBI_INFOHDR + dataSizePad + rsrcSize == (uint32)stream.size() || MBI_INFOHDR + dataSizePad + rsrcSizePad == (uint32)stream.size())
```

fedor4ever:
Can you trace with debugger pegasus engine? Maybe we find better solution.

fedor4ever:
Well, I try to write small testcase because it hard to undersatnd whats goin on.

Found suspicious code:
bool QuickTimeDecoder::loadFile(const Common::String &filename) {
	if (!Common::QuickTimeParser::parseFile(filename))
		return false;

	init();
	return true;
}

Compiles and runs but:

int main()
{
    for(auto x: data_files){
        cout << x << endl;
        initFromMovieFile(x);
    }
    return 0;
}

void initFromMovieFile(const Common::String &fileName)
{
	if (!Common::QuickTimeParser::parseFile(fileName)) { // <<< cannot call member function without object
		// Replace any colon with an underscore, since only Mac OS X
		// supports that. See PegasusEngine::detectOpeningClosingDirectory()
		// for more info.
		Common::String newName(fileName);
		if (newName.contains(':'))
			for (uint i = 0; i < newName.size(); i++)
				if (newName[i] == ':')
					newName.setChar('_', i);

		if (!Common::QuickTimeParser::parseFile(newName))
			error("Could not load video '%s'", fileName.c_str());
	}
};

fedor4ever:
I was surprised there no tests. How this project doesn't fall apart?

digitall:
@fedor4ever: Because people fix it... and I think you are misreading that code hence why it is failing for you in your modified version.

digitall:
```
bool QuickTimeDecoder::loadFile(const Common::String &filename) {
	if (!Common::QuickTimeParser::parseFile(filename))
		return false;
```
looks like a call to a static method... but that is not a static class method: https://github.com/scummvm/scummvm/blob/master/common/quicktime.cpp#L62 which would be a bug... but should be one that the compiler would warn or error on... so why does that work?

Because it is not a call to a static class method, but to a superclass method... QuickTimeDecoder inherits several superclasses:
https://github.com/scummvm/scummvm/blob/master/video/qt_decoder.h#L61

The critical one here is QuickTimeAudioDecoder which inherits from QuickTimeParser hence why calling that method from the superclass is valid:
https://github.com/scummvm/scummvm/blob/master/audio/decoders/quicktime_intern.h#L48

digitall:
Looking at the original code which calls this, it should be fine assuming I am correct in my C++ reading:
https://github.com/scummvm/scummvm/blob/master/engines/pegasus/movie.cpp#L62

fedor4ever:
> Looking at the original code which calls this, it should be fine assuming I am correct in my C++ reading:
> https://github.com/scummvm/scummvm/blob/master/engines/pegasus/movie.cpp#L62

Yes, you correct. I read as call to static method and was wrong.

digitall:
@fedor4ever : No problem. I thought the same thing when I first read it... but no compiler warning / error made me suspect I was missing something.

common/macresman.h
```diff
@@ -72,20 +72,11 @@ class MacResManager {
...
-	/**
-	 * Open a Mac data/resource fork pair from within the given archive.
-	 *
-	 * @param path The path that holds the forks
-	 * @param fileName The base file name of the file
-	 * @note This will check for the raw resource fork, MacBinary, and AppleDouble formats.
-	 * @return True on success
-	 */
-	bool open(const String &fileName, Archive &archive);
+	bool open(const String &fileName, Archive &archive = SearchMan);
```
sev-, line 79: What is the purpose of this change?

common/macresman.cpp
sev-: Indeed, this just breaks things. What is the reasoning behind this change?

sev-:
This PR is a whole dirty hack without proper testing performed and no understanding of the code.

Closing

PR #2795, opened 2021-02-22
SYMBIAN: fix conflict declaration for TKey struct between Lua and SDK…

fedor4ever:
… headers.

Add missed prototypes for math functions.

### DIFF
--- a/backends/platform/symbian/src/portdefs.h
+++ b/backends/platform/symbian/src/portdefs.h
@@ -37,7 +37,9 @@
 #pragma GCC diagnostic ignored "-Wnarrowing"
 #endif
 #include <e32def.h>
-#include <e32std.h>
+#if !defined(__IGNORE__E32STD_H__) // TKey type from system header 
+#include <e32std.h> // doesn't meets with lua ones.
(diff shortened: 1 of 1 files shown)


backends/platform/symbian/src/portdefs.h
```diff
@@ -86,11 +88,13 @@ namespace std
...
 #endif
 
-// TODO: Replace this with the keymapper
-#define GUI_ENABLE_KEYSDIALOG
-
 #define DISABLE_COMMAND_LINE
 #define USE_RGB_COLOR
+#define USE_TINYGL
+
+#ifndef SYMBIAN_DYMANIC_PLUGIN
```
digitall, line 95: Spelling mistake here I think ... Should be #ifndef SYMBIAN_DYNAMIC_PLUGIN ?

fedor4ever:
@digitall, thank you very much! Done!

digitall:
OK. Since this PR only affects the Symbian port which @fedor4ever is maintaining, I think this can be merged as-is as no objections after a week.

PR #2797, opened 2021-02-22
ULTIMA: SYMBIAN: exclude unsupported engines.

fedor4ever:


sev-:
No, this is horribly wrong. Define an engine feature and then guard against it.

What is the reason that these games are excluded on Symbian?

fedor4ever:
With these games ScummVM doesn't start on my Nokia 5800. Debugger can't help. I don't know how handle that.

sev-:
Then my recommendation is either split the engine properly (talk to the engine devs) or not compile it at all.

The proposed solution is not acceptable.

PR #2798, opened 2021-02-23
AGS: Fix text rendering onto surfaces with a transparent pixel color

dreammaster:
This pull request resolves the outstanding problem with rendering TTF text onto surfaces in the AGS engine.. the problem being that with the surfaces having a designated transparent color, partially transparent pixels in the TTF glyphs were doing an alpha blend of the font color with the pink color AGS uses for transparency. Since this affects the core font code, I figure to make it a pull request for review first.

This change introduces a new virtual method on Graphics::Font called "setTransparentColor", and then overrides that in TTFFont.  In the TTFFont glyph drawing code, if it's drawing a partially transparent pixel and detects the transparent color, it forces it to be transparent so it won't be blended.

### DIFF
--- a/graphics/fonts/ttf.cpp
+++ b/graphics/fonts/ttf.cpp
@@ -30,6 +30,7 @@
 #include "graphics/fonts/ttf.h"
 #include "graphics/font.h"
 #include "graphics/surface.h"
+#include "graphics/managed_surface.h"
 
 #include "common/ustr.h"
 #include "common/file.h"
@@ -157,6 +158,8 @@ public:
 	virtual Common::Rect getBoundingBox(uint32 chr) const;
 
 	virtual void drawChar(Surface *dst, uint32 chr, int x, int y, uint32 color) const;
+	virtual void drawChar(ManagedSurface *dst, uint32 chr, int x, int y, uint32 color) const;
+
 private:
 	bool _initialized;
 	FT_Face _face;
@@ -185,6 +188,8 @@ private:
 	int computePointSize(int size, TTFSizeMode sizeMode) const;
 	int readPointSizeFromVDMXTable(int height) const;
 	int computePointSizeFromHeaders(int height) const;
+	void drawChar(Surface *dst, uint32 chr, int x, int y, uint32 color,
+		const uint32 *transparentColor) const;
 
 	FT_Int32 _loadFlags;
 	FT_Render_Mode _renderMode;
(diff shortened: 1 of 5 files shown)


graphics/fonts/ttf.cpp
```diff
@@ -597,7 +608,11 @@ static void renderGlyph(uint8 *dstPos, const int dstPitch, const uint8 *srcPos,
 				sA = *src;
 
 				uint8 dA, dR, dG, dB;
-				dstFormat.colorToARGB(*rDst, dA, dR, dG, dB);
```
lephilousophe, line 600: Shouldn't this be handled in colorToARGB? This would prevent these changes.

OMGPizzaGuy: Are you suggesting PixelFormat know the transparent color? It might make conceptual sense, but I'd be concerned about performance if always checked there.

engines/ags/lib/alfont/alfont.cpp
```diff
@@ -60,7 +60,11 @@ size_t alfont_text_height(ALFONT_FONT *font) {
 
 void alfont_textout(BITMAP *bmp, ALFONT_FONT *font, const char *text, int x, int y, uint32 color) {
 	Graphics::ManagedSurface &surf = **bmp;
-	font->getFont()->drawString(&surf, text, x, y, bmp->w - x, color);
+	Graphics::Font *fnt = font->getFont();
+
+	if (surf.hasTransparentColor())
+		fnt->setTransparentColor(surf.getTransparentColor());
+	fnt->drawString(&surf, text, x, y, bmp->w - x, color);
```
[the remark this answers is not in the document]
dreammaster: That's actually an excellent idea. It would be cleaner than having an explicit setTransparentColor call. Graphics::Font can have a version that simply passes on the internal surface to the existing version, and the TTFFont can overload it to get the transparent color. I'll rework the PR tonight.

dreammaster:
I'd completely forgotten that Font already had a method version that takes in a ManagedSurface. Just internally it simply called the Surface version. So I did a minor rejigging so that the ManagedSurface drawChar version is a virtual method too now, and TTFFont overrides it to get access to the transparent color, if any.

At the same time, I finally got around to moving ManagedSurface's addDirtyRect to be public, and added a method to get access to a ManagedSurface's internal surface if a calling class really needs it. This let me get rid of the "friend class Font" that was previously required just so the Font class could get access to the surface and modify it to write the characters, and then call addDirtyRect itself. So it's all cleaner now.

OMGPizzaGuy:
This looks good to me, but I'd feel more comfortable with someone else approving.
One nitpick is the name of this PR as it is more specific to GRAPHICS than AGS at this point

PR #2800, opened 2021-02-23
SCUMM: Fix verb selection on overlapping verbs

orgads:
On COMI, verbs can vertically overlap. For example, verb 1 is 330-360
and verb 2 is 354-384.

Highlighting is done by the first match, but clicking chooses the last
one, as it is scanned backwards.

So if the mouse is in range 354-360, the highlighted text is not picked.

### DIFF
--- a/engines/scumm/verbs.cpp
+++ b/engines/scumm/verbs.cpp
@@ -952,11 +952,8 @@ int ScummEngine::findVerbAtPos(int x, int y) const {
 	if (!_numVerbs)
 		return 0;
 
-	VerbSlot *vs;
-	int i = _numVerbs - 1;
-
-	vs = &_verbs[i];
-	do {
+	for (int i = 1; i < _numVerbs; ++i) {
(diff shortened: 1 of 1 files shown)


orgads:
I have to admit I couldn't find where the highlighting part is different (looks like it also uses `findVerbAtPos`), but this behavior is consistent.

orgads:
cc @BLooperZ

orgads:
Oops, should start from 1. I'll fix it tomorrow.

Do you have other comments?

bluegr:
``findVerbAtPos()`` is a function used by several SCUMM versions. Have you tested this change with games other than COMI?

orgads:
Yes, I tested with Monkey Island 1 too.

orgads:
The bug only occurs when setting the game language to Hebrew.

orgads:
Superseded by https://github.com/scummvm/scummvm/pull/2817

PR #2803, opened 2021-02-27
STARK: Update detection.cpp

RamizGindullin:
Fargus version wasn't distributed by 1C. Instead, it was done by the Snowball  (now defunct, link: https://www.mobygames.com/game/longest-journey/cover-art/gameCoverId,105165/)

### DIFF
--- a/engines/stark/detection.cpp
+++ b/engines/stark/detection.cpp
@@ -275,7 +275,7 @@ static const ADGameDescription gameDescriptions[] = {
 	// The Longest Journey
 	// Russian 2CD by 1C
 	{
-		"tlj", "2 CD/Fargus",
+		"tlj", "2 CD/Snowball", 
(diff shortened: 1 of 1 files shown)


mgerhardy:
Thanks for the contribution - but the commit message should contain the engine prefix "STARK: your message in here" (like I did it for the pull request title)

RamizGindullin:
Ah, sorry. First time I done it

Should I create a new commit with a correct name and resubmit it?

criezy:
Since you only need to edit the message for the last commit, you can do it with `git --amend` and then force push to your fork with `git push --force`. If you are not using command line, search how to do that with your client.

Note: If you needed to edit the message for more than one commit, or a commit that is not the last one, you would need to use interactive rebase instead of amend.

PR #2804, opened 2021-02-27
RISCOS: Fix internal compiler error in VFP build

lephilousophe:


lephilousophe:
Merging now. If a better solution is found it can be easily reverted.

PR #2805, opened 2021-02-28
JANITORIAL: Update old bug tracker numbers

sluicebox:
I wrote a thing the other day to auto-update all the old bug numbers in comments in the SCI engine (9982c761a2e5e9bd9129bbeb78c0b749b65baaa4) because I had to look them up a lot. Coincidentally, someone just asked how to do this on discord, so now I've run it on the whole codebase. They're yours if you want them!

Sourceforce has http redirects for urls with the old bug numbers to the new ones. That was a simple (and quick!) way to automate lookups. As with SCI, I spot checked a bunch to make sure it worked, and I grepped and eyeballed each line afterwards to make sure it really was a comment about a bug.

For the record, I don't even know the name of the bug tracker that produced these numbers ten years ago, so don't ask me hard questions. =)


bluegr:
You sir are a hero for doing this!
Many thanks for updating all of the bug IDs in the codebase. Merging

PR #2809, opened 2021-03-03
DIRECTOR: implement colorQD builtin

yegortimoshenko:
Normally, `colorQD` is a `TheEntity`. However, "Eastern Mind: The Lost Souls of Tong-Nou" instead uses it as a builtin with no arguments, both in Mac and Windows versions of the game.

cc @sev-


sev-:
Thanks

PR #2811, opened 2021-03-04
DOCS: adds version number to logo

cadihowley:
Going forward, I propose that we add the version number to the ScummVM logo.

As discussed, due to limitations with Read the Docs, the default documentation will simply be referred to as "stable" (or at the moment, "latest") - which will not mean much to most users.

By adding the version to the logo, it is visually obvious which version the documentation relates to.

Logo with version number


mgerhardy:
Should the version maybe just be a string that is automatically added on publishing the website with the docs? I suppose maintaining an image with the version in it is of course doable - but also unnecessary if we do it in a more automatic fashion.

Mataniko:
I don't think we'd want to manually update the image with each release.

cadihowley:
Fair enough - this just seemed the easiest for me (non-developer haha) to implement in the meantime... 

We could use some JS to change the existing “latest” and “stable” text underneath the logo, it’s reasonably easy to add JS to RTD. Seems like it might be easier to do that than to change the RTD theme, which I actually have no idea how to do!

With that we could change “latest” to something like “development build” (or something a little more catchy...? Or even leave it as is...). “Stable” could be changed to actual version number (this is the only part I can’t figure out how to automate, since I’m not sure where to pull the information from, but this would be very easy to do manually). And older versions (which will have a “vx.x-docs” tags) can be changed to just the version number as well. 

There may be more elegant ways to do this, but this pulls the info from the URL and then changes it:

```
    document.addEventListener("DOMContentLoaded", function(){
    var stable = "v2.2"; //Change this to the latest release version
  
    var url = window.location.href;
    var start = url.indexOf("/", 37);
    var end = url.indexOf("/", start+1);
    var endMinus = url.indexOf("-", start+1);
    var versionNumber = url.substring(start+1, endMinus);  
    var found = url.substring(start+1, end);
    var version = null;
    
    switch(found) {
      case 'latest':
        version = "development build"
        break;
      case 'stable':
        version = stable
        break;
      default:
        version = versionNumber
        break;
    }
});
```

 @mgerhardy What exactly did you have in mind in terms of automating the process?

PR #2812, opened 2021-03-04
KEYMAPPER: executeAction - keep original ascii value

ZvikaZ:
If user presses SHIFT+. , it was interpreted as a dot.
That's because there is a `kKeymapMatchPartial` with
"Skip line" action, which caused execution of `executeAction`.
And `executeActio`n replaced the original `incomingEvent`
with `action`, thus 'forgetting' the original ascii code.

The problem was seen (and fixed) on PQ1 (AGI) and SQ3 (SCI).

**NOTE**
I'm not certain that it's a good fix, as I don't fully understand `executeAction`, and all the implications of my change. It's only a single line change, but please review it carefully...

### DIFF
--- a/backends/keymapper/keymapper.cpp
+++ b/backends/keymapper/keymapper.cpp
@@ -275,6 +275,7 @@ Keymapper::IncomingEventType Keymapper::convertToIncomingEventType(const Event &
 
 Event Keymapper::executeAction(const Action *action, const Event &incomingEvent) {
 	Event outgoingEvent = Event(action->event);
+	outgoingEvent.kbd.ascii = incomingEvent.kbd.ascii;
 
 	IncomingEventType incomingType = convertToIncomingEventType(incomingEvent);
 


sev-:
Wouldn't that void potential key remapping when the ASCII value truly differs?

ZvikaZ:
Yes, you're right (if I understand you correctly...)

I have checked SOMI before this PR, and with this PR. Before the PR, when pressing `SHIFT+.`, it skips lines in dialog. With the PR, it doesn't. (of course, just `.` does skip lines). If it's bad - then this PR is wrong...

Another possibility - currently, the `Skip line` action is added for all engines (in `engines\metaengine.cpp`). Maybe it should be added only for engines supporting it? (SCUMM, and what else? I don't know. Grepping for `Skip line` or `SKLI` doesn't help)

ZvikaZ:
I have opened a bug for the issue this PR tries to solve: https://bugs.scummvm.org/ticket/12410

sev-:
I think the proper way to address this is to check the modifier key bitfield. This PR is not a correct approach and would hide the problem, unfortunately. Closing.

PR #2818, opened 2021-03-07
DOCS: Fix link to the iOS Libraries

dcordero:
Just fixing a link that I found broken while following the iOS building documentation


mduggan:
Thanks for fixing!

PR #2820, opened 2021-03-07
JANITORIAL: Replace new[]/memset with new[]()

orgads:
Mostly done using the following Ruby script:

```Ruby
(Dir.glob('**/*.cpp') + Dir.glob('**/*.h')).each do |file|
  s = File.read(file, encoding: 'iso8859-1')
  t = s.gsub(/(([\w_.\[\]]+)\s*=\s*new\s+\S+?\[[^\]]+?\](?!\())([^\{\}]*?)\n\s+memset\(\s*\2\s*,\s*0\s*,[^;]*;/m, '\1()\3')
  if t != s
    File.open(file, 'w') { |io| io.write(t) }
  end
end
```

### DIFF
--- a/audio/decoders/wma.cpp
+++ b/audio/decoders/wma.cpp
@@ -590,9 +590,8 @@ Common::SeekableReadStream *WMACodec::decodeSuperFrame(Common::SeekableReadStrea
 
 		// PCM output data
 		outputDataSize = frameCount * _channels * _frameLen;
-		outputData     = new int16[outputDataSize];
+		outputData     = new int16[outputDataSize]();
 
-		memset(outputData, 0, outputDataSize * 2);
 
 		// Number of bits data that completes the last superframe's overhang.
 		int bitOffset = bits.getBits(_byteOffsetBits + 3);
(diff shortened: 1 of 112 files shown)


lephilousophe:
This is a C++03 construct while ScummVM is restricted to a C++98 subset according to the Wiki [rules](https://wiki.scummvm.org/index.php?title=Coding_Conventions)

orgads:
I see that it's already used in ags, bladerunner, dm, scumm and wintermute. Should these be also changed to use memset instead?

Search pattern: `git grep new.*\]\(\)`

digitall:
@orgads : Ignoring for a second the merits and disadvantages of this change, I would highly recommend splitting this into a commit per engine and subsystem i.e. "JANITORIAL: AGOS: Replace new[]/memset with new[]()" etc. as this makes it easier to review the changes for engine authors and makes future bisection for any regressions a bit less problematic i.e. too many changes in a single commit which you have to split apart to locate the hunk causing the issue.

digitall:
If the changes were related i.e. in a Common API change, then this would need to be a singular commit to avoid unbuildable commits and propagate the change, but these changes are not inter-related and could be split.

orgads:
I agree. I considered splitting it, but didn't know what will you prefer. Anyway, since it is likely to break older compilers, I'll close it. Thanks.

sev-:
Since we're now C++11, could you please split the commit and then we merge?

orgads:
I split it, but I only created separate commits for engines with many changes. All the other engines are packed in a single commit.

It this ok?

sev-:
Yes, thank you!

PR #2823, opened 2021-03-08
README: Update link to MinGW build instructions

orgads:


orgads:
Is there a reason MSYS2 is not officially supported, and the recommended way to build on Windows/MinGW is with MSYS1?

aquadran:
MSYS2 package support is/should be handled on MSYS2 side only.
https://packages.msys2.org/package/mingw-w64-x86_64-scummvm?repo=mingw64

orgads:
If you look at the [wiki page](https://wiki.scummvm.org/index.php/Compiling_ScummVM/MinGW-w64), there is a big red note stating that it is not officially supported. And also on the main [Compiling wiki](https://wiki.scummvm.org/index.php?title=Compiling_ScummVM) links to the old MinGW instructions, with a note: See also unofficial MSYS2/MinGW-w64 instructions.

SupSuper:
I think @lotharsm wrote the current instructions so he might know more.

lotharsm:
I wrote the msys2 instructions, that's correct. While it's still considered as "unsupported" (as in "it works, but I only use it for development purposes and not for the release builds), it works very well. I'm not sure about the old MinGW instructions, they are not written by myself.

For building the release builds, I recommend using the MXE toolchain instead, since it's capable of doing fully static builds instead, so you don't have DLL hell. For development builds, MSYS2 is fine.

PR #2827, opened 2021-03-09
BUILD: Support LLD linker in configure (opt-in)

orgads:
Reduces link time significantly.

Times comparison:

Linux+SSD
=========
ld: 14.2s
gold: 4.5s
lld: 1.7s

Windows+HDD
===========
ld: 110s
lld: 7s

### DIFF
--- a/configure
+++ b/configure
@@ -196,6 +196,7 @@ _no_pragma_pack=no
 _bink=yes
 _cloud=auto
 _pandoc=no
+_lld=no
 # Default vkeybd/eventrec options
 _vkeybd=no
 _eventrec=no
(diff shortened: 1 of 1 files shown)


lephilousophe:
LLD is not a drop-in replacement of Gold or GNU ld (especially on LTO stuff).
Support should be enabled explicitly.
I am not sure about why it should be in configure script though (like specify gold as we currently do).

orgads:
Do you mean using `LDFLAGS=-fuse-ld=lld` env variable?

This works, but is harder to discover for users not familiar with this option.

Fast link saves a lot of time when hacking on scummvm. I don't see that lto is used in the configure script, except for psp2, and didn't notice anything broken with my executable.

Will you accept it if I change the default to no, and remove the `auto` option?

sev-:
Yes, I think this is a good addition. Please change the default to no and I'll merge it.

More testing is required, though, this is why the default should not be 'auto'.

orgads:
Done

sev-:
Thanks, mergin

PR #2828, opened 2021-03-11
Sci0 sound restart

athrxx:
SCI: SCI0 sound restoring/restarting fixes

The main purpose of this is fixing incorrect sound restore after loading a savegame. Permanently hanging notes could sometimes be experienced when saving while a sound was playing and then reloading that savegame. That bug (and how it could be reproduced in KQIV) was pointed out to me by sluicebox. I just went over disasm (in this case mostly QFG, since I have the best disam there) and tried to match the original behavior a bit better. Commit b449564 is mainly for that. Commit cb50e48 aims at fixing inconsistent thread handling I had introduced with the initTrack method. I don't know whether it actually causes glitches. It was just one of the things I examined during my bug fixing attempts.
Commit 2a8289c fixes ScumVM specific things I had broken in the second commit (e. g. entering the GMM after clicking the SCI menu, which can't happen in the original).

I have done some testing of the bug that was reported to me and of the various situations pointed out by comments in the code (LSL3 start scene, ICEMAN room 14, etc). Sluicebox also has been testing, probably way more than me. So we're now confident that this is good to go...

### DIFF
--- a/engines/sci/engine/savegame.cpp
+++ b/engines/sci/engine/savegame.cpp
@@ -755,6 +755,17 @@ void SoundCommandParser::reconstructPlayList() {
 			processPlaySound(entry->soundObj, entry->playBed, true);
 		}
 	}
+
+	// Emulate the original SCI0 behavior: If no sound with status kSoundPlaying was found we
+	// look for the first sound with status kSoundPaused and start that. It relies on a correctly
+	// sorted playlist, but we have that...
+	if (_soundVersion <= SCI_VERSION_0_LATE && !_music->getFirstSlotWithStatus(kSoundPlaying)) {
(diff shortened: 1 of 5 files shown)


sluicebox (approved):
Great job! This fixes a bunch of edge cases, most of which I came up with while debugging the first one that turned up in https://bugs.scummvm.org/ticket/12021

I've been testing this a lot in Colonel's Bequest and KQ4, also some LSL3 and Iceman where known issues have been. These changes all look right to me and match what I'm seeing in SSCI.

bluegr:
Great! Many thanks for your work :)
This follows the original disasm, and since @sluicebox has also tested several SCI0 games with this change, it's safe to merge.
Kudos :) Merging

PR #2835, opened 2021-03-14
COMMON: Fix FriBidiParType failure on 3ds

orgads:
```
/data/src/master/common/unicode-bidi.cpp: In member function 'void
Common::UnicodeBiDiText::initWithU32String(const Common::U32String&)':
/data/src/master/common/unicode-bidi.cpp:78:3: error: invalid conversion
from 'uint32*' {aka 'unsigned int*'} to 'FriBidiParType*' {aka 'long
unsigned int*'} [-fpermissive]
   78 |   &_pbase_dir,
      |   ^~~~~~~~~~~
      |   |
      |   uint32* {aka unsigned int*}
```


sev-:
Thank you for the quick fix

PR #2836, opened 2021-03-14
COMMON: Don't use unavailable macros in unicode-bidi.h

lephilousophe:
Yet another stab at the FriBidiParType bug.

I use long unsigned int as it's the type used by the original enum defined in Fribidi (use of L suffixed integer literals).
According to https://en.cppreference.com/w/cpp/language/types, using uint32 could fail on some platforms where long unsigned int is 64 bits (LP64).

In addition, don't define constants in unicode-bidi.h as they are not available without Fribidi include.
This would fail if someone would want to use these constants.
As a consequence, the constructor with pbase should not be used outside of unicode-bidi.

### DIFF
--- a/common/unicode-bidi.cpp
+++ b/common/unicode-bidi.cpp
@@ -26,6 +26,11 @@
 
 #ifdef USE_FRIBIDI
 #include <fribidi/fribidi.h>
+#else
+/* This constant is used below in common code
+ * fake it here to lighten code
+ */
+#define FRIBIDI_PAR_ON 0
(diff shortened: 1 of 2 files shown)


sev-:
Thank you. merging

lephilousophe:
Type was fixed before.
Just keep the bits around Fribidi constants used without including fribidi.h.
Fix a GCC warning around va_args and enum.

bluegr:
Nice work! Merging

PR #2838, opened 2021-03-14
ULTIMA: Fix Clang warnings

orgads:
```
warning: definition of implicit copy assignment operator for 'X' is deprecated because it has a user-declared copy constructor
```

### DIFF
--- a/engines/ultima/shared/core/str.h
+++ b/engines/ultima/shared/core/str.h
@@ -45,6 +45,7 @@ public:
 	String(const String &str) : Common::String(str) {}
 	String(const Common::String &str) : Common::String(str) {}
 	explicit String(char c) : Common::String(c) {}
+	String& operator=(const String &str) { this->Common::String::operator=(str); return *this; }
 
 	/**
 	 * Returns the index of a given character, or -1 if not present
--- a/engines/ultima/shared/gfx/sprites.cpp
+++ b/engines/ultima/shared/gfx/sprites.cpp
@@ -52,6 +52,12 @@ Sprite::Sprite(const byte *src, uint bpp, uint16 w, uint16 h) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


engines/ultima/shared/core/str.h
```diff
@@ -45,6 +45,7 @@ class String : public Common::String {
 	String(const String &str) : Common::String(str) {}
 	String(const Common::String &str) : Common::String(str) {}
 	explicit String(char c) : Common::String(c) {}
+	using Common::String::operator=;
```
dreammaster, line 48: Unless I'm wrong, the using keyword is a C++11 keyword. The engine doesn't have any C++11 required yet, so I don't want to have to add in the requirement just for this.

engines/ultima/shared/gfx/sprites.cpp
```diff
@@ -52,6 +52,13 @@ Sprite::Sprite(const byte *src, uint bpp, uint16 w, uint16 h) {
 	}
 }
 
+Sprite &Sprite::operator=(const Sprite &src)
```
dreammaster, line 55: It's a minor nitpick, but our style guide would have the { on the same line as the method, as in (const Sprite &src) {

engines/ultima/shared/core/str.h
orgads: On this context it's not C++11. The C++11 form is a typedef replacement.

engines/ultima/shared/gfx/sprites.cpp
orgads: Will amend.

orgads: Done

digitall:
This looks good to me, but will leave this for @dreammaster to merge when happy with the changes.

PR #2839, opened 2021-03-14
MOHAWK: Fix Clang warning

orgads:
```
warning: definition of implicit copy assignment operator for 'LBValue' is deprecated because it has a user-declared copy constructor
```


digitall:
I was working on a similar fix for these warnings. Thanks for this code. Merging.

PR #2840, opened 2021-03-14
TITANIC: Fix Clang warning

orgads:
```
warning: definition of implicit copy assignment operator for 'X' is deprecated because it has a user-declared copy constructor
```


dreammaster:
Looks okay. Thanks

PR #2841, opened 2021-03-14
ULTIMA: Remove wrong conditions

orgads:
button cannot be any of these values. It is not even the right enum.

Detected by Clang.


dreammaster:
You're right. That code is redundant, thanks

PR #2842, opened 2021-03-14
ULTIMA: Use correct enum for button constants

orgads:
All the comparisons are done against Shared::MouseButton. Using values
from SpecialButtons makes no sense.


dreammaster:
Good point, I agree it should use that enum.

PR #2843, opened 2021-03-15
TINYGL: Implement a real operator=

orgads:
With logic similar to the copy ctor.


digitall:
This looks like a better solution than my addition. Merging. Thanks.

PR #2844, opened 2021-03-15
COMMON: Remove explicit ctors and copy assignment operator

orgads:
They're equivalent to the implicit ones anyway.


digitall:
This looks like a better solution than my addition of the copy assignment operator, so merging. Thanks.

PR #2846, opened 2021-03-15
DM: Remove trivial copy ctor and copy assignment operator

orgads:


digitall:
This looks like a better solution than my addition of the copy assignment operator, so merging. Thanks.

PR #2852, opened 2021-03-15
GLK: Fix compilation

orgads:
This amends commit bedba56f24ccb1eaa9381ab36c446bb0b00cfb0c.


criezy:
Thank you. It looks OK.
Let's see if it fixes the compilation with old gcc compilers that do not support c++11...

PR #2854, opened 2021-03-15
GLK: Fix copy assignment operator warning once again

orgads:
Back to square 2 ;)


digitall:
Will merge this as the GCC warnings are giving me a headache :) @orgads: Thanks for taking a look at these.

PR #2857, opened 2021-03-15
TINSEL: Add a missing case to switch statement

orgads:
It's not really needed, only to satisfy the compilers.


digitall:
Looks fine.

PR #2859, opened 2021-03-16
GPH: Don't call DownscaleSdlGraphicsManager anymore

lephilousophe:
Following removal of DownscaleSdlGraphicsManager, there is still some code calling it left.
I don't know enough of graphics code, hence the PR.
I am not sure about how all of this will behave on real hardware.


bluegr:
Nice catch, this will fix the build error on the GP2X Wiz port as DownscaleSdlGraphicsManager doesn't exist anymore. As mentioned, the behavior will need to be checked

PR #2861, opened 2021-03-16
GRAPHICS: Additional asserts for Surface struct

henke37:
This aims to disallow reaching beyond the buffer for surfaces.

Suffice to say, but at least `SCUMM` fails this check instantly. This is gonna take a lot of work to cleanup before it can be merged.

### DIFF
--- a/graphics/surface.cpp
+++ b/graphics/surface.cpp
@@ -89,6 +89,7 @@ void Surface::init(uint16 width, uint16 height, uint16 newPitch, void *newPixels
 	pitch = newPitch;
 	pixels = newPixels;
 	format = f;
+	assert(pitch >= w * format.bytesPerPixel);
 }
 
 void Surface::copyFrom(const Surface &surf) {
--- a/graphics/surface.h
+++ b/graphics/surface.h
@@ -127,6 +127,10 @@ public:
 	 * @return Pointer to the pixel.
 	 */
 	inline const void *getBasePtr(int x, int y) const {
+		assert(x >= 0);
+		assert(x < w);
+		assert(y >= 0);
+		assert(y < h);
 		return (const byte *)(pixels) + y * pitch + x * format.bytesPerPixel;
 	}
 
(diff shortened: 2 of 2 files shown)


sev-:
Yes, I am afraid that without fixing the engines we could lose some games as "crashing" :D

criezy:
I know we can get an out of bound ptr in the `ManagedSurface` blitting code, and I would not be surprised if several engines are also doing something similar. For example you can imagine the following function to blit a sprite on the screen, and the sprite can be partially outside the screen bounds.
```c++
void MyEngine::blit(const Surface *sprite, int x0, int y0) {
    const byte *spritePixels = sprite->getBasePtr(0, 0);
    byte *screenPixels = _screenSurface->getBasePtr(x0, y0);
    for (int y = 0 ; y < sprite->h ; ++y, screenPixels += _screenSurface->pitch, spritePixels += sprite->pitch) {
        if (y0 + y < 0 || y0 + y >= _screenSurface->h)
            continue;
        for (int x = 0 ; x < sprite->w ; ++x) {
            if (x0 + x < 0 || x0 + x >= _screenSurface->w)
                continue;
            *(((uint32*)screenPixels) + x) =  *(((uint32*)spritePixels) + x);
        }
    }
} 
```

I know this code is not perfect as I simplified it (such as assuming that both surfaces have the same 32 bit pixel format and not breaking out of the loop when we are >= to the screenSurface height/width). It is just meant as an illustration of what we can currently do.

Is this something we want to forbid? If so there might be a lot of code we need to change before we can add the assert. And the required changes would make the code slightly more complicated.

I don't have any issue with the assert on the `pitch` added in the `init()` function. But I am not convinced about adding asserts in the `getBasePtr()` functions.

criezy:
Here is a real example of this sort of code: https://github.com/scummvm/scummvm/blob/49455262198130df3bf21c0c114c53c23ba8e661/graphics/managed_surface.cpp#L276-L292

And by the way this example shows the value of possibly having an assert, because in the 1 byte per pixel case it does not correctly handle properly the case where the destRect is partially outside the surface.

mgerhardy:
While this is definitely a good change, we also have a lot of engines that call `getBasePtr` in the loops to blit - and this would most likely make it a lot slower in debug builds.

henke37:
I think that the cost of the comparisons will be dwarfed by the cost of the function call. They are rather trivial.

sev-:
Closing this. This is a drastic change. We already having tons of issues from the HashMap changes, because we do not have proper test coverage, neither even sanity checks were provided by it. This PR has very similar side effects. I am afraid, we will have nasty regressions.

PR #2863, opened 2021-03-18
SCUMM: Fix X-range of selection for Hebrew on scumm<7

orgads:
`curRect.left` is assigned in the script, and it is not modified. `right` is
assigned with the screen width - original left.

This results in bad highlighting of the verbs all over the line width,
instead of being limited to the actual string.

Due to that, sometimes the selection range of the up/down arrows overlaps
with some of the verbs, and then these verbs cannot be selected.

Solve by storing the original left value, and using it as initial x
position for the string (the actual right-to-left manipulation is done in
`drawString()`), and modify the value of `curRect.left` to match the string
that was actually drawn.

This bug is similar to the one that was fixed in 58e921eb87, but the
solution that was done there for v7 and v8 cannot work here, because the
string logic is much more complicated.

### DIFF
--- a/engines/scumm/saveload.cpp
+++ b/engines/scumm/saveload.cpp
@@ -815,7 +815,7 @@ static void syncWithSerializer(Common::Serializer &s, ObjectData &od) {
 }
 
 static void syncWithSerializer(Common::Serializer &s, VerbSlot &vs) {
-	s.syncAsSint16LE(vs.curRect.left, VER(8));
+	s.syncAsSint16LE(vs.origLeft, VER(8));
 	s.syncAsSint16LE(vs.curRect.top, VER(8));
 	s.syncAsSint16LE(vs.curRect.right, VER(8));
 	s.syncAsSint16LE(vs.curRect.bottom, VER(8));
@@ -837,6 +837,8 @@ static void syncWithSerializer(Common::Serializer &s, VerbSlot &vs) {
 	s.syncAsByte(vs.center, VER(8));
 	s.syncAsByte(vs.prep, VER(8));
 	s.syncAsUint16LE(vs.imgindex, VER(8));
+	if (s.isLoading() && s.getVersion() >= 8)
+		vs.curRect.left = vs.origLeft;
 }
 
 static void syncWithSerializer(Common::Serializer &s, ScriptSlot &ss) {
--- a/engines/scumm/verbs.cpp
+++ b/engines/scumm/verbs.cpp
@@ -175,7 +175,7 @@ void ScummEngine_v0::resetVerbs() {
 		vs->center = 0;
 		vs->imgindex = 0;
 		vs->prep = verbPrepIdType(vtable[i - 1].id);
-		vs->curRect.left = vtable[i - 1].x_pos * 8;
+		vs->curRect.left = vs->origLeft = vtable[i - 1].x_pos * 8;
 		vs->curRect.top = vtable[i - 1].y_pos * 8 + virt->topline + 8;
 		loadPtrToResource(rtVerb, i, (const byte*)vtable[i - 1].name);
 	}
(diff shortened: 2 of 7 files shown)


orgads:
I have to say I don't really like this solution. If you can suggest an alternative way to solve this, I'll be more than happy to test it.

engines/scumm/saveload.cpp
```diff
@@ -837,6 +837,7 @@ static void syncWithSerializer(Common::Serializer &s, VerbSlot &vs) {
 	s.syncAsByte(vs.center, VER(8));
 	s.syncAsByte(vs.prep, VER(8));
 	s.syncAsUint16LE(vs.imgindex, VER(8));
+	vs.origLeft = -1;
```
orgads, line 840: Ahh now I see that if I save the game and load it, left and origLeft get wrong values, and the text becomes left-aligned :/

bluegr:
Well, you could work with the original left / right coordinates internally, and only adjust them when displaying the string on-screen

orgads:
I thought about it too, but it's not only on display, but also on findVerbAtPos.

I started to check this approach, but I'm not done yet. Will continue next week hopefully.

Thanks for reviewing.

orgads:
#2898 is an incomplete attempt, but I'm very reluctant about it. What do you think?

orgads:
Looks better now. Tested on `monkey1` and `comi`. I still don't like it much, but couldn't find a better solution. I'm open for other suggestions :)

orgads:
ping?

bluegr:
This is a much better approach!
So, I understand that V8 games can't work with this approach, right? Thus, COMI and Dig won't be affected, and your tests verified this?

orgads:
For COMI the script reads the verbs all the time (I don't know why), and the string handling is very simple. So what I did there was to effectively replace left and right for RTL. I don't really care what's in the save file, but I think it contains correct values. I didn't test Dig.

For scumm<7 the string logic is very complex, and split between verbs.cpp, charset.cpp and string.cpp. As much as I tried, I couldn't find a reasonable way to keep track of left and right without breaking the display, selection or erasing of the text. So I ended up with storing origLeft every time `left` is assigned, and using it on save/load.

I tested COMI to make sure it isn't broken by this, because of the change in save/load.

orgads:
@bluegr ping

engines/scumm/verbs.cpp
```diff
@@ -1084,7 +1084,7 @@ void ScummEngine::drawVerb(int verb, int mode) {
 		restoreVerbBG(verb);
 
 		_string[4].charset = vs->charset_nr;
-		_string[4].xpos = vs->curRect.left;
+		_string[4].xpos = vs->origLeft;
```
bluegr, line 1087: Will this work in all languages, or should there be a check for Hebrew here?

orgads: I think it should be ok with other languages. @BLooperZ?

orgads: Added a condition. If you don't have other rejections, please merge it. We can polish it later.

orgads: I reverted the condition, because I think it can break things. I assign `vs->curRect.left` below (that's the actual fix of the bug), and here we need the original left value.

BLooperZ:
Thanks for the fix.
Adding `origLeft` might also prove useful with other issues as well

engines/scumm/verbs.cpp
bluegr: Great

bluegr:
Thanks for your work on this. Merging

criezy:
The change in this pull request introduced a regression in some non-Hebrew games for which we can no longer click on the first verb column. See https://bugs.scummvm.org/ticket/12552
@orgads Could you take a look at this?

PR #2864, opened 2021-03-19
IOS7: Add 64-bit target

lephilousophe:
This PR adds support of 64-bit iPhone to configure script.


criezy (approved):
I have never compiled for iOS on the command line, but the change seems to make sense and is simple enough.

lephilousophe:
Thanks.
Merging now.

PR #2867, opened 2021-03-20
BACKENDS: Fix MinGW warnings on GetProcAddress

orgads:


bluegr:
This is a known issue with the newest MinGW and WinAPI. I can't say I'm really happy with the way we're silencing the warnings, but there's not much to be done, as this is a Microsoft API. After a brief search on the net, it seems that other projects are following a similar approach for these cases.

Leaving this open for some days in case there's any other suggestion of how to silence these warnings

orgads:
This is not really silencing the warning. It's a real fix.

The warning makes sense actually. If you look at the definition of `FARPROC` it's:

```
typedef INT_PTR (FAR WINAPI *FARPROC)();
```

Meaning a pointer to a function that accepts no arguments. But in fact, the returned function *does* accept argument, so you have to cast it.

Now, generally casting to an incompatible function is wrong, and that's what the compiler warns about. For example, suppose you have this:

```c++
typedef void (*func)();
typedef void (*funcWithInt)(int);

void foo() {
    std::cout << "foo" << std::endl;
}

func bar() { return foo; }

int main() {
    int a = 42;
    funcWithInt f = (funcWithInt)bar();
    f(42);
    return 0;
}
```

Calling `f`, which is `foo`, with 42 is wrong, but the cast allows it. This actually hides a bug - you're using C-style cast which falls back to `reinterpret_cast` in case it cannot `static_cast`. If you replace the cast with `funcWithInt f = static_cast<funcWithInt>(bar());` compilation just fails.

On this case however, when "you know what you're doing", it's totally fine to suppress the warning by double casting.

Actually I'm surprised that MSVC doesn't warn about this.

orgads:
Done

bluegr:
Thanks for your work on this, and for the explanation provided. Merging

PR #2868, opened 2021-03-20
AUDIO: Fix compiler warning

orgads:
size_t and long are not the same on Win64.

### DIFF
--- a/audio/soundfont/vgmcoll.cpp
+++ b/audio/soundfont/vgmcoll.cpp
@@ -185,7 +185,7 @@ SynthFile *VGMColl::CreateSynthFile(VGMInstrSet *theInstrSet) {
 				newRgn->SetWaveLinkInfo(0, 0, 1, (uint32) realSampNum);
 
 				if (realSampNum >= finalSamps.size()) {
-					debug("Sample %u does not exist", unsigned(realSampNum));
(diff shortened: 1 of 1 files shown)


audio/soundfont/vgmcoll.cpp
```diff
@@ -185,7 +185,7 @@ SynthFile *VGMColl::CreateSynthFile(VGMInstrSet *theInstrSet) {
 				newRgn->SetWaveLinkInfo(0, 0, 1, (uint32) realSampNum);
 
 				if (realSampNum >= finalSamps.size()) {
-					debug("Sample %lu does not exist", realSampNum);
+					debug("Sample %u does not exist", unsigned(realSampNum));
```
bluegr, line 188:
```suggestion
					debug("Sample %u does not exist", (uint32) realSampNum);
```

orgads: I agree that `uint32` is more portable than `unsigned`, but why do you prefer C-style cast over ctor-like cast, which is usually a better choice for C++?

bluegr: In this case, mostly for portability reasons, as well as uniformity with the rest of the code

orgads:
Thanks for merging and fixing, and sorry for the delay. I was on vacation.

PR #2869, opened 2021-03-20
COMMON: Fix compiler warning on MinGW

orgads:
Disallow long long HashMap (it is not used anyway), as the lli and llu formats are not supported in MinGW.

### DIFF
--- a/common/hashmap.cpp
+++ b/common/hashmap.cpp
@@ -58,11 +58,11 @@ template<> void unknownKeyError(::Common::String key) {
 }
 
 template<> void unknownKeyError(signed char key) {
-	error("Unknown key \"%hhi\"", key);
+	error("Unknown key \"%i\"", key);
 }
 
 template<> void unknownKeyError(unsigned char key) {
-	error("Unknown key \"%hhu\"", key);
+	error("Unknown key \"%u\"", key);
 }
 
 template<> void unknownKeyError(short signed key) {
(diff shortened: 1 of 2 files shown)


henke37:
Is there a need to also change the format strings for the `char` overloads?

orgads:
Yes, hh* is not supported too.

common/hashmap.cpp
```diff
@@ -58,11 +58,11 @@ template<> void unknownKeyError(::Common::String key) {
 }
 
 template<> void unknownKeyError(signed char key) {
-	error("Unknown key \"%hhi\"", key);
+	error("Unknown key \"%d\"", int(key));
```
lephilousophe, line 61: I would say that specifying int and unsigned is useless as key will be promoted to int.
cf. https://en.cppreference.com/w/cpp/language/variadic_arguments
>  bool, char, short, and unscoped enumerations are converted to int or wider integer types as in integer promotion

orgads: Done

common/hashmap.h
```diff
@@ -734,6 +730,11 @@ void HashMap<Key, Val, HashFunc, EqualFunc>::erase(const Key &key) {
 
 /** @} */
 
+template<class Val, class HashFunc, class EqualFunc>
+class HashMap<signed long long, Val, HashFunc, EqualFunc>;
+template<class Val, class HashFunc, class EqualFunc>
+class HashMap<unsigned long long, Val, HashFunc, EqualFunc>;
+
```
bluegr, line 737: So, since these do not have an implementation, there should be a comment here that they should not be used

orgads: What for? Compilation will fall anyway.

criezy:
I don't understand why you want to disallow long long hashmap simply because of a format issue in `unknownKeyError()`. Why not just remove the specialisations of that function for long long but still allow having hashmap with long long? It will then use the generic implementation of `unknownKeyError()`. That one does not print the key value, but since this is only meant as a debug helper this is not really an issue.

orgads:
Actually if #2887 is accepted, this can be closed.

bluegr:
Closing, as #2887 has been merged

PR #2871, opened 2021-03-23
GRAPHICS: Expose ascent value in font class

criezy:
This pull request adds a `getFontAscent()` function to the `Font` class and implements it various classes that derives from it. I decided to make it optional with a default implementation that returns -1 to avoid having to add it to all the fonts implemented in engines where it is not needed (I saw that at least bladeruner and cryomni3d are implementing other font classes). It is implemented for all the font classes in the Graphics namespace.

This is needed in the AGS engine to properly align vertically the text drawn with TTF fonts.
Since this is a change in common code I am proposing this as a pull request. But I would like to merge this quickly so that I can push the changes in the AGS engine to properly display the text.

### DIFF
--- a/graphics/font.cpp
+++ b/graphics/font.cpp
@@ -28,6 +28,10 @@
 
 namespace Graphics {
 
+int Font::getFontAscent() const {
+	return -1;
+}
+
 int Font::getKerningOffset(uint32 left, uint32 right) const {
 	return 0;
 }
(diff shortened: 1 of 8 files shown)


mgerhardy:
Looks fine to me.

bluegr:
Looks good to me too

lephilousophe:
OK for me, should not disturb cryomni3d.

criezy:
Thank you for the quick review. I will merge this now as it is a relatively trivial change that should not break any existing code.

PR #2872, opened 2021-03-23
SCI: Added fan made Hebrew translation for QFG1VGA

ZvikaZ:
HebrewAdventure team has finished translation of QFG1VGA:

image


bluegr:
Nice, thanks!

PR #2874, opened 2021-03-23
CONFIGURE: Avoid needless overwrite of header files

orgads:
They trigger extensive rebuild of source files without a reason every time
the configure script changes, even if the result is exactly the same.


orgads:
This is not enough. Need to avoid retriggering configure too...

configure
```diff
@@ -6481,10 +6493,13 @@ EOF
...
+		cat >> engines/plugins_table.h.new << EOF
 #if PLUGIN_ENABLED_STATIC($j)
 LINK_PLUGIN($j)
 #endif
 EOF
 	fi
 done
+copy_if_changed engines/plugins_table.h.new engines/plugins_table.h
+
+touch configure.stamp
```
mgerhardy, line 6505: not sure if `configure.stamp` is already part of `.gitignore` by some pattern - but if not, this should get added. Other than that: :rocket:

orgads: Good point, thanks.

orgads:
🚀?

mgerhardy:
Sorry. The rocket was an agreement (i like the change)

orgads:
Ah, I thought it was "let's merge it" :)

configure
```diff
@@ -6306,6 +6315,8 @@ typedef $type_ptr  uintptr;
 
 #endif /* CONFIG_H */
 EOF
+copy_if_changed config.h.new config.h
+
 
 echo "Creating config.mk"
 cat > config.mk << EOF
```
criezy, line 6322: Is there a reason not to create a `config.mk.new` here and then use `copy_if_changed config.mk.new config.mk`?

Currently that does not really matter as a modified `config.mk` does not trigger any files to be rebuilt. But I am thinking this is a bug, as it causes build errors in partial rebuild due to inconsistent flags used to build files. With your changes it also means that running again `configure` with only changing options that impact this `config.mk` file (for example `--enable-asan` does not trigger any rebuild, when it should probably rebuild everything.

orgads: No problem. I'll change it tomorrow.

lephilousophe:
This change will impact buildbot. Can we wait next Saturday (2 says) to merge? I want to be ready to update it. 
Thanks

configure
orgads: Done

lephilousophe:
I will merge this in 1 hour as problems raised got fixed.

lephilousophe:
Merging now

PR #2875, opened 2021-03-23
COMMON: Change JSON integer storage to int32

orgads:
Almost the last piece of `%lld`.

Fixes GCC warnings on MinGW-64.

### DIFF
--- a/common/json.h
+++ b/common/json.h
@@ -90,13 +90,17 @@ enum JSONType { JSONType_Null, JSONType_String, JSONType_Bool, JSONType_Number, 
 class JSONValue {
 	friend class JSON;
 
+private:
+	// Not implemented, and should not be used.
+	JSONValue(long long int numberValue);
+
 public:
 	JSONValue(/*NULL*/);
 	JSONValue(const char *charValue);
 	JSONValue(const String &stringValue);
 	JSONValue(bool boolValue);
 	JSONValue(double numberValue);
-	JSONValue(long long int numberValue);
+	JSONValue(int32 numberValue);
 	JSONValue(const JSONArray &arrayValue);
 	JSONValue(const JSONObject &objectValue);
 	JSONValue(const JSONValue &source);
@@ -113,7 +117,7 @@ public:
 	const String &asString() const;
 	bool asBool() const;
 	double asNumber() const;
-	long long int asIntegerNumber() const;
+	int32 asIntegerNumber() const;
 	const JSONArray &asArray() const;
 	const JSONObject &asObject() const;
 
@@ -140,7 +144,7 @@ private:
 	union {
 		bool _boolValue;
 		double _numberValue;
-		long long int _integerValue;
+		int32 _integerValue;
 		String *_stringValue;
 		JSONArray *_arrayValue;
 		JSONObject *_objectValue;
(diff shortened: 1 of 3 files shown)


common/json.h
```diff
@@ -90,13 +90,17 @@ enum JSONType { JSONType_Null, JSONType_String, JSONType_Bool, JSONType_Number,
 class JSONValue {
 	friend class JSON;
 
+private:
+	// Not implemented, and should not be used.
+	JSONValue(long long int numberValue);
```
bluegr, line 95: If the implementation is missing, won't the compiler complain?

orgads: No. Only the linker, in case it is actually used (can only be used by JSONValue itself, since it is private).

criezy: This should indeed not be an issue for the compiler, and I have not seen any linker complain about non-implemented functions when they are not used for a while. However I remember linkers on some platforms throwing an error about this a while ago. I am not sure if I misremember, if the standard has changed, or if this is simply an undefined behaviour and up to the linker to give an error or not (and most do not). It might be good to clarify that this is valid code with a defined behaviour before me merge this.

orgads: It is a typical pattern used for non-copyable classes. Copy ctor and copy assignment operator are declared as private and not implemented. I never had a problem with this even with very old toolchains.

Maybe what you saw was link error on unimplemented *virtual* function? Because virtual functions must be resolved at link time.

criezy: Right. I have seen a lot of copy constructor or assignment operators declared as private and with an empty implementation (or deleted in code using c++11), but I don't see non-implemented ones as often so I forgot. This is indeed perfectly valid (and we actually do this in ScummVM with the `NonCopyable` class).

You may be right that the errors I saw in the past were for virtual functions. My memory is a bit hazy on that as that was around 20 years ago for platforms that no longer exist (and I would also not necessarily trust the vendor compilers and linkers for this platforms - I still have nightmares about trying to write and compile template code with the SGI compiler on IRIX).

orgads: :)

orgads: By the way, the linker error for a missing implementation of virtual functions is rather cryptic:

```c++
struct Foo { virtual void func(); };
int main() { Foo foo; return 0; }
```

This is ld:
```
undefined reference to `vtable for Foo'
```

lld actually has a better description:
```
ld.lld: error: undefined symbol: vtable for Foo
>>> referenced by example.cpp:1 (<source>:1)
>>>               /tmp/example-6ae871.o:(Foo::Foo())
>>> the vtable symbol may be undefined because the class is missing its key function (see https://lld.llvm.org/missingkeyfunction)
```

bluegr:
The change from int to long long int was made when support for Google Drive storage was added:
https://github.com/scummvm/scummvm/commit/da3b7bd8d9f3d3828b8cea6dff60e5f43e7ad4b1#diff-eeecac4c684bef0d040e6efb78e0a7df765ec84fe6644c6f6255a8cfcc92eb85

In this commit, the change is noted as a fix. Presumably, it has been made to support Dropbox timestamps. Here is a comment which sheds some light:
https://github.com/dropbox/json11/blob/master/json11.hpp

```
 * Fortunately, double-precision IEEE754 ('double') can precisely store any integer in the
 * range +/-2^53, which includes every 'int' on most systems. (Timestamps often use int64
 * or long long to avoid the Y2038K problem; a double storing microseconds since some epoch
 * will be exact for +/- 275 years.)
```

As such, I believe that it shouldn't be reverted

orgads:
Thanks for the careful review. I just filed #2887 to avoid the format warnings on MinGW. Closing this change.

PR #2876, opened 2021-03-23
BUILD: Disable global-constructors warning for Clang

orgads:
Most (all?) of the global objects have trivial and independent ctor and
dtor, so this is not a real problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ endif
 
 ifeq "$(HAVE_CLANG)" "1"
 	CXXFLAGS+= -Wno-conversion -Wno-shorten-64-to-32 -Wno-sign-compare -Wno-four-char-constants
+	CXXFLAGS+= -Wno-global-constructors
 	# We use a anonymous nested type declaration in an anonymous union in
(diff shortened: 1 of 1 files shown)


bluegr:
This is not a good idea. We *need* warnings about global constructors, because it's not certain if such objects are destroyed correctly when the engine exits.

We had such cases in the tinsel engine, where there were a lot of globally constructed objects in the engine, which were then not destroyed fully upon the engine's exit. The result was that when the player tried to play, return to launcher and then start the game again, the engine would have kept an invalid state, so the game had all sorts of odd behavior after relaunching the engine within the same session.

orgads:
Not sure I follow.

This warning appears on global objects that are only destroyed on shutdown. The most common problem with early initialized objects is that you don't have control over the initialization/destruction order, and if they depend on other globals they might be accessed uninitialized.

digitall:
orgads: The problem here is not a problem of uninitialized memory, but that globally initialized objects cause game state issues for ScummVM engines. The fix is to replace the globally initialized objects with members of the game engine object.

orgads:
Ok, I'll close it then.

PR #2877, opened 2021-03-23
PRIVATE: Fix potentially uninitialized variable on release build

orgads:
assert is no-op.


bluegr:
A false positive, but initializing this won't hurt. Merging

PR #2887, opened 2021-03-25
COMMON: Use __gnu_printf__ instead of __printf__ for gcc on MinGW

orgads:
With `__printf__`, MinGW prefers the portable old ms-style printf, and
warns if formats like `%lld` and `%hhd` are used.

In practice, if __USE_MINGW_ANSI_STDIO is not defined, it defaults to 1
on C++11, and these formats are supported. Using `__gnu_printf__`
suppresses bad format warnings.

### DIFF
--- a/common/scummsys.h
+++ b/common/scummsys.h
@@ -339,7 +339,11 @@
 //
 #ifndef GCC_PRINTF
 	#if defined(__GNUC__) || defined(__INTEL_COMPILER)
-		#define GCC_PRINTF(x,y) __attribute__((__format__(__printf__, x, y)))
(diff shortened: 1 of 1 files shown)


orgads:
Clang doesn't like it. I'll fix it later/tomorrow.

orgads:
Fixed. Ready for review.

orgads:
ping?

bluegr:
Thanks, the issue here looks to be the mixture of gcc and Microsoft's runtime library.
More information can be found here:
https://stackoverflow.com/questions/29710064/a-code-that-doesnt-work-on-gcc

Thanks for your work. Any issues with this can be checked via buildbot.
Merging

PR #2891, opened 2021-03-26
GUI: HiDPI theme support for OpenGL SDL backend

sev-:
Finally, I was able to finish the bulk of the implementation for HiDPI support.

This change detects if the DPI of the monitor, as reported by the system and scales GUI if it is present.

For achieving that, a concept of the base resolution is introduced, that is the resolution of our GUI theme. Those are currently hardcoded and are catching heights of 239, 399, 479 and 719. Then it tries to switch to the relevant height processed by our GUI, those are 1x, 2x and 3x theme. The logic sits at https://github.com/sev-/scummvm/blob/gui-scale/gui/gui-manager.cpp#L110

As a result, the image gets bigger to a certain window height and then switches to the bigger theme.

All widgets get scaled, so are the fonts, Thus, the general proportions stay useable.

There are several things that are currently missing, but optionally could be addressed after the merge:

- [x] Testing it on non-HiDPI display. I do not have those at my home, I cannot confirm I did not break it
- [x] Cursor has 2x position scale in non-HiDPI mode
- [x] Transparency in LoDPI mode is broken: black squares around SVG
- [ ] ManagedSurface::blitFromInner() is super slow, primarily colorToARGB/ARGBtoColor slowness
- [x] "Unsupported bpp in blitManagedSurface: 2!" when switching to non-OpenGL renderer
- [x] Wrong upper parts of widgets in non-HiDPI
- [x] ScummVM logo in GMM is not drawn on the first launch
- [x] ScummVM logo in GMM is not scaled properly or aspect ratio is not maintained when window is resized
- [x] Similarly, ScummVM logo at the launcher sometimes disappears when window is resized
- [x] Scrollbar is not drawn
- [x] Tab controls are not drawn
- [x] Oversized keymapper controls
- [x] Spacing between buttons in Launcher is not scaled
- [x] Tab labels padding is incorrect
- [x] Confirmation dialogs have wrong size
- [x] DropDownButton has wrong rendering, drop down is too narrow
- [x] Save/Load dialog is not properly scaled
- [x] I plan to add SVG support for the images. Mataniko already converted our images to vector. The current code scales bitmaps.

Along the way, I had to fix our vector rendered because the shadow code was very buggy.
Also, I added `scalable` attribute to the theme constants to mark those which need to be scaled without touching the rest.

The code is tied to OpenGL SDL because this is the only backend that currently supports arbitrary window resizing. But the code itself is all in GUI, so you can enable it for any platform. Just indicate it by setting `OSystem::kFeatureHiDPI`.

That is basically it, please, review and provide your feedback.

The screenshots (look at the native res files, here they are downscaled):

ScummVM master:
master
This branch:
branch


ScummVM-Translations:
Looks amazing! A couple of issues I noticed:

- [ ] Buttons in Keymaps are oversized
- [ ] Classic theme is extra small

criezy:
It does look great. The text in particular is so much crisper.
I can do some tests on a non-HiDPI display this weekend.

I noticed just looking at the pictures that the spacing between the buttons doesn't seem to be scaled? Is that on purpose?

Mataniko:
Looks amazing! A couple of issues I noticed:

- [ ] Buttons in Keymaps are oversized
- [ ] Classic theme is extra small
- [ ] GMM Options dialog takes over the entire screen
- [ ] Dialog after saving a game via GMM text is rendered is out of bounds

mgerhardy:
Unfortunately this doesn't change anything for me.

Screenshot from 2021-03-26 14-46-51
Screenshot from 2021-03-26 14-46-23

mgerhardy:
These are the details for the display
```
INFO: (0) possible display modes:
INFO: (0) 3840x2160@60Hz SDL_PIXELFORMAT_RGB888
INFO: (0) resolution (3840:2160) (pixel)
INFO: (0) resolution (3840:2160) (screen)
INFO: (0) dpi factor: 2.945155
INFO: (0) dpi factor h: 2.944927
INFO: (0) dpi factor v: 2.945876
```

Mataniko:
Are you running in OpenGL mode?

sev-:
@criezy no, spacing is not intentional. Everything should look practically 1:1

sev-:
@mgerhardy run with -d3 and try to change the window size. It should print something like

```
Setting 1280 x 960 -> 640 x 480 -- 2
Setting 1386 x 1042 -> 638 x 480 -- 2.17083
```

First is your current window resolution as reported by the backend, second is the computed base GUI resolution, and final is the scale factor. What are your numbers?

sev-:
...and for the classic theme this thing is currently useless because we use bitmap fonts. The possible solution is to scale those too, because it looks like this:

backends/graphics/sdl/sdl-graphics.cpp
```diff
@@ -208,6 +208,11 @@ bool SdlGraphicsManager::notifyMousePosition(Common::Point &mouse) {
 	mouse.y = CLIP<int16>(mouse.y, 0, _windowHeight - 1);
 
 	int showCursor = SDL_DISABLE;
+	// DPI aware scaling to mouse position
```
mgerhardy, line 211: Wouldn't it be better to use a float here, too - you might get rounding issues in the higher coordinate ranges otherwise.

gui/ThemeEval.cpp
```diff
@@ -169,6 +169,14 @@ ThemeEval &ThemeEval::addSpace(int size) {
 	return *this;
 }
 
+#define SCALEVALUE(val) (val > 0 ? val * _scaleFactor : val)
```
mgerhardy, line 172: Most likely due to the fact that I'm not that much into the scummvm ui code - but why only for positive values?

Mataniko:
> 
> 
> ...and for the classic theme this thing is currently useless because we use bitmap fonts. The possible solution is to scale those 

Might be easier to package 3 versions of the font, small, medium and large to match our scale breakpoints

gui/ThemeEval.cpp
sev-: -1 is a special value, meaning "compute it"

sev-:
> Might be easier to package 3 versions of the font, small, medium and large to match our scale breakpoints
A sound idea, however, BDF fonts are huge. Still, somebody could need to try it, but is there a sense in having that ugly historical version supported?

criezy:
I did some tests today on macOS with no Retina display, and it is broken:
- The GUI is too big (it is scaled x2).
- The mouse is off compared to the system mouse. At the top left they are at the same position, but the ScummVM cursor reaches the bottom-left when the system cursor is in the middle of the window (so again a factor 2).

I will try to debug this and fix this.

criezy:
I also made a quick hack to enable the scaling in iOS and it looks nice:

Current master:
IMG_0162

With HiDPI support:
IMG_0164

I will need to clean a bit the code. In particular we cannot assume a scaling of x2 in `GuiManager::computeScaleFactor()` as the usual scale factor for iPhone/iPad retina display is x3 and using x2 displays everything too small. So I will need to expose the HiDPI screen scaling in `OSystem` (which should be trivial to do - and this will probably also fix the display on non-retina macOS).

@sev- By the way, what is the reasoning for snapping the GUI base resolution to a few predefined resolutions in `GuiManager::computeScaleFactor()` rather than jut use the overlay size / scaling factor?

sev-:
> By the way, what is the reasoning for snapping the GUI base resolution to a few predefined resolutions in GuiManager::computeScaleFactor() rather than jut use the overlay size / scaling factor?

I wanted 640x480 resolution to stay as long as possible in order to avoid the situation as you have on your iPad screenshots when things are looking way too small.

criezy:
The screenshot is actually for a case where it looks good on the iPad (click on the image to see it as its full resolution - but possibly if your screen has a different dpi as the iPad it will look smaller/bigger than it does on the real device). It actually looks very similar to how it looks on master in term of scaling.

But I was also thinking of adding a user scale option, that default to 1, and is multiplied to the screen scale factor to let the user display the GUI bigger or smaller based on his preferences/needs.

ScummVM-Translations:
One thing that's completely missing is the tab switching arrows in the options dialog
image

antoniou79:
I don't want to confuse things, as I am unsure if this is the scaling that would be applied for Android platforms as well (where we currently have issues with OpenGL scaling). I'll briefly post my test feedback on Android, but you can ignore it if it's not relevant with the current work. 

I've tested by declaring support of System::kFeatureHiDPI. The device is an android 10 RedMi, and by Droid Info app, display info is: Resolution 1080x2400px, Software Density 440 dpi (gsm arena says ~395 ppi density).

Main issues include 
- an oversized mouse cursor
- small ui and text elements (too high resolution for screen)

Launcher - Standard:
Standard-Launcher

Launcher - Gui-scale:
hiDPI-Launcher

In game GMM - Standard:
Standard-DigInGameGMM

In game GMM - Gui-scale:
hiDPI-DigIntroGMM
hiDPI-DigInGameGMM

I also have no idea what is going on with the ScummVM image logo vs ScummVM text logo vs no ScummVM text nor logo in the launcher and GMM, but that is probably a separate issue.

sev-:
@antoniou79 in order to understand and simulate what is going on (particularly I am curious about the cursor issue), I need resolution, base resolution and the computed scale: https://github.com/sev-/scummvm/blob/gui-scale/gui/gui-manager.cpp#L135

ScummVM logo is a known issue, it has its own tickbox above.

@criezy ha, I did not notice the absence of the tab controls. This is exactly why I launched this PR, to collect all of this information

The spacing between buttons in the launcher is kind of nasty: our current theme code appeared to be _assuming_ the DPI and relying on the low DPI resolution. Particularly, in this case, there is _NO_ space between the buttons, it is just space left for the hardcoded shadow. I'm still puzzled on how to solve it and keep the looks unchanged.

antoniou79:
> @antoniou79 in order to understand and simulate what is going on (particularly I am curious about the cursor issue), I need resolution, base resolution and the computed scale: https://github.com/sev-/scummvm/blob/gui-scale/gui/gui-manager.cpp#L135
> 
I should note that what I did to "set the OSystem::kFeatureHiDPI" was to:
1. return true for it in OSystem_Android::hasFeature()
2. return true for it in OSystem_Android::getFeatureState() 

From my log messages:
OpenGLGraphicsManager::handleResizeImpl: width:2298 height:1080 xdpi:397 ydpi:395!
Setting 2298 x 1080 -> 1021 x 480 -- 2.25!

I am including also the line above from OpenGLGraphicsManager::handleResizeImpl() (opengl-graphics.cpp) because we call that initializing the surface (DPI values and dimensions are retrieved in java code and passed through JNI to the c++ code), and those are the specific arguments for my testing device.

In the OpenGLGraphicsManager::handleResizeImpl there's also a call to OpenGLGraphicsManager::recalculateCursorScaling() which seems relevant. I put some log messages in there. It seems that 
_cursorDontScale is: true 
and _cursorWidthScaled, _cursorHeightScaled start from:
_cursorWidthScaled: 72 _cursorHeightScaled: 72
and end up: 
 _cursorWidthScaled: 317 _cursorHeightScaled: 316

From a quick lookup I found that we had merged a PR by @ccawley2011 for high DPI support on Android, a couple of years ago:
https://github.com/scummvm/scummvm/pull/1895/commits/9036ff37d16711f79a9510fa6cc48cbbb9321ff0

Maybe there's some conflict with that? I'm really under-qualified on this topic to say for sure.

sluicebox:
Tested with default ScummVM settings on Win7 (not OpenGL mode) with SDL2. The borders of text boxes and dropdowns are barely there or don't appear at all. 

`Setting 640 x 400 -> 640 x 400 -- 1`

master
scummvm-00000


gui-scale
scummvm-00001

Mataniko:
Overall starting to look much better - found another issue, with GMM help screen - not only the layout is off, the mouse interaction on the buttons has a weird offset to make them active

image

Mataniko:
One other big issue: mouse coordinates going in and out of the window get mangled when changing renderer

antoniou79:
The warning we get when starting Blade Runner restored content version has bad formatting and as a result has its buttons obscured (this is from a windows build, msys2/minGW 64bit)
WarningIssue

It is a bit worse on Android, but this is probably due to Android having more issues with the scaling with the code:

backends/graphics/sdl/sdl-graphics.cpp
sev-: I am not sure what you mean here, `scale` could be 1 or 2 only.

sev-:
All the currently detected rendering artefacts seems to be fixed.

What is left to do, is GUI controls for selecting the scale factor, figuring our mouse cursor position scaling and SVG images rendering.

Of course, a side task is to look into Android, so @antoniou79 we would need to team up for that since I do not have a build environment here.

All, please make another round of testing/checking, whatever issues you spot. All possible dialogs, some sitting in some engines, etc.

lotharsm:
I spotted two minor rendering issues. On Win32, using a "standard DPI" display and the regular windowed mode, the mouse cursor seems to be scaled down in a bad way:
cursor_1x_artifact
Speaking of the cursor, are there plans to vectorize it as well? (like the plans for the logo itself). This would be a nice touch imho.

When playing around with the scaling in general, sometimes I can randomly trigger a rendering issue with the dropdown buttons:
dropdown_artifact

This one is really random, I can't trigger it in a reliable way. I can't test it on a highdpi display for now unfortunately.

Mataniko:
Another issue: button highlight offset is wrong on both main and down-arrow for "add game" and the popup off set is also off from where the mouse cursor is.

image

Mataniko:
There are some additional issues when switching from hidpi to normal dpi monitor:

- [ ] Layout is not recalculated when moved to the 2nd monitor
- [ ] Scale value does not change if resizing on the regular monitor - leading to mouse coordinate issues similar to what was reported above
- [ ] Cursor looks weird - same as above

sev-:
@lotharsm yes, all graphics is going to be vectorized. That includes cursor and logo.

I fixed the dropdown buttons.

antoniou79:
A possible bug I've noticed building today, is that switching from OpenGL renderer to something like "2x' or "HQ2x" (possibly other options too), exits to desktop with a message: 
 Unsupported bpp in blitManagedSurface: 2!

Another issue I'm unsure of is the stretching options. Other than the "Fit to Window (4:3)" all the other options seem to have the same effect.

sev-:
@mgerhardy thanks, I put as a TODO for the assert() you trigger when switching to non-OpenGL.

As of the stretching options, those have effect only on the games, e.g. where resolution is fixed. Our GUI is occupying all the available resolution. So, you need to test it while running some game and then invoking GMM.

sev-:
Full Screen, Mac 2880x1800

sev-:
Full Screen Mac, 2880x1800

sev-:
On the screenshot above everything is vectorized. Fonts, buttons, widgets, logo.

Things we will work on: 

* thicker borders for empty checkbox/radiobutton.
* Still 2x cursor movement in LoDPI mode.
* Need to recheck every engine which uses custom dialogs and see that they're rendered properly.

I noticed that at the high resolutions our GUI makes a pause between switching tabs/dialogs, need to investigate what does it do, but it is out of scope for this PR.

Basically, it is ready.

sev-:
Any objections for me regarding merging this and continuing work in-tree?

What is left to do: speed up blitting routines and test all different dialogs in engines.

PR #2883, opened 2021-03-25
CLOUD: Fix format for uint64 variables

orgads:

PR #2892, opened 2021-03-26
CLOUD: Fix compiler warnings on linux64

orgads:
`uint64` is unsigned long on linux64 (first match in `configure:find_type_with_size`), so the format should be `lu`, but this is incompatible with Win64, so just cast to make it portable and warning-clean.

Amends commit 1496ad6bb5.

### DIFF
--- a/backends/cloud/cloudmanager.cpp
+++ b/backends/cloud/cloudmanager.cpp
@@ -124,7 +124,7 @@ void CloudManager::save() {
 		Common::String name = getStorageConfigName(i);
 		ConfMan.set(kStoragePrefix + name + "_username", _storages[i].username, ConfMan.kCloudDomain);
 		ConfMan.set(kStoragePrefix + name + "_lastSync", _storages[i].lastSyncDate, ConfMan.kCloudDomain);
-		ConfMan.set(kStoragePrefix + name + "_usedBytes", Common::String::format("%llu", _storages[i].usedBytes), ConfMan.kCloudDomain);
+		ConfMan.set(kStoragePrefix + name + "_usedBytes", Common::String::format("%llu", (unsigned long long)_storages[i].usedBytes), ConfMan.kCloudDomain);
 	}
 
 	ConfMan.set("current_storage", Common::String::format("%u", _currentStorageIndex), ConfMan.kCloudDomain);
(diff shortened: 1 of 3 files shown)


backends/cloud/cloudmanager.cpp
```diff
@@ -124,7 +124,7 @@ void CloudManager::save() {
 		Common::String name = getStorageConfigName(i);
 		ConfMan.set(kStoragePrefix + name + "_username", _storages[i].username, ConfMan.kCloudDomain);
 		ConfMan.set(kStoragePrefix + name + "_lastSync", _storages[i].lastSyncDate, ConfMan.kCloudDomain);
-		ConfMan.set(kStoragePrefix + name + "_usedBytes", Common::String::format("%llu", _storages[i].usedBytes), ConfMan.kCloudDomain);
+		ConfMan.set(kStoragePrefix + name + "_usedBytes", Common::String::format("%llu", (unsigned long long)_storages[i].usedBytes), ConfMan.kCloudDomain);
```
sev-, line 127: It is better use `uint64` here and in other places.

orgads: But that's exactly the problem I tried to solve... `uint64` is defined as `unsigned long` on Linux64, so `%llu` doesn't match the type. And on Windows and Linux32 it is `unsigned long long`, so `%lu` doesn't match it.

orgads:
?

sev-:
What does happen when you use `uint64` instead of `unsigned long long`?

orgads:
Compiler warning on Linux, because uint64 is unsigned long, and the format is llu.

digitall:
This casting to (unsigned long long) ensures that it matches the format %llu as per @dreammaster 's fix to the AGS engine so probably viable. The other option would be to use PRIu64 for the format. If that is missing, it could be provided by an ifndef based on the platform as per: http://web.mit.edu/freebsd/head/contrib/llvm/include/llvm/Support/DataTypes.h.in

digitall:
I mooted that solution on the earlier PR discussion for #2883 but this solution is fine with me as well..._

orgads:
Well? Merge it then?

digitall:
OK. Merging as-is. We can always migrate to a different solution in future.

PR #2896, opened 2021-03-29
TWINE: Fix struct forward-declarations

orgads:


mgerhardy:
Thanks a lot - but this is already fixed in https://github.com/scummvm/scummvm/commit/0657fdbd93e499d7a19663002c5c27a607451529

PR #2898, opened 2021-03-30
WIP: SCUMM: Fix selection x-range of verbs on Hebrew

orgads:
TODO:
* Artifacts on centered text after it should be erased.
* Short verbs on COMI are split to 2 lines, and the next verb overwrites it.
* It possibly breaks other things I haven't noticed yet.


orgads:
@BLooperZ @bluegr @ZvikaZ Do you think it's worth proceeding with this change, or try to find a solution that involves less changes?

PR #2900, opened 2021-03-30
SCUMM: Fix calculation of arrow position on Hebrew

orgads:
The text buffer starts on the i offset. This was overlooked in the original patch.

This amends commit 89c5cd5e5d81d780c74bf5da01ac6fb227e0695c.


bluegr:
Thanks, merging

PR #2902, opened 2021-03-30
MAKEFILE: Fix build without implicit rules

orgads:
Running `make -r` used `.o` (with no base name) for the compilation output of `base/version.o`.


sev-:
Thanks, good catch.

PR #2905, opened 2021-04-02
SCUMM: Add some missing includes

orgads:
Pre-included files are not reliable.

### DIFF
--- a/engines/scumm/scumm-md5.h
+++ b/engines/scumm/scumm-md5.h
@@ -3,6 +3,12 @@
   DO NOT EDIT MANUALLY!
  */
 
+#ifndef SCUMM_MD5_INTERNAL_H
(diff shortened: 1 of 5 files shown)


engines/scumm/scumm-md5.h
```diff
@@ -3,6 +3,12 @@
   DO NOT EDIT MANUALLY!
  */
 
+#ifndef SCUMM_MD5_INTERNAL_H
```
mgerhardy, line 6: For this you should edit `c_header` in `./devtools/md5table.cpp`

orgads: Thanks, done.

bluegr:
Thanks for your work on this! Merging

PR #2908, opened 2021-04-02
GRIFFON: Adding TTS to Intro Story

taylorzhancher:
By passing the words to Text To Speech Manager, this code allows TTS support to the beginning of the game.

### DIFF
--- a/engines/griffon/cutscenes.cpp
+++ b/engines/griffon/cutscenes.cpp
@@ -167,6 +167,7 @@ const char *story3 =
 	"We believed we could find"
 	"shelter here, only to find"
 	"every last griffon dead."
+	"the town burned to the ground,"
 ;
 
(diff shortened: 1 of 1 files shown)


bluegr:
Thanks for your work! However, the way this is done, there is duplicate text for the intro. You could concatenate the existing story array items dynamically into a new string, and pass that to the TTS handler instead.

Also, the commit messages do not follow our guidelines - in this case, each commit should be prefixed with GRIFFON

taylorzhancher:
@bluegr Thank you! Your instructions are helpful. I will come back with it fixed.

sev-:
It would be great if other things like menus also have voice overs.

Also, this is incomplete: this lacks the relevant GUI options, so it could not be added as is.

bluegr:
This has been redone in #2912. Closing this one

PR #2910, opened 2021-04-04
AGI: fixed typo in music volume settings

ZvikaZ:
The typo caused music and sound effects volumes values to be swapped for PCjr (but not to MIDI, because `sound_midi` doesn't call MixerImpl::getVolumeForSoundType)


bluegr:
Nice catch! Thanks for spotting and fixing it. Merging

PR #2912, opened 2021-04-04
GRIFFON: Text To Speech (TTS) for the Game

taylorzhancher:
Text To Speech (TTS) service for intro story by concatenating strings then passing into TTS handler. Texts are passed as paragraphs.

There is also a GUI for enabling TTS in the options menu of Griffon

Related Minor Fixes:
- Fixed the display of the story. In this version the sky background disappears only after all the text has scrolled over the screen. (`if (yy < 10 && i == ARRAYSIZE(story) - 1)`)
- Used `ARRAYSIZE(story)` `ARRAYSIZE(story2)` for the loops


engines/griffon/cutscenes.cpp
```diff
@@ -182,6 +186,20 @@ void GriffonEngine::showLogos() {
 }
 
 void GriffonEngine::intro() {
+	#ifdef USE_TTS
+		Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
+		const char *story1 = story[0];
+		char result[1000];
+
+		strcpy(result,story1);
```
mgerhardy, line 194: This could really need some buffer overflow checks - or use Common::String and the append operator for this.

engines/griffon/cutscenes.cpp
```diff
@@ -234,7 +252,7 @@ void GriffonEngine::intro() {
 			y--;
 		}
 
-		for (int i = 0; i <= 37; i++) {
+		for (int i = 0; i <= 47; i++) {
```
mgerhardy, line 255: Using `for (int i = 0; i < ARRAYSIZE(story); i++) {` would maybe be better - to not duplicate the size of the story array.

engines/griffon/cutscenes.cpp
```diff
@@ -182,6 +186,20 @@ void GriffonEngine::showLogos() {
...
 
 void GriffonEngine::intro() {
+	#ifdef USE_TTS
+		Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
+		const char *story1 = story[0];
+		char result[1000];
+
+		strcpy(result,story1);
+
+		for(int i=10; i<=47; ++i){
```
mgerhardy, line 196: Using for (int i = 0; i < ARRAYSIZE(story); ++i) { would maybe be better - to not duplicate the size of the story array.

engines/griffon/cutscenes.cpp
```diff
@@ -182,6 +186,20 @@ void GriffonEngine::showLogos() {
 }
 
 void GriffonEngine::intro() {
+	#ifdef USE_TTS
+		Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
+		const char *story1 = story[0];
```
mgerhardy, line 191: `_ttsMan` should get a null check - this is done in the other locations, too

engines/griffon/cutscenes.cpp
taylorzhancher: Thank you for this feedback!  I have modified it as you said, and will proceed to fix the code as you have reviewed them.

engines/griffon/cutscenes.cpp
taylorzhancher: Thank you for this feedback!  I have modified it as you said, and will proceed to fix the code as you have reviewed them.

engines/griffon/cutscenes.cpp
```diff
@@ -182,6 +186,21 @@ void GriffonEngine::showLogos() {
 }
 
 void GriffonEngine::intro() {
+	#ifdef USE_TTS
```
mgerhardy, line 189: The indention here is off - the preprocessor line should be at the beginning of the line - and the guarded code itself is one level too deep.

engines/griffon/cutscenes.cpp
```diff
@@ -182,6 +186,21 @@ void GriffonEngine::showLogos() {
...
+
+		Common::String result;
+
+		result+=story[0];
+
+		for(int i=10; i< ARRAYSIZE(story); ++i){
+			result+=story[i];
+		}
+
+		if (_ttsMan != nullptr)
```
mgerhardy, line 200: The whole code should be guarded by that null check - the constructed string is not needed if `ttsMan == nullptr`

engines/griffon/cutscenes.cpp
```diff
@@ -182,6 +186,21 @@ void GriffonEngine::showLogos() {
 }
 
 void GriffonEngine::intro() {
+	#ifdef USE_TTS
+		Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
+
+		Common::String result;
+
+		result+=story[0];
```
mgerhardy, line 194: Formatting doesn't match coding guidelines

taylorzhancher: Thanks! I have a hard time understanding what is expected in this case though. The code conventions I have consulted are here: https://wiki.scummvm.org/index.php?title=Code_Formatting_Conventions

athrxx: `result += story[0];`

engines/griffon/cutscenes.cpp
```diff
@@ -182,6 +186,17 @@ void GriffonEngine::showLogos() {
 }
 
 void GriffonEngine::intro() {
+#ifdef USE_TTS
+	Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
+	if (_ttsMan != nullptr) {
+		Common::String result = story[0];
+		for(int i=10; i< ARRAYSIZE(story); ++i){
```
mgerhardy, line 193: `for (int i = 10; i < ARRAYSIZE(story); ++i) {`

This would e..g. be the right formatting. Watch out for the missing whitespaces. See the existing code to get an idea about the formatting. Keep formatting it the same way.

Or try to use something like clang-formatter to perform the base formatting (if you ever have more than this to format)

Also if you use clang-format on a whole file, you should get used to `git add -p` to only use the relevant parts for your particular commit.

engines/griffon/cutscenes.cpp
```diff
@@ -190,7 +190,7 @@ void GriffonEngine::intro() {
 	Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
 	if (_ttsMan != nullptr) {
 		Common::String result = story[0];
-		for(int i=10; i< ARRAYSIZE(story); ++i){
+		for(int i = 10; i < ARRAYSIZE(story); ++i){
```
sev-, line 193: This is not correct. Please pay attention to the details. As it was pointed out, the correct formatting is
`for (int i = 10; i < ARRAYSIZE(story); ++i) {`

sev-:
This PR is still missing the GUI part.

sev- (changes requested):
The changes are awkwardly hardcoded, I explained how to process the story properly with a loop.

Also, I do not see changes to our GUI, only to the in-game settings which are nice thing, but different from what I requested.

Also, it would be great is the game settings also have voice-overs.

engines/griffon/cutscenes.cpp
```diff
@@ -133,6 +139,27 @@ const char *story2[27] = {
 	"and I am free to die as I please."
 };
 
+#ifdef USE_TTS
+int textToSpeech(int nextparagraph, const char *story[], int arraysize) { 
+Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
```
criezy, line 144: You seem to be missing indentation on this line (and the next ones).

engines/griffon/cutscenes.cpp
```diff
@@ -133,6 +139,27 @@ const char *story2[27] = {
...
+if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
+	if ((_ttsMan->isSpeaking()) == false) {
+		Common::String paragraph;
+		while (nextparagraph < arraysize && story[nextparagraph][0] != ' ') {
+			paragraph += story[nextparagraph];
+			paragraph += " ";
```
criezy, line 150: Doing it this way will leave a trailing space at the end of the prarapgaph. This probably doesn't matter, but this can be easily fixed by moving this at the start of the while loop if the parapraph is not empty:
```c++
	while (nextparagraph < arraysize && story[nextparagraph][0] != ' ') {
		if (!paragraph.empty())
			paragraph += " ";
		paragraph += story[nextparagraph++];
	}
```

engines/griffon/cutscenes.cpp
```diff
@@ -133,6 +139,27 @@ const char *story2[27] = {
 	"and I am free to die as I please."
 };
 
+#ifdef USE_TTS
+int textToSpeech(int nextparagraph, const char *story[], int arraysize) { 
+Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
+if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
+	if ((_ttsMan->isSpeaking()) == false) {
+		Common::String paragraph;
+		while (nextparagraph < arraysize && story[nextparagraph][0] != ' ') {
```
criezy, line 148: You could also check that `story[nextparagraph][0] != 0` to handle the empty lines in the first story array.

engines/griffon/cutscenes.cpp
```diff
@@ -233,15 +262,21 @@ void GriffonEngine::intro() {
 			cnt = 0;
 			y--;
 		}
-
-		for (int i = 0; i <= 37; i++) {
+
```
criezy, line 265: You have a line with a trailing tab here. You may want to check the settings in your text editor as it can help you avoid those (for example Xcode has a setting `Automatically trim trailing whitespace` and a separate child option `Including whitespace-only lines` and I would recommend to toggle both on.

engines/griffon/cutscenes.cpp
```diff
@@ -207,6 +234,8 @@ void GriffonEngine::intro() {
 	int cnt = 0;
 	float xofs = 0.0;
 	float ld = 0.0;
+	int nextparagraph = 10;
```
criezy, line 237: Don't we want to read the first 10 lines as well before the actual start of the story?

engines/griffon/cutscenes.cpp
```diff
@@ -133,6 +139,27 @@ const char *story2[27] = {
 	"and I am free to die as I please."
 };
 
+#ifdef USE_TTS
+int textToSpeech(int nextparagraph, const char *story[], int arraysize) { 
+Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();
+if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
+	if ((_ttsMan->isSpeaking()) == false) {
```
criezy, line 146: You probably don't want this test here because it means that if it has not finished speaking the previous paragraph it will skip this paragraph, and all the remaining ones afterward (as you test for `i == nextparagraph` to call this function).

What you want to do instead is to always execute this code and queue the speech so that it speaks this paragraph as soon as it can.

If you look at https://github.com/scummvm/scummvm/blob/master/common/text-to-speech.h you can see the following (I simplified a bit to keep only the relevant code):
```c++
class TextToSpeechManager {
public:
	enum Action {
		INTERRUPT,
		INTERRUPT_NO_REPEAT,
		QUEUE,
		QUEUE_NO_REPEAT,
		DROP
	};

	bool say(const String &str, Action action, CodePage charset = kUtf8);

	bool say(const String &str, CodePage charset = kUtf8) {
		return say(str, INTERRUPT_NO_REPEAT, charset);
	}
};
```

This means you can use one of two functions to speak some text, one where you pass only the text (the one you are currently using) and that interrupts the current speech, and another one where in addition to the text you also specify an action to indicate what you want it to do if it is currently speaking.  When using that one with the `QUEUE` action you indicate that it should continue the current speech and once it has finished immediately start speaking the new text. See the documentation at [on lines 169-178](https://github.com/scummvm/scummvm/blob/master/common/text-to-speech.h#L169-L178) for more details on the actions.

criezy (changes requested):
Overall it looks quite good now, but there are still a few things to fix.

engines/griffon/dialogs.cpp
```diff
@@ -224,27 +241,45 @@ void GriffonEngine::title(int mode) {
...
 		"", "", "",
 		"On", "Off", "",
 		"On", "Off", "",
+		"On", "Off", "",
 		"[----------]", "",
 		"[----------]", "",
 		"Exit + Save", "",
 		"Exit"
 	};
+	static const int curselMapTitles[15] = {
```
criezy, line 272: I am worried that this array might be too small and you may need to pad it with a couple of 0 to avoid out of bound access which could have unexpected consequences.

Looking at the code, it seems that both `curselMapTitles` and `curselMapValues` can be accessed with an index between `MINCURSEL` and `MAXCURSEL` inclusive. And `MAXCURSEL` is defined as 16. Since indexes start from 0, this means the arrays need to have at least 17 values to avoid an out of bound access when used with index 16.

criezy (changes requested):
As I indicated o Discord, but I will write it here as well so that it does not get lost. The engine needs to set the language to use when speaking text (in this case English) at some point. Currently it will assume whatever language is set by the GUI, which may not be English.

The way to do that is to have the following before speaking the first time:
```c++
ttsMan->pushState();
ttsMan->setLanguage("en");
```

And
```c++
ttsMan->popState();
```
When closing the engine (only if `pushState()` was used previously).

Another possible improvement that was briefly discussed on Discord is that we may want to:
 - interrupt the speech if the intro or endOfGame is interrupted
 - wait for the speech to finish when at the end of the intro and endOfGame

engines/griffon/dialogs.cpp
```diff
@@ -439,10 +450,20 @@ void GriffonEngine::configMenu() {
...
+						if (!ConfMan.getBool("tts_enabled")) {
+							ConfMan.setBool("tts_enabled", true);
+						}
+						break;
+					case 12:
+						if (ConfMan.getBool("tts_enabled")) {
+							ConfMan.setBool("tts_enabled", false);
+						}
+						break;
+					case 16:
```
criezy, line 509: I think this one is supposed to be 15, and the one below 16 (and not 17).
You only increased `MAXCURSEL` by 2, and its maximum value is 16, so `cursel` can never be 17. When you select `Exit + Save` it is 15, which does nothing as you have no case 15. And when you select `Exit`, it is 16, and it does a save and exit and not just exit.

engines/griffon/dialogs.cpp
```diff
@@ -404,12 +453,20 @@ void GriffonEngine::configMenu() {
 					cursel--;
 					if (cursel < MINCURSEL)
 						cursel = MAXCURSEL;
+					if (ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
+						ttsMan -> say(optionTitles[curselMapTitles[cursel]]);
+						ttsMan -> say(optionValues[curselMapValues[cursel]], Common::TextToSpeechManager::QUEUE);
+					}
```
criezy, line 459: This block of code, and the similar one below, does not quite work properly with the `Exit + Save` and `Exit` menus for me when I use the arrow keys quickly to go down or up in the menu. I think the issue is with calling `say()` with an empty text when `curselMapTitles[cursel]` is 0, and with queuing the following text.
What seems to work properly is to check that there is some text to speak, and only use queuing when there was an interrupt before. That would give something like this:
```c++
if (ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
	Common::TextToSpeechManager::Action valueAction = Common::TextToSpeechManager::INTERRUPT;
	if (curselMapTitles[cursel] != 0) {
		ttsMan->say(optionTitles[curselMapTitles[cursel]]);
		valueAction = Common::TextToSpeechManager:: QUEUE;
	}
	if (curselMapValues[cursel] != 0)
		ttsMan->say(optionValues[curselMapValues[cursel]], valueAction);
}
```

Also note that we are not expecting a space before and after `->` (so it should be `ttsMan->say` and not `ttsMan -> say`).

engines/griffon/cutscenes.cpp
```diff
@@ -234,15 +259,23 @@ void GriffonEngine::intro() {
...
 				drawString(_videoBuffer, story[i], x, yy, 4);
 			}
-
-			if (yy < 10 && i == 37)
-				return;
+			if (yy < 10 && i == ARRAYSIZE(story) - 1) {
+				if (ttsMan != nullptr)
+					if (ttsMan->isSpeaking() == false)
+						return;
+			}
```
criezy, line 278: That looks strange. Now we return only if `ttsMan` is not nullptr, and `ttsMan->isSpeaking()` is false. It seems to me that it should return in the case where `ttsMan` is nullptr.
So I would expect a test like:
```c++
if (yy < 10 && i == ARRAYSIZE(story) - 1) {	
	if (ttsMan == nullptr || ttsMan->isSpeaking() == false)
		return;
}
```

engines/griffon/cutscenes.cpp
```diff
@@ -367,15 +407,25 @@ void GriffonEngine::endOfGame() {
...
 				drawString(_videoBuffer, story2[i], x, yy, 4);
 			}
 
-			if (yy < 10 && i == 25)
-				break;
+			if (yy < 10 && i == ARRAYSIZE(story2)-1) {
+				if (ttsMan != nullptr)
+					if (ttsMan->isSpeaking() == false)
+						break;
+			}
```
criezy, line 428: Same comment as above. I would expect to break as well is `ttsMan` is a nullptr.

criezy (changes requested):
There are a few more issues that need to be fixed before this can be merged.

engines/griffon/cutscenes.cpp
criezy: Why change the `break` into a `return` here? Is this a mistake?

criezy (commented):
That looks good now. I only have one comment left on a `break` that you changed into a `return` in 027c012. To me it looks like this might be a mistake due to doing a copy/paste of similar code.

criezy (approved):
This looks good to me now.

The history is a bit messy now, but I think we can just squash everything. So I don't see a need to cleanup the history.

I will wait a couple of days in case somebody else wants to review the final changes, and if there is no new request to make changes I will squash+merge the pull request.

criezy:
Time to squash and merge...

PR #2914, opened 2021-04-05
DONT MERGE - WIN32: Use Unicode for paths

ZvikaZ:
Before this commit, Windows returned path names with local computer encoding, and then ScummVM converted them to `U32String`, without knowing what is the proper encoding.

This commit ensures that Windows returns Unicode paths.

When scummvm.ini is opened in GVim, the Hebrew path isn't rendered correctly. However, `:e! ++enc=utf8` makes it show correctly. I'm not certain about this - is it normal? Or should we so something for scummvm.ini so editors will automatically know that it's UTF8? (maybe add a `BOM`? but as far as I know, it's frowned upon in the Linux world)

EDIT
-----
Codacy fails, with:

	backends/fs/windows/windows-fs.cpp

	Does not handle strings that are not \0-terminated; if given one it may perform an over-read (it could cause a crash if unprotected) (CWE-126).
	WideCharToMultiByte(CP_ACP, 0, str, _tcslen(str) + 1, asciiString, sizeof(asciiString), NULL, NULL);

	Does not handle strings that are not \0-terminated; if given one it may perform an over-read (it could cause a crash if unprotected) (CWE-126).
	MultiByteToWideChar(CP_ACP, 0, str, strlen(str) + 1, unicodeString, sizeof(unicodeString) / sizeof(TCHAR));

However, I have only changed the codepage. The rest of these lines is similar to before. Do we want to change something in these lines? Or can we safely ignore these warnings?

(BTW, it's weird that Codacy complains about the **original** lines, before my changes - I have modified `CP_ACP` to `CP_UTF8`)

EDIT2
-----
grepping found that there's another instance of `CP_ACP` usage, at `engines/ags/shared/util/path.cpp`
I assume it should also be changed to `CP_UTF8`. Either as part of this PR, or maybe after its acceptance. However, I'm completely unfamiliar with AGS engine. @PaulGilbert , what do you say?

EDIT3
-----
We should be aware that this PR doesn't provide safe migration for users that already have non ascii characters in their game paths, as they are encoded in the local encoding, which ScummVM isn't aware of. However, this PR still *improves* the situation, because anyway, currently these games simply won't work as ScummVM doesn't understand the old path encoding. So, either with this PR, or without it, we're not backward compatible. This PR at least allows the users to re-add those games.

### DIFF
--- a/backends/fs/windows/windows-fs.cpp
+++ b/backends/fs/windows/windows-fs.cpp
@@ -43,16 +43,17 @@
 #define W_OK 2
 #endif
 
+
 bool WindowsFilesystemNode::exists() const {
-	return _access(_path.c_str(), F_OK) == 0;
+	return _taccess(toUnicode(_path.c_str()), F_OK) == 0;
 }
 
 bool WindowsFilesystemNode::isReadable() const {
-	return _access(_path.c_str(), R_OK) == 0;
+	return _taccess(toUnicode(_path.c_str()), R_OK) == 0;
 }
 
 bool WindowsFilesystemNode::isWritable() const {
-	return _access(_path.c_str(), W_OK) == 0;
+	return _taccess(toUnicode(_path.c_str()), W_OK) == 0;
 }
 
 void WindowsFilesystemNode::addFile(AbstractFSList &list, ListMode mode, const char *base, bool hidden, WIN32_FIND_DATA* find_data) {
@@ -91,7 +92,7 @@ char* WindowsFilesystemNode::toAscii(TCHAR *str) {
 	return (char *)str;
 #else
 	static char asciiString[MAX_PATH];
-	WideCharToMultiByte(CP_ACP, 0, str, _tcslen(str) + 1, asciiString, sizeof(asciiString), NULL, NULL);
+	WideCharToMultiByte(CP_UTF8, 0, str, _tcslen(str) + 1, asciiString, sizeof(asciiString), NULL, NULL);
 	return asciiString;
 #endif
 }
@@ -101,7 +102,7 @@ const TCHAR* WindowsFilesystemNode::toUnicode(const char *str) {
 	return (const TCHAR *)str;
 #else
 	static TCHAR unicodeString[MAX_PATH];
-	MultiByteToWideChar(CP_ACP, 0, str, strlen(str) + 1, unicodeString, sizeof(unicodeString) / sizeof(TCHAR));
+	MultiByteToWideChar(CP_UTF8, 0, str, strlen(str) + 1, unicodeString, sizeof(unicodeString) / sizeof(TCHAR));
 	return unicodeString;
 #endif
 }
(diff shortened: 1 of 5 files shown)


sluicebox:
I can't speak to the existing encoding situation or its problems, but unconditionally defining UNICODE is a problem.

windows-fs.cpp is full of `#ifndef UNICODE` code and this change unconditionally defines UNICODE a few lines earlier so now all that code is dead. That doesn't sound right.

UNICODE (and _UNICODE) should be consistently defined/undefined across an entire program and not per cpp. Otherwise TCHAR usage in headers (windows-fs.h, SDK headers, etc) could be char or wchar_t depending on which cpp gets compiled first. If you're unconditionally defining UNICODE then you don't need TCHARs at all because that gives the illusion of char/wchar_t compatibility when it doesn't exist, but you don't want to be unconditionally defining UNICODE in source at all because...

Most importantly, this will break the beloved Win98 build. (Okay maybe just @SupSuper loves it.) The Win32 API calls that take strings will now always link to the wide-char versions.

I'm not familiar with any of this code, just saw that the change started with `#define UNICODE` and had Win32 flashbacks. I don't know if this is relevant but since we have code to handle when UNICODE is defined, I'd want to know under what circumstances that occurs.

ZvikaZ:
> I can't speak to the existing encoding situation or its problems, but unconditionally defining UNICODE is a problem.
> 
> windows-fs.cpp is full of `#ifndef UNICODE` code and this change unconditionally defines UNICODE a few lines earlier so now all that code is dead. That doesn't sound right.
> 
> UNICODE (and _UNICODE) should be consistently defined/undefined across an entire program and not per cpp. Otherwise TCHAR usage in headers (windows-fs.h, SDK headers, etc) could be char or wchar_t depending on which cpp gets compiled first. If you're unconditionally defining UNICODE then you don't need TCHARs at all because that gives the illusion of char/wchar_t compatibility when it doesn't exist, but you don't want to be unconditionally defining UNICODE in source at all because...
> 
> Most importantly, this will break the beloved Win98 build. (Okay maybe just @SupSuper loves it.) The Win32 API calls that take strings will now always link to the wide-char versions.
> 
> I'm not familiar with any of this code, just saw that the change started with `#define UNICODE` and had Win32 flashbacks. I don't know if this is relevant but since we have code to handle when UNICODE is defined, I'd want to know under what circumstances that occurs.

That's a good point!
I have moved the (_)UNICODE defines to create_project. @SupSuper , is it enough for Win98 support, or should I find another place to put them?

Anyway, now there are many compilation errors, as it seems we have a lot of code that depends on UNICODE being not defined.

I assume I can fix them all, but I'd like first to be sure that I'm on the right path, before investing time and energy in this...

bluegr:
So, this approach is problematic, because UNICODE is unconditionally defined.
The approach followed in PR #2929 does not have this problem, so this one can be closed

PR #2880, opened 2021-03-23
GRIM: add RTL support for hebrew

orgads:

PR #2917, opened 2021-04-05
GRIM: improve RTL support for Hebrew fan translation

BLooperZ:
This amends #2880

### DIFF
--- a/engines/grim/textobject.cpp
+++ b/engines/grim/textobject.cpp
@@ -20,6 +20,7 @@
  *
  */
 
+#include "common/unicode-bidi.h"
 #include "engines/grim/debug.h"
 #include "engines/grim/grim.h"
 #include "engines/grim/textobject.h"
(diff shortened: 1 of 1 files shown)


orgads:
Can you squash the commits?

BLooperZ:
squashed

engines/grim/textobject.cpp
```diff
@@ -278,16 +279,11 @@ void TextObject::setupText() {
 			nextLinePos = message.size();
 			cutLen = nextLinePos;
 		}
-		Common::String currentLine;
+		Common::String currentLine(message.c_str(), message.c_str() + nextLinePos);
```
orgads, line 282: Needless copy for RTL (and the comment is now redundant). Suggestion:

```c++
		Common::String currentLine;
		if (g_grim->getGameLanguage() == Common::HE_ISR)
			currentLine = Common::convertBiDiString(currentLine, Common::kWindows1255);
		else
			currentLine = Common::String(message.c_str(), message.c_str() + nextLinePos);
```

orgads (changes requested):

engines/grim/textobject.cpp
BLooperZ: please note that coverBiDiString requires the created string (as first parameter), so the copy is needed

orgads: Ah, right. Sorry for the noise :)

orgads (approved):

orgads:
Thank you. Merging.

PR #2919, opened 2021-04-05
MAKEFILE: Add dependencies for scummvm.rc

orgads:
When an engine is added or removed, it affects the embedded resources in
scummvm.rc, so config.mk is required.

Apart from that, list all the files it uses and set them as dependencies,
so if a file changes, the resource file will be rebuilt.

### DIFF
--- a/backends/platform/sdl/win32/win32.mk
+++ b/backends/platform/sdl/win32/win32.mk
@@ -2,7 +2,17 @@
 # Windows specific
 #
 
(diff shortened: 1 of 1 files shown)


orgads:
Don't merge yet. I need to run some more tests.

orgads:
@ccawley2011 @digitall (others are welcome too ;))

This change currently lists all the required dependencies. Do you think it is valuable to have automatic tracking? I have a preliminary proposal in #2921 but I'm a bit reluctant about it. See my comment there.

What do you think?

digitall:
@orgads: This change looks reasonable. I would suggest that it is not an exclusive-or situation. This PR should be merged and tested as an improvement... then the merits and issues with PR #2921 for automatic dependency tracking can be dealt with.

orgads:
I agree. Merging then.

PR #2923, opened 2021-04-06
IMAGE: Gif decoder using libgif

mgerhardy:
This adds an external dependency to libgif - but also supports palettes for gifs. Overall it might have more features than what `stb_image` is offering right now.

Also see the `stb_image` PR: https://github.com/scummvm/scummvm/pull/2911 (Obviously only one should get merged)

- [x] Added configure support for libgif
- [x] Extended create_project to add libgif support
- [x] Verified functionality in TwinE
- [x] Added unittest for loading a small 2x2 gif buffer
- [x] Updated github actions and travis packages
- [x] Add libgif to buildbot after merging

This already adds full support for the floppy disk `fla_gif.hqr` version of Little Big Adventure 1


rsn8887:
Why are you changing gitignore?

test/image/gif.h
```diff
@@ -0,0 +1,41 @@
...
+#if defined(HAVE_CONFIG_H)
+#include "config.h"
+#endif
+
+#include "common/memstream.h"
+#include "image/gif.h"
+#include "graphics/surface.h"
+
+class GIFDecoderTestSuite : public CxxTest::TestSuite {
+#ifdef USE_GIF
```
lephilousophe, line 12: OK, this fails on platforms without USE_GIF because the test is nonexistent anymore.
The cxxtestgen.py script doesn't track #ifdef and when it generates a runTests method, it tries to use this test. This obviously fails.
Like in test/common/array.h for USE_CXX11, I think it's simpler to move the #ifdef/#endif inside the function.

mgerhardy: Sorry for the trouble - should be fixed now.

lephilousophe: Working now.

lephilousophe:
Buildbot is ready. Merging now.

PR #2924, opened 2021-04-06
OpenGL: Fix warning for hidden overladed virtual saveScreenshot() function

criezy:
This reverts commit 0f118d0210d1dc1d4e4174fd72179de82b263ed8.
Instead it attempt to fix the `hiding overloaded virtual function` warning in a cleaner way.

### DIFF
--- a/backends/graphics/opengl/opengl-graphics.h
+++ b/backends/graphics/opengl/opengl-graphics.h
@@ -265,31 +265,6 @@ protected:
 	 */
 	bool saveScreenshot(const Common::String &filename) const;
 
-	/**
-	 * @brief Dummy implementation of saveScreenshot() (ie. no file argument) for OpenGLGraphicsManager
-	 *
-	 * Currently we have a backend OpenGLSDLGraphicsManager, which inherits from both
-	 * OpenGLGraphicsManager and SdlGraphicsManager
-	 *
-	 * OpenGLSDLGraphicsManager "implements" saveScreenshot(const Common::String &filename) (can be overridden)
-	 * by calling the saveScreenshot(const Common::String &filename) of OpenGLGraphicsManager
-	 * and saveScreenshot() by calling saveScreenshot() of SdlGraphicsManager
-	 *
-	 * SdlGraphicsManager provides an implementation of saveScreenshot() (which can be overridden)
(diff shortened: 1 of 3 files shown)


antoniou79:
> This reverts commit [0f118d0](https://github.com/scummvm/scummvm/commit/0f118d0210d1dc1d4e4174fd72179de82b263ed8).
> Instead it attempt to fix the `hiding overloaded virtual function` warning in a cleaner way.

Indeed it fixes the warning and it's cleaner. 
I kind of think my tentative fix was more symmetrical, but this is cleaner. 

Still I don't think this addresses the mess/confusion with the two signatures and why we are keeping them separate in this providing implementations for one and not the other, when one (the one with the file argument) can basically use the other (no argument signature) in every (?) case.

criezy:
> when one (the one with the file argument) can basically use the other (no argument signature)

That's rather the other way around. The one without argument can call the one with argument passing the default path for a new screenshot file. The issue is that currently this "default path" is backend-dependent. In the SDL backend you have `getScreenshotsPath()`, but that is currently not available in non-SDL backends.

The main issue here is that the `saveScreenshot` that takes an argument is only really an implementation detail in the SDL backend. But it gets exposed to non-SDL backends as well. While on the other hand `getScreenshotsPath()` is only available in the SDL backend.

So I agree that this could probably be cleaned. But I don't see an obvious way to do it.

antoniou79:
> > when one (the one with the file argument) can basically use the other (no argument signature)
> 
> That's rather the other way around. The one without argument can call the one with argument passing the default path for a new screenshot file. The issue is that currently this "default path" is backend-dependent. In the SDL backend you have `getScreenshotsPath()`, but that is currently not available in non-SDL backends.
> 
Yes, that's correct. I've kind of rethought that sentence (hence the edit) and the typical approach is what you describe.

> The main issue here is that the `saveScreenshot` that takes an argument is only really an implementation detail in the SDL backend. But it gets exposed to non-SDL backends as well. While on the other hand `getScreenshotsPath()` is only available in the SDL backend.
> 
> So I agree that this could probably be cleaned. But I don't see an obvious way to do it.

I completely understand. For now, this PR will do.

criezy:
Thanks for confirming this fixes your warning.
I am merging this now and we can try to improve the code later.

PR #2926, opened 2021-04-07
DOCS: Fix misprint for iconspath

antoniou79:
The configuration key "iconpath" is actually "iconspath"

Also, there's more to explain about this feature, but it would probably bloat the details table cell for a minor feature that probably not many people use. 

Anyway, for reference the code seems to accept naming formats of:
 * (gameid).(extension), where gameid is as it appears in scummvm.ini for the game entry
 * (engineid)-(gameid).(extension), where engineid and gameid are as they appear in scummvm.ini for the game entry
 * (gameTarget).(extension), where gameTarget is the "tag" as it appears in scummvm.ini for this game's entry section, which is also what is exposed to the ScummVM launcher GUI as "ID". (Our documentation currently actually points to this one).

If the _iconspath_ is empty or missing, the code will also look for an _extrapath_ --another configuration key, but one which is exposed to the ScummVM launcher GUI, and has a global value and a per-game entry, under Options->Paths (tab)->Extra Path (field) or Edit Game... ->Paths (tab)->Extra Path (field) ).

Finally, within the path specified either in iconspath or the extrapath (if the iconspath is empty/undefined) the icons can either reside directly into that folder path or within a subfolder named "icons".

This is based on the source code in file: common/taskbar.h

### DIFF
--- a/doc/docportal/advanced_topics/configuration_file.rst
+++ b/doc/docportal/advanced_topics/configuration_file.rst
@@ -244,7 +244,7 @@ There are many recognized configuration keys. In the table below, each key is ei
 		":ref:`herculesfont <herc>`",boolean,false,
 		":ref:`hpbargraphs <hp>`",boolean,true,
 		":ref:`hypercheat <hyper>`",boolean,false,
-		iconpath,string,, "Specifies the path to icons to use as overlay for the ScummVM icon in the Windows taskbar or the macOS Dock when running a game. The icon files should be named after the :ref:`gameid <gameid>`, and be an ICO file on Windows, or a PNG file on macOS."
+		iconspath,string,, "Specifies the path to icons to use as overlay for the ScummVM icon in the Windows taskbar or the macOS Dock when running a game. The icon files should be named after the :ref:`gameid <gameid>`, and be an ICO file on Windows, or a PNG file on macOS."
 		":ref:`improved <improved>`",boolean,true,
 		":ref:`InvObjectsAnimated <objanimated>`",boolean,true,
 		":ref:`joystick_deadzone <deadzone>`",integer, 3


cadihowley (approved):
Ok... so there's more than one thing wrong in the docs then - as you said, gameid is not the same as what's displayed in the GUI ID field, and therefore those two shouldn't be linked in the docs! Another item on the to-do list...but for now, we could probably try and give as much info for iconspath as possible?

What about:
````
Icon files can follow any of these naming conventions, where xxx is the file extension:
  - gameid.xxx
  - engineid-gameid.xxx
  - gametarget.xxx. Gametarget is the name of the game listed in square brackets in the configuration file, or in the :ref:`ID <gameid>` field in the GUI.
Icons must be ICO files on Windows, and PNG files on macOS.

```````

Or is that too much? Happy to merge just the fix for the spelling mistake for now!

criezy:
@cadihowley Your proposed change for the text looks fine to me.
An alternative to adding that text to the table would be to add a small section on taskbar/dock integration and link to it from the table. The taskbar integration has several features:
 - Add an overlay icon in the taskbar / dock when running a game (macOS and Windows).
 - Update a list of recently played games that can be started from the taskbar / doc (macOS and Windows).
 - Show a progress when doing a mass add of games, and the number of games found at the end (Linux, macOS, and Windows).
 - Show an error state in the taskbar if an error occurs when running a game (only on Windows).

I made a blog post a while ago that has images and a screen recording of those in action: https://criezy.blogspot.com/2016/03/nsdocktileplugin-for-scummvm-on-os-x.html

And maybe it would be worth adding a link to https://github.com/scummvm/scummvm-icons in the documentation?

Note that on both Windows and macOS, the icons are also stored in the list of recent games. But on macOS at least it is not used (the list of recent items in the doc is only the game names with no icons - as visible in my blog post linked above). I don't know on Windows if the icon actually appears in the recent games list. So for now it is probably best to not mention it (unless somebody can confirm that the icons indeed appear there).

cadihowley:
@criezy That's awesome information, thank you. I think it's worth adding a section in the docs!

sev-:
Good change, merging.

PR #2929, opened 2021-04-08
WIN32: Use UTF8 instead of local codepage

ZvikaZ:
Third attempt to solve the same problem as https://github.com/scummvm/scummvm/pull/2914 and https://github.com/scummvm/scummvm/pull/2928

This one is quite similar to https://github.com/scummvm/scummvm/pull/2914 , but without using `define UNICODE`.

@sluicebox , at #2914  you've mentioned problems with Win98. Do you think this PR is also problematic with Win98? What's the exact problem? Doesn't Win98 support wide chars? But we already have such usage in our codebase, at backends\platform\sdl\win32\win32_wrapper.cpp - so, is it OK? Or maybe that file isn't used at Win98?

EDIT
-----
I've described the problem, and my analysis, at https://bugs.scummvm.org/ticket/12409

### DIFF
--- a/backends/dialogs/win32/win32-dialogs.cpp
+++ b/backends/dialogs/win32/win32-dialogs.cpp
@@ -92,7 +92,7 @@ HRESULT getShellPath(IShellItem *item, Common::String &path) {
 	LPWSTR name = NULL;
 	HRESULT hr = item->GetDisplayName(SIGDN_FILESYSPATH, &name);
 	if (SUCCEEDED(hr)) {
-		char *str = Win32::unicodeToAnsi(name);
+		char *str = Win32::winUnicodeToAscii(name);
 		path = Common::String(str);
 		CoTaskMemFree(name);
 		free(str);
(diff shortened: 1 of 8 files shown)


ZvikaZ:
I've implemented now @SupSuper 's suggestion in Discord to use the safer `unicodeToAnsi` from win32_wrapper.cpp. I've also renamed it, to better reflect its nature.

bluegr:
Thanks for your work!

PR #3003 changes the same parts as this one, but it modifies more parts of the codebase, such as the save game code, the log saving code, the functionality that opens URLs etc.

Also, in the other PR, codepages are handled internally in `ansiToUnicode()` and `unicodeToAnsi()`, which is a better approach, IMHO.

Futhermore, in the other PR, the Unicode functionality is handled as a new feature that can be enabled at compile time, which allows it to be disabled for older non-Unicode systems.

Thus, overall, I believe that the approach taken in PR #3003 to tackle the issue at hand is more feature complete and can work in older systems. Since we can go with one of the two approaches, I shall close this one, so we can check, modify and merge one attempt at this.

For the reasons mentioned above, I'm closing this PR, and we can continue the discussion of the issue described in bug 12409 in PR #3003.

Again, thanks for your work on this, including for the effort you did in identifying the issue, reproducing it and proposing different solutions for it.

Closing, and we can continue the discussion in the implementation done in PR #3003

PR #2930, opened 2021-04-08
GUI: Fix caching of Edit Game dialog layout

DivyamAhuja:
Since different engines may have different number of options we have to recreate "Edit Game" dialog layout every time and avoid caching.


sev-:
Good finding, thank you!

PR #2938, opened 2021-04-10
TSAGE: Support Blue Force Spanish

jdieguez:
This change has two parts

- make the ScummVM TSAGE engine not crash with the Spanish version of Blue Force.
- translate all remaining bits (taken from Spanish BLUE.EXE).

More info provided in the individual commits.


dreammaster:
Looks good to me, except for one minor formatting issue. Could you amend the commits so that } else occurs on the same line rather than on separate lines.

jdieguez:
> Looks good to me, except for one minor formatting issue. Could you amend the commits so that } else occurs on the same line rather than on separate lines.

Done! Sorry for not noticing it myself.

PR #2942, opened 2021-04-11
AUDIO: Add support for 24 bit PCM WAVE

criezy:
Some AGS games use 24 bit WAVE audio. Currently this does not work as our code only supports 8 and 16 bit PCM, but not 24 bit PCM. This is an attempt to implement this. Since this is my first foray into the audio code (and it includes some interesting template code), I would appreciate some review to make sure it looks OK.

This was tested with The Rail (https://www.adventuregamestudio.co.uk/site/games/game/1670-the-rail/), that uses 8, 16, and 24 bit PCM WAVE audio.

### DIFF
--- a/audio/decoders/raw.cpp
+++ b/audio/decoders/raw.cpp
@@ -35,8 +35,8 @@ namespace Audio {
 // us to go with the macro approach. So far this is
 // the only template function that MSVC6 seemed to
 // compile incorrectly. Knock on wood.
-#define READ_ENDIAN_SAMPLE(is16Bit, isUnsigned, ptr, isLE) \
-	((is16Bit ? (isLE ? READ_LE_UINT16(ptr) : READ_BE_UINT16(ptr)) : (*ptr << 8)) ^ (isUnsigned ? 0x8000 : 0))
+#define READ_ENDIAN_SAMPLE(bytesPerSample, isUnsigned, ptr, isLE) \
+	((bytesPerSample == 1 ? (*ptr << 8) : (bytesPerSample == 2 ? (isLE ? READ_LE_UINT16(ptr) : READ_BE_UINT16(ptr)) : (int16)((isLE ? READ_LE_UINT24(ptr) : READ_BE_UINT24(ptr)) >> 8))) ^ (isUnsigned ? 0x8000 : 0))
 
 
 #pragma mark -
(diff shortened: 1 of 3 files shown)


audio/decoders/raw.cpp
```diff
@@ -35,8 +35,8 @@ namespace Audio {
 // us to go with the macro approach. So far this is
 // the only template function that MSVC6 seemed to
 // compile incorrectly. Knock on wood.
-#define READ_ENDIAN_SAMPLE(is16Bit, isUnsigned, ptr, isLE) \
-	((is16Bit ? (isLE ? READ_LE_UINT16(ptr) : READ_BE_UINT16(ptr)) : (*ptr << 8)) ^ (isUnsigned ? 0x8000 : 0))
+#define READ_ENDIAN_SAMPLE(bytesPerSample, isUnsigned, ptr, isLE) \
+	((bytesPerSample == 1 ? (*ptr << 8) : (bytesPerSample == 2 ? (isLE ? READ_LE_UINT16(ptr) : READ_BE_UINT16(ptr)) : (int16)((isLE ? READ_LE_UINT24(ptr) : READ_BE_UINT24(ptr)) >> 8))) ^ (isUnsigned ? 0x8000 : 0))
```
lephilousophe, line 39: This is barely readable.
As the macro is used at only one place, why not putting it back where it's used and make it multiline with proper if structures?

criezy: I can do that. I was not sure it was safe because of the comment above that macro:
> This used to be an inline template function, but buggy template function handling in MSVC6 forced  us to go with the macro approach. So far this is  the only template function that MSVC6 seemed to  compile incorrectly. Knock on wood.

That made me unsure whether moving that to where it is used, which is a template function, would work in MSVC6. But at the same time I don't know if anybody is still using that compiler.

criezy: I have now pushed a commit that does that. If we think this is the way to go then the two commits can be squashed.

lephilousophe: I think the MSVC6 error was due to template function. As it is now, it's only some lines of code in the body of a regular function so I think it's safe.
And it's now really cleaner. :D Thanks.

bluegr:
The changes are great, and the usage of templates is nicely implemented. Kudos! +1 from me

sev-:
Thank you, all looks proper. Merging.

PR #2946, opened 2021-04-12
GRAPHICS: MACGUI: Window border drawing fixes

ysj1173886760:
* move title box drawing to 9-patch image
* use lazyLoad to load built-in borders
* move border stuff into macwindowborder


graphics/macgui/macwindowborder.h
```diff
@@ -29,9 +29,23 @@
 #include "graphics/nine_patch.h"
 #include "graphics/managed_surface.h"
 #include "graphics/transparent_surface.h"
+#include "graphics/macgui/macfontmanager.h"
```
sev-, line 32: In general, avoid using includes in other includes unless absolutely necessary.

I see that these three added includes could go to macwindowborder.cpp

Also, please remove nine_patch.h from here and use forward declaration instead, e.g. add: `class NinePatchBitmap;` after you forward-declare `MacWindow`, because you are using only pointers here.

This way, compilation goes faster and less files get recompiled when you change a header.

graphics/macgui/macwindowborder.h
```diff
@@ -29,9 +29,23 @@
...
+#include "graphics/primitives.h"
+
+#include "image/bmp.h"
 
 namespace Graphics {
 
+class MacWindow;
+
+enum {
+	kWindowBorderActive = 1 << 0,
```
sev-, line 42: Align these horizontally by adding spaces, so `=` are all in the same column.

graphics/nine_patch.cpp
```diff
@@ -113,21 +127,29 @@ bool NinePatchSide::init(Graphics::TransparentSurface *bmp, bool vertical) {
...
-void NinePatchSide::calcOffsets(int len) {
+void NinePatchSide::calcOffsets(int len, int titleIndex, int titleWidth) {
 	uint i, j;
 	int dest_offset = 0;
-	int remaining_stretch = len - _fix;
+	// if we don't got titleIndex, then we better set titleWidth to 0
+	if (titleIndex == 0) titleWidth = 0;
+	int remaining_stretch = len - _fix - titleWidth;
+	// if titleWidth is too big, which may cause the remaining_stretch be 0, so we check it here
+	if (remaining_stretch < 0) remaining_stretch = 0;
```
sev-, line 137: Please split this and the `if` above into two lines.

graphics/nine_patch.cpp
```diff
@@ -208,11 +239,18 @@ NinePatchBitmap::NinePatchBitmap(Graphics::TransparentSurface *bmp, bool owns_bi
...
+	_titleWidth = titleWidth;
+	_h.calcOffsets(_cached_dw, _titleIndex, _titleWidth);
+}
+
 void NinePatchBitmap::blit(Graphics::Surface &target, int dx, int dy, int dw, int dh, byte *palette, int numColors, MacWindowManager *wm, uint32 transColor) {
 	/* don't draw bitmaps that are smaller than the fixed area */
 	if (dw < _h._fix || dh < _v._fix)
 		return;
 
+	if (dw < _h._fix + _titleWidth) dw = _h._fix + _titleWidth;
```
sev-, line 253: The same thing, try to avoid statements like this, they are not easy to read. Split it into two lines.

sev- (commented):
Very good job. Only few minor notes.

sev-:
Thank you! Merging

PR #2947, opened 2021-04-12
BACKENDS: Fix incorrect cursor scaling on high-DPI Windows

SupSuper:
This tries to correct handling coordinates from mouse events and window events on high-DPI systems:
- on some platforms (eg. Windows), mouse and window events need to be unscaled 
- on some platforms (eg. OSX), mouse and window events need to be scaled by the DPI

We try to determine this by comparing the WindowSize and DrawableSize and scaling appropriately.

Only tested on Windows, appreciate testing and feedback from other systems.

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -289,10 +289,10 @@ void OpenGLSdlGraphicsManager::updateScreen() {
 void OpenGLSdlGraphicsManager::notifyVideoExpose() {
 }
 
-void OpenGLSdlGraphicsManager::notifyResize(const int w, const int h) {
+void OpenGLSdlGraphicsManager::notifyResize(const int width, const int height) {
 #if SDL_VERSION_ATLEAST(2, 0, 0)
-	// We sometime get outdated resize events from SDL2. So check that the size we get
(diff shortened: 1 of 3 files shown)


OMGPizzaGuy:
Tested on Windows 10 - display scaled to 175%
Seems much better on both windowed and full screen mode. Mouse cursor tracks better in windowed transitioning from in-and-out of bounds. Full screen no longer loses the cursor on right or bottom edge.
Windowed mode does start with a pretty small window by default still, but resize does occur a moment after changing the window size

OMGPizzaGuy:
Text scaling gets a bit odd when stretch mode is Fit to Window (4:3)
image

bluegr:
Thanks! This is definitely an improvement on Windows, and makes the UI usable again.
The mouse coordinates are correct
I'm using Windows 10 64-bit with a 4K resolution and a 150% system scaling.

There's another issue, although it's not that important: if I maximize the window and then restore it, the ScummVM logo's dimensions do not adjust back correctly. It gets adjusted if I manually drag the window borders and change its size.

lephilousophe:
> Text scaling gets a bit odd when stretch mode is Fit to Window (4:3)
> image

I saw this one before the hidpi PR was merged. I don't think it's related.

sev-:
Thanks!

PR #2948, opened 2021-04-12
GUI: Text display issue when editing text 

Rohit-v-raj:

### DIFF
--- a/gui/widgets/editable.cpp
+++ b/gui/widgets/editable.cpp
@@ -340,7 +340,7 @@ void EditableWidget::drawCaret(bool erase) {
 		// possible glitches due to different methods used.
 		width = MIN(editRect.width() - caretOffset, width);
 		if (width > 0) {
-			g_gui.theme()->drawText(Common::Rect(x, y, x + width, y ), character,
+			g_gui.theme()->drawText(Common::Rect(x, y, x + width, y), character,
 			                        _state, _drawAlign, _inversion, 0, false, _font,
 			                        ThemeEngine::kFontColorNormal, true, _textDrawableArea);
 		}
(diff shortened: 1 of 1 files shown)


gui/widgets/editable.cpp
```diff
@@ -340,7 +340,7 @@ void EditableWidget::drawCaret(bool erase) {
 		// possible glitches due to different methods used.
 		width = MIN(editRect.width() - caretOffset, width);
 		if (width > 0) {
-			g_gui.theme()->drawText(Common::Rect(x, y, x + width, y + editRect.height()), character,
+			g_gui.theme()->drawText(Common::Rect(x, y, x + width, y ), character,
```
sev-, line 343: This change does not follow our Code Formatting conventions. Please, pay attention to the spaces around braces.

sev-:
The commit log message does not conform to our Commit Guidelines.
The code change itself breaks the Code Formatting Guidelines.

The PR is not descriptive, what is "the issue"?

Rohit-v-raj:
When editing some text, for example in the search box, of the game name in the Edit Game dialog, the letter immediately after the text cursor is displayed lower than the rest of the text.

criezy:
See https://wiki.scummvm.org/index.php?title=Commit_Guidelines for the commit guidelines. And pay special attention to the commit message formatting section.

You can use interactive rebase with git to fix your commit message (and squash the two commits) and then force push to your fork. This will automatically update this pull request.

criezy (changes requested):
Your commit is not a proper fix for the issue.
What I suspect it does is that it no longer redraw the character when the caret is blinking (since the draw rect height is now 0). But there is a reason why we are drawing the character here after erasing the caret. What you need to find is why the character is not drawn at the correct position and fix that position.

criezy:
@Rohit-v-raj Since there was no reply from you my request to change this pull request, and went ahead and fixed the bug myself. You can see the correct fix, and the reason of the bug in ba0dabd5c.

PR #2950, opened 2021-04-12
ACCESS: Support Amazon Spanish in access.dat R/W

jdieguez:
I assume we want to use stable language codes in access/resources.cpp
even if language.h keeps changing, so I've used language code "23" for
ES_ESP since this was the enum value in language.h the last time
access.dat was generated.

Note: we still need access.dat to be regenerated for the game to work,
but create_access should now be able to extract the needed info
from the Spanish version of AMAZON.EXE and the game should run fine
afterwards.


dreammaster:
Thanks, merging in. If you could contact me via Discord or email, I'd like to co-ordinate generating a fresh access.dat

PR #2951, opened 2021-04-13
SAGA: Fix a hitzone detection bug in ITE

a-yyg:
At Alamma's cottage, the wrong object's hitzone is selected when using it with an item (cottage door's 8295 vs door's 24578). In order to make it more easy to progress, we reset the objectId to ID_NOTHING in this case.
-This fixes [Ticket #10369](https://bugs.scummvm.org/ticket/10369)


sev-:
Thank you!

PR #2956, opened 2021-04-14
BACKENDS: Fix aspect ratio bug when switching render

lamebanana:
BACKENDS: Fix aspect ratio bug when switching render 

Detailed description of the bug: https://bugs.scummvm.org/ticket/12394


sev-:
I rechecked the logic here, and you caught it properly, thank you!

Merging.

criezy:
This change actually caused a regression. When using hotkeys to toggle the aspect ratio correction this is now ignored. I think the `OpenGLGraphicsManager` was just missing an initialisation of `currentState.aspectRatioCorrection` from `ConfMan.getBool("aspect_ratio")` in its constructor. I will check and if I am correct revert this commit and implement the proper fix.

criezy:
I have now reverted this commit.
And I have checked that it does not reintroduced the bug that it original fixed as since commit ceca0b5ae the window size in OpenGL mode no longer depends on whether the Aspect Ratio correction is enabled or not.

PR #2959, opened 2021-04-14
GUI: Hide "-unknown" candidates if there are known

lolbot-iichan:
This fixes one of issues mentioned at
https://bugs.scummvm.org/ticket/12377

"Chivalry is Not Dead" contains:
* data.dcp mentioned at Wintermute detection tables
* game.exe unrelated to game.exe files mentioned at AGS detection tables

So, detection result is 100% hit for "chivalry" and possible unknown version hits for a dozen of AGS games.

This PR changes "Add" button behaviour to show list of "xxx-unknown" candidates only if there are no known candidates.

### DIFF
--- a/gui/launcher.cpp
+++ b/gui/launcher.cpp
@@ -571,7 +571,10 @@ bool LauncherDialog::doGameDetection(const Common::String &path) {
 		g_system->logMessage(LogMessageType::kInfo, report.encode().c_str());
 	}
 
-	Common::Array<DetectedGame> candidates = detectionResults.listDetectedGames();
+	Common::Array<DetectedGame> candidates = detectionResults.listRecognizedGames();
+	if (candidates.empty()) {
+		candidates = detectionResults.listDetectedGames();
+	}
 
 	int idx;
 	if (candidates.empty()) {


sev-:
As explained on IRC, the proper fix would be to fix AGS engine detection entry.

This has been bugging me for a while. I want to implement a "blacklist" of detection filenames, which would require the engine dev to use at least 2 entries for the game detection.

The possible blacklisted games are:

* game.exe
* demo.exe
* game
* demo
* data.z
* data1.cab
* data.cab

These are way too generic and could present in many titles across many engines.

criezy:
@lolbot-iichan Pull request #1493 is similar there is a discussion there on the possible caveat and reason why it was rejected.

@sev- The issue with AGS games is that sometimes there is only an .exe file (as the game data are embedded in the exe). There is usually also a `acsetup.cfg` file, but it is an editable text file with configuration for the game, and thus not a good candidate for detection. So we may not be able to add a second file to the detection. My idea instead is to use the code we already have in the fallback detector to check that the .exe file is indeed an AGS game and thus avoid having false positive for unknown variants.

lolbot-iichan:
> These are way too generic and could present in many titles across many engines.

@sev- Please consider "engine.exe" as a blacklist candidate as well.

sev-:
@criezy Indeed, checking AGS file validity would be a great thing. Still, what do you think in regards to issuing a warning when  a "blacklisted" file is present alone in an entry?

criezy:
> @criezy Indeed, checking AGS file validity would be a great thing. Still, what do you think in regards to issuing a warning when a "blacklisted" file is present alone in an entry?

Can you clarify who you see as the target audience for this warning? Would it be for users to tell them the unknown variant detection is likely to be a false positive? Or would it be for developers to encourage them to use an additional file for the detection? In the later case I am not sure a warning at runtime when detecting games would be the best approach, but I don't really have anything better to propose.

An alternative could be to drop unknown variant detections that are based on a single file that is blacklisted in the advanced detector, possibly with a warning indicating that possible unknown variant detections were dropped. So if an engine wants to have such detection, only the known ones will be reported, and the engine will need a fallback detector to handle the unknown variants.

sev-:
I like your idea about dropping those. Will do it this way. The warning was intended for developers because I was planning to call `warning()`

sev-:
@criezy Re: on AGS detection.

If `acsetup.cfg` is always present, then that is the perfect candidate for the second file. Add it as `("acsetup.cfg", "abc-md5", 0, -1")`. This way its size and content will be ignored and only presence checked.

criezy:
> @criezy Re: on AGS detection.
> 
> If `acsetup.cfg` is always present, then that is the perfect candidate for the second file. Add it as `("acsetup.cfg", "abc-md5", 0, -1")`. This way its size and content will be ignored and only the presence checked.

Unfortunately that file is not always present. It is not required by the AGS engine, and some games were released without it (for example for _5 Days a Stranger_ only the multi-lingual version has this file).

criezy:
Since this was discussed here, the issue with the AGS detection should now be fixed (it should no longer have false positive when detecting unknown variants).

lolbot-iichan:
Closing this PR since the issue was worked around another way.

PR #2964, opened 2021-04-18
GUI: Allow text selection

srikeerthis:
This PR adds the enhancement requested at
https://bugs.scummvm.org/ticket/6948

The key combination Shift+Left Arrow will highlight and select text towards left of the caret.

Two methods have been added:
- initHighlight() - resets highlight.
- getHighlightOffset() - used for scrolling text animation as text is selected.

Files changed:
- editable.h: declared variables and methods.
- edittext.cpp: Code segment that prevents highlight from being drawn on any position where mouse is clicked.
- editable.cpp:
  - Select and highlight text towards left.
  - Copy selected text using Ctrl-C.
  - Remove selected characters on Backspace/Delete key.
  - Replace selected text with a character.
  - Disable and re-initialize highlight on press of Left Arrow/Right Arrow/Home/End keys

### DIFF
--- a/gui/widgets/editable.cpp
+++ b/gui/widgets/editable.cpp
@@ -20,11 +20,11 @@
  *
  */
 
+#include "gui/widgets/editable.h"
 #include "common/rect.h"
 #include "common/system.h"
-#include "gui/widgets/editable.h"
-#include "gui/gui-manager.h"
 #include "graphics/font.h"
+#include "gui/gui-manager.h"
 
 namespace GUI {
 
@@ -46,6 +46,11 @@ void EditableWidget::init() {
 	_caretInverse = false;
 
 	_editScrollOffset = 0;
+
+	_highlightVisible = false;
+	_highlightPos = 0;
+	_highlightSize = 0;
+	_highlightCharacterCount = 0;
 
 	_align = g_gui.useRTL() ? Graphics::kTextAlignRight : Graphics::kTextAlignLeft;
 	_drawAlign = _align;
@@ -86,7 +91,11 @@ bool EditableWidget::tryInsertChar(byte c, int pos) {
 
 void EditableWidget::handleTickle() {
 	uint32 time = g_system->getMillis();
-	if (_caretTime < time && isEnabled()) {
+	// prevent blinking, if text is highlighted
+	if (_highlightVisible && isEnabled()) {
+		_caretTime = 0;
+		drawCaret(_caretVisible);
+	} else if (_caretTime < time && isEnabled()) {
 		_caretTime = time + kCaretBlinkTime;
 		drawCaret(_caretVisible);
 	}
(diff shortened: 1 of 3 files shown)


gui/widgets/editable.cpp
```diff
@@ -20,11 +20,11 @@
  *
  */
 
+#include "gui/widgets/editable.h"
```
sev-, line 23: These changes must be reverted as they're just reordering of includes

gui/widgets/editable.cpp
```diff
@@ -129,49 +138,86 @@ bool EditableWidget::handleKeyDown(Common::KeyState state) {
 	switch (state.keycode) {
 	case Common::KEYCODE_RETURN:
 	case Common::KEYCODE_KP_ENTER:
+		initHighlight();
```
sev-, line 141: the name of the method is misleading. "init" normally means start. So, better name it `clearHighLight()`

gui/widgets/editable.cpp
```diff
@@ -181,23 +227,31 @@ bool EditableWidget::handleKeyDown(Common::KeyState state) {
...
 		forcecaret = true;
 		dirty = true;
 		break;
 
 	case Common::KEYCODE_UP:
 	case Common::KEYCODE_HOME:
+		initHighlight();
 		// Move caret to start
 		setCaretPos(0);
+		sendCommand(_cmd, 0);
```
sev-, line 244: What is the purpose of this call?

gui/widgets/editable.cpp
```diff
@@ -262,8 +316,25 @@ bool EditableWidget::handleKeyDown(Common::KeyState state) {
 }
 
 void EditableWidget::defaultKeyDownHandler(Common::KeyState &state, bool &dirty, bool &forcecaret, bool &handled) {
-	if (state.ascii < 256 && tryInsertChar((byte)state.ascii, _caretPos)) {
+	//  if highlighted, replace selected text with inserted character
+	if (_highlightVisible) {
+		if (state.ascii < 256 && tryInsertChar((byte)state.ascii, _caretPos)) {
+			for (uint32 i = _highlightPos + _highlightCharacterCount; i > (uint32)_highlightPos; i--) {
+				_editString.deleteChar(i - 1);
```
sev-, line 323: please use `Common::String::erase(uint32 pos, uint32 len)` instead of this loop.

gui/widgets/editable.cpp
```diff
@@ -341,8 +441,8 @@ void EditableWidget::drawCaret(bool erase) {
 		width = MIN(editRect.width() - caretOffset, width);
 		if (width > 0) {
 			g_gui.theme()->drawText(Common::Rect(x, y, x + width, y + editRect.height()), character,
-			                        _state, _drawAlign, _inversion, 0, false, _font,
```
sev-, line 344: This whitespace change should be reverted, or put into a separate commit

gui/widgets/editable.cpp
```diff
@@ -389,4 +503,19 @@ void EditableWidget::makeCaretVisible() {
...
 }
 
+void EditableWidget::initHighlight() {
+	// disable highlight
+	_highlightVisible = false;
+	// Empty the temporary string of previous values
+	_highlightString.clear();
+	// insert blank characters
+	for (int i = 0; i < (int)_caretPos; i++)
+		_highlightString.insertChar(' ', i);
```
sev-, line 513: What is the logic behind this?

sev- (changes requested):
I reviewed the code and put my comments and questions.

A quick test showed the following:

- [ ] The highlight interferes with the tooltip. If I move the mouse over e.g. game id and start highlighting, the highlight will blink when the tooltip is shown
- [ ] On initial launch and starting highlighting right away, it leads to the visual glitch when whole line becomes green, with flickering white spot where cursor is supposed to be, and doubled input line. I'll attach a screenshot
- [ ] When selecting, clicking "shift-right" clears the whole selection and moves cursor right. The logical thing would be to decrease the highlight size by one
- [ ] If the line is wider than the input (scrollPos is non-zero), highlighting on the left edge leads to the whole line being highlighted
- [ ] When in the middle of a line, Shift-Right does not start the highlighting

sev-:
Screenshot 2021-04-19 at 13 36 29

Screenshot of glitched highlight. The input string is "kyra1"

gui/widgets/editable.cpp
srikeerthis: I have tried using append() method to add characters, but had difficulty in reversing the string and displaying the same. So the logic thought was to have an empty temporary string equal to the size of the caret position and add characters from the end of the temporary string.

sev-:
Closing this due to inactivity, lack of handling of the edge cases.

PR #2972, opened 2021-04-20
PINK: fix pink to suit for new macgui, add win95 mode for it.

ysj1173886760:
fix the kWMModeNoDesktopflag
fix the behaviour of appending text when there is an empty string(to suit for new logic of splitString)
add the intersection check of macwindow and spirites of pink
add win95 mode for macgui(macmenu, scrollbar)
make the algorithm where we calc the scrollbar more robust
fix the pink to suit for new macgui


sev-:
Great work, merging.

PR #2973, opened 2021-04-20
ENGINES: Add a template for 3D engine render init (RFC)

lolbot-iichan:
There are currently 4 functions that are very similar and complex, and yet they are copy-pasted and maintained separately:
* Stark::Gfx::Driver::create() that check rendererConfig and OpenGL state & defines to return OpenGLSDriver / OpenGLDriver / CreateTinyGLDriver
* Myst3::createRenderer() that check rendererConfig and OpenGL state & defines to return CreateGfxOpenGLShader / CreateGfxOpenGL / CreateGfxTinyGL
* Grim::GrimEngine::createRenderer that check rendererConfig and OpenGL state & defines to return CreateGfxOpenGLShader / CreateGfxOpenGL / CreateGfxTinyGL
* Wintermute::BaseGame::initialize2 that check rendererConfig and OpenGL state & defines to assign _render to makeOpenGL3DShaderRenderer / makeOpenGL3DRenderer / makeTinyGL3DRenderer

While code was very similar, the implentations had some differencies and bugs:
* Wintermute code was missing a warning when unable to use the desired renderer
* Wintermute code was missing getBestMatchingAvailableRendererType call, leading to bug https://bugs.scummvm.org/ticket/12377, because code didn't handle kRendererTypeDefault
*  Stark code had a check for OpenGLContext.shadersSupported raising error for missing shaders without check for matchingRendererType == Graphics::kRendererTypeOpenGLShaders - it looks like Stark may have failed to work on all the systems without shaders, even if non-shaders OpenGL was manually selected
*  Stark code was showing a warning if unable to use the desired renderer before checking for shaders

This PR moves all the logic to a single template function and switches all 4 engines to it.

What I strongly don't like about this PR is unused "void *" arguments added to some engine functions just to use "T *(*function)(A *)" pattern.

Maybe someone could suggest a more graceful way to do the same,but while keeping "T *(*function)()" signature for some functions?


aquadran:
I don't like template idea. it's putting into one logic with loosing capability custom renderer cases

sev-:
@aquadran In fact, this is done as a utility function. Nothing prevents you from not calling it and implementing your own shenanigans when needed. At this moment, I see only benefit in unifying the code which came to the same state, just slightly got desynced.

aquadran:
@sev- it's not utility, it's just replacement. if it's not flexible, then better not touch it

PR #2975, opened 2021-04-23
COMMON: Fall back to engine icon in taskbar

trembyle:
When an icon for game ID is not found, use the engine icon instead. This provides a fallback so that every game doesn't need an icon.

When discussing this with @SupSuper, I was hoping that this could be extended to the icons used by Discord RPC. Unfortunately we have no way of knowing which icons are stored as assets on the Discord dev portal, and so couldn't check whether a game ID is present there or not.


trembyle:
Here's what it looks like on Windows. Hopefully at least Mac icons look better on Retina displays.

taskbar_director

taskbar_sci

criezy (approved):
Looks good to me.

criezy:
To give an idea, here is what it looks like for me on mac with retina screen.

image

image

criezy:
This is a simple change that seems uncontroversial to me. Merging.

PR #2979, opened 2021-04-25
DRAGONS: More visual fixes for big-endian systems

dwatteau:
This continues the work related to https://bugs.scummvm.org/ticket/11710 and previous PR #2784.

With this, Blazing Dragons visuals on big-endian systems are almost identical to those on little-endian systems. The only remaining visual difference I still see is in the copyright menu, where the left creature has some weird pixels around it (I don't know how to fix that part, but everything else seems to be improved).

(In `font.cpp`, I had to revert the `WRITE_SCREEN` change from my previous PR in order to have proper text rendering.)

Note that `void Screen::drawScaledSprite()` was the only place where an uint16 was used for a read. Changing it to an int16 like the rest doesn't seem to cause a regression.

I haven't tested it back on a little-endian system, yet.

### DIFF
--- a/engines/dragons/font.cpp
+++ b/engines/dragons/font.cpp
@@ -160,7 +160,7 @@ Font *FontManager::loadFont(uint16 index, Common::SeekableReadStream &stream) {
 void updatePalEntry(uint16 *pal, uint16 index, uint16 newValue) {
 	newValue = (uint16)(((uint16)newValue & 0x1f) << 10) | (uint16)(((uint16)newValue & 0x7c00) >> 10) |
 			(newValue & 0x3e0) | (newValue & 0x8000);
-	WRITE_SCREEN(pal + index, newValue);
+	WRITE_LE_INT16(pal + index, newValue);
 }
 
 void FontManager::updatePalette() {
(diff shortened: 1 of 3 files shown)


digitall:
Looked over the code. Looks reasonable and I don't think this will cause any code change on LE systems since the READ_SCREEN symbol is just an alias to the previously used symbol. Will merge as-is.

sev-:
@dwatteau what are you using for debugging these issues? We have long-standing problems with one engine and definitely would benefit from some help.

dwatteau:
> @dwatteau what are you using for debugging these issues? We have long-standing problems with one engine and definitely would benefit from some help.

I have a late PowerBook G4 with Mac OS X Leopard, Debian, and OpenBSD/macppc as a triple-boot. I mostly use Debian for ScummVM (Debian/ppc32 is not a release architecture anymore, but "unstable" still exists for it).

I know that Fullpipe still has some issues on big-endian systems and I participated in bugs [9705](https://bugs.scummvm.org/ticket/9705) and [9713](https://bugs.scummvm.org/ticket/9713) (although I haven't found a solution for the latter, yet).

Is there any other engine where you'd like me to have a look?

Also, I'd happily pay for a late G4 (or G5) system off eBay for an official ScummVM developper, if this can help the project.

sev-:
Yes, that is Fullpipe (NGI engine) which has problems. WOuld be nice if you could take a look at that. Even the demo has the problems if you do not have the full game.

raziel-:
I also offer whatever help I can give...not much, I'm afraid...as I'm suffering from BE issues.

Thank you for your work on this @all

PR #2980, opened 2021-04-25
AGOS: Fix format-overflow warnings

mparnaudeau:
Slightly increase the destination buffer to satisfy the theorical
copy of 5 characters for a 16-bit value.


digitall:
Looks reasonable, though might be better to replace with Common::String usage in future, but fine for now. Merging.

PR #2981, opened 2021-04-26
SCUMM: Fixed regression caused by Pete Workaround

LittleToonCat:
The original commit containing the Pete Wheeler workaround for the Baseball games (#2933) has caused a regression where batting with any other bat as Pete Wheeler would slow down his swinging animation after hitting the ball.  Whoops, This fixes the regression by adding another check to determine whether he is bunting or not.  So sorry about that.

### DIFF
--- a/engines/scumm/script_v6.cpp
+++ b/engines/scumm/script_v6.cpp
@@ -1424,6 +1424,9 @@ void ScummEngine_v6::o6_getAnimateVariable() {
 	 		_currentRoom == ((_game.id == GID_BASEBALL2001) ? 4 : 3) && \
 			vm.slot[_currentScript].number == 2105 && \
 			a->_costume == ((_game.id == GID_BASEBALL2001) ? 107 : 99) && \
+			// Room variable 5 to ensure this workaround executes only once at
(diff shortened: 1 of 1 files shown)


engines/scumm/script_v6.cpp
```diff
@@ -1424,7 +1424,7 @@ void ScummEngine_v6::o6_getAnimateVariable() {
 	 		_currentRoom == ((_game.id == GID_BASEBALL2001) ? 4 : 3) && \
 			vm.slot[_currentScript].number == 2105 && \
 			a->_costume == ((_game.id == GID_BASEBALL2001) ? 107 : 99) && \
-			readVar(0x8000 + 5) != 0)
+			readVar(0x8000 + 5) != 0 && readVar(0x8000 + 22) == 4)
```
bluegr, line 1427: Could you please add a comment here to explain what this is doing?

LittleToonCat: Added in 8c51752

bluegr:
Thanks for your work! Merging

PR #2985, opened 2021-04-28
SCUMM: Fix autosave

orgads:
Autosave was centralized in 30d34fa63d1f5ccf37dfd9c80b0009c4115e78f2, but
SCUMM engine was left behind.

Bugreport #12026

### DIFF
--- a/engines/scumm/scumm.cpp
+++ b/engines/scumm/scumm.cpp
@@ -2315,14 +2315,6 @@ void ScummEngine::scummLoop(int delta) {
 		}
 	}
 
-	// Trigger autosave if necessary.
-	if (!_saveLoadFlag && shouldPerformAutoSave(_lastSaveTime) && canSaveGameStateCurrently()) {
-		_saveLoadSlot = 0;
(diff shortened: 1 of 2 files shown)


engines/scumm/scumm.cpp
```diff
@@ -2315,14 +2315,6 @@ void ScummEngine::scummLoop(int delta) {
 		}
 	}
 
-	// Trigger autosave if necessary.
-	if (!_saveLoadFlag && shouldPerformAutoSave(_lastSaveTime) && canSaveGameStateCurrently()) {
```
sev-, line 2319: I do not see the check for `canSaveGameStateCurrently()` in the global autosave code. Perhaps, it should be added there?

orgads: That's in engine.h:

```
	virtual bool canSaveAutosaveCurrently() {
		return canSaveGameStateCurrently();
	}
```

bluegr:
Thanks for your work! Good to have autosaving in SCUMM in line with the rest of the engines. Merging

PR #2987, opened 2021-04-28
COMMON: removed USE_TTS check from engines

mgerhardy:
This removed the need for the engine authors to add scummvm compile time options checks into their engine code. 
There are however still some places where I've left the check in the engine code - this is e.g. for UI options.

OSystem now just returns a `nullptr` if there is no text to speech manager instance (because none is compiled into the binary, or the system doesn't provide support for it).

### DIFF
--- a/backends/graphics/opengl/opengl-graphics.cpp
+++ b/backends/graphics/opengl/opengl-graphics.cpp
@@ -52,9 +52,7 @@
 #include "image/bmp.h"
 #endif
 
-#ifdef USE_TTS
 #include "common/text-to-speech.h"
-#endif
 
 namespace OpenGL {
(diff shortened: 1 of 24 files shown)


bluegr:
Very nice. All the relevant code segments have null checks for textToSpeechManager, which  is set in the different platforms (macosx.cpp, posix.cpp and win32.cpp).

I've gone through all the changes, and this certainly clears these checks and makes the TTS code much simpler and easier to use. The USE_TTS checks are indeed not needed for the engine code, and should only be left in the backend code that actually initializes the TTS manager.

Many thanks for your work! A straightforward change, which makes things easier for everyone.

Merging

PR #2989, opened 2021-04-29
SDL: Improve multi-monitor support by replacing display 0 with window display

SupSuper:
Currently most SDL2 calls are hardcoded to displayIndex 0 (the primary monitor), so if the ScummVM window is moved to a separate monitor, the results will be incorrect. This attempts to fix this by passing the window display index whenever possible.

This solves the most common bug of resizing / fullscreening on a separate monitor using the wrong resolution, though more work is still necessary to fully support multiple monitors (such as refreshing cached display values when the window is moved).

### DIFF
--- a/backends/platform/sdl/sdl-window.cpp
+++ b/backends/platform/sdl/sdl-window.cpp
@@ -232,9 +232,8 @@ bool SdlWindow::getSDLWMInformation(SDL_SysWMinfo *info) const {
 
 Common::Rect SdlWindow::getDesktopResolution() {
 #if SDL_VERSION_ATLEAST(2, 0, 0)
-	int displayIndex = _window ? SDL_GetWindowDisplayIndex(_window) : 0;
 	SDL_DisplayMode displayMode;
-	if (!SDL_GetDesktopDisplayMode(displayIndex, &displayMode)) {
+	if (!SDL_GetDesktopDisplayMode(getDisplayIndex(), &displayMode)) {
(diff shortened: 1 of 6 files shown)


backends/platform/sdl/sdl-window.cpp
```diff
@@ -265,6 +264,16 @@ SDL_Surface *copySDLSurface(SDL_Surface *src) {
 	return res;
 }
 
+int SdlWindow::getDisplayIndex() const {
+	if (_window) {
+		int displayIndex = SDL_GetWindowDisplayIndex(_window);
```
bluegr, line 269: This is an SDL2 call. The equivalent SDL1 call was SDL_GetCurrentVideoDisplay(). There should be a version check here, so that this compiles with SDL1

bluegr:
Great work overall! Clean and straightforward implementation. Once the SDL2-specific code is handled correctly for SDL1, this is good to be merged

backends/platform/sdl/sdl-window.cpp
SupSuper: The version check is at line 245. Since this function is only used in SDL2 there's no need for a SDL1 version.

bluegr: Indeed. That SDL2 block is quite large, and includes other functions too.

bluegr:
Thanks for your work! I've tried this and it improves behavior with multiple monitors

PR #2990, opened 2021-04-29
SCUMM: Fix format-truncation warning in IMUSE player

mparnaudeau:
The size in snprintf includes the final null character, so here
3 bytes are not enough to store a space, 2 digits and the null
character.
The compiler complained and the output used to display 1 digit
instead of 2, for each byte.
Extend this size from 3 to 4 in the snprintf call.


digitall:
@mparnaudeau : Looks good and since it only affects the buf[] used for debugC output, I doubt this is going to cause any functional issue. Thanks, merging as-is.

PR #2991, opened 2021-04-30
SCUMM: fix bug 4591 Zak keeps walk animation without moving

lkramer:
This fixes a bug described here: https://bugs.scummvm.org/ticket/4594

It seems that if the script tells the player actor to turn towards
something while the player actor is walking, the walk animation is not
stopped.

The fix is to play the standing animation after turning
direction.

The bug is easy to reproduce following the steps in the bug report, and is also reproducible in the PC version (Zak McKracken and the Alien Mindbenders (V2/DOS/English)). I played through a good deal of scripted events in Zak and could see no adverse behaviour. To be on the safe side, we could put a guard around the code change so it's only executed if the game is Zak.

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -2507,6 +2507,7 @@ void Actor::animateActor(int anim) {
 		break;
 	case 4:				// turn to new direction
 		turnToDirection(dir);
+		startAnimActor(_standFrame);
 		break;
 	case 64:
 		if (_vm->_game.version == 0) {


athrxx:
Thank you for working on this.
However, I currently don't see that the original code would do something like this there.
So, where does this call to startAnimActor(_standFrame) come from?

The original code seems to do exactly 2 things:
1) set _targetFacing to the new direction
2) set _actor_moving to MF_TURN

But that might just be me not seeing it...
The fact that our code is based on SCUMM5 and support for the other SCUMM versions has been factored into that doesn't make figuring out things like this any easier (even this newDirToOldDir/fromSimpleDir/toSimpleDir/normalizeAngle stuff drives me nuts every time).

lkramer:
Hi, you're right. I did not do my research well enough. I'll withdraw the PR. Thanks for looking.

PR #2992, opened 2021-04-30
AGS: Remove duplicate achievements entries

lolbot-iichan:
2 games were described twice at achievements_tables.h: "atotkdeluxe" and "theterribleoldman"


sev-:
Thanks!

PR #2994, opened 2021-05-01
MorphOS: cleanup code

BeWorld2018:
- remove utf8toLocal and use internal encode
- save only path (drawer) in browser_lastpath


bluegr:
Thanks for your work. Merging

PR #2995, opened 2021-05-01
TINSEL: numPolygons fixes for big-endian systems

dwatteau:
When testing the original V0 Floppy Demo of Disworld (`dw-dos-floppy-demo-en.zip`) on a big-endian PowerPC machine, the game would immediately trigger `assert(numPolys <= MAX_POLY)` in `MaxPolygons()`.

It appears that numPolygons was left uninitialized, and so it was set to a big value which would trigger this assertion.

Looking at the changes done in commit 24b77f1, I think the original intent was to use the `(cptr != NULL) ? READ_32(cptr) : 0` idiom, like the rest around it.

Tested with the various available Discworld1 demos on a big-endian system.


bluegr:
Thanks for this, MaxPolygons was indeed not initialized in this case. Merging

PR #2997, opened 2021-05-02
DETECTOR: Modify getFileProperties to cache md5 results

a-yyg:
Define a Singleton that stores the md5 and size of files, used during mass add.
This skips the double computing of md5s during a single mass add, with the cache being cleared in subsequent mass adds.

### DIFF
--- a/engines/advancedDetector.cpp
+++ b/engines/advancedDetector.cpp
@@ -450,9 +450,31 @@ void AdvancedMetaEngineDetection::composeFileHashMap(FileMap &allFiles, const Co
 	}
 }
 
+/* Singleton Cache Storage for MD5 */
+
+namespace Common {
+	DECLARE_SINGLETON(MD5CacheManager);
+}
+
 bool AdvancedMetaEngineDetection::getFileProperties(const FileMap &allFiles, const ADGameDescription &game, const Common::String fname, FileProperties &fileProps) const {
 	// FIXME/TODO: We don't handle the case that a file is listed as a regular
 	// file and as one with resource fork.
+	if (!allFiles.contains(fname))
+		return false;
+
+	fileProps.md5.clear();
+
+	const char *numbers = "0123456789";
+	Common::String hashname = allFiles[fname].getPath() + ":";
+	for (uint n = _md5Bytes; n > 0; n/=10) {
+		hashname += numbers[n % 10];
+	}
+
+	if (MD5Man.contains(hashname)) {
+		fileProps.md5 = MD5Man.getMD5(hashname);
+		fileProps.size = MD5Man.getSize(hashname);
+		return true;
+	}
 
 	if (game.flags & ADGF_MACRESFORK) {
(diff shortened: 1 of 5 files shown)


engines/advancedDetector.cpp
```diff
@@ -450,9 +450,31 @@ void AdvancedMetaEngineDetection::composeFileHashMap(FileMap &allFiles, const Co
...
 bool AdvancedMetaEngineDetection::getFileProperties(const FileMap &allFiles, const ADGameDescription &game, const Common::String fname, FileProperties &fileProps) const {
 	// FIXME/TODO: We don't handle the case that a file is listed as a regular
 	// file and as one with resource fork.
+	if (!allFiles.contains(fname))
+		return false;
+
+	fileProps.md5.clear();
+
+	const char *numbers = "0123456789";
+	Common::String hashname = allFiles[fname].getPath() + ":";
```
sev-, line 468:
```C++
Common::String hashname = Common::String::format("%s:%d", allFiles[fname].getPath().c_str(), _md5Bytes);
```

gui/launcher.cpp
```diff
@@ -393,6 +394,7 @@ void LauncherDialog::massAddGame() {
 	MessageDialog alert(_("Do you really want to run the mass game detector? "
 						  "This could potentially add a huge number of games."), _("Yes"), _("No"));
 	if (alert.runModal() == GUI::kMessageOK && _browser->runModal() > 0) {
+		MD5Man.clear();
```
sev-, line 397: You need to run it not only during massAdd, but during the conventional detection too, since it always goes through all the engines.

I would say, the more proper place for this would be somewhere around `DetectionResults EngineManager::detectGames(const Common::FSList &fslist)`

sev- (changes requested):
I added a comment about optimization.

Also, I believe, you need to clear the MD5s before every detection, not just when you're running massAdd.

gui/launcher.cpp
a-yyg: Thank you for the suggestions. I've force pushed the suggestion regarding `Common::String::format`.
As for the second suggestion,
I had put `MD5Man.clear()` inside of `DetectionResults EngineManager::detectGames(const Common::FSList &fslist)` before, but noticed that the amount of cache hits decreased when the engine had `directoryGlobs` defined. This is most likely because the file (located in a subfolder defined in directoryGlobs) is added to `allFiles` and has their md5 computed, but the cache is still cleared between folder jumps inside `void MassAddDialog::handleTickle()`.
To solve this I thought using the entire path as the hash key would be a good choice since there is (supposedly) no chance of the wrong md5 cache being used that way.
I could be understanding the system incorrectly and this solution could be more detrimental than I thought.

a-yyg: Force pushed the second suggestion since not clearing the cache may have unwanted side effects. Thank you for your comments.

sev-:
Thank you!

PR #2998, opened 2021-05-04
BUILD: Deduplicate install targets

orgads:


criezy (approved):
That looks good to me.

The only comment is that it feels a bit strange to create the `$(DESTDIR)$(bindir)` and `$(DESTDIR)$(libdir)/scummvm/` directories in the `install-data` target. I understand it avoids duplicating those two lines in the `install` and `install-strip` targets, but the `install-data` target does not copy any files to those directories, and only the `install` and `install-strip` do. So to me it seems it would maybe be a bit more logical to create the directories in the `install` and `install-strip` targets.

orgads:
Done

PR #2999, opened 2021-05-04
MAKEFILE: Add win32dist-mingw target

orgads:
Installs, and bundles all required dlls in bin directory.


ports.mk
```diff
@@ -70,6 +70,9 @@ ifdef DYNAMIC_MODULES
 	rm -rf "$(DESTDIR)$(libdir)/scummvm/"
 endif
 
+bundle-mingw: install-strip
+	@cp -vu $(shell ldd "$(DESTDIR)$(bindir)"/$(EXECUTABLE) | grep -i mingw | cut -d">" -f2 | cut -d" " -f2 | sort -u) "$(DESTDIR)$(bindir)"
```
lephilousophe, line 74: I am not sure to see how ldd works here as it's a Windows executable. Does it work when cross compiling?

orgads: I didn't test. I'll try to test it tomorrow.

orgads: No, it doesn't work on linux. It is meant for MSYS2 environment.

Finding the dependencies on linux is complicated. Can be done [using objdump](https://stackoverflow.com/a/58671570/764870), but this only gives a raw list. You need to locate each dll from the list on the path.

But anyway, the cross compilation is usually statically linked, isn't it? So this target is not needed at all for this kind of build.

aquadran: I'm not sure if this should be part of scummvm at all

orgads: Why not? It's just an extra target, doesn't affect anyone who doesn't use it directly.

See the [MinGW instructions](https://wiki.scummvm.org/index.php/Compiling_ScummVM/MinGW-w64). They're rather scary.

I'm trying to simplify them by [providing the needed curl package](https://github.com/msys2/MINGW-packages/pull/8469), and creating a simple target for bundling the required dlls (`make bundle-mingw`).

After both are done, the instructions can be reduced to more or less 5 lines.

aquadran: there is nothing scary there. it's just moving step from wiki to makefile.
it's adding mess. at least should be pointed as only MinGW windows only to prevent confusion

bluegr: I agree with @orgads here. We have other targets with platform-specific commands (e.g. ScummVMDockTilePlugin further down uses the macOS ``lipo`` command, iphonebundle uses the ``ldid`` homebrew tool for iOS, osxsnap uses the macOS ``hdiutil`` tool. I don't see why this is different, it's a Windows-only target

aquadran:
We have a lot of mess in Makefile, sure we can add one more. just point it in name description it's windows MinGW only target

orgads:
Renamed bundle-win-mingw, and replaced `$(shell)` with `xargs`.

lephilousophe:
I don't think you should use install-strip as a prerequisite, it implies install-data which will copy many files.
These are not needed because, IIRC, they are already bundled in the executable resources.
@SupSuper or @sluicebox know better than me.

lotharsm:
Indeed, adding all those resource files as well is not necessary, since they are already merged into the binary itself. On Win32, you only need the EXE and the DLL files, but no other files installed on other platforms at all.

lephilousophe:
@lotharsm you still need the docs, licenses, readme,... for the user. :)

lotharsm:
Yes, indeed :-) I just wanted to clarify that the stuff pulled in via the `engine-data` won't be necessary here,

bluegr:
Why are we talking about data files @lephilousophe, @lotharsm? datadir is not used at all in this target, only bindir is, so only the binary files will be copied

orgads:
How about this?

lephilousophe:
OK for me but just note 2 things you may want to improve to fit your needs.
The `win32dist` target installs to `c:\scummvm` and it's not configurable (hard-coded in `configure`).
The target also installs an ISS script for InnoSetup.
I don't know the users of the `win32dist` target, maybe @lotharsm knows.

lotharsm:
> I don't know the users of the `win32dist` target, maybe @lotharsm knows.

Nope, since I don't use the `win32dist` target, since I use a custom build script and a different Inno Setup file.

orgads:
I see that it was initially `c:/scummvm` (3cf5c78c42a287a73ddd6ddb82f9c455f83c5309), then it was changed to `build/x86` (371b9cace7e911a762747a083128b73853632268), and finally changed back to `c:/scummvm` (50f28f2c23d72799bf4e415783650320a432ed97). Last change was 10 years ago :)

I doubt anybody uses it nowadays.

Can I remove the innosetup part from there?

lephilousophe:
Errrm... About the path, I was looking at all of this recently because Amiga also has a hard-coded variable in configure script.
In addition, buildbot produced packages have too many files bundled in it.

There are 2 paths involved here for Windows: STAGINGPATH, relative and not used anymore and WIN32PATH, absolute and expected to be overriden when running the target.
Some cleanup will be needed...

If you agree with that we could modify this to come back to what was 10 years ago with a win32-data target, and your win32dist-mingw both of them using the same variable to output files (maybe DESTDIR as it's already used on Linux ?).
We could add some other target for use by Buildbot which would do the same

What do you think?

orgads:
Is this what you meant?

orgads:
`DESTDIR` is `/usr/local` by default. I don't think this is the appropriate place for these files. What do you say?

lephilousophe:
I overlooked this. I thought DESTDIR was empty by default.
Let see what others think about your change. I'm OK with it.

PR #3002, opened 2021-05-07
SCUMM: Add Macintosh variant of Monkey Island 1

eriktorbjorn:
I noticed yesterday that ScummVM was warning me about missing CD audio tracks for the Macintosh version of The Secret of Monkey Island. This is because while the game is correctly detected as the Mac version, it is labelled as a "CD" variant. This pull request adds a "Macintosh" variant, with the same properties as the CD variant, except without audio tracks. I hope this is the correct approach.

I recently moved and haven't had the time to properly unpack my game CDs, but while the copies I have were distributed on CD there were apparently no CD audio tracks on them. At least not as I recall it, and not according to MobyGames either:

- https://www.mobygames.com/game/lucasarts-mac-cd-game-pack
- https://www.mobygames.com/game/macintosh/lucasarts-mac-cd-game-pack-ii

"Like its predecessor, the games were simply carried over from floppy to CD-ROM with no enhancements. It was bundled with CD-ROM drives, which was still a novelty then."

While both of these contain The Secret of Monkey Island, they are actually two different versions. (One of them has a copy protection screen, but ScummVM doesn't show it by default. I don't know if the code wheel was included originally.) ScummVM can apparently detect a _third_ version, but I don't own that one. Still, I guess the fix should apply to all three, so that' what I've done here.


criezy:
I believe `scumm-md5.h` is generated from the data in `devtools/scumm-md5.txt`. So rather than modify it directly you would need to change the txt file and run the `md5table` tool. Otherwise your changes will be lost the next time we run that tool.

eriktorbjorn:
Ah, I wasn't aware of that. I've force-pushed a new version of the patch.

bluegr:
Nice work, thanks! Merging

PR #3003, opened 2021-05-07
WIN32: Enable building with UNICODE

sluicebox:
This PR adds the ability to configure Windows builds with UNICODE (and _UNICODE) defined. It keeps existing behavior when they are not defined. It does not change default configurations. The idea is that this is safe to merge, `--enable-windows-unicode` can be tested for a while, and then eventually become the default.

Windows builds currently can't handle paths with most unicode characters because we always use the Windows ANSI API. On NT-based Windowses, that means strings are converted to and from the native wide-character API using CP_ACP. This works on some characters but it's a lossy conversion that depends on machine settings. A path that gets returned by Windows might not work when ScummVM passes it back to open a file or list a directory. Games can't be loaded if a path contains these characters and user profile paths break if the username contains one of these. Most long-lived Windows codebases have faced these problems. The way out is to use the native wide-character API by defining UNICODE and do conversions to and from the application with UTF8.

Our current Windows code is a bit incoherent when it comes to UNICODE. There are #if(n)def UNICODE sections that imply some kind of support... except that there really isn't, and the code in these sections wouldn't have compiled if there was. The rest of the code is inconsistent and doesn't build when UNICODE is defined, which isn't surprising since it's never been used.

This PR sorts all that out. We can now build with UNICODE so we can see what it fixes and what still needs work. If all goes well we can eventually enable it by default and then all paths will work. Disabling UNICODE retains the existing behavior and DLL imports, preserving Win9x support. So far it's working well for me. I can browse to paths I couldn't before in the ScummVM UI, and I can add and run games from paths I couldn't before. Unicode paths can now be safely passed in command line arguments. All paths I've tested have persisted correctly in scummvm.ini and displayed properly in the UI.

The strategy:
* Fix all build errors when UNICODE is defined.
* Keep the exact same behavior as now when UNICODE isn't defined.
* Use wide-char APIs with UTF8 conversions when UNICODE is defined.

To enable, pass  `--enable-windows-unicode` to `configure` or `--use-windows-unicode` to `create_project`. `--disable-windows-unicode` / `--use-windows-ansi` can be used should this become a default.

Tested:
* Building with a very old mingw with unicode enabled and disabled
* Building with a very new mingw-w64 with unicode enabled and disabled
* Building with vs2015/vs2019 with unicode enabled and disabled

Scummvm.ini files exist in the wild with path strings created by CP_ACP conversion, so that may need some compatibility handling. If there's anything to be done there, I think it will be easier to figure out and test after this is merged.

https://bugs.scummvm.org/ticket/12409 is an example of one of the many symptoms of CP_ACP.

### DIFF
--- a/backends/audiocd/win32/win32-audiocd.cpp
+++ b/backends/audiocd/win32/win32-audiocd.cpp
@@ -52,6 +52,7 @@
 #include "audio/audiostream.h"
 #include "backends/audiocd/audiocd-stream.h"
 #include "backends/audiocd/default/default-audiocd.h"
+#include "backends/platform/sdl/win32/win32_wrapper.h"
 #include "common/array.h"
 #include "common/config-manager.h"
 #include "common/debug.h"
@@ -197,7 +198,9 @@ bool Win32AudioCDManager::openCD(int drive) {
 
 	// Construct the drive path and try to open it
 	Common::String drivePath = Common::String::format("\\\\.\\%c:", drives[drive]);
-	_driveHandle = CreateFileA(drivePath.c_str(), GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE, 0, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, 0);
+	TCHAR *tDrivePath = Win32::stringToTchar(drivePath);
+	_driveHandle = CreateFile(tDrivePath, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE, 0, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, 0);
+	free(tDrivePath);
 	if (_driveHandle == INVALID_HANDLE_VALUE) {
 		warning("Failed to open drive %c:\\, error %d", drives[drive], (int)GetLastError());
 		return false;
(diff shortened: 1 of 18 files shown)


ZvikaZ:
I have opened https://bugs.scummvm.org/ticket/12409 , and now I tested this PR, and indeed, it solves that use case.
@sluicebox , thanks for addressing this!

lotharsm:
Any objections against merging this?

SupSuper (approved):
Looks good to me, thanks for taking this on and making the dreaded Win32 code more consistent. :)

sluicebox:
Thanks everybody for taking a look and thank you @ZvikaZ for testing. I've been running this branch quite a bit in the past two months and feel good about it. @SupSuper had great suggestions on Discord that made this better. (In between, you know, adding an *entire engine!*)

I've recently come up with an approach for cleanly handling scummvm.ini files that are out there with working CP_ACP paths. I'll work on that soon. As far as I know that is the only remaining issue to take care of before this can be enabled for releases. (Otherwise we risk breaking working paths in scummvm.ini upon upgrade from an ANSI build to a UNICODE build, which is kinda the opposite of what we're trying to do here!)

Merging this now, but just a reminder: this won't automatically change behavior yet, for now you have to enable this in configure / create_project.

PR #3004, opened 2021-05-07
BACKENDS: SDL: Fix stack buffer overflow on empty supported formats list

mparnaudeau:
Still running scummvm compiled with address sanitizer (by default here), running the GRIM engine (GRIM or EMI games) with the default 3D renderer (OpenGL), there is a stack buffer overflow when starting the game:

```
==24643==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffff182b530 at pc 0x55bdad2014b9 bp 0x7ffff182b350 sp 0x7ffff182b348
READ of size 9 at 0x7ffff182b530 thread T0
    #0 0x55bdad2014b8 in SdlGraphicsManager::setState(SdlGraphicsManager::State const&) backends/graphics/sdl/sdl-graphics.cpp:89
    #1 0x55bdac938bf3 in OSystem_SDL::setGraphicsMode(int, unsigned int) backends/platform/sdl/sdl.cpp:825
    #2 0x55bdacf2ae1d in initGraphics3d(int, int) engines/engine.cpp:398
    #3 0x55bdacd1cbc3 in Grim::GrimEngine::createRenderer(int, int) engines/grim/grim.cpp:273
    #4 0x55bdacd43617 in Grim::GrimEngine::run() engines/grim/grim.cpp:389
    #5 0x55bdac94e61c in runGame base/main.cpp:300
    #6 0x55bdac94e61c in scummvm_main base/main.cpp:592
    #7 0x55bdac94221e in main backends/platform/sdl/posix/posix-main.cpp:45
    #8 0x7f12f940509a in __libc_start_main ../csu/libc-start.c:308
    #9 0x55bdac898369 in _start (/home/mathias/Sources/SCUMMVM/scummvm-mpa-clang/scummvm+0x59e369)

Address 0x7ffff182b530 is located in stack of thread T0 at offset 432 in frame
    #0 0x55bdad200bb7 in SdlGraphicsManager::setState(SdlGraphicsManager::State const&) backends/graphics/sdl/sdl-graphics.cpp:79

  This frame has 7 object(s):
    [32, 40) '<unknown>'
    [96, 104) '<unknown>'
    [160, 168) '<unknown>'
    [224, 232) '<unknown>'
    [288, 296) '<unknown>'
    [352, 361) 'format'
    [416, 432) 'supportedFormats' <== Memory access at offset 432 overflows this variable
HINT: this may be a false positive if your program uses some custom stack unwind mechanism or swapcontext
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow backends/graphics/sdl/sdl-graphics.cpp:89 in SdlGraphicsManager::setState(SdlGraphicsManager::State const&)
```

    With GRIM engine and OpenGL renderer, the called getSupportedFormats() is
    from openglsdl-graphics3d.h, that returns an empty list. So, data field
    is read from the front node, that has no real value, what cause an access
    beyond the stack frame.

I just added a check to ignore and keep the default format if the list is empty.

### DIFF
--- a/backends/graphics/sdl/sdl-graphics.cpp
+++ b/backends/graphics/sdl/sdl-graphics.cpp
@@ -85,8 +85,10 @@ bool SdlGraphicsManager::setState(const State &state) {
 		// pixel format instead.
 		Graphics::PixelFormat format = state.pixelFormat;
 		Common::List<Graphics::PixelFormat> supportedFormats = getSupportedFormats();
-		if (Common::find(supportedFormats.begin(), supportedFormats.end(), format) == supportedFormats.end())
-			format = supportedFormats.front();
+		if (!supportedFormats.empty()) {
+			if (Common::find(supportedFormats.begin(), supportedFormats.end(), format) == supportedFormats.end())
+				format = supportedFormats.front();
+		}
 		initSize(state.screenWidth, state.screenHeight, &format);
 #else
 		initSize(state.screenWidth, state.screenHeight, nullptr);


lephilousophe:
The supported format list should not be empty. This fix will just hide another bug.
I think it should raise an error instead.

criezy:
It seems to me that the issue is indeed in the `OpenGLSdlGraphics3dManager` implementation of `getSupportedFormats`.
I am not familiar with the 3D stuff, but I guess it should maybe include the `_overlayFormat` to the list of supported formats it returns. It completely ignores the pixel format it gets in its implementation of `initSize`, so I guess in the end it does not really matter what it returns in `getSupportedFormats`.

And then we can indeed have an `assert()` or `error()` in `SdlGraphicsManager::setState` if the list is empty.

criezy:
I forgot this pull request.
I will close it now has the issue has been fixed in the meantime.

PR #3006, opened 2021-05-10
COMMON: Add support for multi-file InstallShield cabinets

fracturehill:
This PR adds support for versions 6-13 InstallShield cabinets,
which can be both single- and multi-volume (v5 multi-volume
cabinets are still not supported since I have no test cases). This is
used in some of the Nancy Drew games. The code was written
several years ago by waltervn and clone2727, but never made its
way into the main tree.

This makes some minor changes to the AGOS and Tony engines
which I am unable to test since I don't own any of the games;
I'd appreciate if someone who has The Feeble Files or one of
the Simon the Sorcerer's Puzzle Packs could test the
unextracted versions and see if there are any issues.

### DIFF
--- a/common/installshield_cab.cpp
+++ b/common/installshield_cab.cpp
@@ -123,6 +123,9 @@ bool InstallShieldCabinet::open(const String &baseName) {
 	case 0x0100600c:
 		_version = 6;
 		break;
+	case 0x020004B0: // Found in some Russian variants of Nancy Drew games. Possibly a malformed header
+		_version = 6;
+		break;
 	default:
 		warning("Unsupported CAB version %08x", version);
 		close();
--- a/common/installshield_cab.cpp
+++ b/common/installshield_cab.cpp
@@ -48,20 +48,26 @@
 #include "common/hash-str.h"
 #include "common/installshield_cab.h"
 #include "common/memstream.h"
+#include "common/substream.h"
+#include "common/ptr.h"
 #include "common/zlib.h"
 
 namespace Common {
 
+namespace {
+
 class InstallShieldCabinet : public Archive {
 public:
-	InstallShieldCabinet(SeekableReadStream *stream, DisposeAfterUse::Flag disposeAfterUse);
-	~InstallShieldCabinet();
+	InstallShieldCabinet();
+
+	bool open(const String &baseName);
+	void close();
 
 	// Archive API implementation
-	bool hasFile(const String &name) const;
-	int listMembers(ArchiveMemberList &list) const;
-	const ArchiveMemberPtr getMember(const String &name) const;
-	SeekableReadStream *createReadStreamForMember(const String &name) const;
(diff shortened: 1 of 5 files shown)


sluicebox:
This is great, it makes it possible to read the resource files from the Polish version of the Leisure Suit Larry 7 CD. That was reported four years ago, and with this I was able to get it wired up and working in a few hours. https://bugs.scummvm.org/ticket/10066

In that game's case, the Polish patch files are packaged in DATA1.CAB and DATA1.HDR.

bluegr:
This is great! Sad that this was not merged up to now.
+1 from me

common/installshield_cab.cpp
```diff
@@ -48,114 +48,177 @@
...
+	uint32 version = header->readUint32LE();
+
+	switch(version) {
+	case 0x01000004:
+		_version = 5;
+		break;
+	case 0x0100600c:
+		_version = 6;
+		break;
+	case 0x020004B0: // Found in some Russian variants of Nancy Drew games. Possibly a malformed header
```
bluegr, line 126: This is actually InstallShield version 12, not a malformed header. Check https://github.com/twogood/unshield/blob/master/lib/libunshield.c and https://www.onlinegdb.com/mXzjHjXAt for a small program that extracts the version, based on that repo (note that magic bytes 0x01000004 are actually transformed into version 0 using that algorithm).

Please change this

common/installshield_cab.cpp
```diff
@@ -48,114 +48,177 @@
...
+		return false;
 	}
 
-	uint32 version = _stream->readUint32LE();
-
-	if (version != 0x01000004) {
+	uint32 version = header->readUint32LE();
+
+	switch(version) {
+	case 0x01000004:
```
bluegr, line 120: Based on the previous comment, I would suggest to change this switch into something like this:

    int shift = magicBytes[i] >> 24;
    _version = shift == 1 ? (version >> 12) & 0xf : (version & 0xffff) / 100;
    _version = (_version == 0) ? 5 : _version;

The last check for version 0 has been commented out in unshield, but we are treating this version as 5 in our code, so it's needed

fracturehill: Done. Based on the unshield code we should now (mostly) support versions 5-13, though I'd need some more test cases to verify if that's actually true.

sluicebox: This correctly handles the LSL7 Polish files. (Data1.cab/hdr) They are version 5 with magicBytes 0x01005201. I confirmed in the installer's version strings that it really is version 5.

bluegr:
Thanks for the changes! :)

Overall, this is a straightforward implementation, and we know it works for two games already. The changes to the agos and tony engines are minimal, so overall this should be OK to merge.

Thanks to all the involved persons for the work done on this feature. Merging

PR #3008, opened 2021-05-12
GRAPHICS: Fix leak in ManagedSurface, not forcing _disposeAfterUse to NO

mparnaudeau:
With the commit just merged about leak from PicButtonWidget, the ManagedSurface object is freed but not its pixel buffer because _disposeAfterUse was overwritten with value NO after the allocation with copyFrom().

That causes this indirect leak of thumbnail buffers (160x100x2 bytes):

```
Indirect leak of 96000 byte(s) in 3 object(s) allocated from:
    #0 0x7f21b8b93518 in calloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9518)
    #1 0x55cedbde0908 in Graphics::Surface::create(unsigned short, unsigned short, Graphics::PixelFormat const&) graphics/surface.cpp:75
    #2 0x55cedbde14dc in Graphics::Surface::copyFrom(Graphics::Surface const&) graphics/surface.cpp:96
    #3 0x55cedbd82895 in Graphics::ManagedSurface::copyFrom(Graphics::Surface const&) graphics/managed_surface.cpp:218
    #4 0x55cedbd7b331 in Graphics::ManagedSurface::ManagedSurface(Graphics::Surface const*) graphics/managed_surface.cpp:105
    #5 0x55cedb209847 in GUI::PicButtonWidget::setGfx(Graphics::Surface const*, int, bool) gui/widget.cpp:613
    #6 0x55cedb0e9241 in GUI::SaveLoadChooserGrid::updateSaves() gui/saveload-dialog.cpp:1110
    #7 0x55cedb0def72 in GUI::SaveLoadChooserGrid::open() gui/saveload-dialog.cpp:913
    #8 0x55cedaf2b498 in GUI::Dialog::runModal() gui/dialog.cpp:74
```

In the case of the ManagedSurface constructor from a Surface object,
_disposeAfterUse is not initialized. But more, copyFrom() sets
_disposeAfterUse to YES after allocation of buffer. And then,
after the call to copyFrom, _disposeAfterUse was set to NO, what
made the pixels buffer not freed.

### DIFF
--- a/graphics/managed_surface.cpp
+++ b/graphics/managed_surface.cpp
@@ -101,8 +101,8 @@ ManagedSurface::ManagedSurface(const Surface *surf) :
 		return;
 	}
 
-	copyFrom(*surf);
 	_disposeAfterUse = DisposeAfterUse::NO;
+	copyFrom(*surf);
 }
 
 ManagedSurface::~ManagedSurface() {
--- a/graphics/managed_surface.h
+++ b/graphics/managed_surface.h
@@ -136,7 +136,7 @@ public:
 	/**
 	 * Create a managed surface from plain Surface.
 	 *
-	 * If disiposeAgter use flag is set (default), the surface will reuse all structures
+	 * If disposeAfterUse flag is set (default), the surface will reuse all structures
 	 * from the surface and destroy it, otherwise it will make a copy.
 	 */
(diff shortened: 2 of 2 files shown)


bluegr:
Nice! Thanks for your work. Merging

PR #3012, opened 2021-05-17
BASE: add getDebugChannels in metaEngineDetection

ysj1173886760:
add new method getDebugChannels() in metaEngineDetection, which will get engine debug flags even engine is not instantiated.
move adding debugChannels from the constructor of engine to detection.cpp
add list-debugflags and list-all-debugflags in commandLine
original feature request https://bugs.scummvm.org/ticket/7535


sev-:
Thanks! Now we need to update Quux engine (see quux branch)

PR #3014, opened 2021-05-17
SCUMM: Workaround missing subtitle in The Dig

orgads:
Fixes #4410

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -2774,8 +2774,11 @@ void ScummEngine_v7::actorTalk(const byte *msg) {
 	playSpeech((byte *)_lastStringTag);
 
 	if (_game.id == GID_DIG || _game.id == GID_CMI) {
-		if (VAR(VAR_HAVE_MSG))
+		if (VAR(VAR_HAVE_MSG)) {
+			if (_game.id == GID_DIG && _roomResource == 58 && msg[0] == ' ' && !msg[1])
+				return;
 			stopTalk();
+		}
 	} else {
 		if (!_keepText)
 			stopTalk();


athrxx:
Is this a bug that also happens with the original interpreter?
If so, a workaround like that would be fine IMO.
However, if this occurs in ScummVM only, it would be preferable to fix the underlying issue.

orgads:
I don't know, and not sure how to check.

digitall:
The bug referenced here is https://bugs.scummvm.org/ticket/4410 ... This is a missing subtitle in the Tomb which is a fair way into the game IIRC. The room number is 58, but The Dig's original interpreter lacks a debug mode so no way to "warp" there:
https://wiki.scummvm.org/index.php/SCUMM/Debug_Codes#The_Dig

The bug has a savegame for ScummVM, but not for the original interpreter. @orgads : You would need to set up the original interpreter in Dosbox or Windows VM and play through to this point to verify. I can look at doing this and attach some original engine savegames to the bug... but it might take me a while.

digitall:
If you need a walkthrough, https://steamcommunity.com/sharedfiles/filedetails/?id=504761213 is reasonable.

orgads:
Please do it and send me an original savegame. Thanks.

orgads:
Hi. @antoniou79 found gameplay on youtube, and it looks like the problem occurs with the original interpreter. So can we merge this fix?

digitall:
@orgads : The issue is usually working out if the gameplay videos are using ScummVM or the original interpreter ... and which exact game version. Have got DIG working in DOSBox pretty easily... Just going to play through now and will attach tomb savegame to the bug.

orgads:
The French version referenced in the bug report is executed on DOSBox, you can see it clearly in the video's beginning. And the issue happens there too.

orgads:
https://youtu.be/LYYjila4-E8?t=6468

orgads:
@antoniou79 was quicker ;)

Reproduces with DOSBox.

digitall:
Ah cool... Well, happy to merge here. I think even if this is a "workaround", the SCUMM engine developers can look at this in future and better than broken :)

PR #3016, opened 2021-05-18
TRECISION: New engine for Nightlong: Union City Conspiracy

SupSuper:
This engine currently supports all PC versions of [Nightlong: Union City Conspiracy](https://wiki.scummvm.org/index.php?title=Nightlong:_Union_City_Conspiracy). Other adventure games made by Trecision may be added in the future.
If you don't own the full game, you can test using [the demo](https://www.gamepressure.com/download.asp?ID=985).

Thanks to the original devs for providing us source code, and @Strangerke @bluegr @madmoose for all their help. :)


raziel-:
OMG...yes, yes, yes :-)
Another of my all-time favourites to be supported.

Thank you @SupSuper 

Question: I own the Amiga version from PXL computers/ clickBOOM and could provide md5sums for all data files, if needed?
The data directories seem to be the same as with the PC version and as far as I know did they only provide a different binary.

bluegr:
@raziel-: The Amiga version is quite rare, glad you have it! :)
Unfortunately, it's not supported at the moment, as it uses a different encoding for the game's videos and animation (not Smacker)

raziel-:
@bluegr 
Yes, so glad I kept hold of it after it stopped working on the NG amigas :-)

Bummer...I really hope they didn't degrade the quality to save space?
It still comes on three CDs, did the PC version aswell?

Spam me with testing (I want to help get the amiga version supported) :-)

Strangerke:
@raziel- The Amiga version is documented to be of a lower quality than the PC version, and we don't know much about it so it may, or may not, require reverse engineering. So far, I had difficulties to extract the FMV video files from it. I expect it to be a Bink format, but what I retrieve is plain crap (I see artefacts coming from bitmaps instead). When you have some time, I'd love to discuss about that :) Thanks for the help offer!

raziel-:
@Strangerke 

Shoot...I'm here all night :-)

digitall:
@bluegr: Mea culpa... My force push to fix my braino in the last commit message has nuked https://github.com/scummvm/scummvm/commit/498d6e059bc2aa82158b91731ff5bcd377545fd5 off the branch... Do you want me to restore that or will you recommit? :|

This is the second braino this evening after the accidental branch push to scummvm/scummvm as a branch instead of this PR. I think I am going for dinner before I foul anything else up :)

P.S. I have a list of GCC warnings on this branch, which I will take a look at and push fixes for later... if someone doesn't get to them first :)

digitall:
Hmm... We have the old "This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. " warning from GH again. Will check the rebase history as well to see, but I wonder if the issue is using a force-push on a PR branch breaks GH's "Rebase and merge" function :|

SupSuper:
> Hmm... We have the old "This branch cannot be rebased due to conflicts
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. " warning from GH again. Will check the rebase history as well to see, but I wonder if the issue is using a force-push on a PR branch breaks GH's "Rebase and merge" function :|

Nope, I got the error from the start of the PR. So I rebased to try and fix it, and still same error. I think GitHub just doesn't like these big PRs...

digitall:
@SupSuper : Hmm... OK. Not optimal since it will need a manual rebase to merge as @sev- did for the Asylum engine PR... but we shall see.

digitall:
@bluegr : I have restored the commit of yours which I accidently nuked and pushed as d3497278a0f0a5b2fc2e780eed90e1c51541de04 ... Hope that is OK.

engines/trecision/3d.cpp
```diff
@@ -0,0 +1,2687 @@
...
+	int16 y;   // looping variable
+
+	if (y1 > _maxYClip)
+		y1 = _maxYClip;
+	if (y1 < _minYClip)
+		y1 = _minYClip;
+
+	int16 yBottom = y1;
+	int16 yTop = y1;
+	uint8 *texture = t->_texture;
```
mgerhardy, line 140: this buffer could be marked as const, as well as the `STexture` pointer

engines/trecision/actor.cpp
```diff
@@ -0,0 +1,481 @@
...
+		_face[i]._mat = ff->readUint16LE();
+	}
+
+	delete ff;
+
+	_curFrame = 0;
+	_curAction = hSTAND;
+
+	// fixup Microprose head correction
+#define P1 306
```
mduggan, line 231: I have a slight preference for using `static const int` for this sort of thing as it should compile to the same thing but avoids the potential problems of defines which stick around for the rest of the file

engines/trecision/dialog.cpp
```diff
@@ -0,0 +1,717 @@
...
+	}
+
+	_vm->_graphicsMgr->copyToScreen(0, 0, MAXX, TOP);
+
+	_vm->_flagDialogMenuActive = true;
+	_vm->_graphicsMgr->showCursor();
+}
+
+void DialogManager::updateChoices(int16 dmx, int16 dmy) {
+	if ((dmy >= MAXDISPCHOICES) && (dmy < (CARHEI * (_curDispChoice)+5)))
```
mduggan, line 87: nit: spaces around the + here

engines/trecision/actor.cpp
mgerhardy: +1 as this is not part of the namespace and will potentially break the unity cmake build that I'm using from time to time.

common/bitarray.h
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	~BitArray() {
+		free(_bits);
+	}
+
+	void set_size(uint bitcount) {
+		free(_bits);
+		_bitcount = bitcount;
+		_bits = (byte*)malloc((bitcount + 7) / 8);
```
mgerhardy, line 49: should imo be guarded by `bitcount != 0` check as `malloc(0)` might produce problems on some platforms.

eriktorbjorn:
I noticed that there are a couple of Valgrind warnings, just from looking at the intro and walking around a bit in the first room. I don't know if that's something that you want to deal with now or later.

[valgrind.txt](https://github.com/scummvm/scummvm/files/6506500/valgrind.txt)

engines/trecision/video.h
```diff
@@ -0,0 +1,117 @@
...
+class TrecisionEngine;
+
+class NightlongSmackerDecoder : public Video::SmackerDecoder {
+public:
+	bool loadStream(Common::SeekableReadStream *stream) override;
+	void muteTrack(uint track, bool mute);
+	void setMute(bool mute);
+	bool forceSeekToFrame(uint frame);
+	void forceSkip(uint frame);
+	bool endOfVideo() const;
```
mgerhardy, line 63: isn't an `override` missing here?

mgerhardy:
Kudos guys - nice work.

engines/trecision/actor.cpp
Strangerke: I've moved that fix in a separate function and turn those defines into static const as suggested.

engines/trecision/dialog.cpp
Strangerke: Done

Strangerke:
@eriktorbjorn i have fixed a couple of uninitialized values, but a little refactoring changed all the line numbers in 3d.cpp, so... hum, I'm a bit lost :) Could you check if things are slightly better now and send me an update valgrind trace? Thanks a lot! :)

eriktorbjorn:
> Could you check if things are slightly better now and send me an update valgrind trace? Thanks a lot! :)

@Strangerke Sure. I added some flags to Valgrind to maybe make it a bit more informative. It certainly made it a lot slower, so I didn't have the patience to let it play all the intro movies.

[valgrind.txt](https://github.com/scummvm/scummvm/files/6507665/valgrind.txt)

eriktorbjorn:
@Strangerke By the way, the main reason why I ran Valgrind was that I've seen the blue car disappear a few times while walking around it. I was hoping to find some clues as to why, but I don't know...

Strangerke:
@eriktorbjorn The disappearing car doesn't sound good :/ I never noticed, I'll check with @bluegr and @SupSuper if they know about it. I fixed a couple of things, one of them most likely fixed several issues... Could you recheck please?

eriktorbjorn:
@Strangerke To keep Valgrind happy, _compStream should be freed with `delete`, not `delete[]`, and _compBuffer should be allocated with `malloc()`, not `new`.

As for the disappearing car bug, I don't have any exact way to reproduce it. But if I walk around it for a bit, I can  usually get it to happen. And at least so far, it's always happened while walking near the doors on the shadowy side of the care. The object is apparently still there, because if I keep walking near where the car was, it will redraw it piece by piece.

eriktorbjorn:
@Strangerke Actually, I didn't even have to walk near the car for it to happen. If I just skip the intro movies and then wait for about half a minute - until the police (?) car in the background in the lower left corner of the picture drives away - both the blue car and the door to the building disappear for me.

Well, at least that makes it easy to reproduce.

Strangerke:
@eriktorbjorn yep, we replicated the issue, and I fixed _compBuffer :) sorry for the confusion. BTW, if it can help (who know?): if you switch on caps lock, the character moves his ass faster

bluegr:
@eriktorbjorn Thanks for reporting this issue, it has been fixed now

eriktorbjorn:
@bluegr I can confirm that the car seems to stay put now. Thanks!

The only Valgrind warnings I saw were possible memory leas:

It claimed that some stream allocated by loadAudioWav() wasn't freed. I'm a bit fuzzy on how that "dispose after use" thing actually works. I guess it could at the very least be a problem if the sound is loaded but never played?

It also claimed that the looping audio stream created in fadeIn() wasn't freed.

It claimed that doIdle() created a thumbnail that wasn't freed.

bluegr:
@eriktorbjorn Thanks! I believe that the memory leaks you reported should be plugged now. Could you please re-check?

eriktorbjorn:
@bluegr Hmm... Actually it seems a bit worse now, as far as the SoundManager is concerned.

[valgrind.txt](https://github.com/scummvm/scummvm/files/6509782/valgrind.txt)

eriktorbjorn:
@SupSuper Almost. This time I walked into a second room before quitting (from the in-game menu), and Valgrind claimed there was a leak in fadeIn(). Apparently the memory allocated by calling makeLoopingAudioStream() isn't freed? Other than that, it didn't say anything about leaks in the engine from this (admittedly trivial) test case. Good job!

bluegr:
@eriktorbjorn The sound code has been refactored, so could you try again now?

eriktorbjorn:
@bluegr Memory leaks are slippery things, apparently. I guess this latest log means that the memory stream used to create _stepLeftStream and _stepRightStream isn't freed?

[valgrind.txt](https://github.com/scummvm/scummvm/files/6513309/valgrind.txt)

bluegr:
@eriktorbjorn Thanks! My mistake. That leak has been fixed now

eriktorbjorn:
@bluegr I can confirm that this time, Valgrind did not warn about any memory leaks in the engine for my test case. Thanks!

engines/trecision/video.h
bluegr: Indeed, thanks! Fixed

engines/trecision/3d.cpp
```diff
@@ -0,0 +1,2687 @@
...
+		b++;
+	}
+
+	return len;
+}
+
+/*------------------------------------------------
+	Check if a point is inside a panel
+--------------------------------------------------*/
+bool PathFinding3D::pointInside(int pan, float x, float z) {
```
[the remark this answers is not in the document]
bluegr: Thanks, fixed

engines/trecision/3d.cpp
bluegr: Thanks! Fixed

digitall:
Couple of quick minor bugs / notes against a quick test with the latest i.e. 59579bc.
Firstly, the engine emits a couple of warnings at start which I think are associated with the autosave in slot #0:
````
WARNING: couldn't find thumbnail header type!
WARNING: trying to save thumbnail with bpp 0!
````
The thumbnail header type warning also occurs after load/save later so I think that needs a check.

digitall:
Secondly, quite a few of this warning after just walking through the first two rooms:
````
WARNING: setPosition: Unknown error : null light!
````

digitall:
Thirdly, if you hit "ESC" to access the original game settings menu screen which has save/load and then do Save from there using the ScummVM dialogs... in some cases which are a little unstable, the "continue" button on that screen to allow you to return to the game has no effect, leaving the player stuck at that screen. It is still possible to use "load" button to load a game and thus return to the game. Apart from this, saving on that settings screen saves the thumbnail as the settings screen rather than in the game, which is a bit wrong in terms of user expectation... i.e. many savegames showing progress though games... but all with thumbails of the settings screen :|

digitall:
Mea culpa. Wrong button... Wish GH had an "are you sure" on that "Close with comment" button :/

bluegr:
@digitall: Thanks, the warning during startup has been fixed - it was caused by the autosaving code, which was creating an invalid autosave. You may need to delete your autosave (nl-win.000) for the warning to go away

Strangerke:
@digital : regarding "WARNING: setPosition: Unknown error : null light!" - Fixed - It was a regression from cc5354521589fb8c2d441faefd5a3456dc1de4f6. Luckily, no impact except this warning displayed :)

bluegr:
@digitall The third issue with the saved games has been fixed now

digitall:
@bluegr @Strangerke: Thanks for sorting those issues. Will give it a bit more of a playtest when I get a chance.
One further issue: the ticket machine in the subway, when you try to use it with the card, the game plays two dialog lines over the top of each other. Will try to replicate again

digitall:
Yes, it is replicable. Just use the ticket machine. Ego says "I haven't got time for this"... then look at the Map to see "Rottmall 17" which you found in the apartment. Use the ticket machine again. When it asks for destination, Ego says "Rottmall 17", but also at the same time, "I haven't got time for this" :|

digitall:
Hmm... Another odd bug. If you look at the "Pen Pad" on Ruby's apartment door, Ego says "The fuse is burnt out. The wire between the contacts is broken" ... but that is the wrong audio. That should be played when you look at the Button / Panel in the previous scene with the elevator. This is fixed by using the foil on the fuse...

digitall:
Another minor bug, there is a graphical glitch when you leave Ruby's apartment back to the Elevator / Vending Machine room. It occurs to the right of the Ego.

digitall:
Still get the same double audio dialog with the ticket machine...
And still getting a few of these:
````
WARNING: trying to save thumbnail with bpp 0!
WARNING: couldn't find thumbnail header type!
````

SupSuper:
Manually merged.

PR #3017, opened 2021-05-19
GRAPHICS: MACGUI: fix the incorrect behaviour of macmenu in wage

ysj1173886760:
add intersection check with macmenu and macwindow
fix the offset problem


sev-:
Thanks!

PR #3018, opened 2021-05-19
ENGINES: Replace checkCD with more isolated partial methods

antoniou79:
Improve re-usability of the new partial methods, document each one's behavior and clarify the role of the previously generic void method named checkCD().

Original checkCD() as it was before this commit, would:
- Check for existence of extracted supported audio files (and return if such were found)
- Check if a specific issue where game data and audio CD data were read from the same CD drive, applied for Windows-only platform and if so, it would display a warning about the windows specific issue and return.
- If none of the above checks returned, it would display a warning about missing extracted audio files.

Prior to commit 669b9a0ffd75c8517c08c18a0f251a075ca4f7a1, the function would only run for Windows platform. After that commit, the first and last tasks of the function logic (as listed above) are ran for every platform, which fixes the issue whereby eg. Android or Linux ScummVM port would not display the warning for missing extracted audio files and would still play no sound, whereas Windows ScummVM would display the warning as intended. 

This PR attempts to split the generic named Engine::checkCD() into three partial methods, with descriptive names and better isolated behavior. This will hopefully improve re-usability and clarify exactly what checkCD() was doing when called. 

The logic of the original method should remain the same as before (after commit 669b9a0ffd75c8517c08c18a0f251a075ca4f7a1).

However, instead of calling the  (removed)  checkCD() method, we now replace it with an equivalent if () clause. 

I am unsure as of yet if the full checkCD() logic actually was supposed to apply in all of the places that it is in the code -- so developers familiar with the affected engines (CINE, DRASCULA, GOB, GROOVIE, KYRA, MADE, SCUMM) could re-check if the full new if () clause is needed or cuts/tweaks can be made on a case by case basis.


engines/engine.cpp
```diff
@@ -507,21 +521,30 @@ void Engine::checkCD() {
...
 			"See the Documentation (CD audio) for details."), _("OK"));
 		dialog.runModal();
-	} else {
-#endif // defined(WIN32) && !defined(__SYMBIAN32__)
-		// If we reached here, the game has audio tracks,
-		// it's not ran from the CD and the tracks have not
-		// been ripped.
-		GUI::MessageDialog dialog(
-			_("This game has audio tracks in its disk. These\n"
-			"tracks need to be ripped from the disk using\n"
```
athrxx, line 517: I wouldn't write "in its disk". Rather "on the disk" or "on the CD". However, I am not a native speaker of English, so be careful with my advice here...

engines/engine.cpp
```diff
@@ -507,21 +521,30 @@ void Engine::checkCD() {
...
-			"an appropriate CD audio extracting tool in\n"
-			"order to listen to the game's music.\n"
-			"See the Documentation (CD audio) for details."), _("OK"));
-		dialog.runModal();
```
athrxx, line 521: *documentation (small letter)

engines/engine.h
```diff
@@ -587,9 +587,16 @@ class Engine {
 	inline Common::SaveFileManager *getSaveFileManager() { return _saveFileMan; }
 
 public:
-	/** On some systems, check whether the game appears to be run from CD. */
-	void checkCD();
-
+	/** Check if extraced CD Audio files are found.
+	*/
```
bluegr, line 591: Nitpicking: Multi-line comments should be like this:



    /**
      * Check if extracted CD Audio files are found.
      */

bluegr: Also, it's "extracted"

engines/engine.h
```diff
@@ -587,9 +587,16 @@ class Engine {
...
+	bool existExtractedCDAudioFiles();
+	/** On some systems, check whether the game appears to be run
+	* from the same CD drive, which also should play CD audio.
+	*/
```
bluegr, line 595: Same here, concerning multi-line comments

antoniou79: Fixed in next commit: https://github.com/scummvm/scummvm/pull/3018/commits/f76acd499e81a06ec603e3beed8e97114d30230e

engines/engine.h
antoniou79: Fixed in commit: https://github.com/scummvm/scummvm/pull/3018/commits/f76acd499e81a06ec603e3beed8e97114d30230e

engines/engine.cpp
antoniou79: Fixed in https://github.com/scummvm/scummvm/pull/3018/commits/3ec3820ddbf70905a2aaba6d7c0374d1f77f796a

engines/engine.cpp
antoniou79: I put "on its CD" because it sounds better to me. Also replaced the "disk" in the next sentence with "CD". These changes are in commit:
https://github.com/scummvm/scummvm/pull/3018/commits/3ec3820ddbf70905a2aaba6d7c0374d1f77f796a

bluegr:
I'm wondering if it's worth breaking down the checkCD() function into several smaller ones.
The end result is the same, and these functions are used in place of the current checkcD() call, so this:

    checkCD();

becomes this:

    if (!existExtractedCDAudioFiles() && !areDataAndCDAudioReadFromSameCDDrive()) {
        warnMissingExtractedCDAudio();
    }

I personally don't see any benefit with this split, as it needs more verbose code to be used at the engine level. Perhaps I'm missing something?

antoniou79:
> I'm wondering if it's worth breaking down the checkCD() function into several smaller ones.
> The end result is the same, and these functions are used in place of the current checkcD() call, so this:
> 
> ```
> checkCD();
> ```
> 
> becomes this:
> 
> ```
> if (!existExtractedCDAudioFiles() && !areDataAndCDAudioReadFromSameCDDrive()) {
>     warnMissingExtractedCDAudio();
> }
> ```
> 
> I personally don't see any benefit with this split, as it needs more verbose code to be used at the engine level. Perhaps I'm missing something?

I understand that this would be a question, but I did try for a detailed explanation in the PR post, and added comments in the code for this purpose.

Currently, with this PR the if-clause code is equivalent but apparently more verbose -- obviously I agree it's not as short as a checkCD(). 

However, it "exposes" what checkCD() was doing in its final iteration (before the split), which was 3 things grouped albeit being heterogeneous. 

In the relevant discord discussion, there was confusion as to the method's purpose. It was not only addressing the Window's specific bug, it did more stuff but (wrongly) hid them under the macros guarding the Windows issue.

Also, maybe some engines currently using it, don't actually need the full "checkCD()" logic, so the replacement code would be simpler that the current if clause. I can't do this optimization in this PR, but the affected engine developers could do it.

Anyway, in the discord discussion, I suggested that the method should be split; sev wrote he agreed, hence the PR of how I thought the split should be made.

I'm ok if we end up dismissing this, although in that case the issue with checkCD() being obscure will remain.

antoniou79:
@sev- did you have the time to look into this?

sev-:
Thank you very much for looking into it.

The verbosity is exactly what is needed because the previous method name is not just misleading: it already led to a problem when It was misused.

This method was created in 2005 but then extended in 2008 which changed its behaviour.

This change reflects the actual functionality.

Thus, if there are no objections in several days, I'd like to merge this.

athrxx:
You did manage to find a shorter name for ``warnForGameHasCDAudioButNoExtractedFilesFound()``,
but there is still ``areDataAndCDAudioAreReadFromSameCDDrive()``.

How about naming
``areDataAndCDAudioAreReadFromSameCDDrive()``
to
``isDataReadFromAudioCD()``
?
(or to whatever else you can come up with that is shorter)

antoniou79:
> You did manage to find a shorter name for `warnForGameHasCDAudioButNoExtractedFilesFound()`,
> but there is still `areDataAndCDAudioAreReadFromSameCDDrive()`.
> 
> How about naming
> `areDataAndCDAudioAreReadFromSameCDDrive()`
> to
> `isDataReadFromAudioCD()`
> ?
> (or to whatever else you can come up with that is shorter)
Maybe areDataAndCDAudioReadFromSameCDDrive() to isDataAndCDAudioReadFromSameCD() ?
I don't want to obscure the actual issue that this method checks for (which is whether data *and* CD Audio read/played from the same CD)

engines/engine.cpp
```diff
@@ -454,25 +454,39 @@ void GUIErrorMessageFormat(Common::U32String fmt, ...) {
 	GUIErrorMessage(msg);
 }
 
-void Engine::checkCD() {
-#ifdef USE_VORBIS
+/**
+ * Checks if supported (extracted) audio files are found.
+ *
+ * @return			true if audio files of the expected naming scheme are found, as long as ScummVM
+ *					is also build with support to the respective audio format (eg. ogg, flac, mad/mp3)
```
bluegr, line 461:
```suggestion
 *					is also built with support to the respective audio format (eg. ogg, flac, mad/mp3)
```

bluegr:
> Thank you very much for looking into it.
> 
> The verbosity is exactly what is needed because the previous method name is not just misleading: it already led to a problem when It was misused.
> 
> This method was created in 2005 but then extended in 2008 which changed its behaviour.
> 
> This change reflects the actual functionality.
> 
> Thus, if there are no objections in several days, I'd like to merge this.

I believe it was me who extended its behavior. Since we want the engine code to be verbose in this case, we could rename the ``checkCD()`` function into something like ``checkAndWarnForMissingExtractedCDAudio()``, which would reflect its actual functionality. I have no strong preference on splitting the function into three parts with distinct functionality, I just feel it's unnecessarily verbose common code in engine code, plus these three functions will always be used together for this check anyway. Don't get me wrong, overall I'm fine with the change, I just wanted to voice my opinion on it :)

engines/engine.cpp
antoniou79: Fixed in https://github.com/scummvm/scummvm/pull/3018/commits/45acce61e2002b1f0f6bea5562ca93dd1709a350

sluicebox:
You'll need to rebase, as I made changes to Engine::checkCD() in PR #3003. It shouldn't be too bad, the "return" that you're changing to "return false" just moved.

antoniou79:
Ok, I think I've totally messed up the rebase to fix the conflict. I'll close this one and open a new PR with only the necessary commits.

PR #3020, opened 2021-05-20
AGOS: Fix restoreBlock() regression (needs testing!)

eriktorbjorn:
This is an attempt at fixing a regression in The Feeble Files: Pressing the "Off" button in the Oracle interface, causes the game to crash.

This appears to have been caused by the changes surrounding Elvira/PC-98/Jp. The restoreBlock() function generates an invalid Rect. The whole function looks a bit strange: It claims to take x, y, w and h, but in reality it seems to take left, top, right and bottom. Except in one case, it is called with the parameters in the wrong order, as far as I can tell.

This needs testing, of course. At a quick glance, things seem ok, but I'd like to hear some feedback on whether or not my changes look right. Also, I don't have Elvira/PC-98/Jp to test with.

### DIFF
--- a/engines/agos/agos.h
+++ b/engines/agos/agos.h
@@ -1212,7 +1212,7 @@ protected:
 	void colorBlock(WindowBlock *window, uint16 x, uint16 y, uint16 w, uint16 h);
 
 	void restoreWindow(WindowBlock *window);
-	void restoreBlock(uint16 x, uint16 y, uint16 w, uint16 h);
+	void restoreBlock(uint16 left, uint16 top, uint16 right, uint16 bottom);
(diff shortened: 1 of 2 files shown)


athrxx:
Looks good to me. I can confirm the bug in FF (Rect constructor assert due to invalid coordinates).

I quickly tested FF, ELVIRA1 PC-98 and PP with these changes. Seems like the issue is gone without any regressions, at least from what I tested.

digitall:
I can replicate the regression in FF here as well... Merging this to allow further testing by users to avoid possible AGOS game breaking bugs in the next release.

PR #3022, opened 2021-05-21
DIRECTOR: fix the builtin command NthFileName of lingo

ysj1173886760:
after the normilization of path style, we either get a relative path to the current directory, or get an absolute path, so we try to match the path with the game directory, and then do the things we want


sev-:
Thanks!

PR #3023, opened 2021-05-21
AGS: Fix file size in detection

aviloria:
Fix size value for "Bad Bunker"


criezy:
The current detection was reported in https://forums.scummvm.org/viewtopic.php?p=95421#p95421
I am wondering if it is possible that there are several versions of this game, or if that was an error (maybe a truncated file). In other word should the current detection be changed, or should a new one be added? Since we only use the first 5000 bytes for the MD5 hash, it is entirely possible that two different versions would have the same hash.

Also according to our commit message guidelines, the prefix for the commit message should be `AGS:` and not `[AGS]`, and we want commits to be rebased and not merged (so the merge commit is an issue). I can easily fix both issues by merging this pull request manually if you don't know how to fix it on your side (typically with an interactive rebase).

aviloria:
"Bad Bunker" was a game from the MAGS competition:
Link: https://archive.org/details/bad-bunker
Forum: https://www.adventuregamestudio.co.uk/forums/index.php?topic=21132.msg257734#msg257734
The original forum link (pointing to geocities) can be saved from archive.org, but it the same as the previous one:
https://www.adventuregamestudio.co.uk/forums/index.php?topic=21132.msg262827#msg262827
http://web.archive.org/web/20091027135531/http://geocities.com/colxfile/Stuff/bad_bunker.zip
I couldn't find any other link to the game.

MAGS competition was a montly amateur challenge to grant visibility to AGS devs (mostly amateurs), so it's possible there would be different versions, but quite rare.
Furhermore, it's also feasible there would be 2 different versions of the game, but it's really odd they have 1 exact digit difference in size.

About fixing the commit, could you please, do it for me? I'm not as proficient in Git to do that fast. Thank you in advance!

criezy:
Oh right. I didn't notice that only the first digit was changed.  It indeed looks like this was a mistake in the forum post.

criezy:
This was merged manually.
Thank you for the pull request.

PR #3025, opened 2021-05-23
AMIGAOS: Support the new reduced README.md for AmigaGuide conversion

raziel-:
I need help here, this PR is not complete.

I cannot delete, add or rename files, so please, whoever reviews this could you please

1) Rename dists/amigaos/RM2AG.rexx to dists/amigaos/md2ag.rexx
2) Rename dists/amigaos/RM2AG.rexx.in to dists/amigaos/md2ag.rexx.in
3) Delete dists/amigaos/convertRM.sed

I know the changes in the script files looks like a complete mess, but it was more or less a rewrite
and i figured that this way it would take away the least amount of time for the reviewer.

I hope i didn´t break anything

Thank you


lephilousophe:
Hi Hubert,
your changes would break the buildbot because the REXX interpreter there is ReginaREXX and not AREXX
You must use CALL for WRITELN.
I already did the compatibility work before: https://github.com/scummvm/scummvm/commit/8cb34f6c7da7a80a63614af89be904c7809a7eb7

Could you try to keep the compatibility? Thanks

raziel-:
@lephilousophe 

Truly sorry, my bad

Would the changes work with REGINA?

lephilousophe:
I will give it a try tomorrow

lephilousophe:
This works well with additional modifications.
Could you please test these and make sure I didn't break anything?

I suppose we will squash changes before merging in branch.

raziel-:
@lephilousophe 

On it, few changes needed

lephilousophe:
We've got a problem here. :smiley: 

On buildbot, build is done out of tree and README.md isn't here.
Specifying full path with "$(src_dir)/README.md" doesn't work. I don't know why but OPEN seems to fail silently.

Alternative workaround: copy the file to README.md.tmp and remove it afterwards.
That's really hackish but I don't really know how to use REXX to make it open arbitrary files.
What do you think?

raziel-:
@lephilousophe 
Yep, that´s fine. Just added a comment for my future forgettable self :-)

lephilousophe:
Let's squash all commits into one and merge this?

dists/amigaos/md2ag.rexx
```diff
@@ -0,0 +1,136 @@
...
+	*/
+	IF POS('[',v_line)>0 THEN DO
+		IF POS(']',v_line)=0 THEN DO
+			rejoin_line=READLN(read_md)
+			v_line=v_line''rejoin_line
+		END
+	END
+
+	/*
+	Replace any  '�' characters with '`'.
```
raziel-, line 88: This character was unfortunately changed during copying (or utf did something to it).
On my system it looks like an "a" with "^" on top, but it seems it gets converted somehow.

raziel-:
See my comment and i´m still doing some rewriting to the script, it doesnt pick up one weblink, have to find out whats wrong

If that is ok?

dists/amigaos/md2ag.rexx
lephilousophe: Huuuum, I wondered what it was.
It's mangled because Github editor expects UTF-8 but AmigaOS is still on old single-byte encodings.
Do you know which byte code it is?

lephilousophe:
No problem. I thought you had finished. :)

raziel-:
No idea about the bytecode, but it´s "â"

raziel-:
Problem is mainly that the README.md features some ' characters which get falsely translated on AmigaOS...i guess it might not be worth the hassle...maybe changing it in the README source to " would help...there´s a typo in the README.md too :-)

line 40 - completeable , superfluous "e"

I don´t know if it would help fixing it and also change the ' to "?

dists/amigaos/md2ag.rexx
lephilousophe: OK found it. It's the UTF-8 sequence for ’ character. It's a 3 bytes UTF-8 sequence e2 80 99. And you would like to translate it to a 1 byte.
It seems OVERLAY doesn't do this. Do you have an idea?
Else, we could patch the README to make it use ASCII only characters. We must see with the team for this though.

raziel-:
I'd like to display it as it is, but that won't work.
Let's go with the readme change.
I'll keep the change in place, no harm as it isn't found anyway and adapt it once there was a decision

lephilousophe:
Do you have iconv installed and if so, do you expect anyone building on AmigaOS to have it?
When copying README.md to README.tmp.md, we could convert it to ASCII with iconv.

raziel-:
I do...and if anyone's missing it, it can be installed later on

raziel-:
That should do it

lephilousophe:
Here is a fix using iconv.
It has bugs on Debian I worked around.
Could you test? Thanks.

lephilousophe:
Let's do this like you did previously with sed.
Not very elegant but if it worked for sed, it should work for iconv; :smile:

raziel-:
It may not be elegant, but your solution is :-)

Will test tomorrow

Thank you

raziel-:
Unfortunately, it doesnt work

iconv: (stdin): I/O error

raziel-:
Would that work on debian?

raziel-:
If the change doesn´t work on debian, i can try to convert it inside the rexx file, if that helps?

lephilousophe:
I added some small changes to make it work properly on Debian.
The LC_ALL workaround is back, the redirection in REXX script must be stuck to the filename (don't ask me why but with a space it fails) and the path was not correct in cp command.
Oh and my imperfect shim for DELETE doesn't handle more than 1 file so I split it.

raziel-:
Nice...and really embarrassing what I missed.

Actually...I could probably rewrite both rexx scripts to use binutils instead of AmigaOS commands.

The only workaround that has to stay unfortunately, is the makedir/mkdir one in amigaos.mk.
But it would make most shims obsolete.
Would that help?

btw, I can't test, change or add anything till 2nd of June...back at work, so go ahead and merge.
I'll test once I'm back home

lephilousophe:
I think it's better to leave it like this.
If it doesn't work when you come back, we will see but let's not fix something that works.
I will squash it because 21 commits is far too much :smile:

PR #3026, opened 2021-05-24
DIRECTOR: fix the cursor blinking problem in warlock

ysj1173886760:
in this movie
./scummvm --start-movie=STAMBUL/FNCOPS warlock
cursor is not visible when it opens the chatbox
the reason is it has a very frequent switch in _wm->_activewidget
for the speed and correctness, i've add the check when we are setting castmember editable
and the problem is solved


sev-:
Thanks!

PR #3028, opened 2021-05-28
MSVC: Add compiler printf validation

SupSuper:
Currently ScummVM uses the GCC_PRINTF annotation to validate strings in printf-like functions. This doesn't work for MSVC which has a different annotation for this, so I added it, and corrected some incorrect usages.

Sadly MSVC requires annotating the parameter, not the function, so I couldn't combine both macros into one.

### DIFF
--- a/backends/midi/timidity.cpp
+++ b/backends/midi/timidity.cpp
@@ -95,7 +95,7 @@ private:
 	int connect_to_server(const char* hostname, const char* tcp_port);
 
 	/* send command to the server; printf-like; returns reply string */
-	char *timidity_ctl_command(const char *fmt, ...) GCC_PRINTF(2, 3);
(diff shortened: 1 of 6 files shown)


bluegr:
Thanks for this. Indeed, this is the recommended way to mark such parameters in MSVC:
https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values?view=msvc-160

This will help identify incorrect usage of printf and will aid in development.

Thanks!

digitall:
@SupSuper @bluegr ... Looks like this broke builds on the NDS toolchain. The error is "error: attributes are not allowed on a function-definition" ... but not sure why that happens there and not on other POSIX type builds:
https://buildbot.scummvm.org/#/builders/87/builds/1007

SupSuper:
@digitall Nice catch, seems to be a bug with old GCC versions, should be fixed now.

PR #3029, opened 2021-05-28
GUI: Add delete/backspace support to save/load list

sluicebox:
As someone who accumulates a lot of temporary saves, I push the delete key a lot to try and delete them. Now it works!

This revealed that ListWidget::scrollToCurrent() has been doing a less-than comparison that includes the sentinel value -1, causing the list to jump to to the top when there is no currently selected item, such as right after deleting a save and clearing the selection. As that's behavior I fixed in the Delete GUI button in #2628, I added a check so that scrollToCurrent() doesn't scroll when there isn't a current item. This also fixes some existing unusual scroll position resets in the save/load list. I tested the other lists and didn't detect any behavior changes.

### DIFF
--- a/gui/saveload-dialog.cpp
+++ b/gui/saveload-dialog.cpp
@@ -463,7 +463,8 @@ void SaveLoadChooserSimple::handleCommand(CommandSender *sender, uint32 cmd, uin
 		updateSelection(true);
 		break;
 	case kDelCmd:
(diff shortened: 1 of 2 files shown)


bluegr:
Very nice, I really like that the delete key now has the same functionality in both the main game list and the save/load list.

Great work! Merging

PR #3034, opened 2021-05-30
GRAPHICS: MACGUI: selection fix in mactext

ysj1173886760:
fix the incorrect behaviour of selection in mactext and mactextwindow
provide basic edit operations for mactext, e.g. copy, cut, paste


sev-:
Thanks!

PR #3035, opened 2021-05-31
ACHIEVEMENTS: Improve AGS achievements support and add Statistics tab for game that are using SteamAPI Stats

lolbot-iichan:
1. AchievementsInfo is extended to contain Stats info, which are used in 2 scenarios:
i. When active domain is set in AchMan, all Stats are created with default values if they are not present in the INI-File
ii. UI tab is added to Game Options, both in the launcher and in-game options, featuring Stats names and current values

2. steam_achievements.py devtool is extended to provide Stats as well.

3. AGS achievements tables are updated to have Stats lists for game integrated with SteamAPI Stats.

4. Testbed engine is updated to have a sample stat, which is displayed as well.

\+ minor improvements at ags_galaxy_steam and achievements.cpp

Obligatory screenshots:
изображение
изображение
изображение
изображение

### DIFF
--- a/common/achievements.h
+++ b/common/achievements.h
@@ -50,6 +50,15 @@ enum AchievementsPlatform {
 };
 
 /**
+ * Information structure for game-specific statistics.
+ */
+struct StatDescription {
+	const char *id;      //!< Stat internal ID, such as "ITEMS_THROWN".
+	const char *comment; //!< Optional stat comment, such as "Items Thrown".
+	const char *start;   //!< Stat default value, such as "0".
+};
+
+/**
  * Information structure for game-specific achievements.
  */
 struct AchievementDescription {
@@ -65,6 +74,7 @@ struct AchievementDescription {
 struct AchievementsInfo {
 	Common::AchievementsPlatform platform;              //!< Achievements platform, such as "STEAM_ACHIEVEMENTS".
 	Common::String appId;                               //!< Achievements application ID of the given platform.
+	Common::Array<StatDescription> stats;               //!< Descriptions of all game stats.
 	Common::Array<AchievementDescription> descriptions; //!< Descriptions of all game achievements.
 
 	AchievementsInfo() { platform = Common::UNK_ACHIEVEMENTS; }
--- a/common/achievements.h
+++ b/common/achievements.h
@@ -166,7 +166,7 @@ public:
 	 *
 	 * @param[in] id	Internal ID of the achievement.
 	 */
-	String getStatRaw(const String &id);
+	const String getStatRaw(const String &id);
 
 	/**
 	 * Get an average rate statistic (float).
--- a/common/achievements.h
+++ b/common/achievements.h
@@ -162,6 +162,13 @@ public:
 	bool setStatFloat(const String &id, float value);
 
 	/**
+	 * Get a statistic (raw string).
+	 *
+	 * @param[in] id	Internal ID of the achievement.
+	 */
+	String getStatRaw(const String &id);
+
+	/**
 	 * Get an average rate statistic (float).
 	 * Calcucated by devision the sum of count by the sum of times.
 	 *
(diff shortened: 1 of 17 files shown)


common/achievements.h
```diff
@@ -151,6 +161,13 @@ class AchievementsManager : public Singleton<AchievementsManager> {
 	 */
 	bool setStatFloat(const String &id, float value);
 
+	/**
+	 * Get a statistic (raw string).
+	 *
+	 * @param[in] id	Internal ID of the achievement.
+	 */
+	String getStatRaw(const String &id);
```
mgerhardy, line 169: can be `const`

gui/options.cpp
```diff
@@ -1261,6 +1261,38 @@ void OptionsDialog::addAchievementsControls(GuiObject *boss, const Common::Strin
...
+
+	GUI::ScrollContainerWidget *scrollContainer;
+	scrollContainer = new GUI::ScrollContainerWidget(boss, prefix + "Container", "");
+	scrollContainer->setBackgroundType(GUI::ThemeEngine::kWidgetBackgroundNo);
+
+	uint16 nMax = info.stats.size();
+
+	uint16 lineHeight = g_gui.xmlEval()->getVar("Globals.Line.Height");
+	uint16 yStep = lineHeight;
+	uint16 ySmallStep = yStep/3;
```
mgerhardy, line 1276: formatting

engines/dialogs.cpp
```diff
@@ -331,13 +331,17 @@ ConfigDialog::ConfigDialog() :
...
 	//
-	// The Achievements tab
+	// The Achievements & The Statistics tabs
 	//
 	Common::AchievementsInfo achievementsInfo = metaEngine->getAchievementsInfo(gameDomain);
 	if (achievementsInfo.descriptions.size() > 0) {
 		tab->addTab(_("Achievements"), "GlobalConfig_Achievements");
 		addAchievementsControls(tab, "GlobalConfig_Achievements.", achievementsInfo);
 	}
+	if (achievementsInfo.stats.size() > 0) {
```
mgerhardy, line 341: it's a style thing, and doesn't make any difference for most of our containers - but what you really wanna check is imo

`if (!achievementsInfo.stats.empty()) {`

for containers that doesn't cache its own size, it makes a difference (as I said, it's not issue here - more a style thing)

devtools/steam_achievements.py
```diff
@@ -31,31 +31,56 @@ def cleanup_text(text):
...
 
 	scummvm_game_id = args.gameid
 	if not scummvm_game_id:
 		scummvm_game_id = game[0].lower().replace(' ', '').replace('-', '')
 		if args.verbose:
 			sys.stderr.write('missing scummvm game id - assuming {0}\n'.format(scummvm_game_id))
 
-	print("\t{\n\t\t\"%s\",\n\t\tCommon::STEAM_ACHIEVEMENTS,\n\t\t\"%s\",\n\t\t{" % (scummvm_game_id, args.steamid))
-	for i in range(entries):
+	print("\t{\n\t\t\"%s\",\n\t\tCommon::STEAM_ACHIEVEMENTS,\n\t\t\"%s\"," % (scummvm_game_id, args.steamid))
```
mgerhardy, line 58: It would be sweet if you could update the twine achievement table, too - as it would otherwise produce an invalid table for lba1 enhanced achievements.

lolbot-iichan: Good idea. Added at 6affb8e

engines/dialogs.cpp
lolbot-iichan: Very good point, since it's a common engine cpp, we need to keep it neat =)
Fixes at  081ad8c

gui/options.cpp
lolbot-iichan: Fixes at  081ad8c

common/achievements.h
lolbot-iichan: Fixed at 6fb6579

lolbot-iichan:
All checks have passed & review issues are fixed.

PR #3037, opened 2021-05-31
DETECTION: Fix MD5 cache when starting engine

SupSuper:
This fixes a bug with PR #2997 
Currently the MD5 cache is cleared when adding new games. However, starting games uses detection as well. So starting different games with same-name files will result in a "data not found" error, as the cache won't update.

This makes sure to clear the cache before starting new games. @a-yyg mentioned in the PR clearing the cache in detectFiles was a bad idea, so I put it here instead. Feedback welcome.


sev-:
good catch, thanks!

PR #3038, opened 2021-06-01
RFC: AUDIO: Add missing (?) mutex locking to MidiDriver_MT32GM::isReady()

eriktorbjorn:
While looking into why Lure of the Temptress takes so long to start with MT-32 audio (turns out that it's because it is setting up a lot of custom sounds?), I noticed what to me looks like a missing mutex lock in MidiDriver_MT32GM::isReady().

Lure of the Temptress calls isRead() to see if the driver has finished processing all the SysEx data. This checks if the queue is empty, but it does so without waiting for the mutex to unlock. Doesn't that mean the queue can be accessed simultaneously from two different threads?

Of course, such problems are notoriously hard to debug. I added some debug() calls, and it _looked_ like it happened a couple of times... assuming I can count on such debug messages to be printed in the correct order.

I tried asking around, but there was no one there who could answer me. So now I'm creating this pull request to see if there's any feedback.

### DIFF
--- a/audio/mt32gm.h
+++ b/audio/mt32gm.h
@@ -298,7 +298,10 @@ public:
 	virtual int open(MidiDriver *driver, bool nativeMT32);
 	void close() override;
 	bool isOpen() const override { return _isOpen; }
-	bool isReady() override { return _sysExQueue.empty(); }
+	bool isReady() override {
+		Common::StackLock lock(_sysExQueueMutex);
+		return _sysExQueue.empty();
+	}
 	uint32 property(int prop, uint32 param) override;
 
 	using MidiDriver_BASE::send;


NMIError:
I posted a reply a little while ago in Discord, but that might have gotten overlooked in the traffic...

The use case for the SysEx queue is:
1. Put SysExes on the queue
2. Check isReady while an SDL timer sends the SysExes (possibly using another thread)

isReady uses Queue::isEmpty, which is a read-only operation, so it does not affect the queue in any way. The only possible issue I can see is if Queue::isEmpty returns true when the queue is not actually empty due to some concurrent modification, but I don't think it does that.

eriktorbjorn:
Maybe you're right. Threads and mutexes scare me, so I figured it was better to report it and ask for feedback than to just keep quiet and worry.

NMIError:
Sure, thanks for taking an interest!

Maybe another dev can add an opinion on possible concurrency issues between Queue::pop and Queue::isEmpty?

sev-:
The concurrency in this case is quite possible because we're running sound in a separate thread. In any case, this mutex wouldn't hurt.

Merging.

PR #3040, opened 2021-06-02
SDL: Add new pixel perfect stretch mode

criezy:
The commit in this pull request add a new pixel perfect mode. This mode ensure that an integral scaling from the original game resolution is used both for the width and for the height, thus providing homogeneous pixels. Compared to the old pixel perfect stretch mode it thus differs in the way it handles the aspect ratio correction as it may stretch or squeeze the vertical direction to snap it to a multiple of the original game height. When aspect ratio correction is off, the two modes should behave in exactly the same way.

For example with DOTT, which has an original resolution of 320x200, when I use aspect ratio correction and fullscreen on my laptop (with a screen resolution of 2880x1800) I get:
 - Old Pixel Perfect: Stretch (320, 200) -> (2240, 1680) with scaling x7 and x8.4
 - New Pixel Perfect: Stretch (320, 200) -> (2240, 1600) with scaling x7 and x8

So the new mode provides a compromise between the aspect ratio correction and getting perfect pixels.

I am creating this as a draft pull request as I have two questions:
 1. Should it replace the old pixel perfect stretch mode or do we want both? I would tend to keep both as some players may want a perfect aspect ratio and others perfect pixels.
 2. If we want both, how should they be named? For now I have used a temporary name that is not meant to be the final name. Maybe use `Pixel-perfect scaling` for the new one and renamed the old one to `Almost pixel-perfect scaling`? (I am joking here, but I am really struggling to find descriptive names for both).

See https://forums.scummvm.org/viewtopic.php?f=2&t=16297 for background on why I am proposing this change.

Also the documentation will need to be updated, but I am waiting on suggestions and opinions on the two questions above before finalising the change and updating the documentation.

### DIFF
--- a/backends/graphics/opengl/opengl-graphics.cpp
+++ b/backends/graphics/opengl/opengl-graphics.cpp
@@ -241,6 +241,7 @@ namespace {
 const OSystem::GraphicsMode glStretchModes[] = {
 	{"center", _s("Center"), STRETCH_CENTER},
 	{"pixel-perfect", _s("Pixel-perfect scaling"), STRETCH_INTEGRAL},
+	{"even-pixels", _s("Even pixels scaling"), STRETCH_INTEGRAL_AR},
 	{"fit", _s("Fit to window"), STRETCH_FIT},
 	{"stretch", _s("Stretch to window"), STRETCH_STRETCH},
 	{"fit_force_aspect", _s("Fit to window (4:3)"), STRETCH_FIT_FORCE_ASPECT},
--- a/backends/graphics/windowed.h
+++ b/backends/graphics/windowed.h
@@ -33,9 +33,10 @@
 enum {
 	STRETCH_CENTER = 0,
 	STRETCH_INTEGRAL = 1,
-	STRETCH_FIT = 2,
-	STRETCH_STRETCH = 3,
-	STRETCH_FIT_FORCE_ASPECT = 4
+	STRETCH_INTEGRAL_AR = 2,
+	STRETCH_FIT = 3,
+	STRETCH_STRETCH = 4,
+	STRETCH_FIT_FORCE_ASPECT = 5
 };
 
 class WindowedGraphicsManager : virtual public GraphicsManager {
@@ -194,11 +195,11 @@ protected:
 			return;
 		}
 
-		populateDisplayAreaDrawRect(getDesiredGameAspectRatio(), getWidth() * getGameRenderScale(), _gameDrawRect);
+		populateDisplayAreaDrawRect(getDesiredGameAspectRatio(), getWidth() * getGameRenderScale(), getHeight() * getGameRenderScale(), _gameDrawRect);
 
 		if (getOverlayHeight()) {
 			const frac_t overlayAspect = intToFrac(getOverlayWidth()) / getOverlayHeight();
-			populateDisplayAreaDrawRect(overlayAspect, getOverlayWidth(), _overlayDrawRect);
+			populateDisplayAreaDrawRect(overlayAspect, getOverlayWidth(), getOverlayHeight(), _overlayDrawRect);
 		}
 
 		if (_overlayVisible) {
@@ -363,7 +364,7 @@ protected:
 	int _cursorX, _cursorY;
 
 private:
-	void populateDisplayAreaDrawRect(const frac_t displayAspect, int originalWidth, Common::Rect &drawRect) const {
+	void populateDisplayAreaDrawRect(const frac_t displayAspect, int originalWidth, int originalHeight, Common::Rect &drawRect) const {
 		int mode = getStretchMode();
 		// Mode Center   = use original size, or divide by an integral amount if window is smaller than game surface
 		// Mode Integral = scale by an integral amount.
@@ -372,7 +373,7 @@ private:
 		// Mode Fit Force Aspect = scale to fit the window while forcing a 4:3 aspect ratio
 
 		int width = 0, height = 0;
-		if (mode == STRETCH_CENTER || mode == STRETCH_INTEGRAL) {
+		if (mode == STRETCH_CENTER || mode == STRETCH_INTEGRAL || mode == STRETCH_INTEGRAL_AR) {
 			width = originalWidth;
 			height = intToFrac(width) / displayAspect;
 			if (width > _windowWidth || height > _windowHeight) {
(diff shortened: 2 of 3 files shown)


Mataniko:
For pixel perfect stretch mode you'd want to do:
(320, 200) -> (1600, 1200) with scaling x5 and x6. This gives you an aspect ratio corrected 4:3 ratio.

2240/1600 gives you a 7:5 ratio.

the problem with 1600x1200 scaling is that the most common resolution for monitors is 1920x1080, so this would only work for 2560x1440 monitors and above.

We can change aspect ratio correct from a checkbox to a radio -> "Aspect Ratio Correction, Pixel Perfect Aspect Ratio Correction"

Note that we can not achieve pixel perfect resolution for 640x400 resolution games without resizing to a ridiculous resolution (3200, 2400)

Mataniko:
Ideally we can implement arbitrary "near perfect" scaling with the following algorithm:

If 320x200 and resolution > 1600x1200:
  1. return (320*5, 200*6)
else if 
  1. scale res to 640x1200 using nearest neighbor (works for both 320x200 and 640x400)
  2. scale down using bilinear algorithm to 640x480

And then we can just give the best aspect ratio correction regardless of resolution. In this case we would only have "aspect ratio correction" checkbox like we do now

See details here: https://www.gamasutra.com/blogs/FelipePepe/20150423/241730/No_MSDOS_games_werent_widescreen_Tips_on_correcting_aspect_ratio.php

criezy:
I was not really looking for feedback on the algorithm itself. The one I implemented here is simple and generic and will give you the "best" aspect ratio (as in closest to the target aspect ratio) that fits the window/screen and provide integral scaling both horizontally and vertically. For your 1920x1080 screen for example, it will display the game at a resolution of 1280x1000 (so x4 scaling horizontally and x5 scaling vertically). This means an aspect ratio of 320x250 instead of 320x240, but you get even pixels.

The idea here is indeed not to get exactly a 4:3 aspect ratio, but something that approximate it while filling the screen as much as possible and not introducing any scaling artefact.

I am aware of the link you give here, but that would be more complicated to implement, and would also introduce a slight blurring vertically. Currently for the OpenGLGraphicsManager we let SDL do all the scaling. We only need to tell SDL in which area of the window to draw the game screen. Implementing what you suggest would require manually doing the scaling ourself in software (or at least part of it). And I have no intention to look at that at this point.

Note also that to implement what you suggest you don't need to scale to 640x1200 first. You can directly do the scaling from the original game screen to 640x480 and get exactly the same result. The article does it this way because it uses existing software to scale an image. But when you scale to 640x1200 you are only duplicating pixels, so all the pixels are already in the original image. And you can easily map pixels back to the original ones (e.g. divide x by 2 and y by 6 with an integer division if the original is a 320x200 image). This means you can easily rewrite the equations to do a bilinear downscale from 640x1200 to instead do an upscale from 320x200 (or 640x400). This means that for the SurfaceSdlGraphicsManager, where the aspect ratio correction and stretching are separate, and we already do the former in software, we could possibly improve the aspect ratio correction when filtering is enabled by changing the equations in `stretch200To240()` to use this approach (assuming this is not what `kSlowAndPerfectAspectMode` does already, I never really looked at that code).

Mataniko:
I'm not sure I fully understand, are you saying that we scale (2240, 1600) then let SDL handle scaling to fit the surface?

Why would we call this pixel perfect scaling where we're messing up the aspect ratio from the get go?

Wouldn't it still be better to just scale to 1600x1200 which would at least give a chance for monitors with capable resolutions to have actual pixel perfect scaling? 

In that scenario, we change the internal resolution to 1600x1200 and either: 
- let SDL fit to surface "near pixel perfect"
- Letter/Pillarbox  "real pixel perfect"

At 2240x1600 we will always perform some sort of scaling, based on the wrong pixel ratio.

criezy:
> I'm not sure I fully understand, are you saying that we scale (2240, 1600) then let SDL handle scaling to fit the surface?

No, we don't scale at all and let SDL handle the scaling and stretching from the original size, for example 320x200, to whatever size we tell it to scale to, for example 1600x1200 or 2240x1600. The stretch modes control this final size.

Note: What I wrote above is correct with OpenGL, but for the other graphics mode we can do a 2x, 3x, or 4x scaling, and aspect ratio correction stretching, before we let SDL handle the final stretch to the final size (which is one of the reason why I don't like those other graphics mode and think that once we have shaders implemented for OpenGL we should get rid of the other graphics mode; which would also greatly simplify the code and decrease the maintenance burden; but this is a separate topic).

> Why would we call this pixel perfect scaling where we're messing up the aspect ratio from the get go?

Because it gives perfect even pixels (all the original pixels are scaled to even-sized pixels in the final image with no blurring). The aspect ratio of those pixels is not exactly 5:6, but then that might have been the case on the originals as well depending how you setup your CRT screen (e.g. with overscan).

> Wouldn't it still be better to just scale to 1600x1200 which would at least give a chance for monitors with capable resolutions to have actual pixel perfect scaling?

We can already do that with our current pixel perfect mode (or the new one proposed here) if you set your window size to something between 1600x1200 and 1920x1440. We could indeed consider a case where we want a strict pixel-perfect and aspect ratio correction stretch which would only give 1600x1200 for any resolution between 1600x1200 and 3200x2400, which could be useful for full screen modes. And this can easily be implemented with another stretch mode.

It's difficult to say which would be better though as different users will have different opinion depending on what they want to see. Some users want big pixels that all have the same size and have no blur at all, and do not mind some small errors on the aspect ratio. Some other users prefer an exact aspect ratio even if that means some blurring of the pixels or pixels of different size, and others might prefer an exact aspect ratio with perfectly crisp and even pixels, even if that means having big black bars all around the screen. We have those various stretch modes to cater to various user preferences.

OmerMor:
If I understand correctly, it seems like the users need to rank their top 2 options out of the following 3:
- Pixel-Perfect Scaling
- Aspect Ratio Correction
- Full Screen

criezy:
> Note: What I wrote above is correct with OpenGL, but for the other graphics mode we can do a 2x, 3x, or 4x scaling, and aspect ratio correction stretching, before we let SDL handle the final stretch to the final size

And I realised when writing that that the new pixel-perfect stretch mode proposed here is useless for non-OpenGL graphics modes since for other graphics modes the pixels will already have been messed with by the aspect ratio correction applied before the stretch. So I have forced-pushed a change to only have this new stretch mode for OpenGL.

backends/graphics/opengl/opengl-graphics.cpp
```diff
@@ -241,6 +241,7 @@ namespace {
 const OSystem::GraphicsMode glStretchModes[] = {
 	{"center", _s("Center"), STRETCH_CENTER},
 	{"pixel-perfect", _s("Pixel-perfect scaling"), STRETCH_INTEGRAL},
+	{"pixel-perfect-ar", _s("Pixel-perfect scaling AR"), STRETCH_INTEGRAL_AR},
```
sev-, line 244: This requires an ```// I18N: AR means Aspect Ratio(?)``` comment for translators.

sev-:
Looks good to me, besides a required explanation for the GUI translators

backends/graphics/opengl/opengl-graphics.cpp
criezy: As I noted in the PR description, this was a placeholder name as I could not find a good name for it. I have now changed it to "Even pixels scaling"

criezy:
I am going to merge this now. Further changes can be done in master directly if needed.

PR #3042, opened 2021-06-03
JANITORIAL: Remove extraneous period from math/matrix.h

BallM4788:
When the ```math``` folder is added to ```doc/doxygen/scummvm.doxyfile``` as an input directory, the period would cause the doxygen build to fail.


bluegr:
Thanks for this! I guess it wasn't noticed before, as this is the only place where a namespace is defined inside a DocBlock

PR #3043, opened 2021-06-04
ACHIEVEMENTS: Move achievements data to achievements.dat

lolbot-iichan:
Introduced create_achievements.sh to generate achievements.dat (using python script based on original steam_achievements.py).

Technically, achievements.dat is a ZIP-archive, containing INI-files for different games.
All the INI files are generated using SteamDB except for hand-written INI-file for Testbed.

File achievements.dat is added to various dat file lists, so I hope it will get to the ScummVM builds.
AchMan loads this dat file each time the active domain is changing. 

Metaengine's GetAchivementsInfo() now just returns a pair of platform+appId. So, data string are removed from achivements tables. Wintermute and GUI now get achivements descriptions from AchMan, not from metaengine's AchivementsInfo.

Example INI-file for Steam statistics and achivements details:
```
[stats:en]
item_0_id=ExamStrikes
item_0_comment=Strikes on the Center 7 Exams
item_0_start=0
item_1_id=Mementos
item_1_comment=Hidden notes found in Center 7
item_1_start=0
item_2_id=LostHealth
item_2_comment=Health lost during the fights
item_2_start=0
item_3_id=MagsFound
item_3_comment=Extra ammo magazines found
item_3_start=0
item_4_id=EggsFound
item_4_comment=Easter Eggs Discovered
item_4_start=0
[achievements:en]
item_0_id=GRADE_A_STUDENT
item_0_title=Grade A Student
item_0_comment=Complete all Center 7 Exams without any strikes
item_1_id=MEMENTOS
item_1_title=Mementos
item_1_comment=Find all 6 hidden notes in Center 7
item_2_id=GOLDEN_GUN
item_2_title=Golden Gun
item_2_comment=Complete all fights without losing any health
item_3_id=SCAVENGER
item_3_title=Scavenger
item_3_comment=Find all 3 extra ammo magazines
item_4_id=SPACE_COWBOY
item_4_title=Space Cowboy
item_4_comment=Find all 4 hidden Easter Egg Characters
item_5_id=DIPLOMACY
item_5_title=Diplomacy
item_5_comment=Evade the Boryokudan rather than fight them
item_6_id=STAIR_MASTER
item_6_title=Stair master
item_6_comment=Take the stairs instead of the elevator
item_7_id=GOOD_SAMARITAN
item_7_title=Good Samaritan
item_7_comment=Heal the fellow Man on Barracus
item_8_id=RESEARCH_PAPER
item_8_title=Research Paper
item_8_comment=Find all Database Articles
item_9_id=GREAT_ESCAPE
item_9_title=The Great Escape
item_9_comment=Escape Center 7 without losing any health
item_10_id=MEMORIES
item_10_title=Memories
item_10_comment=Access the Memory Database in the Director’s Office in Center 7
item_11_id=GEMINIS
item_11_title=Geminis
item_11_comment=Be able to control both Azriel & Delta-Six
item_12_id=HARD_COPY
item_12_title=Give me a Hard Copy
item_12_comment=Analyze a photo using the terminal
item_13_id=RUE_DAY
item_13_title=Rue the Day
item_13_comment=Complete the game once
item_14_id=RUE_COMMENTARY
item_14_title=Rue the Commentary
item_14_comment=Complete the game once in Commentary mode
```


devtools/create_achievements/create_achievements.sh
```diff
@@ -0,0 +1,87 @@
...
+python steam_achievements.py -v --steamid 291710
+python steam_achievements.py -v --steamid 337130
+python steam_achievements.py -v --steamid 340370
+python steam_achievements.py -v --steamid 349140
+python steam_achievements.py -v --steamid 405780
+python steam_achievements.py -v --steamid 574420
+python steam_achievements.py -v --steamid 1064660
+
+zip -9j achievements.dat gen/* static/*
+mv -vf achievements.dat ../../dists/engine-data
```
mgerhardy, line 87: Missing newline. 

I'm not sure about the guidelines here - but we should maybe check for 7zip being available. It produces smaller files. But this is most likely something for a new PR - as this might touch more locations.

devtools/create_achievements/steam_achievements.py
```diff
@@ -0,0 +1,99 @@
...
+	achievements_rows = response.html.xpath("//tr[starts-with(@id, 'achievement-')]/td")
+	achievements_columns = 3 # id, text, img
+	achievements_entries = int(len(achievements_rows) / achievements_columns)
+	if achievements_entries == 0:
+		sys.exit(127)
+
+	if args.verbose:
+		sys.stderr.write("found {0} achievements\n".format(achievements_entries))
+
+	stats_rows = response.html.xpath("//tr[starts-with(@id, 'stat-')]/td")
```
mgerhardy, line 44: As this is now more a less a tool that is needed in order to generate the dat file. We should add some more error checking and error messages here. If for example (as it happened in the past) steam changes the internal representation of their steamdb generated html - we might produce invalid files here. So it would be nice to add an option to error out if e.g. no achievement was found (as we expect to find them for the specified game ids). And also check that the format we expect is still found in the downloaded document. Please add exit codes != 0 for those error cases and make sure to check the return value in the `create_achievements.sh` script

devtools/create_achievements/create_achievements.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/sh
+
+#AGS games:
+python steam_achievements.py -v --steamid 80310
```
mgerhardy, line 4: This should get added to a function. It reduces duplication but more importantly it adds the ability to perform error checking and error reporting in a single location. The exit codes of `steam_achievements.py` should get checked here - see my comment below

gui/options.cpp
```diff
@@ -1276,12 +1273,10 @@ void OptionsDialog::addStatisticsControls(GuiObject *boss, const Common::String
...
 
 	for (uint16 idx = 0; idx < nMax ; idx++) {
-		Common::String key = info.stats[idx].id;
-		if (info.stats[idx].comment) {
-			key = info.stats[idx].comment;
-		}
+		const Common::StatDescription *descr = AchMan.getStatDescription(idx);
 
-		Common::String value = AchMan.getStatRaw(info.stats[idx].id);
+		Common::String key = descr->comment.empty() ? descr->id : descr->comment;
```
mgerhardy, line 1278: There is imo no need to copy the string - so using a const ref here would be enough (`id` and `comment` are `String` instances now, no?).

mgerhardy:
Nice job - kudos

Thunderforge:
It looks like some of the descriptions for the hidden achievements aren't being loaded. For instance, [Kathy Rain Steam](https://github.com/scummvm/scummvm/blob/997140c4dde9020cab24d36a16d113173ef2d9d1/devtools/create_achievements/gen/steam-370910.ini) has them for the non-hidden achievements, but not the hidden ones.

devtools/create_achievements/steam_achievements.py
lolbot-iichan: Fixed at b6049e16f2d855a589a178d84efaac64e25ef37f
Shell script now work with "set -e" and fails on first error.
Some error-checking added to refactored python script.

devtools/create_achievements/create_achievements.sh
lolbot-iichan: Introduced functions at 7b46c8b1b5b8780d94164d2c08810e932cf5b845
No need to check $? in them since script works at "set -e" mode now.

devtools/create_achievements/create_achievements.sh
lolbot-iichan: Newline fixed.

Thanks for pointing at 7zip. I have added a TODO to remember about it at 5f5f63777760fef5d920b4d0c206896aceea6f9b
Also, i added `touch --date` and `zip -X` to get the same dat file on different runs.

gui/options.cpp
lolbot-iichan: Thanks! Fixed at 312394e0802888ef62b983c9ab59ddd3578a7265

lolbot-iichan:
> It looks like some of the descriptions for the hidden achievements aren't being loaded. For instance, [Kathy Rain Steam](https://github.com/scummvm/scummvm/blob/997140c4dde9020cab24d36a16d113173ef2d9d1/devtools/create_achievements/gen/steam-370910.ini) has them for the non-hidden achievements, but not the hidden ones.

Thanks for sharing this.
I believed that steamdb and steamcommunity are not showing achievements descriptions since it's not supported.
I'll surely switch to additional / another information source some time later.

devtools/create_achievements/create_achievements.sh
```diff
@@ -0,0 +1,146 @@
...
+    echo "Usage: $0 [OPTION]..."
+    echo "    -f, --force    Forced regeneration of all files"
+    echo "    -h, --help     Show usage text"
+    echo "    -v, --verbose  Verbose output"
+    exit 42
+}
+
+add_steam() {
+    if [[ "$FORCE" == "1" || ! -f "gen/steam-$1.ini" ]]; then
+        python steam_achievements.py ${VERBOSE:+"-v"} --steamid "$1"
```
mgerhardy, line 17: These are bash extensions, no? If so we should say /bin/bash - not /bin/sh (but I might be wrong here)

digitall: @mgerhardy You are not wrong.
@lolbot-iichan : There is a tool to check for POSIX shell compliance called "checkbashisms": https://sourceforge.net/projects/checkbaskisms/

digitall: Current output from my run is:
````
possible bashism in create_achievements.sh line 5 (trap with ERR|DEBUG|RETURN):
trap "echo FAILURE: $0 failed to create achievements.dat" ERR
possible bashism in create_achievements.sh line 16 (alternative test command ([[ foo ]] should be [ foo ])):
    if [[ "$FORCE" == "1" || ! -f "gen/steam-$1.ini" ]]; then
possible bashism in create_achievements.sh line 16 (should be 'b = a'):
    if [[ "$FORCE" == "1" || ! -f "gen/steam-$1.ini" ]]; then
possible bashism in create_achievements.sh line 23 (alternative test command ([[ foo ]] should be [ foo ])):
    if [[ "$FORCE" == "1" || ! -f "gen/galaxy-$2.ini" ]]; then
possible bashism in create_achievements.sh line 23 (should be 'b = a'):
    if [[ "$FORCE" == "1" || ! -f "gen/galaxy-$2.ini" ]]; then
possible bashism in create_achievements.sh line 42 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ "$FORCE" == "1" ]]; then
possible bashism in create_achievements.sh line 42 (should be 'b = a'):
if [[ "$FORCE" == "1" ]]; then
possible bashism in create_achievements.sh line 133 (alternative test command ([[ foo ]] should be [ foo ])):
if [[ "$VERBOSE" == "-v" ]]; then
possible bashism in create_achievements.sh line 133 (should be 'b = a'):
if [[ "$VERBOSE" == "-v" ]]; then
````

lolbot-iichan: @mgerhardy Thanks for raising this. I always forget that sh != bash. Fixed at 2ee2335

@digitall Thanks for sharing an excellent tool. Used it to check commit before sending.

digitall: No problem!

PR #3044, opened 2021-06-04
DIRECTOR: allow setting of video casts to castNum

rvanlaar:
The game Majestic dynamically sets the castNum to video casts and expects them to play directly. Videos were also stopped erroneously. Video stoppage is assumed when movieRate is 0. The comment indicated that it was probably broken for D4.

The corresponding lingo code is: *anim is a cast label*

```
  set the castNum of sprite 48 to cast(anim)
  updateStage()
  set the movieRate of sprite 48 to 1
  repeat while the movieRate of sprite 48
    updateStage()
  end repeat
```

### DIFF
--- a/engines/director/channel.cpp
+++ b/engines/director/channel.cpp
@@ -291,16 +291,6 @@ void Channel::setClean(Sprite *nextSprite, int spriteId, bool partial) {
 			}
 		}
 
-		// If the previous sprite in this channel was a video, but the new sprite is different,
-		// stop the playback.
-		// TODO: In D4 *any* channels playing the same cast member across a frame change seems
-		// to be enough to keep the same playback going, but the behaviour seems too glitchy
-		// to be reliable? Hopefully nothing relies on this.
-		if (_sprite && _sprite->_cast && _sprite->_cast->_type == kCastDigitalVideo &&
-				_sprite->_castId != nextSprite->_castId) {
-			((DigitalVideoCastMember *)_sprite->_cast)->stopVideo(this);
-		}
-
 		if (_sprite->_puppet || partial) {
 			// Updating scripts, etc. does not require a full re-render
 			_sprite->_scriptId = nextSprite->_scriptId;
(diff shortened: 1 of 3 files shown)


engines/director/lingo/lingo-the.cpp
```diff
@@ -1278,6 +1278,20 @@ void Lingo::setTheSprite(Datum &id1, int field, Datum &d) {
 	case kTheCastNum:
 		{
 			int castId = d.asCastId();
+			CastMember *castMember = g_director->getCurrentMovie()->getCastMember(castId);
+
+			if (castMember && castMember->_type == kCastDigitalVideo){
```
sev-, line 1283: Space before the opening bracked it missing

engines/director/channel.cpp
```diff
@@ -291,16 +291,6 @@ void Channel::setClean(Sprite *nextSprite, int spriteId, bool partial) {
 			}
 		}
 
-		// If the previous sprite in this channel was a video, but the new sprite is different,
-		// stop the playback.
-		// TODO: In D4 *any* channels playing the same cast member across a frame change seems
```
sev-, line 296: I wonder, what kind of side effects it could bring, since now we're not stopping the video playback.

Could you please elaborate a bit?

sev- (changes requested):
Added couple of comments

engines/director/channel.cpp
rvanlaar: I'll elaborate. The logic used to check if a video should be stopped is broken for Majestics case. 
The video is stopped before it can be played. 

What I've seen so far is that a video stops playing when the movieRate is 0. 
Either because the movie was finished, or the movieRate as set to 0.

It seems to me the .stop() in [1] replaced the functionality meant here.
Which calls [2] directly, instead of indirectly. 

Do you know of any Director files that rely on the original behavior that I could check against?

[1]https://github.com/scummvm/scummvm/blob/master/engines/director/castmember.cpp#L421
[2]https://github.com/scummvm/scummvm/blob/master/video/video_decoder.cpp#L422

sev-:
Thanks!

PR #3046, opened 2021-06-04
DISTS: Initial Emscripten Demo / Proof of Concept

chkuendig:
This is an initial version of my Emscripten/Webassembly target as mentioned on Discord a few days ago (see video https://discord.com/channels/581224060529148060/581224061091446795/849353211877523536)

There's still a lot to do depending on how such a build would be deployed. I summarized quick build instructions and a summary of what's done and what's still missing at [./dists/emscripten/README.md](https://github.com/chkuendig/scummvm/blob/emscripten/dists/emscripten/README.md). 

I built a demo based on this (running `./dists/emscripten/build.sh all`) and uploaded it to http://scummvm.kuendig.io/scummvm.html. This includes some freeware games, some demos and the ScummVM testbed.

I only tested this on macOS, but at least in theory this should also build on Linux and WSL (or anything else that can run shell scripts and is supported by emsdk).

**Update 7th June:** I fixed a few more minor issues and updated the demo at http://scummvm.kuendig.io/scummvm.html - feel free to retest.  Let me know what you think. I'll keep improving this in the coming days and address any further review comments.

**Update 21st July:** I didn't have much time in the last few weeks to work on this. I still plan though to get this merged this summer/asap. But it might take a few more weeks until I'll be able to address the remaining reviews.


engines/grim/grim.cpp
```diff
@@ -1117,6 +1117,10 @@ void GrimEngine::mainLoop() {
 			g_imuseState = -1;
 		}
 
+		#if defined(EMSCRIPTEN)
```
aquadran, line 1120: there is already yield here. is playback too much slow?

graphics/opengl/framebuffer.cpp
```diff
@@ -158,12 +158,20 @@ void FrameBuffer::init() {
 		glBindRenderbuffer(GL_RENDERBUFFER, 0);
 	} else {
 		glBindRenderbuffer(GL_RENDERBUFFER, _renderBuffers[0]);
+		#if defined(EMSCRIPTEN)
```
aquadran, line 161: I really don't like ifdefs like this. maybe some function which return GL type

graphics/opengl/shader.cpp
```diff
@@ -33,6 +33,10 @@ namespace OpenGL {
 static const GLchar *readFile(const Common::String &filename) {
 	Common::File file;
 
+	#if defined(EMSCRIPTEN)
```
aquadran, line 36: you can not do that. this is only for development purpose.

aquadran (changes requested):

graphics/opengl/framebuffer.cpp
chkuendig: I changed this and replaced it with a check against GL_VERSION at runtime

graphics/opengl/shader.cpp
chkuendig: I actually forgot about removing this. I just tried to adjust this and I still struggle though about how to handle these. For some reason non of the shaders are found although they are all placed in the subdirectory `shaders/`. I quickly checked the macOS nightly here and that seems to be the same way there (the individual shaders aren't in subdirectories but flat as they are placed in `data/shaders` by `make dist-generic`). Weirdly enough all other assets are loaded just fine, so I'm really at a loss here why this doesn't work.

chkuendig:
I tried to fix most codacy errors. I'm not sure if the remaining once have any relevance.

engines/grim/grim.cpp
chkuendig: _I deleted my previous commit as it wasn't relevant anymore_

I once again looked into this, there's no way to avoid a forced yield here. The logic which is yielding here depends on the frame rate and the limit, which means that as soon as we don't hit the speed limit, the whole browser tab freezes (as the browser doesn't get control over it's main loop back).

graphics/opengl/shader.cpp
chkuendig: _deleted my previous comment as it's not relevant anymore_

This is now fixed. Thank you for pointing out!

engines/grim/grim.cpp
aquadran: ok then

aquadran (approved):

configure
```diff
@@ -2880,6 +2899,24 @@ EOF
...
 		;;
+	emscripten)
+		# mandatory emscripten flags
+		append_var LDFLAGS "-s FULL_ES2=1 -s MAX_WEBGL_VERSION=1 -s ASYNCIFY -s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1"
+		# enable emscripten-ports and set up paths accordingly
+		append_var LDFLAGS "-s USE_SDL=2 -s USE_SDL_MIXER=2  -s USE_OGG=1 -s USE_VORBIS=1  -s USE_LIBJPEG=1  -s USE_FREETYPE=1 -s USE_ZLIB"
+		_sdlpath="$EMSDK/upstream/emscripten/system/bin/"
+		_pngpath="$EMSDK/upstream/emscripten/cache/"
+		_freetypepath="$EMSDK/upstream/emscripten/cache/ports-builds/freetype/"
+		VORBIS_CFLAGS="-I$EMSDK/upstream/emscripten/cache//ports-builds/vorbis/include"
```
aquadran, line 2910: "cache//" unecessary double '/'

chkuendig: thanks, i fixed this.

chkuendig:
I pushed a few more minor improvements and bug fixes today. There's still a few minor bugs and quite a lot potential improvements and I could probably work on this branch for the foreseeable future. 

OTOH I was wondering what would be a good point to get this initial state merged from the projects perspective? I could then focus on this and postpone a few of the less important improvements.

dists/emscripten/build.sh
```diff
@@ -0,0 +1,207 @@
...
+if [[ ! -d "$DIST_FOLDER" ]]; then
+  echo "/dists/emscripten/ not found. Please make sure to run this script from the root of the project - ./dists/emscripten/build.sh "
+  exit 1
+fi
+if [[ "$1" =~ ^(clean)$ ]]; then
+  make clean
+  rm -rf ./dists/emscripten/libs/build
+  rm -rf ./dists/emscripten/libs/*/
+  rm -rf ./dists/emscripten/emsdk*/
+  find . -name "*.o"
```
sev-, line 24: This has a side effect of removing gph/gp2x/mmuhack/mmuhack.o (which reminds me that I need to clean that port up). Why `make distclean` is not working for you?

dists/emscripten/build.sh
```diff
@@ -0,0 +1,207 @@
...
+    wget -nc https://downloads.scummvm.org/frs/extras/Beneath%20a%20Steel%20Sky/BASS-Floppy-1.3.zip
+    unzip -n BASS-Floppy-1.3.zip -d "${ROOT_FOLDER}/build-emscripten/games/bass-floppy/"
+    wget -nc https://downloads.scummvm.org/frs/extras/Drascula_%20The%20Vampire%20Strikes%20Back/drascula-1.0.zip
+    unzip -n drascula-1.0.zip -d "${ROOT_FOLDER}/build-emscripten/games/drascula/"
+    wget -nc https://downloads.scummvm.org/frs/extras/Drascula_%20The%20Vampire%20Strikes%20Back/drascula-audio-mp3-2.0.zip
+    unzip -n drascula-audio-mp3-2.0.zip -d "${ROOT_FOLDER}/build-emscripten/games/drascula/"
+    wget -nc https://downloads.scummvm.org/frs/extras/Flight%20of%20the%20Amazon%20Queen/FOTAQ_Floppy.zip
+    unzip -n FOTAQ_Floppy.zip -d "${ROOT_FOLDER}/build-emscripten/games/fotaq-floppy/"
+    wget -nc https://downloads.scummvm.org/frs/extras/Lure%20of%20the%20Temptress/lure-1.1.zip
+    unzip -n lure-1.1.zip -d "${ROOT_FOLDER}/build-emscripten/games/lure/"
```
sev-, line 181: Would you perhaps also need a pre-made scummvm.ini? Or how are you handling the initial run?

dists/emscripten/build.sh
```diff
@@ -0,0 +1,207 @@
...
+  fi
+
+  cd "${ROOT_FOLDER}//build-emscripten/games/"
```
sev-, line 184: I see a double slash.

sev- (changes requested):
I made a couple more minor comments.

I think we may merge it as soon as those are addressed. Would you like to join the team as an official porter?

A question, is it possible to somehow create a standalone version from it? E.g. basically a release build?

I love this port, and while of course, it is a resource hog, I was able to just open it and play on my non-jailbroken iPad. Yay!

dists/emscripten/build.sh
chkuendig: Yes, a pre-made scummvm.ini would make the most sense. Hows the best way to go about that? I currently inject a default into the browsers local storage if there's nothing there. That's a bit of a hack though if there's a concept of a default ini file.

chkuendig:
> I made a couple more minor comments.

> I think we may merge it as soon as those are addressed. Would you like to join the team as an official porter?

I'll address these as soon as possible. Would love to join. Does this involve anything I should know about that?

> A question, is it possible to somehow create a standalone version from it? E.g. basically a release build?

Yes, my idea was to get this to build some type of bare bone minimal release. I'll see that I'll add this. At the moment this is a bit hobbled together with the bundling of the demos.

neuromancer:
I tested the live demo from http://scummvm.kuendig.io/scummvm.html in the Switch browser and it doesn't work. Nintendo probably disabled some HTML5 APIs to avoid people playing games there :disappointed:

sev-:
Not much is involved in being a porter except continue to show love for your port, address blocker issues (if any) and be around the release time which is coordinated via scummvm-devel mailing list.

Looking forward to the review issues addressed and the port merged.

dists/emscripten/build.sh
mgerhardy: Not sure if this should really be the way to go - maybe we should think about adding a download button for the freeware games or/and demos to the launcher. That way every scummvm installation could benefit from it. And the best thing would be that the emscripten port would automatically get new releases to its launcher without us having to maintain a scummvm.ini or the download urls like here.

it would imo only be a temporary solution.

graphics/opengl/framebuffer.cpp
```diff
@@ -158,12 +158,23 @@ void FrameBuffer::init() {
...
-		glRenderbufferStorage(GL_RENDERBUFFER, useDepthComponent24() ? GL_DEPTH_COMPONENT24 : GL_DEPTH_COMPONENT16, _texWidth, _texHeight);
-		glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, _renderBuffers[0]);
 
-		glBindRenderbuffer(GL_RENDERBUFFER, _renderBuffers[1]);
-		glRenderbufferStorage(GL_RENDERBUFFER, GL_STENCIL_INDEX8, _texWidth, _texHeight);
-		glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_RENDERBUFFER, _renderBuffers[1]);
+		const char *glVersion = (const char *)glGetString(GL_VERSION);
+		if (strstr(glVersion,"WebGL 1.0") != NULL) {
+			// See https://www.khronos.org/registry/webgl/specs/latest/1.0/#FBO_ATTACHMENTS
+			#define GL_DEPTH_STENCIL 0x84F9
```
mgerhardy, line 165: this should be guarded by `#ifnded` imo. other gl headers might define these constants (to the same value) - but you would get compiler warnings about redefining those, no?

sev-:
@chkuendig any update?

chkuendig:
Sorry, I didn't have much free time in the last few weeks to work on this. I still plan though to get this merged asap. But it might take a few more weeks until I'll be able to address everything.

graphics/opengl/framebuffer.cpp
chkuendig: You are right. This is actually only needed because emscripten doesn't have all the WebGL headers (see emscripten-core/emscripten#4832 ). In theory this might be fixed in the future. I added `ifndefs` for this.

dists/emscripten/build.sh
chkuendig: You are right. This was overaggressive. I changed it to only remove a single file that isn't taken care of by `make distclean`.

dists/emscripten/build.sh
chkuendig: The great thing with a web port is that such a download wouldn't be necessary. All files can be downloaded on request (and to a certain point should be - browsers have their own caching system and allow web apps to store limited amounts of data persistently).

The handling of this is definitely something I'd like to improve a lot on.

chkuendig:
Fixed the last minor issues. There's three larger topics I plan to tackle later:
- How do best bundle demos / scope of an official release build
- Better file loading in general (range requests, async loading etc.)
- Dynamic plugin support (so we can have one build supporting all games without blowing up build and load times)

Happy to join the team as a porter.

dists/emscripten/build.sh
chkuendig: I'll mark this as resolved since I decided this is probably better solved later then in this initial PR (since it will also require some thinking about how demos are distributed/packed)

engines/grim/grim.cpp
```diff
@@ -1117,6 +1117,13 @@ void GrimEngine::mainLoop() {
 			g_imuseState = -1;
 		}
 
+		#if defined(__EMSCRIPTEN__)
```
sev-, line 1120: The preprocessor directives must have zero indentation, e.g. start from the column 1 per our code formatting giudelines.

graphics/opengl/framebuffer.cpp
```diff
@@ -158,12 +158,28 @@ void FrameBuffer::init() {
...
 	} else {
 		glBindRenderbuffer(GL_RENDERBUFFER, _renderBuffers[0]);
-		glRenderbufferStorage(GL_RENDERBUFFER, useDepthComponent24() ? GL_DEPTH_COMPONENT24 : GL_DEPTH_COMPONENT16, _texWidth, _texHeight);
-		glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, _renderBuffers[0]);
 
-		glBindRenderbuffer(GL_RENDERBUFFER, _renderBuffers[1]);
-		glRenderbufferStorage(GL_RENDERBUFFER, GL_STENCIL_INDEX8, _texWidth, _texHeight);
-		glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_STENCIL_ATTACHMENT, GL_RENDERBUFFER, _renderBuffers[1]);
+		const char *glVersion = (const char *)glGetString(GL_VERSION);
+		if (strstr(glVersion,"WebGL 1.0") != NULL) {
```
sev-, line 163: Please, add space after the comma

graphics/opengl/framebuffer.cpp
```diff
@@ -158,12 +158,28 @@ void FrameBuffer::init() {
...
+			// See https://www.khronos.org/registry/webgl/specs/latest/1.0/#FBO_ATTACHMENTS
+			// and https://github.com/emscripten-core/emscripten/issues/4832
+			#ifndef GL_DEPTH_STENCIL
```
sev-, line 166: Same thing regarding the preprocessor directives

sev-:
Just minor formatting issues and we're fine to merge.

sev-:
Actually, let me merge it and then fix after by myself.

PR #3047, opened 2021-06-05
COMMON: Simplify default MetaEngine::listSaves

SupSuper:
Currently for an engine to support custom savegame formats it needs to overload both `MetaEngine::listSaves` and `MetaEngine::querySaveMetaInfos`.

This simplifies the process by just having the default `listSaves` call `querySaveMetaInfos`, since the functionality is identical.


sev-:
Yep, makes all sense, thank you!

PR #3048, opened 2021-06-07
DREAMWEB: Text To Speech (TTS) for the game

taylorzhancher:
The TTS feature is implemented to the following:

- GUI (ScummVM main menu > edit game > "Enable Text To Speech")
- Opening bible quote ("And I heard...")
- Opening intro story ("Unseen....")
- Actions in the game ("Open inventory", "Examine book", "Walk to stone floor", etc.)
- Everyday objects in the game and actions upon ("Take watch", "Place gum", "Use wallet", etc.)
- In game options ("Return to map", "Disk options", "Select page", etc.)


engines/dreamweb/detection.cpp
```diff
@@ -25,12 +25,15 @@
...
 #include "common/system.h"
 #include "common/translation.h"
+#include "common/text-to-speech.h"
 
 #include "engines/advancedDetector.h"
 
 #include "dreamweb/detection.h"
 #include "dreamweb/dreamweb.h"
 
+#define GAMEOPTION_TTS_NARRATOR 	GUIO_GAMEOPTIONS1
```
sev-, line 35: Could you please rename it to GAMEOPTION_TTS?

engines/dreamweb/stubs.cpp
```diff
@@ -929,7 +930,9 @@ void DreamWebEngine::processTrigger() {
 	}
 }
 
+const char* lastText = "";
```
sev-, line 933: The code formatting here is incorrect.

engines/dreamweb/stubs.cpp
```diff
@@ -963,6 +966,14 @@ void DreamWebEngine::useTimedText() {
 
 	const uint8 *string = (const uint8 *)_timedTemp._string;
 	printDirect(string, _timedTemp._x, _timedTemp._y, 237, true);
+	const char* theText = (const char*)string;
```
sev-, line 969: The code formatting is incorrect. Please follow our code formatting conventions

engines/dreamweb/stubs.cpp
sev-: Also, why put this on a global level? Please move it to the engine class.

engines/dreamweb/stubs.cpp
```diff
@@ -1261,6 +1277,9 @@ void DreamWebEngine::copyName(uint8 type, uint8 index, uint8 *dst) {
 }
 
 void DreamWebEngine::commandWithOb(uint8 command, uint8 type, uint8 index) {
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+	Common::String theText = "";
```
sev-, line 1281: The initialization is redundant. Common::String is already initialized with an empty string upon creation

engines/dreamweb/stubs.cpp
```diff
@@ -1273,13 +1292,23 @@ void DreamWebEngine::commandWithOb(uint8 command, uint8 type, uint8 index) {
 
 	if (getLanguage() != Common::RU_RUS) {
 		printDirect(string, _textAddressX, _textAddressY, textLen, (bool)(textLen & 1));
+    	if (ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
+			const char *temp = (const char*)string;
```
sev-, line 1296: What is the purpose of this variable, why not add `string` right away to `theText`?

engines/dreamweb/stubs.cpp
```diff
@@ -1273,13 +1292,23 @@ void DreamWebEngine::commandWithOb(uint8 command, uint8 type, uint8 index) {
...
+		}
 
 		copyName(type, index, commandLine);
 
 		uint16 x = _lastXPos;
 		if (command != 0)
 			x += 5;
 		printDirect(commandLine, x, _textAddressY, textLen, (bool)(textLen & 1));
+		if (ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
+			const char *temp = (const char*)commandLine;
```
sev-, line 1307: Same question here. The variable looks redundant to me.

engines/dreamweb/titles.cpp
```diff
@@ -99,6 +101,9 @@ void DreamWebEngine::gettingShot() {
 }
 
 void DreamWebEngine::bibleQuote() {
+	const char *story = "And I heard a great voice out of the temple saying to the seven angels. Go your ways and pour out the vails of the wrath of god upon the earth. Book of revelation. Chapter 16 verse 7.";
```
sev-, line 104: Please split this line, so it doesn't go too far off the screen. you may write like this:

```C++
story = "This is beginning, "
        "this continues "
        "even more."
```

sev- (changes requested):
Added few notes

engines/dreamweb/stubs.cpp
taylorzhancher: Hello sev, thank you for reviewing. For your question: Because it would make the uttered speech sound more natural. For example, "Walk to Jeans", instead of saying: "Walk to", "Jeans", it would say: "Walk-tuh-jeans", so I decided to concatenate before saying.

engines/dreamweb/stubs.cpp
taylorzhancher: My reasoning to putting it at global level, is because the function is called many times to print the same string, so it needs something outside the scope to detect if TTS should speak. Would you suggest me to pass an external variable to the function instead of declaring a global variable?

engines/dreamweb/stubs.cpp
criezy: The question is not about the `theText` variable that you are using to concatenate the verb and the object, but about the `temp` variable. You could write directly `theText += (const char *)string` without using a temporary variable.

And by the way you are missing a space before the `*` in `(const char*)string`.

criezy:
I only had a quick look, but here are some comments:
 - The CD version of the game already has speech for the opening story ("Unseen..."). So you should only use TTS for the floppy version.
   - Alternatively, and it might be better, you could also use TTS for the CD version if speech is disabled (there is an option to select speech, subtitles or both). The advantage of that is that the speech in the CD version is always in English while TTS would be in the language of the game (the subtitles are translated).
 - You are missing some code to set the TextToSpeechManager language to the game language (you can see on https://www.scummvm.org/games/#games-dreamweb that the game is available in French, German, Italian, and Spanish as well, and trying to use TTS with those version currently will not work properly).
   - However the bible quote is always in English I think (it is in English at least in the French version of the game).
   - We could consider asking our community to provide translations for that quote though so that we can speak it in the language of the game.

engines/dreamweb/titles.cpp
```diff
@@ -99,6 +101,11 @@ void DreamWebEngine::gettingShot() {
 }
 
 void DreamWebEngine::bibleQuote() {
+	const char *story = "And I heard a great voice out of the temple saying to the seven angels. "
+						"Go your ways and pour out the vails of the wrath of god upon the earth. "
+						"Book of revelation Chapter 16 verse 7.";
```
criezy, line 106: This is verse 1, not 7.

Also the punctuation is not quite correct. In game the first line ends with a coma: `saying to the seven angels, go your ways`

engines/dreamweb/dreamweb.cpp
```diff
@@ -392,6 +394,27 @@ void DreamWebEngine::processEvents(bool processSoundEvents) {
...
+			break;
+		case Common::DE_DEU:
+			ttsMan->setLanguage("de");
+			break;
+		default:
+			ttsMan->setLanguage("en");
+			break;
+		}
+	}
+
```
criezy, line 417: Does that work when the language is Russian?

Also you can probably use `Common::getLanguageCode` instead to simplify the code:
```c++
ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
```

engines/dreamweb/dreamweb.h
```diff
@@ -161,6 +161,7 @@ class DreamWebEngine : public Engine {
 	Common::RandomSource			_rnd;
 	Common::String _datafilePrefix;
 	Common::String _speechDirName;
+	Common::String lastText;
```
criezy, line 164: Formatting.
https://wiki.scummvm.org/index.php?title=Code_Formatting_Conventions#Naming

> **Class member variables**
> Prefixed with '_' and in camel case (Yo! no underscore separators), starting with lowercase.
> `char *_someVariableName;`

engines/dreamweb/stubs.cpp
```diff
@@ -963,6 +965,14 @@ void DreamWebEngine::useTimedText() {
 
 	const uint8 *string = (const uint8 *)_timedTemp._string;
 	printDirect(string, _timedTemp._x, _timedTemp._y, 237, true);
+	const char *theText = (const char*)string;
```
criezy, line 968: Formatting (space between `char` and `*`).
```c++
const char *theText = (const char *)string;
```

engines/dreamweb/stubs.cpp
```diff
@@ -963,6 +965,14 @@ void DreamWebEngine::useTimedText() {
...
+	if (lastText != theText) {
```
criezy, line 969: Is this check needed when using the `QUEUE_NO_REPEAT` option below? I would expect the NO_REPEAT to skip it if it is the same as the last text.

engines/dreamweb/stubs.cpp
```diff
@@ -1172,9 +1182,14 @@ void DreamWebEngine::commandOnlyCond(uint8 command, uint8 commandType) {
 }
 
 void DreamWebEngine::commandOnly(uint8 command) {
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
 	delTextLine();
 	const uint8 *string = (const uint8 *)_commandText.getString(command);
 	printDirect(string, _textAddressX, _textAddressY, _textLen, (bool)(_textLen & 1));
+	if (ttsMan != nullptr && ConfMan.getBool("tts_enabled")) {
+		ttsMan->say((const char*)string);
```
criezy, line 1190: Formatting (space between `char` and `*`).

engines/dreamweb/talk.cpp
```diff
@@ -85,13 +87,18 @@ uint16 DreamWebEngine::getPersFrame(uint8 index) {
...
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
 	_talkMode = 0;
 
 	const uint8 *str = getPersonText(_character & 0x7F, 0);
 	uint16 y;
 
 	_charShift = 91+91;
 
+	if (ttsMan != nullptr && ConfMan.getBool("tts_enabled") && !hasSpeech()) {
+        ttsMan->say((const char*)str);
```
criezy, line 99: Formatting.

engines/dreamweb/talk.cpp
```diff
@@ -115,6 +122,11 @@ void DreamWebEngine::startTalk() {
 }
 
 const uint8 *DreamWebEngine::getPersonText(uint8 index, uint8 talkPos) {
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+	if (ttsMan != nullptr && ConfMan.getBool("tts_enabled") && !hasSpeech()) {
+        ttsMan->say((const char*)_personText.getString(index*64 + talkPos),
+		Common::TextToSpeechManager::INTERRUPT);
+    }
```
criezy, line 129: The indentation seems off on a couple of lines here. You may be using spaces instead of tabs.

Also maybe it would be worth using a variable for the text.
```c++
const uint8 *text = (const uint8 *)_personText.getString(index*64 + talkPos);
if (ttsMan != nullptr && ConfMan.getBool("tts_enabled") && !hasSpeech())
	ttsMan->say((const char*)text),	Common::TextToSpeechManager::INTERRUPT);
return text;

Comment (criezy) [reply engines/dreamweb/talk.cpp:129]: Also I am wondering if it might be better to use the `TextToSpeechManager` in places where this function is called rather than here. And I see you already do that in `startTalk()`. Doesn't that cause some sentences to be spoken twice?

File: engines/dreamweb/titles.cpp [hunk 16]
```diff
@@ -99,6 +101,42 @@ void DreamWebEngine::gettingShot() {
...
+	const char *esStory = "Oí una gran voz que decía desde el templo a los siete ángeles: "
+						"Id y derramad sobre la tierra las siete copas de la ira de Dios. "
+						"Libro del Apocalipsis, capítulo 16, versículo primero.";
+	const char *deStory = "Dann hörte ich, wie eine laute Stimme aus dem Tempel den sieben Engeln zurief: "
+						"Geht und gießt die sieben Schalen mit dem Zorn Gottes über die Erde. "
+						"Offenbarung des Johannes. Kapitel 16 Vers 1";
+	const char *itStory = "Udii una gran voce dal tempio che diceva ai sette angeli: "
+						"Andate e versate sulla terra le sette coppe dell’ira di Dio. "
+						"libro della rivelazione. capitolo 16 versetto 1";
+
```
criezy, line 119: For the French text you can drop the guillemets. They should not not difference for the TTS engine.

Also to avoid encoding issues it would be better to encode the non-ASCII characters. By default the TextToSpeechManager expects UTF-8 encoding. And for example the UTF-8 encoding for `è` is `0xC3 0xA8`. So using hex notation you can write `"la col\xC3\xA8re de Dieu."`

engines/dreamweb/titles.cpp
```diff
@@ -99,6 +101,42 @@ void DreamWebEngine::gettingShot() {
...
+		theStory = itStory;
+		break;
+	case Common::DE_DEU:
+		theStory = deStory;
+		break;
+	default:
+		theStory = enStory;
+		break;
+	}
+
```
criezy, line 139: Apparently there are Czech and Russian fan translations as well (we can see those in the detection table).
If we want to default to English for those here, we would need to temporarily change the language of the TextToSpeechManager to English.

criezy (changes requested):

engines/dreamweb/stubs.cpp
taylorzhancher: The reason is that for somehow `QUEUE_NO_REPEAT` doesn't take effect, and the sentences keeps repeating itself until the next line arrives. Everything is actually good with only `_lastText != theText`. Do we want to just remove `QUEUE_NO_REPEAT`?

taylorzhancher:
Would it be advisable to add `Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();` to `dreamweb.h`, and remove all ttsMan declarations in other functions?

criezy:
> Would it be advisable to add `Common::TextToSpeechManager *_ttsMan = g_system->getTextToSpeechManager();` to `dreamweb.h`, and remove all ttsMan declarations in other functions?

You could indeed do that, but not exactly that way.
Add `Common::TextToSpeechManager *_ttsMan` in dreamweb.h, and initialize in dreamweb.cpp.
You have two main options here:
 - set it to `nullptr` in the constructor and then to `g_system->getTextToSpeechManager()` in `DreamWebEngine::run()` (where you set its language).
 - or directly initialize it to `g_system->getTextToSpeechManager()` in the constructor.

taylorzhancher:
It seems that not all of the object descriptions (specifically, the text that shows up after we use something) are at the same place. So far I've played inside Eden's bedroom and Sparky's bar, and I found that the button outside & inside the elevator, is different from 2nd page descriptions of all the other objects in the room.

So far there is TTS for the descriptions passed by `void DreamWebEngine::useRoutine()` & `void DreamWebEngine::showFirstUse()`, but there is another function, `void DreamWebEngine::showSecondUse()`, that looks strikingly the same, and I'm not sure if it's best to add TTS right away, to fully test the game, or to skip it.

engines/dreamweb/stubs.cpp
criezy: We discussed for a separate game, and indeed I forgot that `QUEUE_NO_REPEAT` prevent repeating the sentence currently being spoken, but as soon as it has finished, it can speak it again. So the additional check here is indeed needed.

criezy:
> So far there is TTS for the descriptions passed by `void DreamWebEngine::useRoutine()` & `void DreamWebEngine::showFirstUse()`, but there is another function, `void DreamWebEngine::showSecondUse()`, that looks strikingly the same, and I'm not sure if it's best to add TTS right away, to fully test the game, or to skip it.

It seems to me that we will indeed want to have TTS in `showSecondUse()` as well.

engines/dreamweb/stubs.cpp
```diff
@@ -1941,6 +1968,13 @@ void DreamWebEngine::doLook() {
 	dumpTextLine();
 	uint8 index = _roomNum & 31;
 	const uint8 *string = (const uint8 *)_roomDesc.getString(index);
+
+	const char *placeName = (const char *)string;
+	auto colon_pos = strchr(placeName, ':');
```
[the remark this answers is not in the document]
criezy: And since this is only used for the TTS, you should move this inside the `if (_ttsMan != nullptr && ...)` block.

engines/dreamweb/stubs.cpp
```diff
@@ -2337,6 +2371,13 @@ void DreamWebEngine::obsThatDoThings() {
 	}
 }
 
+void DreamWebEngine::textToSpeech(const char *text) {
+	auto colon_pos = strchr(text, ':');
+	Common::String result(text, colon_pos ? colon_pos - text : strlen(text));
+	if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled_objects"))
+		_ttsMan->say(result);
+}
+
```
criezy, line 2380: Same remarks again:
 - Please do not use `auto`
 - Please move this inside the `if` block.

engines/dreamweb/talk.cpp
```diff
@@ -92,6 +94,13 @@ void DreamWebEngine::startTalk() {
 
 	_charShift = 91+91;
 
+	const char *text = (const char *)str;
+	auto colon_pos = strchr(text, ':');
```
criezy, line 98: Another one to change

engines/dreamweb/use.cpp
```diff
@@ -179,6 +185,13 @@ void DreamWebEngine::showFirstUse() {
 	findNextColon(&obText);
 	findNextColon(&obText);
 	useText(obText);
+
+	const char *text = (const char *)obText;
+	auto colon_pos = strchr(text, ':');
```
criezy, line 190: And another one

criezy (changes requested):
There are a number of changes to make. In addition to the comment I added in the code, there are two other main issues to fix:
1. The `DreamWebEngine` class seems to be extremely sensitive to the order in which the member variables are declared. Your commit 0dc806567 introduces a crash for me when starting the game. I have tracked this down to the addition of the `_ttsMan` and `_lastText` variables to the `DreamWebEngine` class. I have no idea why that is, but moving those variables at the end of the class fixes the issue for me. Maybe somebody who is more familiar with the engine than me would make sense of this.
```c++
	uint8 _linePointer;
	uint8 _lineDirection;
	uint8 _lineLength;

	Common::String _lastText;
	Common::TextToSpeechManager *_ttsMan;
```

2. TTS only works for English and for the bible quote. This is because for everything else the encoding of the text is incorrect.

By default the `TextToSpeechManager` assumes UTF-8 encoding when passing a `Common::String` to the `say()` function:
```c++
bool say(const String &str, Action action, CodePage charset = kUtf8) {
	U32String textToSpeak(str, charset);
	return say(textToSpeak, action);
}

bool say(const String &str, CodePage charset = kUtf8) {
	return say(str, INTERRUPT_NO_REPEAT, charset);
}
```

See the default value for the `charset` argument.
This is fine for the bible quote as you used UTF-8 encoding for those.
This is also fine for English as anything compatible with ASCII would work.
However everything else is broken, and you need to pass explicitly the correct encoding when calling `_ttsMan->say()`. For French, and probably Italian, Spanish, and German, the encoding used is the DOS code page 850. So you for example need:
```c++
	if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled_objects"))
		_ttsMan->say((const char*)string, Common::kDos850);
```
However for Russian, and probably Czech the text encoding is different. I don't know what it is though, but somebody familiar with those languages or with the dreamweb engine might have an idea.

I would suggest adding a `Common::CodePage _textEncoding` member variables to the `DreamWebEngine` class that you initialize at the same time as you set the `_ttsMan` language (in `DreamWebEngine::run()`). Then you need to pass it whenever you call `_ttsMan->say()` (except for the bible quote).

taylorzhancher:
@criezy Thanks for the detailed review!! Except for the encoding of Russian & Czech we'll find out later, everything else should have been updated with the 3 commits

engines/dreamweb/newplace.cpp
```diff
@@ -126,7 +126,7 @@ void DreamWebEngine::lookAtPlace() {
 	findNextColon(&string);
 
 	if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled_objects"))
-		_ttsMan->say((const char *)string);
+		_ttsMan->say((const char *)string, Common::TextToSpeechManager::DROP, _textEncoding);
```
criezy, line 129: Is there a reason to change the behaviour here (and in the other places where you made the change) from `INTERRUPT` (the default) to `DROP`?

If not you can simply write `_ttsMan->say((const char *)string, _textEncoding);`

engines/dreamweb/dreamweb.cpp
```diff
@@ -396,7 +396,13 @@ void DreamWebEngine::processEvents(bool processSoundEvents) {
...
 Common::Error DreamWebEngine::run() {
 	if (_ttsMan != nullptr) {
-		_ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
+		Common::String languageString = Common::getLanguageCode(getLanguage());
+		_ttsMan->setLanguage(languageString);
+		if (languageString != "en") {
+			_textEncoding = Common::kDos850;
+		} else {
+			_textEncoding = Common::kUtf8;
+		}
```
criezy, line 405: `kDos850` should work for English as well.
It will definitely not work for Russian though. If I had to take a guess it would be that it might be using `kDos866`. But since this is a fan translation, it may be using something more recent (e.g. `kWindows1251`). You probably want to ask sev. Since he added support for it (in https://github.com/scummvm/scummvm/commit/93dfd39) he might know what encoding is used.

For the checks it would also be better to use the language code rather than string comparisons:
```c+
switch (getLanguage()) {
case Common::RU_RUS:
    _textEncoding = Common::kDos866;
    break;
default:
    _textEncoding = Common::kDos850;
    break;
}
```

criezy (changes requested):
I have added two more comments.
Also you used the wrong prefix (`SCI` instead of `DREAMWEB`) for bf51cce.

engines/dreamweb/dreamweb.cpp
taylorzhancher: After checking, the encoding for Russian is `kDos866`
As for Czech, I see this message popping up when I tried to open the game
```
The game in 'dreamweb-cz' seems to be an unknown game variant.

Please report the following data to the ScummVM team at https://bugs.scummvm.org/ along with the name of the game you tried to add and its version, language, etc.:

Matched game IDs for the dreamweb engine: dreamweb-cd

  {"dreamweb.r00", 0, "3b5c87717fc40cc5a5ae19c155662ee3", 152918},
  {"dreamweb.r02", 0, "28458718167a040d7e988cf7d2298eae", 210466},
```

engines/dreamweb/newplace.cpp
taylorzhancher: I thought misunderstood the default as `DROP`, and resolved to add it to all the places I intended with `_ttsMan->say((const char *)string, _textEncoding);`. It seemed that I need 3 inputs for that.

engines/dreamweb/dreamweb.cpp
criezy: > As for Czech, I see this message popping up when I tried to open the game
> What should I do? Do I go post at https://bugs.scummvm.org/?
> 
> ```
> The game in 'dreamweb-cz' seems to be an unknown game variant.
> 
> Please report the following data to the ScummVM team at https://bugs.scummvm.org/ along with the name of the game you tried to add and its version, language, etc.:
> 
> Matched game IDs for the dreamweb engine: dreamweb-cd
> 
>   {"dreamweb.r00", 0, "3b5c87717fc40cc5a5ae19c155662ee3", 152918},
>   {"dreamweb.r02", 0, "28458718167a040d7e988cf7d2298eae", 210466},
> ```
This matches the detection we already have for the Czech version, but you are probably missing the `dreamweb.exe` file that is also used for the detection.

The Czech version download is not the full game. This is a patch that you apply on top of another version. As indicated in https://bugs.scummvm.org/ticket/7078, you can for example use the CD DOS Version (English US) that you can find on https://www.scummvm.org/games/#games-dreamweb and replace some of the files with those in the Czech downloads.

taylorzhancher: Here's what I get printed when setting the encoding to Windows1250
Looks a lot like what is printed on the screen & Google translate recognizes it
```
Neviden...
Obydli samotnych myslenek formuje nasi budoucnost...  Dreamweb.
..  Dreamweb.
Strazena svymi sluhy.......
Miliardy let stara......
Po staleti byla Dreamweb neotresena..........
Nyni ale chce zlo prevzit vladu.........
"Strazci. Pavucina snu se pomalu borti."         
"Sedm sil zla na Zemi se pomalu spojuji v jednu"   
"Pokud se stanou prilis silnymi, Dreamweb bude znicen"    
"Kdo bude ten vykonavatel ?"
"Kdy se to stane?"
"TICHO! Ti, kdoz byli vyvoleni, prichazeji."  
"Pokud odhali svoji silu, pak se stanou prilis silnymi"
"Bylo seme zasazeno?"
"Ano. Mohutne vzrostlo a ma se cile k svetu." 
"Prisel cas a ja ho probudim."
"Tohle je budoucnost.."
"Vim, co ode mne chcete. Myslim, ze vam nyni rozumim."
"Tech sedm se stalo silnejsimi."
"Strasili mne v mych snech."  
"Vim. Maji moc nad sedmi misty pavuciny snu - Dreamweb."
"Museji byt zniceni?"    
"Ano."
"Citim blizkost vsech sedmi. Ktery z nich ma byt prvni?"     
"Jmenuje se  Crane."
"Je nejslabsi ze vsech sedmi."
"Jak ho mam najit?"
"Je velmi blizko. Opatri si zbran a zacni ho hledat." 
"Cas se nam kvapem krati, bratre...."
```

engines/dreamweb/newplace.cpp
taylorzhancher: I was wrong! They're all good now

engines/dreamweb/titles.cpp
```diff
@@ -99,6 +101,54 @@ void DreamWebEngine::gettingShot() {
 }
 
 void DreamWebEngine::bibleQuote() {
+	const char *enStory = "And I heard a great voice out of the temple saying to the seven angels, "
+						"Go your ways and pour out the vails of the wrath of god upon the earth. "
+						"Book of revelation Chapter 16 verse 1.";
+	const char *frStory = "Puis j'entendis une voix forte qui venait du temple et disait aux sept anges: "
+						" Allez verser sur la terre les sept coupes de la col\xC3\xA8re de Dieu."
```
criezy, line 108: You can remove the space at the start of this  line (since there is a space at the end of the previous one).
And you need to add a space at the end of this line or at the start of the next one. Currently for me the TTS makes no pause between `Dieu` and `L'apocalypse` (which sounds strange) because of the missing space.

criezy (commented):
Almost there! 😃 
You just have one final small change to make for the French bible text to make me happy. It otherwise works very well.

engines/dreamweb/titles.cpp
taylorzhancher: The lastest update applies the principle to other versions of translation as well

criezy (approved):
This looks good now.

The history is a bit messy and will need to be cleaned, but we can maybe just squashed everything in a single commit when merging.

Will wait until master is unfreezes before merging. And this will give time for others to review and request more changes if needed.

whoozle (approved):
Great work, thanks! :)

engines/dreamweb/use.cpp
```diff
@@ -179,6 +185,14 @@ void DreamWebEngine::showFirstUse() {
 	findNextColon(&obText);
 	findNextColon(&obText);
 	useText(obText);
+
+	if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled_objects")) {
+		const char *text = (const char *)obText;
+		const char *colon_pos = strchr(text, ':');
+		Common::String result(text, colon_pos ? colon_pos - text : strlen(text));
+		_ttsMan->say(result, _textEncoding);
+	}
```
bluegr, line 194: You can use `textToSpeech()` here

engines/dreamweb/use.cpp
```diff
@@ -188,6 +202,14 @@ void DreamWebEngine::showSecondUse() {
...
```
bluegr, line 211: You can use `textToSpeech()` here

engines/dreamweb/newplace.cpp
```diff
@@ -153,6 +160,12 @@ void DreamWebEngine::locationPic() {
 
 	const uint8 *string = (const uint8 *)_travelText.getString(_destPos);
 	printDirect(string, 50, 20, 241, 241 & 1);
+
+	const char *placeName = (const char *)string;
+	const char *colon_pos = strchr(placeName, ':');
+	Common::String result(placeName, colon_pos ? colon_pos - placeName : strlen(placeName));
+	if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled_objects"))
+		_ttsMan->say(result, _textEncoding);
```
bluegr, line 168: You can use `textToSpeech()` here

engines/dreamweb/stubs.cpp
```diff
@@ -963,6 +964,14 @@ void DreamWebEngine::useTimedText() {
 
 	const uint8 *string = (const uint8 *)_timedTemp._string;
 	printDirect(string, _timedTemp._x, _timedTemp._y, 237, true);
+	const char *theText = (const char *)string;
+	if (_lastText != theText) {
```
bluegr, line 968: Could `_ttsMan->isSpeaking()` be used instead of checking if the string changed?

bluegr (changes requested):

engines/dreamweb/stubs.cpp
criezy: The short answer is no.

By default `TextToSpeechManager::say()` already does nothing when trying to speak again the same string that it is currently speaking. But the issue here is that we also want to avoid repeating again and again the same string. Without this check, as soon has it has finished speaking the text, it will start speaking it again because `useTimedText()` gets called again and again (multiple times per second). We only want to speak the text when a new text is displayed.

engines/dreamweb/use.cpp
criezy: Indeed.
And I would also suggest renaming `textToSpeech()` into something like `speakObjectName()` to clarify that it cannot be used to speak conversation text.

engines/dreamweb/stubs.cpp
bluegr: Fair enough :) I'm just not a fan of keeping a state of the currently playing sentence

taylorzhancher:
@bluegr Thanks for pointing out. It should be done as said now.
p.s. I named it `speakObject` instead of `speakObjectName` because it's used for descriptive text instead of just titles.

bluegr (approved):
Looks good now, thanks! Well done

criezy:
I will squash and merge this now that the trunk is unfrozen.

PR #3049, opened 2021-06-07
DIRECTOR: Expand and organize detection entries

trembyle:
This adds 623 new game IDs, most of which are available for both Mac and Win.

Apple / Bandai Pippin is now available as a platform. There are some differences to how these games will be handled vs their Mac counterparts.

In an effort to clean up the multi-line detection entries, there are new macros for Pippin and FM Towns and new macros across all platforms to detect two files when necessary. And finally, since we now have many hundreds of games listed, the game descriptors have been moved to `detection_tables.h` to make the actual detection code easier to find in `detection.cpp`.


engines/director/director.cpp
```diff
@@ -79,12 +79,60 @@ DirectorEngine::DirectorEngine(OSystem *syst, const DirectorGameDescription *gam
 
 	SearchMan.addSubDirectoryMatching(_gameDataDir, "data");
 	SearchMan.addSubDirectoryMatching(_gameDataDir, "install");
-	SearchMan.addSubDirectoryMatching(_gameDataDir, "main");		// Meet Mediaband
+	SearchMan.addSubDirectoryMatching(_gameDataDir, "main");							// Meet Mediaband
 	SearchMan.addSubDirectoryMatching(_gameDataDir, "l_zone");
-	SearchMan.addSubDirectoryMatching(_gameDataDir, "win_data", 0, 2);	// L-ZONE
+	SearchMan.addSubDirectoryMatching(_gameDataDir, "win_data", 0, 2);					// L-ZONE
```
sev-, line 84: Ugh, this is now too ugly. Could you instead expose the list from detection.cpp file, then have a smaller list here with dirs that require depth and iterate over it?

trembyle: Done. Now I need to fix the existing subdirectories based on our Discord discussion and my incorrect understanding of the depth parameter.

trembyle: This is now fixed with the latest commit.

engines/director/director.cpp
```diff
@@ -77,14 +78,13 @@ DirectorEngine::DirectorEngine(OSystem *syst, const DirectorGameDescription *gam
 	// Meet Mediaband could have up to 5 levels of directories
 	SearchMan.addDirectory(_gameDataDir.getPath(), _gameDataDir, 0, 5);
 
-	SearchMan.addSubDirectoryMatching(_gameDataDir, "data");
-	SearchMan.addSubDirectoryMatching(_gameDataDir, "install");
-	SearchMan.addSubDirectoryMatching(_gameDataDir, "main");		// Meet Mediaband
-	SearchMan.addSubDirectoryMatching(_gameDataDir, "l_zone");
-	SearchMan.addSubDirectoryMatching(_gameDataDir, "win_data", 0, 2);	// L-ZONE
```
sev-, line 84: Looks like we are losing '2' here. I'll test after the merge and see if the game still works.

engines/director/detection.cpp
```diff
@@ -29,163 +29,93 @@
...
 
-static const char *directoryGlobs[] = {
-	"install",
-	"l_zone",
-	"win_data",	// L-ZONE
-	0
-};
+namespace Director {
+	const char *directoryGlobs[] = {
+		"install",
```
sev-, line 36: According to our code formatting guidelines, `namespace` does not increase the indentation. I'll fix it after the merge

sev-:
Thank you!

PR #3050, opened 2021-06-08
AGS: Add ags_fire and ags_agi plugin stubs (AGS_Fire.dll + ags_agi.dll)

scureuil:
AGS_Fire: Doesn't do anything, but allows games using plasma fire effects to load
and run.

All methods return 0

A sample game using this plugin is Maniac Mansion Mania 46.

AGS_AGI: Allows games who use AGI lowres effects to load and run.

Games using scaling mode 0 will be squashed on the screen, but others
should be playable without the screen effect.

Sample working games: Gladiator Quest, The Secret of Hutton Grammar School

### DIFF
--- a/engines/ags/module.mk
+++ b/engines/ags/module.mk
@@ -287,12 +287,14 @@ MODULE_OBJS = \
 	engine/script/systemimports.o \
 	plugins/agsplugin.o \
 	plugins/plugin_base.o \
+	plugins/ags_agi/ags_agi.o \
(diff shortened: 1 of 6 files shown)


criezy:
Thank you. The code itself looks OK.
However:
 - commit 51d883a and 10a079c should be squashed (let us know if you need help with that).
 - the merge commit should be removed.

scureuil:
Rebased + squashed commits, it should be cleaner now.

PR #3051, opened 2021-06-08
BASE: Added --engine=ID for --list-all-games and --list-games

scureuil:
For the command line, lists games only for the specified engine.
(optional)

Usage:

--list-all-games --engine=sci 
--list-games --engine=ags

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -86,6 +86,8 @@ static const char HELP_STRING[] =
 	"                           Use --path=PATH to specify a directory.\n"
 	"  --game=ID                In combination with --add or --detect only adds or attempts to\n"
 	"                           detect the game with id ID.\n"
+	"  --engine=ID              In combination with --list-games or --list-all-games only lists\n"
+	"                           games for this engine.\n"
 	"  --auto-detect            Display a list of games from current or specified directory\n"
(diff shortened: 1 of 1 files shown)


criezy:
Also I just tried it and it works nicely. I think this is a good addition.

And I forgot to write that the two commits should also be squashed. But Github has an option to do that when merging the pull request, so it's up to you if you want to do it or let us do it.

bluegr:
+1 from me too, nice work, and I agree with the feedback provided by @criezy

scureuil:
Thanks for the feedback. 

I did the modifications: rebasing, fixing alignments, and squashing the commits, so it should be cleaner.

criezy:
It is indeed cleaner. However it looks like you fixed the alignment for only one of the two I commented on (it is now correct in `listAllGames` but the issue is still there in `listGames`).

scureuil:
Done. (I remember doing this line too, but I think I wrongly reversed a local commit.) 

Changed (engineID=="") to engineID.empty() too.

As it's not using a default engineID like "all" (who could conflict with a future engine, like an engine named "global" would conflict with --list-debugflags), empty() is more true to the meaning of "no engine specified".

criezy (approved):
Thank you. It looks all good now.

PR #3052, opened 2021-06-09
DOCS: Fix Category:Supported_Games links

tipabu:
Previously, these would display like

>  A list of required data files for each game is found on the *wiki entry \<https://wiki.scummvm.org/index.php?title=Category:Supported_Games>* for that game.

(i.e. in italics, with literal angle brackets and the full URL displayed inline.)


criezy:
Thanks!

PR #3053, opened 2021-06-09
GRAPHICS: MACGUI: MacText selection fix

ysj1173886760:
bring the behaviour of mactext selection the same as original director
introduce _selEnd and _selStart to director. the initial selection of mactext will be determined by those two variable.
fix drawing selection and calculating row and col when we have text alignment
clear some TODO in mactext.

### DIFF
--- a/engines/director/lingo/lingo.h
+++ b/engines/director/lingo/lingo.h
@@ -443,9 +443,6 @@ public:
 
 	Common::String _floatPrecisionFormat;
 
-	int _selStart;
-	int _selEnd;
-
 	bool _hadError;
 
 	bool _inCond;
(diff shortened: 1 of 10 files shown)


engines/director/lingo/lingo.h
```diff
@@ -443,6 +443,9 @@ class Lingo {
 
 	Common::String _floatPrecisionFormat;
 
+	int _selStart;
+	int _selEnd;
```
djsrv, line 447: These properties should probably belong to Movie, not Lingo, to avoid conflicts with movies in other windows.

ysj1173886760: thanks

sev-:
Thank you!

PR #3054, opened 2021-06-09
DIRECTOR: add custom target detection to director

ysj1173886760:
trello card which describe the task https://trello.com/c/jpwJCfZA/332-add-target-name-to-custom-detection
detect custom targets like Director Dictionary.
we need to add some info to those custom targets to get the correct description and version number.

### DIFF
--- a/engines/director/detection.cpp
+++ b/engines/director/detection.cpp
@@ -328,7 +328,7 @@ public:
 		return debugFlagList;
 	}
 
-	ADDetectedGame fallbackDetect(const FileMap &allFiles, const Common::FSList &fslist) const override;
+	ADDetectedGame fallbackDetect(const FileMap &allFiles, const Common::FSList &fslist, ADDetectedGameExtraInfo **extra) const override;
 };
 
 static Director::DirectorGameDescription s_fallbackDesc = {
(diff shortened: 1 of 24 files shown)


engines/director/detection.cpp
```diff
@@ -32,90 +32,100 @@
...
-	"english",
-	"deutsch",
-	"francais",						// Rename from Français
-	"italiano",						// Virtual Tourism Paris
-	"please copy to hd. G3",		// Rename from HDにｺﾋﾟｰして下さい。G3
-	"_files_",						// The Gate
-	"popup",						// Pop Up Computer
-	"alpha",						// Interactive Alphabet
-	"technik",						// Glasklar Technology Interactive
+	const char *directoryGlobs[] = {
```
sev-, line 35: This is incorrect. We do not add indent with `namespace`

engines/director/detection.cpp
```diff
@@ -202,6 +219,34 @@ ADDetectedGame DirectorMetaEngineDetection::fallbackDetect(const FileMap &allFil
...
+			if (!f.open(*file))
+				continue;
+			Common::String targetID = f.readString('\n');
+			Common::String gameName = f.readString('\n');
+			Common::String platform = f.readString('\n');
+			Common::String version = f.readString('\n');
+
+			Common::strlcpy(s_fallbackFileNameBuffer, fileName.c_str(), sizeof(s_fallbackFileNameBuffer) - 1);
+			desc->desc.filesDescriptions[0].fileName = s_fallbackFileNameBuffer;
+			desc->version = atoi(version.c_str());
```
sev-, line 236: Do you also capture this at runtime? Seems that this code was not added.

sev-:
Thank you!

PR #3055, opened 2021-06-09
GUI: Do not show tooltip on hovered focused EditText widget

antoniou79:
Prevents showing a tooltip for a widget is it is an editable field (editText) and has both mouse hovering above it and the current focus.

This is so that typing text is not interrupted / slowed down by a periodical display of the tooltip, if the mouse is hovering over the same text field that the user is editing.

### DIFF
--- a/gui/widget.h
+++ b/gui/widget.h
@@ -153,6 +153,8 @@ public:
 	void lostFocus() { _hasFocus = false; lostFocusWidget(); }
 	virtual bool wantsFocus() { return false; }
 
+	uint32 getType() const {return _type;}
+
 	void setFlags(int flags);
 	void clearFlags(int flags);
 	int getFlags() const		{ return _flags; }
--- a/gui/widget.h
+++ b/gui/widget.h
@@ -153,7 +153,7 @@ public:
 	void lostFocus() { _hasFocus = false; lostFocusWidget(); }
 	virtual bool wantsFocus() { return false; }
 
-	uint32 getType() const {return _type;}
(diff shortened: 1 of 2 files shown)


criezy (approved):
This seems to be a good idea to me and the implementations seems OK.
Maybe the two commits should be squashed, but this can be done when the PR gets merged.

gui/widget.h
```diff
@@ -153,6 +153,8 @@ class Widget : public GuiObject {
 	void lostFocus() { _hasFocus = false; lostFocusWidget(); }
 	virtual bool wantsFocus() { return false; }
 
+	uint32 getType() const {return _type;}
```
bluegr, line 156:
```suggestion
	uint32 getType() const { return _type; }
```

bluegr:
Nice work overall! Good idea and straightforward implementation

antoniou79:
I realise now that this is an issue even if the hovered widget is different than the focused EditText widget, since the manager will keep displaying periodically the tooltip for that other widget while the user is typing in the EditText. 

**If it's ok, I'll close this PR and open a new one.** 

The if clause will be changed to
`Widget *focusedWdg = activeDialog->getFocusWidget();
if (focusedWdg && focusedWdg->getType() == kEditTextWidget) { ... }`
So as to catch the case where there is a focused EditTextWidget at the same time when mouse cursor is hovering over the same or another widget which has a tooltip. 

And also the comments will be updated to reflect this new behavior.

PR #3057, opened 2021-06-10
GUI: Do not show tooltip for hovered widget if there is a focused EditText one

antoniou79:
This PR is a more proper fix to the issue that my previous (now closed PR) was addressing: 
(https://github.com/scummvm/scummvm/pull/3055)

Prevents showing a tooltip for the hovered-widget, if an editText field has the current focus, and hence the user will be typing in it. 

This is so that typing text is not interrupted / slowed down by a periodical display of the tooltip for the hovered-widget. This applies whether the mouse cursor is hovering over the same widget that the user is editing or another tooltip-enabled widget (only while an editText widget has the focus). 

### DIFF
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -440,13 +440,29 @@ void GuiManager::runLoop() {
 				++it;
 		}
 
-		if (_lastMousePosition.time + kTooltipDelay < _system->getMillis(true)) {
+		// Handle tooltip for the widget under the mouse cursor
+		uint32 systemMillisNowForTooltipCheck = _system->getMillis(true);
+		if (_lastMousePosition.time + kTooltipDelay < systemMillisNowForTooltipCheck) {
 			Widget *wdg = activeDialog->findWidget(_lastMousePosition.x, _lastMousePosition.y);
 			if (wdg && wdg->hasTooltip() && !(wdg->getFlags() & WIDGET_PRESSED)) {
-				Tooltip *tooltip = new Tooltip();
-				tooltip->setup(activeDialog, wdg, _lastMousePosition.x, _lastMousePosition.y);
-				tooltip->runModal();
-				delete tooltip;
+				Widget *focusedWdg = activeDialog->getFocusWidget();
+				if (focusedWdg && focusedWdg->getType() == kEditTextWidget) {
+					// If an EditText widget has the focus -- and presumably the user is typing in it,
+					// then don't display the tooltip for any widget (wdg)
+					// currently under the mouse cursor so as not to interfere with text typing.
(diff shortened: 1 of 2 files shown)


antoniou79:
@bluegr , @criezy I'm tagging you on this one, since I closed the previous PR about this issue. 
Do you think we can merge this one?

criezy:
I am not sure about the change. The usual behaviour I think is to allow tooltips on editText widgets when they have the focus, but when the tooltip has been hidden (for example if you start to type) it should not appear again until you move the mouse.

This might be more complicated to implement though, so maybe not having any tooltip when an editText widget has the focus is an acceptable solution. Once issue though is that is seems impossible to remove the focus from an editText widget once it has the focus, unless you give it to another editText widget (clicking on any other types of widget does not seem to remove the focus from the editText widget). And that means that once you have given the focus to the editText you can no longer have a tooltip (until you close and reopen the dialog).

antoniou79:
> I am not sure about the change. The usual behaviour I think is to allow tooltips on editText widgets when they have the focus, but when the tooltip has been hidden (for example if you start to type) it should not appear again until you move the mouse.
> 
> This might be more complicated to implement though, so maybe not having any tooltip when an editText widget has the focus is an acceptable solution. Once issue though is that is seems impossible to remove the focus from an editText widget once it has the focus, unless you give it to another editText widget (clicking on any other types of widget does not seem to remove the focus from the editText widget). And that means that once you have given the focus to the editText you can no longer have a tooltip (until you close and reopen the dialog).

You might be right on that. I use Esc key to lose the focus (physical keyboard or Android on-screen keyboard), so I know that works. 

But I didn't test trying to lose the focus using just the mouse (and not having a virtual/on-screen keyboard available to click Esc from there).

I think it's worth trying for the more standard behavior you mention, although it may take me some time to implement -- not much free time at the moment, but I don't think the implementation itself would be that much complicated.

criezy:
> I use Esc key to lose the focus

You are right. That works. I didn't expect it to be the way to do it, so that was not one of the things I tried.

> Maybe the mouse movement would restore tooltips being shown in other widgets but not on the focused one (if the tooltip was already shown or user started typing).

I think it would not be problematic if the tooltip appear also on the focused widget after moving the mouse cursor. As long as it does not appear again while you don't move the mouse cursor, this should make typing text easier.

bluegr:
I like the implementation, and it looks to be better than the current behavior. I agree that it would be ideal to make the tooltip appear on other widgets (e.g. when hovering the mouse over them) and not show it on focused text widgets

antoniou79:
Hopefully, for the last time, I am closing this PR for this issue, and opening a new one that is, again hopefully, more suitable.

PR #3059, opened 2021-06-10
STARTREK: Fix format-security warnings

xavery:
Fix `format-security` warnings emitted by gcc due to a missing format string literal in two calls to `debug`.

These warnings make it impossible to build the `scummvm-git` package in Arch Linux AUR since a few days ago, when `Werror=format-security` was added to the default `CFLAGS`.


digitall:
@xavery : Looks good. Pretty self-explanatory change. Will merge. Thanks for your contribution.

PR #3060, opened 2021-06-10
Cleanup DebugManager

sluicebox:
I had trouble reading DebugManager to figure out recent regressions so I cleaned it up.

- Renamed methods and members to be consistent and clearer
- Updated doxygen comments to be in sync with method behaviors
- Global debug channels are now automatically added by DebugManager. The code was going to a lot of effort to try to achieve that but still not always succeeding. Now it's explicit and simpler. Callers don't have to always call two methods to add channels.
- Removed unnecessary calls to remove debug channels from Engine destructors. As of PR #3012, Engines are no longer responsible for adding their channels, so this code should have been removed from the destructors. It's redundant; `runGame()` already handles this.

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -966,7 +966,7 @@ static void printDebugFlags(const DebugChannelDef *debugChannels) {
 /** List debug flags*/
 static void listDebugFlags(const Common::String &engineID) {
 	if (engineID == "global")
-		printDebugFlags(globalDebugChannels);
+		printDebugFlags(gDebugChannels);
 	else {
 		const PluginList &plugins = EngineMan.getPlugins();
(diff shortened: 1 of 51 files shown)


ysj1173886760:
Thank you, that reverting part is correct. I didn't managed to get the total sight of this, so i made some mistakes.
Thank you for amending and optimizing those codes. I've learnt a lot though that. Much Appreciated.

sluicebox:
Thanks for reviewing this!

The confusion in this class predates your changes -- channels vs flags, adding vs registering vs setting, etc. Now that all that code is in one place, those existing inconsistencies became clearer. I also like that you've got the channels managed outside of the Engine constructors and destructors.

Now that the expert has taken a look, I'm comfortable merging. =)

PR #3061, opened 2021-06-11
DIRECTOR: offset fix

ysj1173886760:
director didn't handle the size of mactext correctly. e.g. it will accumulate the border size.
thus, we will calculate the pure text size of mactext first.
also fixed some offsets problem in mactext. e.g. cursor offsets in mactext


sev-:
Thanks! merging.

PR #3067, opened 2021-06-15
COMMON: moved AchievementDescriptionList into Common namespace

mgerhardy:
getAchievementsInfo is also moved into the MetaEngine to reduce code duplication


bluegr:
Nice work! +1 from me

lolbot-iichan:
Thanks, merging!

PR #3072, opened 2021-06-18
COMMON: Introduce BiDi base paragraph directions

lolbot-iichan:
Some BiDI games require providing exact base paragraph direction for a string.

For example, some Wintermute games are storing BiDi strings as strings that are only displayed correctly with in LTR.
There are 2 possible techniques to fix this - manually adding RTL/LTR Unicode Marks, or using different base paragraph directions.
It seems that current font renderer actually displays "invisible" Unicode Marks, so I selected another way.

This PR adds optional `BiDiParagraph dir = BIDI_PAR_ON` parameter to BiDi constructors and helpers.
Original behavious is provided by default values, desired WME behaviour is provided by other values.

P.S. Results of some experiments with Persian game that is NOT using RTL paragraph direction:
изображение

### DIFF
--- a/common/unicode-bidi.cpp
+++ b/common/unicode-bidi.cpp
@@ -132,10 +132,13 @@ String convertBiDiStringByLines(const String &input, const Common::CodePage page
 }
 
 String convertBiDiString(const String &input, const Common::Language lang, BiDiParagraph dir) {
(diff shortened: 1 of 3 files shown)


lephilousophe:
Looks good to me

PR #3073, opened 2021-06-18
GUI: Fix tooltip behavior to be less disruptive

antoniou79:
This PR is a more proper fix to the issue that my previous (now closed) PRs were addressing:
(https://github.com/scummvm/scummvm/pull/3055, https://github.com/scummvm/scummvm/pull/3057)

The tooltip will show only if mouse cursor was moved but not on a hovered focused editText field

Changes and implications in this commit:
- Fix _lastMousePosition coordinates and time being changed upon giving focus to the tooltip. They now will not (specifically for "focusing" on a Tooltip).
- Check for mouse cursor movement first in the decision for showing the tooltip, then check if sufficient time for mouse resting position has passed (kTooltipDelay).
- Prevent showing a tooltip for a widget if it is an editable field (editText) and has both mouse hovering above it and the current focus. This is so that typing text is not interrupted / slowed down by a periodical display of the tooltip, if the mouse is hovering over the same text field that the user is editing.
- If mouse cursor is moved but lands on the same widget as the one that had its tooltip shown last, then show the tooltip but after a different (larger) delay kTooltipSameWidgetDelay.
- Still shows tooltip for other widgets, including editText ones, if the mouse is hovered over them and they are not the current focused editText widget.

Minor caveats with this PR:
1. Since we now check based on _lastTooltipShown x,y coordinates, then if mouse is moved but cursor lands back at the exact same coordinates, it will not be detected as "movement" as far as the tooltip check is concerned. However, practically this barely has any significance. The user can slightly move the mouse cursor and the tooltip will be displayed.
2. We don't show the tooltip for the *focused* editText. The tooltip can be shown if the cursor is hovering on top of the unfocused editText widget, but not after user clicks on it, setting the focus to it (and presumably starts writing in it) -- not until that widget is unfocused again.

### DIFF
--- a/gui/Tooltip.cpp
+++ b/gui/Tooltip.cpp
@@ -42,6 +42,8 @@ void Tooltip::setup(Dialog *parent, Widget *widget, int x, int y) {
 
 	_parent = parent;
 
+	setMouseUpdatedOnFocus(false);
+
 	_maxWidth = g_gui.xmlEval()->getVar("Globals.Tooltip.MaxWidth", 100);
 	_xdelta = g_gui.xmlEval()->getVar("Globals.Tooltip.XDelta", 0);
 	_ydelta = g_gui.xmlEval()->getVar("Globals.Tooltip.YDelta", 0);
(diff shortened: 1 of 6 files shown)


sev-:
Thank you! and I like your code documentation also.

PR #3078, opened 2021-06-20
COMMON: Fix BASESTRING::contains()

lolbot-iichan:
There is a mistake in current `contains()` implementation - it does not backtrack after partial match, which causing wrong results.
For example, with current implementation word "Russian" contains "ssian", but does not contain "si".

This PR fixes this, however it proposes the simpliest algorithm possible.
Do we need something more clever and efficient here?

### DIFF
--- a/common/base-str.cpp
+++ b/common/base-str.cpp
@@ -422,6 +422,7 @@ TEMPLATE bool BASESTRING::contains(const BaseString &otherString) const {
 			if (sizeMatch == otherString.size())
 				return true;
 		} else {
+			itr2 -= sizeMatch;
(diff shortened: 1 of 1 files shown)


lephilousophe:
Your patch indeed fix a bug in contains.

For the optimization part, as a start we could specialize contains method for chars by using strstr.
No idea for U32String.

lephilousophe:
LGTM!

lolbot-iichan:
@lephilousophe 
Thanks for the review, merging.

mgerhardy:
it would be nice to get `test_contains` extended by this particular error case.

PR #3079, opened 2021-06-20
GRAPHICS: Fix crash when resizing the window to low width

orgads:


lephilousophe:
The root cause of the bug was fixed earlier.
It may be safer to have this patch as well but it could hide other bugs.

bluegr:
This fix actually makes the while loops that fill the screen in that part of the code much more  readable and robust anyway. I don't really see how this change will mask other bugs.

So I'm going to merge this, as checking for < makes much more sense for such loops and I can't think of any ill effects from this change.

Thanks again for your work!

PR #3080, opened 2021-06-20
GUI: If GraphicsWidget is hidden do not proceed with setGfx()

antoniou79:
Removes interfering invisible thumbnail with list item selection in saveload-dialog (OpenGL, large scale)

This issue is more apparent at least on the Android port, which uses OpenGL rendering and may default to large scale mode. It is not Android specific, though. It can be observed on the Windows builds too.

In large scale mode, ScummVM's Load/Save dialogue screen displays only a list of the saved game entries. No thumbnail is shown when selecting an entry -- the space at the right for thumbnail, playtime and save datetime is removed/hidden. However, in that Large scale mode, before this fix, when selecting an item entry from the list, the SaveLoadChooserSimple::updateSelection() method would be called, and eventually _gfxWidget->setGfx(thumb, true), which essentially would create a region overlayed on top of part of the list (aligned top left) for the (invisible) thumbnail. This new region would then "block" the list from receiving mouse down events when clicking on that part of the list, since the dialogue would detect the _gfxWidget instead and send the "mouse down" to that. The result would be clicks and double clicks not being registered for the list. 

Here's a recent forum post that reports the same issue -- for libretro core build, but essentially it is the same bug and the user provides a video showcasing the issue. 
https://forums.scummvm.org/viewtopic.php?f=2&t=16310

I'm issuing this as a PR because I'm unsure if this fix is ok, since it affects all GraphicsWidget objects, not just the one for thumbnail in saveload-dialog. **Edit**: I've also added the same check in other setGfx() methods in widget.cpp that seemed to be relevant / prone to similar issues with rogue invisible widgets being overlayed over other widgets and "stealing" clicks.

Note, that an alternative fix, which also works, would be to only modify gui/saveload-dialog.cpp and method SaveLoadChooserSimple::updateSelection(), to check for _gfxWidget isVisible status before calling: _gfxWidget->setGfx(thumb, true);

### DIFF
--- a/gui/widget.cpp
+++ b/gui/widget.cpp
@@ -630,7 +630,7 @@ void PicButtonWidget::setGfx(int w, int h, int r, int g, int b, int statenum) {
 	_gfx[statenum].free();
 
 	if (!isVisible() || !_boss->isVisible())
-	return;
+		return;
 
 	if (w == -1)
 		w = _w;
@@ -919,7 +919,7 @@ void GraphicsWidget::setGfx(int w, int h, int r, int g, int b) {
 	_gfx.free();
 
 	if (!isVisible() || !_boss->isVisible())
-	return;
+		return;
 
 	if (w == -1)
 		w = _w;
--- a/gui/widget.cpp
+++ b/gui/widget.cpp
@@ -880,6 +880,9 @@ void GraphicsWidget::setGfx(const Graphics::ManagedSurface *gfx, bool scale) {
 		return;
 	}
 
+	if (!isVisible())
+		return;
+
 	float sf = g_gui.getScaleFactor();
 	if (scale && sf != 1.0) {
 		_w = gfx->w * sf;
--- a/gui/widget.cpp
+++ b/gui/widget.cpp
@@ -598,6 +598,9 @@ void PicButtonWidget::setGfx(const Graphics::ManagedSurface *gfx, int statenum, 
 		return;
 	}
 
+	if (!isVisible() || !_boss->isVisible())
+		return;
+
 	float sf = g_gui.getScaleFactor();
 	if (scale && sf != 1.0) {
 		Graphics::Surface *tmp2 = gfx->rawSurface().scale(gfx->w * sf, gfx->h * sf, false);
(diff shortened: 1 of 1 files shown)


gui/widget.cpp
```diff
@@ -624,14 +627,18 @@ void PicButtonWidget::setGfxFromTheme(const char *name, int statenum, bool scale
 }
 
 void PicButtonWidget::setGfx(int w, int h, int r, int g, int b, int statenum) {
+	_gfx[statenum].free();
+
+	if (!isVisible() || !_boss->isVisible())
+	return;
```
bluegr, line 633: Spacing

gui/widget.cpp
```diff
@@ -906,14 +916,18 @@ void GraphicsWidget::setGfx(const Graphics::Surface *gfx, bool scale) {
 }
 
 void GraphicsWidget::setGfx(int w, int h, int r, int g, int b) {
+	_gfx.free();
+
+	if (!isVisible() || !_boss->isVisible())
+	return;
```
bluegr, line 922: Spacing

bluegr:
The rationale behind this change is solid. I really like the simplicity of the changes. It definitely looks like a step towards the right direction! +1 from me

sev-:
I wonder. If such widget is sitting on a ContainerWidget and is scrolled behind the visible area. Does it mean that this patch will break it?

Particularly, with the new icon launcher, it will always be the case.

If it is truly so, then perhaps we would need to introduce another flag to the widgets like "onScreen()"

antoniou79:
Since this bug is highly disruptive in the Save/load game dialogue on low resolution / large scale mode, I've pushed a commit to master which only addresses it in that screen -- checking for _gfxWidget.isVisible() in SaveLoadChooserSimple::updateSelection() before doing the setGfx for the thumbnail. 
https://github.com/scummvm/scummvm/commit/e128d69748f087584476b4ac0ae6c45b42d9fdca

If we decide to merge this PR, that small fix will be redundant and can be removed. Same goes for if a more efficient fix is suggested/implemented which addresses the cause of the issue at its root. Otherwise, it can remain safely, since I don't think it conflicts with anything in that screen as far as I have tested.

bluegr:
I propose to merge this - by the looks of it, it shouldn't introduce any regressions, but we'll only be 100% sure if it gets merged and thoroughly tested

sev-:
I asked the student, @av-dx to test this for us in his branch

av-dx:
I have tested the patch on my branch and the Icon Launcher is not affected.

sev-:
Great, thank you, merging.

PR #3081, opened 2021-06-21
RFC: Allowing music players to use the Mixer mutex

eriktorbjorn:
This experiment was prompted by https://bugs.scummvm.org/ticket/12649 but also https://bugs.scummvm.org/ticket/12617

It seems that it's quite easy to accidentally cause deadlocks in various music players, and one of the causes seem to be that the mixer and music player tend to have one mutex each. So would it be safer if the music player had access to the mixer's mutex? This is a proof-of-concept branch where I've added a mutex() method to the Mixer to get its mutex, and then use that in the Player_AD music player.

Is this a good idea or not? It would - at least in theory - apply to a lot of other music players, of course.

### DIFF
--- a/audio/mixer.h
+++ b/audio/mixer.h
@@ -23,6 +23,7 @@
 #ifndef AUDIO_MIXER_H
 #define AUDIO_MIXER_H
 
+#include "common/mutex.h"
 #include "common/types.h"
 #include "common/noncopyable.h"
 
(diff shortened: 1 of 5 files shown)


athrxx:
It is a simple and very beneficial fix. Several sound drivers that currently need very careful deconstruction so as to avoid deadlocks could be simplified after  this gets merged. 

This fix won't help with sound drivers that run on other threads, like e.g. SDLTimer thread. But all the emulator based drivers usually run on the mixer (SDLAudioP1) thread.

bluegr:
This is definitely a fix towards the right direction and it removes the need for separate mutexes, which can be very hard to handle and debug. +1 from me

sev-:
Yes, please merge it

PR #3083, opened 2021-06-22
DIRECTOR: speed up transitions for debugFast

rvanlaar:
The number of steps a transition has is calculated based on the
chunksize. A larger chunksize means quicker execution with a downside
of not having pixel perfect frames.

This is only enabled when the debugflag is used.


rvanlaar:
Closing PR. It was for running this code in buildbot. Committed into master.

PR #3084, opened 2021-06-23
WINTERMUTE: Update detection and achievements, use non-Steam achivements for non-Steam Carol Reed 10&11

lolbot-iichan:


bluegr:
Nice work, thanks! Merging

PR #3085, opened 2021-06-23
EVENTRECORDER: fixes for keymapper custom engine events

mgerhardy:
The record file format is changed. But the version wasn't bumped because the old state wasn't working anyway.

This added the custom engine events from the key mapper to the serialization of the event recorder. 

With this change I can record a play in twine.

During a playback there is currently no interaction possible as all event sources except the recorder are disabled. I don't consider this a problem atm - as I see the main benefit on running the recorded plays on the buildbot - but if someone needs to have this back, I would consider doing this in a future PR.

# Follow up tasks:

* create recording for game demos
* integrate into buildbot

https://www.youtube.com/watch?v=SOD97joxFds

### DIFF
--- a/backends/cloud/cloudicon.cpp
+++ b/backends/cloud/cloudicon.cpp
@@ -75,12 +75,12 @@ CloudIcon::Type CloudIcon::getShownType() const {
 }
 
 bool CloudIcon::needsUpdate() const {
-	uint32 delaySinceLastUpdate = g_system->getMillis() - _lastUpdateTime;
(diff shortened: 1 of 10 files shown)


sev-:
Thank you for looking into this.

Answering your question. No, it was not intended. The custom events were introduced later, and since the event recorder is not enabled by default, nobody noticed that it is not processing them.

You do not really need to bump the version, because the feature was not used, so it is safe to keep it a v1 for now.

sev-:
Thanks! Merging.

PR #3086, opened 2021-06-23
TINSEL: Add detection for Discworld 1 PSX Ger/Jap PSX versions

eddebaby:
Follow up on: https://bugs.scummvm.org/ticket/11368

Adds detection for German Playstation version.
Also corrects detection of Japanese Playstation version.

getSampleIndex and getSampleFile in tinsel.cpp now fallback to ENGLISH files if they fail to find (e.g.) GERMAN.IDX.

This was an easy way to add support for the PSX German CD, without breaking the German DOS CD, Japanese PSX CD etc.

The German PSX version should be as compatible as the English PSX version.

The Japanese PSX version subtitles are not supported, and cause ScummVM to crash when you first try to hover over an object. However, you can view the intro cutscene, and hear the audio.

__
_Edit: I only marked as draft as this is my first submission, the code should be ready as is._

### DIFF
--- a/engines/tinsel/tinsel.cpp
+++ b/engines/tinsel/tinsel.cpp
@@ -1355,8 +1355,7 @@ const char *TinselEngine::getSampleIndex(LANGUAGE lang) {
 
 	} else {
 		cd = 0;
-		TinselFile f;
-		if (!f.open(_sampleFiles[lang][cd]))
+		if (!Common::File::exists(_sampleFiles[lang][cd]))
 			lang = TXT_ENGLISH; // fallback to ENGLISH.IDX/.SMP if <LANG>.IDX is not found
 	}
 
@@ -1377,8 +1376,7 @@ const char *TinselEngine::getSampleFile(LANGUAGE lang) {
 
 	} else {
 		cd = 0;
-		TinselFile f;
-		if (!f.open(_sampleFiles[lang][cd]))
+		if (!Common::File::exists(_sampleFiles[lang][cd]))
 			lang = TXT_ENGLISH; // fallback to ENGLISH.IDX/.SMP if <LANG>.IDX is not found
 	}
 
--- a/engines/tinsel/tinsel.cpp
+++ b/engines/tinsel/tinsel.cpp
@@ -1355,8 +1355,9 @@ const char *TinselEngine::getSampleIndex(LANGUAGE lang) {
 
 	} else {
 		cd = 0;
-		if (lang != TXT_JAPANESE)
-			lang = TXT_ENGLISH;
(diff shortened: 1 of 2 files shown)


eddebaby:
Files need to be extracted/split correctly from DISCWLD.LFD/DISCWLD.LFI. 

[discworld_quickbms_script.zip](https://github.com/scummvm/scummvm/files/6704568/discworld_quickbms_script.zip)

The attached QuickBMS script can be used on the German and Japanese CDs to extract the data, or could be used to inform handling the LFD/LFI files directly in ScummVM. 

_The QuickBMS script mention in the above request ticket isn't quite right for the Discworld 1 files, so I revised it. (Instead of an OFFSET, that field is an INDEX of 2KB data chucks, so apart from the first file, the existing `discworld2.bms` script produces junk data with the German and Japanese Discworld 1 PSX discs.)_

engines/tinsel/tinsel.cpp
```diff
@@ -1355,8 +1355,9 @@ const char *TinselEngine::getSampleIndex(LANGUAGE lang) {
 
 	} else {
 		cd = 0;
-		if (lang != TXT_JAPANESE)
-			lang = TXT_ENGLISH;
+		TinselFile f;
+		if (!f.open(_sampleFiles[lang][cd]))
```
bluegr, line 1359:
```suggestion
		if (!Common::File::exists(_sampleFiles[lang][cd]))
```

engines/tinsel/tinsel.cpp
```diff
@@ -1376,8 +1377,9 @@ const char *TinselEngine::getSampleFile(LANGUAGE lang) {
...
```
bluegr, line 1381:
```suggestion
		if (!Common::File::exists(_sampleFiles[lang][cd]))
```

bluegr:
Overall, the changes to the detection entries look OK. Regarding the checks for file existence, you can use the static method Common::File::exists(), instead of needlessly initializing TinselFile

sev-:
Please, rename the commit in accordance with our guidelines (https://wiki.scummvm.org/index.php?title=Commit_Guidelines) e.g. to something like "TINSEL: Add detection for Discworld 1 PSX Ger/Jap PSX versions"

engines/tinsel/tinsel.cpp
eddebaby: Great, thank you.

eddebaby:
Thanks @sev and @bluegr - everything should be good now.

bluegr:
Thanks for your work! Merging

PR #3087, opened 2021-06-24
DIRECTOR: Apartment/User Event fix

ysj1173886760:
re-organize the logic of mactext and macbutton, eliminate many magic offsets
fix the logic of builtin function within.
copying the value instead of pass the pointer of sprites to channel. which will prevent modifing sprites directly.
keep the widgets when we don't have to replace them
keep fixing mactext


sev-:
Yay, merging

PR #3088, opened 2021-06-24
PARALLACTION: make IFCHAR command skip other characters' commands

peres:
The command version of IFCHAR was skipping the commands for the current character, instead of those for the other characters. After this change it behaves like the location version of IFCHAR.

The IFCHAR command appears for the first time in BRA part 3.


digitall:
@peres: Nice to hear from you again :) These changes look pretty simple and self-contained to Parallaction BRA engine, so merging as-is. Thanks.

PR #3089, opened 2021-06-25
SCUMM: Partial support for hi-res font in Indiana Jones and the Last Crusade

eriktorbjorn:
This adds support for using the high-resolution font in Indiana Jones and the Last Crusade. (It also hard-codes a high-resolution cursor, since there doesn't seem to be any cursor resource for one.)

- It is not pixel-perfect. That was never my goal.
- It does not implement the Macintosh verb and inventory GUI. Boy howdy was that never my goal.

But I think it's still an improvement.

Along the way, I found and think I've fixed two glitches specific to the Macintosh version:

- If you try to give an object to someone (e.g. a guard in Castle Brunwald), but change your mind by pressing "Never mind" and then try again to give him something, only the first inventory object is visible. My theory is that the Mac version treats the entire inventory box as a single verb. ScummVM works around that by manually creating the inventory verbs, but that means that if they are ever temporarily disabled the scripts will probably try to re-enable them as a single verb. So I've worked around that by having o5_verbOps() call inventoryScriptIndy3Mac() when verb 101 is turned on. (There seemed to be no need for a similar hack in the SO_VERB_OFF case.)
- When giving the painting to a Castle Brunwald guard, some verbs were left behind. It seems the Mac version expects the verb script to be called twice, because there's one case for removing the verbs and another case for printing the reaction to the offer. In the DOS 256-color version (which is the only other version I own) this is handled by one single call to the verb script. I've worked around that by adding a special case for verb script 18 when clicking on an inventory object.

As with the Mac version of Loom, my goal has been to keep the game working properly even without the high-resolution font. You will get a warning that you're missing the Mac executable, but I saw no reason to actively break the game. (Not that it worked quite properly to begin with.)

Unlike Loom, Monkey Island 1, Monkey Island 2 and Fate of Atlantis, there are no digitized instruments. Music probably sounds a lot different than the original, but I don't have the skill to fix that.

Unfortunately, i still ran across two further glitches while playing through the game:

- When leaving the zeppelin by plane, there is a brief scene where some leftover (but disabled) verbs were visible. This could be similar to the glitch described above. Unfortunately, it's on the other side of the most annoying maze in the entire game. (I swear, it's like a Delphine game with better path finding!)
- The end credits are broken. The old text is not removed between the text screens. In the DOS version, the script calls setState(946,0) between text screens, which apparently trigger a redraw. I counted 14 such calls. But in the Mac version, there's only one and I have a nagging suspicion it was left there by accident. Apparently it's supposed to clear the text some other way?


eriktorbjorn:
Screenshots demonstrating the two remaining glitches:

indy3-glitch1
indy3-glitch2

eriktorbjorn:
Sceenshots demonstrating unimplemented features.

indy3-conversation
indy3-inventory

eriktorbjorn:
Screenshots demonstrating things that look just fine to me:

indy3-train
indy3-copyprotection
indy3-fight
indy3-multicolor
indy3-captioned
indy3-manual
indy3-grail1
indy3-grail2
indy3-grail3
indy3-gameover

eriktorbjorn:
I believe both glitches have been worked around now. So that means I'll have to play through both Indy 3 and Loom again to make reasonably sure that everything's still fine.
indy3-glitch1-fixed
indy3-glitch2-fixed
*You! You did this to me!*

eriktorbjorn:
While playing, I did come across some dodgy path finding. The one that's easy to reproduce is when Donovan's guys picks up Indy near the start of the game. Indy walks up and down for a bit before going to the car. This only happens if you leave your office through the window (i.e. the simple solution to the students puzzle).

Perhaps related to this, on my first playthrough I came across a guard in the zeppelin maze who was walking on air straight towards me. As if that maze wasn't annoying enough anyway! It didn't happen the second time though, or at least I never saw it.

I have no intention to look into these glitches myself. I have no idea how path finding works. At least the first glitch happens in the 256 colour DOS version as well.

eriktorbjorn:
This pull request should fix https://bugs.scummvm.org/ticket/4493 though only if you're using the high-resolution font. I just can't be bothered to figure out how to clear the text in low-resolution mode.

eriktorbjorn:
The early path finding glitch has been filed as https://bugs.scummvm.org/ticket/12666

eriktorbjorn:
Mac Loom seems to still work without regressions. Onwards to Indy 3...

eriktorbjorn:
I've played through Indy 3. I may have seen a zeppelin guard walking on air again, but that's unrelated to this pull request and besides athrxx has been doing some walk fixes on on the master branch. So maybe whatever I saw has been fixed there already.

PR #3099, opened 2021-06-26
GLK: Fix build for Symbian.

fedor4ever:
This allow use remove() function.

### DIFF
--- a/engines/glk/agt/filename.cpp
+++ b/engines/glk/agt/filename.cpp
@@ -20,6 +20,13 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


lephilousophe:
According to @ccawley2011 this function and some others aren't used and could be removed.
This PR would be useless then.

fedor4ever:
What functions should be removed?

sev-:
Obviously, the one, using `remove()`

sev-:
Closing this as incorrect. The `remove()` usage should be removed from the GLK engine

PR #3104, opened 2021-06-26
SCI: Fix Symbian compilation.

fedor4ever:
Before build symbian build system process every source file with prehistoric preprocessor.
I was unable to replace it with a modern version from gcc, since there are special specific commands. There is no description of the commands.

### DIFF
--- a/engines/sci/resource/resource_patcher.cpp
+++ b/engines/sci/resource/resource_patcher.cpp
@@ -39,7 +39,7 @@ namespace Sci {
 #else
 #define _PACKINT32(n) (((uint32)n) >> 24 & 0xFF), (((uint32)n) >> 16 & 0xFF), (((uint32)n) >> 8 & 0xFF), (((uint32)n) & 0xFF)
 #endif
-#define _BYTEOP(op, numBytes, ...) op, _PACKINT32(numBytes), __VA_ARGS__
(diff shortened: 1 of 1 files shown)


fedor4ever:
On hold till next commit

sev-:
You are using an C++ standard extension, I am not sure about compatibility across all platforms:

https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

Maybe some  porters could check it on their platforms.

fedor4ever:
I found that solution for prehistoric preprocessors. 'ARGS...' looks like variadic macro but it is one macro token.

fedor4ever:
Similar commit - https://github.com/scummvm/scummvm/pull/3111

fedor4ever:
Here that blog post - http://blog.shiv.me/2011/11/08/badly-punctuated-parameter-list/ .

sev-:
I merging this one PR and see if it breaks on the buildbot. If everything is fine, other two fill follow.

sev-:
No, just as expected this lead to bunch of warnings:

```C++
engines/sci/resource/resource_patcher.cpp:42:66: warning: token pasting of ',' and __VA_ARGS__ is a GNU extension [-Wgnu-zero-variadic-macro-arguments]
#define _BYTEOP(op, numBytes, ARGS...) op, _PACKINT32(numBytes), ##ARGS
                                                                 ^
```

Reverting

PR #3105, opened 2021-06-26
WINTERMUTE: Init screenmode once

lolbot-iichan:
I have noticed, that for 2D games graphics initialization was done twice - at WintermuteEngine::run() and BaseRenderOSystem::initRenderer(int width, int height, bool windowed).

The first call looks unnecessary, let's remove it.
Probably, this would fix https://bugs.scummvm.org/ticket/11160, since second call seems to be ineffective when this first one is called with different screen dimentions.

### DIFF
--- a/engines/wintermute/base/gfx/osystem/base_render_osystem.cpp
+++ b/engines/wintermute/base/gfx/osystem/base_render_osystem.cpp
@@ -34,6 +34,7 @@
 #include "engines/wintermute/math/math_util.h"
 #include "engines/wintermute/base/base_game.h"
 #include "engines/wintermute/base/base_sprite.h"
+#include "engines/util.h"
 #include "common/system.h"
 #include "common/queue.h"
 #include "common/config-manager.h"
@@ -123,11 +124,9 @@ bool BaseRenderOSystem::initRenderer(int width, int height, bool windowed) {
 	_windowed = !ConfMan.getBool("fullscreen");
 
 	Graphics::PixelFormat format(4, 8, 8, 8, 8, 24, 16, 8, 0);
-	g_system->beginGFXTransaction();
-		g_system->initSize(_width, _height, &format);
-	OSystem::TransactionError gfxError = g_system->endGFXTransaction();
+	initGraphics(_width, _height, &format);
 
-	if (gfxError != OSystem::kTransactionSuccess) {
+	if (g_system->getScreenFormat() != format) {
 		warning("Couldn't setup GFX-backend for %dx%dx%d", _width, _height, format.bytesPerPixel * 8);
 		return STATUS_FAILED;
 	}
--- a/engines/wintermute/wintermute.cpp
+++ b/engines/wintermute/wintermute.cpp
@@ -31,7 +31,6 @@
 #include "common/tokenizer.h"
 #include "common/translation.h"
 
-#include "engines/util.h"
 #include "engines/wintermute/ad/ad_game.h"
 #include "engines/wintermute/wintermute.h"
 #include "engines/wintermute/debugger.h"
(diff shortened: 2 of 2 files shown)


criezy:
> since second call seems to be ineffective when this first one is called with different screen dimentions.

That is on purpose. The idea here is that if a game is going to use multiple resolutions, then the engine should call `initGraphicsModes()` with the list of different resolutions it will use before the first call to `initGraphics()`. Then the backend can take some decision on which resolution to use for the screen or window.

But in the case here it seems that the games are only using one resolution and that the issue was that there is a first call to `initGraphics()` with the wrong size before having a second call with the correct size. In such a case removing the first call is indeed the right thing to do.

lolbot-iichan:
Thanks for explanations, @criezy, I now understand why the second call was ineffective.
In this case, accepting this PR, let's see if it fixes the reported issue.

raziel-:
I just checked with latest master and i still get the wme games opened in a small space on the bottom left.
Moreover GMM is not displaying, if i trigger it from this small game space, it simply locks up scummvm.

lolbot-iichan:
> I just checked with latest master and i still get the wme games opened in a small space on the bottom left.
> Moreover GMM is not displaying, if i trigger it from this small game space, it simply locks up scummvm.
Are you using Software or OpenGL render?
Working in the small bottom left space and not displaying GMM is something that I experience on Windows 10, if OpenGL (with or without shaders) render is selected, see https://bugs.scummvm.org/query?status=!closed&keywords=~Wintermute3D
Could you please make sure that you are using Software render?

raziel-:
Yes, it happens with opengl, but not with software renderer

lolbot-iichan:
> Yes, it happens with opengl, but not with software renderer
> Edit: and I can confirm every item you linked to.
> Happens here on amigaos4 aswell.

@raziel-, could you please confirm that I read you correctly?

1. All those 5 bugs happens on AmigaOS4 with OpenGL:
https://bugs.scummvm.org/ticket/11160 WME: Engine doesn't use correct screenmode 
https://bugs.scummvm.org/ticket/12630 WME3D: rendering does not support screen mode change
https://bugs.scummvm.org/ticket/12631 WME3D: rendering does not support overlays 
https://bugs.scummvm.org/ticket/12664 WME3D: rendering does not support screen size change 
https://bugs.scummvm.org/ticket/12636 WME3D: Alpha Polaris - crash when cycling through available shadow types
?

2. All those 4 bugs don't happen on AmigaOS4 with Software render:
https://bugs.scummvm.org/ticket/11160 WME: Engine doesn't use correct screenmode 
https://bugs.scummvm.org/ticket/12630 WME3D: rendering does not support screen mode change
https://bugs.scummvm.org/ticket/12631 WME3D: rendering does not support overlays 
https://bugs.scummvm.org/ticket/12664 WME3D: rendering does not support screen size change 
(there is currently no way to run Alpha Polaris with Software render)
?

If those are correct, I would rename those bugs from "WME3D: <title>" to "WME: OpenGL: <title>" to distinguish them issues specific to 3D games only.

raziel-:
@lolbot-iichan 

I will check the sw renderer when im back home, at least for opengl it seems to be affected here aswell, yes.

Alpha polaris i have to take back.
I initially thought the crash im I'm getting with this game is related to your report, but instead I'm running out of nemory during the engines loading of the games assets.

raziel-:
Those 5 bugs happens on AmigaOS4 with OpenGL (tested with Mental Repairs Inc)

https://bugs.scummvm.org/ticket/11160 WME: Engine doesn't use correct screenmode
Game Renderer: OpenGL - Happens
Game Renderer: Software - Does not happen


https://bugs.scummvm.org/ticket/12630 WME3D: rendering does not support screen mode change
Game Renderer: OpenGL - Happens
See screenshot on bug item, this is what it looks like for me after the switch to window mode.

Game Renderer: Software - Can't test, due to an
ERROR: 3D software renderered is not supported yet!
(There is also a typo in the error message - "renderered" should be "renderer" or "rendering", i suppose)

I also get these warnings
WARNING: BaseRenderOpenGL3D::onWindowChange not yet implemented!
WARNING: BaseRenderOpenGL3D::enableShadows not implemented yet!
and
WARNING: BaseRenderOpenGL3D::onWindowChange not yet implemented!


https://bugs.scummvm.org/ticket/12631 WME3D: rendering does not support overlays
Game Renderer: OpenGL - Happens
This is a bug that happens for me on every WME game, not only 3D.
As soon as the little game window in the big screen is available, i cannot bring the menu up.
But here the whole game screen becomes black, nothing to be seen anymore.

Although, if i press ALT+ENTER i get the GMM menu, but in a window, i can do everything there as a workaround.
And furthermore, if i now click on Resume i get the same window with a white rectangle
at the top right as added to #12630

I also get this
WARNING: BaseRenderOpenGL3D::indicatorFlip not yet implemented!
after the game's main menu loops.

And i always have the game's menu sound playing, so i suspect the game is also running, but invisible/not drawn.


https://bugs.scummvm.org/ticket/12664 WME3D: rendering does not support screen size change
Yes, happens just like described.


https://bugs.scummvm.org/ticket/12636 WME3D: Alpha Polaris - crash when cycling through available shadow types
Unfortunately i cannot test this game as i'm running out of memory while the game's assets are being loaded.

PR #3109, opened 2021-06-27
PARALLACTION: implement SWAP command for BRA

peres:
The SWAP command switches control from the playing character to another one onscreen (the 'current' character becomes an NPC).

Limitations:
- the mouse cursor is not switched, as the engine has no such capability yet
- the playing character's follower is not switched, as I'd need to reach a place where the feature is used first


bluegr:
Nice to see you again @peres!
Great work, merging

PR #3110, opened 2021-06-27
COMMON: Fix inclusion cstddef.

fedor4ever:
That file included for nullptr_t use. If header not aviable that break build. I think it good use cstddef here.

### DIFF
--- a/common/ptr.h
+++ b/common/ptr.h
@@ -27,10 +27,6 @@
 #include "common/noncopyable.h"
 #include "common/safe-bool.h"
 #include "common/types.h"
-#ifdef USE_CXX11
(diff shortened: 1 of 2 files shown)


bluegr:
That's very strange. That header should only be included when USE_CXX11 is defined. Do you have a specific case where this define is not set and compilation is broken?

fedor4ever:
Ags engine required C++11 for build. But Symbian port lacks standard headers.

sev-:
This is a very inappropriate change as it breaks most of the ports.

You flipped the #define semantics and now including for non-C++11 compilers.

If you need this include on your platform, there is portdefs.h of your port.

PR #3111, opened 2021-06-27
AGS: Fix compilation error.

fedor4ever:
That happens due initialization shared pointers with 0.

### DIFF
--- a/engines/ags/console.cpp
+++ b/engines/ags/console.cpp
@@ -30,7 +30,7 @@
 
 namespace AGS {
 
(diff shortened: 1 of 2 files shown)


fedor4ever:
Shared pointer already initialized - https://stackoverflow.com/questions/57310392/will-a-default-constructed-empty-shared-ptr-automatically-be-initialized-to-nu

sev-:
This makes no sense to me. You say that you have an old compiler, and still referring to the C++17 standard.

Leaving variables uninitalized is not acceptable practice.

PR #3113, opened 2021-06-27
PARALLACTION: implement GIVE command for BRA 

peres:
This PR implements the whole GIVE command, which is used to transfer inventory items among the 3 BRA playable characters.

I haven't verified it in practice, but this change should make PART 3 completable.


sev-:
You could commit directly.

PR #3114, opened 2021-06-27
ACHIEVEMENTS: Fix compilation when translations are disabled

dwatteau:
`common/achievements.cpp` would fail to compile without this, in a `--disable-translation` build.


lotharsm:
Thanks, merging!

PR #3115, opened 2021-06-27
IMAGE: Use 0-bit alpha for non-transparent PNGs

lolbot-iichan:
This PR reverts part of 9f98cddf8dc015e0cdc6b57e303b693bba3cc1fc commit - the byte order part of the commit is kept as is, but I restored the code that checked if alpha-channel is present and providing `isAlpha ? 8 : 0` bits for pixel format.

This fixes the root cause of https://bugs.scummvm.org/ticket/11568 issue, because Wintermute engine decides to use color key for PNG images by checking `getSurface()->format.aBits() == 0`.

### DIFF
--- a/image/png.cpp
+++ b/image/png.cpp
@@ -63,11 +63,11 @@ void PNGDecoder::destroy() {
 	_palette = NULL;
 }
 
-Graphics::PixelFormat PNGDecoder::getByteOrderRgbaPixelFormat() const {
+Graphics::PixelFormat PNGDecoder::getByteOrderRgbaPixelFormat(bool isAlpha) const {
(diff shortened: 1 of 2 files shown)


lolbot-iichan:
**UPD**: This does not fix https://bugs.scummvm.org/ticket/11568 completely.
It looks like Rhiannon game sometimes use 24bpp PNGs and sometime use 32bpp PNGs.
And WME 1.x applies colorkey to both, while WME Lite applies colorkey to 24bpp PNGs only.

Since there are very few WME Lite games and none of them is known to have colorkey issues, I guess this PR can be rejected in case the risks of breaking something are believed to be high.

sev-:
Thanks!

PR #3116, opened 2021-06-28
DIRECTOR: Add new detection entries

trembyle:
Adds 92 new gameIDs and 205 new md5 entries.

See individual commits for more detail on the contents of these entries.

I made a small change to detection in CINE engine. The `os-demo-amiga-gb` entry used the filename `demo`, which was generic enough that it was constantly being matched against Mac demos in Director engine. I adjusted this so that the filename is in line with the OS DOS demo (with a different md5).


sev-:
Thank you! merging

PR #3119, opened 2021-06-29
DIRECTOR: Apartment/Puppets fix

ysj1173886760:
bring a better way to handling sprites
handling macwidget events when we are calling alert in director
add partial color support for mactext
implement matte for quick draw shape
fix offsets for constrain box
support auto expand for mactext
more precise definition for highlight sprites
implement immediateSprite
keep fixing mactext and director rendering


sev-:
All is fine, thank you!

PR #3120, opened 2021-06-29
SAGA2: New engine

sev-:
This is a WIP engine for the game _Faery Tale Adventure II: Halls of the Dead_.

I would like to continue work on the engine in-tree with the student. The game is currently playable, but the work is still obviously active.

The things which are not yet there:

* We haven't fully got rid of the original custom data structures such as DList
* Naming of many many things are not according to our naming standards (maybe some bored soul could help us with this)
* The work on the sound subsystem just has started
* No saves
* Occasional crashes, some things do work weirdly
* Tons of stubs
* Still number of globals, though we're like 200 down

I removed the early history of the engine, till the point when we started to have things compilable. It is for going further away from the original sources from one hand, and for not spoiling people's bisect from the other.

I hope, there will be no major objections to merging this, as I would like to benefit from our infrastructures like buildbot and Coverity.


bluegr:
Nice work! Good approach to get the engine working. +1 from me. Keep it up! :)

sev-:
Okay, merged as discussed

PR #3122, opened 2021-06-30
WIN32: Make UNICODE the default build configuration

sluicebox:
This PR makes UNICODE the default for Windows builds.

In PR #3003 I added the ability to configure Windows builds with UNICODE enabled. This allows ScummVM to be compatible with all file paths containing Unicode characters. My plan was to let it sit in master for a while so we could test it, and then make it the default.

I am now accelerating the timeline because of what my testing found.

I thought that the only obstacle would be existing paths in scummvm.ini. I figured game paths would exist with extended ASCII characters that were a result of CP_ACP conversions that happened to work. I figured that upgrading to UNICODE builds in the next release would break those paths. I coded up a config migration scheme that i thought was clever: upon loading a scummvm.ini, inspect the path values and if any of them were potentially unicode, test if they didn't exist on the file system, and then try converting them from native windows codepage to UTF8 and if that converted path existed on then go ahead and upgrade it to UTF8.

But what I found was that I couldn't test this code. I couldn't come up with a path that had Unicode characters that ANSI ScummVM builds didn't mangle. The UI layer always destroyed them. I tried a lot of characters! That suggests that there aren't scummvm.ini Windows files in the wild with Unicode characters in their paths -- the UI wouldn't have allowed that to work. I'm now under the impression that paths with any characters >= 0x80 simply didn't work in Windows builds at all, which means there's no need for a complex migration scheme that no one benefits from. (That's good news! No need for a migration scheme that runs on every startup if there's nothing to migrate!)

*But* the problem  for me is that I'm an American who never types 0x80 or above except to make jokes. I don't have VMs in other code pages. I PR so that those of you in far away places with interesting character sets can chime in. Merging into master to see if there are any problems might be the most effective way to do that, but that's not something I'm going to do all by myself. But I still think we'll find out that this doesn't break anything, because "interesting" paths wouldn't have worked in Windows builds in the first place.


bluegr:
Nice!
I haven't actually used paths with non-English characters up to now (my games are in a folder called "Games"). I could test this in a week or so, unless someone else beats me to it

lotharsm:
Sorry I come up with the following issue so late in the progress, @sluicebox.

Unfortunately, `--enable-windows-unicode` breaks building with plugins enabled:
```
    C++      backends/plugins/win32/win32-provider.o
In file included from backends/plugins/win32/win32-provider.cpp:27:
./backends/platform/sdl/win32/win32_wrapper.h:29:1: error: 'HRESULT' does not name a type
   29 | HRESULT SHGetFolderPathFunc(HWND hwnd, int csidl, HANDLE hToken, DWORD dwFlags, LPTSTR pszPath);
      | ^~~~~~~
./backends/platform/sdl/win32/win32_wrapper.h:86:1: error: 'TCHAR' does not name a type
   86 | TCHAR *stringToTchar(const Common::String& s);
      | ^~~~~
./backends/platform/sdl/win32/win32_wrapper.h:96:36: error: 'TCHAR' does not name a type
   96 | Common::String tcharToString(const TCHAR *s);
      |                                    ^~~~~
backends/plugins/win32/win32-provider.cpp: In member function 'virtual bool Win32Plugin::loadPlugin()':
backends/plugins/win32/win32-provider.cpp:55:29: error: 'stringToTchar' is not a member of 'Win32'
   55 |   TCHAR *tFilename = Win32::stringToTchar(_filename);
      |                             ^~~~~~~~~~~~~
make: *** [Makefile.common:161: backends/plugins/win32/win32-provider.o] Error 1
```

In theory, this could be fixed by adding `windows.h` to the includes in `backends/platform/sdl/win32/win32_wrapper.h`, but I'm unsure if this is the right place to fix this issue.

sluicebox:
@lotharsm Thank you for so much testing that! You're not late to the party at all, my goal was to get `--enable-windows-unicode` in master without enabling it because I figured there would probably still be minor problems that we (you!) would have an easier time testing.

For that plugin code, I fixed the obvious "this would never have compiled with UNICODE" bugs but I didn't even know how to build it in the first place (obviously neither did the author of that existing UNICODE code!) or if it really was ever possible in win32 yet.

I'm not terribly surprised it only passed my eyeball-compiler =)

Fixed in master in 006b91f45bc23ce891a62383e1750d9dd8c737fa and I rebased this branch with it. I can now build with --enable-plugins with and without unicode. Let me know how if works for you!

lotharsm:
Thanks for fixing this :-)

I can confirm that everything works now as expected. No objections agains merging this from my end.

sluicebox:
Great, @bluegr if you can test whenever you get a chance and don't have problems, I'll be comfortable merging.

SupSuper:
> I thought that the only obstacle would be existing paths in scummvm.ini. I figured game paths would exist with extended ASCII characters that were a result of CP_ACP conversions that happened to work. I figured that upgrading to UNICODE builds in the next release would break those paths. I coded up a config migration scheme that i thought was clever: upon loading a scummvm.ini, inspect the path values and if any of them were potentially unicode, test if they didn't exist on the file system, and then try converting them from native windows codepage to UTF8 and if that converted path existed on then go ahead and upgrade it to UTF8.
> 
> But what I found was that I couldn't test this code. I couldn't come up with a path that had Unicode characters that ANSI ScummVM builds didn't mangle. The UI layer always destroyed them. I tried a lot of characters! That suggests that there aren't scummvm.ini Windows files in the wild with Unicode characters in their paths -- the UI wouldn't have allowed that to work. I'm now under the impression that paths with any characters >= 0x80 simply didn't work in Windows builds at all, which means there's no need for a complex migration scheme that no one benefits from. (That's good news! No need for a migration scheme that runs on every startup if there's nothing to migrate!)

Wrong. :)

image

The ANSI Windows codepage is defined by your language. So on your *American* Windows, you will only be able to type English codepage characters. On a *Portuguese* Windows, you will be able to type Portuguese codepage characters. Etc. It wouldn't work if I tried a path with Russian characters, for example. But I'm guessing most ScummVM users have been using paths in their own language fine up to this point.

That said, a breaking change isn't the worst thing in the world, so long as it's documented. This release will already be changing a lot of configs, so this is another for the pile. If ScummVM automatically greys out the now "broken" entries and users are instructed to manually re-add them (or convert their INI), there should be no nasty side-effects.

Maybe in the future we should version-tag our INI to make these kind of migrations easier.

sluicebox:
@SupSuper thank you! That's the kind of path I was hoping to flush out with this. I'm also inclined to just let this be a breaking change since the alternatives that I see are complex, have to run every time, and are inherently difficult to test. (But I still think it can be done if necessary)

Can you walk me through how one of those characters in your example survives all the conversions between the Windows API, the ScummVM UI, and back? There's a piece I'm missing here, because your explanation restates my understanding but I still don't understand how the ScummVM UI with U32Strings is able to make sense out of Win API CP_ACP results in a char*, regardless of the machine's language. Is the UI/U32String layer treating (converting?) strings differently depending on the user's language or something? (That would be something I've missed) What byte is coming back from the Win API /CP_ACP on your machine for Á? 

To put my question another way: if the UI is receiving Portuguese characters from a WindowsFilesystemNode with a char* that's not UTF8, how do those characters end up getting rendered correctly in the UI with U32Strings?

I can take Windows out of the equation on my machine and just make a scummvm.ini with a path c:\test\Á (where Á is 0xC1) and the ScummVM UI won't render that character.

SupSuper:
To clarify, the screenshot is from 2.2.0 (before the Unicode refactor). Characters survived fine because there were no conversions in the ScummVM UI, only the Windows API. And the UI codepages happened to match, if the language was different, the path would be too (but again, just a visual bug):

image

Anyways I tested this Unicode version with some funky paths and didn't see any issues.

sluicebox:
@SupSuper Ah ha! Thank you, that's the piece I was missing. I had been doing all of my migration tests against the code in master, where the UI mangles things as I described. I didn't realize that this UI behavior was recent and came after 2.2.0. In 2.2.0, c:\test\Á works at all layers even for me. Phew, things make sense again!

That makes me even happier about enabling UNICODE at the Windows layer, since without that the GUI Unicode refactor has made it so some directories that fully worked in 2.2.0 can no longer be added / browsed through the GUI.

Now that I know that I can go back to testing my scummvm.ini ascii => unicode migration scheme and use 2.2.0 for testing. If it works and seems worth it, I'll PR it separately.

trembyle:
I tested some Cyrillic paths on MinGW. I saw no issues adding games from the cmd line or the GUI. It seems to work exactly as expected.

bluegr:
I also tested this with Greek paths on Windows, and everything works as expected. The paths are saved with the correct Unicode encoding in scummvm.ini. Nice work overall!
+1 from me

Since this works in both Visual Studio and MinGW builds, it should be good to be merged. The change and the potential issues that could arise that have been described by @SupSuper should be mentioned in the NEWS and perhaps the documentation.

sluicebox:
Thank you everyone for all the great testing and helping getting this all sorted out. I'm comfortable merging now.

In the near future I will PR my scummvm.ini ansi => unicode auto-migration scheme. If it works and we like it, that should mean there's nothing for us to tell users to do, besides announcing in NEWS that all characters in Windows paths are now supported. If the migrator doesn't work well, or we don't think it's worth it (very possible), then as @bluegr said we can also note that some paths might have to be re-added in the GUI upon upgrade.

lephilousophe:
This broke tests: https://buildbot.scummvm.org/#/builders/22/builds/1584

lotharsm:
Thanks, @lephilousophe! This should be fixed via 1aa1832b59105b94c107c7fe13f9bfead5d86c3b.

lephilousophe:
Yep! Thanks!

PR #3124, opened 2021-07-01
GRAPHICS: 32-bit surfaces, points, and rects

dreammaster:
The genesis of this pull request was a need for one AGS game, Guard Duty, to create a 100,000 x 15 surface, which exceeds the maximum size of both the Surface class as well as the Rect and Point classes. To fix this, I've increased the dimensions of the Point, Rect, Surface, ManagedSurface, and TransparentSurface to 32-bits. This fixes the problem for AGS.

As far as I'm aware, there should be no issues for other engines, since all usage of 16-bit fields will work equally well for 32-bit fields.  Finally, one thing I'll mention is that although the original surface w, h, and pitch were uint16, but I've changed them to be int32 rather than uint32. In most cases when surface w & h are referred to, we're using int fields anyway, such as checking x < 0 || x >= surface->w. So with the expanded field size, it makes things more convenient to have them signed.

### DIFF
--- a/backends/graphics/opengl/texture.cpp
+++ b/backends/graphics/opengl/texture.cpp
@@ -176,8 +176,8 @@ Surface::Surface()
 
 void Surface::copyRectToTexture(uint x, uint y, uint w, uint h, const void *srcPtr, uint srcPitch) {
 	Graphics::Surface *dstSurf = getSurface();
-	assert(x + w <= dstSurf->w);
-	assert(y + h <= dstSurf->h);
+	assert(x + w <= (uint)dstSurf->w);
+	assert(y + h <= (uint)dstSurf->h);
 
 	// *sigh* Common::Rect::extend behaves unexpected whenever one of the two
 	// parameters is an empty rect. Thus, we check whether the current dirty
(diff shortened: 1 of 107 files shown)


dreammaster:
Any comments on this? If not, I'll proceed with merging it in tomorrow

bluegr:
Overall, this is fine. There haven't been any cases where we would need such surfaces up to now, as most surfaces reach up to 800x600 in size in most cases. 

Would it make sense to have an overridden class just for AGS with 32-bit integers? If not, we can go forward with an implementation like this, but the chances that such large values will actually be used are pretty slim

dreammaster:
Ironically, I just found out that the 100,000 may be worked around for AGS, since it wasn't needed for cursors like I thought, but for a buffer building up text lines.. the 100,000 is just the "maximum possible width" passed from the script. There likely won't be any issue with doing a MIN(maxWidth, 16000) within AGS to restrain it. That should handle even multiple lines worth of text.

So I'm no longer planning to do an integration whilst I try doing a work-around, but I still have two questions remaining:
1) Is it worth changing Surface w/h/pitch to int16 rather than uint16. As I mention in the pull request header, having signed w/h would make common boundary checks simpler by not generating signed vs unsigned comparison warnings, and there similarly shouldn't be any engines wanting > 16000 width or height
2) Is it worthwhile cherry-picking in the Common::Rect/Point change still? While no other engines would likely benefit from extra-large surfaces, there are still definitely cases where a 32-bit rect class could prove useful. Particularly for engines doing delta calculations using fixed precision math. I'm still semi-sure that I had issues with Rect in one previous engine simply because it was only 16-bit

sev-:
#1 yes, #2 yes.

SAGA2, for instance, has Rect32 and Rect16 which I could use there.

dreammaster:
Closing this pull request now, as the Rect commit has been merged in, and I did a separate commit from scratch to change the surface classes fields to int16

PR #3127, opened 2021-07-03
IMAGE: Improve BMP support

lolbot-iichan:
1. Allow various DIB headers (Windows v1-v5), see https://en.wikipedia.org/wiki/BMP_file_format#DIB_header_(bitmap_information_header) and fix loading palette for formats with extended DIB header
2. Allow 4bpp BMP images and fix pitch calculation for 4bpp
3. Extend raw decoder to support BMPs with alpha channel and use it for BMP compression method = 3

### DIFF
--- a/image/bmp.cpp
+++ b/image/bmp.cpp
@@ -78,8 +78,8 @@ bool BitmapDecoder::loadStream(Common::SeekableReadStream &stream) {
 	}
 
 	uint32 infoSize = stream.readUint32LE();
-	if (infoSize != 40 && infoSize != 108) {
-		warning("Only Windows v3 & v4 bitmaps are supported");
+	if (infoSize != 40 && infoSize != 52 && infoSize != 56 && infoSize != 108 && infoSize != 124) {
+		warning("Only Windows v1-v5 bitmaps are supported, unknown header: %d", infoSize);
 		return false;
 	}
 
@@ -108,6 +108,8 @@ bool BitmapDecoder::loadStream(Common::SeekableReadStream &stream) {
 	/* uint32 pixelsPerMeterY = */ stream.readUint32LE();
 	_paletteColorCount = stream.readUint32LE();
 	/* uint32 colorsImportant = */ stream.readUint32LE();
+
+	stream.seek(infoSize - 40, SEEK_CUR);
 
 	if (bitsPerPixel == 4 || bitsPerPixel == 8) {
 		if (_paletteColorCount == 0)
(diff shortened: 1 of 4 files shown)


image/bmp.cpp
```diff
@@ -109,6 +109,8 @@ bool BitmapDecoder::loadStream(Common::SeekableReadStream &stream) {
 	_paletteColorCount = stream.readUint32LE();
 	/* uint32 colorsImportant = */ stream.readUint32LE();
 
+	stream.seek(infoSize - 40, SEEK_CUR);
```
criezy, line 112: This addition makes sense, but it makes me wonder whether I am missing something, or if the v4 BMP (with infoSize = 108) was not working properly before. Since @dreammaster added it in 154ec750, I am guessing he needed it for something and may have tested it, so he might be able to confirm.

criezy (approved):
This looks good to me.
I only have one small doubt with one addition for which I added a comment. It looks correct to me, but might need to be double checked.

image/bmp.cpp
lolbot-iichan: @criezy 
Since pixel data is read from `subStream`, the only thing affected by this additional `stream.seek()` is palette (if present).
So it only matters in case if 8bpp BMPv4 was used. In my tests, exactly those kind of images looked broken before current PR: изображение
However, you are correct, @dreammaster might have been dealing with some special resources that might be broken by this PR.

criezy: It seems more likely that his test case only included images higher than 8bpp.

dreammaster: Right. I was using the bitmaps that came with the Nuvie project, and I guess they were/are non-paletted, since they all seemed to render fine afterwards.

yuv422: I think the Nuvie BMPs are paletted non-compressed IIRC.

lolbot-iichan: @yuv422 Paletted BMPv1 (40 bytes header) images are not affected by this change as well.
Could you please point the location of Nuvie BMPs so that I could check their types?

bluegr: @lolbot-iichan The Nuvie BMPs are here:
https://github.com/nuvie/nuvie/tree/master/data
Not sure where they are located in ScummVM

mduggan: The same files are in `devtools/create_ultima/files/ultima6` in the ScummVM tree

bluegr:
@lolbot-iichan any news on this?

sev-:
He is busy with RL, will take a look at it as time permits

sev-:
I will rebase and merge this today.

sev-:
I merged it manually, thanks!

PR #3128, opened 2021-07-04
GUI: THEMES: Use scalable cursor in modern theme

lephilousophe:
Here is how it is without the change:
Screenshot_unscaled
After the change:
Screenshot_scaled

The SVG cursor is taken from remastered theme as the bitmap cursor is the same.
No other image is taken from remastered as they were all reworked to be scalable.


sev-:
Thanks!

PR #3129, opened 2021-07-04
MOHAWK: MYST: Add simulation for CD-ROM loading times

lotharsm:
This patch adds support for artificial delays during scene transitions in order to mimic the behavior of CD-ROM drives.

Usually, games supported by ScummVM are limiting their speed by themselves, e.g. through actor movement. This is not the case for the Myst series where the only transitions are the different scenes/cards. Even though the game itself is pretty slow-paced, the current implementation allows to 'rush' through it since you'll notice basically no loading times at all.

While faster loading speeds are in general a good idea, this is not the case for Myst (at least in my opinion). When this game was first released in 1993, CD-ROM drivers were slow - like double-speed slow. This means not only access times of 80 to 200ms (depending on which source you trust), but also a blazing transfer speed of 300KB/s at ideal conditions.

By adding a new game option to enable additional delays during scene transition, I try to mimic the loading speeds of the old CD drives. The way I obtained the values is by no means _really_ scientific. I played through the game on my trusty Windows 9x computer with the CD-ROM drive throttled to 4x speed. Unfortunately, I can't go lower with the drives I have, and the only drive that's theoretically capable of going 2x is currently broken.

By default, this option is disabled and only activated upon request. I think the same concept could be applied to Riven as well, even though Riven didn't fully ran from CD but rather allowed a full installation on the hard drive.

### DIFF
--- a/engines/mohawk/myst.cpp
+++ b/engines/mohawk/myst.cpp
@@ -933,12 +933,12 @@ void MohawkEngine_Myst::changeToStack(MystStack stackId, uint16 card, uint16 lin
 
 	// Add artificial CD-ROM delay
 	if (addCdRomDelay == true) {
-		if (_stack->getStackId() != kIntroStack || _stack->getStackId() != kMenuStack) {
+		if (_stack->getStackId() != kIntroStack && _stack->getStackId() != kMenuStack) {
 			// Pretty arbitrary delays to mimic a period correct 4x drive
 			// TODO: Since the disc layout of the original CD-ROMs is known,
 			//       it should be possible to adapt the delay depending on the
 			//       target stack in order to replicate the original loading times.
-			g_system->delayMillis(_rnd->getRandomNumberRng(200,500));
+			g_system->delayMillis(_rnd->getRandomNumberRng(1000,1200));
 		}
 	}
 
@@ -974,8 +974,11 @@ void MohawkEngine_Myst::changeToCard(uint16 card, TransitionType transition) {
 			// The original engine disables the mouse cursor when loading new cards.
 			_cursor->hideCursor();
 			_system->updateScreen();
-			// Pretty arbitrary delays to mimic a period correct 4x drive
-			g_system->delayMillis(_rnd->getRandomNumberRng(175,225));
+
+			// Pretty arbitrary delays to mimic a period correct 2x-4x drive
+			// Note: This is not only based on seeking times (only 80-120ms depending
+			//       on the source), but also accounts for loading the next chunk of data.
+			g_system->delayMillis(_rnd->getRandomNumberRng(350,400));
 			_cursor->showCursor();
 		}
 	}
--- a/engines/mohawk/myst.cpp
+++ b/engines/mohawk/myst.cpp
@@ -917,6 +917,17 @@ void MohawkEngine_Myst::changeToStack(MystStack stackId, uint16 card, uint16 lin
 	_cache.clear();
 	_gfx->clearCache();
 
+	// Add artificial CD-ROM delay
+	if (ConfMan.getBool("cdromdelay")) {
+		if (_stack->getStackId() != kIntroStack || _stack->getStackId() != kMenuStack) {
+			// Pretty arbitrary delays to mimic a period correct 4x drive
+			// TODO: Since the disc layout of the original CD-ROMs is known,
+			//       it should be possible to adapt the delay depending on the
+			//       target stack in order to replicate the original loading times.
+			g_system->delayMillis(_rnd->getRandomNumberRng(500,750));
+		}
+	}
+
 	changeToCard(card, kTransitionCopy);
 
 	if (linkDstSound)
@@ -940,6 +951,19 @@ void MohawkEngine_Myst::changeToCard(uint16 card, TransitionType transition) {
 
 	if (_card) {
 		_card->leave();
+	}
+
+	// Add artificial CD-ROM delay
+	if (ConfMan.getBool("cdromdelay")) {
+		if (_stack->getStackId() != kIntroStack && _stack->getStackId() != kMenuStack) {
+
+			// The original engine disables the mouse cursor when loading new cards.
+			_cursor->hideCursor();
+			_system->updateScreen();
+			// Pretty arbitrary delays to mimic a period correct 4x drive
+			g_system->delayMillis(_rnd->getRandomNumberRng(175,225));
+			_cursor->showCursor();
+		}
 	}
 
 	_card = MystCardPtr(new MystCard(this, card));
--- a/engines/mohawk/myst.cpp
+++ b/engines/mohawk/myst.cpp
@@ -102,6 +102,13 @@ MohawkEngine_Myst::MohawkEngine_Myst(OSystem *syst, const MohawkGameDescription 
 	// and to support the drop page and other actions in the options dialog.
 	assert(!_mainMenuDialog);
 	_mainMenuDialog = new MystMenuDialog(this);
+
+	// Enable CD-ROM delay simulation if necessary
+	if (ConfMan.getBool("cdromdelay")) {
+		addCdRomDelay = true;
+	} else {
+		addCdRomDelay = false;
+	}
 }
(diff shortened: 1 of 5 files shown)


engines/mohawk/myst.cpp
```diff
@@ -942,6 +967,21 @@ void MohawkEngine_Myst::changeToCard(uint16 card, TransitionType transition) {
 		_card->leave();
 	}
 
+	// Add artificial CD-ROM delay
+	if (addCdRomDelay == true) {
+		if (_stack->getStackId() != kIntroStack && _stack->getStackId() != kMenuStack) {
+			// The original engine disables the mouse cursor when loading new cards.
```
sev-, line 973: I had an impression that it shows a CD-ROM icon.  Like seen here, for example: https://youtu.be/9F9ahZQ7oP0?t=2793

lotharsm: Yes, you are right. However, the CD-ROM icon was a version exclusive for the PS1 and (IIRC) the Amiga CD32 version. The Windows/Mac versions I had available for testing just blank the cursor until the next card is shown.

engines/mohawk/myst.cpp
```diff
@@ -917,6 +931,17 @@ void MohawkEngine_Myst::changeToStack(MystStack stackId, uint16 card, uint16 lin
 	_cache.clear();
 	_gfx->clearCache();
 
+	// Add artificial CD-ROM delay
+	if (addCdRomDelay == true) {
```
bluegr, line 935:
```suggestion
	if (addCdRomDelay) {
```

engines/mohawk/myst.cpp
```diff
@@ -533,6 +540,13 @@ void MohawkEngine_Myst::applyGameSettings() {
 		_gfx->loadMenuFont();
 		changeToStack(_stack->getStackId(), _card->getId(), 0, 0);
 	}
+
+	// Toggle CD-ROM simulation if necessary
+	if (ConfMan.getBool("cdromdelay")) {
+		addCdRomDelay = true;
+	} else {
+		addCdRomDelay = false;
+	}
```
bluegr, line 549: You can just write:
    addCdRomDelay = ConfMan.getBool("cdromdelay");

engines/mohawk/myst.cpp
```diff
@@ -942,6 +967,21 @@ void MohawkEngine_Myst::changeToCard(uint16 card, TransitionType transition) {
 		_card->leave();
 	}
 
+	// Add artificial CD-ROM delay
+	if (addCdRomDelay == true) {
```
bluegr, line 971:
```suggestion
	if (addCdRomDelay) {
```

engines/mohawk/myst.cpp
```diff
@@ -102,6 +102,13 @@ MohawkEngine_Myst::MohawkEngine_Myst(OSystem *syst, const MohawkGameDescription
 	// and to support the drop page and other actions in the options dialog.
 	assert(!_mainMenuDialog);
 	_mainMenuDialog = new MystMenuDialog(this);
+
+	// Enable CD-ROM delay simulation if necessary
+	if (ConfMan.getBool("cdromdelay")) {
+		addCdRomDelay = true;
+	} else {
+		addCdRomDelay = false;
+	}
```
bluegr, line 111: You can just write:
addCdRomDelay = ConfMan.getBool("cdromdelay");

engines/mohawk/myst.cpp
```diff
@@ -917,6 +931,17 @@ void MohawkEngine_Myst::changeToStack(MystStack stackId, uint16 card, uint16 lin
...
 	_gfx->clearCache();
 
+	// Add artificial CD-ROM delay
+	if (addCdRomDelay == true) {
+		if (_stack->getStackId() != kIntroStack && _stack->getStackId() != kMenuStack) {
+			// Pretty arbitrary delays to mimic a period correct 4x drive
+			// TODO: Since the disc layout of the original CD-ROMs is known,
+			//       it should be possible to adapt the delay depending on the
+			//       target stack in order to replicate the original loading times.
+			g_system->delayMillis(_rnd->getRandomNumberRng(1000,1200));
```
bluegr, line 941:
```suggestion
			g_system->delayMillis(_rnd->getRandomNumberRng(1000, 1200));
```

engines/mohawk/myst.cpp
```diff
@@ -942,6 +967,21 @@ void MohawkEngine_Myst::changeToCard(uint16 card, TransitionType transition) {
...
+	if (addCdRomDelay == true) {
+		if (_stack->getStackId() != kIntroStack && _stack->getStackId() != kMenuStack) {
+			// The original engine disables the mouse cursor when loading new cards.
+			_cursor->hideCursor();
+			_system->updateScreen();
+
+			// Pretty arbitrary delays to mimic a period correct 2x-4x drive
+			// Note: This is not only based on seeking times (only 80-120ms depending
+			//       on the source), but also accounts for loading the next chunk of data.
+			g_system->delayMillis(_rnd->getRandomNumberRng(300,400));
```
bluegr, line 980:
```suggestion
			g_system->delayMillis(_rnd->getRandomNumberRng(300, 400));
```

engines/mohawk/myst.h
```diff
@@ -147,6 +147,9 @@ class MohawkEngine_Myst : public MohawkEngine {
 	uint16 getMainCursor() { return _mainCursor; }
 	void refreshCursor();
 	bool wait(uint32 duration, bool skippable = false);
+	bool addCdRomDelay;
+	uint minCdRomDelay;
+	uint maxCdRomDelay;
```
bluegr, line 152: These two (min/max) aren't used anywhere

lotharsm:
Thanks @bluegr for the review! I followed your recommendations and modified the code accordingly.

digitall:
@lotharsm : Have looked over the code changes and they look fine. An interesting change. Will try a quick build and test later, but no problem here for merging :)

engines/mohawk/myst.cpp
```diff
@@ -917,6 +923,17 @@ void MohawkEngine_Myst::changeToStack(MystStack stackId, uint16 card, uint16 lin
...
 	_gfx->clearCache();
 
+	// Add artificial CD-ROM delay
+	if (addCdRomDelay) {
+		if (_stack->getStackId() != kIntroStack && _stack->getStackId() != kMenuStack) {
+			// Pretty arbitrary delays to mimic a period correct 4x drive
+			// TODO: Since the disc layout of the original CD-ROMs is known,
+			//       it should be possible to adapt the delay depending on the
+			//       target stack in order to replicate the original loading times.
+			g_system->delayMillis(_rnd->getRandomNumberRng(1000, 1200));
```
Mataniko, line 933: Since the number is arbitrary, you could consider removing the addCdRomDelay flag and replace it with a numeric configuration for a delay, and use that+200 random to simulate a delay. That way users can find the "correct" delay for them

sev-: This would require switching to a slider in the GUI, or a textfield.

lotharsm: Indeed, this would be a nice touch. I need to dig deeper in the GUI code to be able to do this, since I've hardly ever done any GUI code.

Maybe this could be done in a second PR?

digitall:
@lotharsm : Have built and tested with Myst ME. Works fine for me. With the setting turned off, you get the current behaviour which is a bit quick to change cards compared to the original experience which was more a more sedate pace and part of the game experience (though hiding the technology limitation of the drive speed). With the setting turned on, you get something which is pretty close to the original engine behaviour on original hardware, so am happy with this.

I have no view on which should be the "default", but fine to leave as-is here and have the option to be turned on for people who really want it :) I'd suggest merging this and dealing with adding a GUI slider for the delay in a different PR...

lotharsm:
@digitall: Thanks a lot for your feedback - I'll merge this tonight then :)

I think for now the option should be disabled by now - we can always make it the default at a later point in time based on user feedback.

PR #3130, opened 2021-07-04
TESTBED: Don't fail when pixel format couldn't be set

lephilousophe:
When pixel format isn't supported an assertion can be raised during
blitting. Don't draw anything instead.


lolbot-iichan:
This is a good fix, thanks! Merging.

PR #3136, opened 2021-07-07
DIRECTOR: warlock fix

ysj1173886760:
the fixing that aiming to bring ss warlock back
also including some operations for text chunk fields
mostly the detail problems


sev-:
Merging as is

PR #3137, opened 2021-07-08
BACKENDS: 3DS; SWORD25, TITANIC: Resolve leftover signedness issues

BallM4788:
Resolves leftover signedness issues to conform to baccbedf50e8d2b08df508170e0c0a4b25b8b0b6.


digitall:
@BallM4788 : Thanks for these fixes. Apart from fixing the conflict, can you split these into a commit per engine and 3DS backend please? This makes future bisection much easier as the changes are not really directly related.

BallM4788:
@digitall there, that should do it. @dreammaster has fixed the MOHAWK instances, so I removed them from this pull.

digitall:
@BallM4788 : Looks good now. Will merge as-is. Thanks again for this work.

PR #3140, opened 2021-07-08
ENGINES: Fix some compiler warnings

mparnaudeau:
- Fix -Wsign-compare in engines/titanic
- Fix -Wformat-truncation in engines/queen
- Fix -Wunused-variable in engines/trecision


digitall:
@mparnaudeau : Thanks for these fixes. I was already looking at these, but you have pretty much the same changes I was intending. One point, can you split these into a commit per engine please? This makes future bisection much easier as the changes are not really related.

mparnaudeau:
@digitall : You're right, I'm about to do that.

digitall:
@mparnaudeau : Thanks. Looks good. Will merge as-is.

PR #3141, opened 2021-07-09
EVENTRECORDER: added new event for OSystem::getTimeAndDate

mgerhardy:
`OSystem::getTimeAndDate()` is used in e.g. SCI and SCUMM - this wires it up to the event recorder.

Introducing a new event and changed the `getTimeAndDate` method

### DIFF
--- a/backends/platform/3ds/osystem.cpp
+++ b/backends/platform/3ds/osystem.cpp
@@ -169,7 +169,7 @@ void OSystem_3DS::delayMillis(uint msecs) {
 	svcSleepThread(msecs * 1000000);
 }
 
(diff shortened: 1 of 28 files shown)


sluicebox:
Wow, such quick turn around from a discord comment to a feature! Thanks!

I tested this with the debugger on a few SCI games and in Codename Iceman there's a part that prints the current date; I changed my clock and recorded and changed it back and the record-date correctly displayed on playback.

I now see that in SCI we call `OSystem::getTimeAndDate()` in situations where we don't use the result in kGetTime. I'll change that after this is merged, since it's a code-path that gets called a lot in some games, and that will cut down on the number of these new events that get generated.

sluicebox (approved):

PR #3142, opened 2021-07-09
SYMBIAN: Final batch to make Symbian port self-builded.

fedor4ever:
This is final batch to make Symbian port self-builded. I hope.

Add missed library dependency in build_app_config.mmh.
Resolve linker errors by compiling certain source files with exe. This happens because these files should depends from specific engines.
Added generator for installer builder.
Added source patcher. It make abld build system happy if these patches break other ports and therefore can't accepted. Currently fixes "..." in macros arguments.
Delete obsolete files.


sev-:
The commit log messages are not conformant with our guidelines

fedor4ever:
Commit log messages are conformant now.

sev-:
Thank you

PR #3143, opened 2021-07-10
AGS: Add missing games

aviloria:
Add Symploke games from Steam store and moved the games to the commercial section


sev-:
Thank you!

PR #3145, opened 2021-07-11
AGS: Remove dupped entry

aviloria:
"symploke1" Steam demo version is the same as the "symploke1" (English) release as the game became free2play.
No symploke1 demo is available anywhere


sev-:
Thanks.

PR #3146, opened 2021-07-11
Add Macintosh black-and-white mode for Mac Loom and Indy3

eriktorbjorn:
This adds a black-and-white mode for Mac Loom and Indy3. This feature was actually requested several years ago [1]. I don't know if that guy is still around, but maybe others want it as well.

Each color is dithered to a 2x2 pattern. I don't know if the actual patterns were some sort of standard, or if LucasArts came up with them themselves. But I believe that the result is pixel-perfect compared to the originals. I have put these, along with the Mac "EGA" palette, in a new header file, graphics/macega.h in case they can be reused elsewhere. When I did it, I was mainly thinking about the Mac AGI games, though looking closer at screenshots I suspect they used 3x2 dither patterns instead. Oh well.

Text is rendered as black or white (I think all of the in-game text is white), except for dark gray which is rendered using a checkerboard pattern. This is used for disabled notes in Loom, and for the replacement verb GUI in Indy3. The Loom notes appear to be pixel perfect, except I still can't figure out exactly if/when/how the notes used shadowed text. Oh well, that'll have to be good enough.

When running in 16 colors, the games use the shadow palette to remap colors. This makes no sense when you only have two colors, so these color changes are handled by the renderer, and any change to the shadow palette triggers a full redraw of the screen. This is used for the lightning flashes around Castle Brunwald in Indy3, and for the scene in Loom where the dragon finds Rusty asleep. (If there are other scenes, I'm not aware of them.)

I have replayed all of Loom, because that's the one where I was the most worried about text color. In Indy3, most text is drawn using a separate functions (for the text boxes) that only knows how to draw white text.

[1] https://forums.scummvm.org/viewtopic.php?f=1&t=12489

### DIFF
--- a/common/rendermode.cpp
+++ b/common/rendermode.cpp
@@ -44,6 +44,7 @@ const RenderModeDescription g_renderModes[] = {
 	{ "2gs", "Apple IIgs", kRenderApple2GS },
 	{ "atari", "Atari ST", kRenderAtariST },
 	{ "macintosh", "Macintosh", kRenderMacintosh },
+	// I18N: Macintosh black-and-white
 	{ "macintoshbw", _s("Macintosh b/w"), kRenderMacintoshBW },
 	{nullptr, nullptr, kRenderDefault}
 };
--- a/common/rendermode.cpp
+++ b/common/rendermode.cpp
@@ -44,6 +44,7 @@ const RenderModeDescription g_renderModes[] = {
 	{ "2gs", "Apple IIgs", kRenderApple2GS },
 	{ "atari", "Atari ST", kRenderAtariST },
 	{ "macintosh", "Macintosh", kRenderMacintosh },
+	{ "macintoshbw", _s("Macintosh b/w"), kRenderMacintoshBW },
 	{nullptr, nullptr, kRenderDefault}
 };
 
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -168,7 +168,7 @@ static const char HELP_STRING[] =
 	"                           (default: enabled)\n"
 	"  --render-mode=MODE       Enable additional render modes (hercGreen, hercAmber,\n"
 	"                           cga, ega, vga, amiga, fmtowns, pc9821, pc9801, 2gs,\n"
-	"                           atari, macintosh)\n"
+	"                           atari, macintosh, macintoshbw)\n"
 #ifdef ENABLE_EVENTRECORDER
 	"  --record-mode=MODE       Specify record mode for event recorder (record, playback,\n"
 	"                           passthrough [default])\n"
--- a/common/rendermode.h
+++ b/common/rendermode.h
@@ -57,7 +57,8 @@ enum RenderMode {
 	kRenderPC9801 = 9,
 	kRenderApple2GS = 10,
 	kRenderAtariST = 11,
-	kRenderMacintosh = 12
+	kRenderMacintosh = 12,
+	kRenderMacintoshBW = 13
 };
 
 struct RenderModeDescription {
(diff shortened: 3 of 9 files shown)


eriktorbjorn:
I'm actually surprised at how good the output looks. You'd think that 16 colors mapped to 2x2 patterns would just come out as a horrid mess. Particularly for Loom, which uses a lot of dithering to begin with.

indy3-bw

loom-bw

eriktorbjorn:
It turns out that LucasArts didn't implement the palette changes in their interpreter:

loomdragon-minivmac

The image is small because I was emulating a 9" Mac screen since that's seemed like the most likely use case back then. This is how it currently looks in ScummVM:

loomdragon-scummvm

Which I think is an improvement. I'll check Indy 3 next.

eriktorbjorn:
You could argue that Rusty is quite nicely shaded in the Mini vMac screenshot, without the palette change, but for the Indy 3 castle exterior I think there's no contest:

indycastle-minivmac

The lightning looks very nice in black and white, and of course it's even better with the sound.
indycastle-scummvm

common/rendermode.cpp
```diff
@@ -44,6 +44,7 @@ const RenderModeDescription g_renderModes[] = {
 	{ "2gs", "Apple IIgs", kRenderApple2GS },
 	{ "atari", "Atari ST", kRenderAtariST },
 	{ "macintosh", "Macintosh", kRenderMacintosh },
+	{ "macintoshbw", _s("Macintosh b/w"), kRenderMacintoshBW },
```
sev-, line 47: Please add a comment preceding this line, for the translators, something like
```C++
// I18N: Macintosh black-and-white
```

sev- (approved):
Cool stuff. Please add that comment for translators and we're good to go.

sev-:
Thanks!

pmaciel:
Hi @eriktorbjorn, I'm the one one with the original request -- it now it looks abolutelly gorgeous :-D
I had no clue it would also work on larger resolutions and with your palette fix, it's a new experience (specifically Loom). Thanks for your great work!
Pedro

eriktorbjorn:
@pmaciel I didn't know you were still around. I'm glad you got to see the results!

I was actually a bit skeptical about the whole thing, but I figured that if (as it turned out) the dithering was just a simple matter of replacing each color with a 2x2 dithering pattern, then it would be a pretty small addition since the graphics already had to be upscaled to accommodate for the high-resolution font. That part was a bit of work, but others had already figured out how to handle Mac resources and render the fonts.

(I briefly toyed with the idea of adding a Macintosh black-and-white mode for the Sierra AGI games as well, but it seems different versions use different dithering methods. So no, I'll happily leave that to someone else. :-)

Please let me know if you run into issues, other than the known ones:

- The games are rendered at 640x400 pixels. I believe this to be a reasonable approximation of how they would look on a Macintosh II, running in 2-color mode. Models with smaller screens would have cropped the image to - I believe - 512x342 pixels, and I believe they would also have had to adjust parts of the GUI to fit. (E.g. Loom would draw the notes box in Practice mode above the distaff instead of below it.) I see little reason to support this.
- At least Loom had different settings for music quality. ScummVM only emulates what I believe is the highest quality version of the music. (I've never heard it on real hardware, so I don't know how close it is.) I'm not sure if it's worth pursuing the lower quality music, though I vaguely seem to recall stumbling over the medium-quality tracks while adding support for the custom instruments. So maybe I'll add that at some point... (I think the low quality music was played through the Mac synthesizer. See below.)
- Indy 3's music doesn't sound like the original, since it's played through some sort of AdLib emulation where the original would use the Mac's synthesizer. But I lack all knowledge needed to emulate that.
- There's that Indy 3 GUI that isn't implemented at all. I guess it would be reasonable to add support for that at some point, but it seems like quite a bit of work.

That's all I can think of at the moment.

PR #3147, opened 2021-07-11
EVENTRECORDER: Move Recorder to a separate button

orgads:
When building with event recorder, the button is hidden as a sub-button under Load. This is hard to discover, and the action is not related to load.

Move the button out of there, and make it a top-level button.


mgerhardy:
getting this at startup - anything I've missed to update?

```
WARNING: Could not find theme 'scummremastered' falling back to builtin!
WARNING: widget h: 21916	w: 0!
Unable to load widget position for 'Launcher.RecorderButton'. Please check your theme files!
```

orgads:
Which platform do you build on?

mgerhardy:
Linux and make

orgads:
Did you run make install?

mgerhardy:
no, didn't do it - never did it before - afaik you need to bump the theme versions to make this working.

mgerhardy:
Bildschirmfoto von 2021-07-11 14-23-44

bluegr:
Looks good! Do the button descriptions look OK with other languages?

sev-:
As I was mentioning on Discord, I do not see much value in un-hiding it. This will take the real estate on the GUI but will be used extremely rarely.

Another idea, if you want to give a bigger exposure to the feature: put the button (or copy of it) to the Game Options -> Game tab. Thoughts?

orgads:
I think that placing it in the Options dialog is misleading. This is not a setting, but a different way to start/load the game.

I agree that this action is uncommon, much less than Start and Load, but on the other hand, it really has nothing to do with Load, so having it as a subaction of Load is questionable.

criezy:
My first thought when I saw the pull request was similar to what @sev- wrote. But I waited a few days before commenting to give me time to sort my thoughts so that I can try to explain why.

I am not sure exposing this button to the launcher is a good idea as to me it seems that the recorder feature is not really useful for users and is almost exclusively for developers (and even for those it was not be used often). Making it more visible might confuse users who will not understand  what it does or how to use it. And it will also make the GUI more cramped on devices with a smalls screen (and this might be an issue with some languages that need longer labels than in English for the buttons). I would tend to prioritise usability for standard users and keep the launcher as simple as possible.

It seems to me that keeping it hidden would not really be an issue as long as the few people for whom it might be useful know it is there and how to access it. Maybe documenting this on the Developer Central wiki page would be a better idea (actually I just checked and it is already mentioned in the section for engine authors).

bluegr:
I concur that placing the record button together with the load button makes more sense, as this is a feature mostly used by developers.

However, this feature could also be used for game walkthroughs, e.g. a player could record a playthrough of part of a hard puzzle and redistribute it to others like a walkthrough

sev-:
so, let's close this for now.

PR #3148, opened 2021-07-11
AGS: Pointers should not be casted to long

aviloria:
New string arithmetic is using pointers diff casted to long...
Trying to avoid C++11 features, the best option is to cast them to intptr_t instead of long


aviloria:
These changes should fix scummvm compilation

digitall:
@aviloria : Thanks for your contribution. Looks reasonable... Will just await a quick review from @dreammaster who is working on this engine.

dreammaster:
Looks good to me

PR #3150, opened 2021-07-13
XEEN RU Localization

Ardash:
I've added support for Russian version of Clouds of Xeen

Edit: There is issue in code analysis. I've fixed it.


sev-:
This must be rebased, so there is no merge commit. Also, it would be nice if the whole thing is moved to a separate branch, so the temporary comments in Russian are not visible to the team.

sev-:
Also, it must be rebased because the early commits do not conform with our commit log guidelines, e.g. they're lacking the engine prefix and there is even a comment in Russian.

PR #3152, opened 2021-07-13
GOB: Avoid an off by one

lu-zero:
Lost in Time (English floppy version but CD version might be affected as well) garbles colours during some clips playback otherwise.

The easiest way to reproduce is in manor section of the game: when it starts, just click the horse at the gate or any object in the tractor. A video will play that will also garble one colour in the picture to red (the same happens in other parts of manor section when small clips play). 

I believe it's caused by an off-by-one error in palette copy (e.g. for PORTAL04.IMD colours 16-127 are copied).

This patch seems to fix it.


digitall:
Can replicate the issue with the horse video. Will merge for testing...

digitall:
@lu-zero : Thanks for your code contribution. As expected, this fixes the issue and probably other minor palette bugs with video playback in GOB games.

PR #3153, opened 2021-07-13
BLADERUNNER: fix encodings for save descriptions

athrxx:
If you type save names like e. g. "äöüß" in the GMM this will now show up correctly in the ingame save/load dialogs.

You still cannot type these characters ingame due to the current implementation of UIInputBox::charIsValid(). I don't know the purpose of the function. If I comment out everything and simply return true it works fine (I can enter all these special characters). So it is not a limitation of the font, even though I have the English version. If the only reason for charIsValid() was the wrong encoding it might make sense to limit or even remove it.

### DIFF
--- a/engines/bladerunner/ui/ui_input_box.cpp
+++ b/engines/bladerunner/ui/ui_input_box.cpp
@@ -88,6 +88,9 @@ void UIInputBox::hide() {
 }
 
 void UIInputBox::handleKeyDown(const Common::KeyState &kbd) {
+	// The values that the KeyState::ascii field receives from the SDL backend are actually ISO 8859-1 encoded. They need to be
+	// reencoded to DOS so as to match the game font encoding (although we currently use UIInputBox::charIsValid() to block most
+	// extra characters, so it might not make much of a difference).
 	char kc = Common::U32String(Common::String::format("%c", kbd.ascii), Common::kISO8859_1).encode(Common::kDos850).firstChar();
 	if (_isVisible) {
 		if (charIsValid(kc) && _text.size() < _maxLength) {
--- a/engines/bladerunner/ui/ui_input_box.cpp
+++ b/engines/bladerunner/ui/ui_input_box.cpp
@@ -88,9 +88,10 @@ void UIInputBox::hide() {
 }
 
 void UIInputBox::handleKeyDown(const Common::KeyState &kbd) {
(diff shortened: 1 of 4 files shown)


bluegr:
@antoniou79 any news on this?

antoniou79:
Sorry, I didn't have the time to look into this properly yet.
I expect to do this in the following days/week.

engines/bladerunner/ui/ui_input_box.cpp
```diff
@@ -88,9 +88,10 @@ void UIInputBox::hide() {
 }
 
 void UIInputBox::handleKeyDown(const Common::KeyState &kbd) {
+	char kc = Common::U32String(Common::String::format("%c", kbd.ascii), Common::kISO8859_1).encode(Common::kDos850).firstChar();
```
peterkohaut, line 91: can you add some description & possible examples to a comment as this looks very confusing

athrxx: Done. It is kind of difficult to test, since UIInputBox::charIsValid() will block all these characters anyway (regardless of encoding). If you want to test it you have to comment out the checks in there, in particular the `kc <= '~'`.
After that it will basically depend on your locale and your keyboard which extra characters you can test (I don't know where you're located). I have tested the 'äüöß' characters. The font does have glyphs for it, even in the English version of the game.

peterkohaut: Looks okay, but SDL is only one backend (and I didn't know that SDL do conversion to 8859-1), there are more like Android or IOS.
Do we really need this feature - support for national character entry in-game? Its adding unnecessary complexity a possible issues for other backends.
Maybe we can omit this one, in ScummVM's GUI is okay, but for game... I don't think benefits surpasses the possible issues and more complex code.

sev-:
The change seems fine. Merging

sev-:
I wonder, shall this one be backported to branch-2-5?

antoniou79:
I share Peter's concerns on this; this fix introduces a support for converting utf characters into game charset glyphs, and that may lead to new issues, especially since we have support for exporting legacy format for save games, the game's glyph sets had some weirdness in the ordering of the glyphs, and that ideally we want to support fan translations that alter those sets.
I have a list of cases I wanted to check, which seemed like they would break things, but did not get around to it unfortunately.
I guess merging with master means I can test faster. 
However, at least at the moment, I think we should wait and not backport it to 2.5 straight away.

PR #3156, opened 2021-07-14
CGE2: Text To Speech (TTS) for the Sfinx Game

taylorzhancher:
There will be TTS for the following
- Options in the game
- Speech in the game

And options to do these things
- Turn TTS for objects on an off
- In the English version, an option to override the original voice over with TTS (in English)

### DIFF
--- a/engines/cge2/cge2.cpp
+++ b/engines/cge2/cge2.cpp
@@ -25,6 +25,7 @@
  * Copyright (c) 1994-1997 Janusz B. Wisniewski and L.K. Avalon
  */
 
+#include "engines/advancedDetector.h"
 #include "engines/util.h"
 #include "common/text-to-speech.h"
 #include "common/config-manager.h"
@@ -197,6 +198,10 @@ Common::Error CGE2Engine::run() {
 	ConfMan.flushToDisk();
 
 	return Common::kNoError;
+}
+
+Common::Language CGE2Engine::getLanguage() const {
+	return _gameDescription->language;
 }
 
 } // End of namespace CGE2
--- a/engines/cge2/cge2.cpp
+++ b/engines/cge2/cge2.cpp
@@ -184,8 +184,8 @@ bool CGE2Engine::hasFeature(EngineFeature f) const {
 }
 
 Common::Error CGE2Engine::run() {
-    // if (_ttsMan != nullptr)
-	// 	_ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
+    if (_ttsMan != nullptr)
+		_ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
 	syncSoundSettings();
 	initGraphics(kScrWidth, kScrHeight);
 
--- a/engines/cge2/cge2.cpp
+++ b/engines/cge2/cge2.cpp
@@ -183,7 +183,7 @@ bool CGE2Engine::hasFeature(EngineFeature f) const {
 
 Common::Error CGE2Engine::run() {
 	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
-    if (ttsMan != nullptr)
+	if (ttsMan != nullptr)
 		ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
 	syncSoundSettings();
 	initGraphics(kScrWidth, kScrHeight);
--- a/engines/cge2/cge2.cpp
+++ b/engines/cge2/cge2.cpp
@@ -26,6 +26,7 @@
  */
 
 #include "engines/util.h"
+#include "common/text-to-speech.h"
 #include "common/config-manager.h"
 #include "common/debug.h"
 #include "common/debug-channels.h"
@@ -105,6 +106,8 @@ CGE2Engine::CGE2Engine(OSystem *syst, const ADGameDescription *gameDescription)
 		_oldSfxVolume = ConfMan.getInt("sfx_volume");
 		_music = _oldMusicVolume != 0;
 	}
+
+	_ttsMan = g_system->getTextToSpeechManager();
 }
 
 void CGE2Engine::init() {
@@ -181,6 +184,8 @@ bool CGE2Engine::hasFeature(EngineFeature f) const {
 }
 
 Common::Error CGE2Engine::run() {
+    // if (_ttsMan != nullptr)
+	// 	_ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
 	syncSoundSettings();
 	initGraphics(kScrWidth, kScrHeight);
 
--- a/engines/cge2/cge2.cpp
+++ b/engines/cge2/cge2.cpp
@@ -106,8 +106,6 @@ CGE2Engine::CGE2Engine(OSystem *syst, const ADGameDescription *gameDescription)
 		_oldSfxVolume = ConfMan.getInt("sfx_volume");
 		_music = _oldMusicVolume != 0;
 	}
-
-	_ttsMan = g_system->getTextToSpeechManager();
 }
 
 void CGE2Engine::init() {
@@ -184,8 +182,9 @@ bool CGE2Engine::hasFeature(EngineFeature f) const {
 }
 
 Common::Error CGE2Engine::run() {
-    if (_ttsMan != nullptr)
-		_ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+    if (ttsMan != nullptr)
+		ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
 	syncSoundSettings();
 	initGraphics(kScrWidth, kScrHeight);
 
--- a/engines/cge2/cge2.h
+++ b/engines/cge2/cge2.h
@@ -157,6 +157,7 @@ public:
 	bool hasFeature(EngineFeature f) const override;
 	bool canSaveGameStateCurrently() override;
 	bool canLoadGameStateCurrently() override;
+	Common::Language getLanguage() const;
 	Common::Error saveGameState(int slot, const Common::String &desc, bool isAutosave = false) override;
 	Common::Error loadGameState(int slot) override;
 	Common::Error run() override;
--- a/engines/cge2/cge2.h
+++ b/engines/cge2/cge2.h
@@ -162,8 +162,6 @@ public:
 	Common::Error loadGameState(int slot) override;
 	Common::Error run() override;
 
-	Common::TextToSpeechManager *_ttsMan;
-
 	WARN_UNUSED_RESULT static bool readSavegameHeader(Common::InSaveFile *in, SavegameHeader &header, bool skipThumbnail = true);
 
 	bool showTitle(const char *name);
--- a/engines/cge2/cge2.h
+++ b/engines/cge2/cge2.h
@@ -161,6 +161,8 @@ public:
 	Common::Error loadGameState(int slot) override;
 	Common::Error run() override;
 
+	Common::TextToSpeechManager *_ttsMan;
+
 	WARN_UNUSED_RESULT static bool readSavegameHeader(Common::InSaveFile *in, SavegameHeader &header, bool skipThumbnail = true);
 
 	bool showTitle(const char *name);
--- a/engines/cge2/detection.cpp
+++ b/engines/cge2/detection.cpp
@@ -38,7 +38,7 @@ static const DebugChannelDef debugFlagList[] = {
 namespace CGE2 {
 
 #define GAMEOPTION_COLOR_BLIND_DEFAULT_OFF  GUIO_GAMEOPTIONS1
-#define GAMEOPTION_TTS_THINGS				GUIO_GAMEOPTIONS2
+#define GAMEOPTION_TTS_OBJECTS				GUIO_GAMEOPTIONS2
 #define GAMEOPTION_TTS_SPEECH				GUIO_GAMEOPTIONS3
 
 static const PlainGameDescriptor CGE2Games[] = {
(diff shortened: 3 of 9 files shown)


engines/cge2/talk.cpp
```diff
@@ -254,6 +261,14 @@ InfoLine::InfoLine(CGE2Engine *vm, uint16 w, ColorBank color)
 }
 
 void InfoLine::update(const char *text) {
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+    if (text) {
+        if (lastText != text) {
```
mgerhardy, line 266: this looks suspicious. `lastText != text` is no string compare. It might be that you are comparing a buffer pointer with different content since the last call - but as the same buffer was used, the check didn't evaluate to true. It would imo be better to perform a string compare here.

engines/cge2/talk.h
```diff
@@ -69,6 +69,7 @@ class Talk : public Sprite {
 public:
 	uint8 *_color;
 
+	const char *lastText;
```
mgerhardy, line 72: lastText is not initialized in the constructor

engines/cge2/vmenu.cpp
```diff
@@ -25,16 +25,22 @@
...
 
+#include "common/config-manager.h"
+#include "common/text-to-speech.h"
 #include "cge2/text.h"
 #include "cge2/vmenu.h"
 #include "cge2/events.h"
 
 namespace CGE2 {
 
+const char *_iWantToStopPlayingAlready;
```
mgerhardy, line 36: these aren't initialized - so they might send garbage in `VMenu::touch` to `ttsMan->say`

taylorzhancher:
Just did the changes

engines/cge2/detection.cpp
```diff
@@ -101,6 +103,28 @@ static const ADExtraGuiOptionsMap optionsList[] = {
...
+			false
+		}
+	},
+
+	{
+		GAMEOPTION_TTS_SPEECH,
+		{
+			_s("Enable Text to Speech for Subtitles and Override Original Voice Overs"),
+			_s("Use TTS to read the descriptions (if TTS is available)"),
+			"tts_enabled_new",
```
mgerhardy, line 122: I'm not sure about the name of the config variable. it might be used in other engines, too - and should imo get a better speaking name.

Also I'm no native speaker, but the upper case writing of the description doesn't look correct for me. Is there a reason behind writing it like this? Like highlighting words with special meanings in the game?

engines/cge2/vmenu.h
```diff
@@ -76,6 +76,7 @@ class VMenu : public Talk {
 	char *_vmgt;
 	static VMenu *_addr;
 	int _recent;
+	int lastN = 2;
```
mgerhardy, line 79: cge2 is not marked as c++11 - this would break in c++98 compilation mode. You should init the var in the constructor.

engines/cge2/detection.cpp
taylorzhancher: The upper case is there just because it was also there in the other engines I worked on, just thinking about uniforming the format.

engines/cge2/vmenu.h
taylorzhancher: Is it this particular variable? or all the variables?

engines/cge2/talk.cpp
```diff
@@ -254,6 +261,14 @@ InfoLine::InfoLine(CGE2Engine *vm, uint16 w, ColorBank color)
...
 
 void InfoLine::update(const char *text) {
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+	if (text) {
+        if (strcmp(lastText,text) != 0) {
+            if (ttsMan != nullptr && ConfMan.getBool("tts_enabled"))
+                ttsMan->say(text);
+            lastText = text;
+        }
+    }
```
criezy, line 271: Is the `lastText` check needed?
I would think that the default action, which is `INTERRUPT_NO_REPEAT`, would take care of not repeating the text.

Also could you not use `_oldText` instead of `lastText`?
Actually it seems to me you could move the whole block after the
```c++
if (!_realTime && (text == _oldText))
 		return;
```
you have just a few lines below and remove the strcmp check as a result.

engines/cge2/talk.h
```diff
@@ -69,6 +69,7 @@ class Talk : public Sprite {
 public:
 	uint8 *_color;
 
+	const char *lastText = "";
```
criezy, line 72: This does not comply with our naming conventions for class member variables (missing leading `_`).
Also this engine does not use c++11, so the initialisation of the variable should be in the constructor initialisation list and not here.
And finally, as noted in my previous comment, it seems to me you don't need to add this variable since we already have an `_oldText` variable in `InfoLine`.

engines/cge2/vmenu.cpp
```diff
@@ -25,16 +25,22 @@
...
+#include "common/config-manager.h"
+#include "common/text-to-speech.h"
 #include "cge2/text.h"
 #include "cge2/vmenu.h"
 #include "cge2/events.h"
 
 namespace CGE2 {
 
+const char *_iWantToStopPlayingAlready = "";
+const char *_thisIsAMistakeKeepPlaying = "";
```
criezy, line 37: Those are not class member variables, and thus should not start with an underscore.

engines/cge2/vmenu.cpp
```diff
@@ -151,6 +158,15 @@ void VMenu::touch(uint16 mask, V2D pos, Common::KeyCode keyCode) {
...
 
+		Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+		if (lastN != n) {
+			if (n == 0)
+				ttsMan->say(_iWantToStopPlayingAlready, Common::TextToSpeechManager::INTERRUPT);
+			else
+				ttsMan->say(_thisIsAMistakeKeepPlaying, Common::TextToSpeechManager::INTERRUPT);
+			lastN = n;
+		}
+
```
criezy, line 169: I don't understand why you need the `_iWantToStopPlayingAlready` and `_thisIsAMistakeKeepPlaying` variables.
Could you not use `_menu[n]->_text`?

engines/cge2/vmenu.cpp
criezy: Also as I noted below I don't think you actually need those variables.

engines/cge2/vmenu.h
criezy: All variables, and I added a comment regarding another variable with the same issue.
Also the variable name should start with an underscore since it is a class member variable.

criezy (changes requested):

engines/cge2/cge2.h
```diff
@@ -157,10 +157,13 @@ class CGE2Engine : public Engine {
 	bool hasFeature(EngineFeature f) const override;
 	bool canSaveGameStateCurrently() override;
 	bool canLoadGameStateCurrently() override;
+	Common::Language getLanguage() const;
 	Common::Error saveGameState(int slot, const Common::String &desc, bool isAutosave = false) override;
 	Common::Error loadGameState(int slot) override;
 	Common::Error run() override;
 
+	Common::TextToSpeechManager *_ttsMan;
```
criezy, line 165: One more comment: I don't think you need to make this a class member variable. It seems to be used in a single place (in `CGE2Engine::run()` to set the language), so it would be simpler to just call `g_system->getTextToSpeechManager()` there rather than store it as a class variable.

taylorzhancher: ok, moved.

engines/cge2/snail.cpp
```diff
@@ -656,7 +657,8 @@ void CGE2Engine::snSay(Sprite *spr, int val) {
 				i -= 100;
 			int16 oldRepeat = _sound->getRepeat();
 			_sound->setRepeat(1);
-			snSound(spr, i, Audio::Mixer::kSpeechSoundType);
+			if (!ConfMan.getBool("tts_enabled_new") && Common::getLanguageCode(getLanguage()) == "pl")
+				snSound(spr, i, Audio::Mixer::kSpeechSoundType);
```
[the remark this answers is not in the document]
taylorzhancher: fixed. sorry for this!

engines/cge2/metaengine.cpp
```diff
@@ -158,6 +158,10 @@ void CGE2MetaEngine::removeSaveState(const char *target, int slot) const {
 	g_system->getSavefileManager()->removeSavefile(fileName);
 }
 
+Common::Language CGE2Engine::getLanguage() const {
+	return _gameDescription->language;
+}
+
```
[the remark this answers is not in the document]
taylorzhancher: moving it into the cge2 file throws this error message

```
engines/cge2/cge2.cpp:203:25: error: member access into incomplete type
      'const ADGameDescription'
        return _gameDescription->language;
                               ^
```

engines/cge2/talk.cpp
taylorzhancher: I just tried both `INTERRUPT_NO_REPEAT` and moving it after `_oldText`. What happens is the object get repeatedly said over if the cursor is not moved.

engines/cge2/vmenu.cpp
taylorzhancher: This makes sense. I tried to do it but my compiler complained. I'll try to see again if I did anything wrong. 
```
engines/cge2/vmenu.cpp:164:12: error: no matching member function for
      call to 'say'
                        ttsMan->say(_menu[0], Common::TextToSpe...
                        ~~~~~~~~^~~
```

engines/cge2/metaengine.cpp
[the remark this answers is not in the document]
criezy: Indeed Dreamweb implements several `DreamWebEngine` functions in meta engine.cpp.
This seems a bit strange to me, but different developers would organise their code in slightly different ways. In the CGE2 engines it was not done this way, so I think it might be better not to introduce engine functions in the metaengine.cpp file.

Any idea why you are getting this error?
Where the the `ADGameDescription` struct declared?

engines/cge2/cge2.cpp
```diff
@@ -184,8 +182,9 @@ bool CGE2Engine::hasFeature(EngineFeature f) const {
 }
 
 Common::Error CGE2Engine::run() {
-    if (_ttsMan != nullptr)
-		_ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+    if (ttsMan != nullptr)
+		ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));
```
criezy, line 187: Indentation seems incorrect. Did you maybe use spaces instead of tabs on some of these lines? Line 186 seems to be the incorrect one.

engines/cge2/snail.cpp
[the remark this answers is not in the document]
criezy: You don't need to convert the language code to a string here. You can directly check the code:
```c++
if (!ConfMan.getBool("tts_enabled_speech") || getLanguage() == Common::PL_POL)
```

engines/cge2/talk.cpp
criezy: Ah indeed. Using INTERRUPT_NO_REPEAT would prevent to interrupt and start again the string that it is currently speaking, but once it has finished speaking it, it can say it again.

I would still expect the `text == _oldText` check to do the job, but I might be wrong there. It's possible that the pointers are different but the text value is the same. You should still be able to use the _oldText variable without having to add a `lastText` variable.

engines/cge2/vmenu.cpp
criezy: You need to pass the text of the menu, not the menu itself:
```c++
ttsMan->say(_menu[n]->_text, Common::TextToSpeechManager::INTERRUPT)
```

engines/cge2/metaengine.cpp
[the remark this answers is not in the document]
taylorzhancher: it's in engines/advancedDetector.h line 108
```
/**
 * Data table describing a game variant.
 */
struct ADGameDescription {
	/**
	 * The gameid that identifies the game.
	 *
	 * Mainly used for taking the game description from the @ref PlainGameDescriptor table.
	 */
	const char *gameId;

	/**
	 * Field that is used to distinguish between different variants of a game.
	 *
	 * The content of this field is inserted in the generated description for the config file game entry.
	 * If the @ref kADFlagUseExtraAsHint ADFlag is set, the contents of this field are stored
	 * in the config file, and are used to additionally distinguish between game variants.
	 * Also, if the ADGF_USEEXTRAASTITLE game flag is set, the contents of this field
	 * will be used as a description instead the one extracted from the @ref PlainGameDescriptor table.
	 */
	const char *extra;

	/**
	 * A list of individual file entries used for detection.
	 *
	 * 13 files (last is zero-terminated) is the maximum number of files currently used in ScummVM.
	 * A hardcoded number is used here due to a C++ limitation for defining const arrays.
	 */
	ADGameFileDescription filesDescriptions[14];

	/**
	 * Language of the game variant.
	 */
	Common::Language language;

	/**
	 * Platform of the game variant.
	 */
	Common::Platform platform;

	/**
	 * A bitmask of extra flags.
	 *
	 * The top 16 bits are reserved for generic flags defined in @ref ADGameFlags.
	 * This leaves 16 bits to be used by the client code.
	 */
	uint32 flags;

	/**
	 * Game features that are user-controllable.
	 *
	 * This list reflects which features of the GUI should be turned on or off in order to minimize user confusion.
	 * For instance, there is no point in changing game language in single-language games
	 * or have MIDI controls in a game that only supports digital music.
	 */
	const char *guiOptions;
};
```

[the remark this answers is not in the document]
criezy: It is indeed.
So do you understand what the error means and how you can solve it?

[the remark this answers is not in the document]
taylorzhancher: Ah HA!

engines/cge2/talk.cpp
taylorzhancher: I think I also tried that and it didn't work as we would like it to

engines/cge2/vmenu.cpp
taylorzhancher: ok

engines/cge2/cge2.cpp
taylorzhancher: Fixed. I would really like to know what I did to make sporadic spaces appear in my code.....

engines/cge2/vmenu.h
```diff
@@ -76,6 +76,7 @@ class VMenu : public Talk {
 	char *_vmgt;
 	static VMenu *_addr;
 	int _recent;
+	int _lastN = 2;
```
criezy, line 79: This variable is still not initialized in the correct place. This should be done in the `VMenu` constructor implementation (in the menu.cpp file).

engines/cge2/talk.h
```diff
@@ -69,6 +69,7 @@ class Talk : public Sprite {
 public:
 	uint8 *_color;
 
+	const char *_lastText = "";
```
criezy, line 72: This variable can be moved to the `InfoLine` class.
Also it should be initialized in the talk.cpp file and not here. And it can probably be initialized to a null pointer and not an empty string (see the `_oldText` variable in the InfoLine class).

engines/cge2/talk.cpp
criezy: I will try to get you branch in the next few days to play a bit with this as I don't know why it would not work.

taylorzhancher:
@criezy I changed the place where the variables are initialized as you have suggested. However I didn't set`_lastText` to null pointer, because it will crash when string compare happens at the very beginning.

criezy:
Please rebase the branch on the latest master. When you created a branch there was a bug that caused access to freed memory when playing wav files. This results in random crashes and makes testing your changes difficult. This bug has already been fixed in master, so rebasing your branch will pick up the fix.

engines/cge2/vmenu.cpp
```diff
@@ -99,6 +101,8 @@ VMenu::VMenu(CGE2Engine *vm, Common::Array<Choice *> list, V2D pos, ColorBank co
 	_bar = new MenuBar(_vm, _siz.x - 2 * kTextHMargin, _color);
 	_bar->gotoxyz(V2D(_vm, _pos2D.x, _pos2D.y + kTextVMargin - kMenuBarVerticalMargin));
 	_vm->_vga->_showQ->append(_bar);
+
+	_lastN = 2;
```
criezy, line 105: It would be better to do this initialisation in the constructor initializer list.

engines/cge2/talk.cpp
```diff
@@ -251,9 +258,18 @@ InfoLine::InfoLine(CGE2Engine *vm, uint16 w, ColorBank color)
 	V2D siz = V2D(_vm, w, kFontHigh);
 	b[0] = Bitmap(_vm, siz.x, siz.y, _color[2]);
 	setShapeList(b, 1);
+	_lastText = "";
```
criezy, line 261: It would be better to do this initialisation in the constructor initializer list.

If you don't know what this means, see for example https://www.learncpp.com/cpp-tutorial/constructor-member-initializer-lists/

engines/cge2/talk.cpp
criezy: I have verified that you don't need the `_lastText` check to prevent repeating the name of objects when the cursor is over an object.

Actually your current implementation is not ideal as when you move the cursor over an object it will speak its name, but then if you move the cursor away from the object, and then a bit later back on the object (without moving it on another object in between) it will not speak the name again. And I think it is wrong. Relying on the check already there with `_oldText` fixes this.

Since you may not have understood properly what I suggested, here is the code I just tested and seems to work properly for me:
```c++
void InfoLine::update(const char *text) {
	if (!_realTime && (text == _oldText))
		return;

	_oldText = text;

	if (text && ttsMan != nullptr && ConfMan.getBool("tts_enabled_objects"))
		ttsMan->say(text);
```

taylorzhancher: It now works for me. Thanks for the suggestion and code!

engines/cge2/talk.h
taylorzhancher: Indeed. Deleting.

taylorzhancher:
@criezy I deleted the `_lastText` variable and initialized `_lastN` in the constructor member initializer list

criezy (approved):
It looks good to me now.

criezy:
I will squash and rebase now.

PR #3158, opened 2021-07-15
DIRECTOR: UTF-8 Lingo

djsrv:
This changes Lingo to use UTF-8 internally, enabling Japanese and other non-Latin characters to be used in scripts. Any strings read from files are re-encoded to UTF-8. A handful of functions (charToNum, numToChar, and Datum::compareTo) vary by platform, so they map char codes between Unicode and the native encoding of whatever platform we're pretending to be.


engines/director/lingo/lingo-preprocessor.cpp
```diff
@@ -26,17 +26,36 @@
 
 namespace Director {
 
-Common::String preprocessWhen(Common::String in, bool *changed);
-
-bool isspec(char c) {
-	return strchr("-+*/%%^:,()><&[]=", c) != NULL;
+bool isspec(Common::u32char_type_t c) {
+	switch (c) {
+	case '-':
```
sev-, line 31: I am not sure it will work reliably, all these characters are ASCII, and you're checking against u32char_type_t.

engines/director/lingo/lingo.cpp
```diff
@@ -46,6 +46,93 @@ namespace Director {
 
 Lingo *g_lingo;
 
+static struct CharNormProto {
```
sev-, line 49: Consider creating a new file for this, like `lingo-utils.cpp` and move this utility function and the table there, so we have better logic separation.

sev- (approved):
All is good, just couple of small suggestions

engines/director/lingo/lingo.cpp
djsrv: Moved to a new file.

engines/director/lingo/lingo-preprocessor.cpp
djsrv: Changed back to the previous implementation with a bounds check on the input.

PR #3159, opened 2021-07-15
Reinstate difficulty selection for the MI2 Mac that doesn't have it

eriktorbjorn:
There are currently two known versions of Monkey Island 2 for the Mac. One was distributed (along with Monkey Island 1 and Fate of Atlantis) on CD as part of the LucasArts Mac CD Game Pack II. The other was presumably distributed on floppies.

For the CD version, the boot script was modified to remove the copy protection screen. Unfortunately, this also removed the difficulty selection screen, and I happen to think that Monkey 2 Lite is actually a fine game. So this pull request patches the boot script to match the older (?) version.

(A different approach to the problem can be found in https://github.com/eriktorbjorn/scummvm/tree/mi2-mac-mode but I like this method better.)

While doing this, I noticed that the Mac version has an extra boot param that skips the copy protection screen but leaves the difficulty selection screen. I'm surprised they didn't use that instead. I've decided to take advantage of that, since ScummVM skips the copy protection check the same was as the DOS version anyway.

Naturally, I'm pretty paranoid about modifying scripts like this so I've taken a number of precautions:

- The boot script is only patched if its length matches that of the CD version's. Not only that, but the patched script has to have the same MD5 sum as the floppy version's, or the patch is discarded.
- The boot param is only used if the boot script (which at this point may have been patched) has the expected length and MD5 sum.

I've played through the game in both modes, and I didn't notice any problems related to my changes. (There is some missing music towards the end, but it's missing when I run it in a Mac emulator as well. Weird.) I've made some refinements along the way, but nothing that would affect the patched script itself.


eriktorbjorn:
By the way, the games [*] in the first LucasArts Mac CD Game Pack (Monkey Island 1, Indiana Jones and the Last Crusade and Loom) all have copy protection, even when running in a Mac emulator. None of the games in the second collection do. I only have the CDs, so I don't know what kind of manuals were originally included.

[*] Well, except Pipe Dream, but that one doesn't count.

engines/scumm/resource.cpp
```diff
@@ -1661,6 +1666,85 @@ void ScummEngine::applyWorkaroundIfNeeded(ResType type, int idx) {
...
+0x28, 0x00, 0x00, 0xf6, 0xff,			//     unless (!VAR_RESULT) goto 0955;
+						// }
+0x48, 0x00, 0x40, 0x3f, 0xe1, 0x1d, 0x00,	// if (Local[0] == -7873) [
+0x1a, 0x32, 0x00, 0x3f, 0x01,			//     VAR_MAINMENU_KEY = 319;
+0x33, 0x03, 0x00, 0x00, 0xc8, 0x00,		//     SetScreen(0,200);
+0x0a, 0x82, 0xff,				//     startScript(130,[]);
+0x80,						//     breakHere();
+0x68, 0x00, 0x00, 0x82,				//     VAR_RESULT = isScriptRunning(130);
+0x28, 0x00, 0x00, 0xf6,	0xff,			//     unless (!VAR_RESULT) goto 0955;
+0x1a, 0x00, 0x40, 0x00, 0x00			//     Local[0] = 0;
```
bluegr, line 1704: Spacing

bluegr:
Nice work! +1 from me

sev-:
Thank you, merging. 

Please, also mention this in the NEWS.md file.

PR #3160, opened 2021-07-15
AGI: Modify Volume to use ScummVM GUI as maximum range for Script Volume

digitall:
This change simplifies the AGI engine volume handling by using the music
and sound fx volumes set in the ScummVM GUI as the maximum levels which
can be set by the scripts. This avoids the complexity of synchronising
changes between these and the scripts (which only had a single 16 step
volume range which is inverted in meaning by quite a large number of
fangames).

This still allows the game scripts to adjust the volumes, but only within
the maximum range set in the Launcher / GMM.

This avoids issues such as https://bugs.scummvm.org/ticket/10732 where
users complained about extremely loud output in some cases.

### DIFF
--- a/engines/agi/agi.cpp
+++ b/engines/agi/agi.cpp
@@ -175,8 +175,7 @@ int AgiEngine::agiInit() {
 	// Reset in-game timer
 	inGameTimerReset();
 
-	// Sync volume settings from ScummVM system settings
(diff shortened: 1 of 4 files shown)


digitall:
See also https://bugs.scummvm.org/ticket/9829 ...

sev-:
Thank you, great improvement.

PR #3164, opened 2021-07-16
CI: Use ccache on github actions

orgads:


bluegr:
Nice!
Here's a reference for the action:
https://github.com/hendrikmuhs/ccache-action

Merging this, hopefully it'll decrease build times via github actions significantly

PR #3165, opened 2021-07-16
COMMON: Change TransMan builtin language to "en" instead of "C"

criezy:
The change in this pull request change `Common::TranslationManager` so that `getCurrentLanguage` and `getLangById` both return "en" instead of "C" for the built-in language.

This seems more logical to me and allows to simplify some code (which is also included in this pull request).
Advantages of this change include:
 - Better readability

For example
```c++
#ifdef USE_TRANSLATION
	Common::String cl = TransMan.getCurrentLanguage();
#else
	Common::String cl("en");
#endif
```
makes more sense than
```c++
#ifdef USE_TRANSLATION
	Common::String cl = TransMan.getCurrentLanguage();
#else
	Common::String cl("C");
#endif
```

 - No need for special cases

We can remove the following in various places
```c++
	if (currentLanguage == "C")
		currentLanguage = "en";
```

 - More language id in themes:

If we ever want to use a custom font for English in a theme, it makes more sense to write
```
<language id = 'en'  scalable_file = 'foo.ttf'/>
```
than
```
<language id = 'C'  scalable_file = 'foo.ttf'/>
```

If we want to provide an English translation that differs from the hardcoded strings in the source code, I would expect it to be something like "en_GB" or "en_ZA" for example. So that would still be possible with this change.

Also current config files are still correctly loaded as `TransMan.parseLanguage("C")` still returns `kTranslationBuiltinId` (as it does for any unrecognised language, such as "en").

### DIFF
--- a/backends/platform/ios7/ios7_osys_misc.mm
+++ b/backends/platform/ios7/ios7_osys_misc.mm
@@ -39,10 +39,14 @@ static inline void execute_on_main_thread_async(void (^block)(void)) {
 }
 
 Common::String OSystem_iOS7::getSystemLanguage() const {
-	NSString *locale = [[NSLocale currentLocale] localeIdentifier];
-	if (locale == nil)
+	NSString *language = [[NSLocale preferredLanguages] firstObject];
+	if (language == nil)
 		return Common::String();
-	return Common::String([locale cStringUsingEncoding:NSISOLatin1StringEncoding]);
+	Common::String lang([language cStringUsingEncoding:NSISOLatin1StringEncoding]);
+	// Depending on the iOS version this may use an underscore (e.g. en_US) or a
+	// dash (en-US). Make sure we always return one with an underscore.
+	Common::replace(lang, "-", "_");
+	return lang;
 }
 
 bool OSystem_iOS7::hasTextInClipboard() {
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -565,16 +565,17 @@ Common::String OSystem_SDL::getSystemLanguage() const {
 	// Detect the language from the locale
 	if (locale.empty()) {
 		return BaseBackend::getSystemLanguage();
+	} else if (locale == "C" || locale == "POSIX") {
+		return "en_US";
 	} else {
 		int length = 0;
 
-		// Strip out additional information, like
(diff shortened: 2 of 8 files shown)


criezy:
I am not sure if I missed something and if there was a good reason to return "C", which is why I created this pull request. If you see a possible issue with this change, please speak up.

As indicated by SupSuper on Discord, the "C" almost certainly comes from the "C" locale. But `TranslationManager::setLanguage()` is expecting an ISO language_COUNTRY code, not a full locale. It for example wants `fr_FR`, not `fr_FR.UTF8`. So using a locale name for the built-in case seems strange to me.

I just saw that `OSystem::getSystemLanguage()` is actually documented as returning a POSIX locale, but the various implementations are explicitly only getting the `language_COUNTRY` part of it and stripping anything else. And I double checked that `TranslationManager` would indeed not properly handle cases were an encoding is included. So we might want to update the `OSystem::getSystemLanguage()` documentation as well to clarify that point.

lephilousophe:
What will happen if C locale is explicitly set by the user before running ScummVM?
For me, on SDL Linux build, it would return "C".
Shouldn't this be patched to change POSIX and C locales to en?

criezy:
> What will happen if C locale is explicitly set by the user before running ScummVM?
> For me, on SDL Linux build, it would return "C".
> Shouldn't this be patched to change POSIX and C locales to en?

You mean if `OSystem::getSystemLanguage()` returns "C" or "POSIX"? With the changes in this pull request it still works as before as `TranslationManager::setLanguage()` sets the languages to the builtin language (i.e. English) if the given string does not match the language for an existing translation. So passing "C", "en", or "foobar" would all have the same effect of selecting English.

But if we change the documentation of `OSystem::getSystemLanguage()` to say it should return a `language` or `language_COUNTRY` code, then yes, I think we should also change its implementation to be consistent (and return "en" instead of "C" or "POSIX".

sev-:
This makes all sense to me and in the case of Linux, I'd modify `OSystem::getSystemLanguage()` behaviour.

criezy:
I added a commit to change the implementations and documentation for `OSystem::getSystemLanguage()`. 

I will leave this open for a few more days in case other developers want to check this additional commit, and merge the pull request mid-week.

criezy:
I will merge this now.

PR #3169, opened 2021-07-19
SCUMM: Add support for Mac Loom low quality music

eriktorbjorn:
This is guesswork, but it seems like the first channel always contains a complete rendition of the music. I think that's intended for low-end Macs, while the other voices (four, but the last one never seems to be used) is the standard version of the music.

It doesn't sound particularly good, and also not quite as with Mini vMac (I'm not fully prepared to rule out emulation problems), but it seems reasonably close.

So this is basically the Mac version of PC Speaker music, and I'm adding it as a curiosity. The pull request is mostly to get feedback on the name/description of the engine option and the way it is added, since I've never done that before.

### DIFF
--- a/engines/scumm/detection.cpp
+++ b/engines/scumm/detection.cpp
@@ -197,7 +197,7 @@ static const ExtraGuiOption macV3LowQualityMusic = {
 	_s("Play simplified music"),
 	_s("This music was presumably intended for low-end Macs, and uses only one channel."),
 	"mac_v3_low_quality_music",
-	false,
+	false
 };
 
 const ExtraGuiOptions ScummMetaEngineDetection::getExtraGuiOptions(const Common::String &target) const {
--- a/engines/scumm/detection.cpp
+++ b/engines/scumm/detection.cpp
@@ -204,18 +204,24 @@ const ExtraGuiOptions ScummMetaEngineDetection::getExtraGuiOptions(const Common:
 	ExtraGuiOptions options;
 	// Query the GUI options
 	const Common::String guiOptionsString = ConfMan.get("guioptions", target);
+	const Common::String gameid = ConfMan.get("gameid", target);
+	const Common::String extra = ConfMan.get("extra", target);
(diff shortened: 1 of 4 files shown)


eriktorbjorn:
I've made recordings of the Loom intro music to demonstrate the difference:

http://www.update.uu.se/~d91tan/tmp/loom-full.ogg
http://www.update.uu.se/~d91tan/tmp/loom-simple.ogg

When running it in a Mac emulator, the simple music plays long notes for the first bar, but I have no idea why. Maybe the emulation is bad, or there is some detail that we're not implementing. The rest sounds reasonably similar though.

engines/scumm/detection.cpp
```diff
@@ -193,6 +193,12 @@ static const ExtraGuiOption fmtownsTrimTo200 = {
 	false
 };
 
+static const ExtraGuiOption macV3LowQualityMusic = {
+	_s("Play simplified music"),
+	_s("This music was presumably intended for low-end Macs, and uses only one channel."),
+	"mac_v3_low_quality_music",
+	false,
```
sev-, line 200: This will fail on older compilers, please remove the comma at end of the list

engines/scumm/detection.cpp
```diff
@@ -209,6 +215,9 @@ const ExtraGuiOptions ScummMetaEngineDetection::getExtraGuiOptions(const Common:
 	if (target.empty() || (ConfMan.get("platform", target) == "fmtowns" && guiOptions.contains(GUIO_TRIM_FMTOWNS_TO_200_PIXELS))) {
 		options.push_back(fmtownsTrimTo200);
 	}
+	if (target.empty() || (ConfMan.get("gameid", target) == "loom" && ConfMan.get("platform", target) == "macintosh" && ConfMan.get("extra", target) != "Steam")) {
```
sev-, line 218: Maybe put "gameid" and "platform" to variables for better code readability?

sev-:
I love this addition!

sev-:
Thank you!

PR #3167, opened 2021-07-18
SYMBIAN: Minor fixes for comments and code.

huhungahuru:
Most of the corrections are from comments, a minor fix has been made to the code (Only a few lines were ordered).
These fixes apply to most Symbian files.

PR #3170, opened 2021-07-19
SYMBIAN: Minor fixes for comments and code

digitall:
This is a fixed up version of #3167 with a cleaned up history, squashed commits together for same file and split the comment cleanup for the SDL, LOG and KEYMAPPER to separate commits. This should be much easier and clearer to review.


digitall:
@huhungahuru : If you check, the final result of this branch should be even with your original code in terms of changes.

digitall:
@fedor4ever : As you are the Symbian porter, any comment on these changes? OK with you?

digitall:
Most of these changes seem to be just adding full stops to the end of various comments. I don't know if this is desired or required, since https://wiki.scummvm.org/index.php?title=Code_Formatting_Conventions#Use_common_sense ... While some of this formatting is reasonable, generating a large amount of formatting churn can make tracing code changes harder later so not sure about this. @sev- : Any thoughts?

Also, I don't think our commit message guidelines use the full stop at the end. While not critical, I can filter this out to the standard form using rebase before committing this branch: https://wiki.scummvm.org/index.php?title=Commit_Guidelines#Commit_message_formatting

digitall:
Particularly, I think the "fixes" to the vsnprintf header in c247138625ffdf1f8f1b437fe10726a0678d90f6 should be removed as this header is imported code and cleaning it up from the upstream original unless critically required makes it much harder to track any changes made to it...

fedor4ever:
No ok.

fedor4ever:
You mixed all backends code.
Edit .pl scripts are waste of time. They are gone as port can selfbuld and upload to sourceforge.

fedor4ever:
in  backends/platform/symbian/help/build_help.mk  should be- `copy ScummVM.hlp $(EPOCROOT)epoc32\$(PLATFORM)\c\resource\help`. It point to SDK directory.

fedor4ever:
backends/platform/symbian/res/scummvm.rss - remain as is:
`RESOURCE RSS_SIGNATURE
	{
	}`

digitall:
@fedor4ever : I assume that "No ok" means that you are not happy with these changes as-is. I am sorry, but it is quite hard to understand you at times. I don't understand what you mean by "You mixed all backends code.". This code has been submitted by @huhungahuru so my only contribution was tidying up the commits into a set of logical changesets rather than a big bucket of commits.

digitall:
@fedor4ever : If you have specific comments, it is best to put those in line comments on the specific commits as it otherwise requires quite a bit of hunting for the exact code you are referring to:
https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request

huhungahuru:
I think the best option is to cancel this request for changes. I apologize for the problems caused.

sev-:
@fedor4ever Please, be constructive with your comment or do not comment at all!

@huhungahuru thank you for your contribution. However, I would recommend splitting this into several PRs and squash similar ones. E.g. there is a number of commits that are adding full stops and fixing grammar, those could be committed as a single commit.

Any functional changes must go separately.

Apologies, that it creates a bit more work, but we need to be careful and it is not easy to review such big PRs.

fedor4ever:
backends/platform/symbian/src/SymbianMain.cpp: has no sense - `// This line was added, we still need to test (07/17/2021).`

digitall:
@huhungahuru : No problem. I would suggest doing a quick git course and learning about feature branches since they make it much easier to avoid merge commit issues on your master / main branch:
https://gist.github.com/vlandham/3b2b79c40bc7353ae95a

To echo @sev- comment, it is just hard to review when it is not broken into logical changesets. Also formatting fixes are usually best if they can be produced from an automated tool so it can be shown to be repeatable i.e. http://astyle.sourceforge.net/

digitall:
@fedor4ever : Since you have quite a few objections to the changes in this PR and you are the Symbian porter anyway, I am going to close this PR and let you cherry-pick (@huhungahuru : another good git tool https://git-scm.com/docs/git-cherry-pick ) any changes which you think are useful or relevant.

huhungahuru:
@fedor4ever : I will make suggestions for changes in my own repository and if the changes convince you, you can add them. Hope I can help with the Symbian port.

fedor4ever:
@huhungahuru, great job. But this task should be splitted on several pull request. Several parts I can accepted as is.
@sev- , I try my best to be constructive.
@digitall , thanks for docs.

PR #3174, opened 2021-07-20
SYMBIAN: Add engines layout for app creation.

fedor4ever:
This file contains rules for placing engines to predefined exe.


sev-:
Merging

PR #3175, opened 2021-07-20
SYMBIAN: Fix race condition in build scripts.

fedor4ever:
Add cleanup from results previous build before new start.
Code style fixes from #3170.
Forced CPU usage.


sev-:
Merging

PR #3176, opened 2021-07-20
CGE: Small adjustments to Text To Speech (TTS) feature

taylorzhancher:
Some adjustments to the way TTS is done
- remove unnecessary header
- move `void textToSpeech(const char *text)` to talk engine
- adjust tooltip description in detection.cpp
- adjustments to the function `void textToSpeech(const char *text)`


taylorzhancher:
I'm creating a new branch to do this instead

PR #3177, opened 2021-07-20
CGE: Change declaration place and descriptions

taylorzhancher:
Some adjustments to the way TTS is done

- remove unnecessary header
- move void textToSpeech(const char *text) to talk engine
- adjust tooltip description in detection.cpp
- adjustments to the function void textToSpeech(const char *text)

### DIFF
--- a/engines/cge/cge.h
+++ b/engines/cge/cge.h
@@ -34,7 +34,6 @@
 #include "cge/console.h"
 #include "cge/bitmap.h"
 #include "cge/sound.h"
-#include "common/text-to-speech.h"
 
 struct ADGameDescription;
 
--- a/engines/cge/detection.cpp
+++ b/engines/cge/detection.cpp
@@ -117,7 +117,7 @@ static const ADExtraGuiOptionsMap optionsList[] = {
 		GAMEOPTION_TTS,
 		{
 			_s("Enable Text to Speech"),
-			_s("Use TTS to read the descriptions (if TTS is available)"),
+			_s("Use TTS to read text in the game (if TTS is available)"),
 			"tts_enabled",
 			false
 		}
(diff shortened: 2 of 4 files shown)


criezy:
I rebased the branch on latest master for you to fix the conflict. Your branch was created on a 3-days old master that did not include the last commit from digitall on the CGE engine.

Since I was experimenting with the code, I also pushed a further commit with some changes. This was faster than commenting on the pull request to ask you to do those, but I encourage you to look at the commit as you might want to do similar changes for the CGE2 pull request. Also don't hesitate to ask questions if the purpose of some of those changes os not clear to you.
The changes include:
 - The `textToSpeech(const char *)` did not need to be virtual since you do not need to reimplement it in classes that derive from `Talk` (unlike `update(const char *)` that is reimplemented in `InfoLine`).
 - I moved the `_lastText` variable in the `InfoLine` class.
 - I moved the call to `TextToSpeech()` from the `Talk` constructor to the `update()` function. This change was done for consistency with what is done in the derived `InfoLine` class.
 - I added a check that the text is not a null pointer in `textToSpeech(). This avoids cutting off the speech immediately when moving the cursor over an object and not stoping on it. I think it sounds nicer that way.

Looking at this I also understood why the `_oldText` check already in place was not sufficient for the TTS and you needed to introduce a `lastText` variable. I actually suspect this is a bug, but I did not want to change the code without more tests. So for now both `_oldText` and `_lastText` are still there. You can see the comment I added in the code to explain this.

criezy:
> Looking at this I also understood why the `_oldText` check already in place was not sufficient for the TTS and you needed to introduce a `lastText` variable. I actually suspect this is a bug, but I did not want to change the code without more tests. 

I have now done more test and more digging and I am almost certain it was indeed a bug in the original source code. I pushed an additional commit with a fix.

criezy:
I see no reason to wait more for this one, so merging now.

PR #3178, opened 2021-07-20
GRAPHICS: Implement horizontal triangles

lephilousophe:
Everything is in the title.

Simplify drawTriangle in the same time and fix a bad copy paste when
drawing outline with w == h

### DIFF
--- a/graphics/VectorRendererSpec.cpp
+++ b/graphics/VectorRendererSpec.cpp
@@ -501,6 +501,54 @@ void colorFillClip(PixelType *first, PixelType *last, PixelType color, int realX
 	}
 }
 
+/**
+ * Fills several pixels in a column with a given color.
+ *
+ * @param first Pointer to the first pixel to fill.
+ * @param last Pointer to the last pixel to fill.
+ * @param pitch Number of pixels in a line.
+ * @param color Color of the pixel
+ */
+template<typename PixelType>
+void colorVFill(PixelType *first, PixelType *last, int pitch, PixelType color) {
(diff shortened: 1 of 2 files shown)


lephilousophe:
Forgot to add that I copy pasted from vertical code and changed x by y and so on everywhere.
Some variable names seem bogus to me but they are in the vertical case too so I decided to keep the same logic.

Thunderforge:
Is there some reason we need to draw triangles instead of just using ► or ▶ (unicode characters [U+25BA](https://www.fileformat.info/info/unicode/char/25ba/index.htm) and [U+25B6](https://www.fileformat.info/info/unicode/char/25b6/browsertest.htm) respectively)?

criezy:
> Is there some reason we need to draw triangles instead of just using ► or ▶ (unicode characters [U+25BA](https://www.fileformat.info/info/unicode/char/25ba/index.htm) and [U+25B6](https://www.fileformat.info/info/unicode/char/25b6/browsertest.htm) respectively), which are supposed to be rendered as black triangles (GitHub seems to substitute them with emojis instead).

I can see a few possible reasons, but one of them is that ScummVM can be compiled without freetype support and use BDF fonts, which will not have these characters (and also would not scale well even if they had them).

lephilousophe:
The drawTriangle primitive is more flexible than the Unicode glyphs, you can set the width and height independently. That may be used by some themes out there.

lephilousophe:
Just fixed Codacy issue

sev-:
Thank you!

PR #3183, opened 2021-07-21
SYMBIAN: Can find and use game data placed by installer on any drive.

fedor4ever:
Apply style fixes from #3170.


sev-:
I am merging this as is, but generally  it is not a good idea to mix functional and non-functional changes in one commit.

PR #3184, opened 2021-07-21
SYMBIAN: Avoid deletion argv if no cmdlne args set.

fedor4ever:
That fix multiple User 42 panics at exit.
Apply style fixes from #3170.

### DIFF
--- a/backends/platform/symbian/src/SymbianMain.cpp
+++ b/backends/platform/symbian/src/SymbianMain.cpp
@@ -98,7 +98,7 @@ int main(int argc, char *argv[]) {
 	if (argc > 1) {
 		debug("console arg count by CCommandLineArguments: %d", argc);
 		argv = new char* [argc];
-		HBufC8 *buf = HBufC8::NewMax(20); // This suffice for most cases.
+		HBufC8 *buf = HBufC8::NewMax(20); // This should suffice for most cases.
 		for (TInt i = 0; i < argc; ++i) {
 			TPtrC arg = cmdline->Arg(i);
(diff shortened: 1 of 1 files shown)


backends/platform/symbian/src/SymbianMain.cpp
```diff
@@ -117,14 +116,16 @@ int main(int argc, char *argv[]) {
 	// Invoke the actual ScummVM main entry point:
 	int res = scummvm_main(argc, argv);
 
-	// clear argv
-	for (TInt i = 0; i < argc; ++i) {
-		delete[] argv[i];
+	// Clear argv.
+	if(argc > 1) { //avoid multiple User 42 panics at exit
```
bluegr, line 120:
```suggestion
	if (argc > 1) { //avoid multiple User 42 panics at exit
```

backends/platform/symbian/src/SymbianMain.cpp
```diff
@@ -75,34 +74,34 @@ int main(int argc, char *argv[]) {
...
 		return -1;
 	}
-	
+
 	argc = cmdline->Count();
 	if (argc > 1) {
 		debug("console arg count by CCommandLineArguments: %d", argc);
 		argv = new char* [argc];
-		HBufC8 *buf = HBufC8::NewMax(20); //this suffice for most cases
+		HBufC8 *buf = HBufC8::NewMax(20); // This suffice for most cases.
```
bluegr, line 101:
```suggestion
		HBufC8 *buf = HBufC8::NewMax(20); // This should suffice for most cases.
```

bluegr:
Nice work, thanks! Merging (squashing to include the style fixes in the commit)

PR #3187, opened 2021-07-21
CI: Enable fribidi

orgads:
Issues in vcpkg are already resolved.

### DIFF
--- a/devtools/create_project/msvc.cpp
+++ b/devtools/create_project/msvc.cpp
@@ -59,7 +59,7 @@ std::string MSVCProvider::getLibraryFromFeature(const char *feature, const Build
 		{      "sdl2", "SDL2.lib",                  "SDL2d.lib",     "winmm.lib imm32.lib version.lib setupapi.lib",    0 },
 		{      "libz", "zlib.lib",                  "zlibd.lib",     0,                                                 0 },
 		{       "mad", "mad.lib",                   0,               0,                                                 "libmad.lib" },
-		{   "fribidi", "libfribidi.lib",            0,               0,                                                 0 },
+		{   "fribidi", "fribidi.lib",               0,               0,                                                 0 },
 		{       "ogg", "ogg.lib",                   0,               0,                                                 "libogg_static.lib" },
 		{    "vorbis", "vorbis.lib vorbisfile.lib", 0,               0,                                                 "libvorbisfile_static.lib libvorbis_static.lib" },
 		{      "flac", "FLAC.lib",                  0,               0,                                                 "libFLAC_static.lib win_utf8_io_static.lib" },
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,16 +13,14 @@ jobs:
           - platform: win32
             triplet: x86-windows
             arch: x86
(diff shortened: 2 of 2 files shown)


devtools/create_project/msvc.cpp
```diff
@@ -59,7 +59,7 @@ std::string MSVCProvider::getLibraryFromFeature(const char *feature, const Build
 		{      "sdl2", "SDL2.lib",                  "SDL2d.lib",     "winmm.lib imm32.lib version.lib setupapi.lib",    0 },
 		{      "libz", "zlib.lib",                  "zlibd.lib",     0,                                                 0 },
 		{       "mad", "mad.lib",                   0,               0,                                                 "libmad.lib" },
-		{   "fribidi", "libfribidi.lib",            0,               0,                                                 0 },
+		{   "fribidi", "fribidi.lib",               0,               0,                                                 "libfribidi.lib" },
```
orgads, line 62: I'm not sure if `libfribidi.lib` is needed. The `vcpkg` library is `fribidi.lib`. Opinions?

orgads: @lephilousophe How does it work on buildbot?

lephilousophe: It works without create_project because we use mingw and configure script so it's not really relevant. We use `-lfribidi` and linker must find libfribidi.a.

orgads: Ah, I see. Thanks.

So can I remove the libfribidi.lib?

SupSuper: I assume vcpkg changed something that renamed libfribidi to fribidi? I'll have to update the prebuilt libraries.

orgads: Has it ever worked?

devtools/create_project/msvc.cpp
```diff
@@ -59,7 +59,7 @@ std::string MSVCProvider::getLibraryFromFeature(const char *feature, const Build
 		{      "sdl2", "SDL2.lib",                  "SDL2d.lib",     "winmm.lib imm32.lib version.lib setupapi.lib",    0 },
 		{      "libz", "zlib.lib",                  "zlibd.lib",     0,                                                 0 },
 		{       "mad", "mad.lib",                   0,               0,                                                 "libmad.lib" },
-		{   "fribidi", "libfribidi.lib",            0,               0,                                                 0 },
+		{   "fribidi", "fribidi.lib",               0,               0,                                                 0 },
```
sluicebox, line 62: Won't this break the build for visual studio users using the precompiled libraries? (Which I believe is, practically all the visual studio users.)

orgads: Which precompiled libraries? The `vcpkg` library is named `fribidi.lib`. Are you aware of any distribution that has it as `libfribidi.lib`?

sluicebox: Yes, the one we provide: https://wiki.scummvm.org/index.php?title=Compiling_ScummVM/Visual_Studio

orgads: Done

PR #3188, opened 2021-07-22
SYMBIAN: Add test UID app for ScummVm.hrh generator.

fedor4ever:


sev-:
There is something wrong with this commit, looks like the changes were already applied

PR #3190, opened 2021-07-22
CI: Change Windows builds to release mode

orgads:
Use a symbolic link for building on drive C. Github runners have ~87G free
on drive C, and only 13G on D. For some reason, release build consumes much
more space, so we run out of space when building on D.


orgads:
Nice. Down to 30-40 minutes for Windows. @SupSuper?

SupSuper:
Can you compare with Debug builds? That will probably be faster than Release and not require messing with disk space.

orgads:
https://github.com/scummvm/scummvm/pull/3195

Took 20 minutes. Looks great!

PR #3192, opened 2021-07-24
MAKE: added target to build via cmake

mgerhardy:
This allows easy `CMakeLists.txt` creation. I would like to check the cmake build against ninja and make as well as the unity build versus the normal build. These targets were produced to ease the build dir creation and compilation. Maybe others find them useful, too.


sev-:
Thanks!

PR #3197, opened 2021-07-25
RFC: SCUMM: Workaround for bug #12734

eriktorbjorn:
This is my attempt at reinstating a line of text in Monkey Island 2 that is printed, then immediately erased, after Rapp gives you his map piece. The comments in the commit and on bug #12734 should provide the necessary background.

I've given it some testing, but I don't know how much time I will have for ScummVM in the next few weeks. And it's hardly the only way of doing it, so any feedback is appreciated.

My main concern was that I want the animation of Rapp turning back to ash to play without any further delay. Rapp doesn't care what you're talking about. Admittedly, it may make Guybrush seem a bit less caring, too...

### DIFF
--- a/engines/scumm/string.cpp
+++ b/engines/scumm/string.cpp
@@ -55,6 +55,23 @@ namespace Scumm {
 void ScummEngine::printString(int m, const byte *msg) {
 	switch (m) {
 	case 0:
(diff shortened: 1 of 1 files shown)


sev-:
The change is very well isolated, so I see no harm in merging it now. Thank you!

PR #3200, opened 2021-07-25
CHEWY: Add Spanish translation

aviloria:
Add Spanish fanmade translation


bluegr:
Thanks! The engine is still in a very early stage, so don't expect the game to be working yet - at all

In any case, more detection entries won't hurt. Merging

PR #3201, opened 2021-07-25
AGS: Savegames for Strangeland

dreammaster:
As I mentioned on Discord, the commercial AGS game Strangeland uses 300Mb+ save files. Currently, savegames are written to an intermediate memory buffer to allow for seeking to be done backwards to fill out chunk lengths and the offset of the screenshot, and not all backends support seekable write streams yet. But for Strangeland, it needs to be able to write directly to the disk, since for Visual Studio at least, the memory allocator throws a wobbly once I try to allocate more than 150Mb of memory at once.

This pull request has two commits. The first is changing OutSaveFile to derive from SeekableWriteStream rather than just WriteStream. This introduces two methods: seek and pos. In case a given backend doesn't support it, it keeps the internal _wrapped as a WriteStream, and does a dynamic cast to SeekableWriteStream in the seek and pos methods. This allows it to gracefully fail when the wrapped stream doesn't support the methods.

The second commit builds on this with some changes to the AGS FileStream class. In it, when the save file is opened, it tries to do a size() call on it. If the save file's underlying stream doesn't support seeking, size() will return -1, allowing the save code to fall back on using a memory buffer.. with an explicit error if the game being run is Strangeland.

### DIFF
--- a/backends/saves/savefile.cpp
+++ b/backends/saves/savefile.cpp
@@ -29,7 +29,8 @@
 
 namespace Common {
 
-OutSaveFile::OutSaveFile(WriteStream *w): _wrapped(w) {}
+OutSaveFile::OutSaveFile(WriteStream *w, bool isCompressed):
+		_wrapped(w), _isCompressed(isCompressed) {}
 
 OutSaveFile::~OutSaveFile() {
 	delete _wrapped;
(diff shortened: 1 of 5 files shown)


engines/ags/shared/util/file_stream.cpp
```diff
@@ -204,15 +206,28 @@ void FileStream::Open(const String &file_name, FileOpenMode open_mode, FileWorkM
...
+		// for backtracking to write out chunk sizes and screenshot offset
+		if (_outSave->size() == 0) {
+			// Supports it, so data can be written directly
+			_file = _outSave;
+			_outSave = nullptr;
+		} else {
+			// It doesn't. So data needs to be written to a memory buffer
+			// to allow for seeking, and will be dumped all at once at the end
+			if (::AGS::g_vm->getGameId() == "strangeland")
+				error("Your platform doesn't support savegame seeking, " \
```
sev-, line 226: Shouldn't this be made user-visible, e.g. `GUI::MessageBox()` and translateable? If it is the latter, POTFILES needs to be modified too.

backends/saves/savefile.cpp
```diff
@@ -56,6 +56,18 @@ int64 OutSaveFile::pos() const {
 	return _wrapped->pos();
 }
 
+bool OutSaveFile::seek(int64 offset, int whence) {
+	Common::SeekableWriteStream *sws =
+		dynamic_cast<Common::SeekableWriteStream *>(_wrapped);
```
sev-, line 61: I don't quite understand this change. All of our `createWriteStream()` calls (used by `DefaultSaveFileManager::openForSaving()` are returning `Common::WriteStream`. How this cast could ever be executed?

dreammaster: As I understand it (being at work, and away from PC), the OpenForSaving by default uses the StdioStream, which was previously changed to derive from SeekableWriteStream back when SeekableWriteStream was first introduced. So for Windows (and presumably Mac and Linux and others that use that standard implementation) the _wrapped writestream can be cast to SeekableWriteStream. It's just any backend, if any, that uses an alternative saving logic that would fail the cast.

dreammaster:
Since last night I verified that the latest commit for MemoryWriteStreamDynamic allows Strangeland to save without crashing, it's less critical to have these changes. On reflection, I'm leaning towards simply closing this PR without merging.. I don't want to get into a situation where other engines in the future start trying to use seek on opened savefiles without realizing that it only works on some systems.

Far better, if ever needed in the future, to have a proper implementation that ensures all the backends support seeking (when creating an uncompressed file). Do you concur?

sev-:
I think seekable saves are pretty convenient. But for that, I would add a big fat warning() to the devs that they must disable the compression in order to be able to seek().

dreammaster:
@sev- - I'm deprecating this pull request in favor of a cleaner implementation that adds seekable write streams to all the backends in: https://github.com/scummvm/scummvm/pull/3208

PR #3202, opened 2021-07-27
RFC: [DO NOT MERGE!] SCUMM: Generate ASCII 1, 2, 3, ... for CTRL+A, B, C... (bug #10644)

eriktorbjorn:
The Monkey Island 1 input script assumes that Ctrl+A, B, C, ... will generate ASCII codes 1, 2, 3, ... This is used for Ctrl+W as a shortcut to "win" the game, for instance.

There are several places where this could be implemented, but it seemed best to limit it to the SCUMM engine. I don't know if it should be limited any further than that, though.

Thoughts? Suggestions?

### DIFF
--- a/engines/scumm/input.cpp
+++ b/engines/scumm/input.cpp
@@ -623,7 +623,29 @@ void ScummEngine::processKeyboard(Common::KeyState lastKeyHit) {
 			// Handle KBD_ALT combos in MI2. We know that the result must be 273 for Alt-W
 			// because that's what MI2 looks for in its "instant win" cheat.
 			_mouseAndKeyboardStat = lastKeyHit.keycode + 154;
+		} else if ((_game.id == GID_MONKEY ||
+		            _game.id == GID_MONKEY_EGA ||
+			    _game.id == GID_MONKEY_VGA ||
+			    _game.id == GID_MONKEY2 ||
+		            _game.id == GID_LOOM ||
+		            _game.id == GID_PASS) &&
+		           (lastKeyHit.flags & Common::KBD_CTRL) &&
+		           lastKeyHit.ascii >= 'a' &&
+		           lastKeyHit.ascii <= 'z') {
+			// Some games games expect Ctrl+A, B, C, etc. to
+			// generate codes 1, 2, 3, etc.
+			//
+			// This is used for several settings in the "ultimate
+			// talkie" versions of Monkey Island 1 and 2. Monkey
+			// Island 1 also uses it for Ctrl+W to immediately win
+			// the game.
+			//
+			// Many (but not all) versions of Loom displays version
+			// information on Ctrl+V, as does the Passport to
+			// Adventure and Monkey Island 1 demo. If there are
+			// others, I don't know about them yet.
 
+			_mouseAndKeyboardStat = lastKeyHit.ascii & 0x1F;
 		} else if (lastKeyHit.keycode >= Common::KEYCODE_UP &&
 		          lastKeyHit.keycode <= Common::KEYCODE_LEFT) {
 			if (_game.id == GID_MONKEY && _game.platform == Common::kPlatformSegaCD) {


sev-:
Maybe limit it to DOS games?

eriktorbjorn:
> Maybe limit it to DOS games?

The Mac version of Monkey Island 1 has the same Ctrl+W shortcut. (Though I'm uncertain exactly if the Mac keyboard had a Ctrl key.) I don't know about the Amiga version.

eriktorbjorn:
With the change, you can use Ctrl+A, Ctrl+B and Ctrl+C to switch between the three characters in Manaic Mansion. Since this is wrong (it's F1, F2, F3, which already work in ScummVM), the patch needs to be restricted a bit. I'll look some more to see what I can find in the other games.

eriktorbjorn:
The EGA, VGA and FM Towns versions of check for ASCII code 22 and prints version information. That's presumably meant to be Ctrl+V, and works with the patch.

The Mac and PC-Engine versions do not, as far as I can tell.

criezy:
> Though I'm uncertain if the Mac keyboard had a Ctrl key, or just some generic options key.

I am not sure about earlier models, but the keyboard that came with my Macintosh Classic in 1990 has both an option and a control key.

eriktorbjorn:
The "Passport to Adventure" and Monkey Island 1 demos also handle Ctrl+V to print version information.

eriktorbjorn:
Since it now looks like it has to be special-cased for just the games where it's known to be used, I've moved the workaround to the spot in the code where we already special case Alt+whatever for Monkey Island 2.

eriktorbjorn:
Note that I've hardly checked any of the Humongous games, since there's an awful lot of them and I just can't be bothered to figure out the proper descumm parameters for them.

eriktorbjorn:
Or maybe it should be dependent on SCUMM version? In which case these game span versions 3-5.

sev-:
I checked with the HE sources and they make no special handling.

Thus, you do not need to filter out this by specific games as it seems to be the common approach in the SCUMM engine.

My initial comment was regarding Windows, apologies for not being explicit, but seems that even that filtering is not required.

eriktorbjorn:
Well, _some_ filtering is probably needed. Because, as noted above, apparently Maniac Mansion uses 1, 2, 3 for F1, F2, F3, and I don't want to make Ctrl+A, B and C behave like that.

I seem to recall the Ctrl+W shortcut in MI1 used to work in older versions of ScummVM, but I haven't investigated if that's true or, in that case, how it used to work.

engines/scumm/input.cpp
```diff
@@ -623,7 +623,29 @@ void ScummEngine::processKeyboard(Common::KeyState lastKeyHit) {
 			// Handle KBD_ALT combos in MI2. We know that the result must be 273 for Alt-W
 			// because that's what MI2 looks for in its "instant win" cheat.
 			_mouseAndKeyboardStat = lastKeyHit.keycode + 154;
+		} else if ((_game.id == GID_MONKEY ||
+		            _game.id == GID_MONKEY_EGA ||
```
mgerhardy, line 627: The formatting is off here.

eriktorbjorn:
When I tried some older versions of ScummVM, I found that in 1.8.1 the Ctrl+W cheat works in Monkey Island 1. In 1.9.0 it does not.

It's probably not a coincidence that 1.8.1 was built against SDL 1.2, while 1.9.0 was built against SDL 2.0.

Perhaps that's a clue for someone who knows more about that transition than I do?

eriktorbjorn:
Oh, and in 1.8.1 it turns out I _can_ use Ctrl+A, Ctrl+B and Ctrl+C to switch between the kids.

eriktorbjorn:
So maybe this is something that should be fixed globally, and not just in the SCUMM engine after all?

eriktorbjorn:
It looks like in 1.8.1 it's SdlEventSource::obtainUnicode() that returns 1, 2, 3 for Ctrl+A, B, C, etc. for SDL 1.2. It simply returns the "unicode" field from the SDL_keysym type. Unicode support is only enabled for SDL 2, so there is no Unicode translation, but apparently the "unicode" field gets set anyway?

In the current version, with SDL 2, obtainUnicode() returns 0 since it's not seeing any SDL_TEXTINPUT event for it. And mapKey() handles unicode values < 32 as 0 anyway.

So maybe it really should be something like this instead after all?

```
diff --git a/backends/events/sdl/sdl-events.cpp b/backends/events/sdl/sdl-events.cpp
index 9cf043ee37..44912626a8 100644
--- a/backends/events/sdl/sdl-events.cpp
+++ b/backends/events/sdl/sdl-events.cpp
@@ -159,6 +159,8 @@ int SdlEventSource::mapKey(SDL_Keycode sdlKey, SDL_Keymod mod, Uint16 unicode) {
        } else if (unicode) {
                // Return unicode in case it's still set and wasn't filtered.
                return unicode;
+       } else if (key >= 'a' && key <= 'z' && (mod & KMOD_CTRL)) {
+               return key & 0x1F;
        } else if (key >= 'a' && key <= 'z' && (mod & KMOD_SHIFT)) {
                return key & ~0x20;
        } else if (key >= Common::KEYCODE_NUMLOCK && key < Common::KEYCODE_LAST) {
```

bluegr:
I believe that the second global fix seems to be the best solution to the problem

bluegr:
#3324 has been merged, so I believe that this can be closed now

orgads:
I tested with more games, and it seems to be supported on all of them. See #3331.

PR #3203, opened 2021-07-27
SCUMM: COMI: Fix iMUSE crashes on Android

AndywinXp:
When the COMI DiMUSE engine encounters a JUMP instruction, a crossfade is executed from the current region of the track to the destination one.
I originally implemented this quirk in a very convoluted and hacky way, and guess what: it crashed the game on Android during certain app pause/resume situations (thanks for the heads up, @antoniou79 ).

I have now deleted my crossfading system and cleaned up the corresponding code: the game now just switches to the destination region without a crossfade, which is a thing that is very much rock solid and still pleasing to the ears.

I think this will suffice until I get around to implement the disasm.

### DIFF
--- a/engines/scumm/imuse_digi/dimuse.cpp
+++ b/engines/scumm/imuse_digi/dimuse.cpp
@@ -647,7 +647,7 @@ void IMuseDigital::switchToNextRegion(Track *track) {
 					// same time as a loop; the result is that the former is prioritized and the latter
 					// is executed without a crossfade. Also, avoid music crossfades for start markers, these are just plain
 					// dangerous and useless since there's already a fade in for those.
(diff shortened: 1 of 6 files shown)


antoniou79:
I have tested with the changes in this PR and the crash as reported in the following ticket does not occur anymore:
https://bugs.scummvm.org/ticket/12095

Note, that there is some extra work needed for the Android pause/resume code, which caused the crash in conjunction with the iMuse changes, but the code as is (pause/resume) worked with the iMuse code before the changes (probably with a few glitches) and works with this PR code (again with one random glitch). I will try to revisit the code as soon as I can make time for it.

If there are no objections and since this resolves a crash issue, I propose this is merged after the weekend.

AndywinXp:
Thanks!

PR #3204, opened 2021-07-27
GUI: Fix shading being applied more than once

sluicebox:
Fixes nested GUI dialogs causing background shading to be applied multiple times. This is a regression that occurred between release 1.8.0 and 1.9.0.  The GUI code tries to prevent this by testing the dialog stack count, so presumably the structure changed but this wasn't updated.

To reproduce and see another unintended effect:

1. Click Options (applies shading)
2. Go to MIDI tab
3. Click FluidSynth Settings (bug: applies more shading)
4. Go to the Chorus tab (extra shading goes away on redraw)


sev-:
Excellent, thank you!

PR #3205, opened 2021-07-28
AGS: recognize alternative name of shell plugin

mausimus:
There is a circulating version of AGS Shell plugin as agsshell.dll
used by some games (Rosewater). Internally it's the same
ags_shell.dll so simply an alias will allow these games to run.


dreammaster:
Thanks for identifying it. Merging it now.

PR #3206, opened 2021-07-28
SCUMM: FT: Fix softlock on negative walk destination

AndywinXp:
This commit fixes bug [#12030](https://bugs.scummvm.org/ticket/12030), but to do so it reverts commit [SCUMM: Fix UB shifting negative integers in Actor](https://github.com/scummvm/scummvm/commit/57084b4a1b23e131be7497c4c97c75db060210c9).

I tried comparing the calculations of these two versions of the code and they do indeed give different results when being fed negative values, and while I don't know if that's intentional or not, with my commit I'm restoring the behavior which matches the disassembly, just in case.

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -604,13 +604,13 @@ int Actor::actorWalkStep() {
 			_walkdata.yfrac -= _v3stepThreshold;
 		}
 	} else {
-		tmpX = (_pos.x * (1 << 16)) + _walkdata.xfrac + (_walkdata.deltaXFactor / 256) * _scalex;
+		tmpX = (_pos.x << 16) + _walkdata.xfrac + (_walkdata.deltaXFactor >> 8) * _scalex;
 		_walkdata.xfrac = (uint16)tmpX;
-		_pos.x = (tmpX / (1 << 16));
+		_pos.x = (tmpX >> 16);
 
-		tmpY = (_pos.y * (1 << 16)) + _walkdata.yfrac + (_walkdata.deltaYFactor / 256) * _scaley;
+		tmpY = (_pos.y << 16) + _walkdata.yfrac + (_walkdata.deltaYFactor >> 8) * _scaley;
 		_walkdata.yfrac = (uint16)tmpY;
-		_pos.y = (tmpY / (1 << 16));
+		_pos.y = (tmpY >> 16);
 	}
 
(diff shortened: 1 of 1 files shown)


athrxx:
Does this also fix your COMI ticket?
https://bugs.scummvm.org/ticket/12029

As the only thing you're doing here is reverting that commit by csnover this PR should be safe to merge. That commit isn't very clear in its commit message, but apparently it doesn't even aim at fixing any actual issue in the SCUMM engine. It seems to be part of a series of commits all over ScummVM which all claim in their commit messages to "Fix UB shifting negative integers".

AndywinXp:
> Does this also fix your COMI ticket?

No, unfortunately not, but I'm looking into that one as we speak, and I'm suspecting that bug is caused by slight calculation inaccuracies which culminate in assigning slightly wrong XY coordinates for the ships in Actor::actorWalkStep().

athrxx:
Okay, that's unfortunate. Since we currently use the same walk step calculations for all of >= SCUMM 5. Maybe there are more version differences here.

Are you looking at COMI disasm or just experimenting on the code?

AndywinXp:
> Okay, that's unfortunate. Since we currently use the same walk step calculations for all of >= SCUMM 5. Maybe there are more version differences here.
> 
> Are you looking at COMI disasm or just experimenting on the code?

I'm doing it in the least fun way possibile: debugging ScummVM and the COMI disasm side by side to see where the calculations fall apart 😆

AndywinXp:
...and solved, should I merge that commit with this PR or should I make another PR altogether (asking since I'm operating on the same chunk of code)?

For the record, this was the problem, in Actor::calcMovementFactor(), row 517, the check:
`if ((uint)ABS(deltaXFactor) > (_speedx << 16)) {`
should be
`if ((uint)ABS(deltaXFactor >> 16) > _speedx) {`

I can attest this is also true at least for FT and DIG, which are the only disasm I have right here.

athrxx:
Nice :thumbsup:  
Please put that in a separate PR, because we might want to counter check with SCUMM5/6, too, before we merge that. Unless you limit it to SCUMM 7/8.

AndywinXp:
Sure thing!

athrxx:
I think this is safe to merge, since all that happens here is the revert of a questionable commit.
And this way it can get some testing before the next release...

PR #3207, opened 2021-07-29
SCUMM: COMI: Fix bug #12029

AndywinXp:
This PR fixes bug [#12029](https://bugs.scummvm.org/ticket/12029). The problem stems from an oversight in `Actor::calcMovementFactor()`:

 `ABS(deltaXFactor) > (_speedx << 16)` is not the same as `ABS(deltaXFactor >> 16) > _speedx`, which is the correct version. This probably went unnoticed because the old calculation pipeline led to slightly wrong (by one or two pixels) values for actor's x and y when walking.

This fix was taken from the disassemblies for COMI, DIG and FT, and it appears to be working fine for old SCUMM versions (I have tried at least MI2 and DOTT).

### DIFF
--- a/engines/scumm/actor.cpp
+++ b/engines/scumm/actor.cpp
@@ -514,7 +514,7 @@ int Actor::calcMovementFactor(const Common::Point& next) {
 			deltaYFactor = 0;
 		}
 
-		if ((uint)ABS(deltaXFactor) > (_speedx << 16)) {
+		if ((uint)ABS(deltaXFactor >> 16) > _speedx) {
 			deltaXFactor = _speedx << 16;
 			if (diffX < 0)
 				deltaXFactor = -deltaXFactor;


athrxx:
I can confirm that this is correct for COMI disasm.

And basically also for MI2 (SCUMM 5). The original actually does a 0x10000 IDIV on the deltaXfactor and then gets the abs of the result. But that should always be exactly the same as the 0x10 SAR that COMI does on the abs of deltaXfactor....

SCUMM 3 code recently got separated/fixed, so that won't be affected.

It's a bit hard to predict whether it will have an impact on SCUMM1/2. We don't pretend to have accurate code for that. But we also don't really use much of the deltaX/Yfactor there...

AndywinXp:
I have just tested Maniac Mansion (enhanced, so it should be SCUMM v2) with and without the fix, and it appears to be working the same: considering the way deltaX/YFactor is used for v2 games, it doesn't appear to change the behavior at all.

athrxx:
I think this is safe to merge. I haven't done any actual game testing, but I have confirmed the changes from COMI and MI2 disasm (for v8 and v5; but I guess it is safe to assume that the versions in between will have the same behavior).

PR #3208, opened 2021-07-30
BACKENDS: Save files support seeking on all backends

dreammaster:
This is a cleaner implementation of my prior pull request for AGS Strangeland that should implement save file seeking across all the backends. Given that we've already integrated in a change to the stream class for 64-bit file support, I think it's the most opportune time to make this change as well. AGS isn't the first engine that's needed to do seeking within the save files being written out, and it's previously required hacky workarounds, like writing data to a MemoryWriteStreamDynamic first and then dumping out to the save files. Now, that will no longer be required.

The pull request is two main parts:
1) Enhancing the file system node createWriteStream() to return a SeekableWriteStream. Most of the backends instantiate a write stream directly using StdIOStream or a class that derives from it, so they didn't require much changes. As for the few remaining ones, the most common was to use a BufferedFileStream wrapper. So I enhanced the class to support SeekableWriteStream as well.
2) OutSaveFile is enhanced to derive from SeekableWriteStream as well. OutSaveFile still takes in a plain WriteStream, because the GZip compressed write stream doesn't support seeking. As with the prior PR, OutSaveFile's seek and size methods try casting the wrapped stream to a SeekableWriteStream, and only pass on the method call if that's successfully.

Finally, the AGS commit is also present to clean up the FileStream class, since the save file can be directly written to.

### DIFF
--- a/backends/fs/abstract-fs.h
+++ b/backends/fs/abstract-fs.h
@@ -190,7 +190,7 @@ public:
 	 *
 	 * @return pointer to the stream object, 0 in case of a failure
 	 */
-	virtual Common::WriteStream *createWriteStream() = 0;
+	virtual Common::SeekableWriteStream *createWriteStream() = 0;
 
 	/**
 	* Creates a directory referred by this node.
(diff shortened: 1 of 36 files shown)


bluegr:
Nice and clean implementation! However, the pull request has conflicts now. Could you please rebase your feature branch?

dreammaster:
If nobody else comments, I'll merge it in later today. Given the minimal changes, I'm not expecting any issues, and I'd like to see it integrated before we get into preparing for release.

PR #3210, opened 2021-07-30
CONFIGURE: Do not pass --prefix to sdl2-config

orgads:
sdl2-config should return the correct paths without it. Trying to determine
the path by name lookup is error-prone.

For example, if both the host and a cross-toolchain have sdl2 installed,
the host executable is typically /usr/bin/sdl2-config, and the
cross-toolchain is <triplet>-sdl2-config and
/path/to/toolchain/triplet/bin/sdl2-config.

The toolchain bin path should not be part of PATH, because this can cause
clashes when building for the host (depends on the PATH order, and if the
tool exists on both).

And even if we do add the toolchain bin path, like we currently do at least
on MXE container, if the host path comes before the toolchain then the
<triplet>-sdl2-config will be detected, but this executable *is not* in the
directory that relates to this toolchain, but in the host bin (for example,
/opt/mxe/bin), so we pass the wrong path to --prefix and sdl2 breaks.

This only worked on buildbot because the lookup for mingw was sdl-config
instead of sdl2-config, and sdl-config just doesn't exist on the system.

### DIFF
--- a/configure
+++ b/configure
@@ -462,11 +462,6 @@ find_sdlconfig() {
 			if test -f "$path_dir/$sdlconfig" ; then
 				_sdlconfig="$path_dir/$sdlconfig"
 				echo $_sdlconfig
-				# Save the prefix
-				_sdlpath=$path_dir
-				if test `basename $path_dir` = bin || test `basename $path_dir` = sbin ; then
-					_sdlpath=`dirname $path_dir`
-				fi
 				# break at first sdl-config found in path
 				break 2
 			fi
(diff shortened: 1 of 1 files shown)


sev-:
@orgads we have port maintainers. Thus, changes to areas which you are not responsible for must be discussed with the relevant maintainers.

This is how our project works.

This particular change will be reverted, but in the future, please discuss first instead of just silently self-merging.

orgads:
I had an extensive discussion with @lephilousophe and we tested it on buildbot before I merged. It's all in team-talk on discord.

sev-:
That is great, but last time I checked, the Win32 port maintainer was @lotharsm. @lephilousophe is our noble buildbot wizard.

orgads:
I'm sorry, I wasn't aware of that. And @lotharsm was tagged several times in this discussion, but didn't show up.

Sorry for the mess anyway.

sev-:
Do not worry, your work is great, just this time was a bit too hasty. I think, after the current discussion in the channel, you may come up with a better solution. As for the reactions, I did not react either. Real-life does not always allow reacting within a few hours time window.

orgads:
It was actually 2 days, but never mind.

PR #3211, opened 2021-07-31
SCUMM: Restore missing cigar smoke to captain Smirk's close-up

eriktorbjorn:
In the EGA and VGA floppy versions, the close-up of captain Smirk has animated cigar smoke. This is missing from the VGA CD version. This pull request fixes that. (Note that it's already fixed in the "Ultimate Talkie" version, but my fix should not interfere with that.)

The animated smoke is set up by the entry script for room 76. In the VGA floppy version (I don't have the EGA one) it looks like this:

```
[001A] (A8) if (VAR_FIXEDDISK) {
[001F] (13)   ActorOps(12,[Costume(76)]);
[0024] (18) } else {
[0027] (13)   ActorOps(12,[Costume(0)]);
[002C] (**) }
[002C] (13) ActorOps(12,[Init(),Palette(2,3),Palette(9,3),Palette(3,7)]);
[0039] (5D) setClass(12,[150,148]);
[0043] (11) animateCostume(12,250);
[0046] (2D) putActorInRoom(12,76);
[0049] (01) putActor(12,172,78);
[004F] (11) animateCostume(12,6);
```

I think this means that when playing the game from floppy, you get costume 0 which doesn't have any animation for the smoke. Perhaps to cut down on disc accesses? In the CD version, it looks like this:

```
[0000] (13) ActorOps(12,[Costume(0)]);
[0005] (13) ActorOps(12,[Init(),Palette(2,3),Palette(9,3),Palette(3,7),IgnoreBoxes(),NeverZClip()]);
[0014] (11) animateCostume(12,250);
[0017] (2D) putActorInRoom(12,76);
[001A] (01) putActor(12,172,78);
[0020] (11) animateCostume(12,6);
```

So it _always_ gets the non-animated costume. Oops. Furthermore, the position of the smoke has been changed for the worse in a few spots. I've adjusted them to match the floppy version. This is handled by script 57. The whole scene can be tested with boot param 888.

The "Ultimate Talkie" version also corrects the position of the smoke, but uses slightly different positions than the floppy version. The difference is never more than a single pixel, so I see no reason to change that here.

The two versions use different colors for the smoke. I see no reason to change that, since both look fine to me. This is what the floppy version looks like:

scummvm-monkeyvga-00000

And this is what the CD version looks like, with this pull request:

scummvm-monkey1-00000

### DIFF
--- a/engines/scumm/script_v5.cpp
+++ b/engines/scumm/script_v5.cpp
@@ -444,7 +444,25 @@ void ScummEngine_v5::o5_actorOps() {
 			getVarOrDirectByte(PARAM_1);
 			break;
 		case 1:			// SO_COSTUME
-			a->setActorCostume(getVarOrDirectByte(PARAM_1));
+			i = getVarOrDirectByte(PARAM_1);
+
+			// WORKAROUND: In the VGA floppy version of the Monkey
+			// Island 1, there are two different costumes for the
+			// captain Smirk close-up: 0 for when the game is run
+			// from floppies, and 76 for when the game is run from
+			// hard disk, I believe.
+			//
+			// Costume 0 doesn't have any cigar smoke, perhaps to
+			// cut down on disk access.
+			//
+			// But in the VGA CD version, only costume 0 is used
+			// and the close-up is missing the cigar smoke.
+
+			if (_game.id == GID_MONKEY && _currentRoom == 76 && act == 12 && i == 0) {
+				i = 76;
+			}
+
+			a->setActorCostume(i);
 			break;
 		case 2:			// SO_STEP_DIST
 			i = getVarOrDirectByte(PARAM_1);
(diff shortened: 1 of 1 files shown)


eriktorbjorn:
I'm guessing the difference in colors may be because of a different palette used for the costume. The script remaps three colors, but the costume may not use all of them?

That might also explain why the floppy and CD versions have different colors for the smoke in the voodoo lady's place. It looks like it may be the same animation with different colors.

The color indexes used in the CD version seems to be 1, 2 and 3, so it may be possible to fix the calls to o5_actorOps(), SO_PALETTE, to set the floppy colors. I'm just not sure exactly what to remap them to at the moment. And is it worth it?

eriktorbjorn:
For reference, these are the color differences I'm talking about. The VGA floppy version:

scummvm-monkeyvga-00001
scummvm-monkeyvga-00002

And the VGA CD version:

scummvm-monkey1-00001
scummvm-monkey1-00002

I don't know if the smoke is used in any other scenes. In the voodoo lady scene, I guess the cauldron (which also uses different colors) is a separate thing. Actually, I haven't even verified that all of these use the same animation. But it seems likely.

So what to do about these? The CD version is probably the most common one today (not counting the special edition), so in a way that became the canonical version. So maybe the colors should be left alone for these two scenes. It does seem to be the colors they went with for the special editions.

But in the case of captain Smirk, we're now enabling something that hasn't been seen before in the official release. (It's not in the special edition either.) So perhaps we should make an effort to match the floppy version here? I'll see what I can do...

eriktorbjorn:
I can't find the appropriate blue-gray color in the existing palette, and patching both the palette and the SO_PALETTE calls seems a bit too invasive to me.

Ah, heck, the CD version's smoke looks fine I guess.

eriktorbjorn:
So the current version of the patch is probably final. I've merged the two commits together, since the second was just changes to the comments anyway.

sev-:
Thank you!

PR #3212, opened 2021-07-31
Change the default for useEllipsis to false in Font::drawString

criezy:
This change was made for a couple of reasons:
 - The default is now consistent with the one in Font::getBoundingBox
 - Engines are unlikely to want ellipsis

I checked the current use of `Font::drawString`:
 - The GUI always specify if it wants ellipses or not. So changing the default does not impact this code.
 - The OSD and virtual keyboard used the default and I changed the code to continue using ellipsis if needed.
 - The MacGUI code (`MacMenu` and `MacText`) used the default and  I changed the code to continue using ellipsis if needed (except for the `MacMenu` accelerator string). I am not familiar with this code, and not sure if it should use ellipsis or not. So I would be happy to change it if you think I got it wrong.
 - The engines always (or almost always - I did not check all the usage) use the default. But I assumed no engines wants ellipsis, so did not change any of the calls there.

This pull request fixes for example the menu in Myst ME:
 
Current master:
image

With this pull request:
image

### DIFF
--- a/backends/graphics/opengl/opengl-graphics.cpp
+++ b/backends/graphics/opengl/opengl-graphics.cpp
@@ -849,7 +849,7 @@ void OpenGLGraphicsManager::osdMessageUpdateSurface() {
 	for (uint i = 0; i < osdLines.size(); ++i) {
 		font->drawString(dst, osdLines[i],
 		                 0, i * lineHeight + vOffset + lineSpacing, width,
-		                 white, Graphics::kTextAlignCenter);
+		                 white, Graphics::kTextAlignCenter, 0, true);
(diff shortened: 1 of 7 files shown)


eriktorbjorn:
What about the ScummVM launcher dialog? That's been the most common case for me where I've seen ellipsis being used.

av-dx:
The launcher dialog GUI is also explicit about whether it requires ellipses or not. Changing the default does not affect the launcher. But I will accordingly need to update the code at few places in my PR for Grid GUI.

sev-:
Thank you!

PR #3215, opened 2021-08-01
COMMON: Fix code style in makefile.

fedor4ever:
That strange makefile structure break Symbian makefile parser.

### DIFF
--- a/common/module.mk
+++ b/common/module.mk
@@ -41,6 +41,7 @@ MODULE_OBJS := \
 	stuffit.o \
 	system.o \
 	textconsole.o \
+	text-to-speech.o \
 	tokenizer.o \
 	translation.o \
 	unarj.o \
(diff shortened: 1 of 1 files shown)


common/module.mk
```diff
@@ -52,7 +52,8 @@ MODULE_OBJS := \
 	winexe_ne.o \
 	winexe_pe.o \
 	xmlparser.o \
-	zlib.o
+	zlib.o \
+	text-to-speech.o
```
criezy, line 56: This list is mostly alphabetical, so it would probably make more sense to move it just after `system.o`.

criezy:
The current structure does not seem particularly strange to me, but I see no issue with moving the text-to-speech.o to the main list. It is currently separate for historical reason because it used to be added only when `USE_TTS` is defined, but this was changed recently (in 9d82fa5).

However as noted in my inline comment I think it would be better to keep the list alphabetical and move it accordingly (and not at the end of the list).

fedor4ever:
@criezy , done!

sev-:
Thanks, merging.

PR #3217, opened 2021-08-01
CONFIGURE: Do not pass --prefix to sdl2-config

orgads:
Reapply a6ded8907570bb4b14101d67f076968f0775f460.

This already works with all the platforms on buildbot. The mingw release
environment had a misconfigured sdl2-config, but this should now be
fixed.


lotharsm (approved):
After manually editing the sdl2-config file on the release server, this now works as intentioned.

PR #3219, opened 2021-08-01
ALL: Fix code style for "End of namespace".

fedor4ever:
I fix small part because it too many (Hundreds?) for one commit review.


sev-:
Why adding a dot for 'End of namespace'? We are pretty consistently avoiding it.

bluegr:

PR #3220, opened 2021-08-01
BACKENDS: Fix codestyle in sdl-graphics.cpp and sdl-graphics.h.

fedor4ever:
See #3170.


sev-:
What is this obsession with the full stops after every comment? Where does it come from?

fedor4ever:
Mindless copy from #3170. I think it's OK...
Less stupid work is good.

bluegr:

sev-:
Indeed, and moreover, the comment is misleading. It is not code style but comment style

fedor4ever:
I stop to waste my time with #3220.

PR #3221, opened 2021-08-01
BACKENDS: KEYMAPPER: Fix code style.

fedor4ever:
See #3170.


sev-:
Same question as for #3220 I do not quite understand the purpose of adding dots after single sentence comments.

bluegr:

sev-:
Closing.

PR #3225, opened 2021-08-01
COMMON: Fix divide by zero in getRandomNumber()

tkln:
Calling RandomSource::getRandomNumber() results in divide by zero
exception when called with max = 4294967295 as max + 1 overflows to zero
and the subsequent modulus operation causes the exception.

Certain game saves in Bladerunner trigger this.

### DIFF
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -20,6 +20,8 @@
  *
  */
 
+#include <climits>
+
 #include "common/random.h"
 #include "common/system.h"
 #include "gui/EventRecorder.h"
@@ -46,6 +48,9 @@ void RandomSource::setSeed(uint32 seed) {
 uint RandomSource::getRandomNumber(uint max) {
 	_randSeed = 0xDEADBF03 * (_randSeed + 1);
 	_randSeed = (_randSeed >> 13) | (_randSeed << 19);
+
+	if (max == UINT_MAX)
+		return _randSeed;
 	return _randSeed % (max + 1);
 }
 


common/random.cpp
```diff
@@ -46,7 +46,7 @@ void RandomSource::setSeed(uint32 seed) {
 uint RandomSource::getRandomNumber(uint max) {
 	_randSeed = 0xDEADBF03 * (_randSeed + 1);
 	_randSeed = (_randSeed >> 13) | (_randSeed << 19);
-	return _randSeed % (max + 1);
+	return _randSeed % MAX(max, 1u);
```
criezy, line 49: That does not look quite right since with this change we now get a result in the range [0, max - 1] instead of [0, max].

criezy (changes requested):

common/random.cpp
tkln: Pushed a fixed version of the commit where the range is preserved (except on when max == UINT_MAX).

criezy: This is better, but you still get the issue that you cannot get max in the case where max == UINT_MAX.

I think you will need something like this:
```c++
if (max == UINT_MAX)
    return _randSeed;
else
    return _randSeed % (max + 1);
```

tkln: Thanks, I've pushed a fixed version.

antoniou79: Nice catch!
The code looks good to me, after the latest commit.

antoniou79:
@tkln It's a good catch of the issue. 
However, since you mention Blade Runner savegames specifically of triggering this, I wonder if the issue goes deeper than this. Do you have such a savegame to share, so I could look further into it?

tkln:
@antoniou79 Yes, I've attached the save file that crashes on load (as .zip, though, because github comments only seem to allow certain known file types).
[bladerunner-final-win.002.zip](https://github.com/scummvm/scummvm/files/6915778/bladerunner-final-win.002.zip)

criezy (approved):

antoniou79:
> @antoniou79 Yes, I've attached the save file that crashes on load (as a .zip, though, because github comments only seem to allow certain known file types).
> [bladerunner-final-win.002.zip](https://github.com/scummvm/scummvm/files/6915778/bladerunner-final-win.002.zip)

Thank you, this is perfect. I'll investigate further if this is something that needs further attention in Blade Runner. 
And thank you for this fix.

antoniou79:
Just for the record, the issue with BladeRunner is that the engine does not really fully initialize some fields if it is not actually needing them -- here it was the ambient sound track fields-- so those fields typically get garbage values. However, it does store those uninitialized tracks with their garbage field values, and on load it would pointlessly run the getRandomNumber() on these fields (the specific culprit were garbage values for track.delayMin, track.delayMax. 
I've added some notes and extra code in the BladeRunner engine to at least sanitize bad loaded values in this commit. Of course we could also init the track fields in the constructor (but the code elsewhere is such so it should ignore the track fields if the track is not active and also that would not solve the issue with old save games or imported classic game saves; I might do that in another commit, just in case, though):
https://github.com/scummvm/scummvm/commit/0c203f7edf855de0c75f9009cc14e6ce47e93191

PR #3226, opened 2021-08-02
DEVTOOL: move engine credits from devtools/credits.pl to engines

fingolfin:
This PR moves engine credits from `devtools/credits.pl` to the respective engines. The credits are then included by `devtools/credits.pl`. They are *always* included, even if an engine is disabled, to match the current behavior. Of course that could be changed in a future PR, if desired (personally, I think always showing all credit has some merit, but YMMV).

There are a few additional minor changes, which I will happily revise or drop based on your feedback:
- the credits listed SCUMM and HE first for historical reasons. As far as I can tell, this has been given up everywhere (???) else, so perhaps it is time to do it in the credits, too? It also made the code simpler (my main motivation, to be honest 😂 ). However, If you prefer to keep the special case for SCUMM/HE, I can add some code to ensure that.
- the credits still listed the "NGI" engine as "Fullpipe", which I changed in one commit
- I renamed "HE" to "SCUMM HE" in the credits, to ensure an overall alphabetical ordering, and also because I felt this might make it easier to discover what "HE" is (as it is not the name of "main" engine, unlike all other engine credits. 

Finally, I marked this as a draft as it should not be merged in its current form, even if you were 100% happy with the change: This is because the first commit contains the script I used to split `devtools/credits.pl`. It is written in Julia (still a bit exotic, but since I code a lot in it these days, it was the quickest way for me to get the job done), and having it here makes it easier to reproduce the splitting (to verify I didn't mess up, and also to redo it based on feedback for this PR). In case this PR gets approved, it should be dropped before merging, as the script is useless after this PR.


sev-:
Good and meaningful change. As of having SCUMM engine first, I did not quite understand what you mean by "this has been given up everywhere", e.g. we still have it on the top in the credits and in the build process. Could you please explain more what you mean?

Other than that, great change, ready for merging.

The side effect (and perhaps the purpose) of it will be the fact that any custom build with a different set of engines will always have its credits in a proper shape.

criezy:
> As of having SCUMM engine first, I did not quite understand what you mean by "this has been given up everywhere", e.g. we still have it on the top in the credits and in the build process. Could you please explain more what you mean?

I don't know which ones Max had in mind, but I can think of several examples:
 - The compatibility page on the website
 - The Engine page on the wiki
 -  The engine list in the `configure --help` output.

None of those list the SCUMM engine first.

Personally I see no issue with changing the credits to also have all engines in alphabetical order.

fingolfin:
Sorry, that was (too) lazily phrased. When I wrote "everywhere (???)", the "???" was meant to indicate doubt; but in retrospect that was too cryptic. I should have written "everywhere that I checked" -- and clearly I did not check all spots, so I should have made explicit what I checked.

@criezy is spot on with his list. But now I went a bit further and tried to find more places that list the engines. Among those that I found, the following listed SCUMM alphabetically:
- the news entries in `NEWS.md`
- output of `./configure --help`
- list on https://wiki.scummvm.org/index.php?title=Engines
- the compatibility page https://www.scummvm.org/compatibility/
- the documentation of available settings: https://docs.scummvm.org/en/v2.2-docs/settings/engine.html
- the list of engines in the bug tracker component field (OK, granted, you have no easy control over it so it could be considered an accident)
- https://wiki.scummvm.org/index.php?title=Category:Engines_Added_to_ScummVM (again, admittedly out of your control)

Places that sort SCUMM first:
- the credits (but I was aware of that, obviously, that's part of what this PR is about)
- `engines/engines.mk`
- the output of `configure` when running it lists SCUMM as first engine (due to code in `engines.awk`)
- the SCUMM engine is built before all others (due to the same code in `engines.awk`)

Anyway: in the end, it doesn't matter how many places follow which convention; after all, you can do it like you want. So, just let me know if I should add code to sort SCUMM first or not.

engines/glk/credits.pl
```diff
@@ -0,0 +1,8 @@
+begin_section("ScummGlk");
```
fingolfin, line 1: By the way, is this intentionally listed as "ScummGlk" ? Elsewhere it seems this is referred to as just "Glk" ?

sev-:
I have no issue with having the about sorted alphabetically either. So, let's not complicate the code.

fingolfin:
Done. Also dropped the TEMP commit. Should I also rename "ScummGlk" to "Glk", though?

sev-:
Yes please, this is an artefact from the pre-scummvm engine existence.

fingolfin:
OK, changed ScummGlk -> Glk

sev-:
Thank you! merging

PR #3229, opened 2021-08-02
MAKEFILE: Fix shadow build of ideprojects

orgads:


ports.mk
```diff
@@ -546,13 +546,11 @@ else ifeq "$(CUR_BRANCH)" ""
...
 	@echo Creating Code::Blocks project files...
-	@cd $(srcdir)/dists/codeblocks && ../../devtools/create_project/create_project ../.. --codeblocks >/dev/null && git add -f engines/plugins_table.h *.workspace *.cbp
+	@cd $(srcdir)/dists/codeblocks && $(PWD)/devtools/create_project/create_project ../.. --codeblocks >/dev/null
 	@echo Creating MSVC project files...
-	@cd $(srcdir)/dists/msvc && ../../devtools/create_project/create_project ../.. --use-canonical-lib-names --msvc-version 12 --msvc >/dev/null && git add -f engines/plugins_table.h *.sln *.vcxproj *.vcxproj.filters *.props
+	@cd $(srcdir)/dists/msvc && $(PWD)/devtools/create_project/create_project ../.. --use-canonical-lib-names --msvc-version 12 --msvc >/dev/null
 	@echo
 	@echo All is done.
-	@echo Now run
-	@echo "\tgit commit -m 'DISTS: Generated Code::Blocks and MSVC project files'"
```
lotharsm, line 554: I think this one should stay. This commit message is the default commit message for the past releases, since it marks the last step in our release process.

sev-: Indeed, it is part of my release procedure. Please, keep it.

orgads: Thanks for the reference. I wasn't familiar with your branching scheme.

Removed this commit, and left only the first one that allows shadow build. Change the PR subject accordingly.

sev-:
Thank you, merging.

PR #3231, opened 2021-08-02
SCUMM: Fix findObject(x,y) for DIG and COMI

AndywinXp:
Today, while I was testing some sound stuff in COMI, I found out that one of the three destroyable turrets (the farthest one) in the cannon minigame in Part 1 couldn't be destroyed.

...and I took that personally.

Turns out that while Full Throttle (and I believe every previous SCUMM game) indeed did what ScummVM already does when checking for object coordinates, The Dig and COMI use the '>=' operator in place of '>'.

And now, when I try to destroy that turret:

image

Yay!

### DIFF
--- a/engines/scumm/object.cpp
+++ b/engines/scumm/object.cpp
@@ -540,9 +540,15 @@ int ScummEngine::findObject(int x, int y) {
 						return _objs[i].obj_nr;
 				}
 #endif
-				if (_objs[i].x_pos <= x && _objs[i].width + _objs[i].x_pos > x &&
-				    _objs[i].y_pos <= y && _objs[i].height + _objs[i].y_pos > y)
-					return _objs[i].obj_nr;
+				if (_game.id == GID_CMI || _game.id == GID_DIG) {
+					if (_objs[i].x_pos <= x && _objs[i].width + _objs[i].x_pos >= x &&
+						_objs[i].y_pos <= y && _objs[i].height + _objs[i].y_pos >= y)
+						return _objs[i].obj_nr;
+				} else {
+					if (_objs[i].x_pos <= x && _objs[i].width + _objs[i].x_pos > x &&
+						_objs[i].y_pos <= y && _objs[i].height + _objs[i].y_pos > y)
+						return _objs[i].obj_nr;
+				}
 				break;
 			}
 		} while ((_objs[b].state & mask) == a);


sluicebox:
Beautiful! This would make a great item in NEWS.md

athrxx:
Thanks! These changes seem to be correct. (I have only looked at COMI and DIG, since these are the ones you're changing). If there is no objection we can merge this...

aquadran:
I would look into older scumm engine games too. is the bug too. explicit adding dig/ft looks like w/a for me

AndywinXp:
I currently have no mean to look at older decomps, but I'm just assuming the current ScummVM version with '>' is correct for older games since Full Throttle uses exactly that, while the new one with '>=' is only used for COMI and DIG.

aquadran (changes requested):
this has to be verified

AndywinXp:
Very well, managed to find my Activision copy of DOTT (the one from around 2003), spent some time remapping SCUMM OP codes on IDA, and indeed the '>' operator is used, like in FT. 

The only difference is that in DOTT x and y values are divided by 8 when doing the check:
`if (_objs[i].x_pos <= (x / 8) && _objs[i].width + _objs[i].x_pos > (x / 8) &&`
`_objs[i].y_pos <= (y / 8) && _objs[i].height + _objs[i].y_pos > (y / 8))`

Can anyone familiar with SCUMM v6 confirm this?

AndywinXp:
Sorry for the many notifications; I've bought the LucasArts Adventure Pack on Steam, which contains the Aaron Giles' versions of LOOM, FOA, DIG and Last Crusade, giving me the chance to look into versions 3 and 4.

Again, x and y are divided by 8 before the check. 🤷 I'll do some testing with this stuff and report back here.

AndywinXp (commented):
Original comment:
_Sorry for the many notifications; I've bought the LucasArts Adventure Pack on Steam, which contains the Aaron Giles' versions of LOOM, FOA and Last Crusade, giving me the chance to look into versions 3 and 4 and 5._

_Again, x and y are divided by 8 before the check. 🤷 I'll do some testing with this stuff and report back here._

_EDIT: Turns out, the division by 8 is not needed in v6 and below (if I put it in, it just breaks everything), it must have been part of some side calculation which ScummVM handles directly, so, to summarize_: 
_**I can confirm this commit is compliant with SCUMM versions 3 to 8**, v0 to 2 are fine for sure because, again, I didn't touch anything other than for COMI and DIG._

Review:

This "if" bisection is compliant with the following disasms:

- Last Crusade
- Fate Of Atlantis
- Loom
- Sam & Max
- DOTT
- Full Throttle

- The Dig
- COMI

aquadran (approved):

PR #3233, opened 2021-08-02
DEVTOOLS: Replace ugly SDL/MinGW hack with a cleaner hack

orgads:
Instead of having `main=SDL_main` defined by `pkg-config`/`sdl2-config --cflags`
and undefining it everywhere, strip it in its origin.


lotharsm (approved):
Looking good to me - thanks!

configure
```diff
@@ -3790,7 +3790,7 @@ if test "$_sdlnet" = auto ; then
 	if test "$_sdl" = no || test "$_sdlMajorVersionNumber" = 2; then
 		if test "$_pkg_config" = "yes" && $_pkgconfig --exists SDL2_net; then
 			append_var SDL2_NET_LIBS "`$_pkgconfig --libs SDL2_net`"
-			append_var SDL2_NET_CFLAGS "`$_pkgconfig --cflags SDL2_net`"
+			append_var SDL2_NET_CFLAGS "`$_pkgconfig --cflags SDL2_net | sed 's/[[:space:]]*-Dmain=SDL_main//g'`"
```
lephilousophe, line 3793: Is it really needed here?

configure
```diff
@@ -3799,7 +3799,7 @@ if test "$_sdlnet" = auto ; then
 	if test "$_sdl" = no || test "$_sdlMajorVersionNumber" = 1; then
 		if test "$_pkg_config" = "yes" && $_pkgconfig --exists SDL_net; then
 			append_var SDL1_NET_LIBS "`$_pkgconfig --libs SDL_net`"
-			append_var SDL1_NET_CFLAGS "`$_pkgconfig --cflags SDL_net`"
+			append_var SDL1_NET_CFLAGS "`$_pkgconfig --cflags SDL_net | sed 's/[[:space:]]*-Dmain=SDL_main//g'`"
```
lephilousophe, line 3802: and here?

lephilousophe:
Overall, I don't understand the idea to request review and not even wait for a day before merging.
I agree PR must not get forgotten but taking some days is not a problem at all.

configure
orgads: Yes, because SDL_net "requires" sdl2, so it returns its flags too.

configure
orgads: Not sure about this one, but it shouldn't hurt.

orgads:
Trust me, I'm an engineer. 😉

Just kidding. I'll wait longer next time. One thing I like about gerrit is that a reviewer can vote +1, still requiring another reviewer to vote +2 before merging. A feature I miss on github.

PR #3234, opened 2021-08-03
COMMON: Create Path type

djsrv:
This introduces `Common::Path`, a type which is used to pass paths to file-related common functions. Previously, these functions took a string as an argument, using '/' as a directory separator. This was a problem for the Director engine, which has several titles with '/' in file names. Path solves this by allowing path strings with arbitrary directory separators to be converted to/from a safer, common form.

For example, a Path can be created with `Common::Path("dir:subdir:file", ':')` where ':' is the directory separator. A Path can be converted back to a string with `path.toString(':')`. If no directory separator is specified, it defaults to '/', allowing existing code to work with almost no modification.

Internally, Path is just a simple wrapper around a string, and it uses '\x1f' (unit separator) to separate path components. As this is a control character, I don't anticipate that it will appear in file names. However, if needed, Path could easily be changed to use a different separator character or to store path components in an array.

### DIFF
--- a/backends/platform/android/asset-archive.cpp
+++ b/backends/platform/android/asset-archive.cpp
@@ -163,11 +163,10 @@ const Common::ArchiveMemberPtr AndroidAssetArchive::getMember(const Common::Path
 }
 
 Common::SeekableReadStream *AndroidAssetArchive::createReadStreamForMember(const Common::Path &path) const {
-	Common::String name = path.toString();
-	if (!hasFile(name)) {
+	if (!hasFile(path)) {
 		return nullptr;
 	}
-	return new AssetInputStream(_am, name);
+	return new AssetInputStream(_am, path.toString());
 }
 
 #endif
--- a/backends/platform/android/asset-archive.cpp
+++ b/backends/platform/android/asset-archive.cpp
@@ -115,7 +115,8 @@ AndroidAssetArchive::AndroidAssetArchive(jobject am) : _hasCached(false) {
 AndroidAssetArchive::~AndroidAssetArchive() {
 }
 
-bool AndroidAssetArchive::hasFile(const Common::String &name) const {
+bool AndroidAssetArchive::hasFile(const Common::Path &path) const {
+	Common::String name = path.toString();
 	AAsset *asset = AAssetManager_open(_am, name.c_str(), AASSET_MODE_RANDOM);
 	bool exists = false;
 	if (asset != NULL) {
@@ -156,15 +157,17 @@ int AndroidAssetArchive::listMembers(Common::ArchiveMemberList &member_list) con
 	return count;
 }
 
-const Common::ArchiveMemberPtr AndroidAssetArchive::getMember(const Common::String &name) const {
+const Common::ArchiveMemberPtr AndroidAssetArchive::getMember(const Common::Path &path) const {
+	Common::String name = path.toString();
 	return Common::ArchiveMemberPtr(new Common::GenericArchiveMember(name, this));
 }
 
-Common::SeekableReadStream *AndroidAssetArchive::createReadStreamForMember(const Common::String &path) const {
-	if (!hasFile(path)) {
+Common::SeekableReadStream *AndroidAssetArchive::createReadStreamForMember(const Common::Path &path) const {
+	Common::String name = path.toString();
+	if (!hasFile(name)) {
 		return nullptr;
 	}
-	return new AssetInputStream(_am, path);
+	return new AssetInputStream(_am, name);
 }
 
 #endif
--- a/common/fs.cpp
+++ b/common/fs.cpp
@@ -35,15 +35,16 @@ FSNode::FSNode(AbstractFSNode *realNode)
 	: _realNode(realNode) {
 }
 
-FSNode::FSNode(const String &p) {
+FSNode::FSNode(const Path &p) {
 	assert(g_system);
 	FilesystemFactory *factory = g_system->getFilesystemFactory();
 	AbstractFSNode *tmp = nullptr;
+	String s = p.toString();
 
-	if (p.empty() || p == ".")
+	if (s.empty() || s == ".")
 		tmp = factory->makeCurrentDirectoryFileNode();
 	else
-		tmp = factory->makeFileNodePath(p);
+		tmp = factory->makeFileNodePath(s);
 	_realNode = SharedPtr<AbstractFSNode>(tmp);
 }
 
@@ -175,25 +176,25 @@ FSDirectory::FSDirectory(const FSNode &node, int depth, bool flat, bool ignoreCl
 	_includeDirectories(includeDirectories) {
 }
 
-FSDirectory::FSDirectory(const String &prefix, const FSNode &node, int depth, bool flat,
+FSDirectory::FSDirectory(const Path &prefix, const FSNode &node, int depth, bool flat,
 						 bool ignoreClashes, bool includeDirectories)
   : _node(node), _cached(false), _depth(depth), _flat(flat), _ignoreClashes(ignoreClashes),
 	_includeDirectories(includeDirectories) {
 
-	setPrefix(prefix);
+	setPrefix(prefix.rawString());
 }
 
-FSDirectory::FSDirectory(const String &name, int depth, bool flat, bool ignoreClashes, bool includeDirectories)
+FSDirectory::FSDirectory(const Path &name, int depth, bool flat, bool ignoreClashes, bool includeDirectories)
   : _node(name), _cached(false), _depth(depth), _flat(flat), _ignoreClashes(ignoreClashes),
 	_includeDirectories(includeDirectories) {
 }
 
-FSDirectory::FSDirectory(const String &prefix, const String &name, int depth, bool flat,
+FSDirectory::FSDirectory(const Path &prefix, const Path &name, int depth, bool flat,
 						 bool ignoreClashes, bool includeDirectories)
   : _node(name), _cached(false), _depth(depth), _flat(flat), _ignoreClashes(ignoreClashes),
 	_includeDirectories(includeDirectories) {
 
-	setPrefix(prefix);
+	setPrefix(prefix.rawString());
 }
 
 FSDirectory::~FSDirectory() {
@@ -202,8 +203,8 @@ FSDirectory::~FSDirectory() {
 void FSDirectory::setPrefix(const String &prefix) {
 	_prefix = prefix;
 
-	if (!_prefix.empty() && !_prefix.hasSuffix("/"))
-		_prefix += "/";
+	if (!_prefix.empty() && _prefix.lastChar() != DIR_SEPARATOR)
+		_prefix += DIR_SEPARATOR;
 }
 
 FSNode FSDirectory::getFSNode() const {
@@ -222,7 +223,8 @@ FSNode *FSDirectory::lookupCache(NodeCache &cache, const String &name) const {
 	return nullptr;
 }
 
-bool FSDirectory::hasFile(const String &name) const {
+bool FSDirectory::hasFile(const Path &path) const {
+	String name = path.rawString();
 	if (name.empty() || !_node.isDirectory())
 		return false;
 
@@ -230,7 +232,8 @@ bool FSDirectory::hasFile(const String &name) const {
 	return node && node->exists();
 }
 
-const ArchiveMemberPtr FSDirectory::getMember(const String &name) const {
+const ArchiveMemberPtr FSDirectory::getMember(const Path &path) const {
+	String name = path.rawString();
 	if (name.empty() || !_node.isDirectory())
 		return ArchiveMemberPtr();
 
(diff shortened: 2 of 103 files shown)


sev-:
Thank you for working on this. I was always bothered by the not-enough abstraction of paths in ScummVM and this PR cleanly approaches it.

I requested number of people to review it, hope to hear the feedback soon

dreammaster:
I've reviewed the code, and it all looks excellent, it'll certainly make path handling cleaner, as well as easier to access files in a subfolder off the game path in the future. I'll admit that in the past I've done things like wrapping ConfMan.get("path") in a Common::FSNode, and then calling getChild("folder").getChild("file1.txt") to a node that I can use to open the file. Now, though, I'll be able to safely do something like:
```
	Common::Path p3("folder/file1.txt");
	if (f.open(p3))
		error("wow");

```
which I tested to confirm worked. So I'm quite happy with this change. So I'm in favor of merging it.

sev-:
@criezy, @aquadran, @bluegr Your thoughts?

criezy (commented):
I had a quick look and this looks a good idea and implementation. I will try to look at it in more details this weekend when I have more time.

But I have one request already: can you please add some doxygen documentation for the Path class?

I noticed that the `+=` operator does not add separators. So for example:
```c++
Path path("/foo");
path += Path("bar"); // "/foobar"
```
And I was wondering if we would expect it to give "/foo/bar" instead. I don't think that is necessarily the case, but it would be good to have documentation to clarify what the functions do.

Other than that I am also wondering if it would make sense to add some utility function to the path class to split the path. For example:
```c++
    String getFileName() const;
    Path getDirectory() const;
    StringArray splitPath() const;
```

```c++
Path path("/path/to/file");
String fileName = path.getFileName(); // "file"
Path directory = path.getDirectory(); // "/path/to/"
StringArray parts = path.splitPath(); //  { "path", "to", "file" }
```

And similarly it could have functions to build a path from its parts, or for example add a subdirectory of file, which would take care of adding the separators.

But at this point those might not be needed and could be added later if we find it would help to simplify some code.

djsrv:
I've added Doxygen comments.

I think it's reasonable to expect that `+=` would insert a directory separator, but there are some cases where it doesn't make much sense. For example, if you wanted to add a file extension:
```
Path path("/foo");
path += ".bar"; // "/foo.bar"
```
If a path separator was automatically inserted, this would become much more difficult, so I think the current behavior is probably the most reasonable. In any case, I've documented the behavior in the Doxygen comments.

I agree that more utility functions would be nice. But they weren't an immediate necessity for the Director engine's use case, and I didn't want to prematurely design unneeded APIs.

dreammaster:
I too toyed with the idea of whether it would be useful for the appending to automatically add directory separators, before I decided against suggesting it. Because, similar to my prior example, if a directory separator is needed, one can just include it manually in whatever string is being appended (using whatever separator is set for the path). So there's no real benefit to doing it automatically, and as djsrv says, it would prevent building up filenames from working if it did so

backends/platform/android/asset-archive.cpp
```diff
@@ -156,15 +157,17 @@ int AndroidAssetArchive::listMembers(Common::ArchiveMemberList &member_list) con
...
+const Common::ArchiveMemberPtr AndroidAssetArchive::getMember(const Common::Path &path) const {
+	Common::String name = path.toString();
 	return Common::ArchiveMemberPtr(new Common::GenericArchiveMember(name, this));
 }
 
-Common::SeekableReadStream *AndroidAssetArchive::createReadStreamForMember(const Common::String &path) const {
-	if (!hasFile(path)) {
+Common::SeekableReadStream *AndroidAssetArchive::createReadStreamForMember(const Common::Path &path) const {
+	Common::String name = path.toString();
+	if (!hasFile(name)) {
```
somaen, line 167: Wasn't this changed to take a Common::Path as parameter?

common/fs.cpp
```diff
@@ -35,15 +35,16 @@ FSNode::FSNode(AbstractFSNode *realNode)
...
 
-FSNode::FSNode(const String &p) {
+FSNode::FSNode(const Path &p) {
 	assert(g_system);
 	FilesystemFactory *factory = g_system->getFilesystemFactory();
 	AbstractFSNode *tmp = nullptr;
+	String s = p.toString();
 
-	if (p.empty() || p == ".")
+	if (s.empty() || s == ".")
```
somaen, line 44: This sounds like something that a Path-class should have members for.

common/installshieldv3_archive.cpp
```diff
@@ -107,7 +107,8 @@ void InstallShieldV3::close() {
 	_map.clear();
 }
 
-bool InstallShieldV3::hasFile(const Common::String &name) const {
+bool InstallShieldV3::hasFile(const Common::Path &path) const {
+	Common::String name = path.toString();
```
somaen, line 111: All of this .toString()-inging, can't that be done inline in the various callsites? (Or where it makes sense, possibly make the called function take a Path, if that's the only thing it needs to deal with anyhow)

criezy:
> if a directory separator is needed, one can just include it manually in whatever string is being appended (using whatever separator is set for the path).

That actually points to something else that is bothering me and is potentially dangerous: the `Path` does not remember the separator used. So the following code for example probably does not do what you expect:
```c++
Path path(":foo", ":");
path += ":bar";
```

The `+` and `+=` operators with a `String` or `const char *` assume we are using the default separator `/` for those. I think most developer would assume that it continues to use the separator we used when creating the path, as it seems you do when you write "whatever separator is set for the path". But with the current code we don't really "set" a separator and it is forgotten as soon as the input is converted to the internal separator and not used when appending further strings.

The code above would work by explicitly using paths everywhere:
```c++
Path path(":foo", ":");
path += Path(":bar", ":");
```

So the issue is not that we cannot do it, but rather that the behaviour might be unexpected and thus potentially dangerous.

common/installshieldv3_archive.cpp
```diff
@@ -118,11 +119,13 @@ int InstallShieldV3::listMembers(Common::ArchiveMemberList &list) const {
 	return _map.size();
 }
 
-const Common::ArchiveMemberPtr InstallShieldV3::getMember(const Common::String &name) const {
+const Common::ArchiveMemberPtr InstallShieldV3::getMember(const Common::Path &path) const {
+	Common::String name = path.toString();
 	return Common::ArchiveMemberPtr(new Common::GenericArchiveMember(name, this));
```
somaen, line 124: I.e. here, shouldn't the argument to GenericArchiveMember be a Common::Path, and not a Common::String?

common/macresman.cpp
```diff
@@ -606,39 +606,41 @@ void MacResManager::readMap() {
 	}
 }
 
-String MacResManager::constructAppleDoubleName(String name) {
+Path MacResManager::constructAppleDoubleName(Path name) {
 	// Insert "._" before the last portion of a path name
-	for (int i = name.size() - 1; i >= 0; i--) {
+	String rawName = name.rawString();
```
somaen, line 611: Might be overkill, but how about adding functions to Common::Path that output something like Common::List<Common::String> pathComponents, and a similar function for creating a Common::Path from such a list?

That would make it easier to transform each path-component, without having to care about the DIR_SEPARATOR directly.

common/path.h
```diff
@@ -0,0 +1,158 @@
...
+
+	/** @overload */
+	Path &operator=(const String &str);
+
+	/** @overload */
+	Path &operator=(char c);
+
+	/**
+	 * Append the given path to this path. Does not automatically
+	 * add a directory separator.
```
somaen, line 125: I think that concatenating two paths should yield a path with directory separator added. (Similar to Python's os.path.join). But then again, what if the two instances have different separators defined?

somaen:
Overall this makes dealing with paths a lot cleaner, my comments mostly point toward wishes for further abstractions for some use cases I saw changing.

As for the += case, I guess the safest thing might be to drop the operator-overload, in favour of something like a 
.append() - does not add separator
.join() - adds separator

That would be less surprising and more explicit than the operator overload. BUT, it has the same issue in cases where the separator happens to differ between the join-ing Paths, perhaps we should just assert against that? (But that would also require all the instances to retain knowledge of the separator used.

djsrv:
> That actually points to something else that is bothering me and is potentially dangerous: the `Path` does not remember the separator used.

My original implementation worked like this, but I think it defeats the purpose of having a Path type that's mostly agnostic to which directory separator is used. It also has the capacity for dangerous, more invisible behavior. For example:
```
Common::Path a() {
	return Common::Path(":foo", ':');
}

Common::Path b() {
	return Common::Path("/foo", '/');
}

Common::Path c(Common::Path path) {
	path += "/bar";
	return path;
}

c(a()); // returns a path to the file "foo/bar"
c(b()); // returns a path to file "bar" in directory "foo"
```
The initial paths created in `a` and `b` are the same exact path, only differing by the separator used in the input string. Yet the same transformation to them results in totally different paths. I don't think that makes sense at all, especially since the directory separator is supposed to be taken out of the equation here.

Forgetting which directory separator was originally used is slightly annoying, but it at least results in consistent behavior.

common/installshieldv3_archive.cpp
djsrv: I tried doing that initially, but I found that in a lot of cases, it resulted in code much uglier (and slightly less efficient) than stringifying the path once at the beginning of the function.

As for the necessity of stringifying the paths in the first place, there are a ton of `Archive` subclasses which were designed to use strings as input. Ideally, they should all be changed to fully work with paths, but doing that would require modifying *a lot* of code I'm unfamiliar with, which would eat up my remaining GSoC time and possibly introduce bugs. Stringifying the path and leaving the rest of the code as-is guarantees that the code's behavior will be the same as before, so I don't think it's worth the effort of making further changes to most of these `Archive` subclasses right now.

djsrv:
> As for the += case, I guess the safest thing might be to drop the operator-overload, in favour of something like a
> .append() - does not add separator
> .join() - adds separator
> 
> That would be less surprising and more explicit than the operator overload. BUT, it has the same issue in cases where the separator happens to differ between the join-ing Paths, perhaps we should just assert against that? (But that would also require all the instances to retain knowledge of the separator used.

I agree, having separate `.append()` and `.join()` is a safer solution. I still don't think that Path should remember which directory separator was initially used, though, for the reason stated above.

criezy:
Thank you for explaining your reasons for the current design. They make sense.

I do like the proposal from @somaen though as it helps reduce the ambiguity in the behaviour for both issues I raised (should it add a separator, and what separator is used for the string being appended).
```c++
    Path &append(const String &str, char separator = '/') {
        _str += Path(str, separator).rawString();
        return *this;
    }

    Path &join(const String &str, char separator = '/') {
        if (!_str.empty() && !_str.hasSuffix(DIR_SEPARATOR) && !str.hasPrefix(separator))
            _str += DIR_SEPARATOR;
        _str += Path(str, separator).rawString();
        return *this;
    } 
```

I also just noticed that your `setString()` doesn't clear the member `_str` before adding the given string, which might actually be a good thing as it allows avoiding the creation of a temporary `Path` object in the example above.
```c++
    Path &append(const String &str, char separator = '/') {
        setString(str.c_str(), separator);
        return *this;
    } 
```
However this is confusing because of the function name. You might want to rename it `appendString()` instead.

djsrv:
> I also just noticed that your `setString()` doesn't clear the member `_str` before adding the given string

Oh, that's a mistake. I'll rename that to `appendString()` and make a `setString()` which properly clears `_str`.

One additional design consideration: do we want `append()` and `join()` to modify the path in-place? If there was an out-of-place version too, how should they be named?

djsrv:
Thoughts on these function names?
```
/**
 * Appends the given path to this path (in-place).
 * Does not automatically add a directory separator.
 */
void appendInPlace(const Path &x);
void appendInPlace(const String &str, char separator = '/');
void appendInPlace(const char *str, char separator = '/');

/**
 * Returns this path with the given path appended (out-of-place).
 * Does not automatically add a directory separator.
 */
Path append(const Path &x);
Path append(const String &str, char separator = '/');
Path append(const char *str, char separator = '/');


/**
 * Joins the given path to this path (in-place).
 * Automatically adds a directory separator.
 */
void joinInPlace(const Path &x);
void joinInPlace(const String &str, char separator = '/');
void joinInPlace(const char *str, char separator = '/');

/**
 * Returns this path joined with the given path (out-of-place).
 * Automatically adds a directory separator.
 */
Path join(const Path &x);
Path join(const String &str, char separator = '/');
Path join(const char *str, char separator = '/');
```

bluegr:
Overall, the implementation is nice!
I too agree that overloading the += operator will be confusing and counterintuitive. A specialized method like `.join()` will make the implementation more usable and straightforward

criezy:
> Thoughts on these function names?

They look OK. I am not loving them, but I don't have any better suggestions.
But I have two comments on your code example:
- The not-in-place functions should probably be const.
- The in-place functions could return a reference to this, as in my previous code example, which would allow chaining them (e.g. `path.joinInPlace("subdir").joinInPlace("file")`).

sev-:
Thank you, merging. The future API enhancements could be implemented in-tree as needed.

PR #3235, opened 2021-08-03
JANITORIAL: Fix some typos

fingolfin:


dreammaster:
Looks good to me, and good catch that it should be global-constructors rather than global-destructors. Just one minor thing.. could you reword the commit to have a "JANITORIAL: " prefix?

fingolfin:
Done

sev-:
Thanks!

PR #3238, opened 2021-08-04
JANITORIAL: add *.o.tmp to .gitignore

fingolfin:
Such files are created during compilation, and should not appear in the
output of `git status` nor should `git add` add them.


sev-:
Thanks!

PR #3241, opened 2021-08-04
SCUMM: COMI: Disable saving/loading during SMUSH videos

AndywinXp:
This commit prevents an edge case in COMI, in which SMUSH videos were terminated in case of an event of autosaving.
Also, saving and loading was disabled in the original exe, so I guess we're not losing much here...


athrxx:
What's the reason for preventing the loading?

bluegr:
It would be better to allow loading, if possible

AndywinXp:
Sure! Done :)

bluegr:
Thanks! Merging

PR #3243, opened 2021-08-06
CONFIGURE: Suppress output of a52 test app

orgads:
`a52_init()` unconditionally calls `a52_imdct_init()`, which prints "No
accelerated IMDCT transform found". This interferes with the normals
configure output, which looks like this:

```
Checking for liba52... No accelerated IMDCT transform found
yes
```


bluegr:
Thanks! Merging

PR #3244, opened 2021-08-06
DEVTOOLS: Minor fixes in ideprojects target

orgads:
* Add also `engines/detection_table.h` to Git
* Replace literal `\t` with a tab in output.


bluegr:
Nice work and very straightforward implementation! Thanks, merging

PR #3245, opened 2021-08-06
CONFIGURE: Improve readline detection

sluicebox:
I was trying to get `--enable-text-console` working on mac and ran into some issues:

1. configure detected apple's libedit as readline, even though doing that produces a broken build.
2. Passing `--with-readline-prefix=/opt/homebrew/Cellar/readline/8.1` caused configure to correctly detect the presence of readline, but then incorrectly detect the readline function signature, which also produced a broken build. The readline tests are run with different include orders and the second one used Apple's libedit even though the previous test had just correctly found and used readline with the prefix I specified.

Now configure's detection test includes a reference to `rl_hook_func_t` which gui/debugger.cpp depends on so that it fails early instead of producing a broken build.

The function signature test now uses the same parameters / include-order as the detection test so that the results are consistent and `--with-readline-prefix` is respected.

I can now `--enable-text-console --with-readline-prefix=/opt/homebrew/Cellar/readline/8.1` on a mac M1 and it works great.

### DIFF
--- a/configure
+++ b/configure
@@ -5037,6 +5037,7 @@ if test "$_text_console" = yes ; then
 #include <readline/history.h>
 
 int main(void) {
+	rl_hook_func_t f; // not present in apple's libedit
 	char *x = readline("");
 }
 EOF
--- a/configure
+++ b/configure
@@ -5055,9 +5055,6 @@ else
 fi
 
 if test "$_readline" = yes ; then
-	append_var LIBS "$READLINE_LIBS $_READLINE_LIBS"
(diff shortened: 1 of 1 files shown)


sev-:
Please, do not push the PR branches directly to the main repo.

Unfortunately, GitHub does not allow preventing doing that, thus I expect that the developers will pay extra attention.

sluicebox:
Oh my gosh I'm so sorry, this is a new machine and I totally blew it. =(

PR #3247, opened 2021-08-07
MOHAWK: MYST: Reword accessibility option for Selenitic Age

lotharsm:
This is just a small patch to reword the new accessibility option in Myst's game options.

Opening as PR since english is not my first language and there might be better solutions.

### DIFF
--- a/engines/mohawk/dialogs.cpp
+++ b/engines/mohawk/dialogs.cpp
@@ -132,7 +132,8 @@ MystOptionsWidget::MystOptionsWidget(GuiObject *boss, const Common::String &name
 		 * We change it to use fuzzy logic.
 		 * By default the option is off.
 		 */
-		_spaceshipFuzzyLogicCheckbox = new GUI::CheckboxWidget(widgetsBoss(), "MystGameOptionsDialog.FuzzyMode", _("~F~uzzy Logic in SpaceShip Active"));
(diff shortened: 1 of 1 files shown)


digitall:
@lotharsm : The original English for this option is a bit colloquial and fuzzy as in fuzzy logic is not really clear for non-specialist users. Your change is good in that respect, though you are losing the ~F~ which I think indicates the keyboard shortcut for control... Might want to add a similar one as this is about accessability.

digitall:
@lotharsm : You can shorten English sentences much more than German so "Enable accessibility feature for the Selenitic Age" can be shortened to "Improve Selenitic Age puzzle accessibility". I would suggest rewording the description to "Allow solving Selenitic Age audio puzzles with more error margin".

lotharsm:
@digitall Thank you very much for your suggestion which is way better than mine :-)

I adopted it and also brought back the hotkey. I went with "~a~" since it can be easily attached to "accessibility".

digitall:
@lotharsm No problem! https://github.com/scummvm/scummvm/commit/e31558ae51ca999896beb5f0f4872971c0358cfd looks fine to me. +1 here to merge.

bluegr:
Thanks for the suggestions guys :) This looks much better now. Let's merge it then

PR #3248, opened 2021-08-08
PRINCE: Add initial support for Spanish translation

aviloria:
Add initial support for Spanish translation.


bluegr:
Thanks! Squashing to simplify history

PR #3239, opened 2021-08-04
SCUMM: Add missing MI1 Lemonhead lines

eriktorbjorn:
Apparently some lines for Lemonhead went missing when The Secret From Monkey Island was updated for the CD version. It looks like a mistake, since one of the message that's still there isn't even displayed long enough to see. This pull request re-inserts the missing lines for the English CD version. Apparently localized versions are affected too, but I don't have them so all I can do is trying to make it as easy as possible to extend the workaround for them as well.

The "Ultimate Talkie" also has the bug. This pull request does nothing to fix that. I don't even know if the missing lines were recorded.

Unfortunately, my only savegame is based on boot params, not playing. (I did replay the game recently, but that was on a tablet and frog knows where those savefiles were hidden!)

[monkey1.s22.gz](https://github.com/scummvm/scummvm/files/6930299/monkey1.s22.gz)

PR #3252, opened 2021-08-09
SCUMM: Restore old clock tower behavior to the CD version of MI1

eriktorbjorn:
By popular demand...

It was pointed out that the clock tower in The Secret of Monkey Island behaves slightly differently in the VGA CD version than it did in the VGA floppy (and presumably also EGA floppy, but I don't have that one) version. While I don't think it _completely_ ruins the joke, I agree that the old behavior is preferable.

In the floppy version, if you examine the clock tower you get a new description (there are three possible ones, with a random variation on the last one) when you examine it, but only if you've left the room in between. The reasoning, I guess, is that you've been away for a while and the clock still hasn't changed.

In the CD version, the description changes _immediately_ when you examine the clock tower, with no need to leave in between. Guybrush's remark that it's still/always 10 o'clock makes less sense. And since the description _still_ changes when you leave the room (if you've examined the clock tower) it's very likely that you'll never even see second description.

This pull request works around that by now allowing room-35-210 to increment Var[248]. (I could have patched the script when it's loaded, but since it contains text strings the offset to patch would be different depending on which version of the game you are running.) This way, it even works with the "ultimate talkie" version. Hopefully with the localized versions as well.

### DIFF
--- a/engines/scumm/script_v5.cpp
+++ b/engines/scumm/script_v5.cpp
@@ -619,6 +619,35 @@ void ScummEngine_v5::o5_add() {
 		}
 	}
 
+	// WORKAROUND: The clock tower is controlled by two variables: 163 and
+	// 247 in the floppy VGA version, 164 and 248 in the CD version. I
+	// don't know about the EGA version, but this fix only concerns the
+	// CD version.
+	//
+	// Whenever you enter the room, the first variable is cleared. It is
+	// then set if you examine the clock tower. The second variable
+	// determines which description you see, e.g. "Ten o'clock.", "Hmm.
+	// Still ten o'clock.", etc.
+	//
+	// If the first variable was set, the second is incremented when you
+	// leave the room. That means that every time you examine the clock
+	// tower, you get a new description (there are three of them, with a
+	// random variation on the last one) but only if you've been away from
+	// the room in between.
+	//
+	// But in the CD version, someone has attempted to "fix" this behavior
+	// by always incrementing the second variable when the clock tower is
+	// examined. So you don't have to leave the room in between, and if
+	// you examine the clock tower once and then leave, the second variable
+	// is incremented twice so you'll never see the second description.
+	//
+	// We restore the old behavior by adding 0, not 1, to the second
+	// variable when examining the clock tower.
+
+	if (_game.id == GID_MONKEY && vm.slot[_currentScript].number == 210 && _currentRoom == 35 && _resultVarNumber == 248 && a == 1) {
+		a = 0;
+	}
+
 	setResult(readVar(_resultVarNumber) + a);
 }
 


bluegr:
Nice! However, since we're deviating from the original behavior, it would be preferable to add a game option to toggle this

eriktorbjorn:
> it would be preferable to add a game option to toggle this

We could do that, I guess. Though where do we draw the line between things that were probably changed by mistake, and things that were changed deliberately? There have been two other recent changes that modify the CD version to be closer to the floppy version:

* https://github.com/scummvm/scummvm/pull/3211
* https://github.com/scummvm/scummvm/pull/3239

With those two, it seemed very likely to me that the changes weren't deliberate. Here I _guess_ it's less clear. But since it will easily skip the second description of the clock tower, I'm leaning towards it being an accidental change.

Another difference between the CD and floppy version (that is mention in the Captain Smirk pull request) is the color of the smoke from the voodoo lady's cauldron. If there's a reasonable way to fix that (which I currently don't know), it should probably be an option too. Particularly since the special edition uses the colors from the CD version.

Hypothetically, if I ever submitted a patch to restore the stump joke to the CD version (which I don't plan on doing), I'd _definitely_ make it an option since that joke only makes sense with floppies.

bluegr:
The two changes that you mention are features that do indeed look like they were omitted by mistake. This one looks to be intentional. 

So, we can choose either of the two options:
- Apply changes to behavior like this one
- Add a single game option for MI1 that would toggle changes to behavior of the CD version, like this one or the tree stump joke (e.g. something like "Use the floppy version logic")

To be honest, I don't mind going with either option. This is a quite specific behavioral change. The only consideration for the inclusion of a game option would be for the game purists out there.

eriktorbjorn:
> This one looks to be intentional.

It's possible - I'm not saying it isn't - but since the CD-ROM conversion (according to the opening credits) was made by a different team than the one that programmed the original game, I can also see how it could be an accident. If you just repeatedly examine the clock tower (without leaving the room in between) in the floppy version, you could easily come to the conclusion that you won't ever see anything else than the first message and decide to "fix" that in the CD version.

What makes it feel like a bug to me is the following scenario:

Enter the room and examine the clock tower: "Ten o'clock."
Leave the room and come back later to examine the clock: "Seems like it's ALWAYS ten o'clock on this island."

The only way you'll see the middle message, "Hmm.  Still ten o'clock.", is if you examine the clock tower twice without leaving the room in between. Which makes it unlikely that you'll ever see it.

In the floppy version, there's no way you see the last message without also having seen the middle one.

JohnnyWalkerDesign:
Nice job. This definitely no more intentioned then the other weird CD-ROM changes. Why would Guybrush complain that it was "ALWAYS 10 O'Clock on this island!" when only seconds had passed. I remember someone saying they'd spotted the point in the script where something had been commented out, too.

I wonder who was responsible for all these odd changes to the script.

bluegr:
After replaying this scene, I agree that the logic of the floppy version makes much more sense here. Making this a game option would be overkill. So, it's much more preferable and straightforward to adjust the logic for this in-game joke.

Thanks for your work once again! Merging

JohnnyWalkerDesign:
This has bugged me for decades. Finally ScummVM has fixed it for everyone. Thank you!

PR #3253, opened 2021-08-09
SCUMM: MI1: Use correct colors for Smirk's cigar smoke in CD version

eriktorbjorn:
This pull request is a continuation of https://github.com/scummvm/scummvm/pull/3211 and is an attempt at getting the same colors for the cigar smoke in the CD version as in the floppy version.

My reasoning here is that while we probably can't do anything about the colors in the other cases where this animation is used, and the special edition made the new colors canonical anyway, the cigar smoke has never before been seen in the CD version. Therefore, the floppy version is the canonical version in this particular case.

The CD version without this pull request:

cd-before

The CD version with this pull request:

cd-after

The floppy version:

floppy


bluegr:
Nice work, once more! I love the end result of all this :)

Merging

PR #3254, opened 2021-08-10
NEWS: Mention DiMUSE and SCUMM fixes

AndywinXp:
I didn't know I could do it before, but I've finally included my fixes (some were minor so I didn't include those) in NEWS.md, here you go.


bluegr:
Thanks!

PR #3256, opened 2021-08-10
SCI: Text To Speech (TTS) for Laurabow2 and Castle of Dr. Brain

taylorzhancher:
Tentative attempt in adding TTS to two games that are powered by SCI. 
The functions are far from perfect, but here's what we have
- GUI option that comes up only when Laurabow2/ Castle of Dr.Brain is played
- TTS for Quit game options
- TTS for subtitles that appear in the game
- TTS for some clickable options that appear in the game


engines/sci/engine/kgraphics.cpp
```diff
@@ -1210,6 +1212,22 @@ reg_t kShakeScreen(EngineState *s, int argc, reg_t *argv) {
 	return s->r_acc;
 }
 
+Common::String textKeeper[10];
```
bluegr, line 1215: Why is this needed at all?

sluicebox:
I haven't tested this out yet since it doesn't compile: https://github.com/scummvm/scummvm/pull/3256/checks

At a high level, there's two things I'd like to see:

1. All of this SCI TTS stuff moved into its own class, which the kernel functions you're hooking can then call
2. Much comments =)

Right now the implementation is spread across graphics and control code, and Text To Speech is neither of those things. Any time you're repeating big logic tests like which game this applies to, that's a sign that it should be in one place. This SCI TTS functionality has state and if you make a class then you can put it there instead of having some of it in GfxPaint16 and some of it in a global array of strings in kgraphics.cpp. And you'll probably be needing more state as you continue. SCI kernel functions are already pretty complicated, so lets keep the TTS complexity separate. (And I suspect it will get more complex...)

I have quite a few questions but most will be taken care of by first centralizing and documenting/commenting this.

- What specific text events are tied to these particular kernel functions?
- kDisplay draws text on the screen but it's also called again to remove text from the screen. How is that handled?
- Why LB2 and Castle of Doctor Brain? (Is there something special about those two that this relies on?)
- What's the plan for other SCI games?
- Why are there game-id and demo tests? The config setting can only be true for the games that have been added in the detection table.
- I see that you've included the CD (talking) version of LB2. How does that work, and what does TTS add to that game?

Step 1 though is to get this compiling =)

I'd also take a look at the GCC warnings that this code is producing and fix those.

I'm excited to hear how this works and try it out!

engines/sci/engine/kgraphics.cpp
taylorzhancher: Thanks for taking a look. 
It's for the text that appear in the question book of Laurabow2 (shown below). It's here to avoid passing the wrong/repeated text into ttsMan. The text that gets printed every time a new option gets highlighted is regularly more than the option itself.
Screen Shot 2564-08-11 at 22 19 35

taylorzhancher:
@sluicebox Thanks for looking, and sorry for the compilation problem. I hope it's working after the merge conflict is resolved. 

- What specific text events are tied to these particular kernel functions?
`reg_t kDisplay(EngineState *s, int argc, reg_t *argv)` -> the question book & opening menu options for Laurabow2 
`void GfxControls16::kernelDrawButton(Common::Rect rect, reg_t obj, const char *text, uint16 languageSplitter, int16 fontId, int16 style, bool hilite)` -> the "buttons" of "would you like to quit" pop up for the demo of Laurabow2 & Castle of Dr.Brain
`void GfxControls16::kernelDrawText(Common::Rect rect, reg_t obj, const char *text, uint16 languageSplitter, int16 fontId, TextAlignment alignment, int16 style, bool hilite) ` -> all text in boxes, including subtitles, the "prompt" of "would you like to quit" pop up for Laurabow2 & Castle of Dr.Brain
`reg_t GfxPaint16::kernelDisplay(const char *text, uint16 languageSplitter, int argc, reg_t *argv)` -> text that appear in Laurabow2 & Castle of Dr.Brain that are not in a textbox

- kDisplay draws text on the screen but it's also called again to remove text from the screen. How is that handled?
`reg_t GfxPaint16::kernelDisplay(const char *text, uint16 languageSplitter, int argc, reg_t *argv)`  check if it's a repetition of the last text
`reg_t kDisplay(EngineState *s, int argc, reg_t *argv)` Using the same way, but handled by an array

- Why LB2 and Castle of Doctor Brain? (Is there something special about those two that this relies on?)
Nothing really special. I started with them because I could see text in their demo games.

- What's the plan for other SCI games?
I think it's open. The reason I started with these is solely to their demo's accessibility

- Why are there game-id and demo tests? The config setting can only be true for the games that have been added in the detection table
the game-id test is for restricting the functions to only happen on the games I've tested and have made sure they are working. I think if a game hasn't been added to the detection table yet, it should not be working with TTS yet.

- I see that you've included the CD (talking) version of LB2. How does that work, and what does TTS add to that game?
TTS works for other text that are not subtitles in that game, such as the question book + don't ask this question prompt

When you said that you wanted TTS moved to its own class, do you mean you would like to get rid of the `if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled")` check? Or because we would like to have our code work with the Sierra game script?

bluegr:
As @sluicebox said, the TTS functionality should be placed into a self-contained class, e.g. SciTts. That class will contain all the functions with the TTS functionality, which will be hooked per case. The way it is implemented now, the TTS functionality is scattered throughout the engine, which gives no clear indication of the unit of work pattern that has been followed. The functions from that class will be hooked into the relevant places in the engine. Also, future changes to TTS will be self-contained in that class. Finally, please bear in mind that SCI has major versions, with quite different functionality. For example, text in earlier SCI16 games is handled in a different way from the newer SCI32 games

taylorzhancher:
@bluegr Thanks. I'll create a class for TTS and push as soon as possible

taylorzhancher:
I've moved all the TTS code into a new class now. Let me know if this what we have in mind.

sev-:
I believe, someone else should take this patch and wrap it into a class. @taylorzhancher seems to be busy with the studies. @bluegr ?

taylorzhancher:
@sev- I'll take ownership and try to wrap it into a class. Sorry for putting the project off for 2 months.

sev-:
@taylorzhancher any progress? :)

sluicebox:
Before @taylorzhancher puts more work into refactoring this, we should have the awkwardly overdue conversation about whether this should be merged in any form.

We may have dropped the ball on this one. I'm not sure if anyone asked an SCI dev if SCI-TTS was a feasible task for a student or anyone. And I wasn't paying attention to the TTS tasks until @taylorzhancher asked me about SCI. I said it sounded hard and I personally didn't know how it could be done. I should have been more vocal about that, but I also didn't want to discourage just because I'm too thick to figure it out. I've now spent some time with this problem.

When I look at how TTS was added to other engines, it seems straightforward: find the engine code that prints messages and hook that with TTS. There's usually not much code affected and I don't get the impression that a lot of engine-specific knowledge is required. TTS is easy to test and a great feature. It's a good student task.

But the SCI engine doesn't have a high level print-message function to hook. It only has low-level text drawing kernel functions that game scripts use in all kinds of different and interesting ways. Those are the kernel functions that have been hooked for TTS, which makes sense because they're the only ones that get passed strings. But all the high level message code is in game's own scripts. Those get customized and change between games and versions. I don't see how the interpreter (ScummVM) can ever know when it's appropriate to TTS based off that.

For example, how would this work with games like King's Quest VI (the closest relative to LB2) and King's Quest V? Those games don't draw the first letter of their strings. Their scripts strip the first letter and draw a fancy big graphic instead, like an old book, and pass the rest of the to the interpreter for normal drawing. There's no kernel function that gets passed that first letter. And the exceptions just keep coming. What about games that display countdown timers that update every cycle or second? Symbols like copy protection that are drawn as text with symbol-fonts? We haven't even touched the parser games yet. Text gets drawn to the screen in all kinds of interesting ways.

the-beach

I'm not convinced an SCI expert could fully achieve this task. If they could I suspect it would take a mountain of difficult game-specific and version-specific workarounds. And I doubt anyone would ever spend that kind of time on it.

My overall concerns:

- SCI TTS may not be feasible in the first place (there's no good place in the engine to hook)
- We can't advertise that we've added TTS support for a game unless it works well across the whole game (imagine the disappointment and predicable feedback)
- I suspect that full TTS support for any game, if possible, would require a ton of game-specific workarounds by an SCI expert, and I'm skeptical that anyone will ever do that.
- This proof of concept only affects two SCI games, of which LB2 already has a CD talkie version.
- This proof of concept is incomplete because the hard parts like the LB2 menu-intensive final act (where good TTS would be helpful) aren't handled, and the result is practically nonstop TTS. And the only reason that doesn't happen on the notebook, which still has some problems, is because of the way that script just happens to draw its selected text. All of that makes me skeptical that this is a general solution, or that anyone will build upon it.
- The current code would require quite a lot of comments explaining what it's doing for it to be acceptable. But I suspect the answers are going to be something like: "The game seems to call this function three times for every message so we only TTS on the third, except for here when it's four for some reason." And if that's the case, I don't see that being a general solution or anyone building upon that.

None of this is a reflection on @taylorzhancher's hard work. But SCI TTS is a <i>really</i> hard problem, and I'm not sure we presented it that way.

bluegr:
From my end, I believe that TTS is feasible for most of the in-game texts of SCI16 games.
It's not possible to achieve a 100% TTS coverage of all game texts. But a very large chunk of SCI game texts can be hooked, and this improves the experience for games without speech.

I'm not fond of the game-specific logic that has been added for LB2, so IMHO that should be removed.

Regarding the latest comments from @sluicebox:
- It is feasible to hook into the places where textboxes are shown, which is where most of the in-game text is shown. Some games have their own logic for displaying text, but I believe that we should aim for the general cases here
- As mentioned, TTS is a nice to have feature, which wasn't available in the original game versions. It's not feasible to add it everywhere and cover all cases - although it would be cool to do so, it's not a requirement
- We shouldn't need to add workarounds, if we are aiming for the general cases where in-game text is shown
- I will add a new PR that will target all SCI16 (SCI0 - SCI1.1 games) without speech
- I got no saved game from the LB2 endgame menus, let's check this one
- Again, I'm not fond of game-specific logic. It's easier to go for a very generalized solution

Since this PR has bitrot, I'll add a new PR with what I got in mind, and close this one.

engines/sci/engine/tts.cpp
```diff
@@ -0,0 +1,78 @@
...
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+	if (ttsMan != nullptr && ConfMan.getBool("tts_enabled") && (g_sci->getGameId() == GID_LAURABOW2 || g_sci->getGameId() == GID_CASTLEBRAIN))
+		ttsMan->say(text, Common::TextToSpeechManager::INTERRUPT);
+}
+
+const char *thelastText;
+
+void ttsDisplay(const char *text) {
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+	if (text != thelastText && ttsMan != nullptr) {
```
bluegr, line 72: This comparison is probably not what was intended - either use strcmp() here, or the != operator from Common::String

bluegr:
I've opened a new PR for this, #3623, with the following changes:
- Grouped all TTS functions under a new class, SciTTS
- Generalized the code so that it's reusable for all SCI16 games
- Removed all the game-specific logic (e.g. for LB2)
- Changed all TTS functions to use Common::String for their parameters
- Simplified ttsManager checks

Closing this PR, in favor of the other

PR #3257, opened 2021-08-10
XEEN: RU Translated Mirror aliases

Ardash:
Here is translated aliases for teleport mirror


dreammaster:
Looks good to me. Merging

PR #3258, opened 2021-08-10
SCUMM: Add italian translation for MI1 Cannibal script patch

AndywinXp:
This commit adds the italian translation for the patch implemented by @eriktorbjorn for the CD version of MI1.

### DIFF
--- a/engines/scumm/script_v5.cpp
+++ b/engines/scumm/script_v5.cpp
@@ -2875,6 +2875,11 @@ void ScummEngine_v5::printPatchedMI1CannibalString(int textSlot, const byte *ptr
 "Oooh, that's nice.\xFF\x03"
 "Simple.  Just like one of mine.\xFF\x03"
 "And little.  Like mine.";
+	} else if (strncmp((const char *)ptr, "/LH.ITA/", 8) == 0) {
+		msg =
+"Oooh, che bello.\xFF\x03"
+"Semplice.  Proprio come uno dei miei.\xFF\x03"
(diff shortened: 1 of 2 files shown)


eriktorbjorn:
Nice! I'm glad to see that adapting it for other languages was as (relatively) simple as I had hoped.

AndywinXp:
Given that I'm not really the sharpest tool in the shed, I'd say that your guide was very simple to follow and fool proof! Thanks again.

engines/scumm/script_v5.cpp
```diff
@@ -2875,6 +2875,11 @@ void ScummEngine_v5::printPatchedMI1CannibalString(int textSlot, const byte *ptr
 "Oooh, that's nice.\xFF\x03"
 "Simple.  Just like one of mine.\xFF\x03"
 "And little.  Like mine.";
+	} else if (strncmp((const char *)ptr, "/LH.ITA/", 8) == 0) {
+		msg =
+"Oooh, che bello.\xFF\x03"
+"Semplice.  Proprio come uno dei miei.\xFF\x03"
+"E piccolo.  Come il mio.";
```
bluegr, line 2882: Could you please fix the indentation here, together with the English strings above?
These should be indented one level after msg (so, two levels to the right of the if block)

AndywinXp: Sure! I indented it like that only because the english one was approved like that :P

bluegr:
Nice work, thanks! Merging

PR #3259, opened 2021-08-10
SCUMM: COMI: Fix getStringWidth behavior

AndywinXp:
While researching for another bug ([the object line being out of place in COMI](https://bugs.scummvm.org/ticket/12796)), I found a bug in how `o8_getStringWidth()` calculates the width for object lines.

Originally, ScummVM calls `translateText()` on the string (say, `/SSBT001/` for objects). This works fine when the input is dialogue string (since it contains not only the `/*/` code, but also the full string. 
And it pretty much falls apart when sending just the  `/*/` code 🤕 This means that if by any chance there's an object line whose width should exceed 640 pixel, the script looking for it will never catch that (e.g.: global script 39).

A call to `convertMessageToString()` is enough to fix that and get the expected width... almost.

For some reason, `getStringWidth()` always, always, always returns the expected width+1 for COMI. Since that code is shared with other games I think adding the `- 1` is enough for sake of accuracy.

### DIFF
--- a/engines/scumm/charset.cpp
+++ b/engines/scumm/charset.cpp
@@ -456,6 +456,56 @@ int CharsetRendererClassic::getCharWidth(uint16 chr) {
 }
 
 int CharsetRenderer::getStringWidth(int arg, const byte *text) {
+	if (_vm->_game.id == GID_CMI) {
+		int numBytesMax = 100000; // Also hardcoded in the exe
+		int maxWidth = 0;
+		int width = 0;
+
+		while (*text && numBytesMax) {
+			// Some localizations may override colors
+			// See credits in Chinese COMI
+			if (_vm->_language == Common::ZH_TWN &&
+				text[0] == '^') {
+				if (text[1] == 'c') {
+					text += 4;
+				}
+			}
+
+			while (text[0] == '^') {
+				switch (text[1]) {
+				case 'f':
+					// We should change the font on the fly at this point
+					// which would result in a different width result.
+					// This has never been observed in the game though, and
+					// as such, we don't handle it.
+					text += 4; 
+					break;
+				case 'c':
+					text += 5;
+					break;
+				default:
+					error("CharsetRenderer::getStringWidth(): Invalid escape code in text string");
+				}
+			}
+
+			if (is2ByteCharacter(_vm->_language, *text)) {
+				width += _vm->_2byteWidth + (_vm->_language != Common::JA_JPN ? 1 : 0);
+				++text;
+				--numBytesMax;
+			} else if (*text == '\n') {
+				maxWidth = MAX<int>(width, maxWidth);
+				width = 0;
+			} else if (*text != '\r' && *text != _vm->_newLineCharacter) {
+				width += getCharWidth(*text);
+			}
+
+			++text;
+			--numBytesMax;
+		}
+
+		return MAX<int>(width, maxWidth);
+	}
+
 	int pos = 0;
 	int width = 1;
 	int chr;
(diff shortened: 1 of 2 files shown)


AndywinXp:
@athrxx this change indeed lets script 39 know if a line is too large to fit on screen (and as a result puts the line at y = 415 instead of 440). 
But since we're not printing the string word by word like the original, the line doesn't actually wrap and goes out of bounds instead (an example of this is using the keg-bomb on the big yeti in the last part of COMI, at least in the italian version).

I saw that there's a function (in charset.cpp I think) which inserts line breaks if the line is too long, maybe it can be put to use into a possibile solution to this problem?

athrxx:
The original uses the same getStringWidth() method that we have in SmushFont::getStringWidth(), except that we don't handle the ^ control codes. Maybe try to copy over the code from that method and see if that solves the issue? If yes, then we can think about how to go at this...

Also, you got rid of the text translation in your PR, so you won't get the length of the translated string (or am I missing something here?)

AndywinXp:
> The original uses the same getStringWidth() method that we have in SmushFont::getStringWidth(), except that we don't handle the ^ control codes. Maybe try to copy over the code from that method and see if that solves the issue? If yes, then we can think about how to go at this...
Fair point, I'll give that a look.

> Also, you got rid of the text translation in your PR, so you won't get the length of the translated string (or am I missing something here?)
convertMessageToString actually calls for translateText for SCUMM v7 and 8

AndywinXp:
> Maybe try to copy over the code from that method and see if that solves the issue?

Yes it does!

Also, do you have any in-game (non-SMUSH related) examples of strings with ^ control codes? If any, in the english version. I suppose this is how they should be handled:

	while (str[0] == '^') {
		switch (str[1]) {
		case 'f':
		{
			fontId = str[3] - '0';
			str += 4;
		}
		break;
		case 'c':
		{
			color = str[4] - '0' + 10 *(str[3] - '0');
			str += 5;
		}
		break;
		default:
			error("invalid escape code in text string");
		}
	}

AndywinXp:
Still, `convertMessageToString()` is very much needed in place of `translateText()`

athrxx:
Honestly, I don't know if the ^codes are even used ingame. I just see that the original handles them. But that doesn't have to mean anything, since the original uses these routines also for Smush, and Smush definitely has the ^codes. We don't really handle them correctly in Smush, too. We just assume that they're always at the very beginning of the string. So we set the color and the font and then strip these codes off (see `SmushPlayer::handleTextResource()`). And there is also a hack in case the assumption is wrong... :-)

Your handling of the ^codes in your example is almost right. Only that the original immediately selects the font from the ^f code. So the following characters would have a different width. But it seems like just a theoretical issue for us, since I don't think the feature is actually used. In `CharsetRenderer::getStringWidth()` (charset.cpp, line 514) and `ScummEngine_v6::drawBlastTexts()` (string.cpp, line 234), we handle at least one special ^ case for Chinese COMI. But this actually gives me the impression that this is the only case were it ever turned up ingame. That code is not from me, so I don't know why it was needed.
So I would keep the handling of these codes simple, just like in your example...

AndywinXp:
Very well, this seems to work as expected with a big number of strings I tested.

Also, this is why the previous version gave width+1 as a result:
[https://github.com/scummvm/scummvm/blob/055135acdbe98138eec29d8655e417ed496faf8a/engines/scumm/charset.cpp#L510](https://github.com/scummvm/scummvm/blob/055135acdbe98138eec29d8655e417ed496faf8a/engines/scumm/charset.cpp#L510)

I suppose that's intentional so I'm not gonna touch that.

athrxx:
Yes, please don't touch that, unless we definitely know that it is wrong. The method seems to be adapted to work with the existing charset code, so we should be very careful.

The new commit looks good. The special handling of the COMI Chinese ^c code can probably go. This looks fishy anyway (byte size of the code), but I'll have to check that when I am not on my phone like right now.

I am also not sure if it is ideal to have the SCUMM7+ version of getStringWidth() in the common charset renderer. But that's also not something I'll figure out on my phone... :-)

AndywinXp:
Very good then, thanks :-)

athrxx:
I can merge it like this. Maybe I add some comments and clean it up a bit. But no need for further nitpicking here...
Now, I haven't done any major testing. But the dialogues in the first scene seem fine...

PR #3260, opened 2021-08-10
NEWS: Mention macOS improvements and big-endian fixes

dwatteau:
Here's some stuff I had in mind for the next release.

For Apple M1, I'm inferring that the download page for 2.3.0 will provide an M1 build which is going to be close to what's currently available in the buildbot, but that's just a guess.


bluegr:
Thanks!

criezy:
I have not yet decided whether there will be a specific M1 version for the  2.3.0 release.
The are a few complications, for example with Sparkle, and I don't have a lot of time to look at those currently.

PR #3261, opened 2021-08-10
ENGINES: Streamline auto-save write/delete protection

orgads:
Some engines call setAutosave and some don't. isAutosave is used to
determine if a saved game is an autosave, but in fact, on most cases it
just falls back to comparing the name to "Autosave".

This is wrong for several reasons:
* Older versions of ScummVM used Autosave 0.
* The name "Autosave" is translated, so if you change the language, it
  won't be detected.

Instead of relying on the name, use the well-known getAutosaveSlot() from
Engine/MetaEngine.

Fixes #12735.

### DIFF
--- a/engines/savestate.cpp
+++ b/engines/savestate.cpp
@@ -21,7 +21,9 @@
  */
 
 #include "engines/savestate.h"
+#include "engines/engine.h"
 #include "graphics/surface.h"
+#include "common/config-manager.h"
 #include "common/textconsole.h"
 #include "common/translation.h"
 
@@ -32,16 +34,24 @@ SaveStateDescriptor::SaveStateDescriptor()
 	_thumbnail(), _saveType(kSaveTypeUndetermined) {
 }
 
-SaveStateDescriptor::SaveStateDescriptor(int s, const Common::U32String &d)
-	: _slot(s), _description(d), _isDeletable(true), _isWriteProtected(false),
-	  _isLocked(false), _saveDate(), _saveTime(), _playTime(), _playTimeMSecs(0),
-	_thumbnail(), _saveType(kSaveTypeUndetermined) {
+SaveStateDescriptor::SaveStateDescriptor(int slot, const Common::U32String &d)
+	: _slot(slot), _description(d), _isLocked(false), _playTimeMSecs(0) {
+	initSaveType();
 }
 
-SaveStateDescriptor::SaveStateDescriptor(int s, const Common::String &d)
-	: _slot(s), _description(Common::U32String(d)), _isDeletable(true), _isWriteProtected(false),
-	_isLocked(false), _saveDate(), _saveTime(), _playTime(), _playTimeMSecs(0),
-	_thumbnail(), _saveType(kSaveTypeUndetermined) {
+SaveStateDescriptor::SaveStateDescriptor(int slot, const Common::String &d)
+	: _slot(slot), _description(Common::U32String(d)), _isLocked(false), _playTimeMSecs(0) {
+	initSaveType();
+}
+
+void SaveStateDescriptor::initSaveType()
+{
+	// Do not allow auto-save slot to be deleted or overwritten.
+	const bool autosave =
+			g_engine && ConfMan.getInt("autosave_period") && _slot == g_engine->getAutosaveSlot();
+	_isWriteProtected = autosave;
+	_saveType = autosave ? kSaveTypeAutosave : kSaveTypeRegular;
+	_isDeletable = !autosave;
 }
 
 void SaveStateDescriptor::setThumbnail(Graphics::Surface *t) {
--- a/engines/savestate.cpp
+++ b/engines/savestate.cpp
@@ -44,8 +44,7 @@ SaveStateDescriptor::SaveStateDescriptor(int slot, const Common::String &d)
 	initSaveType();
 }
 
-void SaveStateDescriptor::initSaveType()
-{
+void SaveStateDescriptor::initSaveType() {
 	// Do not allow auto-save slot to be deleted or overwritten.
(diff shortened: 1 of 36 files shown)


lephilousophe:
getAutosaveSlot() tells where autosaves are supposed to be saved.
If a user has saved on the autosave slot, it won't be an autosave and I am not sure what will happen with your changes

orgads:
Most (all?) engines write-protect and delete-protect this entry, so the user cannot save on it.

orgads:
Rebased to trigger CI.

orgads:
Ok, I discovered the global `g_engine`, and simplified this patch a lot. Please review.

lephilousophe:
> Most (all?) engines write-protect and delete-protect this entry, so the user cannot save on it.

This is not the case on at least Versailles. I always thought that autosaving support were optional and never did anything to implement it.
This is a quite a huge change so I think some stance should be taken on this.

orgads:
> This is a quite a huge change so I think some stance should be taken on this.

It's much smaller now 🙂

Did you see the latest revision?

engines/savestate.cpp
```diff
@@ -32,16 +33,23 @@ SaveStateDescriptor::SaveStateDescriptor()
...
-	_thumbnail(), _saveType(kSaveTypeUndetermined) {
+SaveStateDescriptor::SaveStateDescriptor(int slot, const Common::String &d)
+	: _slot(slot), _description(Common::U32String(d)), _isLocked(false), _playTimeMSecs(0) {
+	initSaveType();
+}
+
+void SaveStateDescriptor::initSaveType()
+{
+	// Do not allow auto-save slot to be deleted or overwritten.
+	const bool autosave = g_engine ? _slot == g_engine->getAutosaveSlot() : false;
```
lephilousophe, line 49: This must take autosave_period configuration into account like it's done in MetaEngine::listSaves

orgads: I thought about it, but I think it's better not to. Consider the following use-case:

* Autosave is disabled - slot 0 is a regular slot.
* The user happily saves on this slot.
* After several weeks, the user discovers the autosave feature and enabled it.
* 5 minutes into the game and Poof - the saved game is gone!

sev-: In this case it would be best to warn the user when they enable this option, something like: "The autosave uses slot XX, it will be lost. Proceed?"

orgads: Good point. So when enabling this option, I'll iterate over all the engines and check if they have a saved game which is not named "Autosave" on the autosave slot, and warn on this case?

sev-: Exactly, something like this.

sev- (approved):
This is a good and simple change, and we could still squeeze it into the release because it is easy to check and it is obvious when it is broken or not.

So, please, implement more sanity check when switching the option and we merge it.

engines/savestate.cpp
orgads: I have another idea, which I'm working on.

When autosave is triggered, I check if this slot is occupied by a non-autosave entry. If it is, I prompt the user to choose between:
* Move the currently saved game to a free slot
* Overwrite
* Disable autosave (for this session)

A similar approach can also be useful to prevent accidental overwrite of autosave. For example, if the existing autosave was saved after 3 hours of gameplay, then the user started a new game and autosave was triggered, we can warn before overwriting the existing one.

What do you say?

orgads: Done

engines/savestate.cpp
```diff
@@ -32,16 +33,23 @@ SaveStateDescriptor::SaveStateDescriptor()
...
-	: _slot(s), _description(Common::U32String(d)), _isDeletable(true), _isWriteProtected(false),
-	_isLocked(false), _saveDate(), _saveTime(), _playTime(), _playTimeMSecs(0),
-	_thumbnail(), _saveType(kSaveTypeUndetermined) {
+SaveStateDescriptor::SaveStateDescriptor(int slot, const Common::String &d)
+	: _slot(slot), _description(Common::U32String(d)), _isLocked(false), _playTimeMSecs(0) {
+	initSaveType();
+}
+
+void SaveStateDescriptor::initSaveType()
+{
```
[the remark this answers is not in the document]
orgads: Done

orgads:
Factored out the prompt as a function, and changed the check to be contains("Autosave") instead of equals, because previous versions of ScummVM saved it as "Autosave 0".

orgads:
@sev- I have another pending commit for warning on options change. https://github.com/orgads/scummvm/tree/autosave-improved

Waiting for final approval of this PR before opening a new PR for that.

macca8:
Activating a warning message if the slot is deemed to be occupied by a non-autosave saved game at the time the autosave is triggered is a great inclusion, however, allowing the player to disable autosaving as an option seems odd. 
Isn't the objective here to encourage the player to take some action, when necessary, to allow the autosave to proceed?

At worst, I'd recommend that 'Cancel autosave' should only cancel the current autosave (the message could be simplified to "or cancel this autosave."). By default, failed autosaves are repeated at 5 minute intervals (which covers all other 'saveFlag = false' instances). It makes sense for this option to follow the same pattern.

orgads:
Thanks for the feedback.

Pressing Escape cancels the current save, but will prompt again next time. I know this is not so discoverable :)

Do you think it's worth adding another button for this action? I don't want to clutter the dialog too much.

sev-:
Thank you, merging.

@macca8 temporary disabling autosave for a session when there is a problem seems like the most logical approach to me, instead of getting nagged every 5 minutes.

orgads:
Great! Follow-up PR: #3304

PR #3262, opened 2021-08-10
SCUMM: DiMUSE: Implement actual setVolume and setPan constraints

AndywinXp:
I hope this is gonna be the last PR for the old Digital iMUSE system, I really want to get the new one done soon.
This commit implements constraints for setPan and setVolume (see [here](https://github.com/AndywinXp/Digital-iMULATOR/blob/57255e3a4637dfbc7cff61e700ef3fa505a4fcbd/iMUSE/tracks.c#L369)). 
The decomp ignores any attempt to set a volume or a pan greater than 127 so we do that. In ScummVM (because of how these parameters are then handled by the Mixer class) we also have to account for negative values which, at least for panning, do appear sometimes.

As a result, all three DiMUSE games behave as they should, and I also managed to delete some nasty COMI hacks :-)


bluegr:
Thanks! Merging

PR #3264, opened 2021-08-11
Fix HiDPI scaling on macOS and iOS

criezy:
The main purpose of this pull request is to fix issues with the HiDPI handling on macOS and iOS.

On macOS the main issue is that on some non-retina screens, the code in master currently think that the screen is a HiDPI screen with a scale of 2. This results in various issues, such as:
 - the GUI appearing by default in lowres layout despite the window being big enough for the normal layout
 - the window get smaller and smaller each time we close and restart ScummVM (https://bugs.scummvm.org/ticket/12590)

On iOS the issue is the other way around. The code in master currently assume the HiDPI scaling is 2. But on some iOS device it can be 3, and that causes the GUI to be too small by default (and it makes it difficult to go into the option to change the GUI scale).

Both are fixed by adding a new `float getHiDPIScreenFactor()` to `OSystem` and implementing it on both iOS and macOS using the system API. On other SDL backends the implementation is the same as in master (looks at the dpi and returns 1 or 2).

I also added an implementation in the Android backend, and it currently always return 2. This should preserver the current behaviour from master, but I suspect the scaling should actually depend on the device (as on iOS). A developer familiar with that backend should take a look to implement it properly. According to https://developer.android.com/training/multiscreen/screendensities, it seems that DisplayMetrics.density might gives us what we want:
> On a medium-density screen, DisplayMetrics.density equals 1.0; on a high-density screen it equals 1.5; on an extra-high-density screen, it equals 2.0; and on a low-density screen, it equals 0.75. This figure is the factor by which you should multiply the dp units in order to get the actual pixel count for the current screen.

I tested this on macOS on both a retina laptop and a non-retina one, and both are now working properly. In particular I confirmed that this fixes [bug #12590](https://bugs.scummvm.org/ticket/12590).

I have not yet tested on iOS, but since we have limited time before the next release I decided to open this pull request as soon as possible so that others can have a look and test it while I confirm that the behaviour on iOS is also fixed.

Finally I have added one last commit that is not required but in my opinion improves the behaviour of the GUI scale option. Instead of indicating a base resolution, it indicates a scaling. This scaling is multiplied with the HiDPI scaling, and is 100% by default. This change means that when resizing the window it now preserves the text size instead of making it bigger or smaller. This might also work better on some devices with unusual sizes since it no longer relies on a small set of hardcoded base resolutions.

### DIFF
--- a/backends/graphics/graphics.h
+++ b/backends/graphics/graphics.h
@@ -101,6 +101,7 @@ public:
 	virtual void copyRectToOverlay(const void *buf, int pitch, int x, int y, int w, int h) = 0;
 	virtual int16 getOverlayHeight() const = 0;
 	virtual int16 getOverlayWidth() const = 0;
+	virtual float getHiDPIScreenFactor() const { return 1.0f; }
 
 	virtual bool showMouse(bool visible) = 0;
 	virtual void warpMouse(int x, int y) = 0;
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -251,6 +251,12 @@ bool OpenGLSdlGraphicsManager::getFeatureState(OSystem::Feature f) const {
 	}
 }
 
+float OpenGLSdlGraphicsManager::getHiDPIScreenFactor() const {
+	float scale;
+	getDpiScalingFactor(&scale);
+	return scale;
+}
+
 void OpenGLSdlGraphicsManager::initSize(uint w, uint h, const Graphics::PixelFormat *format) {
 	// HACK: This is stupid but the SurfaceSDL backend defaults to 2x. This
 	// assures that the launcher (which requests 320x200) has a reasonable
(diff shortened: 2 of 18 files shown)


dwatteau:
Hi @criezy,

I don't know if this other problem is known/related or if your PR is supposed to fix it, but I'm still seeing this after switching to the OpenGL graphics mode on your branch:

criezy-hidpi-opengl-renderer-retina

criezy-hidpi-opengl-renderer-retina2

i.e. the text becomes almost unreadable.

It's OK with the default SDL renderer (although I need to restart ScummVM, otherwise nothing happens, while the change from OpenGL to SDL is immediate), though.

iMac 2017 5K running Mojave 10.14.6 (18G9323), SDL2 2.0.16 from Homebrew.

(If this needs to be addressed in a different issue instead, I'll do it.)

bluegr:
Overall, nice work! This should still work as before, with a better interface.

One suggestion: please change `getDpiScalingFactor()` so that it actually returns a value, instead of modifying an input variable by reference

criezy:
I have now tested this on iOS, and it works nicely.
I also tested on a retina mac with a non-retina screen attached as a secondary screen, and this works very well on both screens. Even moving the window from one screen to the other works well.

The behaviour on other platforms (e.g. Windows, Linux, Android) should not be changed by this pull request. So I think it is good to go and I am planning to merge this on Friday evening (before the code freeze for the release).

> **EDIT:** Ignore this, sorry. I didn't know that the fonts where part of the "Extra" paths and not the "Theme" paths, and so the good fonts weren't loaded.

I actually had the same issue yesterday when testing on my non-retina mac. I was wondering why the fonts were not getting scaled when changing the GUI scale option, until I realised that the extra path was not set. 🤦 

> One suggestion: please change `getDpiScalingFactor()` so that it actually returns a value, instead of modifying an input variable by reference

I though about it when I changed the type of that variable. I have no idea why it was implemented that way to start with, but this actually turned out to be useful since this resulted in an error for the one place where I forgot to change the `scale` from an `uint` to a `float`. If it was returning the value instead, that would have been a warning at best. So you could argue that the current way is actually more robust (you cannot use the wrong type by mistake).

But I still agree that changing it would probably be best as it is more in line with the usual practice. I will add one commit for that.

criezy:
I will merge this now.

PR #3265, opened 2021-08-11
WIN32: Remove F_OK/R_OK/W_OK hack

carlo-bramini:
Into `backends/fs/windows/windows-fs.cpp`, the macros `F_OK`/`R_OK`/`W_OK` are not defined by system includes of MSVC, so they have been added manually into the code.
While this solution works, in my opinion it would be much cleaner to use `GetFileAttributes()` for getting this information.
Actually, this is what the `_access()`/`_waccess()` functions do and, afterall, this is a piece of code expected to work on Windows only.

### DIFF
--- a/backends/fs/windows/windows-fs.cpp
+++ b/backends/fs/windows/windows-fs.cpp
@@ -28,31 +28,21 @@
 #include "backends/fs/windows/windows-fs.h"
 #include "backends/fs/stdiostream.h"
 
-// F_OK, R_OK and W_OK are not defined under MSVC, so we define them here
-// For more information on the modes used by MSVC, check:
-// http://msdn2.microsoft.com/en-us/library/1w06ktdy(VS.80).aspx
-#ifndef F_OK
-#define F_OK 0
-#endif
-
-#ifndef R_OK
-#define R_OK 4
-#endif
-
-#ifndef W_OK
-#define W_OK 2
-#endif
-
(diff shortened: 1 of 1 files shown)


carlo-bramini:
This is an an additional information about file attributes, taken from _chmod()/_wchmod() at:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/chmod-wchmod?view=msvc-160

immagine

lotharsm:
At the first glance, this looks like a good change to me. However, I recommend that someone with more experience approves this as well.

I have two comments:

1. Commit guidelines: Please prefix the commits with `WIN32:`, e.g. `WIN32: Remove F_OK/R_OK/W_OK hack` to match our commit guidelines.
2. Does `GetFileAttributes()` have aby requirements regarding the Windows version, e.g. being limited to the Windows NT series? Since we even have support for Win9x, `GetFileAttributes` should be available there as well.

carlo-bramini:
Hello, thank you for the reply.
I changed the commit message according to your suggestion.
The `GetFileAttributes()` is also supported on Windows 95/98/ME, at least until you won't build an UNICODE executable since those operating systems just support ASCII APIs.

bluegr:
Nice work @carlo-bramini! @lotharsm asked you to change the commit message, not the description of the pull request. Squashing to fix the commit message

PR #3266, opened 2021-08-11
BACKENDS: OPENGLSDL: Fix logic errors in window size management

lotharsm:
The current implementation of of checking if we have a maximized window was pretty unreliable. This lead to the ScummVM window randomly being started up with the dimensions of the maximized window, effectively rendering the image too large.

The main issue was that `_windowIsMaximized` was never properly set _unless_ a previous resizing event occured. This is fixed by getting rid of this helper variable and instead querying the return values of `SDL_WINDOW_MAXIMIZED` at runtime.

I wasn't able to reproduce any issues with this, but since we are pretty close to branching off for the next release, I'm submitting this as PR.

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -348,6 +348,7 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 	// Fetch current desktop resolution and determining max. width and height
 	Common::Rect desktopRes = _window->getDesktopResolution();
 
+#if SDL_VERSION_ATLEAST(2, 0, 0)
 	if ((SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED) && ConfMan.hasKey("window_maximized_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("window_maximized_height", Common::ConfigManager::kApplicationDomain)) {
 		// Set the window size to the values stored when the window was maximized
 		// for the last time.
@@ -374,6 +375,18 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 		ConfMan.setInt("last_window_height", requestedHeight, Common::ConfigManager::kApplicationDomain);
 		ConfMan.flushToDisk();
 	}
+
+#else
+		// Set the basic window size based on the desktop resolution
+		// since we cannot reliably determine the current window state
+		// on SDL1.
+		requestedWidth  = desktopRes.width()  * 0.3f;
+		requestedHeight = desktopRes.height() * 0.4f;
+
+		// Apply scaler
+		requestedWidth  *= _graphicsScale;
+		requestedHeight *= _graphicsScale;
+#endif
 
 	// Determine current aspect ratio
 	uint maxAllowedWidth   = desktopRes.width();
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -302,17 +302,13 @@ void OpenGLSdlGraphicsManager::notifyResize(const int width, const int height) {
 		currentWidth /= scale;
 		currentHeight /= scale;
 #endif
-		// Reset maximized flag
-		_windowIsMaximized = false;
 
 		// Check if the ScummVM window is maximized and store the current
 		// window dimensions.
-		if ((SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED) == 128) {
-			_windowIsMaximized = true;
+		if (SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED) {
 			ConfMan.setInt("window_maximized_width", currentWidth, Common::ConfigManager::kApplicationDomain);
 			ConfMan.setInt("window_maximized_height", currentHeight, Common::ConfigManager::kApplicationDomain);
 		} else {
-			_windowIsMaximized = false;
 			ConfMan.setInt("last_window_width", currentWidth, Common::ConfigManager::kApplicationDomain);
 			ConfMan.setInt("last_window_height", currentHeight, Common::ConfigManager::kApplicationDomain);
 		}
@@ -352,13 +348,13 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 	// Fetch current desktop resolution and determining max. width and height
 	Common::Rect desktopRes = _window->getDesktopResolution();
 
-	if (_windowIsMaximized == true) {
+	if ((SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED) && ConfMan.hasKey("window_maximized_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("window_maximized_height", Common::ConfigManager::kApplicationDomain)) {
 		// Set the window size to the values stored when the window was maximized
-		// for the last time. We also need to reset any scaling here.
+		// for the last time.
 		requestedWidth  = ConfMan.getInt("window_maximized_width", Common::ConfigManager::kApplicationDomain);
 		requestedHeight = ConfMan.getInt("window_maximized_height", Common::ConfigManager::kApplicationDomain);
 
-	} else if (ConfMan.hasKey("last_window_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("last_window_height", Common::ConfigManager::kApplicationDomain)) {
+	} else if (!(SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED) && ConfMan.hasKey("last_window_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("last_window_height", Common::ConfigManager::kApplicationDomain)) {
 		// Restore previously stored window dimensions.
 		requestedWidth  = ConfMan.getInt("last_window_width", Common::ConfigManager::kApplicationDomain);
 		requestedHeight = ConfMan.getInt("last_window_height", Common::ConfigManager::kApplicationDomain);
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -348,6 +348,9 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 	// Fetch current desktop resolution and determining max. width and height
 	Common::Rect desktopRes = _window->getDesktopResolution();
 
+	// Determine current desktop aspect ratio
+	float ratio = (float)desktopRes.width() / (float)desktopRes.height();
+
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	bool _isMaximized = (SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED);
 	if (_isMaximized && ConfMan.hasKey("window_maximized_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("window_maximized_height", Common::ConfigManager::kApplicationDomain)) {
@@ -364,8 +367,8 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 	} else {
 		// Set the basic window size based on the desktop resolution
 		// since we have no values stored, e.g. on first launch.
-		requestedWidth  = desktopRes.width()  * (1.0f / 4.0f) * 1.2f;
-		requestedHeight = desktopRes.height() * (1.0f / 3.0f) * 1.2f;
+		requestedWidth  = desktopRes.width()  * (1.0f / 3.0f) * 1.5f / ratio;
+		requestedHeight = desktopRes.height() * (1.0f / 4.0f) * 1.5f;
 
 		// Apply scaler
 		requestedWidth  *= _graphicsScale;
@@ -381,18 +384,17 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 		// Set the basic window size based on the desktop resolution
 		// since we cannot reliably determine the current window state
 		// on SDL1.
-		requestedWidth  = desktopRes.width()  * (1.0f / 4.0f) * 1.2f;
-		requestedHeight = desktopRes.height() * (1.0f / 3.0f) * 1.2f;
+		requestedWidth  = desktopRes.width()  * (1.0f / 3.0f) * 1.5f / ratio;
+		requestedHeight = desktopRes.height() * (1.0f / 4.0f) * 1.5f;
 
 		// Apply scaler
 		requestedWidth  *= _graphicsScale;
 		requestedHeight *= _graphicsScale;
 #endif
 
-	// Determine current aspect ratio
+	// Set allowed dimensions
 	uint maxAllowedWidth   = desktopRes.width();
 	uint maxAllowedHeight  = desktopRes.height();
-	float ratio = (float)requestedWidth / (float)requestedHeight;
 
 	// Check if we request a larger window than physically possible,
 	// e.g. by starting with additional launcher parameters forcing
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -360,7 +360,7 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 		requestedHeight = ConfMan.getInt("window_maximized_height", Common::ConfigManager::kApplicationDomain);
 
 	} else if (!_isMaximized && ConfMan.hasKey("last_window_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("last_window_height", Common::ConfigManager::kApplicationDomain)) {
-		// Restore previously stored window dimensions.
+		// Load previously stored window dimensions.
 		requestedWidth  = ConfMan.getInt("last_window_width", Common::ConfigManager::kApplicationDomain);
 		requestedHeight = ConfMan.getInt("last_window_height", Common::ConfigManager::kApplicationDomain);
 
@@ -384,6 +384,14 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 		requestedHeight = MIN<uint>(requestedWidth * 3 / 4, desktopRes.height());
 
 #endif
+
+	// In case we detected an increase in the requested window dimensions (e.g. when launching
+	// a game in 800x600 while having a smaller screen size stored in the configuration file),
+	// we override the window dimensions with the "real" request.
+	if ((requestedWidth < _lastRequestedWidth || requestedHeight < _lastRequestedHeight) && ConfMan.getActiveDomain()) {
+		requestedWidth  = _lastRequestedWidth  * _graphicsScale;
+		requestedHeight = _lastRequestedHeight * _graphicsScale;
+	}
 
 	// Set allowed dimensions
 	uint maxAllowedWidth   = desktopRes.width();
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -383,9 +383,6 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 		requestedWidth  = MAX<uint>(desktopRes.width() / 2, MIN<uint>(640, desktopRes.width()));
 		requestedHeight = MIN<uint>(requestedWidth * 3 / 4, desktopRes.height());
 
-		// Apply scaler
-		requestedWidth  *= _graphicsScale;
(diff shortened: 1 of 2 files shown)


lotharsm:
Great, `SDL_GetWindowFlags` is SDL2-only...

lotharsm:
Thanks for your review, @criezy!

> The changes in this pull request make sense, although I made a minor suggestion.
> 
> I am however wondering about two things I noticed:
> 
>     1. Do we need to store and restore the maximized window size? If the window is maximized I would think that we do not want to change the window size in `loadVideoMode`, and thus setting the requested size to the current SDL window size (that you can get with `getWindowSizeFromSdl()`) should work?
>        Edit: storing the size in `ConfMan` at least means we don't need to repeat the `__APPLE__` hack in two different places. So maybe this is not a bad idea after all.

I tried this before, and I never managed to get it "right" there - setting the window size in maximized was always broken for me. If we should get rid of this config values, I'll obviously try again ;)

> 
>     2. Why are the default width and height 60% and 80% of the desktop size respectively (taking into account that the default `_graphicsScale` is 2)? That results in a square ScummVM window on my desktop mac, which is strange since most game have a 4:3 ratio. On my laptop the screen is a bit wider (16:10 - 2880x1800), but the window resulting window is still more narrow than I would expect (12:10 - 1728x1440). If we assume that 16:10 is a common ratio for computer screens, you would need to multiply the desktop width by 1/3 and not 0.3 to get a 4:3 ScummVM window.

Yes, that's totally right. While working on the aspect ratio calculation, it seems my brain ran into a deadlock. I reworked it and now the window is _always_ drawn in a 4:3 aspect ratio, since as you said, this is basically the common denominator for the games of that era.

The additional scaling factor of 1.2 is completely arbitrary here, since at least on Low-Res screens, the window is drawn a tad too small IMHO.

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -352,13 +348,14 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
...
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+	if ((SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED) && ConfMan.hasKey("window_maximized_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("window_maximized_height", Common::ConfigManager::kApplicationDomain)) {
 		// Set the window size to the values stored when the window was maximized
-		// for the last time. We also need to reset any scaling here.
+		// for the last time.
 		requestedWidth  = ConfMan.getInt("window_maximized_width", Common::ConfigManager::kApplicationDomain);
 		requestedHeight = ConfMan.getInt("window_maximized_height", Common::ConfigManager::kApplicationDomain);
 
-	} else if (ConfMan.hasKey("last_window_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("last_window_height", Common::ConfigManager::kApplicationDomain)) {
+	} else if (!(SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED) && ConfMan.hasKey("last_window_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("last_window_height", Common::ConfigManager::kApplicationDomain)) {
```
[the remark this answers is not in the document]
lotharsm: Yes, good idea indeed. I changed this so it now uses a variable instead and only queries the value once.

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -379,10 +377,22 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 		ConfMan.flushToDisk();
 	}
 
+#else
+		// Set the basic window size based on the desktop resolution
+		// since we cannot reliably determine the current window state
+		// on SDL1.
+		requestedWidth  = desktopRes.width()  * ((float)1 / (float)4) * 1.2f;
```
bluegr, line 384:
```suggestion
		requestedWidth  = desktopRes.width()  * (1.0f / 4.0f) * 1.2f;
```

lotharsm:
@bluegr Thanks for your comments - I applied the fixes you proposed :-)

I must confess that my previous method of calculating the aspect ratio was wrong - _again_. It worked on my 16:9 screen at work, but here at home with my low-res 16:10, I get greeted with an aspect ratio of 1,2. Turns out my lack of a proper methology hit me again - it doesn't make sense to simply base everything on the desktop resolution since then you simply carry around the "wrong" aspect ratio.

The new logic _should_ now properly draw a 4:3 window no matter of the desktop resolution or aspect ratio. I tested on various desktop resolutions with various aspect ratios - the only thing left is to try it on a HiDPI display.

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -352,22 +348,27 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
...
 		requestedWidth  = ConfMan.getInt("last_window_width", Common::ConfigManager::kApplicationDomain);
 		requestedHeight = ConfMan.getInt("last_window_height", Common::ConfigManager::kApplicationDomain);
 
 	} else {
 		// Set the basic window size based on the desktop resolution
 		// since we have no values stored, e.g. on first launch.
-		requestedWidth  = desktopRes.width()  * 0.3f;
-		requestedHeight = desktopRes.height() * 0.4f;
+		requestedWidth  = desktopRes.width()  * (1.0f / 3.0f) * 1.5f / ratio;
+		requestedHeight = desktopRes.height() * (1.0f / 4.0f) * 1.5f;
```
criezy, line 372: Isn't  that more complicated than needed?
Unless I am mistaken, your code is equivalent to:
```c++
requestedHeight = desktopRes.height() * 3 / 8;
requestedWidth = requestedHeight * 4 / 3;
```
Note: I also changed it to use integer computations to avoid float to int conversions (since `requestedWidth` and `requestedHeight` are integers).

But you could try to handle a bit better small screens as well with a minimum width of 640 (or the screen width if smaller):
```c++
requestedWidth  = MAX(desktopRes.width() / 2, MIN(640, desktopRes.width()));
requestedHeight = MIN(requestedWidth * 3 / 4, desktopRes.height());
```

I am however wondering if we should really ignore the `requestedWidth` and `requestedHeight` passed to the function. These might for example be the size of the game when starting a game (and if we start the game from the command line with an empty ini file, we will hit this case). So I am thinking that we should use a size that:
 - Has the same aspect ratio than the provided ones (rather than force 4:3).
 - Is at least as big as the given size (provided it fits on the screen).

This last lines also applies to the case where we have a `last_window_width` and `last_window_height` in ConfMan. I think they should only be used if they are bigger than the requested size.

One suggestion for the default (if the window is not maximized and there is no `last_window_width` and `last_window_height` in ConfMan) would be to use the multiple of the requested size passed to the function that is closest to the half screen size, with a minimum of 640:
```c++
frac_t ratio = intToFrac(requestedWidth) / requestedHeight;
int scale = MIN(1, (desktopRes.width() / 2 + requestedWidth / 2) / requestedWidth);
requestedWidth = MAX(requestedWidth * scale, MIN(640, desktopRes.width()));
requestedHeight = MIN(intToFrac(requestedWidth) / ratio, desktopRes.height());

Comment (lotharsm) [reply backends/graphics/openglsdl/openglsdl-graphics.cpp:372]: > Isn't that more complicated than needed?
> Unless I am mistaken, your code is equivalent to:
> 
> ```c++
> requestedHeight = desktopRes.height() * 3 / 8;
> requestedWidth = requestedHeight * 4 / 3;
> ```
> 
> Note: I also changed it to use integer computations to avoid float to int conversions (since `requestedWidth` and `requestedHeight` are integers).
> 
> But you could try to handle a bit better small screens as well with a minimum width of 640 (or the screen width if smaller):
> 
> ```c++
> requestedWidth  = MAX(desktopRes.width() / 2, MIN(640, desktopRes.width()));
> requestedHeight = MIN(requestedWidth * 3 / 4, desktopRes.height());
> ```
> 

Yes, indeed this is a great idea - thank you! I'll adapt the code accordingly.

> I am however wondering if we should really ignore the `requestedWidth` and `requestedHeight` passed to the function. These might for example be the size of the game when starting a game (and if we start the game from the command line with an empty ini file, we will hit this case). So I am thinking that we should use a size that:
> 
>     * Has the same aspect ratio than the provided ones (rather than force 4:3).
> 
>     * Is at least as big as the given size (provided it fits on the screen).
> 
> 
> This last lines also applies to the case where we have a `last_window_width` and `last_window_height` in ConfMan. I think they should only be used if they are bigger than the requested size.

I'm not sure about this since the main idea I had while implementing this originally was to avoid any "size switching" and only stick to the dimensions either calculated initially or set by the user. I see that we shouldn't disrespect the `requestedWidth` and `requestedHeight` values, but for now, the only downside (unless you use the "Stretch to window" option) is that you might notice some black borders around games that have an "odd" resolution.

> 
> One suggestion for the default (if the window is not maximized and there is no `last_window_width` and `last_window_height` in ConfMan) would be to use the multiple of the requested size passed to the function that is closest to the half screen size, with a minimum of 640:
> 
> ```c++
> frac_t ratio = intToFrac(requestedWidth) / requestedHeight;
> int scale = MIN(1, (desktopRes.width() / 2 + requestedWidth / 2) / requestedWidth);
> requestedWidth = MAX(requestedWidth * scale, MIN(640, desktopRes.width()));
> requestedHeight = MIN(intToFrac(requestedWidth) / ratio, desktopRes.height());
> ```

The main issue here is that `desktopRes.*` returns the full desktop resolution, ignoring any space used by e.g. the Windows taskbar or window borders. When using this code, this leads to the image being drawn too large on high-res games unfortunately, and especially on lower resolutions.

Looking at the upcoming release and feature-freeze, I'm worried to run out of time here. The current upstream `master` branch is clearly broken here, so for me it has highest priority to improve the situation, even if the solution is not perfect :(

Comment (criezy) [reply backends/graphics/openglsdl/openglsdl-graphics.cpp:372]: > > I am however wondering if we should really ignore the `requestedWidth` and `requestedHeight` passed to the function. These might for example be the size of the game when starting a game (and if we start the game from the command line with an empty ini file, we will hit this case). So I am thinking that we should use a size that:
> > 
> > * Has the same aspect ratio than the provided ones (rather than force 4:3).
> > * Is at least as big as the given size (provided it fits on the screen).
> >   
> > This last lines also applies to the case where we have a `last_window_width` and `last_window_height` in ConfMan. I think they should only be used if they are bigger than the requested size.
> 
> I'm not sure about this since the main idea I had while implementing this originally was to avoid any "size switching" and only stick to the dimensions either calculated initially or set by the user. I see that we shouldn't disrespect the `requestedWidth` and `requestedHeight` values, but for now, the only downside (unless you use the "Stretch to window" option) is that you might notice some black borders around games that have an "odd" resolution.

I agree that at this point we can aim for something that at least works and look a improving it further later. The only point that I am worried about is my second point. If your current ScummVM window size is 640x480 and you are starting a game with a 800x600 resolution for example, I think we want to resize the window to be at least 800x600. But maybe I am missing something and this is already happening somewhere else?

Comment (lotharsm) [reply backends/graphics/openglsdl/openglsdl-graphics.cpp:372]: Ah, no, this is not happening. If you start with a 640x480 window, then your game will be downscaled.

**Comment by lotharsm** [comment]
@criezy I adopted the code you provided in order to calculate the initial window dimensions. Now, no matter of the screen resolution, we should always end up with a true 4:3 window as common denominator until we get the fully dynamic size calculation right.

Now the only thing left is to prevent downscaling as you said - any idea on how I can check for the resolution the game requests _itself_?

**Comment by criezy** [comment]
> Now the only thing left is to prevent downscaling as you said - any idea on how I can check for the resolution the game requests _itself_?

I am assuming this is the `requestedWidth` and `requestedHeight` passed to the `loadVideoMode()`? (which is why I suggested they should not be completely ignored)

**Comment by lotharsm** [comment]
The main issue here is that the ScummVM launcher itself initially always reports requested dimensions of 320x200. Then, e.g. when you either start a game via the Launcher or via the command line, it goes through various `loadVideoMode()` calls until it finally passes the correct dimensions :/

So for now, I don't really have an idea on how to "scan" for the correct value. One possibility is to compare the values stored in `ConfMan` with the `requestedWidth` and `requestedHeight` and brutally overwrite it...

**Comment by criezy** [comment]
How about what I suggested earlier: use the `requestedWidth` and `requestedHeight` values we get only if they are bigger than the values in ConfMan (i.e. don't make the window smaller, but make it bigger if needed)?

**Comment by lotharsm** [comment]
Ah, yes, sorry if I wasn't clear - that's the problem I'm trying to solve - the plan was to use your proposed solution :-)

**Comment by criezy** [comment]
Also if you want to know if the size comes from a game or the launcher, you can probably just check if an engine is running, and there are a couple of ways to do that:
 - Check `ConfMan.getActiveDomain()`. This is a nullptr when an engine is not running.
 - Check `g_engine`. This is a nullptr when an engine is not running.

**Comment by lotharsm** [comment]
I think I got it now - thank you so much for your patience :-)

File: backends/graphics/openglsdl/openglsdl-graphics.cpp [hunk 4]
```diff
@@ -352,37 +348,54 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
...
+		// on SDL1.
+		requestedWidth  = MAX<uint>(desktopRes.width() / 2, MIN<uint>(640, desktopRes.width()));
+		requestedHeight = MIN<uint>(requestedWidth * 3 / 4, desktopRes.height());
+
+#endif
+
+	// In case we detected an increase in the requested window dimensions (e.g. when launching
+	// a game in 800x600 while having a smaller screen size stored in the configuration file),
+	// we override the window dimensions with the "real" request.
+	if ((requestedWidth < _lastRequestedWidth || requestedHeight < _lastRequestedHeight) && ConfMan.getActiveDomain()) {
```
criezy, line 391:
```suggestion
	if ((requestedWidth < _lastRequestedWidth * _graphicsScale || requestedHeight < _lastRequestedHeight * _graphicsScale) && ConfMan.getActiveDomain()) {
```

This only matters for game with a width <= 320 (as then `_graphicsScale` is 2) and when the window is bigger than the requested size but smaller than double the requested size. In such a case we could say that the current size is fine (and not double the size), which is what your code is doing. But then we probably want to also change the two lines below to not use the `_graphicsScale`.

The current inconsistency using `_graphicsScale` in the two lines below, but not in the `if` condition, mean a strange behaviour where a smaller current window can result in a bigger window.
| current size | requested size | final size |
|-------------|----------------|----------|
|   310x200    |      320x200      | 640x400|
|   340x220    |      320x200     | 340x220|

So either we use the `_graphicsScale` in both places, or in neither.

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -352,37 +348,54 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 	// Fetch current desktop resolution and determining max. width and height
 	Common::Rect desktopRes = _window->getDesktopResolution();
 
-	if (_windowIsMaximized == true) {
+	// Determine current desktop aspect ratio
+	float ratio = (float)desktopRes.width() / (float)desktopRes.height();
```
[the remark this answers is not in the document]
criezy: Ha, I see now that it is still used, as you removed the other ratio computed from the computed size instead.
But we could argue that if we need to resize the window because the computed size is too big, then we should probably use either the computed ratio, or the requested ratio, but not the desktop ratio. I think using the requested ratio might make more sense.
So either remove this ratio and add back the other one, or use `float ratio = (float) _lastRequestedWidth / (float) _lastRequestedHeight;`

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -352,37 +348,54 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
...
+	// we override the window dimensions with the "real" request.
+	if ((requestedWidth < _lastRequestedWidth || requestedHeight < _lastRequestedHeight) && ConfMan.getActiveDomain()) {
+		requestedWidth  = _lastRequestedWidth  * _graphicsScale;
+		requestedHeight = _lastRequestedHeight * _graphicsScale;
+	}
+
+	// Set allowed dimensions
 	uint maxAllowedWidth   = desktopRes.width();
 	uint maxAllowedHeight  = desktopRes.height();
-	float ratio = (float)requestedWidth / requestedHeight;
```
criezy, line 385: See comment above. I think we probably want to use `float ratio = (float) _lastRequestedWidth / (float) _lastRequestedHeight;` here rather than the ratio computed above from the desktop resolution.

criezy (commented):
This looks almost good now. I think there are still a couple of things that could be improved and added comments in the corresponding places.

backends/graphics/openglsdl/openglsdl-graphics.cpp
criezy: Ha, I think I understand now why it was using `requestedWidth / requestedHeight` and not `_lastRequestedWidth / _lastRequestedHeight`. With the later you could end up with a size that is still bigger than the allowed size if decreasing the height causes the width to increase due to the requested ratio being wider than the computed one.

So reinstating the line above seems actually the better option.

backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -352,37 +348,54 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
...
 		// Set the basic window size based on the desktop resolution
 		// since we have no values stored, e.g. on first launch.
-		requestedWidth  = desktopRes.width()  * 0.3f;
-		requestedHeight = desktopRes.height() * 0.4f;
-
-		// Apply scaler
-		requestedWidth  *= _graphicsScale;
-		requestedHeight *= _graphicsScale;
+		requestedWidth  = MAX<uint>(desktopRes.width() / 2, MIN<uint>(640, desktopRes.width()));
+		requestedHeight = MIN<uint>(requestedWidth * 3 / 4, desktopRes.height());
```
criezy, line 371: Oh, one last comment. I have now noticed that there is some code below already to reduce the size if it is bigger than the desktop size (with the `maxAllowedWidth` and `maxAllowedHeight`). So we don't need to do it here and the code could be simplified to
```c++
requestedWidth  = MAX<uint>(desktopRes.width() / 2, 640);
requestedHeight = requestedWidth * 3 / 4;
```
And the same simplification can be done in the `#else` block below.

lotharsm: Thanks, good point. Initially, the code with `maxAllowedWidth` and `maxAllowedHeight` was meant as a safeguard, but it turned out to be _really_ helpful, especially when handling maximized windows. I'll adopt your changes here.

backends/graphics/openglsdl/openglsdl-graphics.cpp
lotharsm: Very nice, I was confused already why the scaling randomly didn't work properly for the 320x200 games. Turns out that the missing `_graphicsScale` in the comparison was indeed the culprit here.

Fixed by using `_graphicsScale` at both instances now.

backends/graphics/openglsdl/openglsdl-graphics.cpp
lotharsm: Thanks, I moved the aspect ratio calculation back to it's old place.

backends/graphics/openglsdl/openglsdl-graphics.cpp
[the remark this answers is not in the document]
lotharsm: As you proposed below, I reinstated this line.

lotharsm:
Thanks again, I changed the code according to your comments - I hope I got everything right :-)

criezy:
It looks all good to me now. Good job!

bluegr:
I have a question here: with the new code, the minimum  width is set to 640 pixels... can you please elaborate on this?

criezy:
> I have a question here: with the new code, the minimum width is set to 640 pixels... can you please elaborate on this?

You mean the minimum width for the *default* size.
That was already the case before due to both the Surface SDL and OpenGL graphics manager using a scaling of x2 by default for 320x200 games (and also for the gui with a `system.initSize(320, 200)` when starting ScummVM). The main change compared to previous releases is that we are now using a bigger window by default when using a big screen (bigger than 1280 pixels wide).

But you can still use a smaller window, either by resizing it manually, setting the `last_window_width` and `last_window_height` in the config file manually, or using the hotkeys to decrease the scaler (ctrl-alt-minus by default, which I actually fixed yesterday https://bugs.scummvm.org/ticket/12187).

lotharsm:
@criezy: Thanks a lot for your feedback, your patience and your mentoring :-)

@bluegr: Basically, this is the default right now as well, since the initial size is 320x200 which gets then upscaled to 640x400 (since we apply a 2x scaler).

bluegr:
@criezy @lotharsm: thanks for clearing this out :)

This looks very good to me, nice work overall!

bluegr (approved):

criezy:
I will merge this manually to squash some commits. I don't think we need to keep all the intermediate commits for the default window size. 😃

lotharsm:
Thank you - I think we can even squash everything into one single commit - the changes are not _that_ extensive.

criezy:
I am doing something slightly more advanced to keep some of the changes separate with a logical grouping, for example keep the `_windowIsMaximized` fix separate from the changes to improve the default size.

criezy:
This has now been merged.

bluegr:
Thanks! Closing, as this has been merged manually

PR #3267, opened 2021-08-12
SUPERNOVA: Text To Speech (TTS) for the game

taylorzhancher:
Tentatively adding TTS to the game
This includes
- Speech for opening intro story
- Speech for dialogue trees
- GUI option for TTS


engines/supernova/game-manager.cpp
```diff
@@ -553,15 +555,22 @@ bool GameManager::isNullObject(Object *obj) {
...
 	else {
-		for (int r = _rowsStart[number]; r < _rowsStart[number] + _rows[number]; ++r)
+		for (int r = _rowsStart[number]; r < _rowsStart[number] + _rows[number]; ++r) {
 			_vm->renderText(_texts[r], 1, 142 + r * 10, brightness ? kColorGreen : kColorDarkGreen);
+			if (!string.empty())
+				string += " ";
+			string += _vm->getGameString(_texts[r]);
+		}
 	}
+	ttsMan->say(string, Common::TextToSpeechManager::QUEUE_NO_REPEAT);
```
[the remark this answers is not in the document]
criezy: To test this, here is a savegame for the first game (you will need to unzip it). After loading this, just talk to the doorman.
[msn_save.006.zip](https://github.com/scummvm/scummvm/files/6978973/msn_save.006.zip)

engines/supernova/detection.cpp
```diff
@@ -82,7 +93,7 @@ static const ADGameDescription gameDescriptions[] = {
 		Common::IT_ITA,
 		Common::kPlatformDOS,
 		ADGF_NO_FLAGS,
-		GUIO2(GAMEOPTION_IMPROVED, GUIO_NOMIDI)
+		GUIO3(GAMEOPTION_TTS, GAMEOPTION_IMPROVED, GUIO_NOMIDI)
 	},
 	// Mission Supernova 2
```
criezy, line 98: You forgot to add it for the msn2 detection entries as well.

engines/supernova/screen.cpp
```diff
@@ -286,12 +287,16 @@ void Screen::renderMessage(int stringId, MessagePosition position,
...
 	renderMessage(text, position);
 }
 
 void Screen::renderMessage(const Common::String &text, MessagePosition position) {
 	if (!text.empty())
 		renderMessage(text.c_str(), position);
+	Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+	if (ttsMan != nullptr && ConfMan.getBool("tts_enabled"))
+		ttsMan->say(text,  Common::TextToSpeechManager::INTERRUPT_NO_REPEAT);
 }
```
criezy, line 300: Why did you add the TTS here and not in `Screen::renderMessage(const char *text, MessagePosition position, int positionX, int positionY)`?

That other render message is called from here, but it is also sometimes called directly. And when called directly it will be missing the TTS. So it would be better to move the TTS to that other one.

I am also wondering if queuing might not be better than interrupt here.

criezy: Here is a savegame for Mission Supernova 2 that will show the issue. You will need to unzip it. After loading the savegame, go to the Entrance (top right) and then click through the dialog. You will see that you are missing the reaction of the audience, but if you move the TTS to the place I suggest it should work.
[ms2_save.014.zip](https://github.com/scummvm/scummvm/files/6979064/ms2_save.014.zip)

criezy (changes requested):
This is a good start.
In addition to the comments I added, here are a few other changes to do:
 - You are missing the call to `TextToSpeechManager::setLanguage()` to either English or German depending which version we are playing.
 - Similarly to Dreamweb, you need to specify the text encoding. When playing in English it does not matter, but when playing in German it does. Here the encoding is again kDos850 (for both German and English).

In addition it would also be good to add TTS for some other text: 
  - Moving cursor on objects in the inventory or a hotspot.
  - Automatic dialogs (when there is no choice). There is an example with the savegame I provided for Mission Supernova 2 when you are on the scene ("One moment, I have to think about what I am playing for you").

For the inventory and hotspot I will need to have a look as I don't remember how it is handled.
For the automatic dialogs, I think it is using `GameManager::say()`.

taylorzhancher:
@criezy There are 2 things I wish to share & ask

1. The action panel can have TTS to highlighted options by doing this in `void GameManager::processInput()` Which seems great (see line 529)
```
				_guiInventoryArrow[_mouseField - 768].setHighlight(true);
				_ttsMan->say(_guiInventoryArrow[_mouseField - 768].getText(), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
				break;
			case onInventory:
				_guiInventory[_mouseField - 512].setHighlight(true);
				_currentInputObject = _inventory.get(_mouseField - 512 + _inventoryScroll);
				_ttsMan->say(_guiInventory[_mouseField - 512].getText(), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
				break;
			case onCmdButton:
				_guiCommandButton[_mouseField - 256].setHighlight(true);
				_ttsMan->say(_guiCommandButton[_mouseField - 256].getText(), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
				break;
			case onObject:
				_currentInputObject = _currentRoom->getObject(_mouseField);
```
- On the last 2 lines of the code above, it is found that the objects in the scene are also able to have TTS by `case onObject`
- However, I don't know how to get the text from the last line? In the API documentation I didn't find something that looked like text


2. I tried to add a get language function but was in vain. This was my attempt. The error complained that it doesn't recognize ` _gameDescription`. 

- I hope to push the changes here soon
- In case the reason is obvious, I've attached the code below
from supernova.cpp
```
#include "engines/advancedDetector.h"
#include "common/text-to-speech.h"

@ -99,6 +101,8 @@ SupernovaEngine::SupernovaEngine(OSystem *syst)
		_MSPart = 0;

	_improved = ConfMan.getBool("improved");

	_ttsMan = g_system->getTextToSpeechManager();
}

SupernovaEngine::~SupernovaEngine() {
@ -110,6 +114,9 @@ SupernovaEngine::~SupernovaEngine() {
}

Common::Error SupernovaEngine::run() {
	 if (_ttsMan != nullptr)
	 	_ttsMan->setLanguage(Common::getLanguageCode(getLanguage()));

@ -830,5 +837,8 @@ void SupernovaEngine::stopSound() {
	_sound->stop();
}

 Common::Language SupernovaEngine::getLanguage() const {
  	return _gameDescription->language;
 }

}
```

from supernova.h
```
@ -137,6 +137,8 @@ public:
	void renderBox(const GuiElement &guiElement);
	void setColor63(byte value);
	void stopSound();
	Common::Language getLanguage() const;
	Common::TextToSpeechManager *_ttsMan;
};
```

criezy:
> @criezy There are 2 things I wish to share & ask
> 
> 1. The action panel can have TTS to highlighted options by doing this in `void GameManager::processInput()` Which seems great (see line 529)

This is indeed a good place to add TTS

>```c++
> 	case onInventoryArrowUp:
> 	case onInventoryArrowDown:
> 		// Fallthrough
> 		_guiInventoryArrow[_mouseField - 768].setHighlight(true);
> 		_ttsMan->say(_guiInventoryArrow[_mouseField - 768].getText(), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
> 		break;
>```

Are you sure those up and down arrows (the buttons to scroll the inventory) have a text that can be spoken?

>```c++
> 	case onInventory:
> 		_guiInventory[_mouseField - 512].setHighlight(true);
> 		_currentInputObject = _inventory.get(_mouseField - 512 + _inventoryScroll);
> 		_ttsMan->say(_guiInventory[_mouseField - 512].getText(), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
> 		break;
>```

Does this actually work?
I don't think the `_guiInventory` elements have the text set, or at least not always. I think you might need to use the `_currentInputObject` instead.

```c++
		if (!isNullObject(_currentInputObject))
			_ttsMan->say(_vm->getGameString(_currentInputObject->_name), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
```

Also the `Common::TextToSpeechManager::INTERRUPT_NO_REPEAT` action is the default, so you don't need to specify it.
```c++
		if (!isNullObject(_currentInputObject))
			_ttsMan->say(_vm->getGameString(_currentInputObject->_name), Common::kDos850);
```

>```c++
> 	case onCmdButton:
> 		_guiCommandButton[_mouseField - 256].setHighlight(true);
> 		_ttsMan->say(_guiCommandButton[_mouseField - 256].getText(), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
> 		break;
>```

Same remark as above about the `Common::TextToSpeechManager::INTERRUPT_NO_REPEAT` that can be removed.

>```c++
> 	case onObject:
> 		_currentInputObject = _currentRoom->getObject(_mouseField);
> ```
> 
>  * On the last 2 lines of the code above, it is found that the objects in the scene are also able to have TTS by `case onObject`
> * However, I don't know how to get the text from the last line? In the API documentation I didn't find something that looked like text

Same as for the inventory object case above:
```c++
		if (!isNullObject(_currentInputObject))
			_ttsMan->say(_vm->getGameString(_currentInputObject->_name), Common::kDos850);
```

> 1. I tried to add a get language function but was in vain. This was my attempt. The error complained that it doesn't recognize ` _gameDescription`.

There is indeed no `_gameDescription` member variable in the `SupernovaEngine`. You can get the language from `ConfMan` instead.
```c++
if (_ttsMan != nullptr)
	_ttsMan->setLanguage(ConfMan.get("language"));
```

engines/supernova/game-manager.cpp
```diff
@@ -360,6 +362,8 @@ void GameManager::processInput() {
 		onInventoryArrowDown
 	} mouseLocation;
 
+	_ttsMan = g_system->getTextToSpeechManager();
```
criezy, line 365: If you are going to make this a member variable for the `GameManager` class, don't set it every time  `processInput()` gets called. Instead it should be set in the `GameManager` constructor.

But I don't see the benefit of having it as a member variable here.

engines/supernova/game-manager.cpp
```diff
@@ -525,13 +529,16 @@ void GameManager::processInput() {
 			case onInventoryArrowDown:
 				// Fallthrough
 				_guiInventoryArrow[_mouseField - 768].setHighlight(true);
+				_ttsMan->say(_guiInventoryArrow[_mouseField - 768].getText(), Common::TextToSpeechManager::INTERRUPT_NO_REPEAT, Common::kDos850);
```
criezy, line 532: Don't forget to test that the `_ttsMan` is not a nullptr and that tts is enabled.

criezy: Also see my comments in the pull request. Quick summary as a reminder:
 * You don't need to specify the `Common::TextToSpeechManager::INTERRUPT_NO_REPEAT` since this is the default.
 * I am not sur the inventory arrow text can actually be spoken.
 * For the case below I think you need to get the text from the `_currentInputObject` and not from the `_guiInventory` as the later may not be set.

engines/supernova/game-manager.cpp
```diff
@@ -555,13 +562,24 @@ bool GameManager::isNullObject(Object *obj) {
 void GameManager::sentence(int number, bool brightness) {
 	if (number < 0)
 		return;
+
+	_ttsMan = g_system->getTextToSpeechManager();
```
criezy, line 566: Same remark as above. Either set the `_ttsMan` in the `GameManager` constructor, or change it to a local variable.

engines/supernova/screen.cpp
```diff
@@ -535,6 +532,10 @@ void Screen::renderMessage(const char *text, MessagePosition position, int posit
 	int y = 0;
 	byte textColor = 0;
 
+	_ttsMan = g_system->getTextToSpeechManager();
+	if (_ttsMan != nullptr && ConfMan.getBool("tts_enabled"))
+		_ttsMan->say(text,  Common::TextToSpeechManager::QUEUE_NO_REPEAT, Common::kDos850);
+
```
criezy, line 538: Either set `_ttsMan` in the `Screen` constructor, or change it to a local variable.
But again I don't see the benefit of having this as a class member variable, so I would suggest making it a local variable.

engines/supernova/supernova.h
```diff
@@ -137,6 +137,8 @@ class SupernovaEngine : public Engine {
 	void renderBox(const GuiElement &guiElement);
 	void setColor63(byte value);
 	void stopSound();
+	// Common::Language getLanguage() const;
+	Common::TextToSpeechManager *_ttsMan;
```
criezy, line 141: I don't see the benefit of having it as a class member variable since it is only used once in one place. I would suggest to use a local variable in `SupernovaEngine::run()` instead.

criezy (changes requested):

taylorzhancher:
@criezy I've made changes to comply with what you said
Would like to add that `case onInventory:` , although my previous code doesn't actually produce tts, but `case onInventory:` refers to the objects in the inventory, which is unique to `case onObject`, and I've just added TTS to the former with `_vm->getGameString(_currentInputObject->_name)`

engines/sci/tts.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "sci/graphics/controls16.h"
+
+namespace Sci {
+
+void TTS_SAY(const char *text) {
+    Common::TextToSpeechManager *ttsMan = g_system->getTextToSpeechManager();
+    ttsMan->say(text);
+}
+
+}
```
criezy, line 37: I think you added that file by mistake here. Please remove it.

engines/sci/tts.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "sci/tts.cpp"
+#include "common/text-to-speech.h"
+#include "common/config-manager.h"
+#include "common/system.h"
+
+
+namespace Sci {
+    void TTS_SAY(const char *text);
+}
```
criezy, line 31: And this one as well.

engines/supernova/game-manager.cpp
```diff
@@ -529,12 +533,18 @@ void GameManager::processInput() {
 			case onInventory:
 				_guiInventory[_mouseField - 512].setHighlight(true);
 				_currentInputObject = _inventory.get(_mouseField - 512 + _inventoryScroll);
+				if (ttsMan != nullptr && ConfMan.getBool("tts_enabled"))
```
criezy, line 536:
```suggestion
				if (ttsMan != nullptr && ConfMan.getBool("tts_enabled") && !isNullObject(_currentInputObject))
```

engines/supernova/game-manager.cpp
```diff
@@ -529,12 +533,18 @@ void GameManager::processInput() {
...
+					ttsMan->say(_vm->getGameString(_currentInputObject->_name), Common::kDos850);
 				break;
 			case onCmdButton:
 				_guiCommandButton[_mouseField - 256].setHighlight(true);
+				if (ttsMan != nullptr && ConfMan.getBool("tts_enabled"))
+					ttsMan->say(_guiCommandButton[_mouseField - 256].getText(), Common::kDos850);
 				break;
 			case onObject:
 				_currentInputObject = _currentRoom->getObject(_mouseField);
+				if (!isNullObject(_currentInputObject))
```
criezy, line 546:
```suggestion
				if (ttsMan != nullptr && ConfMan.getBool("tts_enabled") && !isNullObject(_currentInputObject))
```

criezy (changes requested):

criezy:
There are a few more changes needed, and you need to remove the changes in the SCI engine.

> Would like to add that `case onInventory:` , although my previous code doesn't actually produce tts, but `case onInventory:` refers to the objects in the inventory, which is unique to `case onObject`, and I've just added TTS to the former with `_vm->getGameString(_currentInputObject->_name)`

I don't understand that sentence. Is that a question?

taylorzhancher:
> There are a few more changes needed, and you need to remove the changes in the SCI engine.
> 
> > Would like to add that `case onInventory:` , although my previous code doesn't actually produce tts, but `case onInventory:` refers to the objects in the inventory, which is unique to `case onObject`, and I've just added TTS to the former with `_vm->getGameString(_currentInputObject->_name)`
> 
> I don't understand that sentence. Is that a question?

No, it's not a question. Sorry for the many mistakes in the previous commit. Hope they're all good now.

criezy (approved):
Thank you. This looks good now.
The history will need to be cleaned, but we can do that by squashing all the commits when merging this pull request.

Since we are in feature freeze currently to prepare the next release, I will wait before merging this.

criezy:
Now that trunk is unfrozen, I will squash and merge this.

PR #3268, opened 2021-08-12
BASE: Convert version strings to const objects. 

carlo-bramini:
The following variables:

```
gScummVMVersion
gScummVMBuildDate
gScummVMVersionDate
gScummVMFullVersion
gScummVMFeatures
```

are declared as pointers to strings but there is no need to do so.
They can be simply declared as const strings, since they will be never modified when SCUMMVM runs.

### DIFF
--- a/base/version.cpp
+++ b/base/version.cpp
@@ -55,14 +55,14 @@
  * I don't know VC enough to be sure). And of course it must be robust enough
  * to properly work in exports (i.e. release tar balls etc.).
  */
-const char *gScummVMVersion = SCUMMVM_VERSION SCUMMVM_REVISION;
+const char gScummVMVersion[] = SCUMMVM_VERSION SCUMMVM_REVISION;
 #if defined(__amigaos4__) || defined(__MORPHOS__)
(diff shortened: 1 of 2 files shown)


carlo-bramini:
> I fail to see the benefit of this change... why is the proposed code better?

The difference is that in the first case you have a string and a pointer, while in the second case you have just the string.
For example, in this code:
`const char *my_string = "foo";`
the `my_string` identifier is a pointer that points to a location that stores the string "foo".
Instead:
`const char my_string[] = "foo";`
the `my_string` identifier is directly the string "foo".
In the first case, keep in mind that the strings are not const objects actually.
If you wanted to do so, the expression needs to be changed to:
`const char * const my_string = "foo";`
but in my opinion there is no need to keep also a pointer for the purpose that those strings have.
Of course, this is a minor fix, I just noticed this for pure case since I have seen that the strings were not into the `.rodata` section.

sev-:
This is kind of an unusual change, but for the aesthetical purity, it is OK... However, the commit log messages are not in conformance with our commit guidelines (https://wiki.scummvm.org/index.php?title=Commit_Guidelines). Could you please add the prefix "BASE: " in front of the commits? And also, please squash the commits into one, since one of them is not compilable.

sev-:
Thanks, merging

PR #3269, opened 2021-08-13
DIRECTOR: Add name for c64pack

einstein95:
Game name taken from [MobyGames entry](https://www.mobygames.com/game/windows/activisions-commodore-64-15-pack) as well as the ["About" window title](https://cdn.discordapp.com/attachments/870528165255188530/875558971136671844/image.png).


sev-:
Thanks!

PR #3272, opened 2021-08-14
SDL: Fix HiDPI scaling on Windows (and others?)

SupSuper:
Previously SdlWindow::getDpiScalingFactor only returned 100% or 200% scale. This is pretty limiting, as OSes have more fine-grained scales, for example on Windows, the "DPI" (or text scale) is set in increments of 25%: 100% (96dpi), 125% (120dpi), 150%, 175%, etc.

This changes SDL HiDPI to return the exact scaling factor, bringing it line with #3264 for macOS/iOS. I've only tested it on Windows, so I'd appreciate testing on other platforms, and seeing if there's any potential side-effects.

### DIFF
--- a/backends/platform/sdl/sdl-window.cpp
+++ b/backends/platform/sdl/sdl-window.cpp
@@ -269,10 +269,7 @@ float SdlWindow::getDpiScalingFactor() const {
 	getDisplayDpi(&dpi, &defaultDpi);
 	debug(4, "dpi: %g default: %g", dpi, defaultDpi);
 	float ratio = dpi / defaultDpi;
-	if (ratio >= 1.5f)
-		return 2.f;
-	else
-		return 1.f;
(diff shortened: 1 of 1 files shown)


criezy:
That looks fine to me.
This was discussed on Discord last night and all the Windows developers present agreed it was a good change on Windows.

This change does not impact macOS as `getDpiScalingFactor` is reimplemented in `SdlWindow_MacOSX`.
The change will however also impact other *nix system, but I think it should be fine for those as well. But maybe a safeguard to make sure the ratio is at least 1.f could be added.

bluegr:
I tried it on Windows and it looks much better indeed 😀
Well done!

criezy:
@lotharsm has indicated on Discord that he also tested this on Linux.
I think we can merge this now.

PR #3274, opened 2021-08-15
SCUMM: Add German MI1 Lemonhead lines

ghoost82:
This commit adds a patch that will readd the missing Lemonhead lines for the German CD version of Monkey Island


bluegr:
Nice work, thanks! Merging

PR #3275, opened 2021-08-16
GRIM: Add Monkey4 Steam/GOG version detection

Frohlix:
Hi

This PR makes it so the Steam and GOG versions of Escape from Monkey Island can be added and played.

The actual commit is simple enough, but there are a few points to note:

The files used for the Steam and GOG versions are mostly the same as for the CD versions, but of course there is no duplicate data situation. In particular there is only one `FullMonkeyMap.imt`, but as far as I can tell this works fine.

When trying to launch the game, the resources check will complain that `monkey4-patch.m4b`is missing. This is because the file is called `patch.m4b`. Simply renaming it fixes this.
Now, it's not ideal that one has to rename `patch.m4b`. But to be completely honest, I don't understand the codebase well enough to know whether there is a good solution for this, and if there is, how/where to implement it. I hope somebody more knowledgeable can help out here :)

I tested this rudimentarily by playing the game a bit.

Even if the PR doesn't make it, I hope it helps in supporting the Steam/GOG versions, because those are currently by far the easiest to get in a convenient and legal manner.

### DIFF
--- a/engines/grim/detection.cpp
+++ b/engines/grim/detection.cpp
@@ -461,6 +461,71 @@ static const GrimGameDescription gameDescriptions[] = {
 		GType_MONKEY4
 	},
 	{
+		// Escape from Monkey Island English Steam/GOG
+		{
+			"monkey4",
+			"",
+			AD_ENTRY1s("artAll.m4b", "8c7db9dab564854f2c4bab0571104780", 18515664),
+			Common::EN_ANY,
+			Common::kPlatformWindows,
+			ADGF_UNSTABLE,
+			GUI_OPTIONS_GRIME
+		},
+		GType_MONKEY4
+	},
+	{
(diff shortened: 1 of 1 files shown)


criezy:
> When trying to launch the game, the resources check will complain that `monkey4-patch.m4b`is missing. This is because the file is called `patch.m4b`. Simply renaming it fixes this.

Renaming a file from your copy of the game does not seem correct.
The `monkey4-patch.m4b` is a file that is distributed with ScummVM and not one that is supposed to come with your game. If you use your own compiled version of ScummVM without installing (or creating a .app bundle on macOS) you need to set your extra path in the ScummVM option to the `dist/engine-data/` inside your scummvm source code folder so that ScummVM can find those files.

Frohlix:
Oh yes, that worked flawlessly! Thank you @criezy for clearing that up :)

PR #3276, opened 2021-08-17
SCUMM: COMI: Fix string wrapping when wrap flag is active

AndywinXp:
This commit is aimed at fixing the fact that we have a wrap flag, but the wrapping per se (for non-subtitles blastText) is not implemented at all:
scummvm-comi-00020

There's a very lengthy comment in the commit explaining what I did; this implementation, while not being the same as the disasm (which changes x and y values word by word on the fly) works on the same principles:
* Split the string in multiple lines (as many as needed);
* Draw every substring, each time adding a getStringHeight() factor to y position;
* Starting from the bottom, compensate the y position (in other words, correct the string vertical placement as always), maintaining a getStringHeight() distance between each string.

This does in fact produce a result which is 1:1 pixel perfect (even for the x position, I don't know why) with respect to the original interpreter:

image

Actually the x position also seems to be pixel perfect everywhere now, why is that (not that I'm complaining)? 🤷‍♂️


AndywinXp:
@athrxx could you please give me your opinion on this PR when you've got time?
Code Analysis doesn't like strlen and strncpy but I've seen those being used in the codebase several times.

athrxx:
Fixing the wrapped texts for non-Smush texts is certainly  a good thing to do. Let's do it for all of SCUMM7/8, as it should be (we do currently wrap the dialog lines, but not always in a pixel perfect way).

- let's make enqueueText() virtual, make a new ScummEngine_v7::enqueueText() and
change ScummEngine_v6::enqueueText() to what it was.
- ScummEngine_v7::processSubtitleQueue()  will then also have to pass the wrapping parameter...
- Please do the measuring/adjusting stuff in the right order. You now have the vertical fixing hack that I made a couple days ago at the top. And then you still use that clipHeight value again at the bottom. You can remove the hack and just do everything like in SmushFont::drawStringWrap().
- Get rid of bt_secondary. If we also want to fix the other texts we need more lines. Just make a loop and enqueue all strings as needed. The arraysize for the blast text queue is currently at 50., that should be totally sufficient.

AndywinXp:
Thanks a lot! I will work on this on monday

athrxx:
You can use `Common::strlcpy()` and `Common::strnlen()` if you want. I don't know whether the static analyzer will like those better. Also, it doesn't really matter what the analyzer thinks... :-)

AndywinXp:
I managed to do some clean-up while on lunch break. Which means, everything on your list except copying SmushFont logic over. I will get back on this as soon as I can.

AndywinXp:
Ok, at the moment I don't have enough time to move that chunk of code from Smush (or, to put it more frankly, extensively test three games for regressions) so I have to drop that idea, but I managed to get pixel-perfect subtitle wrapping by using `getStringHeight()` (from SMUSH) instead of `getFontHeight()` in `CHARSET_1()`.

This particular change has been verified in the disasms for The Dig and COMI.

Please let me know if this PR at the current state is still mergeable after the last three commits, or it needs more work.

orgads:
cc @BLooperZ

athrxx:
In its current form the PR breaks English DIG (the '^' is actually used as a printable character there for '...'). The ^-commands are only present in COMI and the newer interpreter version of DIG...

AndywinXp:
I currently stopped working on anything non-audio engine related and that is until I'm finished with DiMUSE, so it might take a while for me to get back on this PR.

If anyone in the meantime manages to get to a satisfactory result without me, please feel free to close this PR of course and open a new one :)

AndywinXp:
> In its current form the PR breaks English DIG (the '^' is actually used as a printable character there for '...'). The ^-commands are only present in COMI and the newer interpreter version of DIG...
> 
> Cherrypicking the first commit (which fixes the issue) might work out fine. Should definitely get some testing, though.

I'm back on this! Could you please provide me with an example of DIG using ^ as '...', so I can find the issue and attempt to solve it properly with the disasm at hand?

athrxx:
At least one situation where the '^' / '...' issue can be tested is the situation where Brink is stuck with his hand in the wall if you talk to him.

I have to admit that I am a bit out of the loop with the text stuff. I had refactored and rewritten lots of the SCUMM7/8 text code which included your commits, also fixed the broken English DIG text formatting and even improved the Hebrew...

https://github.com/athrxx/scummvm/tree/scumm-7-8-text

The thing left to do was including the FT texts. Which was the point where I wasn't sure any more if it is such a good idea to move around such amounts of code. And found other projects... :-)

Maybe just polish up your first commit for the actual bug fix and we'll leave it at that...

AndywinXp:
Honestly? After looking at all your commits in your `scumm-7-8-text` branch It seems a real shame for your great work to be thrown away, and it makes me put this PR in another perspective: it's half-assed and incomplete! :-P (Or at least it only targets COMI correctly)

I can either limit the escape characters code to COMI, so to fix DIG, and leave it at that, or I'd be willing to give up on this PR in favour of your more complete and general version.

If you decide that none of those option is valid, I'd rather close this PR for the time being (since it can be seen as quite hackish, frankly) and see if it's possible to refactor the codebase in a distant future. ;-)

athrxx:
You're right, of course. I shouldn't let this bitrot. I believe to have basically pixel-perfect text for DIG and COMI, it would probably just take a couple of sessions to include FT, too. I'll see that I get back to it....

AndywinXp:
This is great news! :-D Take your time of course, but it's true, you've done so much for this subsystem with great results that it would really be a shame to abandon it

PR #2683, opened 2020-12-13
KYRA: support hebrew fan translation

BLooperZ:
adds support for hebrew fan translation
changes includes:
* Add hebrew fan translation detection entry
* Add hebrew strings to KYRA.DAT
* Bump KYRA.DAT version
* Add converter utility for escape sequences[1]
* Write hebrew text lines right to left
* Align hebrew text to the right
* Typewriter effect in intro Right-to-left in hebrew

[1] maybe there is more suitable place to put it in?

thanks @phcoder and @athrxx for their support

PR #3277, opened 2021-08-17
KYRA: support hebrew fan translation for Legend of Kyrandia 2: The Hand of Fate

BLooperZ:
Following #2683 
Here is Kyrandia 2 :)

this PR adds support for hebrew fan translation
changes includes:
Add kyra2 hebrew fan translation detection entry
Add kyra2 hebrew strings to KYRA.DAT (+Bump KYRA.DAT version)
Convert strings to be written Right-to-Left in main menu + intro sequence + spellbook


engines/kyra/detection_tables.h
```diff
@@ -829,6 +829,20 @@ const KYRAGameDescription adGameDescs[] = {
...
+			"CD",
+			AD_ENTRY2s("FATE.PAK", "28cbad1c5bf06b2d3825ae57d760d032", -1,
+					   "8FAT.FNT", "12424362a537e1335b10323c4013bb1d", -1),
+			Common::HE_ISR,
+			Common::kPlatformDOS,
+			ADGF_DROPLANGUAGE | ADGF_CD,
+			GUIO5(GUIO_MIDIADLIB, GUIO_MIDIMT32, GUIO_MIDIGM, GUIO_MIDIPCSPK, GUIO_RENDERVGA)
+		},
+		KYRA2_CD_FLAGS
+	},
```
athrxx, line 844: Does it work properly for you like that? Normally you would use KYRA2_CD_FAN_FLAGS(Common::HE_ISR, Common::EN_ANY) to tell the engine which of the three included languages gets replaced.

athrxx:
Nice :-) 
It looks pretty safe to me (no changes to the non-Hebrew parts).
Please have a look at the detection entry, the flags might need to be changed like I said in the comment.

engines/kyra/detection_tables.h
BLooperZ: It does (after investigating why, seems like _lang is initialized to 0). 
Still updated the flags as recommended, though.
Thanks you.

athrxx:
Thanks. I'll merge it right away so that it gets into the upcoming release...

engines/kyra/gui/gui_v1.cpp
```diff
@@ -599,26 +599,34 @@ void MainMenu::printString(const char *format, int x, int y, int col1, int col2,
 	Common::String string = Common::String::vformat(format, vaList);
 	va_end(vaList);
 
+	Common::String revBuffer;
+	const char *cstr = string.c_str();
+	if (_vm->gameFlags().lang == Common::HE_ISR) {
+		for (int i = string.size() - 1; i >= 0; --i)
+			revBuffer += string[i];
```
orgads, line 606: Why don't use you `Common::convertBiDiString`?

BLooperZ: I have tried, but IIRC it had issues with brackets (like in "[ Autosave ]"), so I just opted for reversing the texts, knowing the result will always be the exact same text spelled backwards is very predictable.

PR #3278, opened 2021-08-17
GUI: Support multiple alt buttons in MessageDialog

orgads:
This is an infrastructure for introducing autosave prompts. See #3261.


sev-:
Thanks!

PR #3281, opened 2021-08-18
COMMON: Fix c++ header usage.

fedor4ever:


bluegr:
Thanks! Other files follow this paradigm, so this was an exception. Merging

PR #3236, opened 2021-08-03
SYMBIAN: Implement openUrl() method.

fedor4ever:

PR #3286, opened 2021-08-18
SYMBIAN: Allow usage by default OSystem_SDL_Symbian::openUrl() 

fedor4ever:
Fix printf() usage for logging

### DIFF
--- a/backends/platform/symbian/src/SymbianOS.cpp
+++ b/backends/platform/symbian/src/SymbianOS.cpp
@@ -173,9 +173,7 @@ Common::String OSystem_SDL_Symbian::getDefaultConfigFileName() {
 bool OSystem_SDL_Symbian::hasFeature(Feature f) {
 	if (f == kFeatureFullscreenMode)
 		return false;
-#if defined(USE_CLOUD)
 	if (f == kFeatureOpenUrl) return true;
-#endif
 
 	return OSystem_SDL::hasFeature(f);
 }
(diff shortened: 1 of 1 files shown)


bluegr:
So why don't we want to use the USE_CLOUD define here?

fedor4ever:
I have no time to port  properly curl. Also Symbian lacks tls 1.2+, limited devices support tls 1.1. So I planned add local sync only for 2.3.0.
Fiodar Stryzhniou

исходное сбщ
Тема: Re: [scummvm/scummvm] SYMBIAN: Allow usage by default OSystem_SDL_Symbian::openUrl()  (#3286)
От: Filippos Karapetis ***@***.***>
Дата: 18.08.2021 21.55

So why don't we want to use the USE_CLOUD define here?

fedor4ever:
In  #3236 @criezy says "It is not tied to the cloud storage access and is used for various features...". So I decide allow usage for that.

bluegr:
Yes, that is correct, it's not tied to the cloud storage.

Does this work in all Symbian-based phones? Because you removed the #ifdef __S60_3X__ check

fedor4ever:
Maybe. Code itself compatible from Symbian 6.1 upwards but devices too numerous.

bluegr:
Right. Thanks for your work and feedback! Merging

fedor4ever:
@bluegr , @criezy , what game has local help files in html?

sev-:
Many, but so far, it was implemented in Pink Panther. Why?

fedor4ever:
@sev- , because test app and real life may differ. I don't know what arg comes to browser. They should come in right form. 

For opening local files exist more robust way.

sev-:
Then I do not understand why you did not run `git grep openUrl`

fedor4ever:
@sev- , I hear it for the first time.

PR #3287, opened 2021-08-18
XEEN fixes and further Russian translation

Ardash:
XEEN:
Click on 'Auto Notes' shows 'Items' button animation - Fixed
'Up'/'Down' pressed buttons in create_char dialog shows wrong icon - Fixed

XEEN: RU 
Items button not working because it uses 'd' alias, same as down button ('d', 'down', kp2) - Fixed
Fixed PARTY_DIALOG_TEXT margins 
Translated map title area names 
Translated moster names
Translated spell names


dreammaster:
Looks mostly fine overall, apart from some housekeeping I'd like to request
1) The first two listed commits don't have proper descriptions after XEEN: prefix. Even if there's further description below, a summary commit line is still advised
2) Only have a single commit that regenerates xeen.ccs. Since it's a binary blob file I want to keep down the number of times it gets redundantly committed in. Do you intend there to be any more changes in the near future? If so, I request rebasing the commits to remove it from both. Or if you think this is going to be the final version, just rebase and remove it from whichever commit is older.

sev-:
@Ardash Please hurry up, so we could put these changes into the releaae.

Ardash:
@dreammaster @sev- fixed

>> Do you intend there to be any more changes in the near future?
Some people testing it now, but there is no more reports for RU at this moment.
There is some non-critical reports for BASE now, but I think I can fix it later. 
I'm working on DARKSIDE RU now, but it is still a very long way.

dreammaster:
That's fine. So the changes done so far would be fine to go in this release, and some minor changes would be included in a later patch for Darkside. If so, I return to my original point - could you rebase the pull request commits to remove the regeneration of xeen.ccs from the oldest commit. Thereafter, there should be only the one commit that has xeen.ccs in it. Also, at the same time, flesh out the commit messages. Thanks.

dreammaster:
Oh, wait. I see the commits have been squashed. My bad. Let me review it further immediately..

PR #3288, opened 2021-08-18
SCUMM: Add SegaCD MI1 Lemonhead lines

BenCastricum:
This adds the missing Lemonhead lines to English SegaCD version of Monkey Island as well using @eriktorbjorn's resource patch for this. Playtested it.


bluegr:
Thanks! Merging

PR #3289, opened 2021-08-18
BASE: Purge keyboard events before running Engine

sluicebox:
When pressing Enter to start a game in the GUI, the key-up event remains in the event queue when an engine starts. In the SCI game Phantasmagoria 1, the game scripts for the introduction respond to any input event by skipping the introduction. This means that you only see the introduction by starting the game from the ScummVM GUI with a mouse. The introduction skips if you start the game by pressing Enter.

This PR adds EventManager::purgeKeyboardEvents() and calls it before Engine::run() so that GUI keyboard events no longer leak into engines. There is already a purgeMouseEvents() so I modeled it after that.

Tested on Windows and Mac.

### DIFF
--- a/base/main.cpp
+++ b/base/main.cpp
@@ -303,6 +303,10 @@ static Common::Error runGame(const Plugin *plugin, const Plugin *enginePlugin, O
 	// Inform backend that the engine is about to be run
 	system.engineInit();
 
+	// Purge queued input events that may remain from the GUI (such as key-up)
+	system.getEventManager()->purgeKeyboardEvents();
+	system.getEventManager()->purgeMouseEvents();
+
 	// Run the engine
 	Common::Error result = engine->run();
 
--- a/backends/events/default/default-events.cpp
+++ b/backends/events/default/default-events.cpp
@@ -296,6 +296,27 @@ void DefaultEventManager::purgeMouseEvents() {
 	_eventQueue = filteredQueue;
 }
 
+void DefaultEventManager::purgeKeyboardEvents() {
+	_dispatcher.dispatch();
+
+	Common::Queue<Common::Event> filteredQueue;
+	while (!_eventQueue.empty()) {
+		Common::Event event = _eventQueue.pop();
+		switch (event.type) {
+		// Update keyboard state even when purging events to avoid desynchronisation with real keyboard state
+		case Common::EVENT_KEYDOWN:
+		case Common::EVENT_KEYUP:
+			_modifierState = event.kbd.flags;
+			break;
+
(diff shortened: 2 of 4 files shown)


bluegr:
Nice find! I confirm the issue in Phantasmagoria, the intro logo would be skipped sometimes, and I always thought it was my imagination. Turns out it was caused by starting the game with the Enter key and by left clicking.

Overall, nice work, this will avoid weird issues in all engines when starting some games. +1 from me

base/main.cpp
```diff
@@ -303,6 +303,9 @@ static Common::Error runGame(const Plugin *plugin, const Plugin *enginePlugin, O
 	// Inform backend that the engine is about to be run
 	system.engineInit();
 
+	// Purge all queued keyboard events that may remain from the GUI
+	system.getEventManager()->purgeKeyboardEvents();
```
mgerhardy, line 307: I wonder why only keyboard events are purged here? What about mouse and/or joystick? Wouldn't it make a lot more sense to wipe all those those input related events?

sluicebox: I started with keyboard because those are the known problem and I can't vouch for the effects of anything else.

Purging others event types sounds reasonable but may someone familiar with the other Common::EventTypes should do that. There's often an EVENT_SCREEN_CHANGED in the queue, is removing that a problem? I have no idea.

SupSuper: There's a `purgeMouseEvents()` which should be safe to add. Other inputs are probably safe too, but most engines probably don't have any special joystick/touch handling.

For clearing everything there's already `getEventDispatcher()->clearEvents()` but like @sluicebox pointed out, don't know which events we wanna keep.

sluicebox: I'm comfortable updating this to also call purgeMouseEvents().

I could also add purgeJoystickEvents() to handle EVENT_JOYAXIS_MOTION, EVENT_JOYBUTTON_DOWN, EVENT_JOYBUTTON_UP if we want, but that's something I wouldn't be able to test. I think that plus purgeMouseEvents() would cover all the traditional input events.

What do you think? My goal is just to get the very real key-up bug safely fixed in time for release.

bluegr: Keyboard and mouse events are the most straightforward events to handle here, so they should be safe to be cleared in order to fix the issue at hand. Research for the other event types can be done post release

sluicebox:
Thanks for the feedback, everyone. I've added a call to purgeMouseEvents() and as @bluegr said, this fix can be researched and possibly expanded after the release.

PR #3290, opened 2021-08-19
ENGINES: keep existExtractedCDAudioFiles in sync with DefaultAudioCDM…

mgerhardy:
…anager::play()

out of sync since: f5164c911758e84804ef7d4e18c72cef21185daf

Reported on discord by eientei: https://discord.com/channels/581224060529148060/581224061091446795/877779182363500604

### DIFF
--- a/backends/audiocd/default/default-audiocd.cpp
+++ b/backends/audiocd/default/default-audiocd.cpp
@@ -23,6 +23,7 @@
 #include "backends/audiocd/default/default-audiocd.h"
 #include "audio/audiostream.h"
 #include "common/config-manager.h"
+#include "common/file.h"
 #include "common/system.h"
 #include "common/util.h"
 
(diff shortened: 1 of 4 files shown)


criezy (approved):

bluegr (approved):

engines/engine.cpp
```diff
@@ -465,30 +465,42 @@ void GUIErrorMessageFormat(Common::U32String fmt, ...) {
...
+	};
+
+	for (int i = 0; i < ARRAYSIZE(trackName); ++i) {
+		for (const char **ext = extensions; *ext; ++ext) {
+			const Common::String &filename = Common::String::format("%s.%s", trackName[i], *ext);
+			if (Common::File::exists(filename)) {
+				return true;
+			}
+		}
+	}
```
bluegr, line 498: It would be preferable to use the same code in both places, if possible

mgerhardy: Tried to do that in a follow up commit. The problem is that the supported file format extensions are still duplicated. We need a way to query those extensions. I've seen them in a few other locations spread over the code. Would be easier (as in less `#ifdef USE_THISANDTHAT` in the engine code) for the engine coders.

There is e.g. `SoundHE::tryLoadSoundOverride` - and it looks like it is missing `m4a` support - that might be on purpose though... not sure.

mgerhardy:
There was a discussion on discord about it which might be related (thanks trembyte for pointing me to this):

https://discord.com/channels/581224060529148060/581224061091446795/802369853968351262

backends/audiocd/default/default-audiocd.cpp
```diff
@@ -68,15 +107,13 @@ bool DefaultAudioCDManager::play(int track, int numLoops, int startFrame, int du
 		// Try to load the track from a compressed data file, and if found, use
 		// that. If not found, attempt to start regular Audio CD playback of
 		// the requested track.
-		Common::String trackName[4];
-		trackName[0] = Common::String::format("track%d", track);
-		trackName[1] = Common::String::format("track%02d", track);
-		trackName[2] = Common::String::format("track_%d", track);
-		trackName[3] = Common::String::format("track_%02d", track);
+		Common::Array<Common::String> trackNames;
```
criezy, line 110: Minor remark: you can use `Common::StringArray`.
In practice this doesn't change anything since this is a type for `Array<String>`.

bluegr:
Thanks! This is much cleaner now and easier to maintain. Merging

PR #3291, opened 2021-08-19
DOCS: added notes about the extracted audio cd files

mgerhardy:
Reported on discord by eientei: https://discord.com/channels/581224060529148060/581224061091446795/877767492146196541


cadihowley:
Looks good to me. There SHOULD be a preview of the documentation built with every PR, not sure why that’s still not happening…

mgerhardy:
It also looks like m4a is missing from the list. I've added it now. Hope it wasn't missing on purpose...

criezy:
> It also looks like m4a is missing from the list. I've added it now. Hope it wasn't missing on purpose...

My guess is that this was for historical reasons because support for it in `SeekableAudioStream` was added after the other three.

The change looks good, so I will merge it now.

PR #3292, opened 2021-08-19
SDL: Cleanup HiDPI macOS hacks

criezy:
This pull request take advantage of the changes I made in feac996b50 cleanup the hacks to handle HiDPI scaling on macOS.

The SDL library handles HiDPI differently depending on the system. On some systems, such as macOS, the drawable area and the SDL window have a different size (the window is on low-dpi size) while on other systems such as Windows they have the same size. Because of that we sometimes need to scale sizes or coordinates between the two, and sometimes we don't.

This was handled in two different ways.
 - In some places we queried the SDL window size and the SDL drawable area to compute the scaling between the two. As far as I know currently the scaling is 1 for all the systems except macOS with Retina screen.
 - In other places we used preprocessor directives to apply a scaling only on macOS.

This commit change the code to handle it consistently everywhere, and also should be more future proof should SDL change the way it handles HiDPI in the future (as we now query the size from SDL itself to find out if the scaling is needed instead of using the `__APPLE__` hack).

I have verified that this still works on macOS, but it might be worth checking that this does not introduce regressions on other systems (I don't think it does as `SdlWindow:: getSdlDpiScalingFactor()` should be 1, but I prefer to be on the safe side by opening a pull request for the change).

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -290,19 +290,15 @@ void OpenGLSdlGraphicsManager::notifyResize(const int width, const int height) {
 	// event is processed after recreating the window at the new resolution.
 	int currentWidth, currentHeight;
 	getWindowSizeFromSdl(&currentWidth, &currentHeight);
-	float scale = _window->getDpiScalingFactor();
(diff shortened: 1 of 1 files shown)


bluegr:
Looks good to me! +1

sev-:
Thanks! All looks good.

PR #3293, opened 2021-08-20
SYMBIAN: Fix command order in mmp_gen.py

fedor4ever:
Add ARM assembly to project file from Scumm engine
Avoid compiling ARM assembly for x86 emulator


sev-:
Thanks.

PR #3294, opened 2021-08-20
Fix various audio issues in Strangeland

criezy:
I looked at the original code of the ags_wave plugin for reference (https://github.com/Dualnames1/AGSWaves) but it handles the audio in quite a different way from this is implemented in ScummVM, so I am not really sure I understood everything properly.

These changes seem to fix all the audio issues I was having at the start of the game though.
So this also fix https://bugs.scummvm.org/ticket/12781

One thing that is still missing is the audio commentary.

### DIFF
--- a/engines/ags/plugins/ags_waves/ags_waves.cpp
+++ b/engines/ags/plugins/ags_waves/ags_waves.cpp
@@ -152,7 +152,8 @@ int64 AGSWaves::AGS_EngineOnEvent(int event, NumberPtr data) {
 		break;
 
 	case AGSE_ENTERROOM:
-		stopAllSounds();
+		// The original unloads sfx that are not playing and are not on repeat
+		// I don't think we need to do anything here.
 		break;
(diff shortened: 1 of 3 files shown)


bluegr:
Note that this contains the HiDPI commit

criezy:
> Note that this contains the HiDPI commit

Thanks! Fixed.

bluegr:
Overall, the changes look quite straightforward and reasonable

criezy:
Please do not merge yet.

I double checked the implementation in the original ags_wave plugin and commit ddbbfa2 is not correct. I will need to rework that part. The logic is actually more complicated than that when trying to play a sound that is already playing.

The other two changes should be fine.

criezy:
After checking the original source code for the ags_plugin, I am now confident with the changes in this pull request.
Some features are not implemented, but this is out of scope for this pull request.

Merging.

PR #3295, opened 2021-08-20
BACKENDS: OPENGLSDL: Remember maximized window state, fix off-screen positioning of window

lotharsm:
_Disclaimer: Not sure if this change is allowed right now or if it's affected by the current freeze. Would be nice to get it into the next release though._

Last time, I didn't know about the various SDL window flags, but this is my next try. This patch makes ScummVM remember if we left off with a maximized window and restores it properly.

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -550,7 +550,9 @@ bool OpenGLSdlGraphicsManager::setupMode(uint width, uint height) {
 
 	// WORKAROUND: Prevent (nearly) offscreen positioning of the ScummVM window by forcefully
 	// trigger a re-positioning event to center the window.
-	SDL_SetWindowPosition(_window->getSDLWindow(), SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
+	if (!_wantsFullScreen && !(SDL_GetWindowFlags(_window->getSDLWindow()) & SDL_WINDOW_MAXIMIZED)) {
+		SDL_SetWindowPosition(_window->getSDLWindow(), SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
+	}
 	return true;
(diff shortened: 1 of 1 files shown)


bluegr:
Looks good! +1 from me

lotharsm:
Thanks @bluegr!

55908159df600449fb11c9a9e32b1cfcb932d607 fixes a bug that's also present in the current master branch. When the Launcher runs in fullscreen mode (_not_ maximized, but true fullscreen instead) and you have a game set to override the graphic options and launch in windowed mode instead, you'll notice that (at least) games with 640x480 and higher might have issues where the ScummVM window is partially drawn outside the desktop viewport.

lotharsm:
Okay, there still seems to be an (unrelated?) issue when switching from fullscreen mode to windowed mode.

It seems that at least on Windows, SDL2 randomly and forcefully draws the window at coords 0 x 0, sacrificing even the top window bar until a resize event occurs.

It's possible we hit an SDL2 bug here, since even forcing `SDL_WINDOWPOS_CENTERED` in `sdl.cpp` doesn't help....

lotharsm:
...because I was looking at the wrong place.

2c61691d0498750d366980ed7ecd94fa7ab77ceb feels pretty ugly, but it works and fixes all known remaining issues with the OpenGL renderer.

lotharsm:
Since the creation of the next release branch is pretty soon and I wasn't able to reproduce the issues I had on Linux, I'm limiting the scope of the window positioning workaround to Win32.

lotharsm:
Any objections against merging this despite the freeze? I see this more as a bugfix and the new option is more or less a nice side effect ;)

lotharsm:
Mh, hold on - this is currently not really nice, since the re-positioning is also triggered if it's not strictly necessary, e.g. when running in plain windowed mode.

lotharsm:
If there are no objections, I'd like to merge this either today or early tomorrow, so it lands in the next release branch.

sev-:
Thank you, merging.

PR #3296, opened 2021-08-20
SCI: Fix GK2 patch dialog button behavior

tsoliman:
The button behavior was reversed.
"Download patch" did nothing and "ok" downloaded the patch.


bluegr:
Thanks! Merging

PR #3298, opened 2021-08-21
PRINCE:  Only the first phrase was displayed when examining an object

aviloria:
When examining an object, if the description has multiple phrases, only the first one was displayed on the screen (but speech audio continues).
Thant's because multiple phrases are encoded in the same string using '\0' separator, so strcpy/strncpy should be avoided.

### DIFF
--- a/engines/prince/script.cpp
+++ b/engines/prince/script.cpp
@@ -1025,6 +1025,10 @@ void Interpreter::O_GETMOBTEXT() {
 	int32 mob = readScriptFlagValue();
 	debugInterpreter("O_GETMOBTEXT mob %d", mob);
 	_currentString = _vm->_locationNr * 100 + mob + 60001;
+	// Use of memcpy() instead of strncpy() because the examination text can contain
+	// different phrases separated by '\0' characters followed by an ID.
+	// The examination text ends for a mob if the ID is 0xFF.
+	// Strings are properly extracted by the interpreter in that format.
 	memcpy((char *)_stringBuf, _vm->_mobList[mob]._examText.c_str(), 1023);
 	_string = _stringBuf;
 }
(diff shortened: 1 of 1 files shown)


engines/prince/script.cpp
```diff
@@ -1025,7 +1025,7 @@ void Interpreter::O_GETMOBTEXT() {
 	int32 mob = readScriptFlagValue();
 	debugInterpreter("O_GETMOBTEXT mob %d", mob);
 	_currentString = _vm->_locationNr * 100 + mob + 60001;
-	strncpy((char *)_stringBuf, _vm->_mobList[mob]._examText.c_str(), 1023);
+	memcpy((char *)_stringBuf, _vm->_mobList[mob]._examText.c_str(), 1023);
```
bluegr, line 1028: Please add a comment here, indicating the reason that `memcpy()` is used

aviloria: Done

engines/prince/script.cpp
```diff
@@ -1025,7 +1025,11 @@ void Interpreter::O_GETMOBTEXT() {
 	int32 mob = readScriptFlagValue();
 	debugInterpreter("O_GETMOBTEXT mob %d", mob);
 	_currentString = _vm->_locationNr * 100 + mob + 60001;
-	strncpy((char *)_stringBuf, _vm->_mobList[mob]._examText.c_str(), 1023);
+	// Use of memcpy() instead of strncpy() because the examination text can contain
```
bluegr, line 1028:
```suggestion
	// Use memcpy() instead of strncpy() because the examination text can contain
```

engines/prince/script.cpp
```diff
@@ -1025,7 +1025,11 @@ void Interpreter::O_GETMOBTEXT() {
...
+	// different phrases separated by '\0' characters followed by an ID.
```
bluegr, line 1029:
```suggestion
	// different phrases separated by '\0' characters, followed by an ID.
```

bluegr:
Thanks! Merging

PR #3299, opened 2021-08-21
SCUMM: MONKEY-VGA fix bug #346 and #10571, Object stopped with active cutscene

BenCastricum:
Both bugs also occur in original.

The crashes are triggered by "background scripts" resetting inventory objects in cutscene state. Details are in the commits.

### DIFF
--- a/engines/scumm/script_v5.cpp
+++ b/engines/scumm/script_v5.cpp
@@ -2088,6 +2088,21 @@ void ScummEngine_v5::o5_setCameraAt() {
 }
 
 void ScummEngine_v5::o5_setObjectName() {
+	// WORKAROUND bug #10571 (also occurs in original) Object stopped with active cutscene
+	// Script 68 contains the code for handling the mugs. The issue occurs when a mug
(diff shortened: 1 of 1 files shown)


eriktorbjorn:
I wonder if that also has any effect on https://bugs.scummvm.org/ticket/10571 ... but I don't remember if I ever tried to reproduce that one.

bluegr:
This is a very specific fix, so I believe it's safe to add

BenCastricum:
> I wonder if that also has any effect on https://bugs.scummvm.org/ticket/10571 ... but I don't remember if I ever tried to reproduce that one.

This patch will not fix that bug. Bug#10571 is caused by the melting cup containing grog somehow, most likely because it also contains a cutscene. But I'll have a closer look at that one later.

BenCastricum:
I added a fix for the other bug too. Hope that's ok. They are pretty similar but not in a way that a generic fix was possible.

bluegr:
Thanks! The fixes are quite targeted, so they are safe to add. Merging

PR #3304, opened 2021-08-23
GUI: Warn when enabling autosave on options

orgads:
...if non-autosave games are stored in autosave slot

### DIFF
--- a/gui/options.cpp
+++ b/gui/options.cpp
@@ -38,6 +38,7 @@
 #include "common/config-manager.h"
 #include "common/gui_options.h"
 #include "common/rendermode.h"
+#include "common/savefile.h"
 #include "common/system.h"
 #include "common/textconsole.h"
 #include "common/translation.h"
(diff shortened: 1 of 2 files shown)


bluegr:
Looks to be a very straightforward implementation. Nice work! +1 from me

bluegr:
Typo on the commit message: enaling -> enabling

gui/options.cpp
```diff
@@ -2476,6 +2477,83 @@ void GlobalOptionsDialog::addAccessibilityControls(GuiObject *boss, const Common
...
+
+bool GlobalOptionsDialog::updateAutosavePeriod(int newValue) {
+	const int oldAutosavePeriod = ConfMan.getInt("autosave_period");
+	if (oldAutosavePeriod != 0 || newValue <= 0)
+		return true;
+	typedef Common::Array<ExistingSave> ExistingSaveList;
+	ExistingSaveList saveList;
+	using Common::ConfigManager;
+	const ConfigManager::DomainMap &domains = ConfMan.getGameDomains();
+	const Common::String autoSaveName = Common::convertFromU32String(_("Autosave"));
```
criezy, line 2500: Is this variable actually used anywhere?

criezy (commented):
The only worry that I have is about performances if the user has added a lot of games in ScummVM and disk access is slow. But I don't know if that can be an issue in practice.

orgads:
> Typo on the commit message: enaling -> enabling

Fixed. Thanks for spotting.

gui/options.cpp
orgads: Good point. Leftovers from previous revisions.

orgads:
> The only worry that I have is about performances if the user has added a lot of games in ScummVM and disk access is slow. But I don't know if that can be an issue in practice.

Good point. Notice it is only executed when changing from zero to non-zero.

But it might be a good idea to limit the entry count. I suppose 10 entries are enough?

sev-:
I think this is fine. It is more important to not lose a savegame than wait a little bit. Merging.

PR #3308, opened 2021-08-24
GUI: Warn when saving a "young" game over an "older" one

orgads:


sev- (approved):
Very good addition, thank you.

bluegr:
Very nice indeed! Thanks, +1 from me too

PR #3310, opened 2021-08-26
fix rng seeding

Die4Ever:
Common::RandomSource fix rng seeding

Rng seeding on Windows had a small range of possible values, because it was set in the constructor based on milliseconds since process start. This was greatly reducing the number of possible seeds. @TehGelly noticed this issue and pointed out that the only seeds he ever gets when playing Shivers are between about 900 to 1050.

```
uint32 OSystem_NULL::getMillis(bool skipRecord) {
#ifdef POSIX
	timeval curTime;

	gettimeofday(&curTime, 0);

	return (uint32)(((curTime.tv_sec - _startTime.tv_sec) * 1000) +
			((curTime.tv_usec - _startTime.tv_usec) / 1000));
#elif defined(WIN32)
	return GetTickCount() - _startTime;
#else
	return 0;
#endif
}
```

### DIFF
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -37,7 +37,13 @@ RandomSource::RandomSource(const String &name) {
 #ifdef ENABLE_EVENTRECORDER
 	setSeed(g_eventRec.getRandomSeed(name));
 #else
-	setSeed(g_system->getMillis());
+	TimeDate time;
+	g_system->getTimeAndDate(time);
+	uint32 newSeed = time.tm_sec + time.tm_min * 60 + time.tm_hour * 3600;
+	newSeed += time.tm_mday * 86400 + time.tm_mon * 86400 * 31;
+	newSeed += time.tm_year * 86400 * 366;
+	newSeed = newSeed * 1000 + g_system->getMillis();
+	setSeed(newSeed);
 #endif
 }
 


Mataniko:
This is only for the initial random seed. The seed is changed every time getRandomNumber() is called. Sure, 150 possible initial seeds are not much, but since we don't require cyptographically secure generation it should be enough.

More so, most games request a RNG within a small range, and specifically SCI games have their own abstraction on top of RNG. 

Are you seeing an issue with Shivers due to RNG? that may be a script bug, as we've seen several of those with regards to RNG. I recommend opening a bug since this won't necessarily fix anything other than increase the range of potential initial seeds.

Die4Ever:
getRandomNumber does not get a "new" seed really, it just gets the next seed in the sequence, it will always follow the same sequence

so if you have a game where getRandomNumber is called every frame or something then it's fine because no player is perfect so they'll have getRandomNumber be called a different number of times in between events that the players care about

if you have a game like Earthbound (just an example) where the rng is called every time you bump into something then that is acceptable for casual players but manipulatable by speedrunners

if you have a game like Shivers which only calls rng at specific points, then you have very low entropy

this was not identified by reading code, it was identified by speedrunners noticing they're getting the same things pretty frequently

Mataniko:
You're chasing a red herring here. The initial seed might be limited but the randomness of the function is decent enough.

For example, generating 10,000 random numbers for each initial seed between 900 and 1050 creates 1500000 unique numbers, with no repeats.

It's more likely that the range for a random number in a specific scenario is too restrictive. 

Are you able to provide a specific example? If a particular puzzle in Shivers only has 8 different combinations, yes you will see things pretty often, but the distribution is relatively spread consistently.

criezy:
I am not sure this is a red herring. I think you should not be looking at the number of unique random numbers you get, but at the number of unique sequences. And this is solely controlled by the starting seed. So in this case you get at most about 150 possible sequences, and this might be reduced further by the resolution of `GetTickCount()` on Windows (between 10 and 16ms according to the Microsoft documentation).

In any case I see no harm in trying to increase the range and randomness of this initial seed.

criezy:
However the commit cannot be accepted as is as it does not respect our commit message format convention.
https://wiki.scummvm.org/index.php?title=Commit_Guidelines

Also the code does not seem to increase the initial seed much since you are adding a number between 0 and 141 (and with a non-uniform distribution, so some values are more likely than others). There might be a better way to do it.

Mataniko:
I've done additional tests with a more restrictive set of numbers to decrease randomness in both distribution and sequences and here's what i got:

For an initial seed between 900 and 1250 generating a 10 number sequence with a maximum with a random range between 0 and 10, we end up generating 351 unique sequences and no duplication.

Here is the distribution of unique numbers within each sequence:
image

92% of the sequences have more than 6 unique numbers while 8% have 5 unique numbers or less. Just under 35% of the sequences have 6 unique numbers or less.

I'm not arguing that increasing the range of numbers for the seed isn't good - but our distribution and uniqueness seem relatively good. This tells me that if there is an issue with something random happening more often than it should, it could be a bug in the engine/script and this change would not fix it.

sev-:
The point here is regarding the possible easiness of manipulation for the RNG seed which could be exploited by, say, speedrunners. And ScummVM is widely used by speedrunners.

sev-:
@Die4Ever please add a `COMMON: ` prefix to your commit log message and perhaps be a bit more descriptive there and then we merge it.

Die4Ever:
Yes the issue is repeatability. PRNG is stable and deterministic, so if you start the game with the game seed then you get all the same stuff again, I've made a quick video to demonstrate this https://youtu.be/QOYH6HhXkhg

I can fix my commit message soon, I'll also properly multiply the numbers because I did forget to do that (like tm_min*60)

TehGelly:
Additionally, with further testing on modern SSDs and caching, that range can be decreased from 350 possible seeds down to about 10 or 20. While 350 may not have been noticeable, 20 definitely has been.

sev-:
Thank you, merging

PR #3311, opened 2021-08-26
DEVTOOLS: COMPANION: Fix masking lower bits of flag

einstein95:
Properly gets the "new" Finder flags from the flags property, covers the flags introduced in Finder 5.4.

### DIFF
--- a/devtools/dumper-companion.py
+++ b/devtools/dumper-companion.py
@@ -130,7 +130,7 @@ def decode_macjapanese(text: ByteString) -> str:
 
 def file_to_macbin(f: machfs.File, name: ByteString) -> bytes:
     oldFlags = f.flags >> 8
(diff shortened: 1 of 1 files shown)


rvanlaar:
Please update your commit message. Why is 0xFF more correct than 15, 0xF?

rvanlaar:
Hi,

Can you make this into one commit? Then I'll commit it to scummvm.

rvanlaar:
> Please update your commit message. Why is 0xFF more correct than 15, 0xF?

Reply was: the flags are one full byte, so 0xFF it is.

rvanlaar:
Thank you for the PR. I've added the change here: https://github.com/scummvm/scummvm/commit/898dfcdae3570664e559482a190eada89fadf6a6

PR #3313, opened 2021-08-26
SDL: Set better default scaler for high-resolution games

eriktorbjorn:
Now that we have higher scale factors than 3, it makes sense to try and figure out a default scaler for high-resolution games that approximate the window size of a low resolution game.

That also means that we are not necessarily restricted to the normal scaler, since AdvMame has both 2x and 4x versions.

Since my own default scaler is Normal 3x, this change won't affect me much personally, though.

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -716,12 +716,20 @@ void SurfaceSdlGraphicsManager::initSize(uint w, uint h, const Graphics::PixelFo
 
 	if ((int)w != _videoMode.screenWidth || (int)h != _videoMode.screenHeight) {
 		const bool useDefault = defaultGraphicsModeConfig();
+		int scaleFactor = ConfMan.getInt("scale_factor");
+		int mode = _videoMode.scalerIndex;
 		if (useDefault && w > 320) {
-			// Only the normal scaler has a 1x mode
-			setScaler(ScalerMan.findScalerPluginIndex("normal"), 1);
-		} else {
-			setScaler(_videoMode.scalerIndex, ConfMan.getInt("scale_factor"));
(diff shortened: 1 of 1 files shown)


backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -716,12 +716,20 @@ void SurfaceSdlGraphicsManager::initSize(uint w, uint h, const Graphics::PixelFo
...
-			setScaler(_videoMode.scalerIndex, ConfMan.getInt("scale_factor"));
+			// The default scaler is assumed to be for low
+			// resolution games. For high resolution games, pick
+			// the largest scale factor that gives at most the
+			// same window width.
+			scaleFactor = MAX<uint>(1, (scaleFactor * 320) / w);
+
+			// Only the normal scaler has a 1x mode.
+			if (scaleFactor == 1)
+				mode = ScalerMan.findScalerPluginIndex("normal");
```
criezy, line 730: Maybe replace this special case for the factor 1 with a more robust check that the factor exists for the current scaler?

This is not tested, but I think it would look like this:
```c++
const Common::Array<uint> &factors = _scalerPlugins[mode]->get<ScalerPluginObject>().getFactors();
if (Common::find(factors.begin(), factors.end(), scaleFactor) == factors.end())
	mode = ScalerMan.findScalerPluginIndex("normal");
```

I think your code probably works with all the currently existing scaler plugins and supported game resolutions. So this is a suggestion to make is more future-proof rather than to fix an existing issue.

eriktorbjorn: Sounds reasonable, and seems to work. (I get the idea of your implementation, but I'm not fluent in that kind of C++ trickery so I'm copying your implementation of it.)

sev-:
Thank you!

PR #3317, opened 2021-08-28
BUILD: Add a target to download manual when not present

lephilousophe:
Don't use special characters in file name as make and shell don't have
same rules.

When we use parentheses, shell expects them to be escaped (or quoted) while make doesn't want to hear about quotes nor escaping (it doesn't see file).
Spaces are not ideal too but they can be handled as we don't do heavy processing with the variables.

Then, add the $(DIST_FILES_DOCS) as a prerequisite to all targets using it. This will download the manual when it's not present and curl is available.


sev-:
Thank you

PR #3318, opened 2021-08-28
STARK: Detect TLJ Steam version to fix missing bg resource

antoniou79:
This is an alternative fix for bug https://bugs.scummvm.org/ticket/12762

This fix will set a flag at detection time (when adding the game to ScummVM)
so that the Steam version of The Longest Journey will skip trying to load a background image resource (147.bmp) from the exe file (game.exe), since that specific build is missing this resource and this leads to an assertion fault.

Edit: It wont be necessary to re-add (redetect) the game, if it has already been added before this fix.

The previous fix was this commit and it was based only on version info of the exe file.
That one also works without having to re-add/re-detect the game, but perhaps just checking
version info is not the best approach.
https://github.com/scummvm/scummvm/commit/565a0559ed09660daa83cd529afa4e0efb6949c5

### DIFF
--- a/engines/stark/ui/dialogbox.cpp
+++ b/engines/stark/ui/dialogbox.cpp
@@ -228,8 +228,6 @@ Graphics::Surface *DialogBox::loadBackground() {
 	// GOG version of The Longest Journey is 1.0.0.142 "RC1" (Special Build: "Paper Sun")
 	// Steam's game.exe does not contain a valid resource for the background bitmap id 147
 	// so we skip trying to retrieve it based on build number.
-	// Important Note: For this fix to have any effect the game may have to be re-added to ScummVM 
-	//      so that it can be detected as the Steam version and set the flag.
 	if (_vm->getGameFlags() & GF_MISSING_EXE_RESOURCES) {
 		warning("Steam version does not contain the modal dialog background bitmap in 'game.exe'. Using fallback color for dialog background...");
 		return nullptr;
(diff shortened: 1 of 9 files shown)


sev-:
The fix looks fine, but I am confused by your note regarding the necessity of re-add the game. It is not needed as per the current AdvancedDetector implementation. And we run detection on every run and returning the relevant detection entry. Thus, now, with you adding additional flag, it is all fine and should work as is.

antoniou79:
> The fix looks fine, but I am confused by your note regarding the necessity of re-add the game. It is not needed as per the current AdvancedDetector implementation. And we run detection on every run and returning the relevant detection entry. Thus, now, with you adding that additional flag, it is all fine and should work as-is.

You are correct. I thought I tested it yesterday and the game I had added before the fix, didn't launch after the fix, so I assumed that it needed to be re-added (re-detected to set the flag). I re-added it and it definitely worked then. However, testing again today it seems I did something wrong yesterday, because this case now works without re-adding the game. Which is confirmed by your explanation that this is how the AdvancedDetector works, so this is good.

I'll correct the comment in the code.

engines/stark/ui/dialogbox.cpp
```diff
@@ -228,12 +228,7 @@ Graphics::Surface *DialogBox::loadBackground() {
 	// GOG version of The Longest Journey is 1.0.0.142 "RC1" (Special Build: "Paper Sun")
 	// Steam's game.exe does not contain a valid resource for the background bitmap id 147
 	// so we skip trying to retrieve it based on build number.
```
criezy, line 230: The `based on build number` part of the comment should probably be removed?

criezy (approved):
It looks good and cleaner than the previous fix.

engines/stark/ui/dialogbox.cpp
antoniou79: Indeed. I removed it now.

toby63:
I can confirm that it is working.

Thx for fixing the bug.

PR #3319, opened 2021-08-28
Manual manual download

criezy:
Fixes trying to copy the user manual pdf file when it is not present.
This does so by checking if the file is present before adding it to `DIST_FILES_DOCS`.

This is an alternative to pull request #3317 that instead automatically downloads the manual if not present (or at least try to do so). Since that pull request had already been merged, this pull request includes reverting the commits it contained.

I am not sure which approach is best, which is why I am proposing this as a pull request. With the changes here, if we want to bundle the user manual, we need to run manually `make manual` first. But in the case of a release, the file could be included in the release tarball.

One issue with the automatic download is that it requires an internet access. Without one trying to do `make bundle` or `make install` will fail.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -132,10 +132,12 @@ $(EXECUTABLE).dwp: $(EXECUTABLE)
 endif
 
 # Grab the ScummVM Manual from Read the Docs
-manual:
 ifdef USE_CURL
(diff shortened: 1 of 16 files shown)


sev-:
Well, I am now merging this and backporting. The previous automatic approach was problematic in at least two scenarios:

  - When the host has no internet connection, like Raspberry Pi
  - When the host does not have `curl`, like Amiga

PR #3320, opened 2021-08-28
CONFIGURE: Drop `u` flag for ar

orgads:
On replace mode (r), u tells ar to replace object files only if the source
file is newer than the one in the archive.

When binutils is configured with --enable-deterministic-archives, the
default mode for ar is D (deterministic), which means that all the files
are stored with zero for UID, GID and timestamp. This means that ar has no
way to tell if the archived file is older than the source file, so it just
archives all the input files and issues the following warning:
```
`u' modifier ignored since `D' is the default (see `U')
```

Fix by dropping this flag. The performance gain is negligible anyway.

### DIFF
--- a/configure
+++ b/configure
@@ -227,7 +227,7 @@ _windows_unicode=yes
 # Default commands
 _ranlib=ranlib
 _strip=strip
-_ar="ar cru"
+_ar="ar cr"
 _as="as"
 _dwp=dwp
 _windres=windres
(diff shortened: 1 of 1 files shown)


lotharsm:
+1 for this. I already dropped the `u` flag for the Win32/MinGW builds a while ago - good idea to do this globally.

lotharsm:
I wonder if there might be a reason why specific builds specifically ask for `cru`? Maybe it is better to just change the default value here and leave modifying the port-specific settings to the porters instead.

orgads:
My guess is copy/paste, either from other builds or other projects.

lotharsm:
Thank you for this patch, @orgads! Merging now - let's see if buildbot complains ;-)

PR #3323, opened 2021-08-29
DOCS: moves Mac game data page to "handling game files" page

cadihowley:
@criezy Thanks for expanding the Mac section, much appreciated! I've done some edits to align more closely with the style guide. 

What are your thoughts on combining it with the Handling game files page? It seems like it belongs there and with proper navigation it isn't tooo unwieldy...


criezy:
That seems fine.
The reason I created a separate file was for two reason:
1. Highlight the browser dumper companion method by having it on its own on the"handling game files" page with the other methods relegated to the more detailed page. The browser dumper companion should work in most cases and is easier to use than the other methods.
2. I was worried that having everything on the "handling game files" page would make it difficult to navigate, with too much content on the page. In the end because we used tabs, not all the content is visible and the page is not as big as I anticipated.

But if you think it works better with a single page, I have no objections.

One other thing I considered to make the dumper companions more prominent was to have the section on creating the iso image at the end of the chapter on the dumper companions. And for the browser version for example, before the step 1 ("Upload your iso here"), add a step 0 "Create an ISO image of your floppy disks or CDs" with a link to the relevant section. In the end I decided to keep the content in the logical order (first create the ISO, then use the dumper companion), but I am not completely convinced this was the right decision.

And maybe you have other ideas to make the browser dumper companion method more prominent so that it is more clear that this is the recommended method for the majority of cases.

criezy:
@cadihowley Any idea on how we can make it clear that the browser dumper-companion is the preferred method when it is applicable? I am worried that this is getting lost with your changes.

Other than that the change looks good, and we might want to merge it before tagging the release.

cadihowley:
@criezy Sorry for the delay on this - few things going on my end with work etc!

I think having all the information (including creating the ISO file) in order is a good idea, so I think you made the right call on that. Personally when I read docs I like being able to follow along without linking all over the place. Kind of following on from that, I suspect that having the browser method listed first means that most people will try that first by default. However I agree it might not be obvious enough. What about making a reference to it being the preferred method in the section title and/or the paragraph explaining the dumper?

The other thought I had is that we could still put the mac stuff on its own page, but make it the page that immediately follows the general "handling game files" page. I might try that now and see how it feels...

criezy (approved):
That looks good to me.

The added reference to `preferred method` in the section title as well as in the paragraph explaining the dumper seems good enough to me.

Regarding moving the content back to a separate page, I am more unsure. There is quite a lot of content, so having a dedicated page is maybe better. But on the other hand having two separate pages for the game data handling could maybe confuse a bit the users (although is is linked from the other page and they are listed next to each other in the side menu). So I will let you decide if that is better or not.

cadihowley:
I've added a statement which will hopefully direct users to the more general game data file page if they end up on the Mac page first. I reverted to having it on one page again, and after some comparison I do prefer it being on a separate page I think. It's definitely easier to navigate! So I will leave it as that for now, if we change our mind in the future I'd be happy to move it around again :)

PR #3324, opened 2021-08-29
SCUMM: Fix handling of Ctrl+Key in monkey1 and monkey2

orgads:
There are several Ctrl+Key combinations that are accepted by the engine.
Most famous one is Ctrl+W, which is Instant Win on monkey1 (monkey2 uses
Alt+W, and has a similar hack for passing it correctly).

This used to work with ScummVM 1.9.0/SDL1, but broke with SDL2, and was
later broken further even for SDL1 in f5ed14e93d85.

Fixes #10644.

### DIFF
--- a/engines/scumm/input.cpp
+++ b/engines/scumm/input.cpp
@@ -619,6 +619,10 @@ void ScummEngine::processKeyboard(Common::KeyState lastKeyHit) {
 		    lastKeyHit.keycode <= Common::KEYCODE_F9) {
 			_mouseAndKeyboardStat = lastKeyHit.keycode - Common::KEYCODE_F1 + 315;
 
+		} else if ((_game.id == GID_MONKEY || _game.id == GID_MONKEY2) &&
+				   (lastKeyHit.flags & Common::KBD_CTRL) &&
+				   (lastKeyHit.keycode >= Common::KEYCODE_a && lastKeyHit.keycode <= Common::KEYCODE_z)) {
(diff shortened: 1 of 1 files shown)


eriktorbjorn:
See also https://github.com/scummvm/scummvm/pull/3202 which is an earlier attempt at addressing the same issue. (The comments are probably more relevant than the actual commit.)

orgads:
Thanks @eriktorbjorn. Why didn't you implement the last suggestion then?

eriktorbjorn:
Because I was hoping for some further feedback on whether or not it was the correct approach.

orgads:
I also considered this option, but it looked too scary, so I limited it to monkey1/2.

But given that this was the behavior with SDL1 until 1.9, I suppose it should be ok.

I'm not an expert though :)

eriktorbjorn:
> I'm not an expert though :)

Me neither. I just wanted to make a reference to the older pull request here, so one of them wouldn't get left behind if the other is accepted. (A reference to this one was added automatically to mine when I did that.)

sev-:
Well, I see no better solution to this. Thank you, I'll merge and backport.

orgads:
I tested with more games, and it seems to be supported on all of them. See #3331.

eriktorbjorn:
Note that this change did not enable the Ctrl+W instant win cheat for older versions of Monkey Island 1 since it only checks GID_MONKEY, not GID_MONKEY_VGA or GID_MONKEY_EGA. (I haven't verified that the cheat is in the EGA version of the game, but it would surprise me if it isn't.)

orgads:
That's covered in #3331

PR #3325, opened 2021-08-29
SCUMM: Indy4 - Put coat/window in office, fixes bug #12420

BenCastricum:
WORKAROUND bug #12420 (also occurs in original) Broken window and coat missing

This happens when you skip the cutscenes in the beginning, in particular
the one where Indy enters the office for the first time. If object 23 (National
Archeology) is in possession of Indy (owner == 1) then it's safe the force the
coat (object 24) and broken window (object 25) into the room to compensate for
the skipped code.

### DIFF
--- a/engines/scumm/script_v5.cpp
+++ b/engines/scumm/script_v5.cpp
@@ -1386,7 +1386,7 @@ void ScummEngine_v5::o5_loadRoom() {
 	// WORKAROUND bug #12420 (also occurs in original) Broken window and coat missing
 	// This happens when you skip the cutscenes in the beginning, in particular
 	// the one where Indy enters the office for the first time. If object 23 (National
-	// Archeology) is in possession of Indy (owner == 1) then it's safe the force the
+	// Archeology) is in possession of Indy (owner == 1) then it's safe to force the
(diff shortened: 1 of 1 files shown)


eriktorbjorn:
I can confirm that the workaround seems to work equally well with the English Macintosh versions, both talkie and non-talkie.

engines/scumm/script_v5.cpp
```diff
@@ -1383,6 +1383,16 @@ void ScummEngine_v5::o5_loadRoom() {
 
 	room = getVarOrDirectByte(PARAM_1);
 
+	// WORKAROUND bug #12420 (also occurs in original) Broken window and coat missing
+	// This happens when you skip the cutscenes in the beginning, in particular
+	// the one where Indy enters the office for the first time. If object 23 (National
+	// Archeology) is in possession of Indy (owner == 1) then it's safe the force the
```
bluegr, line 1389:
```suggestion
	// Archeology) is in possession of Indy (owner == 1) then it's safe to force the
```

bluegr:
Nice and straightforward. I'm not sure if we still have time to add this to the 2.5.0 release.

Thanks! Merging

PR #3328, opened 2021-08-30
ASYLUM: Add detection entry for German Steam version

lotharsm:
This patch adds a detection entry for the German Steam version of Sanitarum.

Unfortunately, this version is currently completely broken and crashes with
```File::open: opening 'palette' failed
Assertion failed: _handle, file common/file.cpp, line 142
```
on startup. I'm adding this version basically just for reference at this point. Hopefully, I'll get my hands on a proper German CD release as well.


alxpnv:
@lotharsm  It should work now. Please make sure that the game path you've specified is something like `<steamdir>/steamapps/common/Sanitarium`

lotharsm:
Thanks @alxpnv, it's working now properly :-)

PR #3331, opened 2021-08-31
SCUMM: Extend Ctrl handling to all the games on versions 3-7

orgads:
Tested with Loom and The Dig. It seems to be consistent.

Sam&Max seems to have different mapping for Ctrl. The DOS version
has the following actions:

Ctrl-I: Rotates the current action
Ctrl-J: Switches to Joystick Mode
Ctrl-K: Shows memory stats
Ctrl-C: Prompts to quit
Ctrl-M: Switches to Mouse Mode

Anyway, since it doesn't break anything, I did not add an exclusion for it.

### DIFF
--- a/engines/scumm/input.cpp
+++ b/engines/scumm/input.cpp
@@ -619,10 +619,16 @@ void ScummEngine::processKeyboard(Common::KeyState lastKeyHit) {
 		    lastKeyHit.keycode <= Common::KEYCODE_F9) {
 			_mouseAndKeyboardStat = lastKeyHit.keycode - Common::KEYCODE_F1 + 315;
 
-		} else if ((_game.id == GID_MONKEY || _game.id == GID_MONKEY2) &&
-				   (lastKeyHit.flags & Common::KBD_CTRL) &&
+		} else if (lastKeyHit.flags & Common::KBD_CTRL && _game.version >= 3 && _game.version <= 7 &&
 				   (lastKeyHit.keycode >= Common::KEYCODE_a && lastKeyHit.keycode <= Common::KEYCODE_z)) {
-			_mouseAndKeyboardStat = lastKeyHit.keycode - Common::KEYCODE_a + 1;
+			// Some games (at least their DOS variants)
+			// expect Ctrl+A, B, C, etc. to generate codes 1, 2, 3, etc.
+			//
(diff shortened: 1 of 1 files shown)


eriktorbjorn:
Unless I messed up while testing, this makes it possible to switch between the three kids in Maniac Mansion with Ctrl+A, Ctrl+B and Ctrl+C. I can't do that when I run the original in DOSBox. There, it's F1, F2 and F3 instead, which already works in ScummVM even without this change.

As mentioned in my old pull request, outside of debug modes the games I could verify that use Ctrl+_key_ for anything are:

- Monkey Island 1 and 2 "ultimate talkies" use it for various settings.
- Monkey Island 1 uses Ctrl+W for an instant win cheat. I've verified this with the DOS CD, DOS VGA and Macintosh versions. Presumably it's also in the DOS EGA version, but I don't have that. Note that the current code only checks GID_MONKEY, not GID_MONKEY_EGA and GID_MONKEY_VGA.
- The EGA, VGA and FM Towns versions of Loom print version information on Ctrl+V. Apparently the Mac and PC-Engine versions don't?
- The "Passport to Adventure" and Monkey Island 1 demos also print version information on Ctrl+V.

Maybe it should depend on SCUMM version? According to my earlier notes, these games span versions 3-5.

orgads:
It worked for me on DOTT, CMI, FT, and The Dig, which are 6-8.

I just tested Maniac Mansion on Dosbox, and indeed Ctrl+Key doesn't do anything, except Ctrl+M, which acts like pressing Enter. So I'll limit it to `_game.version >= 3`.

orgads:
hmmm... Actually Sam&Max seems to behave differently. On Dosbox Ctrl-I switches between actions, and I brings up the inventory... Also, Ctrl-K shows memory usage (Heap 414K, ems 2000K). These don't work with the current mapping. And some of the items I included in the comment are actually handled by ScummVM before reaching this point.

orgads:
Wait before merging, needs more tests... And the commit message is wrong now.

orgads:
Ready for review.

bluegr:
Thanks! Clean implementation, which adds proper handling for control key combos in newer SCUMM versions. Thanks for testing this with all SCUMM games

Merging

PR #3332, opened 2021-08-31
AUDIO OPL2/3 EMULATION: Fix DOS_BOX OPL3 mode runnin as OPL2

Raffaello:
AUDIO OPL2/3 EMULATION: Fix DOS_BOX OPL3 mode runnin as OPL2

If running an OPL2 track on a DOS_BOX OPL3 emulation,
The buffer will be divided by 2 like if it is stereo,
but it will generate a mono audio stream, as opl3Active
flag is disabled, so it will act as an OPL2.
The error results in wrong sound produced.
This fix the issue when OPL3 is used as OPL2 and therefore
as a mono audio source.


Raffaello:
probably is not correct. i've just noticed that the pitch is higher..

PR #3335, opened 2021-09-01
GUI: persist the debug console history

mgerhardy:
Persisting the debug console history in the save dir of the engine (history.txt)

### DIFF
--- a/gui/console.cpp
+++ b/gui/console.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "gui/console.h"
+#include "common/savefile.h"
 #include "gui/widgets/scrollbar.h"
 #include "gui/ThemeEval.h"
 #include "gui/gui-manager.h"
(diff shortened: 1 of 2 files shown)


gui/console.cpp
```diff
@@ -582,6 +591,45 @@ void ConsoleDialog::killLastWord() {
 	}
 }
 
+void ConsoleDialog::loadHistory() {
+	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
+	Common::InSaveFile *loadFile = saveFileMan->openRawFile("history.txt");
```
sev-, line 596: Could we please name it `scummvm-history.txt` for lesser ambiguity

sev- (approved):
Great idea and feature. My only suggestion is regarding the file name.

gui/console.cpp
```diff
@@ -582,6 +591,45 @@ void ConsoleDialog::killLastWord() {
...
+	if (!loadFile) {
+		debug("Could not open history.txt");
```
bluegr, line 598: This will incorrectly pop up the first time the user tries to open the console (where there's no history).

It would be better if we could check for file existence instead, and throw this if the file exists but can't be loaded

bluegr:
Nice idea!

gui/console.cpp
mgerhardy: I think I'll just remove them

gui/console.cpp
mgerhardy: done

PR #3336, opened 2021-09-01
ASYLUM: Add detection for alternate demo

trembyle:
Found on PC Gamer covermount March 1998.

It uses AVI files (Cinepak 24-bit) instead of SMK.

Original disc is available here:
https://archive.org/details/pcgamer-disc-3.12-mar-1998


trembyle:
Lets

trembyle:
I'm going to resubmit this with commits for demos in other engines.

PR #3337, opened 2021-09-02
BURIED: Fix missing research points (bug 12881)

eriktorbjorn:
This should fix the missing research points described in https://bugs.scummvm.org/ticket/12881

Basically, the bug boiled down to the points only being awarded when pressing the "previous" button. And that would never take you to the pages that awarded the points.

Note that since I've never gotten around to playing the game, it's quite difficult for me to tell if I've accidentally mixed up anything. I'd be grateful if someone more familiar with the game could test it.

### DIFF
--- a/engines/buried/biochip_view.cpp
+++ b/engines/buried/biochip_view.cpp
@@ -746,10 +746,6 @@ void FilesBioChipViewWindow::onLButtonUp(const Common::Point &point, uint flags)
 
 		if (_curPage == 6)
 			((SceneViewWindow *)(_parent->getParent()))->getGlobalFlags().scoreResearchBCJumpsuit = 1;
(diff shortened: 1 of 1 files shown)


Thunderforge:
I've tested it out and it looks good to me. 

I don't see any unexpected side effects either, so I'm in favor of merging (and backporting to 2.3.x if possible).

Thunderforge:
For completeness, here are screenshots showing that the bug is fixed.

## Following PR instructions

Looking at the correct screens in Biochip - Files and then dying shows a Research Bonus of 300.

Following PR Instructions

## Full Game

A complete 15000 out of 15000 points can now be achieved.

100% Completion

sev-:
Excellent, thank you! I will also backport it

PR #3339, opened 2021-09-03
BURIED: Clear the live text window when moving, as advertised (bug 12887)

eriktorbjorn:
The comment says it will clear the live text window, but instead disabled the navigation arrows (I think), which it then does again right aftewards. A cut and paste error, perhaps? Anyway, I hope this produces the correct behavior as described in https://bugs.scummvm.org/ticket/12887

I also noticed that there are a number of updateLiveText("") calls (to clear the text), even though an empty string is the default value for it. But I didn't bother cleaning that up.


Thunderforge:
The live text now clears when moving forward, just like in the original engine. I tried it under several different situations and they all seem to be fixed now.

sev-:
Thank you!

PR #3340, opened 2021-09-04
COMMON: avoid bad rng seeds

Die4Ever:
RandomSource has some seeds that produce short loops

<details>
<summary>I wrote a test for this</summary>

```

#include <iostream>
#include <thread>
#include <vector>
#include <mutex>

std::mutex mtx;
typedef unsigned int uint32;

uint32 rng(uint32 _randSeed) {
	_randSeed = 0xDEADBF03 * (_randSeed + 1);
	_randSeed = (_randSeed >> 13) | (_randSeed << 19);
	return _randSeed;
}

int test_seed(uint32 orig_seed) {
	uint32 s = orig_seed;
	uint32 prev = s;
	for (int i = 0; i < 50; i++) {
		s = rng(s);
		if (s == orig_seed) return i+1;
		if (s == prev) return i + 1;
		prev = s;
	}
	return -1;
}

void test_seeds(uint32 first, uint32 last) {
	for (uint32 s = first; ; s++) {
		int loops = test_seed(s);
		if (loops != -1) {
			std::lock_guard<std::mutex> lck(mtx);
			std::cout << s << " failed with " << loops << " iterations\n";
		}
		if (s == last) break;
	}
}

void test_rng() {
	std::vector<std::thread> threads;
	uint32 start = 0;
	int num_threads = 12;
	uint32 max = UINT_MAX;
	//uint32 max = 100000000;
	uint32 each = max / num_threads;
	for (int i = 0; i < num_threads-1; i++) {
		threads.push_back(std::thread(test_seeds, start, start +each));
		start += each + 1;
	}
	threads.push_back(std::thread(test_seeds, start, max));

	for (uint32 i = 0; i < threads.size(); i++) {
		threads[i].join();
	}
}

int main()
{
	std::cout << "start!\n\n";
	test_rng();
}

```

</details>

<details>
<summary>which produced this output</summary>

```

1800586416 failed with 48 iterations
1087576404 failed with 48 iterations
3593762925 failed with 11 iterations
730316237 failed with 48 iterations
2527622093 failed with 48 iterations
3969000752 failed with 48 iterations
3997635582 failed with 48 iterations
782266141 failed with 48 iterations
2923531432 failed with 48 iterations
3285614266 failed with 48 iterations
790957358 failed with 48 iterations
2228155406 failed with 11 iterations
2934411674 failed with 48 iterations
3295602753 failed with 48 iterations
3662902891 failed with 11 iterations
450572919 failed with 48 iterations
2598359503 failed with 48 iterations
[...]

PR #3341, opened 2021-09-04
COMMON: Change PRNG Function to Xorshift

TehGelly:
The old RNG method had non-standard periods, ranging from some seeds looping on themselves (seed = 1184201285) to some seeds having periods as low as 11 or 48, as listed in https://github.com/scummvm/scummvm/pull/3340. This is a problem even for games that run the RNG once a frame, as the possibilities for random events is greatly reduced should the initial seed be in one of these sets of small periods.

Xorshift is a standard, fast, non-cryptographic PRNG with academic backing that has period 2^32-1 (all seeds lead to another seed except 0, which is excluded from the initial seeds).  Many different flavors are possible, as listed in the paper, but the choice implemented in this pull request uses only a single 32-bit integer as a state, like the old PRNG.

### DIFF
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -73,7 +73,7 @@ int RandomSource::getRandomNumberRngSigned(int min, int max) {
 	return getRandomNumber(max - min) + min;
 }
 		
-private inline void RandomSource::scrambleSeed() {
+inline void RandomSource::scrambleSeed() {
 	//marsaglia's paper says that any of 81 triplets are feasible
 	//(11,21,13) was chosen, with (cba) and (>>,<<,>>)
 	_randSeed ^= _randSeed >> 13;
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -48,7 +48,7 @@ RandomSource::RandomSource(const String &name) {
 }
 
 void RandomSource::setSeed(uint32 seed) {
-	if(seed == 0)
+	if (seed == 0)
 		seed++;
 	_randSeed = seed;
 }
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -52,20 +52,12 @@ void RandomSource::setSeed(uint32 seed) {
 		seed++;
 	_randSeed = seed;
 }
-	
-inline void RandomSource::scrambleSeed() {
-	//marsaglia's paper says that any of 81 triplets are feasible
-	//(11,21,13) was chosen, with (cba) and (>>,<<,>>)
-	_randSeed ^= _randSeed >> 13;
-	_randSeed ^= _randSeed << 21;
-	_randSeed ^= _randSeed >> 11;
-}
 
 uint RandomSource::getRandomNumber(uint max) {
 	scrambleSeed();
 	if (max == UINT_MAX)
 		return _randSeed;
-	return _randSeed % (max + 1);
+	return (_randSeed * 0xDEADBF03) % (max + 1);
 }
 
 uint RandomSource::getRandomBit() {
@@ -79,6 +71,14 @@ uint RandomSource::getRandomNumberRng(uint min, uint max) {
 
 int RandomSource::getRandomNumberRngSigned(int min, int max) {
 	return getRandomNumber(max - min) + min;
+}
+		
+private inline void RandomSource::scrambleSeed() {
+	//marsaglia's paper says that any of 81 triplets are feasible
+	//(11,21,13) was chosen, with (cba) and (>>,<<,>>)
+	_randSeed ^= _randSeed >> 13;
+	_randSeed ^= _randSeed << 21;
+	_randSeed ^= _randSeed >> 11;
 }
 
 } // End of namespace Common
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -56,7 +56,7 @@ void RandomSource::setSeed(uint32 seed) {
 uint RandomSource::getRandomNumber(uint max) {
 	scrambleSeed();
 	if (max == UINT_MAX)
-		return _randSeed;
+		return (_randSeed * 0xDEADBF03);
 	return (_randSeed * 0xDEADBF03) % (max + 1);
 }
 
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -43,13 +43,13 @@ RandomSource::RandomSource(const String &name) {
 	newSeed += time.tm_mday * 86400 + time.tm_mon * 86400 * 31;
 	newSeed += time.tm_year * 86400 * 366;
 	newSeed = newSeed * 1000 + g_system->getMillis();
-	if(newSeed == 0)
-		newSeed++;
 	setSeed(newSeed);
 #endif
 }
 
 void RandomSource::setSeed(uint32 seed) {
+	if(seed == 0)
+		seed++;
 	_randSeed = seed;
 }
 	
--- a/common/random.cpp
+++ b/common/random.cpp
@@ -43,6 +43,8 @@ RandomSource::RandomSource(const String &name) {
 	newSeed += time.tm_mday * 86400 + time.tm_mon * 86400 * 31;
 	newSeed += time.tm_year * 86400 * 366;
 	newSeed = newSeed * 1000 + g_system->getMillis();
+	if(newSeed == 0)
+		newSeed++;
 	setSeed(newSeed);
 #endif
 }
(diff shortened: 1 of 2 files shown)


Die4Ever:
looks good to me, I used my code to verify no short loops

<details>
<summary>test code</summary>

```

#include <iostream>
#include <thread>
#include <vector>
#include <mutex>

std::mutex mtx;
typedef unsigned int uint32;

/*uint32 rng(uint32 _randSeed) {
	_randSeed = 0xDEADBF03 * (_randSeed + 1);
	_randSeed = (_randSeed >> 13) | (_randSeed << 19);
	return _randSeed;
}*/

uint32 rng(uint32 _randSeed) {
	//marsaglia's paper says that any of 81 triplets are feasible
	//(11,21,13) was chosen, with (cba) and (>>,<<,>>)
	_randSeed ^= _randSeed >> 13;
	_randSeed ^= _randSeed << 21;
	_randSeed ^= _randSeed >> 11;
	return _randSeed;
}

int test_seed(uint32 orig_seed) {
	uint32 s = orig_seed;
	uint32 prev = s;
	for (int i = 0; i < 100; i++) {
		s = rng(s);
		if (s == orig_seed) return i+1;
		if (s == prev) return i + 1;
		prev = s;
	}
	return -1;
}

void test_seeds(uint32 first, uint32 last) {
	for (uint32 s = first; ; s++) {
		int loops = test_seed(s);
		if (loops != -1) {
			std::lock_guard<std::mutex> lck(mtx);
			std::cout << s << " failed with " << loops << " iterations\n";
		}
		if (s == last) break;
	}
}

void test_rng() {
	std::vector<std::thread> threads;
	uint32 start = 0;
	int num_threads = 12;
	uint32 max = UINT_MAX;
	//uint32 max = 10000000;
	uint32 each = max / num_threads;
	for (int i = 0; i < num_threads-1; i++) {
		threads.push_back(std::thread(test_seeds, start, start +each));
		start += each + 1;
	}
	threads.push_back(std::thread(test_seeds, start, max));

	for (uint32 i = 0; i < threads.size(); i++) {
		threads[i].join();
	}
}

int main()
{
	/*auto r = rand();
	for (int i = 0; i < 1000000; i++)
		r = std::max(r, rand());
	std::cout << "r == " << r << "\n";
	return 0;*/
	std::cout << "start!\n\n";
	test_rng();
}

```

</details>

TehGelly:
It would be possible to upgrade this to Xorshift*, a much more secure PRNG, by changing the return line in getRandomNumber to `return (_randSeed * 0xDEADBF03) % (max + 1);`, and would also re-introduce the magic constant 0xDEADBF03 back into the PRNG - one of my favorite hallmarks of the old PRNG.

I leave it open, however, as it would introduce an additional multiplication, which seems to be important to this project -  getRandomBit was intentionally shifted to save on computation.

common/random.h
```diff
@@ -39,8 +39,8 @@ namespace Common {
 class String;
 
 /**
- * Simple random number generator. Although it is definitely not suitable for
- * cryptographic purposes, it serves our purposes just fine.
+ * Xorshift PRNG. Replaces the previous PRNG, as
+ * the cycles within it were detrimental to gameplay.
```
criezy, line 43:
```suggestion
 * Xorshift random number generator. Although it is definitely not suitable for
 * cryptographic purposes, it serves our purposes just fine.
```
I don't think we need information about the previous state of the code in our doxygen documentation. What matters is the current state. Also we might want to avoid acronyms as some developers who simply want a random number for their engine might not know what `PRNG` means.

common/random.h
```diff
@@ -60,6 +60,13 @@ class RandomSource {
 	uint32 getSeed() const { /*!< Get a random seed that can be used to initialize the RNG. */
 		return _randSeed;
 	}
+	
+	/**
+	* Scrambles the seed in order to get a new result.
+	* Code is shared between getRandomNumber and getRandomBit,
+	* so it is split off for clarity.
+	*/
+	inline void scrambleSeed();
```
criezy, line 69: I would suggest to move this to the end of the class declaration and make it private.

Since c++ requires an inline function to be defined in every translation unit in which it is used, but you define it only in `random.cpp`, it is not a good idea to open the door to the possibility to use the function elsewhere. Making it private will ensure that does not happen.

criezy (commented):
The change looks good to me, although I suggested a couple of change.

Also the first two commits (627ee43 and 7a1414a) will need to be squashed as they are two parts of a single change (if we checkout 627ee43, it will not compile). You can either do that before we merge the pull request, or we can squash all three commits when we decide to merge it.

Using an Xorshift* generator seems fine to me. ScummVM can run on devices with low processing power, but I don't expect that an additional multiplication would be an issue even for those. Somebody with more experience than me with those devices might disagree though. And the change as it is in this pull request should be sufficient for what we need.

common/random.h
TehGelly: Fixed in commit 17c0c6f

common/random.h
TehGelly: Fixed in commit 17c0c6f, and random.cpp was adjusted with commit aa765e4

TehGelly:
I have made the changes as suggested. Squashing the commits on merge seems to be the easiest and cleanest way to handle that issue, as I think the two commits I have just made are similarly problematic.

I have also made the PRNG into Xorshift* with your approval 👍

common/random.h
```diff
@@ -90,6 +90,13 @@ class RandomSource {
 	 * @return	a random number in the interval [min, max]
 	 */
 	int getRandomNumberRngSigned(int min, int max);
+	
+	/**
+	* Scrambles the seed in order to get a new result.
+	* Code is shared between getRandomNumber and getRandomBit,
+	* so it is split off for clarity.
+	*/
+	private inline void scrambleSeed();
```
criezy, line 99:
```suggestion
private:
	inline void scrambleSeed();
```

common/random.cpp
```diff
@@ -48,21 +48,20 @@ RandomSource::RandomSource(const String &name) {
...
 	_randSeed = seed;
 }
 
 uint RandomSource::getRandomNumber(uint max) {
-	_randSeed = 0xDEADBF03 * (_randSeed + 1);
-	_randSeed = (_randSeed >> 13) | (_randSeed << 19);
-
+	scrambleSeed();
 	if (max == UINT_MAX)
 		return _randSeed;
```
Die4Ever, line 59: should this also be multiplied by 0xDEADBF03 for this return?

TehGelly: Fixed in a0bc664.

common/random.cpp
```diff
@@ -48,21 +48,20 @@ RandomSource::RandomSource(const String &name) {
 }
 
 void RandomSource::setSeed(uint32 seed) {
+	if(seed == 0)
```
bluegr, line 51:
```suggestion
	if (seed == 0)
```

common/random.cpp
```diff
@@ -73,5 +72,13 @@ uint RandomSource::getRandomNumberRng(uint min, uint max) {
 int RandomSource::getRandomNumberRngSigned(int min, int max) {
 	return getRandomNumber(max - min) + min;
 }
+		
+inline void RandomSource::scrambleSeed() {
+	//marsaglia's paper says that any of 81 triplets are feasible
+	//(11,21,13) was chosen, with (cba) and (>>,<<,>>)
```
bluegr, line 78: Reading through the paper, Marsaglia suggests a specific triplet for 32-bit seeds, which will pass almost all sets of randomness: << 13 , >> 17, << 5. Further down he does say that any of the choices he provides in the paper will do as well, but he has extensively tested specific ones like this. Perhaps we should use that triplet?

TehGelly: The exact wording in the paper is "It uses one of my favorite choices, [a, b, c] = [13, 17, 5], and will pass almost all tests of randomness, except the binary rank test in Diehard." For context, Diehard is an older battery of tests developed to test PRNGs, and has been superseded by more modern tests, such as Dieharder and TestU01. 

Using TestU01's SmallCrush on the Xorshift* version of this algorithm, all triplets and variations pass the exact same tests and fail BirthdaySpacings, a test that would be impossible to feel in a non-cryptographic context. So, personal preference notwithstanding, there's no real benefit to choosing certain triplets above others for this.

Marsaglia's comment above on the matter is that (13,17,5) is one of his personal favorites, and in much the same way, (11,21,13) is a personal favorite of mine :)

TehGelly: I decided to throw TestU01 at all possible xorshift* pairings, with initial seed = 1, and the best result for BirthdaySpacings still fails, with p = 1.6*10^-16 (>>2, <<7, >>9). This is better than Marsaglia or my preferences, but still fails the test overall. I would not mind using this set of values instead, but it certainly would feel like splitting hairs.

bluegr (approved):

bluegr:
Overall, very nice work! Thanks for your research and for your work on this

criezy (approved):
It looks good to me now.
We just have to remember it needs to be squashed when merging.

bluegr:
Thanks again. Squashing the commits on this one

PR #3342, opened 2021-09-04
EVENTRECORDER: Fix synchronisation, add 'update' mode

moralrecordings:
- Screenshot frequency can now be specified from the command line.
- Screen redraws are now tracked, and all events and screenshots are force-aligned with a screen redraw.
- A few cases of recursion in the event processing stream have been removed.
- Compatibility with the previous version of Event Recorder file format has been dropped, on the advice that no testing efforts use this feature yet.
- A new 'update' mode has been added to the EventRecorder. This new mode is functionally identical to playback mode, however
a new recording file is used to track the actual output of ScummVM. This feature can be used to update a suite of existing recordings after a renderer or a timing change.

### DIFF
--- a/common/recorderfile.cpp
+++ b/common/recorderfile.cpp
@@ -132,7 +132,7 @@ bool PlaybackFile::parseHeader() {
 }
 
 bool PlaybackFile::checkPlaybackFileVersion() {
-	_version = _readStream->readUint32LE();
+	_version = _readStream->readUint32BE();
 	switch (_version) {
 	case 2:
 		break;
@@ -167,8 +167,8 @@ String PlaybackFile::readString(int len) {
 }
 
 bool PlaybackFile::readChunkHeader(PlaybackFile::ChunkHeader &nextChunk) {
-	nextChunk.id = (FileTag)_readStream->readUint32LE();
-	nextChunk.len = _readStream->readUint32LE();
+	nextChunk.id = (FileTag)_readStream->readUint32BE();
+	nextChunk.len = _readStream->readUint32BE();
 	return !_readStream->err() && !_readStream->eos();
 }
 
@@ -286,7 +286,7 @@ void PlaybackFile::readHashMap(ChunkHeader chunk) {
 
 void PlaybackFile::processRndSeedRecord(ChunkHeader chunk) {
 	String randomSourceName = readString(chunk.len - 4);
-	uint32 randomSourceSeed = _readStream->readUint32LE();
+	uint32 randomSourceSeed = _readStream->readUint32BE();
 	_header.randomSourceRecords[randomSourceName] = randomSourceSeed;
 }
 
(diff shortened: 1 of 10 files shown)


gui/EventRecorder.h
```diff
@@ -151,13 +154,23 @@ class EventRecorder : private Common::EventSource, public Common::Singleton<Even
...
-	void resumeRecording() {
-		_initialized = _savedState;
+	void releaseRecording() {
+		assert(_acquireCount > 0);
+		_acquireCount -= 1;
+		if (_acquireCount == 0)
+			_initialized = _savedState;
+	}
+
+	RecordMode getRecordMode() {
```
mgerhardy, line 172: this should be `const`

gui/EventRecorder.h
```diff
@@ -151,13 +154,23 @@ class EventRecorder : private Common::EventSource, public Common::Singleton<Even
 	void deleteRecord(const Common::String& fileName);
 	bool checkForContinueGame();
 
-	void suspendRecording() {
-		_savedState = _initialized;
-		_initialized = false;
+	void acquireRecording() {
+		if (_acquireCount == 0) {
```
mgerhardy, line 158: maybe an assert with `_acquireCount >= 0` would be helpful here?

common/recorderfile.h
```diff
@@ -157,6 +158,9 @@ class PlaybackFile {
 	PlaybackFileHeader &getHeader() {return _header;}
 	void updateHeader();
 	void addSaveFile(const String &fileName, InSaveFile *saveStream);
+
+	uint32 getVersion() {return _version;}
```
mgerhardy, line 162: These could be `const`

common/recorderfile.cpp
```diff
@@ -130,12 +132,20 @@ bool PlaybackFile::parseHeader() {
...
+	case 2:
+		break;
+	case 1:
+		_trackScreenUpdate = false;
+		break;
+	default:
+		warning("Unknown playback file version %d. Maximum supported version is %d.", _version, RECORD_VERSION);
 		return false;
 	}
+	if (_version != RECORD_VERSION) {
```
mgerhardy, line 146: Last time modified the event recorder @sev-  said that bumping the version of the recorded file isn't needed, as there basically aren't any wild living recorded files around yet. I suppose this is still true.

mgerhardy (commented):
Awesome work - next step to automate engine tests.

common/recorderfile.cpp
moralrecordings: No worries, I've removed the backwards-compatibility code. On the basis that it's okay for older files not to work, I've taken the opportunity to switch the file format to using big-endian everywhere, mostly so that external tooling doesn't need to make a special parsing exception for the screenshot THMB block.

mgerhardy (approved):

PR #3343, opened 2021-09-04
COMMON: Use const references where possible in punycode

mduggan:
I saw this in the coverity report (CID 1461548 and 1461550) and it seemed easy enough to fix.


bluegr:
Thanks! Quite straightforward. Merging

PR #3347, opened 2021-09-05
COMMON: Don't decompress more bytes than the file specifies (bug 12900)

eriktorbjorn:
When running my CD version of Buried in Time, the loadFromCompresedEXE() function would try to decompress beyond the allocated buffer, causing it to crash. Up to that point, the decompressed data was identical to what I would get if I let the original installer decompress the EXE and DLL files.

See https://bugs.scummvm.org/ticket/12900 for details.


bluegr:
Great! Thanks :)

eriktorbjorn:
Should this be backported to the release branch?

PR #3354, opened 2021-09-07
[GRIM] Fix subtitle alignment for OpenGL with shaders

somaen:
This is a quick attempt at trying to fix the broken positioning of text in Grim when using the OpenGL shader renderer.

@eriktorbjorn noted that this may stem from the OpenGL (with shaders)-renderer lacking a proper implementation of getScreenBoundingBox, so I went ahead and adapted one.

This was done by comparing the getScreenBoundingBox implementations for EFMI/Grim in the OpenGL-renderer, and noticing that those were only different in how they got at their data, as there was already an implementation for EFMI in the shader-renderer, I copy pasted that code, and adapted the differences I found when comparing the OpenGL-implementation, which seems to improve things.

### DIFF
--- a/engines/grim/gfx_opengl_shaders.cpp
+++ b/engines/grim/gfx_opengl_shaders.cpp
@@ -518,8 +518,75 @@ void GfxOpenGLS::flipBuffer() {
 	g_system->updateScreen();
 }
 
-void GfxOpenGLS::getScreenBoundingBox(const Mesh *mesh, int *x1, int *y1, int *x2, int *y2) {
+void GfxOpenGLS::getScreenBoundingBox(const Mesh *model, int *x1, int *y1, int *x2, int *y2) {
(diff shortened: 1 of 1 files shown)


somaen:
There, fixed the rebasing.

eriktorbjorn:
Either I screwed up, or this still puts the subtitles at the bottom of the screen with the OpenGL with shaders renderer. (Though with the release branch, this is an improvement since there I don't get any subtitles at all without the patch. Perhaps because I compiled it in release mode, i.e. with optimizations? I seem to recall that I came to the tentative conclusion that the current behavior might be undefined.)

aquadran:
I checked that, this is not correct. model and projection matrix is different

aquadran:
let's merge this for now, and fix it later with another PR

PR #3356, opened 2021-09-09
AGS: Change a few char variables to int8

antoniou79:
This is done to explicitly specify that they should be treated as signed

Some ports may treat an (unspecified) "char" as unsigned by default. Android does this for its ARM architectures and while we did fix this by enforcing -fsigned-char to its compiler, the issue may come up for other ports. Also it should be better to clarify in the engine when a variable is not actually storing string characters.

I've spotted and changed the most "safe" cases I could find for this. As far as I can tell, out of these, only the Outline field and fontoutline[] array could potentially be assigned valid negative values which would cause bugs for ports treating chars as unsigned.

### DIFF
--- a/engines/ags/engine/ac/character_extras.h
+++ b/engines/ags/engine/ac/character_extras.h
@@ -52,8 +52,8 @@ struct CharacterExtras {
 	short tint_b;
 	short tint_level;
 	short tint_light;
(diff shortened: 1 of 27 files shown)


dreammaster:
Thanks for doing review. It's appreciated.

PR #3359, opened 2021-09-10
SLUDGE: Fix build with Clang >= 11.0.0

knobix:
Clang 11 and newer check the types of return values more strictly and
generate an error if there is a discrepancy.

However the "bool loadGame()" function, returns the value "NULL", if
loading of a savegame was unsucessful, whereas Clang expects a return
value of type "bool".

Remedy the issue by using "false" as return value instead of "NULL".

### DIFF
--- a/engines/sludge/saveload.cpp
+++ b/engines/sludge/saveload.cpp
@@ -192,7 +192,7 @@ bool loadGame(const Common::String &fname) {
 		headerBad = true;
 	if (headerBad) {
 		fatal(ERROR_GAME_LOAD_NO, fname);
-		return NULL;
+		return false;
 	}
 	char c;
 	c = fp->readByte();


knobix:
On `FreeBSD 13.0` (amd64 or i386), which has `Clang 11.0.1`, the build with `--enable-all-engines` for the pre-release `2.3.0` fails with following error:

```
engines/sludge/saveload.cpp:195:10: error: cannot initialize return object of type 'bool' with an rvalue of type 'nullptr_t'
                return NULL;
                       ^~~~
/usr/include/sys/_null.h:37:14: note: expanded from macro 'NULL'
#define NULL    nullptr
                ^~~~~~~
1 error generated.
gmake[2]: *** [Makefile.common:164: engines/sludge/saveload.o] Error 1
```

This issue does not appear with `FreeBSD 11.4` and `12.2`, which have `Clang 10.0.0` and `10.0.1`, respectively.

bluegr:
Nice catch, thanks! Merging

PR #3360, opened 2021-09-11
BASE: Don't check for config key presence when loading gfx mode

lotharsm:
I noticed that while bilinear filtering is disabled by default with ``RegisterDefault();`` at startup, bilinear filtering is always applied when starting a game (at least on the OPENGLSDL backend) unless you explicitly set ``filtering=false`` in the configuration file by simply opening the options dialog, applying the settings and closing it again.

While researching this, I think the culprit is that main.cpp checks for ``ConfMan.hasKey("filtering")``, but ``registerDefault()`` technically won't set a configuration key. This means that this checks override our defaults.

I think that this check is not necessary. With the checks for ``hasKey`` removed, it will first try to load the values set in the configuration file and otherwise fall back to the values set by ``registerDefault()``  instead.

Since we only deal with ``bool`` values here, it shouldn't even break in case we don't use ``registerDefault`` for those values since in this case, ``ConfMan.getBool`` will simply return ``false`` as well (which are the default values we set with ``registerDefault``.

### DIFF
--- a/base/main.cpp
+++ b/base/main.cpp
@@ -351,12 +351,10 @@ static void setupGraphics(OSystem &system) {
 
 		system.initSize(320, 200);
 
-		if (ConfMan.hasKey("aspect_ratio"))
-			system.setFeatureState(OSystem::kFeatureAspectRatioCorrection, ConfMan.getBool("aspect_ratio"));
-		if (ConfMan.hasKey("fullscreen"))
-			system.setFeatureState(OSystem::kFeatureFullscreenMode, ConfMan.getBool("fullscreen"));
-		if (ConfMan.hasKey("filtering"))
-			system.setFeatureState(OSystem::kFeatureFilteringMode, ConfMan.getBool("filtering"));
+		// Parse graphics configuration, implicit fallback to defaults set with RegisterDefaults()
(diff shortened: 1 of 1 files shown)


Die4Ever:
for reference
```
bool ConfigManager::getBool(const String &key, const String &domName) const {
    String value(get(key, domName));
    bool val;
    if (parseBool(value, val))
        return val;

    error("ConfigManager::getBool(%s,%s): '%s' is not a valid bool",
          key.c_str(), domName.c_str(), value.c_str());
}
```

```
bool parseBool(const String &val, bool &valAsBool) {
    if (val.equalsIgnoreCase("true") ||
        val.equalsIgnoreCase("yes") ||
        val.equals("1")) {
        valAsBool = true;
        return true;
    }
    if (val.equalsIgnoreCase("false") ||
        val.equalsIgnoreCase("no") ||
        val.equals("0") ||
        val.empty()) {
        valAsBool = false;
        return true;
    }

    return false;
}
```

the `parseBool` function checks for `val.empty()` and will set `valAsBool` to `false`, and return `true` for success
so this code is good

Die4Ever (approved):
looks good, tested it out real quick with
```
    if (ConfMan.getBool("fake_config_option"))
        error("BAD!");
```

PR #3361, opened 2021-09-11
AGS: Only load ttf font as outline for ttf font

antoniou79:
Possible fix for https://bugs.scummvm.org/ticket/12923

This is quite an oblivious fix, though, since I use FontInfo Flags field (which as far as I can tell is unused) to set bit flags to track if a loaded font uses a TTF or WFN renderer and also I renamed a method (get_outline_font() to get_outline_ttf_font()) to specify that it's specifically for TTF fonts use cases.

Maybe there's a better way of fixing this.

### DIFF
--- a/engines/ags/shared/ac/game_struct_defines.h
+++ b/engines/ags/shared/ac/game_struct_defines.h
@@ -226,6 +226,10 @@ struct SpriteInfo {
 	}
 };
 
+// Font renderer flags
+#define FONT_TTF_RENDERER   0x01
+#define FONT_WFN_RENDERER   0x02
+
 // Various font parameters, defining and extending font rendering behavior.
 // While FontRenderer object's main goal is to render single line of text at
 // the strictly determined position on canvas, FontInfo may additionally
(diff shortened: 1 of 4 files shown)


dreammaster:
Actually, I had to revert the commit, sorry. It reintroduced the very problem that the "outilne font" commits worked around in the first place. Try running Maniac Mansion Deluxe, and skipping the intro. As the three selected characters chat as they arrive at the mansion, you'll see outlines getting offset from the solid fonts on top of them.

antoniou79:
> Actually, I had to revert the commit, sorry. It reintroduced the very problem that the "outilne font" commits worked around in the first place. Try running Maniac Mansion Deluxe, and skipping the intro. As the three selected characters chat as they arrive at the mansion, you'll see outlines getting offset from the solid fonts on top of them.

Right, sorry for that. Thank for looking into it. 
For reference, I checked my changes again today and there were a number of issues:
1. Setting the fonts as TTF or WFN in wloadfont_size() was (almost always?) overridden by the code below that set the Info filed from the font_info parameter of wloadfont_size():
https://github.com/scummvm/scummvm/blob/dc3919b413b52bb4f3137becc5535720613d932f/engines/ags/shared/font/fonts.cpp#L344
I completely missed this.
2. Searching for fontinfo in the engine code, I can find numerous other places where fontinfo is set explicitly from Flags, or Flags in fontInfo are set from data parameters and there are already defined for bit flags for the Flags field (ie. ). Eg. apparently FFLG_SIZEMULTIPLIER (1) is a legacy one which directly conflicts with me setting FONT_TTF_RENDERER to 1.
- SetFontInfoFromLegacyFlags() and AdjustFontInfoUsingFlags() in ags\shared\ac\game_setup_struct.cpp
- set_fontinfo() (which seems unused) in ags\shared\font\fonts.cpp
- UpgradeFonts() in ags\shared\game\main_game_file.cpp
There's just potential that my fix introduces regressions with those too.
I also seem to completely have missed this.
So overall, this PR would need a big rework and check all the affected cases to work. 
Since just removing the assert in TFFontRenderer::RenderText() is a more simple fix, it's probably better to ignore this PR.

PR #3369, opened 2021-09-15
COMMON: Add error for unsupported manual save slot

antoniou79:
Addresses ticket https://bugs.scummvm.org/ticket/12766

Currently used by SKY engine.

Allow engines to prevent user from manually saving at selected slots, if they do not support them.

### DIFF
--- a/common/error.cpp
+++ b/common/error.cpp
@@ -72,6 +72,9 @@ static String errorToString(ErrorCode errorCode) {
 	case kUserCanceled:
 		return _s("User canceled");
 
+	case kManualSaveSlotNotSupportedByNativeEngine:
+		return _s("Manual saving to this particular slot is not supported. Please try another slot");
+
 	case kUnknownError:
 	default:
 		return _s("Unknown error");
--- a/common/error.h
+++ b/common/error.h
@@ -73,6 +73,15 @@ enum ErrorCode {
 
 	kUserCanceled,			///< User has canceled the launching of the game.
 
+	/** Manually saving to this slot is not supported (yet or at all) by the game engine.
+	 * For example:
+	 * Some engines (eg. SKY) show their save slots starting from slot 1.
+	 * GMM allows saving to slot 0, and also that slot can optionally be used for autosaves, if the game engine supports them.
+	 * If the user manually saves, using GMM, to an unsupported slot by the in-game native UI, 
+	 * they won't be able to find and restore their save game from the native UI.
+	 */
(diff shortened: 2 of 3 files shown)


common/error.cpp
```diff
@@ -72,6 +72,9 @@ static String errorToString(ErrorCode errorCode) {
 	case kUserCanceled:
 		return _s("User canceled");
```
lotharsm, line 74: Minor nitpick - maybe go with "Canceled by user" instead?

antoniou79: > Minor nitpick - maybe go with "Canceled by user" instead?
I agree, it would be better to phrase it that way. However, this message is not part of the changes of this PR.

lotharsm: Uh, right, I totally missed that. Sorry for the noise ;)

bluegr:
I believe it would be easier to write protect the autosave slot instead.

We already have support for this, e.g. check agi/metaengine.cpp:303:

		SaveStateDescriptor emptySave;
		// Do not allow save slot 0 (used for auto-saving) to be overwritten.
		if (slotNr == 0) {
			emptySave.setAutosave(true);
			emptySave.setWriteProtectedFlag(true);
		} else {
			emptySave.setWriteProtectedFlag(false);
		}
		return emptySave;

antoniou79:
> I believe it would be easier to write protect the autosave slot instead.
> 
> We already have support for this, e.g. check agi/metaengine.cpp:303:
> 
> ```
> 	SaveStateDescriptor emptySave;
> 	// Do not allow save slot 0 (used for auto-saving) to be overwritten.
> 	if (slotNr == 0) {
> 		emptySave.setAutosave(true);
> 		emptySave.setWriteProtectedFlag(true);
> 	} else {
> 		emptySave.setWriteProtectedFlag(false);
> 	}
> 	return emptySave;
> ```

Currently looking into that and have a tentative implementation ready, but I will test a bit more. Also I've discovered another save game related bug (basically save deletion bug) that I intend to fix. 

From what I understand, setting the slot as write protected will disable the option to "Save" in it, rename it or "Delete" it. It does not show any message when attempting to do any of these actions, but the buttons get disabled. 

I think it's close enough to what I want so I will close this PR in favor of the setWriteProtectedFlag() approach.

antoniou79:
Closing as superseded by commit:
https://github.com/scummvm/scummvm/commit/1712cceb205e0fcc94e8327406b15279e8a5e227

PR #3371, opened 2021-09-15
EVENTRECORDER: don't show the splash screen if event recorder is active

mgerhardy:
The splash screen must get disabled or it consumes events and milliseconds of the event recorder.


sev-:
No, splash screen must always be present. Pause the event recorder playback for it.

PR #3373, opened 2021-09-16
GROOVIE: Clandestiny fixes

Die4Ever:
Fixes crashes when you have saved games. Also makes the preview images on saved games work.

Also fixes Mac/iOS cursors and music, because they use slightly different filenames for them.

### DIFF
--- a/engines/groovie/cursor.cpp
+++ b/engines/groovie/cursor.cpp
@@ -384,7 +384,7 @@ GrvCursorMan_v2::GrvCursorMan_v2(OSystem *system) :
 
 	// Open the icons file
 	Common::File iconsFile;
-	if (!iconsFile.open("icons.ph"))
+	if (!iconsFile.open("icons.ph") && !iconsFile.open("icons.bin"))
 		error("Groovie::Cursor: Couldn't open icons.ph");
 
 	// Verify the signature
--- a/engines/groovie/cursor.cpp
+++ b/engines/groovie/cursor.cpp
@@ -385,13 +385,13 @@ GrvCursorMan_v2::GrvCursorMan_v2(OSystem *system) :
 	// Open the icons file
 	Common::File iconsFile;
 	if (!iconsFile.open("icons.ph") && !iconsFile.open("icons.bin"))
-		error("Groovie::Cursor: Couldn't open icons.ph");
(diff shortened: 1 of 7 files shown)


engines/groovie/cursor.cpp
```diff
@@ -384,7 +384,7 @@ GrvCursorMan_v2::GrvCursorMan_v2(OSystem *system) :
 
 	// Open the icons file
 	Common::File iconsFile;
-	if (!iconsFile.open("icons.ph"))
+	if (!iconsFile.open("icons.ph") && !iconsFile.open("icons.bin"))
 		error("Groovie::Cursor: Couldn't open icons.ph");
```
Thunderforge, line 388: Shouldn't the error message be updated to also include `icons.bin`?

Die4Ever: hmm yea, maybe it should just say

"Groovie::Cursor: Couldn't open icons.ph or icons.bin"?

Thunderforge: Either that or just "Couldn't open icons"

engines/groovie/music.cpp
```diff
@@ -793,8 +793,15 @@ bool MusicPlayerTlc::load(uint32 fileref, bool loop) {
 	unload();
 	_file = new Common::File();
 
-	// Create the audio stream from fileref
 	Common::String filename = getFilename(fileref);
+
+	// Apple platforms use m4a files instead of mpg
+	if (_vm->getPlatform() == Common::kPlatformMacintosh || _vm->getPlatform() == Common::kPlatformIOS)
+		filename += ".m4a";
```
Thunderforge, line 800: Turns out that if you load a `.m4a` with `Audio::makeMP3Stream()` below, then the first note keeps repeating. Instead, `Audio::makeQuickTimeStream()` does load the music properly. So you'll need to switch which way you are loading the file based on which extension we have (or as we speculated in Discord, see if there's a way ScummVM has to automatically pick what type of stream to use).

bluegr:
Nice work, thanks! Merging

scott-t:
What's the general procedure for these 'universal' apps? GRL published both the Mac App Store and iOS variants more or less at the same time, so it makes sense the releases would be basically identical.

PR #3375, opened 2021-09-16
ENGINES: Fix moving savegames to a new free slot

orgads:
This was broken for most engines.


bluegr:
I see that you've written this part of the code. Thanks for the fix! Merging

PR #3376, opened 2021-09-16
ENGINES: Warn when autosave is overwritten by a newer game

orgads:
If the previous autosave had a longer gameplay duration, and the user
started a new game, warn before overwriting the more advanced autosave.

### DIFF
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -570,7 +570,7 @@ bool Engine::warnBeforeOverwritingAutosave() {
 	if (desc.getSaveSlot() == -1)
 		return true;
 	if (desc.hasAutosaveName())
-		return true;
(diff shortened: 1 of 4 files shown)


bluegr:
This could result in a constant nagging every time an autosave gets created, if the user declines to overwrite the previous autosave. Are we sure we want this? It could turn out to be an annoying popup for users.

orgads:
I started a discussion on discord. Should have been here.

In case the user chooses "No", the dialog will re-appear every 5 minutes, until the play duration becomes longer than the original autosave (or another longer-duration savegame is loaded).

Do you think I should disable autosave until loading or restarting?

Or have another button (this can be confusing)?

@SupSuper claimed that this feature can be more annoying than useful, and I see his point.

Other opinions before I reject and close it?

PR #3380, opened 2021-09-18
WINTERMUTE: Replace convert_utf with built-in conversions

SupSuper:
Addresses https://bugs.scummvm.org/ticket/12926
Wintermute is already using U32String so we can just use our Unicode rather than third-party code.


Die4Ever (approved):

bluegr:
Good move! Yes, now that we got the common UTF and the associated string conversion code, we no longer need the engine-specific code in Wintermute. Thanks! Merging

lotharsm:
Can we backport this to `branch-2-3`, so that the next release will be available in Debian with all engines enabled?

PR #3381, opened 2021-09-19
BASE: detection as a dynamic plugin

Die4Ever:
proof of concept to fix https://bugs.scummvm.org/ticket/12345

### DIFF
--- a/base/main.cpp
+++ b/base/main.cpp
@@ -314,6 +314,11 @@ static Common::Error runGame(const Plugin *plugin, const Plugin *enginePlugin, O
 	system.getEventManager()->purgeKeyboardEvents();
 	system.getEventManager()->purgeMouseEvents();
 
+	#if defined(UNCACHED_PLUGINS) && defined(DYNAMIC_MODULES) && !defined(DETECTION_STATIC)
+	PluginManager::instance().unloadDetectionPlugin();
+	_plugin = NULL;
+	#endif
+
 	// Run the engine
 	Common::Error result = engine->run();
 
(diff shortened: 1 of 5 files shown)


Die4Ever:
@lephilousophe check it out
I know it has a memory leak, and it doesn't properly determine the size of the game descriptor... but it does work lol

engines/advancedDetector.cpp
```diff
@@ -392,6 +400,16 @@ Common::Error AdvancedMetaEngineDetection::createInstance(OSystem *syst, Engine
 		return Common::kNoGameDataFoundError;
 
 	DetectedGame gameDescriptor = toDetectedGame(agdDesc);
+	ADGameDescription *desc = (ADGameDescription *)new byte[sizeof(ADGameDescription)+128];
```
Die4Ever, line 403: since we know the length of the strings inside the struct, we could allocate it all in a single allocation, but if there's any pointers in the subclass then that would be a problem

base/main.cpp
```diff
@@ -314,6 +314,11 @@ static Common::Error runGame(const Plugin *plugin, const Plugin *enginePlugin, O
 	system.getEventManager()->purgeKeyboardEvents();
 	system.getEventManager()->purgeMouseEvents();
 
+	#if defined(UNCACHED_PLUGINS) && defined(DYNAMIC_MODULES) && !defined(DETECTION_STATIC)
+	PluginManager::instance().unloadDetectionPlugin();
```
Die4Ever, line 318: this could probably be moved higher in the function, hopefully before the engine is allocated, as long as you copy the data that's needed first

lephilousophe: You can't call unloadDetectionPlugin like this as the Engine may require its MetaEngine to work.

PR #3384, opened 2021-09-20
Clean translations charsets

criezy:
In the past we used to require 8-bit encodings for translations (typically iso-8859-#), with corresponding localised bds fonts in the themes to support those translations.

When we switched the translations encoding to Unicode, some leftovers were left behind, most notably all the localised bds fonts that were no longer used, but still bloated the themes.

This pull request:
 - Removes the `TranslationManager::getCurrentCharset` function (since it is now utf-32, and was used only to support loading the localised iso-8859-* bds fonts).
 - Removes the *-iso-8859-*.bdf and *-iso-8859-*.fcc files.
 - Cleans various code related to those.

The main benefit is that the theme files are noticeably smaller. But this is simplifies a bit the `ThemeEngine` font-related code.

I tried to do the changes step by step in a logical order so that each commit is easy to understand. This should also help to find the issue if this introduces regressions.

I am not completely confident with all the changes, so I would appreciate another pair of eyes looking at this before this gets merged.


ccawley2011 (approved):

criezy:
I will merge this now before I get another conflict in the themes 😛

PR #3387, opened 2021-09-25
RFC: (Don't merge without discussion) Disable fullscreen/windowed toggle on 3D games on release branch

eriktorbjorn:
Since we have at least two reports about how toggling between fullscreen and windowed mode is completely broken in 3D games...

https://bugs.scummvm.org/ticket/12473
https://bugs.scummvm.org/ticket/12476

...and since no one has proposed a fix in the past five months, perhaps we should disable fullscreen/windowed switching in the release branch. It's beyond useless in its current state.

This seems like the cleanest way of doing it, if I understand things correctly. You can still switch fullscreen/windowed mode if you're using a software renderer.

### DIFF
--- a/backends/graphics/sdl/sdl-graphics.cpp
+++ b/backends/graphics/sdl/sdl-graphics.cpp
@@ -398,6 +398,9 @@ bool SdlGraphicsManager::notifyEvent(const Common::Event &event) {
 void SdlGraphicsManager::toggleFullScreen() {
 	if (!g_system->hasFeature(OSystem::kFeatureFullscreenMode) ||
 	   (!g_system->hasFeature(OSystem::kFeatureFullscreenToggleKeepsContext) && g_system->hasFeature(OSystem::kFeatureOpenGLForGame))) {
+#ifdef USE_OSD
+		displayMessageOnOSD(_("Switching between fullscreen and windwed mode is not supported"));
+#endif
 		return;
 	}
 
--- a/backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
+++ b/backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
@@ -81,7 +81,7 @@ bool OpenGLSdlGraphics3dManager::hasFeature(OSystem::Feature f) const {
 		(f == OSystem::kFeatureFullscreenMode) ||
 		(f == OSystem::kFeatureOpenGLForGame) ||
 #if SDL_VERSION_ATLEAST(2, 0, 0)
-		(f == OSystem::kFeatureFullscreenToggleKeepsContext) ||
+//		(f == OSystem::kFeatureFullscreenToggleKeepsContext) ||
 #endif
 		(f == OSystem::kFeatureVSync) ||
 		(f == OSystem::kFeatureAspectRatioCorrection) ||


sev-:
I agree, this is an acceptable workaround for the time being.  We should document it somewhere, though.

macca8:
I'm in favour of the proposed workaround, however, the potential exists for users to post bug reports for games where the shortcut doesn't do anything. 
As a suggestion, would it be appropriate to perhaps temporarily add a timed On-Screen-Display - e.g. “Toggling fullscreen is not currently available for this game engine.” -  that supports this workaround, until the underlying issue can be properly fixed?

eriktorbjorn:
I added a displayMessageOnOSD() call when fullscreen toggling isn't supported. Unfortunately, it seems that the 3D backend doesn't support OSD messages, so that didn't do any good here...

eriktorbjorn:
In all fairness, I should add that aquadran said that "fullscreen switching works pretty well here in SDL2", with Windows, Mac and Linux.

But that does not hold true for me, so it would be interesting to hear how it works for others. For me it only works with software rendering. OpenGL turns the screen mostly white (if there are models they are untextured), and OpenGL with shaders turns the screen all black. I've tested this with the Grim, Myst 3, Stark and Wintermute engines.

criezy:
> it would be interesting to hear how it works for others

For me on macOS with The Longest Journey and OpenGL with Shaders, switching from fullscreen to windowed with Alt+Enter caused ScummVM to crash with an `EXC_BAD_ACCESS (code=1, address=0x0)` error in `GLEngine gleRunVertexSubmitImmediate`.

Trying to switch from windowed to fullscreen was a lot worse with a black screen and the mac completely unresponsive. I had to hard-reboot it. This might have been bad luck, but I don't want to try it again...

So I am in favour of disabling that for the release if it cannot be fixed in the next few days.

raziel-:
On AmigaOS4:

Myst III
I get a white window on an Alt+Enter switch, but that can be fixed by entering a new room (or take a step so that it loads anew).
Switching back to fullscreen makes the screen stay white (also fixed with the above method.

Grim
White window aswell on a switch, but Manny's figure is still visible (albeit without any textures).
Stepping out of his office, makes the background and surrounding textures load in, but Manny stays bland

Stark doesn't work here due to missing shader support
Wintermute i don't know which games are 3D

raziel-:
I forgot to mention
SDL2, OpenGL without shaders here

macca8:
My experience on macOS 10.11.6 with Myst III is described above in ticket 12476.

But to expand on the detail, when using OpenGL with shaders and switching from fullscreen to windowed, I'm getting the same detail in the crash report for Myst III as criezy does for Longest Journey (not worth the risk to try windowed to fullscreen).

With OpenGL without shaders, I was able to (eventually) reproduce raziel's recovery for Myst III, but only after a lot of blind clicking and dragging of the mouse, and subject to the screen image having changed to another area. However, the game cursor is now an empty square, and the entries in the Myst III save/load dialogs are blank, so an exit is still required to properly reset the game.

eriktorbjorn:
> Wintermute i don't know which games are 3D

I don't know much about the Wintermute engine, but if I try J.U.L.I.A. (either the original, or the re-release) with a software renderer it will fail with a "3D software rendered is not supported yet!" message. So that's what I used for a test case.

macca8:
Just curious, but why is the focus on the relationship between fullscreen switching and SDL2?
Alt+Enter worked perfectly well on Myst III with daily builds built with SDL2, before Refactor OpenGLSdlGraphics3dManager was committed.

raziel-:
@macca8 

> With OpenGL without shaders, I was able to (eventually) reproduce raziel's recovery for Myst III.
However, the game cursor is now an empty square, and the entries in the Myst III save/load dialogs are blank, so an exit is still required to properly reset the game.

https://bugs.scummvm.org/ticket/11895
Take a look at my bug report from years ago, which describes the exact same behaviour regarding the save/load entries.
This is a genuine bug and seems to have nothing to do with the latest regression (but I'm awfully glad someone else than just me was able to finally reproduce).

Seems like no one tried Myst III on OpenGL w/o shaders yet.

raziel-:
> Just curious, but why is the focus on the relationship between fullscreen switching and SDL2?

At least for me this bug
https://bugs.scummvm.org/ticket/12426
prevents me from giving feedback for SDL1.

And now it not only affects GRIM/EMI, but MystIII as well.

I'm assuming it's a general regression in the SDL1 code.

raziel-:
@eriktorbjorn 

Thanks.
Same behaviour with J.U L.I.A.
Getting a white rectangle instead of the game screen which is not recoverable.

See also here why my tests may be flawed for WME (still applies)
https://bugs.scummvm.org/ticket/11160

eriktorbjorn:
> I'm assuming it's a general regression in the SDL1 code.

I'm not really familiar with the ins and outs of the graphics backends but from what I understand, for full screen toggling to work the backend has to support fullscreen mode (obviously) and for 3D games they also have to declare that fullscreen toggling "keeps context". See SdlGraphicsManager::toggleFullScreen().

The SDL graphics managers (that I've looked at) that declare the "keeps context" feature at all only do so for SDL 2.0.0 and higher. So presumably SDL1 doesn't allow fullscreen toggling ever? (Unless using OpenGL for 2D games makes a difference?)

This pull request simply makes OpenGLSdlGraphics3dManager never declare that it supports the "keeps context" feature. That seemed like the cleanest way of disabling any feature that requires it.

aquadran:
I do not agree on this. That worked before. This should be fixed or change reverted.

eriktorbjorn:
> I do not agree on this. That worked before. This should be fixed or change reverted.

I'm not disagreeing, but I don't know how to fix it properly, no one has suggested how in the five months since the bug was reported, and https://github.com/scummvm/scummvm/commit/d33487f64 may not be easily revertable at this point.

So this is the cheap way out, _if_ no proper fix can be made in time for the release.

Another fullscreen regression since ResidualVM is https://bugs.scummvm.org/ticket/11894 / https://bugs.scummvm.org/ticket/12475 but I haven't checked when that was introduced. (And I didn't know until yesterday that it uses the aspect ratio correction setting instead, apparently?)

macca8:
@raziel- 

> https://bugs.scummvm.org/ticket/11895
Take a look at my bug report from years ago, which describes the exact same behaviour regarding the save/load entries.
This is a genuine bug and seems to have nothing to do with the latest regression

Unfortunately, I have to disagee with you... this behaviour only exists now because of the latest regression.
Testing daily builds from 17 April 2021 (prior to the commit I mentioned being added), Alt+Enter works as expected in Myst III using 3D OpenGL (with or without shaders), and consequently, there's no issues with the save/load entries.
Daily builds from 18 April 2021 (that include the commit) tested under the same conditions introduce all the Myst III issues reported here, including the save/load entries (but only until restarted).

Perhaps your bug had been fixed in the meantime, and this commit reversed that fix. I'll post a couple of images of the save screens in your bug report so that you can see if it's indeed similar behaviour.

> Seems like no one tried Myst III on OpenGL w/o shaders yet.

Completed the game using normal 3D OpenGL, and random tested using OpenGL with shaders.

> At least for me this bug
https://bugs.scummvm.org/ticket/12426
prevents me from giving feedback for SDL1.
I'm assuming it's a general regression in the SDL1 code.

Correct. Alt+Enter did nothing in Myst III with previous SDL 1.2.15 daily builds, but I considered it outdated, and not an issue.

raziel-:
@macca8

Updated the Myst item.

...and would it be ok to close the other report then?

bluegr:
I checked the differences in commit d33487f64, which introduced the crash/black screen when switching 3D games to/from full screen

The difference I found is that in the removed SdlGraphics3dManager, the call to setFeatureState() was not enclosed between beginGFXTransaction() and endGFXTransaction().

Removing the calls to beginGFXTransaction() and endGFXTransaction() fixes switching to and from full screen for me in Myst 3. That's in sdl-graphics.cpp:398, SdlGraphicsManager::toggleFullScreen()

beginGFXTransaction() sets _transactionMode to kTransactionActive. But then OpenGLSdlGraphics3dManager::setFeatureState() checks if _transactionMode == kTransactionNone in order to call createOrUpdateScreen().

bluegr:
Changing `SdlGraphicsManager::toggleFullScreen()` fixes the issue for me and retains the previous behavior of  `SdlGraphics3dManager::toggleFullScreen()`:

    void SdlGraphicsManager::toggleFullScreen() {
	    if (!g_system->hasFeature(OSystem::kFeatureFullscreenMode) ||
	       (!g_system->hasFeature(OSystem::kFeatureFullscreenToggleKeepsContext) && g_system->hasFeature(OSystem::kFeatureOpenGLForGame))) {
		    return;
	    }
    
	    bool is3D = g_system->hasFeature(OSystem::kFeatureOpenGLForGame);
    
	    if (!is3D)
		    beginGFXTransaction();
	    setFeatureState(OSystem::kFeatureFullscreenMode, !getFeatureState(OSystem::kFeatureFullscreenMode));
	    if (!is3D)
		    endGFXTransaction();
    #ifdef USE_OSD
	    if (getFeatureState(OSystem::kFeatureFullscreenMode))
		    displayMessageOnOSD(_("Fullscreen mode"));
	    else
		    displayMessageOnOSD(_("Windowed mode"));
    #endif
    }

bluegr:
Opened PR #3392 with the aforementioned change

bluegr:
Closing, as PR #3392 has been merged

PR #3392, opened 2021-09-28
SDL: Fix full screen toggling for 3D games

bluegr:
The previous SdlGraphics3dManager::toggleFullScreen() was not using
beginGFXTransaction / endGFXTransaction in toggleFullScreen().
Restore this logic for 3D games, as using transactions in such games
results in a black screen for 3D games when switching to full screen

A regression from d33487f64

Alternate fix for the situation reported in PR #3387

### DIFF
--- a/backends/graphics/sdl/sdl-graphics.cpp
+++ b/backends/graphics/sdl/sdl-graphics.cpp
@@ -401,9 +401,13 @@ void SdlGraphicsManager::toggleFullScreen() {
 		return;
 	}
 
-	beginGFXTransaction();
+	bool is3D = g_system->hasFeature(OSystem::kFeatureOpenGLForGame);
+
+	if (!is3D)
+		beginGFXTransaction();
 	setFeatureState(OSystem::kFeatureFullscreenMode, !getFeatureState(OSystem::kFeatureFullscreenMode));
-	endGFXTransaction();
+	if (!is3D)
+		endGFXTransaction();
 #ifdef USE_OSD
 	if (getFeatureState(OSystem::kFeatureFullscreenMode))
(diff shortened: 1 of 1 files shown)


bluegr:
Tried this with Grim and Myst 3, and switching to and from full screen works as expected under Windows, using OpenGL with shaders

eriktorbjorn:
Seems to work with Grim, Myst 3 and Stark under Linux. Grim and Stark don't care about the "stretch mode" setting, but it was pointed out to me that this can be worked around with the aspect ratio correction setting. A bit inconsistent, but workable.

raziel-:
@blugr

AmigaOS4 - SDL2 - OpenGL without shaders (btw, are shaders supported in OpenGL 1.3ish?)

It works in Grim again
But i still get a black window (not-recoverable, unless i switch back to fullscreen) in Myst III.

Wintermute i still have to test

aquadran:
> @blugr
> 
> AmigaOS4 - SDL2 - OpenGL without shaders (btw, are shaders supported in OpenGL 1.3ish?)
> 
> It works in Grim again But i still get a black window (not-recoverable, unless i switch back to fullscreen) in Myst III.
> 
> Wintermute i still have to test

aquadran:
> @blugr
> 
> AmigaOS4 - SDL2 - OpenGL without shaders (btw, are shaders supported in OpenGL 1.3ish?)
> 
> It works in Grim again But i still get a black window (not-recoverable, unless i switch back to fullscreen) in Myst III.
> 
> Wintermute i still have to test

Grim should not be affected at all. It reload game state at fullscreen switch. So no idea.

aquadran:
tested with SDL2 on windows, linux,macos. with and without shaders. works fine. merging

PR #3393, opened 2021-09-29
NEVERHOOD: Possible fix for bug #12930 (crash on exit)

eriktorbjorn:
This is the fix I suggested on https://bugs.scummvm.org/ticket/12930

I don't know if it's correct - I never got any feedback on it - but it does seem to fix the game crashing on exit for me. If I'm right, it was trying to keep playing a sound after the audio data for it had been freed. With the change, sound 3 is handled the same way as sounds 1 and 2, i.e. stopped first.


bluegr:
This is indeed a correct fix, as sounds in this category were not properly cleared. Your implementation follows the handling done for the other two. Thanks for spotting and fixing this! Merging

PR #3394, opened 2021-09-30
STARK: Adding mechanism to correct subtitles

Thunderforge:
Fixes [#12967](https://bugs.scummvm.org/ticket/12967)

If there are any additional subtitles that don't match the spoken dialogue, we can correct them too.

Corrected subtitle

### DIFF
--- a/engines/stark/resources/speech.cpp
+++ b/engines/stark/resources/speech.cpp
@@ -149,8 +149,16 @@ int32 Speech::getPauseAfterSpeechDuration() const {
 void Speech::readData(Formats::XRCReadStream *stream) {
 	Object::readData(stream);
 
-	_phrase = stream->readString();
+	_phrase = correctSubtitle(stream->readString());
 	_character = stream->readSint32LE();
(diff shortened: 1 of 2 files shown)


engines/stark/resources/speech.cpp
```diff
@@ -149,10 +149,18 @@ int32 Speech::getPauseAfterSpeechDuration() const {
 void Speech::readData(Formats::XRCReadStream *stream) {
 	Object::readData(stream);
 
-	_phrase = stream->readString();
+	_phrase = correctSubtitle(stream->readString());
 	_character = stream->readSint32LE();
 }
 
+Common::String Speech::correctSubtitle(Common::String phrase) {
```
mgerhardy, line 156: maybe do

```c++
_phrase = stream->readString();
correctSubtitle(_phrase);
```

then you don't have to copy the strings.

Thunderforge:
Thanks @bluegr for finding a better solution than mine!

PR #3398, opened 2021-10-01
STARK: Fix bad subtitle for Cortez in theater

antoniou79:
Fix for bug https://bugs.scummvm.org/ticket/12967

### DIFF
--- a/engines/stark/resources/speech.cpp
+++ b/engines/stark/resources/speech.cpp
@@ -25,6 +25,7 @@
 #include "engines/stark/formats/xrc.h"
 
 #include "engines/stark/services/services.h"
+#include "engines/stark/services/settings.h"
 #include "engines/stark/services/dialogplayer.h"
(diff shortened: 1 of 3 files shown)


bluegr:
I'm confused. Isn't that a similar fix with the one done in #3394?

antoniou79:
> I'm confused. Isn't that a similar fix with the one done in #3394?

Sorry, I did not notice there was an existing PR.

Just to defend my approach. 

It addresses and fixes the same issue but I wanted to make the check for the change as specific as possible, so I filter by game language, character who speaks the line, and then the index (int), subType (int) and finally specific resource name ("Cortez_Laying low #1"), in order to avoid checking every loaded string against the faulty one.

aquadran:
I agree with @antoniou79, this has better checks

Thunderforge:
I agree, this is a better fix than the one I was doing. Thanks!

PR #3400, opened 2021-10-02
Fix restoring window size when exiting fullscreen

criezy:
There were two different issues:
 - On macOS the window was always restored to the maximum window size and not to the previous window size.
 - The window was restored to the default size and not to the previous window size.

The first issue might be specific to macOS, for which fullscreen more or less means maximized, and in particular when requesting a `SDL_WINDOW_FULLSCREEN_DESKTOP` window, the `SDL_WINDOW_MAXIMIZED` flag gets also set. As a result, when we exit fullscreen, `loadVideoMode()` gets called while the window is still fullscreen, and checking the window flag indicate it is maximized. As a result it tries to restore the maximized size. Since we store in the config if the window was maximized, instead of looking at the SDL window flag we can use the flag we stored in the config.

That issue only occurred if maximized size was stored in the config file (i.e. if the window had been maximized by the user at least once in the past).

The second issue was more general. In `loadVideoMode()` when switching to fullscreen, it was storing the default size in the config, overwriting the current window size. So when exiting fullscreen, the size was restored to the default size (since this is what was stored in the config). I don't think this was intended, and at least it seems to me that restoring to the size the window had before entering fullscreen would be what is expected.

As the SDL behaviour may be different on other systems, it would be good to test that this works as expected on other systems (e.g. Windows and Linux) before merging this change.

Note: On macOS setting the `SDL_WINDOW_MAXIMIZED` does not quite work properly. The window is resized to fill the screen, but with the menubar and title bar still visible. That means that if we maximize the window (maximize button in window title bar), then go fullscreen (Alt+Enter) and then exit fullscreen (Alt+Enter), we get an almost-maximised window but not exactly what we had before entering fullscreen.

### DIFF
--- a/backends/graphics/openglsdl/openglsdl-graphics.cpp
+++ b/backends/graphics/openglsdl/openglsdl-graphics.cpp
@@ -349,27 +349,29 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	bool isMaximized = ConfMan.getBool("window_maximized", Common::ConfigManager::kApplicationDomain);
-	if (isMaximized && !_wantsFullScreen && ConfMan.hasKey("window_maximized_width", Common::ConfigManager::kApplicationDomain) && ConfMan.hasKey("window_maximized_height", Common::ConfigManager::kApplicationDomain)) {
-		// Set the window size to the values stored when the window was maximized
-		// for the last time.
-		requestedWidth  = ConfMan.getInt("window_maximized_width", Common::ConfigManager::kApplicationDomain);
-		requestedHeight = ConfMan.getInt("window_maximized_height", Common::ConfigManager::kApplicationDomain);
+	if (!_wantsFullScreen) {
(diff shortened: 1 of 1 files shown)


backends/graphics/openglsdl/openglsdl-graphics.cpp
```diff
@@ -348,29 +348,30 @@ bool OpenGLSdlGraphicsManager::loadVideoMode(uint requestedWidth, uint requested
 	Common::Rect desktopRes = _window->getDesktopResolution();
 
 #if SDL_VERSION_ATLEAST(2, 0, 0)
-	SDL_Window *window = _window->getSDLWindow();
-	bool isMaximized = window ? (SDL_GetWindowFlags(window) & SDL_WINDOW_MAXIMIZED) : false;
```
lotharsm, line 352: I'm pretty sure we have to keep the `isMaximized` check intact, since this is not covered by `_wantsFullScreen`. At least on Win32 and Linux, fullscreen and maximized are not the same like on macOS, but two different modes.

lotharsm: Now that I think of it, maybe it's better to fix what causes my workaround to be necessary in the first place 😂

bluegr:
Regarding the second issue, I can confirm that with the changes in this PR, the original window's dimensions are preserved correctly after switching from full screen under Windows 10.

PR #3401, opened 2021-10-03
DEVTOOLS: simplify macbinary packing a bit

einstein95:


bluegr:
So, you're essentially changing `xB63s4` to `x64p4`. But why have you removed the length?

PR #3403, opened 2021-10-04
KYRA: HOF: Fix hebrew multiword object action strings

BLooperZ:
Following #3277

This PR fixes the action string for multi-worded object in Hebrew fan translation of Kyrandia 2
as well as reversing the order of the sentence.

The game resource have objects names along with the indefinte article,
then, on command string, the engine removes the article and uppercase the first letter, e.g.
an empty flask -> Empty flask taken.

This logic caused object with multiple words to skip the first word in hebrew.

The reversing of the sentences is to avoid having to consider the object gender,
so the sentence become dependent on the player and not the object
(you (plural, indefinite) have dropped [object].) instead of ([object] placed (gender needs to match).)

Thanks


athrxx:
Thanks!

PR #3405, opened 2021-10-05
CREATE_PROJECT: Msvc dynamic modules

henke37:
This adds support for dynamic modules to `CREATE_PROJECT`, and specifically the `msbuild` target.

Right now there are a few things left to do:
* [ ] Dynamically generate the def file
* [ ] Solve the linking for 3d engines
* [ ] Solve `Common::Singleton` being used outside of `Common` linking issues
* [x] Setup correct output locations for the plugin dlls
* [ ] Move files embedded as resources into the plugin dlls
* [ ] Check that other `create_project` outputs work
* [ ] Implement the `SCUMMVM_EXPORT` define in non `create_project` builds
* [ ] Check that everything works


bluegr:
I agree with @sev-, the changes are too invasive. I'm all up for supporting dynamic modules in MSVC builds, but the changes needed should not be that invasive and affect existing code to such extend

henke37:
The core issue here is telling the compiler & linker to use stuff from the main executable. All code references can be done in the `.def` file, but data requires a declspec in the code, doing it in the `.def` file is not enough.

henke37:
Bluegr pretty much nailed the benefit: It makes it possible to debug the plugin code for MSVC users. And when I get to it, other platforms that also use `create_project`. Those are included in my plan for this PR.

I agree that there should be no major changes to engine code here. Or base/common for that matter.

As for the code that breaks when moved, that I have not seen in my WIP builds. I will of course investigate this. If I break something, then I fix it.

However, I do not understand which changes qualify as "invasive". Are you talking about the need for the `SCUMMVM_EXPORT` define? While regrettable, they are few and tiny. A single "keyword" in less than twenty spots is not too many.

As for the changes to game engine folders, those are indeed worth scrutiny. They are essentially a revert of previous work on the plugin system. I am not an expert in make, so this is an area that I'm intending to examine in close detail.

As for testing, yes this will need careful testing. I'm not rushing this and I don't want to find out "later" that stuff broke.

sev-:
No users need to debug plugins under MSVC. Do you have another, more viable reason for these extremely invasive changes?

henke37:
Again, I don't see how these changes are invasive. But I don't have another reason.

sev-:
45 files changes, _everything_ except Windows is broken. If this is not invasive then what is?

henke37:
Touch one spot and everything breaks. Touch another and it all works again.

As for the number of files, I'd rather measure how much each file is changed and how controversial the change is.

bluegr:
> Touch one spot and everything breaks. Touch another and it all works again.

I really don't understand how this answers @sev-'s question: this PR breaks all ports apart from Windows. The ports should be fixed before these changes are even considered for merging

henke37:
Of course that needs to be done. It's even in the checklist. This is not even close to finished.

PR #3408, opened 2021-10-09
DIRECTOR: Handle 0 byte Fmap in Cast::loadFontMapV4

einstein95:
Stephen Jay Gould demo has a 0 byte Fmap which makes ScummVM crash. This is basically a copy/paste from Cast::loadFont.

[Demo link](https://cdn.discordapp.com/attachments/636495083935629325/896308869599461427/Stephen_Jay_Gould.7z)


rvanlaar:
This change results in a couple of broken games. 
The main error seems to be: `scummvm: common/stream.cpp:266: virtual bool Common::SeekableSubReadStream::seek(int64, int): Assertion `_pos <= _end' failed.`

Have a look at: https://john.scummvm.org/#/changes/24365

PR #3417, opened 2021-10-13
SCUMM: Text improvements for Mac Loom and Indy 3 (bugs #12983 and #12984)

eriktorbjorn:
This began as an attempt, albeit not yet quite successful, at addressing the text centering issues described in https://bugs.scummvm.org/ticket/12984 and then expanded to cover some of the issues in https://bugs.scummvm.org/ticket/12983 as well.

The theory is that the main problem is that ScummVM currently measures the width of a string by dividing the width of every character in the string by two (to get the low-resolution width), and taking the sum of that. With this change, it adds the high-resolution width of all the character widths, and then divides that sum by two.

To do this, I had to make getStringWidth() a virtual function so that each character renderer define their own. I don't know if this is the best way to do it, since that means each character renderer has to handle escape sequences in the strings if they want to define their own. It appears Indy 3 doesn't handle any escape sequences at all, and Loom only needs to handle line breaks.

Unfortunately, while the result seems to be an improvement in general, there are still places where it's not pixel perfect and I have no idea why. I'm attaching a series of screenshots to demonstrate, and will add a comment about the observed results. (I recommend opening the pictures in separate browser tabs to make it easier to compare them.)

For Indy 3, there is now emulation for the (in my opinion) slightly broken font spacing used in some cases, with an option to do it correctly instead.

### DIFF
--- a/engines/scumm/script_v5.cpp
+++ b/engines/scumm/script_v5.cpp
@@ -2766,6 +2766,7 @@ int ScummEngine_v5::getWordVararg(int *ptr) {
 
 void ScummEngine_v5::decodeParseString() {
 	int textSlot;
+	int color;
 
 	switch (_actorToPrintStrFor) {
 	case 252:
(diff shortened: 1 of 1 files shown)


eriktorbjorn:
Screenshot from Mac emulator. The original has incorrect spacing between letters, which makes it hard to say what's right or wrong:
original-00006

Screenshot from unmodified ScummVM.
scummvm-00006

Screenshot from ScummVM after this change. The leftmost character of each line appears in the same position as in the original, which could be seen as an improvement over unmodified ScummVM.
scummvm-new-00006

eriktorbjorn:
Screenshot from Mac emulator. The original has incorrect spacing between letters, which makes it hard to say what's right or wrong.
original-00007

Screenshot from unmodified ScummVM. Only "1938" appears in the same position as the original.
scummvm-00007

Screenshot from ScummVM after this change. The leftmost character of each line appears in the same position as in the original, which could be seen as an improvement over unmodified ScummVM.
scummvm-new-00007

eriktorbjorn:
Screenshot from Mac emulator. The original has incorrect spacing between letters, which makes it hard to say what's right or wrong.
original-00008

Screenshot from unmodified ScummVM. The leftmost characters all appear in the same position as in the original.
scummvm-00008

Screenshot from ScummVM after this change. Nothing has changed compared to before the change.
scummvm-new-00008

bluegr:
This is certainly an improvement, even if overall the differences are small. Nice work overall, +1 from me :)

PR #3420, opened 2021-10-14
GUI: Don't apply values from disabled subtitle toggle

sluicebox:
Fixes bug #13007 where selecting "Override global audio settings" always mutes the (few) speech clips in Hoyle4.

The GUI's subtitle toggle controls are disabled in a game's options when GUIO_NOSUBTITLES or GUIO_NOSPEECH set. But the GUI reads from the subtitle toggle even if it's disabled and writes config values based on it. This means that selecting "Override global audio settings" has side effects that the user can't see or control, such as setting "mute_speech" and muting all kSpeechSoundType. That usually doesn't matter, given the games that disable the controls in the first place, but some games like Hoyle4 are GUIO_NOSPEECH but happen to use speech functions for a small number of things. (Hoyle4 is a floppy disk game and a few short message texts have accompanying speech audio like "I win!" or the rank and suit of a card.)

I made a separate member variable for tracking this state instead of querying the control's enabled status directly because _subToggleGroup is a RadiobuttonGroup and that's not a Widget so it doesn't have isEnabled(). This seemed cleaner than querying one of the subcontrols.

Another response is that if Hoyle4 is using any speech functionality then we should remove GUIO_NOSPEECH so that the toggle is enabled and the user can control it. That may be something we also do on the SCI side, though I don't think it makes much sense for this game, but the current toggle behavior doesn't seem intentional or expected so I'm starting here.

### DIFF
--- a/gui/options.cpp
+++ b/gui/options.cpp
@@ -231,6 +231,7 @@ void OptionsDialog::init() {
 	_speechVolumeLabel = nullptr;
 	_muteCheckbox = nullptr;
 	_enableSubtitleSettings = false;
+	_enableSubtitleToggle = false;
 	_subToggleDesc = nullptr;
 	_subToggleGroup = nullptr;
 	_subToggleSubOnly = nullptr;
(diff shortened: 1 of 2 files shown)


orgads:
This breaks SCUMM games with no speech when the global setting is Speech only. https://bugs.scummvm.org/ticket/13151

scumm.cpp has this:
```c++
	if (ConfMan.getBool("speech_mute") && !ConfMan.getBool("subtitles"))
		ConfMan.setBool("subtitles", true);
```

But with this change, none of them is set, so subtitles are not enabled.

PR #3422, opened 2021-10-17
AMIGAOS: Prepare automated OGLES2 compilation

raziel-:
OGLES2 support has been added by using ptitSeb´s gl4es wrapper.

To automate the process without the least amount of hassle use explicitly named gl4es libraries.

### DIFF
--- a/configure
+++ b/configure
@@ -2631,7 +2631,7 @@ case $_host_os in
 		_builtin_resources=no
 		# OGLES2 needs explicitly named gl4es libs
 		if test "$_opengl_mode" = gles2; then
-			append_var LIBS "-lSDL2_gl4es -lglu_gl4es -lgl4es"
+			append_var LIBS "-lSDL2_gl4es -lgl4es"
 			_opengl_game_es2=yes
 		fi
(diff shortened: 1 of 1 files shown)


lephilousophe:
Looking at configure code, when using OpenGL we need glew for shaders to be enabled.
AmigaOS4 doesn't have glew so this won't work.
This may change if #3463 gets merged.

raziel-:
Oh...
Well, whatever GLAD is, i fear i don't have that either

lephilousophe:
GLAD is a generator of wrappers. The load function used in the PR is the function provided by SDL.
As AmigaOS4 platform uses SDL as backend it should work fine.

raziel-:
@lephilousophe 

Ok, here's hoping.

Should I keep this PR open until the other one gets merged?

Thank you once again for the help

lephilousophe:
I would say that if changes are needed after GLAD gets merged, they will be different from the one you currently have.
So it's OK to close and re-open later.

raziel-:
@lephilousophe 

Thank you...prepare to be annoyed later ;-)

PR #3428, opened 2021-10-19
STARK: Fix moving autosave to a new slot

orgads:


orgads:
Squashed into #3429. Closing this.

PR #3430, opened 2021-10-20
ENGINES: Fix getSavegameFile for almost all engines

orgads:
Use kSimpleSavesNames correctly, add where needed, remove where needed.

[Trac #12977](https://bugs.scummvm.org/ticket/12977)

### DIFF
--- a/engines/ngi/metaengine.cpp
+++ b/engines/ngi/metaengine.cpp
@@ -69,6 +69,13 @@ public:
 	SaveStateDescriptor querySaveMetaInfos(const char *target, int slot) const override;
 
 	Common::Error createInstance(OSystem *syst, Engine **engine, const ADGameDescription *desc) const override;
+	Common::String getSavegameFile(int saveGameIdx, const char *target) const override {
+		const Common::String prefix("fullpipe");
+		if (saveGameIdx == kSavegameFilePattern)
+			return prefix + ".s##";
+		else
+			return prefix + Common::String::format(".s%02d", saveGameIdx);
+	}
 };
 
 bool NGIMetaEngine::hasFeature(MetaEngineFeature f) const {
@@ -79,8 +86,7 @@ bool NGIMetaEngine::hasFeature(MetaEngineFeature f) const {
 		(f == kSavesSupportThumbnail) ||
 		(f == kSavesSupportCreationDate) ||
 		(f == kSavesSupportPlayTime) ||
-		(f == kSupportsLoadingDuringStartup) ||
-		(f == kSimpleSavesNames);
+		(f == kSupportsLoadingDuringStartup);
 }
 
 bool NGI::NGIEngine::hasFeature(EngineFeature f) const {
@@ -93,9 +99,8 @@ bool NGI::NGIEngine::hasFeature(EngineFeature f) const {
 SaveStateList NGIMetaEngine::listSaves(const char *target) const {
 	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
 	Common::StringArray filenames;
-	Common::String pattern("fullpipe.s##");
 
-	filenames = saveFileMan->listSavefiles(pattern);
+	filenames = saveFileMan->listSavefiles(getSavegameFilePattern(target));
 
 	SaveStateList saveList;
 	for (Common::StringArray::const_iterator file = filenames.begin(); file != filenames.end(); ++file) {
--- a/engines/pegasus/metaengine.cpp
+++ b/engines/pegasus/metaengine.cpp
@@ -78,6 +78,18 @@ public:
 	int getMaximumSaveSlot() const override { return 999; }
 	void removeSaveState(const char *target, int slot) const override;
 	Common::KeymapArray initKeymaps(const char *target) const override;
+	Common::String getSavegameFile(int saveGameIdx, const char *target) const override {
+		if (saveGameIdx == kSavegameFilePattern)
+			return Common::String::format("pegasus-*.sav");
+		Common::StringArray fileNames = Pegasus::PegasusEngine::listSaveFiles();
+		if (saveGameIdx < fileNames.size())
+			return fileNames[saveGameIdx];
+		if (fileNames.empty())
+			return Common::String("pegasus-1.sav");
+		Common::String name = fileNames.back();
+		name.insertString("_last", name.size() - 4);
+		return name;
+	}
 };
 
 bool PegasusMetaEngine::hasFeature(MetaEngineFeature f) const {
--- a/engines/pegasus/pegasus.cpp
+++ b/engines/pegasus/pegasus.cpp
@@ -702,8 +702,15 @@ void PegasusEngine::writeContinueStream(Common::WriteStream *stream) {
 }
 
 Common::StringArray PegasusEngine::listSaveFiles() {
+	const Common::String autoSaveName("pegasus-AutoSave.sav");
 	Common::StringArray fileNames = g_system->getSavefileManager()->listSavefiles("pegasus-*.sav");
+	// Autosave must be at slot 0, so remove it, then prepend (even if it doesn't exist,
+	// it will be prepended)
+	Common::StringArray::iterator it = Common::find(fileNames.begin(), fileNames.end(), autoSaveName);
+	if (it != fileNames.end())
+		fileNames.erase(it);
 	Common::sort(fileNames.begin(), fileNames.end());
+	fileNames.insert_at(0, autoSaveName);
 	return fileNames;
 }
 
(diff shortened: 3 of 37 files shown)


criezy:
There was some discussion on Discord, but it is better to have it here. So let me copy the relevant part...

I think that the use of `kSavesUseExtendedFormat` to switch between `<target>.###` and `<target>.s##` is indeed dubious. But I don't think that `kSimpleSavesNames` was intended to be used that way either.

I am planning to spend more time this weekend to look at the code and the history and try to make more sense from it. But from the `kSimpleSavesNames` documentation it seems to me that it was intended as a way to indicate to the cloud savegame synchronization code that the savegame files can be identified with a simple pattern. Currently it hardcodes this pattern to `_target + ".###"` in `SaveLoadChooserDialog::listSaves()`, but it could possibly (and maybe should) use `_metaEngine->getSavegameFile(kSavegameFilePattern, _target)` instead. And the `kSimpleSavesNames` would then mean that `MetaEngine:: getSavegameFile` can be used to get a pattern to identify the savegames files (but not necessarily that the pattern is exactly `<target>.###`).

In such a case we would need something else to differentiate between `<target>.###` and `<target>.s##` in `getSavegameFile`. This could be yet another flag, or deciding that any engine that uses `<target>.s##` (or anything else other than `<target>.###`) need to reimplement getSavegameFile.

orgads:
@criezy Not sure I understand your point. If the savegame doesn't match this pattern, what's the point of setting the `kSimpleSavesNames` feature? And if it does, why not add it?

engines/ngi/metaengine.cpp
```diff
@@ -93,9 +92,8 @@ bool NGI::NGIEngine::hasFeature(EngineFeature f) const {
 SaveStateList NGIMetaEngine::listSaves(const char *target) const {
 	Common::SaveFileManager *saveFileMan = g_system->getSavefileManager();
 	Common::StringArray filenames;
-	Common::String pattern("fullpipe.s##");
 
-	filenames = saveFileMan->listSavefiles(pattern);
+	filenames = saveFileMan->listSavefiles(getSavegameFilePattern(target));
```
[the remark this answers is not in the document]
orgads: Got it. Thanks for the explanation.

sev-:
@criezy it looks like most (all?) the changes were made. Would you give it another round of review?

orgads:
@sev- he already reviewed it. I still have some changes to do.

engines/pegasus/metaengine.cpp
```diff
@@ -78,6 +78,12 @@ class PegasusMetaEngine : public AdvancedMetaEngine {
 	int getMaximumSaveSlot() const override { return 999; }
 	void removeSaveState(const char *target, int slot) const override;
 	Common::KeymapArray initKeymaps(const char *target) const override;
+	Common::String getSavegameFile(int saveGameIdx, const char *target) const override {
+		if (saveGameIdx == kSavegameFilePattern)
+			return Common::String::format("pegasus-*.sav");
+		Common::StringArray fileNames = Pegasus::PegasusEngine::listSaveFiles();
+		return fileNames[saveGameIdx];
```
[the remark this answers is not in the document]
orgads: Implemented your last suggestion.

orgads:
> My issue with it is that it is using `kSimpleSavesNames` for a different purpose than what it was introduced for. That flag was introduced with the cloud code, as that code has a special handling of some savegame files (to allow sync in background) and it needed a way to identify engines that use such savegames. Currently that means engines whose savegames match the `<target>.###` pattern, but that could change in the future. If we for example update the cloud code so that it can also handle the `<target>.s##` for background sync'ing, then we will need to add the `kSimpleSavesNames` feature to those engines as well. And with your proposed implementation that would cause `getSavegameFile()` to return an incorrect pattern/name.

I see your point, but I think that instead of reimplementing `getSavegameFile` so many times, I'd rather add *another* flag.

But since this flag will be equivalent to `kSimpleSavesNames`, I suggest adding a comment next to the existing flag (that it serves 2 purposes), and *in case* the cloud maintainer will decide to cover more cases, we can then split this flag.

What do you say?

criezy:
> But since this flag will be equivalent to `kSimpleSavesNames`, I suggest adding a comment next to the existing flag (that it serves 2 purposes), and _in case_ the cloud maintainer will decide to cover more cases, we can then split this flag.

That seems fine.
Any future changes to the cloud code are hypothetical and may never happen. So we can do that additional work if and when that happens. Flagging that with a comment is sufficient for now.

engines/pegasus/pegasus.cpp
```diff
@@ -702,9 +702,18 @@ void PegasusEngine::writeContinueStream(Common::WriteStream *stream) {
...
+	// Autosave must be at slot 0, so remove it, then prepend (even if it doesn't exist,
+	// it will be prepended)
+	Common::StringArray::iterator it = Common::find(fileNames.begin(), fileNames.end(), autoSaveName);
+	if (it != fileNames.end())
+		fileNames.erase(it);
 	Common::sort(fileNames.begin(), fileNames.end());
-	return fileNames;
+	Common::StringArray res;
+	res.push_back(autoSaveName);
+	Common::copy(fileNames.begin(), fileNames.end(), res.end());
```
criezy, line 715: I think `Common::copy` requires the destination to be already allocated? Which means here it would copy the `fileNames` element to unallocated memory, causing random issues.

You can use `res.push_back(fileNames);`, but I think the easiest solution is to insert the autosave name at the start of `fileNames` and not use another array (i.e. either use `fileNames.insert(fileNames.begin(), autoSaveName)`, or `fileNames.insert_at(0, autoSaveName)`.

engines/pegasus/metaengine.cpp
```diff
@@ -78,6 +78,14 @@ class PegasusMetaEngine : public AdvancedMetaEngine {
 	int getMaximumSaveSlot() const override { return 999; }
 	void removeSaveState(const char *target, int slot) const override;
 	Common::KeymapArray initKeymaps(const char *target) const override;
+	Common::String getSavegameFile(int saveGameIdx, const char *target) const override {
+		if (saveGameIdx == kSavegameFilePattern)
+			return Common::String::format("pegasus-*.sav");
+		Common::StringArray fileNames = Pegasus::PegasusEngine::listSaveFiles();
+		if (saveGameIdx < fileNames.size())
+			return fileNames[saveGameIdx];
+		return AdvancedMetaEngine::getSavegameFile(saveGameIdx, target);
```
criezy, line 87: For non-existent save files, this would return a name that does not follow the expected "pegasus-*.sav" pattern. Is that a good idea? Maybe it would be better to return an unused name (one that is not in the `fileNames` array), and follows the pattern? If the function is going to be used to name a save file (e.g. when we copy one from one slot to another), it could be an issue if the name does not follow the pattern as it will then not be listed the next time we call `listSaveFiles()`. Even better might be to make sure it is after the last element of `fileNames` alphabetically, but that might get complicated.

But that might be all hypothetical if the only case where we use this function with an inexistent slot is to copy a normal savegame from the autosave slot to another slot, as this should never happen with this engine (unless the user happens to create a normal savegame and name it `AutoSave`, but that would be sneaky).

engines/pegasus/pegasus.cpp
criezy: I am thinking we might need to also modify the implementation of `PegasusEngine::saveGameState()` so that if `isAutosave` is true it uses the `"pegasus-AutoSave.sav"` name instead of generating one from the save description (which might be `AutoSave` by default, but I am wondering if it might get translated for example, so not using the description for the AutoSave game and hardcoding the name there as well seems safer).

engines/tinsel/metaengine.cpp
```diff
@@ -76,6 +76,8 @@ class TinselMetaEngine : public AdvancedMetaEngine {
 	int getMaximumSaveSlot() const override;
 	SaveStateDescriptor querySaveMetaInfos(const char *target, int slot) const override;
 	void removeSaveState(const char *target, int slot) const override;
+
+	// TODO: Add getSavegameFile()
```
criezy, line 80: Why add this comment? It seems to me that the default implementation of `getSavegameFile()` should work for this engine?

criezy:
I think we need a couple of changes for the pegasus engine, but other than that it looks good and is almost ready to be merged.

engines/tinsel/metaengine.cpp
orgads: See the comments in `loadGameState` and `removeSaveState`.

engines/pegasus/pegasus.cpp
orgads: I was looking for `push_front`. Didn't consider `insert_at`, thanks for that :)

engines/pegasus/metaengine.cpp
orgads: I doubt this code path can be reached. Handled anyway :)

orgads:
All done, hopefully :)

criezy (approved):
Thank you. It look good to me now.

criezy:
I will merge this now.

But I also have a question: other than the tinsel engine, are you aware of any other engine where `getSavegameFile` might still be incorrect or as far as you know they should all be correct now? I didn't see any myself, but you probably looked at this more carefully than I did.

orgads:
myst3 also has a TODO.

PR #3431, opened 2021-10-21
VIDEO: Update the VSYNC option so it works for 2D games too.

vanfanel:
With SDL2, proper VSYNC is perfectly possible across all major OSes, with or without using OpenGL.
These changes make the VSYNC option into a general one that affects 2D using both OpenGL or SDL Surface graphics modes, and 3D games.


aquadran:
software backend is also for 3D games. if this option is handled now, it would be good to have feature flag OSystem::kFeatureVSync ported too.

vanfanel:
@aquadran I have no time to investigate that. Can you _please_ add what's missing?

PR #3432, opened 2021-10-21
3DS: Set stack size w/o running ScummVM inside a thread

BallM4788:
Currently, the 3DS port's stack size is set by running ScummVM within a thread, a behavior introduced in fb39b94384884050a7007d2752b2c2916a73b176. Changing the stack size is necessary because libctru's default stack size of 32KiB is not large enough for ScummVM.

This commit introduces a simpler method for changing the stack size. By setting a global variable ```__stacksize__```, we can override the identically-named, weak-attributed [```__stacksize__```](https://github.com/devkitPro/libctru/blob/master/libctru/source/system/stack_adjust.s#L39) defined in libctru. [This is an intentional feature](https://github.com/devkitPro/libctru/commit/192b88b6a21d83345a4635deba417c191c156c80), and renders it unnecessary to run ScummVM within a thread.

Explanation: https://gbatemp.net/threads/homebrew-development.360646/post-6362424
Usage in the wild: https://github.com/masterfeizz/ioQuake3DS/blob/master/code/ctr/ctr_main.c#L44

### DIFF
--- a/backends/platform/3ds/main.cpp
+++ b/backends/platform/3ds/main.cpp
@@ -30,28 +30,8 @@ enum {
 	SYSTEM_MODEL_2DS = 3
 };
 
-struct CommandLine {
-	int argumentCount;
-	char** argumentsValue;
-
-	CommandLine(int argc, char** argv): argumentCount(argc), argumentsValue(argv) {}
-};
-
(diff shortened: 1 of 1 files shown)


sev-:
Thank you!

PR #3436, opened 2021-10-21
SDL: Generalize VSYNC option for 2D games too, in both surface and opengl graphic modes.

vanfanel:
With SDL2, proper VSYNC is perfectly possible across all major OSes, with or without using OpenGL.


aquadran:
you missed change in getFeatureState() in SDL OpenGL backend. you can put "SDL_GL_GetSwapInterval() != 0" under SDL2 define condition:
case OSystem::kFeatureVSync:
   return SDL_GL_GetSwapInterval() != 0;
#endif

aquadran:
👍

PR #3440, opened 2021-10-21
JANITORIAL: Fix default line endings

Die4Ever:
optionally if you want to fix all existing files, you can run

```
git add --renormalize .
```


sev-:
Thanks!

PR #3443, opened 2021-10-22
SCI: Clarify passing of parameter: pass instead of parse for `classtable` debug command.

deckarep:
With respect to SCI debug commands:
  * Minor wording change to say pass instead of parse.


sluicebox:
Thanks! While we've got you here, does this command's filter parameter even work? When I pass a parameter it doesn't reduce the number of lines it prints, it just changes all the ones that don't match to the "Class 0x%x (not loaded; can't get name)..." message. I mean, I'm looking at the code and I'm sure that's wrong, but since you seem to be the one using this, what do you think?

(The mad science you must be up to...)

PR #3446, opened 2021-10-22
GROOVIE: dump_resources config option

Die4Ever:
making a pull request for this because idk if there are any conventions for this kind of thing


bluegr:
IMHO, the best way to implement such debug functionality would be via a new console (debugger) command, for example "dump_resources" inside groovie/debug.*

PR #3467, opened 2021-10-25
DIRECTOR: Add detection for a few german games

besentv:
Some childhood games I still had laying around here.


trembyle:
> the games' executables are in a specific subfolder each, which are not in the detection paths file. Not really sure, if it's a great idea to put them there, though.

Sure, try putting them in `detection_paths.h` and see if they're recognized.

PR #3471, opened 2021-10-30
DIRECTOR: add new variants of existing games

mistydemeo:
This adds several new variants of games that are already in ScummVM; some are full versions of demos, others are language variants. Some notes:

Jungle Park:
* ScummVM immediately closes with an error in the console: `Cast::loadConfig(): Wrong format. VWCF resource missing!`

Peace Land:
* ScummVM closes immediately after loading. The console complains about not being able to locate a score with a specific name, several unimplemented xlibs, an unhandled movie resource, and an unhandled `PICT` resource.

Pickle's Book:
* The Windows and Mac versions seem to have been compiled with different Director versions.
* Gets to the title screen correctly, then hangs with glitchy audio.

Pyramid Interactive:
* Immediately throws an in-ScummVM error from inside `Cast::loadCastDataVWCR()`

World Engine Fantazation:
* The Windows version boots up to an error about not being able to load textures.
* The Mac version has a Lingo syntax error.

### DIFF
--- a/engines/director/detection_tables.h
+++ b/engines/director/detection_tables.h
@@ -1364,7 +1364,8 @@ static const DirectorGameDescription gameDescriptions[] = {
 //
 //////////////////////////////////////////////////
 
-	MACGAME1_l("alice", "", "Alice", "e54ec74aeb4355b0acd46320327c1bed", 274274, Common::JA_JPN, 200),
+	MACGAME1_l("alice", "", "Alice", "e54ec74aeb4355b0acd46320327c1bed", 271740, Common::JA_JPN, 200),
+	MACGAME1_l("alice", "Rerelease", "Alice", "e54ec74aeb4355b0acd46320327c1bed", 274274, Common::JA_JPN, 200),
 
 	MACDEMO1("backgrounds", "Demo", "Backgrounds Demo", "d029fcb13fd2f0944f3c04577ad3b848", 2708196, 200),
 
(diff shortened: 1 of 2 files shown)


mistydemeo:
Added the original 1991 version of Alice, and added a comment to the D2 version the tables already had to note that it's a rerelease. I believe that's the 1992 digipak release.

The Mac versions of Alice were victims to the size misdetection bug, so I've updated the table with the correct filesizes.

engines/director/detection_tables.h
```diff
@@ -1364,7 +1364,8 @@ static const DirectorGameDescription gameDescriptions[] = {
 //
 //////////////////////////////////////////////////
 
-	MACGAME1_l("alice", "", "Alice", "e54ec74aeb4355b0acd46320327c1bed", 274274, Common::JA_JPN, 200),
+	MACGAME1_l("alice", "", "Alice", "e54ec74aeb4355b0acd46320327c1bed", 271740, Common::JA_JPN, 200),
+	MACGAME1_l("alice", "Rerelease", "Alice", "e54ec74aeb4355b0acd46320327c1bed", 274274, Common::JA_JPN, 200),
```
trembyle: Is there a game version number for this one - either in the readme or somewhere in the binary?

trembyle (commented):
Thanks, this looks great.

See my question under Alice. If the answer is no, then that's fine too.

engines/director/detection_tables.h
```diff
@@ -2172,7 +2172,7 @@ static const DirectorGameDescription gameDescriptions[] = {
 	WINGAME1("adamtis", "v2.01",	  "ADAMPLAY.EXE",	   "803570c5edef30938a1efb0900f113e0", 695855, 404),
 	WINDEMO1("adamtis", "Skull Demo", "PUZZLE.EXE",		   "d2dcc8b55382fc1807e357af5108f7ba", 1057997, 404),
 
-	MACGAME1("alice", "", "Alice", "3b61149c922f0fd815ca29686e4f554a", 304714, 400),
+	MACGAME1("alice", "", "Alice", "3b61149c922f0fd815ca29686e4f554a", 304458, 400),
```
trembyle, line 2175: Can you confirm that this is the Digipak release? If so, then we can mark it as such.

trembyle (commented):
Just another question on the Alice releases. Otherwise this all looks good to merge.

engines/director/detection_tables.h
mistydemeo: This is the English-language rerelease. The Digipak release is the "Rerelease" one above that's using D2.

engines/director/detection_tables.h
mistydemeo: I can't find a version in the binary or the README, and the version field in macOS is just holding the unchanged copyright info. I've changed "rerelease" to "digipak" though.

PR #3472, opened 2021-10-30
AGOS: RFC: Fix Waxworks AdLib music (bug #13048)

eriktorbjorn:
This is an attempt at fixing the Waxworks AdLib music. See https://bugs.scummvm.org/ticket/13048 for some details.

The problem turned out to be that when calculating the value to be written to the 40 - 55 AdLib registerers, it could underflow so instead of 0 (max volume) it would write some negative value, probably close to the minimum volume.

Much of this calculation is done in the noteOn() function, but the `channelVolumeAdjust` variable (which is clipped to the appropriate interval) is never used. I assume that was the main error. I've also changed it so that the clipping is done after all the adjustments have been made. (And since we're working with small 16-bit integers, there is no need to use floating-point math either, so I've eliminated that.)

Is it correct? I don't know. At least for Waxworks, it seems that most notes end up being played at max volume, but perhaps that's how it should be? I'm just guessing here, and if anyone has a better fix I'd be happy to drop this one. Whatever the fix is, it should almost certainly go into the 2.5 branch as well.

### DIFF
--- a/engines/agos/drivers/accolade/adlib.cpp
+++ b/engines/agos/drivers/accolade/adlib.cpp
@@ -169,12 +169,11 @@ void MidiDriver_Accolade_AdLib::setVolume(byte volume) {
 	// Set the master volume in range from -128 to 127
 	_masterVolume = CLIP<int>(-128 + volume, -128, 127);
 	for (int i = 0; i < AGOS_ADLIB_VOICES_COUNT; i++) {
-		// Adjust channel volume with the master volume and re-set registers
-		byte adjustedVelocity = _channels[i].velocity * ((float) (128 + _masterVolume) / 128);
-		noteOnSetVolume(i, 1, adjustedVelocity);
+		// Re-set registers
+		noteOnSetVolume(i, 1, _channels[i].velocity);
 		if (i <= AGOS_ADLIB_VOICES_PERCUSSION_START) {
 			// Set second operator for FM voices + first percussion
-			noteOnSetVolume(i, 2, adjustedVelocity);
+			noteOnSetVolume(i, 2, _channels[i].velocity);
 		}
 	}
 }
@@ -290,12 +289,11 @@ void MidiDriver_Accolade_AdLib::noteOn(byte FMvoiceChannel, byte note, byte velo
 	byte regValueB0h      = 0;
 
 	// adjust velocity
-	int16 channelVolumeAdjust = velocity + _channels[FMvoiceChannel].volumeAdjust;
-	// adjust velocity with the master volume
-	channelVolumeAdjust = (channelVolumeAdjust * (128 + _masterVolume)) / 128;
-	channelVolumeAdjust = CLIP<int16>(channelVolumeAdjust, 0, 0x7F);
+	int16 volumeAdjust = _channels[FMvoiceChannel].volumeAdjust;
+	volumeAdjust += velocity;
+	volumeAdjust = CLIP<int16>(volumeAdjust, 0, 0x7F);
 
-	byte adjustedVelocity = channelVolumeAdjust;
+	byte adjustedVelocity = volumeAdjust;
 
 	if (!_musicDrvMode) {
 		// INSTR.DAT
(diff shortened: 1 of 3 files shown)


athrxx:
I have tested this and can confirm that it at least fixes the issue. The music seems okay now.

But I can also confirm that mostly everything seems to be played at max volume. It seems irrelevant whether I set the GMM music volume to 255 or to 192, since it is all clipped to the same max vol.
I suspect that the sound driver is not accurate, but it is hard to tell without more research. I haven't compared with DOSBox. And sometimes the original drivers just aren't really good either.

Anyway, that's not the point of this PR. If someone wants to rework the driver he/she can always do it...

engines/agos/drivers/accolade/adlib.cpp
```diff
@@ -286,19 +286,16 @@ void MidiDriver_Accolade_AdLib::setTimerCallback(void *timerParam, Common::Timer
...
 
 	// adjust velocity
-	int16 channelVolumeAdjust = _channels[FMvoiceChannel].volumeAdjust;
-	channelVolumeAdjust += adjustedVelocity;
-	channelVolumeAdjust = CLIP<int16>(channelVolumeAdjust, 0, 0x7F);
-
+	int16 channelVolumeAdjust = velocity + _channels[FMvoiceChannel].volumeAdjust;
 	// adjust velocity with the master volume
-	byte volumeAdjust = adjustedVelocity * ((float) (128 + _masterVolume) / 128);
+	channelVolumeAdjust = (channelVolumeAdjust * (128 + _masterVolume)) / 128;
```
NMIError, line 295: I think the problem is here. _masterVolume has a range of -128 to 127, so after adding 128 0 to 255. It is then multiplied with a MIDI velocity (0-127) and then divided by 128, leaving a value in the range of 0 to 255. It should probably be divided by 255 to again get a MIDI velocity value range (0-127).

engines/agos/drivers/accolade/adlib.cpp
```diff
@@ -286,19 +286,16 @@ void MidiDriver_Accolade_AdLib::setTimerCallback(void *timerParam, Common::Timer
...
+	channelVolumeAdjust = CLIP<int16>(channelVolumeAdjust, 0, 0x7F);
 
-	adjustedVelocity = volumeAdjust;
+	byte adjustedVelocity = channelVolumeAdjust;
```
NMIError, line 298: Note that this value is later set on _channels[FMvoiceChannel].velocity. This is used in setVolume, where a new _masterVolume value is applied. So the value set there should be just velocity and volumeAdjust, without _masterVolume applied. Also, setVolume has the same bugged volume calculation.

eriktorbjorn: Ok, I'm getting pretty confused over here so I'm just going to write down some stuff for my own benefit here... :-)

Before the changes, we have the following variables:

- The **velocity** input parameter.
- **channelVolumeAdjust** which is calculated as CLIP<int16>(velocity + _channels[FMvoiceChannel].volumeAdjust)
- **volumeAdjust** which is calculated as velocity * ((floag) (128 + _masterVolume) / 128)
- **adjustedVelocity**, which is set to **volumeAdjust**.

After this point, **velocity**, **channelVolumeAdjust** and **volumeAdjust** are no longer used. (Which means that **channelVolumeAdjust** isn't used at all.)

**adjustedVelocity** is then further adjusted, before it's stored in _channels[FMvoiceChannel].velocity and passed on to noteOnSetVolume().

So if _masterVolume should not be factored into _channels[FMvoiceChannel].velocity, I guess it would be more sensible to do that adjustment in noteOnSetVolume()?

eriktorbjorn: I've made some further changes. Now the volume seems a bit low to me, but of course we could have a different value than 15 for the master volume.

NMIError: Yes, that will put it to half volume. I thought master volume was set from the user volume setting, but that does not seem to be the case. That should probably be added to MidiPlayer::setVolume (only for this driver).

eriktorbjorn: Just to confuse things - or at least me - even further, this is what AGOS's MidiPlayer::pause() does:

```
	if (musicType == MT_ADLIB && _musicMode == kMusicModeAccolade) {
		static_cast <MidiDriver_Accolade_AdLib*> (_driver)->setVolume(_paused ? 0 : 128);
	}
```

So here it seems like 0 means silence, and 128 means... unchanged volume? Half volume? Either way, 128 won't reset _masterVolume to 15.

I'm not sure how to proceed here...

NMIError: The range for the setVolume parameter is 0 - 255. This is then converted to -128 to 127 for _masterVolume, then back to 0 - 255 for the volume calculations because... reasons. The unpaused value should be the previously set volume, i.e. MidiPlayer::_musicVolume or _sfxVolume, whichever one you decide to use in MidiPlayer::setVolume.

eriktorbjorn:
That's probably as good as it's going to get at my hands. It's still trying to play music at or above the maximum volume, but now the ScummVM music volume setting can bring it down. Clipping is done after the master volume has been applied.

Of course, if anyone has a better idea how to do it I'd be happy to drop this pull request in a heartbeat.

NMIError:
This needs some further investigation, but it looks good enough to fix the immediate problem.

sev-:
@NMIError thus, do you think it is OK to merge?

NMIError:
Yes, looks fine to me.

sev-:
Thank you, @eriktorbjorn for the work and @NMIError for the review. Merging

PR #3473, opened 2021-10-30
SDL: Fix warning with GCC 11 on MinGW

orgads:
```
In file included from C:/msys64/mingw64/include/c++/11.2.0/cstdlib:41,
                 from C:/msys64/mingw64/include/c++/11.2.0/stdlib.h:36,
                 from ../scummvm/common/scummsys.h:120,
                 from ../scummvm/backends/platform/sdl/win32/win32-main.cpp:28:
C:/msys64/mingw64/include/c++/11.2.0/x86_64-w64-mingw32/bits/c++config.h:573:2: warning: #warning "__STRICT_ANSI__ seems to have been undefined; this is not supported" [-Wcpp]
  573 | #warning "__STRICT_ANSI__ seems to have been undefined; this is not supported"
      |  ^~~~~~~
```

### DIFF
--- a/backends/platform/sdl/win32/win32-main.cpp
+++ b/backends/platform/sdl/win32/win32-main.cpp
@@ -23,7 +23,9 @@
 // Disable symbol overrides so that we can use system headers.
 #define FORBIDDEN_SYMBOL_ALLOW_ALL
 // HACK: Mingw32 doesn't define _argc _argv in strict ANSI mode
+#if !defined(__GNUC__) || __GNUC__ < 11
 #undef __STRICT_ANSI__
+#endif
 
 #include "common/scummsys.h"
 


lephilousophe:
This looks like a bit hackish but I don't see other solution.
The main problem is that original mingw32 (not mingw32-w64) doesn't let you access argc and argv with `_STRICT_ANSI_` defined.
On the other hand, g++ doesn't let you disable all GNU extension without disabling the non-standard functions and defines...
For reference, warning comes from gcc-mirror/gcc@767537a8b027bcb5807bb45b0268c5da98c2c7a0.

sev-:
Do I understand correctly, that #3476 should eliminate this problem?

orgads:
No, this is still required.

lephilousophe:
Or we add back the mingw platform to make it use gnu++11 instead of c++11 and it will work without all of this __STRICT_ANSI__ stuff.

orgads:
Whatever works for you is fine with me.

lephilousophe:
Fixed by using gnu++11.

orgads:
Thanks!

PR #3474, opened 2021-10-30
CREATE_PROJECT: Add DETECTION_FULL feature

sluicebox:
Allows disabling full detection with `--disable-detection-full`, just like the configure script. `detection_table.h` now has the same content as the one generated by the configure script. This doesn't change any default behavior.

Tested on msvc and xcode. Speeds up compiling my one-engine local builds. Also makes it possible to build again on older msvc versions that have trouble with some of the newer engine code that detection objects bring in, but that's just a bonus for me.


sev-:
Yay! Thank you!

PR #3475, opened 2021-10-31
MOHAWK: Add sizes, versions to some Myst entries

einstein95:
"Making of Myst" from JP and IT Win 3.11 is in English, so mark it as `EN_ANY`

Doesn't resolve the hash collision with the EN v1.0.1 and DE v? entries.


sev-:
Excellent, thank you!

PR #3476, opened 2021-10-31
Rework configure

lephilousophe:
This series of commit remove several outdated checks and flags from the configure script.

First we move the C++11 test before everything else to make sure the compiler does support C++11, else version check is kind of useless.
Then, the checks for old version of GCC (<2.95, <3, <4.3) are all removed as the first bits of C++11 were included in GCC 4.3 (cf. https://gcc.gnu.org/projects/cxx-status.html#cxx11).
Next, I remove the variable HAVE_GCC3 is deleted as it not used anymore.
Finally, there is some try to rework the `__STRICT_ANSI__` handling as GCC 11 doesn't support anymore to undefine `__STRICT_ANSI__` without enabling GNU extensions. In addition, the pedantic handling is moved at the same place as it's directly related to standards.
This should help with Switch toolchain which uses a GCC11.

@orgads adding back mingw to the list of gnu11 platforms could help with both mingw32 and mingw-w64.

Before moving this PR out of draft I would like to do a build with all buildbot platforms.


orgads:
Looks good, and builds on MSYS2/MinGW64 without warnings.

configure
```diff
@@ -2144,63 +2154,56 @@ echo "$cxx_version"
 if test "$cxx_verc_fail" = yes ; then
 	echo
 	echo "The version of your compiler is not supported at this time"
-	echo "Please ensure you are using GCC >= 2.95"
```
sev-, line 2147: I would add something like "it is too low"

eriktorbjorn: Or perhaps "too old"?

lephilousophe: This check will now fail not because version is bad but because compiler is not recognized.
We support GCC, clang and ICC. If it's not one of them, and not some special SGI compiler, we will end up here.
That's why I removed this sentence.

lephilousophe:
I tested the changes on a buildbot and it works.

sev-:
Great, merging.

configure
```diff
@@ -2178,13 +2191,19 @@ if test "$have_gcc" = yes ; then
...
 
 #
 # Set status about C++11 mode
 #
-append_var CXXFLAGS "-std=c++11"
+append_var CXXFLAGS "-std=${std_variant}11"
+
+if test "$pedantic" = yes ; then
+	# We *do* want the 'long long' extension.
+	append_var CXXFLAGS "-pedantic -Wno-long-long"
```
criezy, line 2205: Is the `-Wno-long-long` still needed? The `long long` type is part  of the c++11 standard, so this is no longer an extension.

lephilousophe: Ah! I didn't check for this, just moved around the lines.
I can throw a test during the week to remove this.
Thanks!

lephilousophe: Done

PR #3480, opened 2021-11-01
COMMON/DIRECTOR: Move semantics for Arrays and some DIRECTOR structs

henke37:
To celebrate the support for c++11 everywhere, I added move semantics to the `director::InfoEntry` struct.

Turns out that didn't work with `Common::Array` as is, so I had to add support for move semantics. While at it I figured that I might as well make use of move semantics where possible without changing semantics to speed things up.

### DIFF
--- a/common/algorithm.h
+++ b/common/algorithm.h
@@ -57,6 +57,19 @@ Out copy(In first, In last, Out dst) {
 }
 
 /**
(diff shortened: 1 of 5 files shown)


sev-:
We move to C++11 but not to STL.

PR #3481, opened 2021-11-02
DEVTOOLS: Work around Mac-Japanese sequences missing low byte

einstein95:
This fixes dumping the MacPeople 1998 #2 coverdisc, which has Mac-Roman filenames on the (Japanese) disc.


gingerbeardman:
There's a typo in line 130 that might make sense to fix in this PR?

sev-:
Thanks

PR #3485, opened 2021-11-03
DEVTOOLS: Add test for previous dumpercompanion workaround, other changes

einstein95:
Use "MacJapanese" instead of "Mac Japanese", remove commented code from previous PR. These were requested by @rvanlaar

### DIFF
--- a/devtools/dumper-companion.py
+++ b/devtools/dumper-companion.py
@@ -102,8 +102,7 @@ def decode_macjapanese(text: ByteString) -> str:
         elif (0x81 <= hi <= 0x9F) or (0xE0 <= hi <= 0xFC):  # two-byte sequence
             lo = next(i_text, None)
             if lo is None:
-                # raise Exception(f"Mac Japanese sequence missing second byte 0x{hi:02x}")
-                print(f"WARNING: Mac Japanese sequence missing second byte 0x{hi:02x}, decoding as Mac-Roman")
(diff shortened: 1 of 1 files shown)


bluegr:
Thanks for the fixes! Merging

gingerbeardman:
I just came across some other characters that are not dealt with, and upon reading the spec it seems Shift+JIS mapping is missing from the Japanese logic.

see: "Unicode mapping issues and notes"
ftp://ftp.unicode.org/Public/MAPPINGS/VENDORS/APPLE/JAPANESE.TXT

I'll try to fix this myself, please expect a PR

gingerbeardman:
Sadly I don't know Python well enough to code a map like that efficiently.

gingerbeardman:
My WIP is at https://github.com/gingerbeardman/scummvm/commit/39f41332712d4daa7893a7659171d703e50db774

rvanlaar:
@gingerbeardman Thank you for bringing this to our attention. The mapping was automatically generated. We're looking into regenerating it with the included Shift-JIS ranges

gingerbeardman:
My final solution to my task of generating directory listings of these old discs is to use hfsutils (for raw directory listings) plus a small Tcl script (to do the re-encoding).

So, all this to say I probably won't find time to clean up my PR. Feel free to use any part of it or abandon it altogether and close it.

PR #3487, opened 2021-11-04
DEVTOOLS: fix encoding with the dumper companion `mac` command

mistydemeo:
The `mac` command in the Python dumper companion tried to handle filename encodings, which is, if I'm being honest, a fool's errand. This led it to error out on certain Japanese filenames, for example:

```
Resource in ÇœÇªÇ∑ÇØPPC
Traceback (most recent call last):
  File "/Users/mistydemeo/github/scummvm/devtools/dumper-companion.py", line 488, in <module>
    exit(args.func(args))
  File "/Users/mistydemeo/github/scummvm/devtools/dumper-companion.py", line 404, in collect_forks
    to_file.write(file_to_macbin(file, to_filename.encode("mac_roman")))
  File "/usr/local/Cellar/python@3.9/3.9.1_6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/encodings/mac_roman.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_table)
UnicodeEncodeError: 'charmap' codec can't encode character '\u0327' in position 1: character maps to <undefined>
```

I've switched the path handling to use pure bytes instead so it can rely on having just the filenames as they are on the filesystem. I've confirmed this lets me use the dumper companion with Japanese Mac discs now. More changes will be needed for punycode encoding to work, since that has to be aware of the underlying string encoding to work.

I've also added a second fix which addresses an overly-aggressive `AttributeError` catch I added previously, which ended up catching attribute errors not related to the argparse code.

### DIFF
--- a/devtools/dumper-companion.py
+++ b/devtools/dumper-companion.py
@@ -550,6 +550,7 @@ def test_decode_name():
         ["G3フォルダ", "xn--G3-3g4axdtexf"],
         ["Where \\ Do <you> Want / To: G* ? ;Unless=nowhere,or|\"(everything)/\":*|\\?%<>,;=", "xn--Where  Do you Want  To G  ;Unless=nowhere,or(everything),;=-5baedgdcbtamaaaaaaaaa99woa3wnnmb82aqb71ekb9g3c1f1cyb7bx6rfcv2pxa"],
         ["Buried in Timeｪ Demo", "xn--Buried in Time Demo-yp97h"],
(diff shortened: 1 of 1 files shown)


sev-:
Cool stuff! Could you please also add this filename as a test to `test_decode_name()` (line 537 or so).

mistydemeo:
Done!

I'm looking into adding correct-encoding punycoding, but that'll take more work so I'd like to do it as a followup PR.

PR #3492, opened 2021-11-06
SCI: add "st" as an alias to the debug "stack" command

deckarep:
Since I am constantly evaluating the stack, please consider adding this alias for the stack command. 
  * Add `st` as an alias to `stack` command
  * I'm hoping this is a reasonable request since it also nicely models stack based addresses such as: `ST:0001`
  * I previously added an alias for the `registers` command as well named: `reg`


bluegr:
Thanks! Good idea. Merging

PR #3493, opened 2021-11-06
BACKENDS: DIALOG: GTK: Keep main window updated during GTK+ dialog

eriktorbjorn:
One thing that's always bothered me about the GTK+ dialogs is that the main window doesn't get updated while they are open. At least for me this causes ugly trails in the main window of anything that's moved across it while the dialog is open.

This patch adds an idle handler to keep the main window updated.

Maybe the idle handler should be part of the GtkDialogManager class, but https://stackoverflow.com/questions/21478803/member-function-as-callback-function-to-g-signal-connect suggested that it might be more portable to keep it outside the class? I don't know for sure.

I'd be grateful for feedback for anyone more familiar with GTK+ than me. (Which would be most people familiar with GTK+ at all.)

### DIFF
--- a/backends/dialogs/gtk/gtk-dialogs.cpp
+++ b/backends/dialogs/gtk/gtk-dialogs.cpp
@@ -34,6 +34,16 @@
 
 #include <gtk/gtk.h>
 
+// TODO: Move this into the class? Probably possible, but I've been told that
+//       this might not necessarily work properly with all compilers.
+
+static gboolean _inDialog;
+
+static gboolean idleCallback(gpointer data) {
+	g_system->updateScreen();
+	return _inDialog;
+}
+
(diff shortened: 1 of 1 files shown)


eriktorbjorn:
Example of what it can look like for me before this change. ScummVM doesn't update the main window, and apparently the window manager doesn't do anything to mitigate it:

unhappy-trails

sev-:
This is a marvellous addition, thank you, merging

eriktorbjorn:
> This is a marvellous addition, thank you, merging

Thanks. Though one thing I just noticed is that while the mouse is moving over the SDL main window, there's no cursor. This is not a regression, mind you, but I do wonder if there's any good way of fixing that.

sev-:
Yes, you need to poll events as well.

PR #3496, opened 2021-11-07
DIRECTOR: close file handles when no stream created

mistydemeo:
`getAudioStream()` was leaking file descriptors. The caller closes the file if a stream could be created, but in the case one isn't then the file will never be closed. In the worst case scenario, it could hit the descriptor table size and prevent any further file open calls from working.

cc @moralrecordings


sev-:
Good catcj. thank you!

PR #3497, opened 2021-11-07
DIRECTOR: handle macbinary audio streams

mistydemeo:
Certain audio streams can (and often do) contain resource forks, so the dumper companion will write them as macbinary. Right now, the audio code will choke on them since it tries to read from the macbinary header assuming it's audio. This adds a check for whether the file is macbinary and handles it appropriately.

In the future, I might run a pass over the Director code to support native Mac files on HFS+/APFS macOS, but there's a lot I'd need to update for that.

To test this, you can try the Pippin/Mac version of Jungle Park run through the dumper companion. Without this patch, it'll fail to recognize `JUNGLE/JINGLE.AIFF`; it won't play the intro audio and will spam the console with errors about it. With this patch. it works fine.

### DIFF
--- a/engines/director/sound.cpp
+++ b/engines/director/sound.cpp
@@ -25,6 +25,7 @@
 // License: GNU GPL v2 (see COPYING file for details)
 
 #include "common/file.h"
+#include "common/macresman.h"
 #include "common/substream.h"
 
 #include "audio/decoders/wave.h"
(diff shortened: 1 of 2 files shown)


sev-:
In fact, MacResMan is handling all the flavours of Mac resource forks, including the native. See `MacResManager::open()`. So, even this code could be simplified. I would write it something like this:

```C++
SeekableReadStream *stream;

if (_vm->getPlatform() != Common::kPlatformWindows) {
  _macresman = new Common::MacResManager(); // put the variable into class
  _macresman->open(Common::Path(pathMakeRelative(_path), g_director->_dirSeparator));
  stream = _macresman->getDataFork();
} else {
  stream = new File();
  stream->open(....
}
```

This is a logically clean solution. However, less logical but perfectly working would be to just use `MacResManager::open()`. It has a fallback mechanism that if the file is not binary, it is treating is as a plain data fork. And Data Fork is what you need.

Thus, in this particular case you could write just the following, and if you would need a resource fork, then the more complicated code from above:

```C++
 _macresman = new Common::MacResManager(); // put the variable into class
 _macresman->open(Common::Path(pathMakeRelative(_path), g_director->_dirSeparator));
 file = _macresman->getDataFork();
```

mistydemeo:
Went for the second solution. No idea what I did wrong last time, but when I tried it before it seemed as though it didn't work - but it seems just fine now!

sev-:
Thank you!

PR #3498, opened 2021-11-07
CREATE_PROJECT: Remove VS2008 project generator

sluicebox:
Now that C++11 is mandatory we can retire our old friend


bluegr:
Thanks!
It should be noted that C++11 is fully supported in VS 2019 - earlier versions were missing parts of the standard:
https://docs.microsoft.com/en-us/cpp/overview/install-c17-support?view=msvc-160

VS2008 definitely doesn't support C++11, as mentioned. So, removing support for this old compiler it keeps the tool codebase tidy.

Thanks again! Merging

PR #3500, opened 2021-11-07
SURFACESDL: Implemented fallback in case lack of VSYNC

aquadran:
Added fallback for SDL surface backend in case lack of VSYNC.
Missing VSYNC is possible in case SDL renderer doesn't support, like dummy driver or system GFX driver not able (broken) or can not enable it (lack VSYNC on output).

Tested only with "SDL_VIDEODRIVER=dummy"

Along with changes added missing kFeatureVSync functionality for getFeatureState()

### DIFF
--- a/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
+++ b/backends/graphics/surfacesdl/surfacesdl-graphics.cpp
@@ -109,7 +109,7 @@ SurfaceSdlGraphicsManager::SurfaceSdlGraphicsManager(SdlEventSource *sdlEventSou
 	_osdIconSurface(nullptr),
 #endif
 #if SDL_VERSION_ATLEAST(2, 0, 0)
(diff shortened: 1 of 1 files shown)


backends/graphics/surfacesdl/surfacesdl-graphics.cpp
```diff
@@ -2585,8 +2589,17 @@ SDL_Surface *SurfaceSdlGraphicsManager::SDL_SetVideoMode(int width, int height,
 
 	_renderer = SDL_CreateRenderer(_window->getSDLWindow(), -1, rendererFlags);
 	if (!_renderer) {
-		deinitializeRenderer();
-		return nullptr;
+		if (_vsync) {
```
bluegr, line 2592: We could check for rendererFlags & SDL_RENDERER_PRESENTVSYNC here, no need for a separate flag

PR #3504, opened 2021-11-08
DIRECTOR: Add patches to get Windows version of A Silly Noisy House in-game

einstein95:
The `changedrive` patch is currently necessary as the current `FileIO` implementation will blindly check for invalid characters in filenames.


trembyle:
I confirmed that we can now reach the main menu. Excellent, thanks!

trembyle:
I confirmed that it can now reach the main menu. Excellent, thanks!

PR #3506, opened 2021-11-08
XEEN: fix caster selection during non-combat spell casting

benoit-pierre:
Steps to reproduce:
- bring the casting dialog while not in combat
- try selecting another character

→ instead of changing the caster, the spell is aborted


dreammaster:
Good catch noticing that regression.

PR #3507, opened 2021-11-09
DIRECTOR: decode movie paths before loading

mistydemeo:
This fixes an issue where movies in `probeMacBinary` weren't being decoded before being loaded. In the case of Japanese games that used MacJapanese filenames, this could lead to errors loading them. Before the patch:

```
WARNING: Couldn't find score with name: WackyRaces: WackyRaces: Main:�Entrance!
```

After:

```
WARNING: Replaced score name with:  WackyRaces: Main:･Entrance (from WackyRaces: WackyRaces: Main:･Entrance)!
```

This fixes loading a scene in the Mac version of Wacky Races, though it still doesn't make it past the intro video.

### DIFF
--- a/engines/director/resource.cpp
+++ b/engines/director/resource.cpp
@@ -153,7 +153,7 @@ void Window::probeMacBinary(MacArchive *archive) {
 			if (num == 0)
 				error("No strings in Projector file");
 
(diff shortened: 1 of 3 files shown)


djsrv:
Thanks!

PR #3512, opened 2021-11-10
VIDEO: Fix QuickTimeDecoder bugs

sluicebox:
The KQ6 Mac QuickTime opening movie doesn't play correctly in ScummVM. There are frozen frames and other problems at transitions between tracks and edits. https://bugs.scummvm.org/ticket/11085

There are two bugs preventing correct playback:

1. A workaround from three years ago for a video in a Spanish version of Riven. The workaround has several problems and it breaks valid movies. (Added in: b8abe400850a23d12fe5cdc24d7106820d0f13fd)
2. QuickTimeDecoder doesn't correctly handle tracks where the first edit is empty (a delay) and the second has a `mediaTime` offset. `mediaTime` is ignored in this case, causing the first frame of the media to be played instead of the intended later frame.

I've limited the Riven workaround to just Riven through a flag, this mitigates that regression. It's important to keep that workaround, even if buggy, since it prevents a crash. (Hopefully it can be fixed someday by someone with that video)

The decoder bug is an initialization/workflow problem. All tracks are constructed at the start where they each enter their first edit. Entering an edit calculates the next frame of the edit and next frame start time. But first, `enterNewEditList()` skips empty edits. That's a problem because the track constructor then resets the first frame, undoing the calculation. The edit is never re-initialized since it's already been entered. In other words, there are accidentally two paths to real edit initialization: the one during playback which correctly handles everything, and the one during track construction where the caller is unaware of skipping past edits or the need to buffer frames in case the first edit has a `mediaTime`.

The fix is to no longer skip empty edits, which are just delays until the next edit starts. Instead, a track's current edit can now happily be in an empty edit state during an empty edit. This is a more accurate representation of the movie playback model. This causes subsequent edits to now always be initialized during playback just like all other edits. The trade-off is to just add a check to `decodeNextFrame()` and `getRateAdjustedFrameTime()` so that they take empty edits into account.

Also we now buffer frames during track initialization, which fixes the case where the first edit in a track has `mediaTime`; otherwise the frame's keyframe wouldn't get processed.

I'm not going to admit how long I spent working on this, but...

image

### DIFF
--- a/common/quicktime.cpp
+++ b/common/quicktime.cpp
@@ -368,7 +368,6 @@ int QuickTimeParser::readTRAK(Atom atom) {
 	Track *track = new Track();
 
 	track->codecType = CODEC_TYPE_MOV_OTHER;
-	track->startTime = 0; // XXX: check
 	_tracks.push_back(track);
 
 	return readDefault(atom);
@@ -837,7 +836,6 @@ QuickTimeParser::Track::Track() {
 	codecType = CODEC_TYPE_MOV_OTHER;
 	frameCount = 0;
 	duration = 0;
-	startTime = 0;
 	mediaDuration = 0;
 }
 
--- a/common/quicktime.h
+++ b/common/quicktime.h
@@ -93,7 +93,7 @@ protected:
 
 	struct TimeToSampleEntry {
 		int count;
-		int duration;
+		int duration; // media time
 	};
 
 	struct SampleToChunkEntry {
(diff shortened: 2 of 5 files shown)


bluegr:
Excellent! This reduces the scope of the QT workaround/hack. Since this is not an applicable workaround for all QT movies, the approach with the quirk flag is the best that can be done for this.

Ideally, this should be applied at the engine level and more specifically, only for that specific QT movie, but that's an extra improvement that can be done in the future.

Thanks for your work! Merging

sev-:
This led to regressions in Director.

sluicebox:
Fixed in 69662ee9a44ef8124ff282c716422e67f06f282a

PR #3513, opened 2021-11-10
SAGA: Fix ITE digial music looping regression

eriktorbjorn:
This attempts to fix the music looping regression described in https://forums.scummvm.org/viewtopic.php?t=16434

See commit message for details about where the regression came from.

I was tempted to commit it right away, but I'd appreciate if someone much more familiar with Inherit the Earth could comment on it first. After that, it's probably a good candidate for the 2.5 branch as well.

### DIFF
--- a/engines/saga/music.cpp
+++ b/engines/saga/music.cpp
@@ -349,7 +349,7 @@ bool Music::playDigital(uint32 resourceId, MusicFlags flags) {
 	int realTrackNumber = 0;
 
 	if (_vm->getGameId() == GID_ITE) {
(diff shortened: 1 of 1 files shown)


bluegr:
Since I did the relevant change back then, I can confirm that the problem in the original commit (b14a2599bcbb5fcec2e860e048081751dbde8162) was a mistake from my part. Thanks for doing the R&D work on this and fixing the music loop behavior!

Merging

eriktorbjorn:
And thank _you_ for confirming that it was the correct fix. I'll merge it to the 2.5 branch and add a NEWS item for it.

PR #3516, opened 2021-11-11
CI: Fix MSVC builds

sluicebox:
The MSVC builds have been failing because PCRE changed their download server. https://github.com/microsoft/vcpkg/issues/21201

Fixed by updating vcpkg commit and dropping the "d" from the libjpeg-turbo debug library. There's no longer a suffix on this library when buildling with vcpkg.

I'm doing this as a PR so that someone with access can update the lib filename in http://downloads.scummvm.org/frs/build/scummvm_libs_2015.zip

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
         uses: lukka/run-vcpkg@v7
         id: runvcpkg
         with:
-          vcpkgGitCommitId: b67fab9861239e33b722b9a44f5c96dec6ca807e
+          vcpkgGitCommitId: 70033dbb31527fb3e69654731f540f59c87787f9
           vcpkgTriplet: '${{ matrix.triplet }}'
           vcpkgArguments: '${{ matrix.vcpkgPackages }}'
 #      - name: Upload libs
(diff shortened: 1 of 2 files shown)


sev-:
Could you please clarify, which exactly files need to be renamed in the archive and to which? There are 4 files with 'jpegd' in them.

sluicebox:
`lib/x64/Debug/jpegd.lb`
`lib/x86/Debug/jpegd.lb`

Both need to be renamed to `jpeg.lib`

(There are the only two files with `jpegd` in their name; I'm not sure how you're seeing four)

sev-:
I also see `turbojpegd.lib`.

Anyway, I renamed the file in the archive and merging this. Please test.

sluicebox:
You're right; wow that search screen I used is... *not* good... =)

Tested with the updated archive and a fresh local build; works great, thanks! I don't see anything that references those "turbo" libs, I guess those are for manually using if you want.

lephilousophe:
I believe the `turbojpeg.lib` file is useless for us.
When you build turbo-jpeg, it provides two interfaces: one compatible with original libjpeg (the `jpeg.lib`, which you invoke on GCC with `-ljpeg`) and a turbojpeg interface with different API (the `turbojpeg.lib` file, linked with `-lturbojpeg` on GCC).
The latter is not used by ScummVM.

PR #3517, opened 2021-11-11
OPENDINGUX: Various improvements to building process

lephilousophe:
This PR embeds the various platforms supported by OpenDingux beta directly in the configure script.
This will allow the buildbot to include builds for these platforms as it can't use the release build script.

In addition some autodetections are enabled back to allow make use of the features on GCW0 (Fluidsynth, Theora, SDL2) and OpenGL ES2 is enabled on GCW0 as the toolchain provides Mesa.

Some small fixes are provided to the make-opk script and the build_odbeta is reworked to make use of configure changes and to make use of arguments instead of environment variables to create the build.

### DIFF
--- a/configure
+++ b/configure
@@ -829,7 +829,9 @@ Special configuration feature:
                                            ios7-arm64 for Apple iPhone / iPad (iOS >= 7, 64-bit)
                                            maemo for Nokia Maemo
                                            n64 for Nintendo 64
-                                           opendingux for Opendingux Beta
+                                           opendingux-gcw0 for GCW0 with Opendingux Beta
+                                           opendingux-lepus for Lepus with Opendingux Beta
+                                           opendingux-rg99 for RG99 with Opendingux Beta
                                            openpandora for OpenPandora
                                            ouya for OUYA
                                            ps3 for PlayStation 3
@@ -1576,7 +1578,7 @@ caanoo)
 	_host_cpu=arm
 	_host_alias=arm-none-linux-gnueabi
 	;;
-dingux | gcw0 | opendingux)
+dingux | gcw0 | opendingux-*)
 	_host_os=linux
 	_host_cpu=mipsel
 	_host_alias=mipsel-linux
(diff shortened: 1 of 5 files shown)


citral23 (approved):
Very well, it is indeed a good option to automate the builds more easily.

citral23 (commented):
Hi, do you have a jz4770 device to test it yourself?

Because in my testing : 

Fluidsynth doesn't work (unfortunately, it would be nice)
Theora is too slow (we have no HW accel at all, mpeg2 in game videos run at 3fps to give you an idea)
SDL2 brings no benefit, as it's much, much slower than SDL1 for software rendering and gles2/es2 crashes (haven't looked why yet)

citral23 (approved):

citral23 (approved):

citral23 (approved):

configure
```diff
@@ -3514,6 +3514,28 @@ if test -n "$_host"; then
 			_backend="opendingux"
 			_port_mk="backends/platform/sdl/opendingux/opendingux.mk"
 			add_line_to_config_mk 'OPENDINGUX = 1'
+			add_line_to_config_mk "OPENDINGUX_TARGET = ${_host#opendingux-}"
+			case "$_host" in
+				opendingux-gcw0)
+					_opengl_mode=gles2
+					_opengl_game=yes
```
aquadran, line 3521: does it have desktop opnegl?

citral23: Nope, gles1/2 and gl4es

aquadran: gl4es acts for desktop opengl. just making sure it was not mistake

lephilousophe:
I don't have the hardware but, as libraries were present in the toolchain, I expected device was powerful enough for them.
I removed the commit which enabled all of this.

citral23:
> I don't have the hardware but, as libraries were present in the toolchain, I expected device was powerful enough for them. I removed the commit which enabled all of this.

Yeah, I hoped and tried too, but while fairly capable for what it is, it's still a single core 1GHz cpu
- Not nearly enough for Fluidsynth, and Timidity doesn't work unfortunately (mixed them up in my previous comment). I think fluidsynth should be removed from the toolchain.
-  Theora works on mpv at around 240/272p but without HW accel it won't go above that.
- GLES2 works with some other software, but crashes on scummvm. I will look into it later down the road, but for now it's better to disable it.

lephilousophe:
OK, that's done.

PR #3520, opened 2021-11-11
DEVTOOLS: add Mac fonts to gitignore

mistydemeo:
A few other devtools-generated data files that aren't tracked in git are in this `.gitignore` to prevent them showing up as untracked files to git. I think it makes sense to add these two files as well. They're generated by two scripts in `devtools`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,8 @@ project.xcworkspace
 /dists/engine-data/testbed-audiocd-files/testbed.config
 /dists/engine-data/testbed-audiocd-files/testbed.out
 /dists/engine-data/playground3d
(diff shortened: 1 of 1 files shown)


sev-:
I wonder, why. Those files should not be in that directory unless you set it as your 'extra' directory.

mistydemeo:
This is the default extras directory when building from source.

sev-:
Interesting, could you point to the code where this default is set?

mistydemeo:
Maybe I'm wrong! I never customized that path AFAICT and it's where it's loading them from though. It's also where I was recommended to put them in Discord.

PR #3531, opened 2021-11-13
ENGINES: Use override

orgads:
Using clang-tidy modernize-use-override

### DIFF
--- a/engines/grim/emi/sound/mp3track.cpp
+++ b/engines/grim/emi/sound/mp3track.cpp
@@ -53,7 +53,7 @@ public:
 			_done = true;
 	}
 
-	int readBuffer(int16 *buffer, const int numSamples) {
+	int readBuffer(int16 *buffer, const int numSamples) override {
 		if (_done)
 			return 0;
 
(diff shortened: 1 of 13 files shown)


digitall:
@orgads : I would suggest splitting this change into a commit per engine as per your PR #3533 ... This makes bisecting for regressions and possible commit reversion in future easier.

orgads:
There are 13 files overall, and the change is mechanical, so I don't think it's worth splitting.

If you insist, I can do it.

orgads:
Note that buried had many changes, so I pushed it as a separate PR #3530.

digitall:
@orgads I know what you mean, but it is neater and as I said makes future bisection / revert much easier and less manual.

digitall:
@orgads : Thanks for looking at these improvements to override / nullptrs... but I just lost a week at work dealing with a difference between the older implementation of __FUNCTION__ and the newer one based on C++-11 __func__ which _should_ be equivalent... but had a very bloody subtle breakage... hence :)

orgads:
Split done.

What are the differences between `__FUNCTION__` and `__func__`?

digitall:
Thanks. It is a little involved, but basically the older \_\_FUNCTION\_\_ behaved like a string literal whereas the newer \_\_func\_\_ is an implicit definition of a static string in the function by the compiler... \_\_FUNCTION\_\_ which is C99 is now an alias to the C++ version. Functionally, they are identical... except when using embedded target GCC, where the older C99 construct allowed you to reallocate to flash memory without using RAM, the newer construct does not, so any code using this ported from C99 to a newer embedded compiler compiles and runs fine... but can run out of RAM at runtime when debug is enabled. *HEADACHE*. There is no way of forcing the strings into program memory with the newer standard.

bluegr:
Nice catch! Thanks for splitting this into different commits. Merging

PR #3532, opened 2021-11-13
JANITORIAL: Use override

orgads:
Using clang-tidy modernize-use-override


bluegr:
clang-tidy did a thorough and straightforward cleanup here. Thanks! Merging

PR #3534, opened 2021-11-13
GUI: Fix cursor stuttering with vsync enabled

lotharsm:
Currently, with vsync enabled, the mouse cursor in the GUI Launcher feels very sluggish on a 60 Hz display. The issue is that the artificial waiting/frame times in the GUI launcher code cause a race condition between VSync an GUI manager's own frame timing.

This patch fixes this by skipping our own frametime calculation with VSync enabled. Hardcoding it to `delayMillis(1);` is required here since otherwise, ScummVM would always max out one CPU thread entirely.

### DIFF
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -548,8 +548,9 @@ void GuiManager::runLoop() {
 		redraw();
 
 		// Delay until the allocated frame time is elapsed to match the target frame rate.
-		// In case we have vsync enabled, we force a minimum frame duration of 1 millisecond
-		// since otherwise, we'd have a very sluggish cursor on 60Hz displays.
+		// In case we have vsync enabled, we should rely on vsync to do take care about frame times.
+		// With vsync enabled, we currently have to force a frame time of 1ms since otherwise
+		// CPU usage will skyrocket on one thread as soon as no updateScreen(); calls happening.
 		if (g_system->getFeatureState(OSystem::kFeatureVSync)) {
 			_system->delayMillis(1);
 		} else {
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -547,10 +547,16 @@ void GuiManager::runLoop() {
 
 		redraw();
 
-		// Delay until the allocated frame time is elapsed to match the target frame rate
-		uint32 actualFrameDuration = _system->getMillis(true) - frameStartTime;
-		if (actualFrameDuration < targetFrameDuration) {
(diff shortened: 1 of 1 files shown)


gui/gui-manager.cpp
```diff
@@ -547,10 +547,16 @@ void GuiManager::runLoop() {
...
 		redraw();
 
-		// Delay until the allocated frame time is elapsed to match the target frame rate
-		uint32 actualFrameDuration = _system->getMillis(true) - frameStartTime;
-		if (actualFrameDuration < targetFrameDuration) {
-			_system->delayMillis(targetFrameDuration - actualFrameDuration);
+		// Delay until the allocated frame time is elapsed to match the target frame rate.
+		// In case we have vsync enabled, we force a minimum frame duration of 1 millisecond
+		// since otherwise, we'd have a very sluggish cursor on 60Hz displays.
+		if (ConfMan.getBool("vsync")) {
```
aquadran, line 553: do not ask for config, but ask backend. since vsync might be actually not enabled

lotharsm: Thanks! I hope to have properly adressed this in f628d7b.

gui/gui-manager.cpp
```diff
@@ -547,10 +547,16 @@ void GuiManager::runLoop() {
...
 
-		// Delay until the allocated frame time is elapsed to match the target frame rate
-		uint32 actualFrameDuration = _system->getMillis(true) - frameStartTime;
-		if (actualFrameDuration < targetFrameDuration) {
-			_system->delayMillis(targetFrameDuration - actualFrameDuration);
+		// Delay until the allocated frame time is elapsed to match the target frame rate.
+		// In case we have vsync enabled, we force a minimum frame duration of 1 millisecond
+		// since otherwise, we'd have a very sluggish cursor on 60Hz displays.
+		if (g_system->getFeatureState(OSystem::kFeatureVSync)) {
+			_system->delayMillis(1);
```
aquadran, line 554: why add 1ms? vsync will limit

lotharsm: In case I don't set the additional 1ms, CPU utilization of one core will stay at 100% all the time with vsync enabled. Not entirely sure why this happens though :/

aquadran: 1ms is needed in case no screen changes and OpenGL backend will not call GL swap buffers (with vsync)

aquadran: can you add comments about why limit to 1ms?

sev-:
Thanks!

PR #3535, opened 2021-11-14
GUI: fix inclusion guard for `AlphaAnimation.h`

benoit-pierre:


bluegr:
Nice catch, thanks! Merging

PR #3537, opened 2021-11-14
XEEN: fix "backpack is full" dialog string

benoit-pierre:
Use the destination character's name.


dreammaster:
Thanks for the fix. Merging

PR #3540, opened 2021-11-15
DIRECTOR: allocate more than 2 audio channels for earlier Director versions

mistydemeo:
I've found a Director 3 game that tries to play back audio on channel 3, which currently leads to an assertion failure - see #3539. I'm not entirely clear how it was determined that older Director versions had only two channels; looking at the commit history D4 was given 4 channels in 88d8577cb3a1634bd23ead85ccfec8375d293aed, but it's not really clear to me if it was determined for sure that pre-D4 should only have two channels.

cc @moralrecordings for thoughts, since you bumped the number of channels for D4.

### DIFF
--- a/engines/director/score.cpp
+++ b/engines/director/score.cpp
@@ -752,7 +752,7 @@ void Score::playSoundChannel(uint16 frameId) {
 	}
 
 	// Channels above 2 are only usable by Lingo.
-	if (g_director->getVersion() >= 400) {
+	if (g_director->getVersion() >= 300) {
 		sound->playPuppetSound(3);
 		sound->playPuppetSound(4);
 	}
--- a/engines/director/sound.cpp
+++ b/engines/director/sound.cpp
@@ -44,7 +44,7 @@ namespace Director {
 
 DirectorSound::DirectorSound(Window *window) : _window(window) {
 	uint numChannels = 2;
-	if (g_director->getVersion() >= 400) {
+	if (g_director->getVersion() >= 300) {
 		numChannels = 4;
 	}
 


djsrv:
There are two ways for Director to play sounds - with the score and with Lingo.

The score can only control sound channels 1 and 2.

In D2-3, Lingo commands for sounds don't allow you to specify which channel to play in, and IIRC they always default to channel 1. In Director 4, Macromedia added a channel argument to some of the sound commands, and they can control channels 3 and 4 without interfering with the score.

Unless I'm missing something, there is *no way* for a movie to interact with sound channels 3 and 4 before Director 4, if they existed at all. If a D2-3 movie is somehow doing that, I'm inclined to believe our sound code has an off-by-one error or something similar. Unless we can confirm there is in fact a legitimate way to use channels 2-4 before Director 4, this should not be merged.

If there is a legitimate way for movies to interact with sound channels 3 and 4, we'd also need to adjust the code that actually plays the sounds in those channels: https://github.com/scummvm/scummvm/blob/02cfffcb674037acabb549456080b46a1855ad2b/engines/director/score.cpp#L754-L758

mistydemeo:
I dug up the source in question. Looks like the original Lingo source is specifying `3` in this case:

```
PlayAiffFile(3,"P07")
```

Where `PlayAiffFile` is a macro that dispatches to `playFile` with the channel number unaltered. From checking, looks like it uses the numbers `1` and `3` in the source, but not `2`.

mistydemeo:
Updated the commit to alter the check in `score.cpp` to check for D3 instead of D4, and restored the original check except made it so the old behaviour works on D2. In testing, I confirmed that Oz gains the ability to play back voices with these two checks.

djsrv:
Great, thanks for confirming.

PR #3541, opened 2021-11-15
GOB: Detect Hebrew fan translation of Gobliins 2

orgads:


orgads:
Wait before merging. I need to verify that the translation is considered as done.

orgads:
The translator (@BLooperZ)  approved this. In case there will be a revised version, we will modify the detection entry.

PR #3543, opened 2021-11-16
AGS: Detection for all the missing RON games

tag2015:
* Detection for all the missing Reality-on-the-Norm games
* Detection for a couple recent games
* Some minor fixes


sev-:
This must be first, rebased, so all those merge commits go away, and then, the commit log messages fixed in accordance with our commit guidelines: https://wiki.scummvm.org/index.php?title=Commit_Guidelines

PR #3544, opened 2021-11-16
GUI: Fix file name in icons set for gui-icons.dat

orgads:


sev-:
Thanks, but I have a different fix in an unpushed commit.

PR #3547, opened 2021-11-17
DIRECTOR: handle method end statements

mistydemeo:
This change was suggested by @djsrv. It turns out that methods can have multiple optional `end` statements, which seem to carry no semantic meaning. One game I was looking at used methods written like:

```
method name
    # etc
end method
```

Which was leading to a parse error.

This, alongside #3545, is one of the three fixes needed to get Difficult Book Game/henachoco03 to full working status.


sev-:
Thank you!

PR #3549, opened 2021-11-17
NGI: Fix memory leak in PictureObject

mparnaudeau:
The default destructor did not freed the picture data belonging
to the PictureObject instances, causing many memory leaks.

PR note: I controlled before and after with Valgrind.


bluegr:
Nice catch, thanks! Merging

PR #3552, opened 2021-11-19
KYRA: Don't require the Mac version of Legend of Kyrandia to be installed

eriktorbjorn:
The Macintosh version of The Legend of Kyrandia that I have is a non-talkie version distributed on CD. This version is supported now that arthxx added a music driver for it.

Annoyingly though, the game data files aren't unpacked on the CD. Instead, everything is inside a self-extracting StuffIt installer. This means that in order to play the game in ScummVM, you have to install the game (probably using a Mac emulator) and then copy the files from the emulated hard disk. Not the smooth kind of experience we want people to have.

This is my attempt at adding support for playing the game directly from the files on the CD, without having to go through the installation process. This is the first time I've worked with the Archive class, but hopefully I got it right.

Obviously it needs more testing. I've only tried it a little bit past the intro. But it appears to work.

### DIFF
--- a/common/stuffit.cpp
+++ b/common/stuffit.cpp
@@ -43,6 +43,7 @@ public:
 	~StuffItArchive() override;
 
 	bool open(const Common::String &filename);
+	bool open(Common::SeekableReadStream *stream);
 	void close();
 	bool isOpen() const { return _stream != nullptr; }
 
@@ -90,9 +91,14 @@ static const uint32 s_magicNumbers[] = {
 };
 
 bool StuffItArchive::open(const Common::String &filename) {
+	Common::SeekableReadStream *stream = SearchMan.createReadStreamForMember(filename);
+	return open(stream);
+}
+
+bool StuffItArchive::open(Common::SeekableReadStream *stream) {
 	close();
 
-	_stream = SearchMan.createReadStreamForMember(filename);
+	_stream = stream;
 
 	if (!_stream)
 		return false;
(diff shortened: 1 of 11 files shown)


athrxx:
Very cool to have that feature for the Mac version.
Also, very clean and unobstrusive implementation.
I'll do a quick test asap and then we can merge this...

eriktorbjorn:
Thank you. The only improvement I'm considering right now is if the music code should ask the engine for a pointer to the StuffIt archive, instead of creating its own.

eriktorbjorn:
Also, I was never quite sure of the difference between _files and _protectedFiles etc, so let me know if anything needs to be fixed in that respect.

eriktorbjorn:
The very fact that it's reading Mac resources from a file inside a StuffIt archive that it got from the data fork of a Macintosh executable, and it all somehow seems to work, is enough to make my head spin. :-)

eriktorbjorn:
That's the last bit of cleanup that I've been meaning to make. I'll try to give it a bit of testing (at least up to entering Timbermist Wood) tomorrow. If that works, it's probably good enough to merge. (According to arthxx, "If it starts up fine with sound it should actually be okay. There isn't really anything happening with regard to resource management afterwards.")

eriktorbjorn:
The very fact that it's reading Mac resources from a file inside a StuffIt archive that it got from the data fork of a Macintosh executable, and it all somehow seems to work, is enough to make my head spin. :-)

eriktorbjorn:
I've played both the extracted and unextracted version up to the point where you meet Darm in the Timbermist Wood. I didn't notice any issues related to this pull request.

PR #3553, opened 2021-11-19
AGS: Detection for RON games + minor fixes

tag2015:
AGS: Detection for RON games + minor fixes

* Detection for Reality-on-the-Norm games
* Detection for some recent games
* Minor naming fixes


tag2015:
Third time's the charm... should be fine now. I figured out how to keep my fork updated avoiding all the merge commits.
Sorry for the inconvenience

sev-:
Thank you! I checked all entries, all looks good.

Though, after looking at the duplicates, I think that the Tail MD5s could help. @dreammaster thoughts?

PR #3554, opened 2021-11-20
DIRECTOR: add stub handler for some events

mistydemeo:
These events are handled by the primary handler above, and in D2/D3 they don't have other handlers. However, not being handled at all in the switch statement means it's falling through to the default case below and printing spurious "unhandled event" logs.


sev-:
Thank you!

PR #3556, opened 2021-11-21
BASE: Allow search of shaders in extra

lephilousophe:
Shaders are searched using a 'shaders' subdirectory.
When adding the extra path, if we don't specify a depth of 2 the shaders
can never been found.


lephilousophe:
Already fixed

PR #3559, opened 2021-11-22
Allow the checking of other audio tracks than number 1

eriktorbjorn:
ScummVM currently warns that the FM Towns SCUMM demos (indyzak and zakloom) are missing the audio tracks. That's because we always check for track 1, and neither demo has that.

This could be extended further, if we want to, e.g. if we want to verify that _all_ tracks are present instead of just one. But I don't know if that's worth bothering with?


bluegr:
Nice work! Clean and simple. Thanks, merging

PR #3561, opened 2021-11-23
ENGINE: Suppress autosave overwrite warning when description is empty

orgads:
Some engines (like kyra) return a SaveStateDescriptor with slot assigned,
even when the slot is free. The indication for an empty save is empty
description on these cases.

### DIFF
--- a/engines/engine.cpp
+++ b/engines/engine.cpp
@@ -567,9 +567,7 @@ void Engine::handleAutoSave() {
 bool Engine::warnBeforeOverwritingAutosave() {
 	SaveStateDescriptor desc = getMetaEngine()->querySaveMetaInfos(
 		_targetName.c_str(), getAutosaveSlot());
-	if (desc.getSaveSlot() == -1)
-		return true;
-	if (desc.hasAutosaveName())
+	if (!desc.isValid() || desc.hasAutosaveName())
 		return true;
 	Common::U32StringArray altButtons;
 	altButtons.push_back(_("Overwrite"));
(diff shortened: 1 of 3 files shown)


athrxx:
Thanks for updating this.

However, the issue that this is difficult to understand still remains: It is not obvious to anyone else who looks at that code for the first time why a slot of -1 or an empty slot name would mean that there is no file.

The comment for `int getSaveSlot() const { return _slot; }` says: 
`@return The saveslot id, as it would be passed to the "-x" command line switch.`
There is no indication that a value of -1 would have a special meaning. Same for the empty description name.

Why not add a property `_fileExists` or similiar and a corresponding query function to SaveStateDescriptor? (unless you don't have the time of course :-) )

orgads:
You looked at the wrong function. Search for `getAutosaveSlot` in engines/metaengine.h.

orgads:
https://github.com/scummvm/scummvm/blob/master/engines/metaengine.h#L307

athrxx:
Okay, that function (`MetaEngine::getAutosaveSlot`) is supposed to return -1 if autosaving is not supported. I did see that. But here (from `Engine::warnBeforeOverwritingAutosave()` your check is: `desc.getSaveSlot() == -1`. That's the method from SaveStateDescriptor. Also, "autosaving not being supported" and "autosaving being supported, but a certain save file not existing" are two different things.

Just to clarify: I am not against merging the PR. I would just prefer clearing this up a bit... But that might happen (or not) at a later point...

orgads:
Better now?

athrxx:
Yes :-)

bluegr:
Looks good to me now as well

PR #3562, opened 2021-11-23
AGS: Small tweaks detection macros + new games detection

tag2015:
AGS: Small tweaks detection macros + new games detection

* Small change to macros to append -demo flag to gameid
* Detection for many games (from ags website)
* Minor typo fixes


sev-:
Thank you!

PR #3563, opened 2021-11-23
BUILD: Support SerenityOS in configure script

GMTA:
As discussed here: https://github.com/SerenityOS/serenity/commit/59b2bac3a53a5b7605fc85d10b307147f1702635

This adds SerenityOS as a valid POSIX target for ScummVM. ScummVM is supported as a port in SerenityOS, and this change allows us to remove the only patch needed to get everything to compile.

CC @sev-


sev-:
Thank you :)

PR #3564, opened 2021-11-26
clearing screen and redrawing to prevent repeated drawing on initial screen

Itachi8374:
Before:
image

After:
Screenshot 2021-11-27 at 1 41 10 AM

### DIFF
--- a/engines/bbvs/minigames/bbloogie.cpp
+++ b/engines/bbvs/minigames/bbloogie.cpp
@@ -230,6 +230,11 @@ void MinigameBbLoogie::drawSprites() {
 	buildDrawList(drawList);
 	_vm->_screen->drawDrawList(drawList, _spriteModule);
 	_vm->_screen->copyToScreen();
+	_vm->_screen->clear();
+    DrawList drawList_test;
+    buildDrawList0(drawList_test);
+    _vm->_screen->drawDrawList(drawList_test, _spriteModule);
+    _vm->_screen->copyToScreen();
 }
 
 void MinigameBbLoogie::initObjs() {


aquadran:
Reword the commit log messages according to our Commit Guidelines, particularly, so the all are more descriptive (maybe squash some of them) and have a prefix "DEVTOOLS: COMPANION: ....."

henke37:
The new code seems to make the preceding code redundant. Should the preceding code be removed in favor of the new code?
Also `drawList_test` is not a good name for code intended to be permanent.

Itachi8374:
@aquadran will this be a good commit message: "ENGINE: bbvs: Fixed redrawing on initial screen in minigame"? 
I read the code guidelines but missed to read commit guidelines

Itachi8374:
@henke37 No, the preceding code is not redundant. There are two things which needs to be drawn, the background and objects. Current implementation keeps on drawing on background without clearing what was drawn before. ```_vm->screen->clear()``` only clears the background and leaves the object, for eg the man, cycle, squirrel thus it does not make preceding code redundant.

Maybe the following workflow explains it better:
```
void MinigameBbLoogie::drawSprites() {
    DrawList drawListBackground;
    buildDrawList0(drawListBackground);
    _vm->_screen->drawDrawList(drawListBackground, _spriteModule);
    _vm->_screen->copyToScreen();
    DrawList drawList;
    buildDrawList(drawList);
    _vm->_screen->drawDrawList(drawList, _spriteModule);
    _vm->_screen->copyToScreen();
    _vm->_screen->clear();
}
```

criezy:
> @aquadran will this be a good commit message: "ENGINE: bbvs: Fixed redrawing on initial screen in minigame"?
> I read the code guidelines but missed to read commit guidelines

No need to have both `engine` and `bbvs` prefix. In this case `BBVS:` is sufficient.
When in doubt about the prefix to use you can have look at the history of commits : https://github.com/scummvm/scummvm/commits/master/engines/bbvs
For recent commits you can see some with the `BBVS:` prefix and others with the `ENGINES:` prefix. The later is used for commits that are related to the `Engine` class and touch multiple engines.

criezy (changes requested):
I took a quick look at the issue, and I think the fix is not correct.

Initially I used the demo from our website (https://www.scummvm.org/demos), and it was interesting to see that that one actually works properly. The one attached to the bug (https://bugs.scummvm.org/ticket/13070) is actually a different demo. And if you look in `engines/bbvs/detection.cpp` you will see that there are indeed two different demos.

And if you run both demos, the first thing you will notice is that the one from the bug report seems to not use the correct sprite for the title screen and instead use the one from the game. And I think that is a very strong hint as to what needs to be changed. If I am correct the faulty code is this in `MinigameBbLoogie::run()`:
```c++
if (_vm->_gameDescription->flags & ADGF_DEMO) {
	_backgroundSpriteIndex = 209;
	_titleScreenSpriteIndex = 210;
} else {
	_backgroundSpriteIndex = 210;
	_titleScreenSpriteIndex = 211;
}
```
A quick hack to use the non-demo sprites for that demo seemed to indeed fix the issue.
But a cleaner fix is needed. The approach I would suggest is to:
1. Add one more flag to the `BBVSGameFeatures` enum to differentiate the two demos.
2. Add that flag to one of the two demos in the detection entries.
3. In `MinigameBbLoogie::run()` you can check for that flag to use the correct set of sprites.

bluegr:
Thanks for the suggestion @criezy, and thanks for your work @Itachi8374.
@criezy's suggestion has been implemented in 29bccc4ab45cc71a49f6b3b22c805f76d2af5f78. Closing this

Itachi8374:
@bluegr Hi, I was trying to implement the suggestion but it wasn't working. I was doing everything you did except change on line:1313 in bbloogie.cpp where you added another condition check. Could you explain that change, please. Thanks

Itachi8374:
Also in detection.cpp, instead of replacing one of the flags like you did, I added the extra flag in this way ``` GF_LOOGIE_DEMO | ADGF_DEMO | ADGF_DROPPLATFORM | GF_LOOGIE_ALT_DEMO ```

criezy:
> Also in detection.cpp, instead of replacing one of the flags like you did, I added the extra flag in this way `GF_LOOGIE_DEMO | ADGF_DEMO | ADGF_DROPPLATFORM | GF_LOOGIE_ALT_DEMO`

Both approaches are possible. What you did was actually what I had in mind.

> I was doing everything you did except change on line:1313 in bbloogie.cpp where you added another condition check. Could you explain that change, please.

That's actually related to the point above. Since @bluegr replaced the `GF_LOOGIE_DEMO` flag with the `GF_LOOGIE_ALT_DEMO` for one of the demo rather than have both, this means that `isLoogieDemo()` is now true for only one of the demo and he needed to add `isLoogieAltDemo()` to the check to catch both demos.

With your approach of using both the `GF_LOOGIE_DEMO` and `GF_LOOGIE_ALT_DEMO` flags for the alt demo, the change on line 1313 would not have been needed.

PR #3565, opened 2021-11-27
Director apply henachoco03 patches to demo

mistydemeo:
This applies the henachoco03 patches to the demo. The same ambiguous syntax occurs there, but on different line numbers and in different filenames. The original patch was added in #3545.

I've also added a second commit that fixes the file sizes for the demos, since they were added before the Mac file size bug was fixed.


bluegr:
Thanks!

PR #3566, opened 2021-11-27
DIRECTOR: add second file to All the Lunar

mistydemeo:
This adds a second file to All the Lunar to fix the ambiguity in its `install.exe` name.


sev-:
Excellent, thank you!

PR #3567, opened 2021-11-28
SHERLOCK: Fix missing Rose Tattoo talk animations (bug #7090)

eriktorbjorn:
This is an attempt at fixing https://bugs.scummvm.org/ticket/7090 ("SHERLOCK: Watson isn't animated while talking"). If I understand it correctly, the problem is simply that we compare the wrong strings when looking up the talking actor, and therefore he's not animated.

I don't know if this affects any other actors than Watson.

If this is the correct fix, I'd suggest we apply it to the 2.5 branch as well.

### DIFF
--- a/engines/sherlock/people.cpp
+++ b/engines/sherlock/people.cpp
@@ -261,10 +261,8 @@ int People::findSpeaker(int speaker) {
 		Object &obj = scene._bgShapes[idx];
 
 		if (obj._type == ACTIVE_BG_SHAPE) {
-			Common::String name(obj._name.c_str(), obj._name.c_str() + 4);
-
-			if (name.equalsIgnoreCase(portrait)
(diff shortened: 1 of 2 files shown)


dreammaster:
I compared to the original, and you're right, it should be using npc_name. However, one other minor thing I noticed was that it was doing a strnicmp(name,NPC_name[x],4), where name is the first 4 digits of the portrait. Which is different from what our code does, which is compare to the entirety of portrait. I'd like to keep everything in the one commit, so if you'd like you could amend your commit, or otherwise, I'll make the correct change tomorrow night and then close this PR.

eriktorbjorn:
You mean like this?

dreammaster:
Yep. That looks better than my original half-hearted attempt. Merging and cherry-picking to 2.5 branch now.

PR #3568, opened 2021-11-28
AGS: Add detection for new games + naming fixes

tag2015:
AGS: Add detection for new games + naming fixes

* Add detection for some games using AGS website as reference
* Improved some titles


sev-:
Thank you. I am squashing your commits into one

PR #3570, opened 2021-11-28
ULTIMA: Fix pointer alignment

lephilousophe:
Multiple fixes for ULTIMA.
2 failures to build on latest Xcode due to unaligned access to members.


lephilousophe:
No warning has been added in MSVC builds in Github actions because of this PR.
It seems the structure packing is useless when fields are properly ordered.

dreammaster:
Seems okay to me; looking at the current codebase, I don't see anything remaining that relies on it being packed

PR #3573, opened 2021-11-28
XEEN: DE: Initial support for german version

dmarks2:
XEEN: DE: Initial support for german version

This pull request adds initial support for the german versions
of Clouds of Xeen and Darkside of Xeen. Some very few texts
remain untranslated. Either they were not translated in the
original game or the text were not found in the original
game.

### DIFF
--- a/devtools/create_xeen/constants.cpp
+++ b/devtools/create_xeen/constants.cpp
@@ -28,6 +28,7 @@
 
 #include "en_constants.h"
 #include "ru_constants.h"
+#include "de_constants.h"
 
 const char *const LangConstants::TERRAIN_TYPES[6] = {
 	"town", "cave", "towr", "cstl", "dung", "scfi"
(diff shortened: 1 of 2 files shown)


dreammaster:
Look okay to me. Good work so far

dmarks2:
@dreammaster sorry to come back to this PR... I just downloaded the latest ScummVM 2.7.1, but when I start Xeen it is still in english, even I use the german game files. Was I missing something at the time of this PR?

dreammaster:
Hmm.. you're right. It looks like German support may have been accidentally lost when preliminary support for a Russian translation was added. I've committed a fix that should resolve the problem, though I don't have the German game handy right now. If you could test in the next daily build, and see how it goes..

dmarks2:
@dreammaster Thanks a lot, I tested the daily build and German support for Xeen is working now!

PR #3575, opened 2021-11-29
Cherry-pick commits from orgads to fix the check for overwriting

criezy:
This cherry-picks to branch-2-5 commits from @orgads to fix the check for overwriting savegames in the autosave slot and moving the savegame to a different slots.

I think it would be a good idea to merge those to branch-2-5 as they fix annoying bugs with the check that autosave are not overwriting normal saves. But since they changes a lot of files in a lot of engines, I prefer to propose this as a pull request to check that others agree that they are good to merge.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@ For a more comprehensive changelog of the latest experimental code, see:
  General:
    - Ported ResidualVM GUI theme to remastered version.
    - Fixed edge case for Punycode.
+   - Fixed checking for savegame overwrite in autosave slot.
+   - Fixed moving savegame to new slot for most engines.
 
  AGOS:
    - Fixed old Waxworks AdLib music regression.
--- a/engines/access/metaengine.cpp
+++ b/engines/access/metaengine.cpp
@@ -135,7 +135,7 @@ SaveStateList AccessMetaEngine::listSaves(const char *target) const {
 
 			if (in) {
 				if (Access::AccessEngine::readSavegameHeader(in, header))
-					saveList.push_back(SaveStateDescriptor(slot, header._saveName));
+					saveList.push_back(SaveStateDescriptor(this, slot, header._saveName));
 
 				delete in;
 			}
@@ -170,7 +170,7 @@ SaveStateDescriptor AccessMetaEngine::querySaveMetaInfos(const char *target, int
 		delete f;
 
 		// Create the return descriptor
-		SaveStateDescriptor desc(slot, header._saveName);
+		SaveStateDescriptor desc(this, slot, header._saveName);
 		desc.setThumbnail(header._thumbnail);
 		desc.setSaveDate(header._year, header._month, header._day);
 		desc.setSaveTime(header._hour, header._minute);
--- a/engines/adl/metaengine.cpp
+++ b/engines/adl/metaengine.cpp
@@ -102,7 +102,6 @@ bool AdlMetaEngine::hasFeature(MetaEngineFeature f) const {
 	case kSavesSupportThumbnail:
 	case kSavesSupportCreationDate:
 	case kSavesSupportPlayTime:
-	case kSimpleSavesNames:
 		return true;
 	default:
 		return false;
--- a/engines/adl/metaengine.cpp
+++ b/engines/adl/metaengine.cpp
@@ -136,7 +136,7 @@ SaveStateDescriptor AdlMetaEngine::querySaveMetaInfos(const char *target, int sl
 		return SaveStateDescriptor();
 	}
 
-	SaveStateDescriptor sd(slot, name);
+	SaveStateDescriptor sd(this, slot, name);
 
 	int year = inFile->readUint16BE();
 	int month = inFile->readByte();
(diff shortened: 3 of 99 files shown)


macca8:
If this PR is approved, could we please restore Save/Delete functionality to the Autosave slot?

Originally set up to prevent the user from saving a non-autosave saved game in the Autosave slot, with these improvements, and the introduction of Warning messages when these saved games are present, these saved games are now (finally) being handled properly, making these user restrictions redundant.

This would allow the user to:
- manually update the Autosave file in situations where an autosave isn't usually performed, in particular, when exiting a game.
- clear the Autosave slot manually, without disabling autosaving, providing backup to the methods associated with the Warning messages.

Generally, the slot will be occupied by either the Autosave file or a dummy "Autosave" entry, so it's unlikely a user would deliberately create a non-autosave saved game in the slot, but even if they did, the Warning message would allow the situation to be resolved quickly and efficiently.

Some engines already have this capability (e.g. Riven, Myst 3), and I'm seeing no interference with regular autosaving.

The isWriteProtected & isDeletable flag settings have already been removed from most engines, and are now set from within engines/savestate.cpp/SaveStateDescriptor::initSaveType(), so hopefully changing the status of these flags to "false" & "true" respectively isn't a major undertaking.

If necessary, I'll post this as a feature request, but it would be nice to get some sort of feedback in case there's something I've overlooked. Thank you.

orgads:
I think it's better to just activate auto save on quit.

Enabling Delete is ok by me, but what for?

macca8:
> I think it's better to just activate auto save on quit.

Really depends on whether a prompt-to-save feature is activated. With the addition of the confirm exit option, all engines have the capability of prompting on exit, either by using the global option or their own always-on native dialog... autosave-on-exit was removed from Myst 3 because of this duplication.
The advantage of enabling Save is that it gives the user the choice of where and when to save, and not just on exit. The Autsave file is just another saved game, and regular autosaves are just a bonus along the way. There's no risk, so why would you want to deny the user this flexibility?

> Enabling Delete is ok by me, but what for?

Currently, the minimum steps to manually remove a saved game from the Autosave slot (regardless of why) is to exit the game, disable autosaving, open the Launcher's Load dialog, delete the saved game, enable autosaving, then relaunch the game.
May never be required, but so much simpler to just open an in-game Save/Load dialog and click the Delete button.

orgads:
The problem with that is that the user can save a game in the autosave slot (and not even rename it), then it will be overwritten the next time without any warning (unless it was renamed), and the user can lose a precious savegame.

Regarding Delete, I don't have a strong opinion.

aquadran:
is this PR wait for anything?

criezy:
> is this PR wait for anything?

I proposed this as a pull request in case others wanted to check it was safe for the 2-5 branch as there are a lot of changes and I didn't have time to review everything.
I should be able to find some time tomorrow to finish that check and merge.

sluicebox:
I want to point out that SCI is still suffering regressions from the auto-save changes. I'm mentioning it here because I don't want them to slip through the cracks. I've mentioned some of this in Discord but I haven't seen any recent momentum.

SCI doesn't use ScummVM's auto-save. SCI does use slot 0 for games that have their own in-game auto-saving mechanism. (QFG3, QFG4, and others.) That is, these game's scripts are automatically saving the game as they did in the original. We detect that and route them to slot 0. This has worked well for years. But the new auto-save changes conflate the two concepts: in-game vs external (ScummVM).

1. Disabling auto-save causes ScummVM to list the wrong save slots when displaying the load dialog from launcher when there's a save in slot 0. Slot 0 is displayed as empty and the auto-save that really is in slot 0 is displayed as slot 1, and then everyone else is off by one. (See `MetaEngine::listSaves`)

2. Re-enabling auto-save triggers the new message box / feature about moving around all everything in slot 0 and applies that to SCI where it makes no sense. This tells users that there are problems which don't exist and, depending on their answer, promotes in-game auto saves to regular save slots I don't know anything about this feature, I just know that it's wrong to apply it to SCI.

From my perspective, it seems like there needs some way for an engine to say "I don't want any part of this and hands off my slot 0."

orgads:
There is a way for that. Return -1 in `getAutosaveSlot` (in both the engine and meta-engine).

I'll try to look into these issues. This is the first time I see them.

orgads:
@sluicebox Are these regressions from one of my commits in this PR, or do they occur also in 2.5?

orgads:
Is #3581 good enough, or do you prefer to completely disable autosave?

I tried it with SQ5, and it did not save automatically at all (even with autosave enabled).

orgads:
Pushed #3582 for disabling it altogether.

criezy:
I did a few tests with QFG4.
The issue with the overwrite check when enabling autosave in the GUI options also happens in the 2.5 release. So it is not something introduced by commits in this pull request. And I think it is probably fixed in pull request #3582 (although I don't know if this would have other side effects). We could consider merging that change as well in the branch-2-5.

The issue with the slot shift appears to be different for me in branch and master. So it is possible that one of the commit in this pull request would change the behaviour. Neither seems correct to me though:
 - In 2.5 branch it doesn't matter if autosave is enabled or not in the GUI. It always lists the SCI autosave in slot 1 and lists an non-existent autosave in slot 0 (it probably tries to reserve the slot, but this should only happen in the save dialog, not when loading).
 - In master the wrong behaviour indeed only happens when autosave is disabled in the ScummVM options. At least in master the list is correct when autosave is enabled.

So the SCI issues are not directly related to the commits in this pull request (although the behaviour might be improved a bit by some of the included changes). I will merge this and let SCI devs review pull requests #3581 and #3582 and decide if they should also be backported to the 2.5 branch.

PR #3576, opened 2021-11-29
XEEN: sound related improvements

benoit-pierre:
* remove some dead code
* fix some sounds not being correctly muted when Efx is off (e.g. vendor voices, some combat FXs)
* fix missing music when loading a save from the main menu
* fix toggling music back on
* fix sound settings not getting saved

### DIFF
--- a/engines/xeen/map.cpp
+++ b/engines/xeen/map.cpp
@@ -901,8 +901,7 @@ void Map::load(int mapId) {
 		} else {
 			musName = "outdoors.m";
 		}
-		if (musName != sound._currentMusic)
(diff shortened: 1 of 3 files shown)


dreammaster:
Good work on all the fixes. I'm accepting this PR now and will manually merge the fixes into 2.5 as well

dreammaster:
At this point, I think you're deserving of being mentioned in the Xeen engine credits for all your work if you'd like. If so, if you could provide your name (if you're willing) so I can add it along with your handle.

benoit-pierre:
Sure, thanks, use: "Benoit Pierre".

PR #3577, opened 2021-11-30
SHERLOCK: Workaround for bad doorbell sound in Serrated Scalpel

eriktorbjorn:
This fixes the doorbell sound at Lord Brumwell's mansion so that it's played at 11000 Hz instead of 1100 Hz. This seems to match the behavior of the 3DO version pretty well, though since that's not supported by ScummVM I've only seen YouTube playthroughs.

This fixes https://bugs.scummvm.org/ticket/10838


dreammaster:
Workaround looks good to me

PR #3578, opened 2021-11-30
SHERLOCK: Fix Rose Tattoo glitch when opening map

eriktorbjorn:
This attempts to fix the bug described in https://bugs.scummvm.org/ticket/10850

Since the map always snaps to its saved position, there's no need to show the upper left corner for a split second first.


dreammaster:
Seems straightforward. Accepting

PR #3581, opened 2021-12-04
SCI: Do not add autosave slot in listSaves

orgads:
It is done in MetaEngine::listSaves only for save. For SCI it was done also for load, which is wrong.

### DIFF
--- a/engines/sci/metaengine.cpp
+++ b/engines/sci/metaengine.cpp
@@ -341,7 +341,6 @@ SaveStateList SciMetaEngine::listSaves(const char *target) const {
 	filenames = saveFileMan->listSavefiles(pattern);
 
 	SaveStateList saveList;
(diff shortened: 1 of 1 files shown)


bluegr:
Is there a reason why we can't use the listSaves() function from MetaEngine in SCI?

bluegr:
@orgads any feedback on this?

orgads:
Depends on #3582. If you want to keep an autosave slot even when autosave is disabled, we'll need to override listSaves. We can call it from the superclass, and only add the autosave slot though.

bluegr:
After checking this with @sluicebox, we need the autosave slot to be unavailable when saving. I'm closing this PR.

As an enhancement, we could call listSaves() from  the superclass and then add protection for the autosave slot

PR #3582, opened 2021-12-04
SCI: Disable autosave

orgads:
SciEngine::canSaveGameStateCurrently() returns false unconditionally, so the feature doesn't work anyway.


orgads:
@bluegr / @sluicebox I don't remember if we had this issue settled. What do you say?

bluegr:
Update: added the gist above in fb9a96d0f8002f108b2e6c0d690849aba80b2fcb, still needs some work because music isn't resumed properly upon game load

bluegr:
Saving via the GMM in SCI has been implemented in d8336a31ffbf12cfc4eb8687e5bb83e2dffa4964, therefore autosave should also work correctly now in the games that this feature is supported.

Closing this

PR #3589, opened 2021-12-05
SCUMM: DiMUSE: Fix bug #13080

AndywinXp:
For [this bug](https://bugs.scummvm.org/ticket/13080) to be fixed, I needed to avoid setting an arbitrary `_maxQueuedStreams` value in `IMuseDigital::tracksCallback()`, and instead do a proper calculation based on the audio buffer size.

To do that I needed the audio backend to pass the buffer size along to the mixer (with which DiMUSE can communicate). This has been done for all the ports/platform I could be able to test. For the others I left a fallback value of 4 for the maximum number of queued streams (until someone can test all the other platforms).

The changes in this PR have also been tested with various manually inserted values for `output_rate` (22050, 44100, 48000) and `audio_buffer_size` (from 512 up to 8192) in the `scummvm.ini` config files, and I can happily report that everything works as it should both in DiMUSE and in a bunch of other games and engines which I didn't touch (just for good measure).

Unfortunately Android works in its own ways, and reports a much higher latency than the real one, resulting in a very high `_maxQueuedStreams`  value (and therefore high latency). I have solved this by getting the Android backend to record the actual audio buffer size, which I retrieved from the AudioTrack object _after_ its creation. But this fix will be put in a separate PR, just for safety.

### DIFF
--- a/audio/mixer_intern.h
+++ b/audio/mixer_intern.h
@@ -65,6 +65,7 @@ private:
 	Common::Mutex _mutex;
 
 	const uint _sampleRate;
+	const uint _outBufSize;
 	bool _mixerReady;
 	uint32 _handleSeed;
 
@@ -81,7 +82,7 @@ private:
 
 public:
 
-	MixerImpl(uint sampleRate);
+	MixerImpl(uint sampleRate, uint outBufSize = 0);
 	~MixerImpl();
 
 	virtual bool isReady() const { Common::StackLock lock(_mutex); return _mixerReady; }
@@ -129,6 +130,7 @@ public:
 	virtual int getVolumeForSoundType(SoundType type) const;
 
 	virtual uint getOutputRate() const;
+	virtual uint getOutputBufSize() const;
 
 protected:
 	void insertChannel(SoundHandle *handle, Channel *chan);
--- a/audio/mixer.cpp
+++ b/audio/mixer.cpp
@@ -177,10 +177,11 @@ private:
 #pragma mark --- Mixer ---
 #pragma mark -
 
-MixerImpl::MixerImpl(uint sampleRate)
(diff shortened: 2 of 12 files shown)


audio/mixer_intern.h
```diff
@@ -81,7 +82,7 @@ class MixerImpl : public Mixer {
 
 public:
 
-	MixerImpl(uint sampleRate);
+	MixerImpl(uint sampleRate, uint outBufSize = -1);
```
aquadran: assign -1 to uint is good idea. 0 is not enough?

aquadran (changes requested):

audio/mixer_intern.h
AndywinXp: Whoops, it should have been an int, not a unit.
But I suppose you are right, I can't think of a reason why -1 might be better than 0 as a special value.

aquadran:
can you squash last and "AUDIO: Set _outBufSize to signed int" commits?

AndywinXp:
Thanks, done

audio/mixer_intern.h
```diff
@@ -82,7 +82,7 @@ class MixerImpl : public Mixer {
 
 public:
 
-	MixerImpl(uint sampleRate, uint outBufSize = -1);
+	MixerImpl(uint sampleRate, uint outBufSize = 0);
```
aquadran, line 85: can you squash this change to earlier commit which introduce -1

AndywinXp: Sure, I'll look into it and do it

AndywinXp: There you go, should be much more cleaner than before

aquadran (approved):

PR #3590, opened 2021-12-05
ANDROID: Keep track of actual AudioTrack buffer size

AndywinXp:
This PR is complementary to [https://github.com/scummvm/scummvm/pull/3589](this one), except in this one I make a small but still sensitive change which doesn't appear to negatively affect other games or engine, but it's still a separate PR for good measure.

The same change has to be done for ANDROID3D, but:

- I admit I don't know which is the minimum SDK version being supported for that;
- I'd better be doing that after being done with the audio work on Grim Fandango.


aquadran:
ANDROID3D is deprecated

AndywinXp:
Thanks, that's good to know.

mduggan:
This calculation was a little bit wrong, because the buffer size variable should be in bytes but the frames are 16 bit stereo so it needs to be multiplied by 4 to be correct.  I pushed a fix in 466c8b8.

PR #3592, opened 2021-12-06
ANDROID: Fix softkeyboard button for Chromebooks

mduggan:
On a Chromebook, the keyboard may be present but hidden (not available for
use) because the device is flipped around in Tablet mode.  In this case we want
the soft keyboard to be available.  This change also hides the soft keyboard
button if the hardware keyboard is available, since it doesn't do anything.

Fixes #13138.


lephilousophe:
Seems OK although I couldn't test it with a Chromebook.

PR #3594, opened 2021-12-07
GUI: tabs with scrollbars and move game settings out of Engine tab

Die4Ever:
image


gui/widgets/tab.cpp
```diff
@@ -122,6 +121,7 @@ int TabWidget::addTab(const Common::U32String &title, const Common::String &dial
 	newTab.title = title;
```
Die4Ever, line 121: we might want to add a bool withScroll argument to addTab, so you can create tabs without the automatic ScrollContainerWidget, or else we will have to fix all the old tabs that used scrollbars like GameOptions_KeyMapper

I do want to figure out why the nested scrollbars aren't working properly before I do either of those above things

gui/object.cpp
```diff
@@ -59,6 +59,12 @@ void GuiObject::resize(int x, int y, int w, int h, bool scale) {
 	}
 }
 
+Widget *GuiObject::addChild(Widget *newChild) {
```
Die4Ever, line 62: maybe we should call setNext here instead of trusting the caller (Widget::init) to handle that?

gui/widget.cpp
```diff
@@ -51,9 +51,7 @@ Widget::Widget(GuiObject *boss, const Common::String &name, const Common::U32Str
 }
 
 void Widget::init() {
-	// Insert into the widget list of the boss
-	_next = _boss->_firstWidget;
-	_boss->_firstWidget = this;
+	_next = _boss->addChild(this);
```
Die4Ever, line 54: maybe the addChild function should call setNext on us, instead of doing it here?

gui/widget.cpp
```diff
@@ -119,10 +118,8 @@ void Widget::draw() {
 				_x = _x + g_gui.getOverlayOffset();
 			}
 
-			Common::Rect r = _boss->getClipRect();
-			r.moveTo(_x, r.top);
-
-			g_gui.theme()->swapClipRect(r);
+			clip.moveTo(_x, clip.top);
+			g_gui.theme()->swapClipRect(clip);
```
Die4Ever, line 122: I'm not entirely sure about this change, but it appears to work with Hebrew and English, and both the old and new style scrollcontainers in tabs

gui/widget.cpp
bluegr: Could be done as a future enhancement

gui/widget.cpp
bluegr: This code is used in RTL languages. Since it works with Hebrew, it should be fine

bluegr:
Nice work overall! Moving the game options to the game tab does make more sense indeed. I've initially added them in a separate tab, because back then we didn't have the functionality we got now in the GUI.

Overall, the code is quite extensible and reusable, and the functionality will hopefully help us for future additions as well.

Thanks again! Merging

PR #3596, opened 2021-12-08
SHERLOCK: Fix missing haze when going to Cleopatra's Needle

eriktorbjorn:
In The Case of the Rose Tattoo, certain rooms have a hard-coded haze. This is loaded in TattooEngine::startScene(), but this was bypassed if the room was loaded through the overhead map. (I think that only happens for Cleopatra's Needle.) This pull request fixes that by handling the map case first.

If this is the correct fix (and I've convinced myself that this is likely), it should probably be fixed in 2.5 as well.


dreammaster:
Looks good to me. Thanks yet again. Merging it in, and I'll copy to 2.5 branch.

PR #3598, opened 2021-12-08
SCUMM: Rewrite player_v3a

quietust:
Back when I first wrote `player_v3a` many, many years ago, I based it on a loose understanding of a poorly-labeled disassembly, and its behavior was only loosely related to the original engines - music was allowed to use as many sound channels as it wanted (with a cap of 24), sounds could play on top of music, and some things were just plain done wrong.

With the aid of newer tools (namely, Ghidra), I've been able to much more closely analyze the sound engines in both Indy3 and Loom (which were actually significantly different from each other in several places) and rewrite `player_v3a` to function exactly as the original games did. In particular, music now correctly plays in stereo and is limited to 4 channels, and sound effects are now _also_ played in stereo (i.e. some play at different frequencies in each ear).

Note that this technically makes some things sound "worse" compared to my original code, but they're _supposed_ to sound that way, and I personally think accuracy is more important. That, and my old code sucked and I wanted to make it better.

A similar rewrite of `player_v2a` is in the works, though it might be a long time before it's ready.

### DIFF
--- a/audio/mods/paula.cpp
+++ b/audio/mods/paula.cpp
@@ -81,6 +81,7 @@ void Paula::clearVoice(byte voice) {
 	_voice[voice].volume = 0;
 	_voice[voice].offset = Offset(0);
 	_voice[voice].dmaCount = 0;
+	_voice[voice].interrupt = false;
 }
 
 int Paula::readBuffer(int16 *buffer, const int numSamples) {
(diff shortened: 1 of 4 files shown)


sev-:
This is awesome! My first thought was "could we keep the old one as well", but in this case, I think the accuracy is more important.

Merging.

PR #3602, opened 2021-12-10
SHERLOCK: Fix looking at room objects while inventory is open in Serrated Scalpel

eriktorbjorn:
This pull request aims to fix two long-standing glitches in The Case of the Serrated Scalpel, both happening when examining objects in the room while the inventory is open.

- When the description window was closed, the inventory would show only empty slots. This appears to be because our putInv() function referred to _backBuffer1 directly, instead of using the current back buffer (which in this case was back buffer 2).
- When examining a room object, only the first page of the description was shown. This bug happens in the original as well, and the easiest way to test this is to examine the violin in the first room of the game. Hopefully fixing this will not cause any unwanted side effects.


PaulGilbert:
I manually cherry picked the commits so I could fix the incorrect formatting of one of the commit messages.

PR #3606, opened 2021-12-11
SCUMM: Show subtitles on talkie versions without sound file

orgads:
If the game is a talkie version, on which the verbs reference sounds, but it is missing the sound file, and the user chose Voice Only, a warning appears for every playback attempt.

Force subtitles on this case, to avoid these warnings.

Fixes [trac 13151](https://bugs.scummvm.org/ticket/13151)

### DIFF
--- a/engines/scumm/scumm.cpp
+++ b/engines/scumm/scumm.cpp
@@ -1487,8 +1487,12 @@ Common::Error ScummEngine::init() {
 	resetScumm();
 	resetScummVars();
 
-	if (_game.version >= 5 && _game.version <= 7)
+	if (_game.version >= 5 && _game.version <= 7) {
 		_sound->setupSound();
+		// In case of talkie edition without sfx file, enable subtitles
+		if (!_sound->hasSfxFile() && !ConfMan.getBool("subtitles"))
+			ConfMan.setBool("subtitles", true);
+	}
 
 	syncSoundSettings();
 
--- a/engines/scumm/sound.cpp
+++ b/engines/scumm/sound.cpp
@@ -1002,6 +1002,11 @@ bool Sound::isSfxFileCompressed() {
 	return !(_soundMode == kVOCMode);
 }
 
+bool Sound::hasSfxFile() const
+{
+	return !_sfxFilename.empty();
+}
+
 ScummFile *Sound::restoreDiMUSESpeechFile(const char *fileName) {
 	Common::ScopedPtr<ScummFile> file;
 	file.reset(new ScummFile());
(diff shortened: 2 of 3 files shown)


orgads:
I'm not sure about this change. It might break other games. Please review carefully.

orgads:
@athrxx

sluicebox:
Just to clarify, although a review was requested from me, I don't know anything about the SCUMM engine and don't have anything to contribute here. My "involvement" is that my GUI fix altered undefined behavior *in certain sitautions*  a SCUMM game is run without all of its necessary resource files.  See: https://bugs.scummvm.org/ticket/13151

But since I've been tagged... =)

@orgads Your PRs need explanations. You complained on discord that 7 weeks have gone by without anyone approving this, but you haven't made this an easy PR to approve. Other than "this fixes bug #_____", there's no justification that someone can evaluate.  What may be obvious in code to you isn't always the case to others, especially on a volunteer project where people need to re-familiarize with context first. My PR strategy is to provide so much information that it's as difficult as possible for anyone to reject the PR, and that has served me well.

sluicebox:
After some swimming, I've come up with a hopefully better way to put this:

SCUMM and SCI are in the same boat: ancient ScummVM engines. As time goes on, the bar keeps getting raised for justifying changes. After ~20 years, the bar is pretty high! How can it be that all these years have passed without needing your alteration? This is how I think when making SCI changes.

In this case, I think it would be worth mentioning that you're addressing the scenario where a user is *missing* necessary resource files. Is that normal? Allowed? Why are we putting effort towards improving the experience for known broken setups? I don't know the answer to any of these questions, but they're all examples of why no one has approved this PR yet.

athrxx:
Since I also got tagged I can say that I agree to everything sluicebox wrote. Also, I am very reluctant to go near a PR (and possibly take the blame for it) if the author himself says "I'm not sure about this change. It might break other games".

And no, we usually don't provide fallbacks for incorrect game setups. It isn't per se forbidden to do that, but we'd rather throw a warning or an error about the missing file.

orgads:
Thanks, I appreciate your comments and will try to improve my PRs.

Regarding my previous comment (I'm not sure...) it was on the first revision (5a9c886). I then found a less intrusive solution and pushed it.

I'll improve the commit message and will re-push.

orgads:
Thanks, I appreciate your comments and will try to improve my PRs.

Regarding my previous comment (I'm not sure...) it was on the first revision (5a9c886). I then found a less intrusive solution and pushed it.

I'll improve the commit message and will re-push.

bluegr:
With the latest changes, this functionality is very straightforward and cleanly written.

Thanks for your work! This is good to be merged now

PR #3612, opened 2021-12-14
SHERLOCK: Disable autosave

orgads:
When using original game dialogs for save/load, the user can save on slot
0, and the when autosave is triggered, a warning appears.

### DIFF
--- a/engines/sherlock/metaengine.cpp
+++ b/engines/sherlock/metaengine.cpp
@@ -82,6 +82,11 @@ public:
 	 * Given a specified savegame slot, returns extended information for the save
 	 */
 	SaveStateDescriptor querySaveMetaInfos(const char *target, int slot) const override;
+
+	/**
+	 * Disable autosave. It conflicts with built-in save dialog, which doesn't block the slot.
+	 */
+	int getAutosaveSlot() const override { return -1; }
 };
 
 Common::Error SherlockMetaEngine::createInstance(OSystem *syst, Engine **engine, const ADGameDescription *desc) const {
--- a/engines/sherlock/sherlock.h
+++ b/engines/sherlock/sherlock.h
@@ -166,6 +166,11 @@ public:
 	void syncSoundSettings() override;
 
 	/**
+	 * Disable autosave
+	 */
+	int getAutosaveSlot() const override { return -1; }
+
+	/**
 	 * Saves game configuration information
 	 */
 	virtual void saveConfig();


dreammaster:
I'm still in favor of only doing this as a last resort, particularly since the original save/load dialog is disabled by default. As I see, it would be preferable to do one of two things:
1) Remap the savegame range to start at 1 rather than 0, this would prevent slot 0 from being used. It would mean that the autosave slot couldn't be loaded original dialog, which I think is fine. But I looked at the code and it could end up requiring a bunch of changes in multiple places to handle both mouse and keyboard selections.
2) Keep the current slot 0 being the first, but prevent them from saving to slot 0 from within the dialog. This would mean disabling the Save button if the first slot (on the first page) is selected first, or if the save button is selected first, ignore any attempts to select the autosave slot.

orgads:
@dreammaster Can you please do this?

dreammaster:
I've done some initial experimentation on both points. Point 1 proved difficult to do, since there were a great many places I need to identify, both in the retrieval of the savegame list, paging, and then selecting/description entering and calling the saving/loading code. The second point would be in theory a bit easier, but I still wasn't happy with my experiments. 

I may return to work on them sometime, but honestly, I don't see this as much of a priority. Firstly, it only happens if a user uses the original save/load dialog/pane, and secondly, I'm not entirely sure what the problem is with giving an autosave warning. Particularly given if autosaves are turned on, then chances are an autosave would already be created by the time the user opens the save dialog. And if not, showing a warning would be legitimate, wouldn't it?

eriktorbjorn:
> Particularly given if autosaves are turned on, then chances are an autosave would already be created by the time the user opens the save dialog.

I made my first save at the scene of the murder, so that was probably long before it tried to autosave. Also, I didn't get the warning very often, which makes me believe that autosaves were infrequent even though the interval was set to 5 minutes. From what I understand, if it can't make a savegame it will wait five minutes before giving it another try. If I understand correctly, Sherlock only allows auto-saves in "standard mode" or at the lab table, i.e. probably not while looking at things, talking to characters, using your inventory, etc.

Maybe if autosaving in general tried to make a first autosave as early as possible (if there isn't already a savegame in that slot)? Then it would be much harder to use that slot by accident.

dreammaster:
That's a good idea, to manually trigger an autosave when the game first starts.

dreammaster:
Pushed a commit to do so

PR #3619, opened 2021-12-16
RFC: SCUMM: Replay text when loading a saved game with active text

orgads:


orgads:
Doesn't work well. If I save on Elaine's 3rd phrase on MI2 (You do turn up in the strangest places), it replays back from the beginning, and the lines are wrapped wrong.

athrxx:
Is this something that one of the original games actually does? Or is it supposed to be an extra feature for ScummVM?

orgads:
It's an extra feature. I checked MI2 on DOS, and it's even worse than ScummVM. :)

The text disappears on *save*, and of course it is not displayed on load (the speech animation plays, but there is no text).

orgads:
It sort of works now, but the text layout is not the same as it appeared before the save.

I'm even *less* confident about this change (mostly because I'm not familiar with the control codes, and it's likely that I missed something). Added RFC to the subject.

What do you say?

PR #3620, opened 2021-12-16
SCI: Patch to fix duplication of text with subtitles

darksoul42:
On QFG1 for PC-98, the previous code would cause undue duplication of text
when using the following languge configuration :
- Japanese as main language
- Subtitles enabled (setting subtitle language to English)

Output would end up looking like :
```
日本語 (Japanese)
----------
日本語 (Japanese)
English
----------
English
```

On every dialog box, including the input box.

This fix has been tested for :
- Quest for Glory I (PC-98/Japanese)
- Space Quest III (Amiga/German)

(Ctrl+L to configure languages, attempt to type a "look" command)

For QFG1 for PC-98, the behavior therefore becomes identical
to what can be observed when running the game on a PC-98 emulator.

This fix is valid for any game relying on kStrSplit(),
and even though the following game has another mechanism
for handling multiple language strings,
it has also been confirmed to be regression free on the first screen :
- Police Quest II (PC-98/Japanese)

### DIFF
--- a/engines/sci/engine/state.cpp
+++ b/engines/sci/engine/state.cpp
@@ -373,7 +373,8 @@ Common::String SciEngine::strSplitLanguage(const char *str, uint16 *languageSpli
 		return retval;
 
 	// Add subtitle, unless the subtitle language doesn't match the languages in the string
-	if ((subtitleLanguage == K_LANG_ENGLISH) || (subtitleLanguage == foundLanguage)) {
+	// Confirmed to work for Amiga SQ3 German/English, QFG1 Japanese/English, PQ2 Japanese/English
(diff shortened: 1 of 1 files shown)


sluicebox:
Thanks! And thank you again for the PC-98 screenshots of how things are supposed to look.

I've fixed this in 46cdbecbc6c4c546f496ff6bae03ea6ea8ef6d35 by fixing `SciEngine::getSciLanguageString` to always initialize its `secondaryLanguage` output parameter. That failure led to undefined behavior in `SciEngine::strSplitLanguage` which affected QFG1 Japanese. (And probably would have been a problem for any future callers.) The game scripts handle subtitles in Japanese games differently than the other dual language games which rely on the kernel (ScummVM) inserting the separator string. This fixes the immediate unambiguous implementation error without altering any of the working logic.

PR #3623, opened 2021-12-19
SCI: Text to Speech (TTS) functionality for SCI floppy games

bluegr:
This is based on the work done for GSoC by @taylorzhancher in PR #3256

Highlights:
- TTS has been hooked globally onto places where text is shown
- TTS is currently performed for textboxes and button texts
- TTS stops when text windows are disposed (which enhances the in-game
  experience)
- No game-specific logic has been added
- This hasn't been extensively tested with all SCI games yet
- There will be cases that are not handled properly yet
- The TTS functions have been grouped under a SciTTS class

TTS has been added to all SCI games without speech.
Testing on this hasn't been thorough, but all games where the
functionality has been hooked seem to be behaving nicely.
More adjustments may be needed.


bluegr:
After some initial testing and reviewing this together with @sluicebox, it's in a good state to be merged. Any game-specific fixes/workarounds that may be needed can be done in-tree

PR #3625, opened 2021-12-21
(WIP) SCUMM: DiMUSE: Implement music stream filling while loading resources

AndywinXp:
Hiya, this PR aims to implement the one thing missing from the original engine (admittedly because I couldn't understand its purpose until recently...).

There's a function in DiMUSE, `diMUSEProcessStreams()`, which is responsible for asking the engine to fill the audio streams (and consequently, the data buffers to be processed) up to a certain point; this function is called at every SCUMM engine loop, ensuring smooth audio playback.

That is, until SCUMM has to load a generic resource (graphics, sound, etc.), a speech file, or even a SMUSH movie: this process could take an indefinite amount of time, depending on the hardware you're playing on, therefore the main engine could not get to the `diMUSEProcessStreams()` call in time, and could leave the audio streams empty for a time which is small enough not to be game-breaking, but big enough to cause some major audio hiccups on some hardware configurations.

I have implemented what the original interpreter does in this case (for all the three variations: FT, DIG & COMI + demos) which is, stupidly enough, blast a lot of `diMUSEProcessStreams()` calls whenever a resource has to be loaded, a speech file has to be played, and a SMUSH movie has to be loaded. And sure enough, it works! (This has been tested on my underpowered-on-purpose Windows and Linux VMs, and on my Android phone which suffered the most from this issue)

There's a catch: doing so, exposed some mutex-related bugs which I'm really glad I found. 
Long story short: having two independent mutex objects (_resourceAccessMutex in `scumm.h`, and _mutex in `dimuse_engine.h`) seemed to work fine being that we previously only had accesses of this kind between the two relevant threads (main and audio callback):

> Do various audio stuff: *lock(_mutex)* ---> Use the resource manager: *lock(_resourceAccessMutex)*

...but: it actually didn't work so fine to begin with, since some very edge case deadlocks could be triggered even without the feature implemented above (I stumbled into them only recently, sorry).

With the new feature in the mix, the situation on both thread becomes the following:

> Do various audio stuff: *lock(_mutex)* ---> Use the resource manager: *lock(_resourceAccessMutex)*---> Process the streams: *lock(_mutex)*

This is highly simplified but I hope you get the idea that having two separate mutexes both being constantly locked and unlocked on two separate threads, was just me asking for trouble.

To solve this I passed the _resourceAccessMutex (which is the most external object, belonging to the SCUMM class) along to the DiMUSE engine, so that every access is regulated under a single mutex object.

I'm currently marking this PR as a WIP: I'd like your opinion on how I solved the mutex problem (because I'm pretty certain I'm doing it in a non-optimal way), and I want to test all three games + demos from scratch to make sure that this new feature works and doesn't crash anything since it will also be featured on Grim Fandango.

### DIFF
--- a/engines/scumm/imuse_digi/dimuse_dispatch.cpp
+++ b/engines/scumm/imuse_digi/dimuse_dispatch.cpp
@@ -830,7 +830,7 @@ void IMuseDigital::dispatchProcessDispatches(IMuseDigiTrack *trackPtr, int feedS
 }
 
 void IMuseDigital::dispatchPredictFirstStream() {
-	Common::StackLock lock(_mutex);
+	Common::StackLock lock(*_mutex);
 
 	for (int i = 0; i < _trackCount; i++) {
 		if (_dispatches[i].trackPtr->soundId && _dispatches[i].streamPtr && _dispatches[i].streamZoneList)
--- a/engines/scumm/imuse_digi/dimuse_engine.cpp
+++ b/engines/scumm/imuse_digi/dimuse_engine.cpp
@@ -191,6 +191,8 @@ int IMuseDigital::startVoice(int soundId, const char *soundName, byte speakingAc
 		if (fileDoesNotExist)
 			return 1;
 
+		fillStreamsWhileMusicCritical(5);
+
 		diMUSEStartStream(kTalkSoundID, 127, DIMUSE_BUFFER_SPEECH);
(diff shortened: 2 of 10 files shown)


AndywinXp:
Hmm, on second thought, after a good night's sleep and lots of testing, I think I'll hold on to this for the moment: in some cases it does more harm than good (like randomly deadlocking when idling for several minutes). 
The idea is there, but the execution is rather poor at the moment. I hope I can revisit this feature soon, but it needs much more work than this.

PR #3628, opened 2021-12-25
ANDROID: Add touch 3D mode to 2D backend

lephilousophe:
These changes allow:
- use of touch 3D mode emulating a D-Pad to be used in 2D games
- the user to switch mode at will

To make this happen, a change in OpenGL backend is necessary to allow drawing of clipped textures.
The interface of Android 3D backend textures has been slightly reworked to avoid exposing internal functions.
Finally, the image used to draw the touch controls has been converted to PNG and the Android runtime is used to load it instead of doing it using our code. This make loading the PNG not depending on PNG support in ScummVM code.


antoniou79:
Looks good to me. 
Tested with Grim Fandango, TLJ , Myst 3 and LBA (not 3d), Secret of Monkey Island on Android arm64-v8a (Android 10) and on an arm-v7a (Android 4.4.4). It works ok. 
Some issues to consider or fix later on (not merge stoppers):
- on LBA switching to the joystick control mode, the overlay icons lack transparency. Maybe this has to do with LBA being a non-3d engine? 
- Would it make sense to have the overlay icons as SVG?

lephilousophe:
> * on LBA switching to the joystick control mode, the overlay icons lack transparency. Maybe this has to do with LBA being a non-3d engine?
I have seen this bug too and was wondering why.
I finally found the reason: the blend mode had to be configured.
For this, I need to make the back buffer object accessible to android graphics backend.

PR #3631, opened 2021-12-27
DIRECTOR: PR review

mistydemeo:
This contains changes requested by @trembyle on #3627 after it was merged.


engines/director/detection_tables.h
```diff
@@ -671,10 +671,10 @@ static const PlainGameDescriptor directorGames[] = {
 
 	// French titles
 	{ "1contre100",			"1 contre 100" },
+	{ "abcpacovska",		"Alphabet: Play with the ABCs"},
```
trembyle, line 674: Can you please move this up to the English language titles?

trembyle (changes requested):
This leaves Just the one small detail on `abcpacovska`. Also, I had suggested on the other PR to move `venetiandeer` to English.

Even though this is basically a revision, it does need to follow the commit guidelines; i.e., it needs a descriptive summary and some detail on what is changed. Something like "adjust titles", or "add second files to detection" would be fine.

https://wiki.scummvm.org/index.php/Commit_Guidelines

mistydemeo:
Made the requested changes.

PR #3634, opened 2021-12-28
SCUMM: DiMUSE: Add a debugger/player and fix oversights

AndywinXp:
Ahoy!
This PR:

1. Adds a debugger for DiMUSE: you can now play and stop SFX/music, change parameters and check various internal information (that is, for all SCUMM v7-8 games);
2. Adds a missing sequence table entry for FT (it actually paid to implement this debugger); no matter how you can triple check a thing, you will always miss something :-)
3. Moves some DiMUSE specific code from SCUMM to DiMUSE itself, and removes a redundant function call to `diMUSEProcessStreams()` (now it is coherent with the interpreters).

EDIT: I forgot to say, the debugger can't currently tamper with speech files because in order to play those you would need the filename; needless to say, there's too many of those and I can't simply print a list on the console :-) I'll have to think of a clever way to do that

### DIFF
--- a/engines/scumm/debugger.cpp
+++ b/engines/scumm/debugger.cpp
@@ -29,6 +29,7 @@
 #include "scumm/boxes.h"
 #include "scumm/debugger.h"
 #include "scumm/imuse/imuse.h"
+#include "scumm/imuse_digi/dimuse_engine.h"
 #include "scumm/object.h"
 #include "scumm/resource.h"
 #include "scumm/scumm.h"
@@ -93,7 +94,12 @@ ScummDebugger::ScummDebugger(ScummEngine *s)
 	registerCmd("show",      WRAP_METHOD(ScummDebugger, Cmd_Show));
 	registerCmd("hide",      WRAP_METHOD(ScummDebugger, Cmd_Hide));
 
-	registerCmd("imuse",     WRAP_METHOD(ScummDebugger, Cmd_IMuse));
+	if (_vm->_game.version < 7)
+		registerCmd("imuse", WRAP_METHOD(ScummDebugger, Cmd_IMuse));
+#if defined(ENABLE_SCUMM_7_8)
+	else
+		registerCmd("imuse", WRAP_METHOD(ScummDebugger, Cmd_DiMuse));
+#endif
 
 	registerCmd("resetcursors",    WRAP_METHOD(ScummDebugger, Cmd_ResetCursors));
 }
@@ -112,6 +118,15 @@ void ScummDebugger::postEnter() {
 	// Boot params often need debugging switched on to work
 	if (_vm->_bootParam)
 		_vm->_debugMode = true;
+}
+
+void ScummDebugger::onFrame() {
+	Debugger::onFrame();
+#if defined(ENABLE_SCUMM_7_8)
+	if (_vm->_imuseDigital && !_vm->_imuseDigital->isEngineDisabled()) {
+		_vm->_imuseDigital->refreshScripts();
+	}
+#endif
 }
 
 ///////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


engines/scumm/debugger.cpp
```diff
@@ -93,7 +94,15 @@ ScummDebugger::ScummDebugger(ScummEngine *s)
 	registerCmd("show",      WRAP_METHOD(ScummDebugger, Cmd_Show));
 	registerCmd("hide",      WRAP_METHOD(ScummDebugger, Cmd_Hide));
 
-	registerCmd("imuse",     WRAP_METHOD(ScummDebugger, Cmd_IMuse));
+#if defined(ENABLE_SCUMM_7_8)
```
aquadran, line 97: this looks like duplication for "imuse"

engines/scumm/debugger.cpp
```diff
@@ -175,6 +193,180 @@ bool ScummDebugger::Cmd_IMuse(int argc, const char **argv) {
...
 
+#if defined(ENABLE_SCUMM_7_8)
+bool ScummDebugger::Cmd_DiMuse(int argc, const char **argv) {
+	if (!_vm->_imuseDigital || _vm->_imuseDigital->isEngineDisabled()) {
+		debugPrintf("No Digital iMUSE engine is active.\n");
+		return true;
+	}
+
+	if (argc > 1) {
+		if (!strcmp(argv[1], "panic")) {
```
aquadran, line 204: maybe some better name instead "panic"?

engines/scumm/debugger.cpp
AndywinXp: Should I just keep "imuse" as the command name for tidiness?

engines/scumm/imuse_digi/dimuse_engine.cpp
```diff
@@ -800,4 +800,97 @@ int IMuseDigital::diMUSESetAttribute(int attrIndex, int attrVal) {
...
+			for (int i = 0; _comiStateMusicTable[i].soundId != -1; i++) {
+				_vm->getDebugger()->debugPrintf("|  %4d   | %20s  |\n", _comiStateMusicTable[i].soundId, _comiStateMusicTable[i].name);
+			}
+		}
+	} else if (_vm->_game.id == GID_DIG) {
+		for (int i = 0; _digStateMusicTable[i].soundId != -1; i++) {
+			_vm->getDebugger()->debugPrintf("|  %4d   | %20s  |\n", _digStateMusicTable[i].soundId, _digStateMusicTable[i].name);
+		}
+	} else if (_vm->_game.id == GID_FT) {
+		for (int i = 0; i < 48; i++) {
```
aquadran, line 825: array size maybe?

engines/scumm/debugger.cpp
AndywinXp: "panic" appears to be a fairly common command name in the audio software domain to denote "stop everything" commands (e.g. MIDI panic).
I can ditch that command and instead make people use "stop all" instead of creating another alias as "stopAll".

engines/scumm/debugger.cpp
aquadran: might be good idea, since they are exclude each other

AndywinXp: Yep, good point thanks, will do that

engines/scumm/imuse_digi/dimuse_engine.cpp
AndywinXp: Whoops, you're right

engines/scumm/debugger.cpp
aquadran: please do

engines/scumm/imuse_digi/dimuse_engine.cpp
AndywinXp: I have added a 
```c++
{"ARR_END",  0,  0,    ""                    },
```
at the end of `_ftStateMusicTable` (and also changed `_ftSeqNames` accordingly), so looping through the states/sequences now looks like this:
```c++
} else if (_vm->_game.id == GID_FT) {
    for (int i = 0; strcmp(_ftStateMusicTable[i].audioName, "ARR_END"); i++) {
	  _vm->getDebugger()->debugPrintf("|  %4d   | %21s |\n", i, _ftStateMusicTable[i].name);
    }
}
```

is it ok?

aquadran: nullptr will not work?

AndywinXp: No, unfortunately it won't let me put a nullptr as the marker

aquadran: or add check for _ftStateMusicTable[i].name if != ""?

AndywinXp: Either one of these works fine:

```c++
		for (int i = 0; strcmp(_ftStateMusicTable[i].name, ""); i++) {
			_vm->getDebugger()->debugPrintf("|  %4d   | %21s |\n", i, _ftStateMusicTable[i].name);
		}
```

```c++
		for (int i = 0; _ftStateMusicTable[i].name[0]; i++) {
			_vm->getDebugger()->debugPrintf("|  %4d   | %21s |\n", i, _ftStateMusicTable[i].name);
		}
```

(by having a `	{"",         0,  0,    ""                    },` row on the state table)

Which one would you prefer?

aquadran: second one

AndywinXp:
Okay, I fixed the debugger and squashed the fix onto the main debugger commit, thanks for the review!

engines/scumm/debugger.cpp
aquadran: it was rather unnecessary one call, this could looks like this:
if (_vm->_game.version < 7)
registerCmd("imuse", WRAP_METHOD(ScummDebugger, Cmd_IMuse));
#if defined(ENABLE_SCUMM_7_8)
else
registerCmd("imuse", WRAP_METHOD(ScummDebugger, Cmd_DiMuse));
#endif

AndywinXp: Very good, done

PR #3638, opened 2022-01-02
SCUMM: Fix CMS not playing a music midi track again

antoniou79:
Possible fix for bug #10696

Reason for not playing the Scumm Bar music again, after entering the kitchen, seemed to be that the _loadedMidiSong var did not clear its value, despite the music being stopped when entering the kitchen. Same issue would happen exiting the bar from the external door and re-entering; Same issue seems to be with church (mentioned in the bug ticket).


bluegr:
Quite a straightforward change. Thanks! Merging

PR #3639, opened 2022-01-03
CGE/CGE2: Don't pass the engine object to EncryptedStream

lephilousophe:
It doesn't need it and when linked with detection and with UBSan enabled build fails because CGEEngine and CGE2Engine aren't linked in


bluegr:
Nice work!
That's quite a straightforward and logical change, there's no real need for a review here, so I'm merging this

PR #3641, opened 2022-01-03
CONFIGURE: added --enable-ubsan to enable the undefined behavior sanitizer

neuromancer:
This new configure flag allows to use the [undefined behavior sanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) easily.


bluegr:
Quite a straightforward implementation and a very useful feature.

Thanks! Merging

PR #3642, opened 2022-01-03
GUI: Fix various memory leaks related to grid widget

lephilousophe:
These were found with ASan.
I am not sure why `themepath` and `iconspath` handling are different.

### DIFF
--- a/gui/gui-manager.cpp
+++ b/gui/gui-manager.cpp
@@ -114,7 +114,7 @@ struct ArchiveMemberListBackComparator {
 	}
 };
 void GuiManager::initIconsSet() {
-	Common::Archive *dat = nullptr;
+	Common::Archive *dat;
 
 	_iconsSet.clear();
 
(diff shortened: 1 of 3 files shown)


bluegr:
Looks good! Well done

bluegr (approved):

gui/gui-manager.cpp
```diff
@@ -138,35 +138,37 @@ void GuiManager::initIconsSet() {
...
 	if (ConfMan.hasKey("themepath")) {
 		Common::FSNode *fs = new Common::FSNode(normalizePath(ConfMan.get("themepath") + "/" + fname, '/'));
-
-		if (!fs->exists()) {
-			delete fs;
-		} else {
+		if (fs->exists()) {
 			dat = Common::makeZipArchive(*fs);
 		}
+		delete fs;
```
bluegr, line 150: I don't understand this change. This is handled internally when ZipFile is destroyed.

lephilousophe: No this is not because we are passing a reference to FSNode not a pointer to a stream.
The called prototype is this one: `Archive *makeZipArchive(const FSNode &node);` which does this:
```
return makeZipArchive(node.createReadStream());
```

bluegr: Thanks!

bluegr:
Thanks for your work. I've tried these changes locally, and everything works as expected. Merging

PR #3644, opened 2022-01-04
AGI: Expand unsupported Mac detection

trembyle:
This one needs some help from @sev- or someone who has access to the disks in the existing Mac detection entries, as my files usually didn't match what was already in the tables. Should I be extracting the data and resource forks separately, or can we use MacBinary? If the latter, should we add `ADGF_MACRESFORK`?

I have matching version numbers and dates but different md5 for KQ2, KQ3, SQ1, SQ2. Additionally, my copy of SQ1 has separate interpreters for black & white and 16 colors. Despite matching version/date, neither of them match what's already in detection.

I have earlier version numbers for Gold Rush and PQ1 but later dates than the entries in detection. These could potentially be explained if the earlier versions have been patched to add color support.

And these are probably okay:
- 3 new entries for LSL1, all with later version/dates than the existing entry
- Early black & white version of KQ1. The existing entry was v2.0C, mostly likely in color.
- Entries for both Manhunters, neither of which were previously in detection

### DIFF
--- a/engines/agi/detection_tables.h
+++ b/engines/agi/detection_tables.h
@@ -114,6 +114,22 @@ namespace Agi {
 		ver \
 	}
 
+#define GAME_LVFPNSt(id,extra,fname,md5,size,lang,ver,features,gid,platform,interp,guioptions) { \
+		{ \
+			id, \
+			extra, \
+			AD_ENTRY1s(fname,md5,size), \
+			lang, \
+			platform, \
+			ADGF_UNSTABLE, \
+			guioptions \
+		}, \
+		gid, \
+		interp, \
+		features, \
+		ver \
+	}
+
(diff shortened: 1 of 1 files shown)


einstein95:
Looking at KQ2, the only data files that have resource forks are `Palette` and `ListMgr.rsrc`. Keeping only the data forks of the other files lets it pass detection.

trembyle:
@einstein95 has brought it to my attention that at least some Mac games do start, albeit without their Mac menus and icon bars. I'll test these further and mark as unstable if they do.

Also, with the confirmation that discrete resource / data forks are required, I'll redump my files and see if that clears up any of the md5 differences.

trembyle:
Upon further review, I've confirmed that I am able to detect and start my Mac files via the existing Mac detection entries. In many cases, I think that this is too permissive, since I probably do have different versions for some games.

But since there is future work to be done to support Mac-specific features, like the Mac GUI menus and icon bars, I'll leave the code mostly as is for now. I'll close the pull request and commit minor changes directly.

PR #3645, opened 2022-01-04
ALL: Rework OpenGL detection

lephilousophe:
With the advent of GLAD, OpenGL libraries don't need to be linked with ScummVM for those platforms.
In addition, GLAD support is known at configure step by knowing the platform to configure.
This PR also tries to merge two configure paths: the one for 2D engines, and the one for 3D engines.

It builds correctly on Linux and Android. No other tests have been done yet.
In addition, create_project still has a USE_GLES2 define and no USE_GLES_MODE.

### DIFF
--- a/backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
+++ b/backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
@@ -83,7 +83,7 @@ OpenGLSdlGraphics3dManager::OpenGLSdlGraphics3dManager(SdlEventSource *eventSour
 		DEFAULT_GLES2_MINOR = 0
 	};
 
-#ifdef USE_GLES2
+#if USE_FORCED_GLES2
 	_glContextType = OpenGL::kOGLContextGLES2;
 	_glContextProfileMask = SDL_GL_CONTEXT_PROFILE_ES;
 	_glContextMajor = DEFAULT_GLES2_MAJOR;
@@ -581,7 +581,7 @@ void OpenGLSdlGraphics3dManager::drawOverlay() {
 
 #if !defined(__amigaos4__) && !defined(__MORPHOS__)
 OpenGL::FrameBuffer *OpenGLSdlGraphics3dManager::createFramebuffer(uint width, uint height) {
-#if !defined(USE_GLES2)
+#if !USE_FORCED_GLES2
 	if (_antialiasing && OpenGLContext.framebufferObjectMultisampleSupported) {
 		return new OpenGL::MultiSampleFrameBuffer(width, height, _antialiasing);
 	} else
--- a/configure
+++ b/configure
@@ -159,10 +159,8 @@ _faad=auto
 _fluidsynth=auto
 _fluidlite=auto
 _opengl_mode=auto
-_opengl_game=auto
 _opengl_game_classic=auto
 _opengl_game_shaders=auto
-_opengl_game_es2=no
 _tinygl=yes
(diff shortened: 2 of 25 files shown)


configure
```diff
@@ -950,8 +948,8 @@ Optional Libraries:
 
   --with-opengl-prefix=DIR      prefix where OpenGL (ES) is installed (optional)
   --disable-opengl-game         disable OpenGL (ES) support in game [autodetect]
-  --disable-opengl-game-shaders disable use of modern OpenGL with shaders
-  --force-opengl-game-es2       only detect OpenGL ES2
+  --disable-opengl-game-classic disable OpenGL (ES) classic support in game
```
aquadran, line 951: "OpenGL (ES) classic" I think it should be "OpenGL classic". on another thought potentially we can support gles1, but we are not ready yet.

backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
```diff
@@ -579,7 +579,7 @@ void OpenGLSdlGraphics3dManager::drawOverlay() {
 
 #if !defined(__amigaos4__) && !defined(__MORPHOS__)
 OpenGL::FrameBuffer *OpenGLSdlGraphics3dManager::createFramebuffer(uint width, uint height) {
-#if !defined(USE_GLES2)
+#if !USE_FORCED_GLES2
```
aquadran, line 582: I wonder what prevents us to support multisample support for gles2. it's capable in general

configure
lephilousophe: ah yes, you are right, that was bad copy paste

backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
lephilousophe: I don't know, I adapted the changes without thinking about it.

graphics/opengl/system_headers.h
```diff
@@ -24,35 +24,69 @@
...
-#if !defined(GL_UNPACK_ROW_LENGTH)
-// The Android SDK does not declare GL_UNPACK_ROW_LENGTH_EXT
-#define GL_UNPACK_ROW_LENGTH 0x0CF2
-#endif
+	#define GL_GLEXT_PROTOTYPES
+	#if defined(IPHONE)
+		#include <OpenGLES/ES2/gl.h>
+		#include <OpenGLES/ES2/glext.h>
+	#elif defined(__ANDROID__)
+		#include <GLES/gl.h>
```
aquadran, line 68: we should not include gles1 for gles2. any reason to do it?

lephilousophe: Indeed, it was a bad cut paste from backends/graphics/opengl/opengl-sys.h where GLES is included for Android.
During the merge of the two files I didn't took attention about this.
Now my question will be: should we try to support platforms using GLESv1 only and without GLAD?
I don't think we need to and from what I understand, it seems we don't support it today.

aquadran: gles1 support could be explored in future. dunno if we ever going to support gles1 platforms only, they might have cpu to slow. about GLAD, it could be supported depending on OS. anyway we could ignore it for now.

lephilousophe:
I did some builds on platforms supported by buildbot and GLAD gets included only when necessary so this seems to be OK.
I think it's ready for review.

aquadran (approved):

bluegr (approved):

PR #3647, opened 2022-01-06
CREATE_PROJECT: Consistent name for zlib

henke37:
The feature name didn't match with other code.

### DIFF
--- a/devtools/create_project/codeblocks.cpp
+++ b/devtools/create_project/codeblocks.cpp
@@ -58,8 +58,10 @@ StringList getFeatureLibraries(const BuildSetup &setup) {
 	for (FeatureList::const_iterator i = setup.features.begin(); i != setup.features.end(); ++i) {
 		if (i->enable && i->library) {
 			std::string libname;
-			if (!std::strcmp(i->name, "libz") || !std::strcmp(i->name, "libcurl")) {
+			if (!std::strcmp(i->name, "libcurl")) {
 				libname = i->name;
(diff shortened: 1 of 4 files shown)


devtools/create_project/codeblocks.cpp
```diff
@@ -58,8 +58,11 @@ StringList getFeatureLibraries(const BuildSetup &setup) {
 	for (FeatureList::const_iterator i = setup.features.begin(); i != setup.features.end(); ++i) {
 		if (i->enable && i->library) {
 			std::string libname;
-			if (!std::strcmp(i->name, "libz") || !std::strcmp(i->name, "libcurl")) {
+			if (!std::strcmp(i->name, "libcurl")) {
 				libname = i->name;
+			} else if (!std::strcmp(i->name, "zlib")) {
+				libname = "libz";
+				libraries.push_back("libvorbisfile");
```
Die4Ever, line 65: looks like libvorbisfile shouldn't be here

Die4Ever:
you're probably gonna need to change some `configure.engine` files, right?

devtools/create_project/codeblocks.cpp
henke37: Whopsie.

henke37:
> you're probably gonna need to change some `configure.engine` files, right?

I don't think I will need to, since they already build correctly on platforms that don't use `create_project`.

bluegr:
Thanks, the naming of the zlib library is more consistent with this. Merging

PR #3648, opened 2022-01-06
ENGINES: Fix autosave name check

Die4Ever:
Save names in Groovie are limited to 15 characters, and it's probably not the only engine with a similar limit, I wanted to see if there was any interest in a global fix that could help many engines.

The length of the German translation for Autosave is 23 characters. This meant you would get a warning on every autosave, even though you just told it to overwrite.

I currently have it set to a minimum of 14 characters for confidence that the name is a match and not a false positive, but this could definitely be reduced.

### DIFF
--- a/engines/savestate.cpp
+++ b/engines/savestate.cpp
@@ -95,7 +95,12 @@ bool SaveStateDescriptor::isAutosave() const {
 
 bool SaveStateDescriptor::hasAutosaveName() const
 {
-	return _description.contains(_("Autosave"));
+	Common::U32String autosave = _("Autosave");
+	if (autosave.size() < 14)
+		return autosave == _description;
+	if (_description.size() < 14)
+		return false;
+	return autosave.substr(0, _description.size()) == _description;
 }
 
 bool SaveStateDescriptor::isValid() const
--- a/engines/savestate.cpp
+++ b/engines/savestate.cpp
@@ -96,6 +96,8 @@ bool SaveStateDescriptor::isAutosave() const {
 bool SaveStateDescriptor::hasAutosaveName() const
 {
 	Common::U32String autosave = _("Autosave");
+	if (_description.size() >= autosave.size())
+		return _description.contains(autosave);
 	if (autosave.size() < 14)
 		return autosave == _description;
 	if (_description.size() < 14)
--- a/engines/savestate.cpp
+++ b/engines/savestate.cpp
@@ -95,7 +95,7 @@ bool SaveStateDescriptor::isAutosave() const {
 
 bool SaveStateDescriptor::hasAutosaveName() const
 {
-	Common::U32String autosave = _("Autosave");
+	const Common::U32String &autosave = _("Autosave");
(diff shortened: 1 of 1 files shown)


engines/savestate.cpp
```diff
@@ -95,7 +95,14 @@ bool SaveStateDescriptor::isAutosave() const {
 
 bool SaveStateDescriptor::hasAutosaveName() const
 {
-	return _description.contains(_("Autosave"));
+	Common::U32String autosave = _("Autosave");
+	if (_description.size() >= autosave.size())
+		return _description.contains(autosave);
```
Die4Ever, line 100: I did contains here so it would be unchanged behavior, but we could also do

`_description.substr(0, autosave.size()) == autosave;`

engines/savestate.cpp
```diff
@@ -95,7 +95,14 @@ bool SaveStateDescriptor::isAutosave() const {
...
+	if (autosave.size() < 14)
+		return autosave == _description;
```
Die4Ever, line 102: I don't know if any engines append a number at the end of the autosave? We could also do

`_description.substr(0, autosave.size()) == autosave;`

engines/savestate.cpp
```diff
@@ -95,7 +95,12 @@ bool SaveStateDescriptor::isAutosave() const {
 
 bool SaveStateDescriptor::hasAutosaveName() const
 {
-	return _description.contains(_("Autosave"));
+	Common::U32String autosave = _("Autosave");
```
orgads, line 98: const

orgads:
Is `_("Autosave")` more than 15 chars (bytes) because of unicode/punycode, translation or something else?

Die4Ever:
> Is `_("Autosave")` more than 15 chars (bytes) because of unicode/punycode, translation or something else?

The German translation was very long, which was how this issue was discovered by my German friend

engines/savestate.cpp
criezy: Seeing that code now I am wondering if there might be a better way to test if a savegame is an autosave. Since the description string can be translated this means that after changing the GUI language in the ScummVM option, existing autosaves will no longer be recognised as such...

Die4Ever: maybe something with SeekableReadStream? or a secondary meta file?

I just wanted to fix the issue my friend was having where he was constantly getting spammed with the autosave warning dialog

orgads: > Seeing that code now I am wondering if there might be a better way to test if a savegame is an autosave. Since the description string can be translated this means that after changing the GUI language in the ScummVM option, existing autosaves will no longer be recognised as such...

I've noticed that too, and I even commented on this issue in #3261, but I wanted to preserve the existing behavior too.

engines/savestate.cpp
orgads: I think it should be safe to replace `contains` with `substr`.

orgads (approved):

PR #3654, opened 2022-01-09
Add korean translation for ecoquest2 and gk2

swkim01:


bluegr:
Thanks! Merging

bluegr:
My mistake: Please, next time, prepend the commit with the engine name. In this case, it should be:

    SCI: Add korean translation for ecoquest2 and gk2

PR #3655, opened 2022-01-12
JANITORIAL: Force LF EOL for shell scripts

orgads:
Needed when core.autocrlf is set on Windows, and building with WSL.


bluegr:
Nice catch, thanks! Merging

PR #3658, opened 2022-01-12
DEVTOOLS: Bump minimum cmake version to 3.13

orgads:
This mutes the following warning:
```
CMake Warning (dev) at CMakeLists.txt:1581 (target_sources):
  Policy CMP0076 is not set: target_sources() command converts relative paths
  to absolute.  Run "cmake --help-policy CMP0076" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

  An interface source of target "scummvm" has a relative path.
This warning is for project developers.  Use -Wno-dev to suppress it.
```


bluegr:
Nice catch!
Here's the relevant article:

https://cmake.org/cmake/help/git-stage/policy/CMP0076.html

Merging

PR #3659, opened 2022-01-12
CREATE_PROJECT: Leverage C++11 string literals for cmake generator

orgads:


devtools/create_project/cmake/CMakeLists.txt
```diff
@@ -21,3 +21,5 @@ add_executable(create_project ${SOURCE_FILES})
 if (WIN32)
     target_link_libraries(create_project rpcrt4 advapi32)
 endif ()
+set_property(TARGET create_project PROPERTY CXX_STANDARD 11)
```
aquadran, line 24: description in commit doesn't reflect this, but make sense to add

devtools/create_project/xcode/create_project.xcodeproj/project.pbxproj
```diff
@@ -180,6 +180,7 @@
 		F9A66C2E1396D36100CEE494 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CLANG_CXX_LANGUAGE_STANDARD="c++0x";
```
aquadran, line 183: I don't understand why c++0x?

orgads: That's what I found in create_project itself, when it generates a project for Xcode. I really have no clue about Xcode. Would you prefer c++11? Change it also in create_project?

devtools/create_project/cmake/CMakeLists.txt
orgads: Do you want me to separate this and the Xcode to a "enable c++11" commit?

devtools/create_project/xcode/create_project.xcodeproj/project.pbxproj
aquadran: we have now c++11 required, so I think yes

SupSuper:
The string changes looks fine to me, though beware it's tempting to rewrite half of create_project once you start using C++11. :)

SupSuper (approved):
The string changes looks fine to me, though beware it's tempting to rewrite half of create_project once you start using C++11. :)

PR #3660, opened 2022-01-12
CREATE_PROJECT: Simplify engines part for CMake

orgads:
Create a function add_engine which handles everything for the engine, and
remove redundant conditions all over the project file.


SupSuper:
Commits seem to be overlapping with #3659

devtools/create_project/cmake.cpp
```diff
@@ -243,38 +281,21 @@ void CMakeProvider::createProjectFile(const std::string &name, const std::string
...
-		project << "if (ENABLE_" << engineName << ")\n";
-		project << "add_library(" << name << "\n";
-	}
-
-	std::string modulePath;
-	if (!moduleDir.compare(0, setup.srcDir.size(), setup.srcDir)) {
-		modulePath = moduleDir.substr(setup.srcDir.size());
-		if (!modulePath.empty() && modulePath.at(0) == '/')
-			modulePath.erase(0, 1);
+		enginesStr << "add_engine(" << name << "\n";
```
orgads, line 290: I'm using a buffer because the engines have to come after the main project. In a follow-up commit, I write the engines to separate files, and this buffer won't be used (when running in-source).

orgads: I can come up with alternatives if you're worried about memory consumption (I suppose it should be several 100k for a full build), or anything else.

bluegr: create_project is a development tool used in desktop PCs, which have more than enough memory. Some extra kilobytes of RAM won't be an issue in this case

somaen (approved):
Looks good to me, anything remaining here before we can merge this?

aquadran:
I think it has comments to improve code

PR #3662, opened 2022-01-13
NGI: Fix Inconsistent character state (bug #9669)

r41k0u:
Implemented Cactus HasGrown check in scene32.cpp which resolves [#9669](https://bugs.scummvm.org/ticket/9669)


sev-:
Thanks!

PR #3663, opened 2022-01-13
CREATE_PROJECT: Replace c++0x with c++11 in Xcode generator

orgads:
@ccawley2011 Please review. I'm not sure if this is compatible with older versions of Xcode.


aquadran:
'c++11' is in current Xcode selection. I may try older version

aquadran:
Xcode 9 also use 'c++11'

orgads:
Is this good to merge?

lephilousophe:
@criezy is the MacOS maintainer, he is the one who should know.

somaen:
While we do attempt to retain portability of ScummVM itself with as old releases as viable, do we have to do the same for dev-environments? XCode 9 is fairly old (2017). I can check a bit further back if we care about supporting something like OS X 10.7 for development, or perhaps 10.6 is a common use case due to it's support for Rosetta?

somaen:
OK, just tested, and this was at least used in 10.7, not sure when it was replaced yet.

criezy:
The oldest development system I use myself is Xcode 6.2 on Mac OS X 10.9. And that version already supports using `-std=c++11`.

The change I guess impacts both macOS and iOS builds. I suspect for iOS it is less likely that an old system will be used. And as far as I am concerned, for macOS the main build system is configure/make, and create_project is mostly here for convenience. So I am not opposed to the change.

criezy (approved):

PR #3664, opened 2022-01-14
CREATE_PROJECT: Create separate CMakeLists.txt for engines

orgads:


orgads:
Notice this PR depends on #3660. If you want to review the relevant changes, inspect only the last commit.

aquadran:
can you explain why separated cmakelists?

orgads:
This is the common practice with CMake. It's also easier to maintain than a single huge file.

aquadran:
this file is generated, so there is no maintenance. having one dirty file in repo looks better for me than many for each engine.
I don't see real value here.

orgads:
The target is to commit the CMake files to the repository and phase out create_project at some stage.

sev-:
This is interesting news for me. Who decided to phase out create_project? What is the rationale behind it?

orgads:
It wasn't decided. I'm in the process of a PoC for cmake.

Quoting @SupSuper on https://discord.com/channels/581224060529148060/711242520415174666/930914060722114580

> personally i think maintaining create_project is hell, if we're gonna have two buildsystems we might as well just use configure/cmake, which are easier to match. i don't have time to pursue this though, if you think it's something worth looking into feel free @orgads

aquadran:
I'm not sure if all ports are capable use cmake. create_project is not easy, but I don't agree it's hell to maintain it. switching to cmake can create maintenance problems, like compatibility a cross various systems.
anyway, if it's part of PoC, then this PR should not be accepted.

sev-:
create_project is currently producing project files for MSVC, XCode and not-so-much-used Code::Blocks. How cmake is going to solve it?

orgads:
CMake can generate project files for all of these and more.

SupSuper:
Figured I should explain here given my comment was taken out of context:

I think it's worth exploring alternatives to create_project. It made sense back when it was originally created for MSVC and there were no real other alternatives, but there's a lot more compilers, tools and options out now. Meanwhile create_project regularly falls behind the main project, doesn't have feature-parity with configure, didn't have any testing until Github CI was introduced, etc. Anytime any new C++ compiler, package, library, system, etc comes out, it has to be manually updated and developers pinged to catch up, and I don't think that's something ScummVM developers should have to worry about, that's the job of build tools.

Since @orgads was poking around create_project and CMake, I figured I would throw the idea out. Maybe the answer is CMake, maybe the answer is simplifying create_project, maybe it's somewhere in the middle, but I think it's worth investigating.

> I'm not sure if all ports are capable use cmake. create_project is not easy, but I don't agree it's hell to maintain it. switching to cmake can create maintenance problems, like compatibility a cross various systems.

@aquadran My understanding is every platform and build is using configure, which would remain as-is. The only ones using create_project are MSVC and XCode devs, so any alternative only needs to support those (at least).

> anyway, if it's part of PoC, then this PR should not be accepted.

But I agree with this, PRs should be self-contained. @orgads if you want to make a PoC then you should save a PR for the end, with pros/cons over the current system. If you want to make create_project cleaner along the way, it might be better saved for that, specially if your commits are all dependent. If you just wanna check developer interest or review as you go, that's better done informally over Discord.

orgads:
Ok. I'll close it for now, and will open a new PR if/when it is ready.

Thanks.

lephilousophe:
@SupSuper 
> @aquadran My understanding is every platform and build is using configure, which would remain as-is. The only ones using create_project are MSVC and XCode devs, so any alternative only needs to support those (at least).

IMO, if we switch from create_project to CMake, that's not to keep two build systems at the same time. It would be to merge configure and CMake, else I don't see the point of this work.
I am not sure create_project by itself adds more complexity than CMake would.

PR #3666, opened 2022-01-14
SKY: Fix saving names for non selected save slots

antoniou79:
Fix for bug #13218 "Can Populate More Than One Save Slot At A Time"

We use a dirty string buffer to temporary store the edited text for the selected slot

### DIFF
--- a/engines/sky/control.cpp
+++ b/engines/sky/control.cpp
@@ -897,8 +897,10 @@ uint16 Control::saveRestorePanel(bool allowSave) {
 		textSprites[cnt] = NULL;
 	_firstText = 0;
 
+	Common::String dirtyBufStr;
 	loadDescriptions(saveGameTexts);
 	_selectedGame = 0;
+	dirtyBufStr = saveGameTexts[_selectedGame];
 
 	bool quitPanel = false;
 	bool refreshNames = true;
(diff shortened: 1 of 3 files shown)


digitall:
@antoniou79 : Thanks for looking at this. My slightly naiver / less invasive change should have worked, but I may have been missing something about how StringArray works. Your code looks much clearer in behaviour. Will give it a test and if it works fine for me, I will merge this PR. Thanks again.

digitall:
@antoniou79 : Tested and this works fine to solve the bug. One further point though, I noticed that we can also save when the savename is blank i.e. dirty string is "" which with the original dialogs is not clear that a savegame is present. We probably should also add a check to saving to prevent save when the name is "".

antoniou79:
@digitall I agree. I'll look into it and probably have something ready in a few hours (I also want to test with the game on Dosbox to see the original's behavior, but it probably did not allow empty named saved games. 
Also, this seems to be something that we should look out for in other engines too (I just tested  Blade Runner and it also has this issue *partially*). ScummVM's GMM supports empty named saved games and will show them as "unnamed save game" or similar, but the game's native UI usually does not (not originally anyway) and there won't be much indication there that a slot has a valid saved game if its name is "").

antoniou79:
@digital: I pushed some commits; one to fix the issue with empty string, the other to fix glitches caused when the player could have opened the game's save or restore panel and still be able to open the GMM on top and save a file name (so now the player cannot Save/Load from GMM when the control panel is open, which also postpones autosaving).

digitall:
@antoniou79 : Code changes look good to me. I did a quick test and it seems to work fine as expected so am going to merge this as-is and close the bug out as fixed. Thanks for working on this.

PR #3668, opened 2022-01-16
CONFIGURE: Strip --no-undefined linker flag from sdl2

orgads:
Not supported by LLD <12.

### DIFF
--- a/configure
+++ b/configure
@@ -3801,9 +3801,9 @@ if test "$_sdl" = auto ; then
 	find_sdlconfig
 	append_var SDL_CFLAGS "`$_sdlconfig --cflags | sed 's/[[:space:]]*-Dmain=SDL_main//g'`"
 	if test "$_static_build" = yes ; then
-		append_var SDL_LIBS "`$_sdlconfig --static-libs`"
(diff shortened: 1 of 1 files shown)


lephilousophe:
This change here should be restricted to LLD case.

Anyway, with MXE we were really in a corner case: we built SDL using BFD or Gold and then we switched to LLD for ScummVM.
This may be better to only mangle this at toolchain level.

orgads:
Is this flag really needed? What for?

lephilousophe:
In a general way, I don't think it's a good thing to mess with flags provided by other libraries. This let us in a position where it's easy to report bugs against it.
IIRC, for now the only thing we remove is -mwindows and -mconsole.
The -Wl,--no-undefined flag forces the linker to raise an error when some symbols are not defined.

BTW, the --no-undefined flag exists in LLD ELF driver.
The error happens when we use LLD with its Cygwin driver until version 13 where the flag is ignored (https://github.com/llvm/llvm-project/commit/f84a4cb0df90098ffde4bd79ace5964d04497cfc).
The proper solution is to build SDL with LLD to make sure the flag doesn't get added by mistake.

PR #3672, opened 2022-01-17
ALL: Fix BufferedWriteStream to fix AGS savefile creation on Nintendo Switch

hackingdemon:
AGS savefile creation routine uses GetPosition() and Seek() on the stream, that is implemented via BufferedWriteStream in Switch port. Since BufferedWriteStream::pos() is broken at the moment, saving the AGS game produces invalid files that cannot be loaded back (the engine skips them).
Potentially fixes other engines as well, but I tried only AGS

PR #3674, opened 2022-01-18
COMMON: Fix BufferedWriteStream::pos() method

hackingdemon:
The pos() method of BufferedWriteStream should return the stream position
but actually returns the buffer position. That completely breaks saving
the game in AGS engine on the platforms using BufferedWriteStream wrapper
for savefile handling, such as Nintendo Switch. AGS engine's save file
writing functions use GetPosition() with later Seek() on stream,
so an invalid return value from GetPosition() results in invalid save files
that cannot be loaded and are skipped by the engine.

### DIFF
--- a/common/stream.cpp
+++ b/common/stream.cpp
@@ -550,7 +550,7 @@ public:
 
 	bool flush() override { return flushBuffer(); }
 
-	int64 pos() const override { return _pos; }
+	int64 pos() const override { return _pos + _parentStream->pos(); }
 
 	bool seek(int64 offset, int whence) override {
 		flush();


hackingdemon:
remade #3672 which closed automatically while I was fixing the commit

hackingdemon:
as a reference:  [forum: AGS games won't save on Switch, but will on Mac (latest daily build)](https://forums.scummvm.org/viewtopic.php?t=16380)

hackingdemon:
Also created a bug, but not sure how to link it properly [ #13226 Saving AGS games is broken on Nintendo Switch ](https://bugs.scummvm.org/ticket/13226)

digitall:
@hackingdemon : If you do a rebase of a PR branch and remove all commits i.e make it level with master, Github closes it and I don't think you can add commits again. I think it is intended to stop swapping out of PRs in a underhanded way by bad actors... or it is a Github bug. *shrug* :|

digitall:
@hackingdemon : The commit message formatting looks fine now.

In a quick impact review, BufferedWriteStream is only used by backends/fs/psp/psp-fs.cpp , backends/fs/posix-drives/posix-drives-fs.cpp and common/recorderfile.cpp. Given that this should have affected more platforms i.e. all POSIX and PSP systems, so not sure why it only manifested with AGS savefiles on Nintendo Switch (which I think is using POSIX FS?)

digitall:
@hackingdemon : I assume that no other cases of the usage of the POSIX FS stream returned using pos() ?

hackingdemon:
@digitall 
> I assume that no other cases of the usage of the POSIX FS stream returned using pos() ?

It is more convoluted: the issue manifests itself in posix-**drives** fs when port configures it with the option DrivePOSIXFilesystemNode::**kBufferingModeScummVM** (I don't know why stdio buffering was not good enough).

See: /backends/platform/sdl/switch/switch.cpp:69:
	fsFactory->configureBuffering(DrivePOSIXFilesystemNode::kBufferingModeScummVM, 2048);

digitall:
@hackingdemon : Ah I see. An audit of the code shows that only two ports set kBufferingModeScummVM, the Nintendo Switch and the 3DS port (which is probably not powerful enough for AGS games) hence why this only affects them.

Right, I think this is fine to merge then...

digitall:
@aquadran : Are you happy this now meets the https://wiki.scummvm.org/index.php?title=Commit_Guidelines ?

hackingdemon:
@digitall , as for backends/fs/psp/psp-fs.cpp, I think nobody tried it with AGS (if it is enabled for psp)

hackingdemon:
What should I do next regarding this PR and https://bugs.scummvm.org/ticket/13226? Just wait patiently for one of the maintainers to merge the PR and close the bug?

digitall:
@aquadran : Thanks for confirming.
@hackingdemon : Don't worry. Will handle it. Thanks again for the debugging.

hackingdemon:
thank you!

PR #3675, opened 2022-01-19
Add Screenshot Option

liberodark:
Hi,

On batocera we have add this option i think is better to upstream this option.

Best Regards
PS : Thanks for your great work

### DIFF
--- a/base/commandLine.cpp
+++ b/base/commandLine.cpp
@@ -144,6 +144,7 @@ static const char HELP_STRING[] =
 	"                           pce, segacd, wii, windows)\n"
 	"  --savepath=PATH          Path to where saved games are stored\n"
 	"  --extrapath=PATH         Extra path to additional game data\n"
+	"  --screenshots=PATH       Path to where screenshots are saved\n"
 	"  --soundfont=FILE         Select the SoundFont for MIDI playback (only\n"
 	"                           supported by some MIDI drivers)\n"
 	"  --multi-midi             Enable combination AdLib and native MIDI\n"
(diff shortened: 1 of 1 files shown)


henke37:
You are missing the commit name prefix.

sev-:
I am not sure how this code is going to work. E.g. you're setting the option, but I do not see where it is used.

base/commandLine.cpp
```diff
@@ -144,6 +144,7 @@ static const char HELP_STRING[] =
 	"                           pce, segacd, wii, windows)\n"
 	"  --savepath=PATH          Path to where saved games are stored\n"
 	"  --extrapath=PATH         Extra path to additional game data\n"
+	"  --screenshots=PATH       Path to where screenshots are saved\n"
```
criezy, line 147:
```suggestion
	"  --screenshotspath =PATH  Path to where screenshots are saved\n"
```

criezy (changes requested):
That's a good idea. But the config parameter we are using for the screenshots path is `"screenshotpath"` and not `"screenshotspath"`. Due to the additional `s` your proposed changes will have no impact on the path being actually used. So you will have to remove that 's' for it to work as expected.

Also please follow our [commit guidelines](https://wiki.scummvm.org/index.php?title=Commit_Guidelines#Commit_message_formatting) for the commit message formatting.

criezy:
> I am not sure how this code is going to work. E.g. you're setting the option, but I do not see where it is used.

It will not 😉 
I added the use of the "screenshotpath" config in e96c057c3d, but in this PR there is a typo on the name and the additional 's' means it will not work. But in principle once this is fixed it should work.

criezy:
I had a quick look at this pull request again. There is another issue beside the the inconsistent name for the option. It gets added to the transient domain in `ConfMan` when starting ScummVM as expected, but it then gets removed when the launcher is started. This means with just the code in this pull request the command line option would only work when starting a game directly from the command line and only until returning to the launcher.
I think we would expect it to works when starting with the launcher, or after returning to the launcher. But that would require more work.

criezy:
Closing this one in favour of #3698.

PR #3676, opened 2022-01-20
COMMON: Moved Node struct from private to public in HashMap

philokev:
Moved Node Struct from private to public in HashMap container so that it can be used with Common::find_if from algorithm.h. 

At the moment it is possible to use find_if when iterating through other containers (e.g. Common::Array) because the type contained is typically public. However, attempting to do this with a Hashmap requires the use of the inner Node struct as part of the lambda or function object signature when it is passed to find_if as a predicate.

Using a range based for loop works in C++11, which is required to use lambdas, but this alternative would not be possible for code bases still relying on earlier versions of C++ where a function object would be used.

### DIFF
--- a/common/hashmap.h
+++ b/common/hashmap.h
@@ -88,16 +88,16 @@ class HashMap {
 public:
 	typedef uint size_type;
 
-private:
-
-	typedef HashMap<Key, Val, HashFunc, EqualFunc> HM_t;
-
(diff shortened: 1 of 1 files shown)


philokev:
Hi, any objections to this edit?

sev-:
Thanks!

PR #3677, opened 2022-01-22
IMAGE: Add 16 bitmap support to raw decoder

philokev:
Added support for 16 bit bitmap images (using RGB555 format). Implemented according to the algorithm documented in the link below:

https://docs.microsoft.com/en-us/windows/win32/directshow/working-with-16-bit-rgb

### DIFF
--- a/image/bmp.cpp
+++ b/image/bmp.cpp
@@ -96,12 +96,18 @@ bool BitmapDecoder::loadStream(Common::SeekableReadStream &stream) {
 	/* uint16 planes = */ stream.readUint16LE();
 	uint16 bitsPerPixel = stream.readUint16LE();
 
-	if (bitsPerPixel != 4 && bitsPerPixel != 8 && bitsPerPixel != 24 && bitsPerPixel != 32) {
+	if (bitsPerPixel != 4 && bitsPerPixel != 8 && bitsPerPixel != 16 && bitsPerPixel != 24 && bitsPerPixel != 32) {
 		warning("%dbpp bitmaps not supported", bitsPerPixel);
 		return false;
 	}
 
 	uint32 compression = stream.readUint32BE();
+
+	if (bitsPerPixel == 16 && compression != SWAP_CONSTANT_32(0)) {
+		warning("only RGB555 raw mode supported for %dbpp bitmaps", bitsPerPixel);
+		return false;
+	}
+
 	uint32 imageSize = stream.readUint32LE();
 	/* uint32 pixelsPerMeterX = */ stream.readUint32LE();
 	/* uint32 pixelsPerMeterY = */ stream.readUint32LE();
--- a/image/codecs/bmp_raw.cpp
+++ b/image/codecs/bmp_raw.cpp
@@ -91,6 +91,20 @@ const Graphics::Surface *BitmapRawDecoder::decodeFrame(Common::SeekableReadStrea
 			stream.read(dst + (_flip ? i : _height - i - 1) * _width, _width);
 			stream.skip(extraDataLength);
 		}
+	} else if (_bitsPerPixel == 16) {
+		byte *dst = (byte *)_surface.getBasePtr(0, _height - 1);
+
+		for (int i = 0; i < _height; i++) {
+			for (int j = 0; j < _width; j++) {
+				uint16 color = stream.readUint16LE();
+
+				*(uint16 *)dst = color;
+				dst += format.bytesPerPixel;
+			}
+
+			stream.skip(extraDataLength);
+			dst -= _surface.pitch * 2;
+		}
 	} else if (_bitsPerPixel == 24) {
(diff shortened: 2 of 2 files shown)


lephilousophe:
This PR miss some corner cases.
16-bit bitmaps can either be `BI_RGB` or `BI_BITFIELDS`.
While `BI_RGB` only allows for RGB555 (as you implemented it), `BI_BITFIELDS` is way more flexible (RGBA4444, RGB565, ...) and would need more changes to `BitmapRawDecoder` and `BitmapDecoder`.
In the PR, `BI_BITFIELDS` bitmaps which are not RGB555 are rendered wrong.

philokev:
Thanks for your feedback. I will look into implementing those other corner cases that you mentioned BI_BITFIELDS.

Do you have any documentation with implementation details that covers those cases?

Am I correct that BI_BITFIELDS is only supported for headers v4 and 5?

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/4e588f70-bd92-4a6f-b77f-35d0feaf7a57

lephilousophe:
I thought `BI_BITFIELDS` was documented in [BITMAPINFOHEADER](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapinfoheader) but in fact it sends back to some `AM_MEDIA_TYPE` structure.
So, either you add support for the full V4/V5 header or you can also just error out if bitmap is 16-bits and `BI_BITFIELDS` is used.
This will avoid rendering bad bitmaps in any case.

philokev:
For now I will go with the second approach as I need this functionality for an engine I'm working on.

It looks like in both cases it will require a modification to the BitmapRawDecoder constructor as it does not currently the header version or compression enum into account.

From bmp_raw.cpp, line 30:

`BitmapRawDecoder(int width, int height, int bitsPerPixel, bool ignoreAlpha, bool flip)`

From bmp.cpp, line 127 - 130:

```
	// Create the codec (it will warn about unhandled compression)
	_codec = createBitmapCodec(compression, 0, width, height, bitsPerPixel);
	if (!_codec)
		return false;
```

From codec.cpp, lines 198 to 215:
```

Codec *createBitmapCodec(uint32 tag, uint32 streamTag, int width, int height, int bitsPerPixel) {
	// Crusader videos are special cased here because the frame type is not in the "compression"
	// tag but in the "stream handler" tag for these files
	if (JYV1Decoder::isJYV1StreamTag(streamTag)) {
		assert(bitsPerPixel == 8);
		return new JYV1Decoder(width, height, streamTag);
	}

	switch (tag) {
	case SWAP_CONSTANT_32(0):
		return new BitmapRawDecoder(width, height, bitsPerPixel, true);
	case SWAP_CONSTANT_32(1):
		return new MSRLEDecoder(width, height, bitsPerPixel);
	case SWAP_CONSTANT_32(2):
		return new MSRLE4Decoder(width, height, bitsPerPixel);
	case SWAP_CONSTANT_32(3):
		// Used with v4-v5 BMP headers to produce transparent BMPs
		return new BitmapRawDecoder(width, height, bitsPerPixel, false);
```

lephilousophe:
Yes, that's what I meant by more changes to `BitmapRawDecoder`.
If some engine needs support for `BI_BITFIELDS` it will be possible to add it at this moment.

philokev:
I added a check for 16 bit bitmaps with bitfields.

If a need arises for this format, then it can be implemented and tested at that point.

image/bmp.cpp
```diff
@@ -96,12 +96,18 @@ bool BitmapDecoder::loadStream(Common::SeekableReadStream &stream) {
...
 
-	if (bitsPerPixel != 4 && bitsPerPixel != 8 && bitsPerPixel != 24 && bitsPerPixel != 32) {
+	if (bitsPerPixel != 4 && bitsPerPixel != 8 && bitsPerPixel != 16 && bitsPerPixel != 24 && bitsPerPixel != 32) {
 		warning("%dbpp bitmaps not supported", bitsPerPixel);
 		return false;
 	}
 
 	uint32 compression = stream.readUint32BE();
+
+	if (bitsPerPixel == 16 && compression == 3) {
```
lephilousophe, line 106: To things here:
First, the compression is read BE so constant should be reversed like it is in `createBitmapCodec` using `SWAP_CONSTANT_32`.
Second, as RLE codec doesn't support 16-bits bitmaps, it would be safer to check that value is really `SWAP_CONSTANT_32(0)`.

lephilousophe (changes requested):
A merge commit has been inserted.
Could you rebase your work (and force push) instead of merging?
It's not a big deal though because at the end the PR will likely be squashed.

image/bmp.cpp
philokev: Thanks! Changed

lephilousophe (approved):

lephilousophe:
Looks good to me!

aquadran (approved):

PR #3679, opened 2022-01-24
AMIGAOS: Let user choose the OpenGL implementation

lephilousophe:
This is a follow-up of #3422 by @raziel-.

Now that we have GLAD which allows dynamic loading of OpenGL functions, it's possible to use set a SDL attribute to select between miniGL and OGLES2 implementation.

I also removed multiple ifdef regarding framebuffers.
I suspect these ifdef were here because miniGL doesn't provide definitions for these functions. As they are now provided by GLAD, code should build.
A runtime check before using the framebuffer was already in place so this should be safe.

Finally, the context created to check for framebuffer presence was assuming an OpenGL implementation. This is not always the case and depends on SDL defaults or what we have requested.

The TODO list:
- [x] make sure it builds on Amiga (I have to run tests on buildbot).
- [ ] do tests on real Amiga (@raziel- it's for you :wink:). I only tested the idea by patching OSystem_POSIX with the same piece of code
- [ ] make sure that if OGLES2 is not available, user won't be locked out and that SDL falls back on MiniGL
- [x] fix the label size to display the full text
- [x] make sure that no incompatible choice is done between 3D mode and OpenGL implementation chosen. (Done in #3781)

### DIFF
--- a/backends/graphics3d/opengl/framebuffer.cpp
+++ b/backends/graphics3d/opengl/framebuffer.cpp
@@ -22,7 +22,7 @@
 #include "common/textconsole.h"
 #include "common/util.h"
 
-#if (defined(USE_OPENGL_GAME) || defined(USE_OPENGL_SHADERS)) && !defined(AMIGAOS) && !defined(__MORPHOS__)
+#if defined(USE_OPENGL_GAME) || defined(USE_OPENGL_SHADERS)
 
 #include "backends/graphics3d/opengl/framebuffer.h"
 
@@ -124,7 +124,7 @@ void FrameBuffer::detach() {
 	glViewport(_prevStateViewport[0], _prevStateViewport[1], _prevStateViewport[2], _prevStateViewport[3]);
 }
 
-#if !USE_FORCED_GLES2 && !defined(AMIGAOS) && !defined(__MORPHOS__)
+#if !USE_FORCED_GLES2
 MultiSampleFrameBuffer::MultiSampleFrameBuffer(uint width, uint height, int samples)
 		: FrameBuffer(width,height) {
 	if (!OpenGLContext.framebufferObjectMultisampleSupported) {
@@ -186,7 +186,7 @@ void MultiSampleFrameBuffer::detach() {
 	glViewport(_prevStateViewport[0], _prevStateViewport[1], _prevStateViewport[2], _prevStateViewport[3]);
 }
 
-#endif // !USE_FORCED_GLES2 && !defined(AMIGAOS) && !defined(__MORPHOS__)
+#endif // !USE_FORCED_GLES2
 
 } // End of namespace OpenGL
 
--- a/backends/graphics3d/opengl/framebuffer.h
+++ b/backends/graphics3d/opengl/framebuffer.h
@@ -34,17 +34,10 @@ class FrameBuffer : public TextureGL {
 public:
 	FrameBuffer(uint width, uint height);
 	FrameBuffer(GLuint texture_name, uint width, uint height, uint texture_width, uint texture_height);
-#if defined(AMIGAOS) || defined(__MORPHOS__)
-	virtual ~FrameBuffer() {}
-
-	void attach() {}
-	void detach() {}
-#else
 	virtual ~FrameBuffer();
 
 	virtual void attach();
 	virtual void detach();
-#endif
 
 protected:
 	GLuint getFrameBufferName() const { return _frameBuffer; }
@@ -56,7 +49,7 @@ private:
 	GLint _prevStateViewport[4];
 };
 
-#if !USE_FORCED_GLES2 && !defined(AMIGAOS) && !defined(__MORPHOS__)
+#if !USE_FORCED_GLES2
 class MultiSampleFrameBuffer : public FrameBuffer {
 public:
 	MultiSampleFrameBuffer(uint width, uint height, int samples);
--- a/backends/graphics3d/opengl/surfacerenderer.cpp
+++ b/backends/graphics3d/opengl/surfacerenderer.cpp
@@ -178,14 +178,23 @@ static const char *boxVertex =
 	"uniform vec2 offsetXY;\n"
 	"uniform vec2 sizeWH;\n"
 	"uniform vec2 texcrop;\n"
+// OGLES2 on AmigaOS doesn't support uniform booleans
+#if defined(AMIGAOS)
+	"uniform mediump int flipY;\n"
+#else
 	"uniform bool flipY;\n"
+#endif
 	"varying vec2 Texcoord;\n"
 	"void main() {\n"
 		"Texcoord = texcoord * texcrop;\n"
 		"vec2 pos = offsetXY + position * sizeWH;\n"
 		"pos.x = pos.x * 2.0 - 1.0;\n"
 		"pos.y = pos.y * 2.0 - 1.0;\n"
+#if defined(AMIGAOS)
+		"if (flipY != 0)\n"
+#else
 		"if (flipY)\n"
+#endif
 			"pos.y *= -1.0;\n"
 		"gl_Position = vec4(pos, 0.0, 1.0);\n"
 	"}\n";
--- a/backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
+++ b/backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp
@@ -356,13 +356,11 @@ void OpenGLSdlGraphics3dManager::createOrUpdateScreen() {
 
 	_screenChangeCount++;
 
-#if !defined(__amigaos4__) && !defined(__MORPHOS__)
 	if (renderToFrameBuffer) {
 		glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
 		_frameBuffer = createFramebuffer(_engineRequestedWidth, _engineRequestedHeight);
 		_frameBuffer->attach();
 	}
-#endif
 }
 
 Math::Rect2d OpenGLSdlGraphics3dManager::computeGameRect(bool renderToFrameBuffer, uint gameWidth, uint gameHeight,
@@ -579,7 +577,6 @@ void OpenGLSdlGraphics3dManager::drawOverlay() {
 	_surfaceRenderer->restorePreviousState();
 }
 
-#if !defined(__amigaos4__) && !defined(__MORPHOS__)
 OpenGL::FrameBuffer *OpenGLSdlGraphics3dManager::createFramebuffer(uint width, uint height) {
 #if !USE_FORCED_GLES2
 	if (_antialiasing && OpenGLContext.framebufferObjectMultisampleSupported) {
@@ -590,7 +587,6 @@ OpenGL::FrameBuffer *OpenGLSdlGraphics3dManager::createFramebuffer(uint width, u
 		return new OpenGL::FrameBuffer(width, height);
 	}
 }
-#endif
 
 void OpenGLSdlGraphics3dManager::updateScreen() {
 	if (_frameBuffer) {
--- a/backends/platform/sdl/amigaos/amigaos.cpp
+++ b/backends/platform/sdl/amigaos/amigaos.cpp
@@ -47,4 +47,58 @@ bool OSystem_AmigaOS::hasFeature(Feature f) {
 
 	return OSystem_SDL::hasFeature(f);
 }
+
+void OSystem_AmigaOS::initBackend() {
+	// AmigaOS4 SDL provides two OpenGL implementations (OpenGL 1.3 with miniGL and OpenGL ES with OGLES2)
+	// This is chosen by setting the profile mask attribute before the first window creation but after init
+	int force = 0;
+	if (ConfMan.hasKey("opengl_implementation")) {
+		Common::String implem = ConfMan.get("opengl_implementation");
+		if (implem == "gl") {
+			force = 1;;
+		} else if (implem == "gles2") {
+			force = 2;
+		}
+	}
+
+	// If not forcing, try OGLES2 first
+	if (!force || force == 2) {
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 2);
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+		if (SDL_GL_LoadLibrary(NULL) < 0) {
+			if (force) {
+				warning("OpenGL implementation chosen is unsupported, falling back");
+				force = 0;
+			}
+			// SDL doesn't seem to be clean when loading fail
+			SDL_GL_UnloadLibrary();
+			SDL_GL_ResetAttributes();
+		} else {
+			// Loading succeeded, don't try anything more
+			force = 2;
+		}
+	}
+	// If not forcing, next try miniGL
+	if (!force || force == 1) {
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, 0);
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 1);
+		SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 3);
+		if (SDL_GL_LoadLibrary(NULL) < 0) {
+			if (force) {
+				warning("OpenGL implementation chosen is unsupported, falling back");
+				force = 0;
+			}
+			// SDL doesn't seem to be clean when loading fail
+			SDL_GL_UnloadLibrary();
+			SDL_GL_ResetAttributes();
+		} else {
+			// Loading succeeded, don't try anything more
+			force = 1;
+		}
+	}
+
+	OSystem_SDL::initBackend();
+}
+
 #endif
--- a/backends/platform/sdl/amigaos/amigaos.h
+++ b/backends/platform/sdl/amigaos/amigaos.h
@@ -32,6 +32,7 @@ public:
 	bool hasFeature(Feature f) override;
 
 	void init() override;
+	void initBackend() override;
 };
 
 #endif
--- a/backends/platform/sdl/amigaos/amigaos.mk
+++ b/backends/platform/sdl/amigaos/amigaos.mk
@@ -41,6 +41,10 @@ ifdef DIST_FILES_THEMES
 	makedir all $(AMIGAOSPATH)/themes
 	cp $(DIST_FILES_THEMES) $(AMIGAOSPATH)/themes
 endif
+ifneq ($(DIST_FILES_SHADERS),)
+	makedir all $(AMIGAOSPATH)/shaders
+	cp $(DIST_FILES_SHADERS) $(AMIGAOSPATH)/shaders
+endif
 	# Strip and copy engine plugins.
 ifdef DYNAMIC_MODULES
 	makedir all $(AMIGAOSPATH)/plugins
--- a/backends/platform/sdl/sdl.cpp
+++ b/backends/platform/sdl/sdl.cpp
@@ -328,7 +328,7 @@ void OSystem_SDL::detectOpenGLFeaturesSupport() {
 	_oglType = OpenGL::kOGLContextGLES2;
 	_supportsFrameBuffer = true;
 	_supportsShaders = true;
-#elif !defined(AMIGAOS) && !defined(__MORPHOS__)
+#else
 	// Spawn a 32x32 window off-screen with a GL context to test if framebuffers are supported
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	SDL_Window *window = SDL_CreateWindow("", SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 32, 32, SDL_WINDOW_OPENGL | SDL_WINDOW_HIDDEN);
--- a/engines/grim/shaders/emi_actor.fragment
+++ b/engines/grim/shaders/emi_actor.fragment
@@ -2,7 +2,7 @@ in vec2 Texcoord;
 in vec4 Color;
 
 uniform sampler2D tex;
-uniform bool textured;
+uniform UBOOL textured;
 uniform float alphaRef;
 uniform float meshAlpha;
 
@@ -11,7 +11,7 @@ OUTPUT
 void main()
 {
 	outColor = Color;
-	if (textured) {
+	if (UBOOL_TEST(textured)) {
 		vec4 texColor = texture(tex, Texcoord);
 		outColor.rgba *= texColor.rgba;
 		outColor.a *= meshAlpha;
--- a/engines/grim/shaders/emi_actor.vertex
+++ b/engines/grim/shaders/emi_actor.vertex
@@ -9,12 +9,12 @@ uniform highp mat4 projMatrix;
 uniform highp mat4 extraMatrix;
 uniform highp mat4 normalMatrix;
 uniform highp vec3 cameraPos;
-uniform bool textured;
-uniform bool useVertexAlpha;
+uniform UBOOL textured;
+uniform UBOOL useVertexAlpha;
 uniform vec4 uniformColor;
 
 struct shadow_info {
-	bool _active;
+	UBOOL _active;
 	vec3 _color;
 	vec3 _light;
 	vec3 _point;
@@ -32,7 +32,7 @@ void main()
 	pos = modelMatrix * pos;
 
 	// See http://en.wikipedia.org/wiki/Line-plane_intersection
-	if (shadow._active) {
+	if (UBOOL_TEST(shadow._active)) {
 		pos /= pos.w;
 		vec3 l = pos.xyz - shadow._light;
 		float d = dot(shadow._point - shadow._light, shadow._normal) / dot(l, shadow._normal);
@@ -49,15 +49,15 @@ void main()
 
 	gl_Position = projectedPos;
 
-	if (shadow._active) {
+	if (UBOOL_TEST(shadow._active)) {
 		Color = vec4(shadow._color, 1.0);
 	} else {
 		Color = color;
 	}
-	if (!useVertexAlpha)
+	if (!UBOOL_TEST(useVertexAlpha))
 		Color.a = 1.0;
 	Color *= uniformColor;
-	if (textured) {
+	if (UBOOL_TEST(textured)) {
 		Texcoord = texcoord;
 	} else {
 		Texcoord = vec2(0.0, 0.0);
--- a/engines/grim/shaders/emi_actorlights.fragment
+++ b/engines/grim/shaders/emi_actorlights.fragment
@@ -2,7 +2,7 @@ in vec2 Texcoord;
 in vec4 Color;
 
 uniform sampler2D tex;
-uniform bool textured;
+uniform UBOOL textured;
 uniform float alphaRef;
 uniform float meshAlpha;
 
@@ -11,7 +11,7 @@ OUTPUT
 void main()
 {
 	outColor = Color;
-	if (textured) {
+	if (UBOOL_TEST(textured)) {
 		vec4 texColor = texture(tex, Texcoord);
 		outColor.rgba *= texColor.rgba;
 		outColor.a *= meshAlpha;
--- a/engines/grim/shaders/emi_actorlights.vertex
+++ b/engines/grim/shaders/emi_actorlights.vertex
@@ -9,10 +9,10 @@ uniform highp mat4 projMatrix;
 uniform highp mat4 extraMatrix;
 uniform highp mat4 normalMatrix;
 uniform highp vec3 cameraPos;
-uniform bool textured;
-uniform bool useVertexAlpha;
+uniform UBOOL textured;
+uniform UBOOL useVertexAlpha;
 uniform vec4 uniformColor;
-uniform bool hasAmbient;
+uniform UBOOL hasAmbient;
 
 const int maxLights = 8;
 uniform vec4 lightsPosition[maxLights];
@@ -21,7 +21,7 @@ uniform vec4 lightsColor[maxLights];
 uniform vec4 lightsParams[maxLights];
 
 struct shadow_info {
-	bool _active;
+	UBOOL _active;
 	vec3 _color;
 	vec3 _light;
 	vec3 _point;
@@ -39,7 +39,7 @@ void main()
 	pos = modelMatrix * pos;
 
 	// See http://en.wikipedia.org/wiki/Line-plane_intersection
-	if (shadow._active) {
+	if (UBOOL_TEST(shadow._active)) {
 		pos /= pos.w;
 		vec3 l = pos.xyz - shadow._light;
 		float d = dot(shadow._point - shadow._light, shadow._normal) / dot(l, shadow._normal);
@@ -56,15 +56,15 @@ void main()
 
 	gl_Position = projectedPos;
 
-	if (shadow._active) {
+	if (UBOOL_TEST(shadow._active)) {
 		Color = vec4(shadow._color, 1.0);
 	} else {
 		Color = color;
 	}
-	if (!useVertexAlpha)
+	if (!UBOOL_TEST(useVertexAlpha))
 		Color.a = 1.0;
 	Color *= uniformColor;
-	if (textured) {
+	if (UBOOL_TEST(textured)) {
 		Texcoord = texcoord;
 	} else {
 		Texcoord = vec2(0.0, 0.0);
@@ -109,7 +109,7 @@ void main()
 		light += lightsColor[i].xyz * intensity;
 	}
 
-	if (!hasAmbient)
+	if (!UBOOL_TEST(hasAmbient))
 		light += vec3(0.5, 0.5, 0.5);
 	light /= max(1.0, max(max(light.x, light.y), light.z));
 	Color *= vec4(light, 1.0);
--- a/engines/grim/shaders/emi_sprite.fragment
+++ b/engines/grim/shaders/emi_sprite.fragment
@@ -2,7 +2,7 @@ in vec2 Texcoord;
 in vec4 Color;
 
 uniform sampler2D tex;
-uniform bool textured;
+uniform UBOOL textured;
 uniform float alphaRef;
 uniform float meshAlpha;
 
@@ -11,7 +11,7 @@ OUTPUT
 void main()
 {
 	outColor = Color;
-	if (textured) {
+	if (UBOOL_TEST(textured)) {
 		vec4 texColor = texture(tex, Texcoord);
 		outColor.rgba *= texColor.rgba;
 		outColor.a *= meshAlpha;
--- a/engines/grim/shaders/emi_sprite.vertex
+++ b/engines/grim/shaders/emi_sprite.vertex
@@ -9,12 +9,12 @@ uniform highp mat4 projMatrix;
 uniform highp mat4 extraMatrix;
 uniform highp mat4 normalMatrix;
 uniform highp vec3 cameraPos;
-uniform bool textured;
-uniform bool useVertexAlpha;
+uniform UBOOL textured;
+uniform UBOOL useVertexAlpha;
 uniform vec4 uniformColor;
 
 struct shadow_info {
-	bool _active;
+	UBOOL _active;
 	vec3 _color;
 	vec3 _light;
 	vec3 _point;
@@ -43,15 +43,15 @@ void main()
 
 	gl_Position = projectedPos;
 
-	if (shadow._active) {
+	if (UBOOL_TEST(shadow._active)) {
 		Color = vec4(shadow._color, 1.0);
 	} else {
 		Color = color;
 	}
-	if (!useVertexAlpha)
+	if (!UBOOL_TEST(useVertexAlpha))
 		Color.a = 1.0;
 	Color *= uniformColor;
-	if (textured) {
+	if (UBOOL_TEST(textured)) {
 		Texcoord = texcoord;
 	} else {
 		Texcoord = vec2(0.0, 0.0);
--- a/engines/grim/shaders/grim_actor.fragment
+++ b/engines/grim/shaders/grim_actor.fragment
@@ -3,8 +3,8 @@ in vec4 Color;
 
 uniform sampler2D tex;
 uniform sampler2D texZBuf;
-uniform bool textured;
-uniform bool hasZBuffer;
+uniform UBOOL textured;
+uniform UBOOL hasZBuffer;
 uniform vec2 texcropZBuf;
 uniform vec2 screenSize;
 uniform float alphaRef;
@@ -35,10 +35,10 @@ void checkZBuffer()
 
 void main()
 {
-	if (hasZBuffer)
+	if (UBOOL_TEST(hasZBuffer))
 		checkZBuffer();
 	outColor = Color;
-	if (textured) {
+	if (UBOOL_TEST(textured)) {
 		outColor *= texture(tex, Texcoord);
 	}
 
--- a/engines/grim/shaders/grim_actor.vertex
+++ b/engines/grim/shaders/grim_actor.vertex
@@ -11,12 +11,12 @@ uniform highp mat4 modelMatrix;
 uniform highp mat4 viewMatrix;
 uniform highp mat4 projMatrix;
 uniform highp mat4 extraMatrix;
-uniform bool textured;
-uniform bool lightsEnabled;
+uniform UBOOL textured;
+uniform UBOOL lightsEnabled;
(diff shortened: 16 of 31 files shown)


raziel-:
@lephilousophe

Thank you so much for this.
I'll build and try as soon as I'm back home

lephilousophe:
So, after testing by @raziel-, OGLES2 is not ready to work with ScummVM.
Mainly because of the shader compiler not supporting boolean uniforms and some bugs in the rendering.
The other option available is to create two builds, one with miniGL and one using GL4ES. This needs testing and switching cannot be done dynamically.

raziel-:
..yet (at least i hope)

aquadran:
> So, after testing by @raziel-, OGLES2 is not ready to work with ScummVM.
> Mainly because of the shader compiler not supporting boolean uniforms and some bugs in the rendering.
> The other option available is to create two builds, one with miniGL and one using GL4ES. This needs testing and switching cannot be done dynamically.

boolean uniforms can be replaced. what other issues in the rendering?

raziel-:
@aquadran 

That would be possible?
Really?
I thought it might have to do with portability to other platforms.
That would be great because I doubt my gfx driver will get boolean uniform support anytime soon.

About the rendering issue I'll have to get back to you, I dont really remember which backend caused what issues (switching around backends, on my end, that much without really having the time to inspect is not really working)
Unfortunately work is taking away even more spare time now.

Once I can use (or try) the shaders without boolean uniforms I can do much more testing and report back...if that is fine with you?

Thank you very much

lephilousophe:
@raziel- After some chat with @aquadran I patched the boolean uniforms to make it more compatible for you.
It's macro based so other platforms should be unaffected.
In addition I had left some ifdef which would prevent the proper resizing of the image. The framebuffer object was never created but the engine was still drawing at the resolution it expects. This should be fixed.
For the Grim Fandango bug you reported, @aquadran told me it may be due to a bug in the shaders and not your driver. Could you confirm it's not wrong everywhere? How is the look of Manny at the game start?

aquadran:
@raziel- I suggest to check monkey4 instead grim fandango

raziel-:
@lephilousophe 
@aquadran 

Thank you very much both of you.
I'll try as soon as I can...unfortunately that wont be before 11th of March (hopefully). 

I hope that is ok?
Sorry for the long waiting time

Any reason why Monkey 4 is better suited for testing?

aquadran:
@raziel- for monkey4 renderer doesn't access zbuffer directly.
You can also check other games.

raziel-:
Unfortunately I have to postpone this.

I need SDL_SetWindouseRect to get ScummVM linked, which is not part of the current SDL2_gl4es release.

Once a new release is available I'll get back to you.

Thank you

kas1e:
@all

I helping Raziel a bit to with scummvm on amigaos4 as i am one who made gl4es sdk for amigaos4 and use it a lot, together with all that opengl/warp3dnova on amigaos4.

So, tried to get things in the shape, and need to ask some questions there (hope that topic will be correct place for).

For first, yeah, in our 3d drivers we do not have boolean uniforms support. But that not a problem, as we just workaround them with "ints". Like for example myst3_box.fragment original is:

```
in vec2 Texcoord;

OUTPUT

uniform bool textured;
uniform vec4 color;
uniform sampler2D tex;

void main()
{
	outColor = color;
	if (textured)
		outColor = outColor * texture(tex, Texcoord);
} 
```

But then, our modified versio is just:

```
in vec2 Texcoord;

OUTPUT

uniform int textured;
uniform vec4 color;
uniform sampler2D tex;

void main()
{
	outColor = color;
	if (textured!=0)
		outColor = outColor * texture(tex, Texcoord);
} 
```

I.e. just swap ```bool``` on ```int```, and adding ```!=0``` where need it. 

Not sure if it correctly to ask at all, but can be just modify all the shaders in the ScummVM, so they will have no "bool" ? Logical everything will works the same as before, just more plaforms will be supported without any workarounds where bool is not implemented.

Next, to issues i noticed for now.

First one which i note, that when i configure scummvm, i have in configure:

```
Checking for OpenGL... yes (runtime detection)
Checking for OpenGL for game... yes (shaders enabled)
```

All ok, we can have opengl and opengl with shaders.

Now, if we will try to run for example GRIM game, with settings 3D-rederer "pure OpenGL" (i.e. Opengl _without_ shaders), it still, will use inbuild box.vertex shader in the backends/graphics3d/opengl/surfacerenderer.cpp  , because USE_OPENGL_SHADERS were set. But then, we choose 3d-rendering "pure" Opengl, so why we even go to that "check" ? Should't we somehow separate it, like, if we choose "pure" opengl for 3d rendering, then its not the same as "opengl with shaders", right ?

So should't it be like this : if 3d renderer for the games choosen as pure "opengl" , then no code should be executed with meaning "opengl_with_shaders", but only, when we choose "opengl_with_shaders" for 3d renderer ? For now it looks like that does not matter what, we still have executed that inbuild part in backends/graphics3d/opengl/surfacerenderer.cpp , just because configure found that we can do opengl with shaders. While, it imho looks unlogical that when we choose 3d renderer as pure opengl, we still have execute that part.

But ok, i then just fix this shader , and replaced there BOOL on INT, and when i run GRIM, i have everything rendered, but not the main person. Can anybody from scummvm team test this with changed shader from backends/graphics3d/opengl/surfacerenderer.cpp, to be that:

```
static const char *boxVertex =
	"attribute vec2 position;\n"
	"attribute vec2 texcoord;\n"
	"uniform vec2 offsetXY;\n"
	"uniform vec2 sizeWH;\n"
	"uniform vec2 texcrop;\n"
	"uniform int flipY;\n"
	"varying vec2 Texcoord;\n"
	"void main() {\n"
		"Texcoord = texcoord * texcrop;\n"
		"vec2 pos = offsetXY + position * sizeWH;\n"
		"pos.x = pos.x * 2.0 - 1.0;\n"
		"pos.y = pos.y * 2.0 - 1.0;\n"
		"if (flipY!=0)\n"
			"pos.y *= -1.0;\n"
		"gl_Position = vec4(pos, 0.0, 1.0);\n"
	"}\n";
```

(there i only change ```bool flipY``` on ```int flipY``` and change ```if (flipY)``` on ```if (flipY!=0)```).

Will you have the same issue in GRIM with missing main person ?

And the last question. Where are the paths to the shaders are if want to place them together with distributive ? I mean, i want to try to fix all those "bools" for all shaders for all games (myst3, grim, etc), to see how it will reacts when i will choose opengl_with_shaders, but so far tried "shaders/grim3", "grim3"  and some others without much luck. Is there some default path from which shaders will be taken for each game they need ?

Thanks!

kas1e:
@all
And just to clarify things a bit (that will help us later to understand problem if they will arise) : boolean uniform is not supported by our 3d drivers called Warp3dNova, while opengles2.library (our ogles2 implementation) work over it. So, if warp3dnova didn't support it, ogles2 also can do nothing (as ogles2 use glsnalgvalidator inbuild in to compile shaders to spirv format, then, transfer it to warp3dnova, and if warpdnova fail to compile them, then we fail).

kas1e:
@all
Sorry for mail flood, but another question. But i see we also have commit in this topic with "AMIGAOS: Let user select the OpenGL implementation", which have inside checks on GL (and set it to 1.3 as it minigl),  and if gles2, then to GL ES2. But then, we also have GL4ES , which is 2.1 (while still being "gl") ..

How this is selected ? Maybe it by default just already sets to 1.3 (even if we use GL4ES which 2.1, and so missing functionality) ?

lephilousophe:
Hello @kas1e,

First, about the uniform bool in game shaders: I think it's cleaner to let implementations supporting them make use of it. I don't know if the int uniforms are as efficient as booleans ones.

For the ShaderSurfaceRenderer, it's defined when shaders are supported but never instantiated if context doesn't support them [here](https://github.com/scummvm/scummvm/blob/master/backends/graphics3d/opengl/surfacerenderer.cpp#L37).
Indeed you are right, I overlooked the bool in this shader. I will push a fix shortly.

For the shaders path, you can look at this PR. I already added the macro everywhere we use booleans.
These files will get picked if you run scummvm from inside your source tree.
Else, they are installed next to all other data files. On AmigaOS we never copy them because I overlooked this.
I will fix this as well.

For the dynamic selection of OpenGL implementation, it only supports miniGL and ogles2 because it's what is supported by SDL2 from [AmigaOS](https://github.com/AmigaPorts/SDL).

GL4ES is not supported currently because it's a complete different implementation and we need to statically link to it. This would need some #ifdef.
If we would like to support the 3 implementations at the same time, one would need to build a dynamic library for GL4ES and merge the SDL support in AmigaPorts one.

kas1e:
@lephilousophe 
Thanks for fast answer !

> For the ShaderSurfaceRenderer, it's defined when shaders are supported but never instantiated if context doesn't support them [here](https://github.com/scummvm/scummvm/blob/master/backends/graphics3d/opengl/surfacerenderer.cpp#L37).
Indeed you are right, I overlooked the bool in this shader. I will push a fix shortly.

I just find it strange that why we even execute "box.vertex" shader if we choose in 3d-renderer "opengl" (but not "opengl with shaders"), i.e. we in any case run this shader. That ok ?

> These files will get picked if you run scummvm from inside your source tree.
> Else, they are installed next to all other data files. On AmigaOS we never copy them because I overlooked this.
> I will fix this as well.

Yeah, creating "engines/grim/shaders" directory in the root of my ScummVM with grim's shaders make them loads up!

> If we would like to support the 3 implementations at the same time, one would need to build a dynamic library for GL4ES and merge the SDL support in AmigaPorts one.

Yeah, sounds logical. So for time being it can be out of question from autobuilds then.

EDIT: and i still can't get what happens with GRIM , it surely works before ( i mean Person were visibly), and i build it with the same gl4es before (just with disabled shaders, if that make difference of course).

raziel-:
@kas1e is talking about the bug i reported via email with Grim Fandango and using gl4es/ogles2/shaders.
Manny is mostly missing/not drawn at the start of the game, wandering around doesn't rectify it (as @leqhilousophe suggested it might be a bug in one of the renderers, or maybe an oversight from the bool macro transition).

Manny is still drawn correctly when using pure OpenGL/MiniGL.

lephilousophe:
> I just find it strange that why we even execute "box.vertex" shader if we choose in 3d-renderer "opengl" (but not "opengl with shaders"), i.e. we in any case run this shader. That ok ?
> 
I see what you mean and @aquadran would know better.
This shader is used by the backend (SDL) and not by the game.
When you select for the game OpenGL without shaders, you mean that the game will render using classic OpenGL calls.
The backend still uses a shader when supported to do the final rendering to the screen. This may be an option to add if it's useful.
Anyway, I added a ifdef for Amiga to not use uniform booleans here.

> EDIT: and i still can't get what happens with GRIM , it surely works before ( i mean Person were visibly), and i build it with the same gl4es before (just with disabled shaders, if that make difference of course).
I have yet to test with latest master but last time I checked (7 days ago), Manny was visible in the underground at startup.
I was testing with French demo version as I don't own the old game (only the remastered unsupported one).

> And just to clarify things a bit (that will help us later to understand problem if they will arise) : boolean uniform is not supported by our 3d drivers called Warp3dNova, while opengles2.library (our ogles2 implementation) work over it. So, if warp3dnova didn't support it, ogles2 also can do nothing (as ogles2 use glsnalgvalidator inbuild in to compile shaders to spirv format, then, transfer it to warp3dnova, and if warpdnova fail to compile them, then we fail).

I see... Maybe the SPIR-V bytecode could be patched after compilation to replace all these uniform booleans by integer ones?
Sounds like a bit complicated but it would remove some compatibility issues.

kas1e:
@raziel
Problem is that i tried to build scummvm now _without_ usage of shaders (i.e. --disable-opengl-game-shaders) and the main person in the GRIM still not rendered. Of course, if --disable-opengl-game-shaders mean complete disabling of shaders.

And i swear it works for me before with GL4ES, when i just add 

```#ifdef __amigaos4__
    g_context.shadersSupported = false;
#endif
```

inside of ```backends/graphics/opengl/context.cpp```, but now it seems didn't help, and person in Grim never rendered. That why i think it can be fresh issue (and not with shaders). Of course, again, if ```--disable-opengl-game-shaders``` disables them completely.

kas1e:
> I see... Maybe the SPIR-V bytecode could be patched after compilation to replace all these uniform booleans by integer ones?
> Sounds like a bit complicated but it would remove some compatibility issues.

Author of Warp3DNova for sure will add boolian support someday, we just don't know when and if. For him it will be more easy to do than to add any kind of patching from our side ,etc. At least macro you do should did the trick for sure for now

kas1e:
@All
So, with help from lephilousophe we find out that issues with wrong rendering in GRIM happens not because of the shaders, but because of the swith to GLAD. I.e. i tested with full shaders disabling the commit before GLAD merge : there GRIM person renders fine, menu when press f1 renders fine too. Then next commit with GLAD, this one:  https://github.com/scummvm/scummvm/commit/21b0d6ded18e6994e43c316759e4ffdb487f244d , make the person in GRIM not renders, and menu calling by pressing "f1" also have nothing to shown.

Now, as GLAD is big thing, it can be very well that it define/not-define something which make things works differently or behave differently on our side.

Question is, how to find out what / where, if all we have now is that issue comes from the GLAD, shaders not involved, and visually we just have main person not renders, and menu calling by f1 not rendered too (sounds like problem with textures/sprites maybe?). 

And also, what is difference with what GLAD have defined by default, and what wasn't enabled before when we have no GLAD merged in ?

kas1e:
@all
So i create separate ticket about issues with GRIM after switching to GLAD to not deral that thread: https://bugs.scummvm.org/ticket/13351

kas1e:
@lephilousophe
I build all your changes you put there over gl4es , and when i run scummvm and choose "opengl" (no shaders, so only inbuild box.vertex will be used)  - i have just a crash in ogles2.library. Stack trace show that crash come from GRIM's opengl_draw_movie_frame().   But should to note, that when i build myself yesterday's ScummVM source , and manually change box.vertex only, then i have no crash. So seems something new because of those new changes.

Maybe that related to fact that in your changes when we have "gl" we set it to 1.3 , and not to 2.1, so that why it crash ? And for gl4es builds i need manually to change it to 2.1 ?

Anyway, i also tried to run the GRIM with "opengl shaders enabled", and using the latest shaders as well, then they used, loads up fine, all compiles, so no more syntax errors for sure, but, still it the same not-rendered normally as we have in that "without_shaders" version because of GLAD change.

But at least in terms of shaders syntax/compilation all is fine now. Now only to understand 2 moments:

1. why we crash now when we run pure "opengl" rendering, but sources from yesterdays main's repo does not
2. why we have issues when switch to GLAD.

lephilousophe:
After some debugging out of Github, the bug occurs because I allow framebuffer support to be built and GL4ES announce support for it.
For an unknown reason, the framebuffer causes a bug when GL4ES flushes commands (especially when calling ogles2.library glDrawArrays) before binding the frame texture.

kas1e:
@lephilousophe

kas1e:
@lephilousophe 
Ok the crash bug about framebuffer fixed inside of GL4ES. So for now all your commits in this ticket seems to be ok for both MiniGL and GL4ES.

For now there some remain issues with rendering persists over GL4ES, so will keep you informed once we fix it.

lephilousophe:
A little update here.
With @kas1e we did a lot of debugging with GL4ES and directly using OGLES2.
To sum up, there are rendering issues with Grim and EMI.
Stark and WME3D seem to render OK (although WME3D doesn't seem to support screen resizing in every platform).

The Grim rendering issue is due to two bugs.
First, when writing the 16-bits depth buffer to the texture using LUM_ALPHA format, fragment shader expected little-endian format. This is fixed in this PR.
Second, the use of texture2D with gl_FragCoord seems flipped vertically while framebuffer is enabled.
A minimal test case is being done and should allow us to report the bug to OGLES2 developer (or fix our shader if the bug comes from ScummVM).

EMI rendering bug (actors rendered in black) hasn't been investigated yet.

lephilousophe:
Yet another update:
GRIM was not rendering this good (lights problems) and EMI had dark shadows instead of actors.
All of this was due to a bug about how we addressed the vectors used for lights settings. Only the first light was set up.
Addressing using brackets is, according to the specification, only for arrays of structures. This is now fixed by setting the whole array in one block.
This revealed another bug in OGLES2 when using VBO and GL_UNSIGNED_BYTE where RGBA was swapped. This should be fixed soon in OGLES2 thanks to @kas1e who reported the bug.
FBO bug with texture reversed is being fixed too.

lephilousophe:
This PR now includes all commits from #3780 and #3781.

With all of these and an unreleased (yet) OGLES2, rendering of Grim, Monkey4, TLJ and Wintermute 3D games should be quite OK.
Some bugs still remain like [#13169](https://bugs.scummvm.org/ticket/13169) and [#13375](https://bugs.scummvm.org/ticket/13375).
First one is quite puzzling and may be a bug in `glCopyTexImage2D` of MiniGL, OGLES2 and MorphOS GL.
Second one is just a missing function in WME3D.

lephilousophe:
@kas1e @raziel- I would like to merge this in the next days.
Normally, with the future update of OGLES2 ScummVM should work like a charm.
For MiniGL it seems it will still have bugs in Myst III.

There are still some checks to do:
- make sure that a user can't be locked down because he doesn't have OGLES2 although he chose it,
- ensure that when a user chooses a 3D rendering mode, if it's not available, a fallback is chosen.

Thanks.

kas1e:
@lephilousophe 
I see you commit already half of things to scummvm, and half of things still left there, so i want tomorrow to do some deep test, and how better i should do them ? Just by downloading this https://github.com/lephilousophe/scummvm/commits/ea521f1b82457d8eba5bdc468eff7610e950e5ca mean everything we need to test will be included ?

Thanks!

lephilousophe:
Yes, this one will do.
Else, you can use this simpler one: https://github.com/lephilousophe/scummvm/tree/amigaos4 :)

raziel-:
@lephilousophe 
@kas1e 

I can see you got everything secured.

I can't test anything before Monday/Tuesday I'm afraid and probably never as thoroughly as kas1e.

Sorry

kas1e:
@[lephilousophe](https://github.com/lephilousophe)

> Else, you can use this simpler one: https://github.com/lephilousophe/scummvm/tree/amigaos4 :)

Tested !

I didn't tried to build "ready arhive for release" (that one for Raziel once he will be about to release it), but i do try to build a binary, and use all the shaders for all engines coming with scummvm by copy them manually to progdir:engines/  , etc.

So i build it like this:

```
./configure --build=x86_64 --host=ppc-amigaos --disable-debug --enable-release --disable-all-engines --enable-engine=grim,myst3,monkey4,stark,wintermute,wme3d,groovie2,trecision,asylum,sword25,ngi
```

Now, after some tests, that what i can say:


1. Setting "default" in backend and restart scummvm resulting in MiniGL being used as default, while it wrong, the better driver should be used instead : so, if we have working ogles2 , then default should be ogles2. If we don't have ogles2, then default should be minigl. 

2. When choose "opengl1.3" renderer ,  and then try to choose opengl or opengl_with_shaders , it fallback to the minigl in both cases and not crashes, that good. Through, do we need to show "opengl_with_shaders" option at all, if we do choose ogl1.3 backends ?

3. When choose "opengles2" renderer, and the try to choose openg or opengl_with_shaders, it always fallback to ogles2, no crash, but then, should we have there "opengl" mode at all then in 3D renderer list ?  And anyway, should't pure "opengl" render also works on ogles2 backend ?

4. For Ogles2 backend i always have a warning from scummvm saying that:

```
WARNING: Could not parse GL version 'OpenGL ES 2 3.2 on top of Warp3D Nova 54.13'!
```

5, Hardcore thing : a crash. To reproduce  i do that : 
 -- run scummvm with "opengl 1.3" backed being set before. 
 -- go to backend tab and choose "opengles2" . Hit "Apply" and then "OK"
 -- go to "The Last Journey game", and choose there Graphics Mode : OpenGL and Game 3D Renderer "OpenGL with Shaders"
-- click run => crash

Crash happens because it fallback to opengl 1.3 still, even if i set in "backend" ogles2.  and "The Last Journey" can't even works on minigl. That ok, but user will be a bit mesed in head wtf happens.

So, my suggestion is : if we will keep that "Backend" thing (but should we ? Isn't it easer on run to check if we have ogles2, then use ogles2, if we have minigl, then use minigl, and that it ?)  , then add a warning window inside of scummvm itself (i.e. not console one, but window inside of scummvm window), arise after one change backend, saying "chaning backend requere a ScummVM restarting, or you will have issues and fallback to old backend".

Alrenatively, whih can be more easy, maybe when we choose a backend in the ScummVM, it will just "auto-restart" ? Or, bring a window like "you will be no auto-restarted, yes/no ?". 

Something of that sort maybe ?

6.  And tested with ogles2 backend and opengl with shaders all our games : GRIM, Monkey, Myst3, etc, etc, so far on new beta of ogles2 everything seems good.

lephilousophe:
> @[lephilousophe](https://github.com/lephilousophe)
> 
> > Else, you can use this simpler one: https://github.com/lephilousophe/scummvm/tree/amigaos4 :)
> 
> Tested !
Cool! Thank you!

> 
> I didn't tried to build "ready arhive for release" (that one for Raziel once he will be about to release it), but i do try to build a binary, and use all the shaders for all engines coming with scummvm by copy them manually to progdir:engines/ , etc.
> 
> So i build it like this:
> 
> ```
> ./configure --build=x86_64 --host=ppc-amigaos --disable-debug --enable-release --disable-all-engines --enable-engine=grim,myst3,monkey4,stark,wintermute,wme3d,groovie2,trecision,asylum,sword25,ngi
> ```
> 
> Now, after some tests, that what i can say:
> 
>     1. Setting "default" in backend and restart scummvm resulting in MiniGL being used as default, while it wrong, the better driver should be used instead : so, if we have working ogles2 , then default should be ogles2. If we don't have ogles2, then default should be minigl.

Well, default meant the default of SDL2 from AmigaPorts.
I could make it fallback automatically if it works.

> 
>     2. When choose "opengl1.3" renderer ,  and then try to choose opengl or opengl_with_shaders , it fallback to the minigl in both cases and not crashes, that good. Through, do we need to show "opengl_with_shaders" option at all, if we do choose ogl1.3 backends ?

Indeed, this is an overlook from my PR related to renderer. I will add this feature.

> 
>     3. When choose "opengles2" renderer, and the try to choose openg or opengl_with_shaders, it always fallback to ogles2, no crash, but then, should we have there "opengl" mode at all then in 3D renderer list ?  And anyway, should't pure "opengl" render also works on ogles2 backend ?

Same as above.
"Pure OpenGL" (that we call OpenGL classic in the team) is OpenGL 1.x or 2.x code.
OpenGL ES is completely different: it's inspired by OpenGL non-ES but it's really lightened and uses a shader based pipeline which came with OpenGL 3 while OpenGL 1.x and 2.x provided way more functions but it also was a fixed pipeline.
Note how GL4ES needs to do many things to make OpenGL 2.1 code work on OpenGL ES 2.
Our "OpenGL with shaders" renderer works both on OpenGL 3.x and OpenGL ES 2.0 implementations.
Our "OpenGL" renderer works only on OpenGL non-ES implementations.
So this last one cannot work on OGLES2 because many functions are missing.

> 
>     4. For Ogles2 backend i always have a warning from scummvm saying that:
> 
> 
> ```
> WARNING: Could not parse GL version 'OpenGL ES 2 3.2 on top of Warp3D Nova 54.13'!
> ```

This is a limitation of our parser. We expect the OpenGL ES version in the form major.minor while here it is major only.
The OpenGL ES2 specification is quite elusive about the version format unlike all other versions where it's more formally specified.
Anyway it let me discover that our parser doesn't parse OpenGL ES 1.x versions correctly.
I will try to do some patch for this outside of this PR.

In any case, the version is checked only in OpenGL (non ES) case so it doesn't really matter.

> 
> 5, Hardcore thing : a crash. To reproduce i do that : -- run scummvm with "opengl 1.3" backed being set before. -- go to backend tab and choose "opengles2" . Hit "Apply" and then "OK" -- go to "The Last Journey game", and choose there Graphics Mode : OpenGL and Game 3D Renderer "OpenGL with Shaders" -- click run => crash
> 
> Crash happens because it fallback to opengl 1.3 still, even if i set in "backend" ogles2. and "The Last Journey" can't even works on minigl. That ok, but user will be a bit mesed in head wtf happens.

Changing implementation on restart is expected and falling back also.
The crash should not happen: I suspect TLJ uses an OpenGL call not available on MiniGL. This should be checked at run time and display an error message instead of hard crashing.
Can you provide me a call stack?

> 
> So, my suggestion is : if we will keep that "Backend" thing (but should we ? Isn't it easer on run to check if we have ogles2, then use ogles2, if we have minigl, then use minigl, and that it ?) , then add a warning window inside of scummvm itself (i.e. not console one, but window inside of scummvm window), arise after one change backend, saying "chaning backend requere a ScummVM restarting, or you will have issues and fallback to old backend".

Well, it really depends on what you plan to do with OpenGL.
If I am correct, there are three implementations usable on AmigaOS:
- MiniGL (OpenGL 1.3, you said it's obsolete and noone works on it),
- OGLES2 (OpenGL ES 2.0, paid),
- GL4ES (OpenGL 2.1, free but over OGLES which is paid).

First question: is there a point of using GL4ES? It needs OGLES2 which we already support.
If there is a place for it, do you plan to add its support on AmigaPorts SDL by creating a GL4ES library and loading it dynamically?
Second question: is there a good reason to prefer MiniGL overs OGLES2 when both are present?

If there is a reason to let the user choose the implementation, we should keep this backend tab.
I could indeed add a message box to make the warning more explicit.

> 
> Alrenatively, whih can be more easy, maybe when we choose a backend in the ScummVM, it will just "auto-restart" ? Or, bring a window like "you will be no auto-restarted, yes/no ?".
> 
> Something of that sort maybe ?

I suspect this would need more specific code in the middle of common one so I am not really fond of this. Current implementation is quite straightforward and doesn't modify common code.

> 
>     6. And tested with ogles2 backend and opengl with shaders all our games : GRIM, Monkey, Myst3, etc, etc, so far on new beta of ogles2 everything seems good.

Great! :)

kas1e:
> Changing implementation on restart is expected and falling back also.
> The crash should not happen: I suspect TLJ uses an OpenGL call not available on MiniGL. This should be checked at run time and display an error message instead of hard crashing.

Did you mean checked inside of The Long Journey's code ?

> Can you provide me a call stack?

Of course. All i do there, is just choose opengl backend, and trying to run The Longest Journey

```
Stack trace:
(0x603B28E0) 0x00000000 [cannot decode symbol]
(0x603B28F0) scummvm:_ZN5Stark3Gfx12OpenGLDriver4initEv()+0x48 (section 1 @ 0x3DF7B4)
(0x603B29F0) scummvm:_ZL7runGamePK6PluginS1_R7OSystemRKN6Common6StringE()+0x1004 (section 1 @ 0x936E0)
(0x603B2C20) scummvm:scummvm_main()+0x1ec4 (section 1 @ 0x96548)
(0x603B30F0) scummvm:main()+0x160 (section 1 @ 0x8F6A8)
(0x603B3D10) native kernel module newlib.library.kmod+0x00002614
(0x603B3D60) native kernel module newlib.library.kmod+0x00003340
(0x603B3F10) native kernel module newlib.library.kmod+0x00003864
(0x603B3F40) scummvm:_start()+0x1e0 (section 1 @ 0x1B3C)
(0x603B3F90) native kernel module dos.library.kmod+0x0002a490
(0x603B3FC0) native kernel module kernel+0x0005ea84
(0x603B3FD0) native kernel module kernel+0x0005eafc
```


> First question: is there a point of using GL4ES? It needs OGLES2 which we already support.

Of course not, we worry with GL4ES before only because of not having proper support from ogles2 side, so now when ogles2 side is fully ok, we not need GL4ES of course.


> If there is a place for it, do you plan to add its support on AmigaPorts SDL by creating a GL4ES library and loading it dynamically?

I think someday we will made it as shared library (as there more and more software which use GL4ES, and recompile them all the time once some major improvements done in GL4ES kind of pain). But in case with scummvm we have no needs to worry about GL4ES anymore, IMHO.

> Second question: is there a good reason to prefer MiniGL overs OGLES2 when both are present?

Of course not. MiniGL is old, buggy, and slow implementation. If ogles2 way works, we should use this one by default of course. And only use MiniGL one when OGLES2 is not found.

> If there is a reason to let the user choose the implementation, we should keep this backend tab.
> I could indeed add a message box to make the warning more explicit.

I think it only needs for some testing purposes, so do not know for what elese .. I assume this "backend" thing was added because of all that mess with minigl/gl4es/ogles2/ etc /etc, but now it's just 2 of them, minigl and ogles2. Ogles2 is main and best, and minigl just when there no ogles2. Manual setting can be keept maybe (just in case something somewhere going wrong), but default need to be ogles2 if one can be used of course. And yeah, warning will be enough for sure as well

Thanks!

lephilousophe:
I think I fixed the TLJ crash on miniGL.
The engine made use of `glBindBuffer` which is not available on old OpenGL implementations.

lephilousophe:
> > First question: is there a point of using GL4ES? It needs OGLES2 which we already support.
> 
> Of course not, we worry with GL4ES before only because of not having proper support from ogles2 side, so now when ogles2 side is fully ok, we not need GL4ES of course.
> 
> > If there is a place for it, do you plan to add its support on AmigaPorts SDL by creating a GL4ES library and loading it dynamically?
> 
> I think someday we will made it as shared library (as there more and more software which use GL4ES, and recompile them all the time once some major improvements done in GL4ES kind of pain). But in case with scummvm we have no needs to worry about GL4ES anymore, IMHO.
> 
> > Second question: is there a good reason to prefer MiniGL overs OGLES2 when both are present?
> 
> Of course not. MiniGL is old, buggy, and slow implementation. If ogles2 way works, we should use this one by default of course. And only use MiniGL one when OGLES2 is not found.
> 
> > If there is a reason to let the user choose the implementation, we should keep this backend tab.
> > I could indeed add a message box to make the warning more explicit.
> 
> I think it only needs for some testing purposes, so do not know for what elese .. I assume this "backend" thing was added because of all that mess with minigl/gl4es/ogles2/ etc /etc, but now it's just 2 of them, minigl and ogles2. Ogles2 is main and best, and minigl just when there no ogles2. Manual setting can be keept maybe (just in case something somewhere going wrong), but default need to be ogles2 if one can be used of course. And yeah, warning will be enough for sure as well

Here is what I've just committed:
- we try OGLES2 first,
- we then fallback on MiniGL if OGLES2 loading failed,
- if none of them worked, we clear all settings and try to continue,
- but if `opengl_implementation` is set, only this implementation is tried and if loading failed, we clear all settings and try to continue.

I removed the backends tab and this setting is not exposed anymore.

If you think it's not a good idea, I still have the old code.

kas1e:
For me looks good !

Is ```but if opengl_implementation is set, only this implementation``` , mean that we can manually set when we want minigl/ogles2 (like backend tab before) ? 

Will test everything now

kas1e:
PS.

> if none of them worked, we clear all settings and try to continue,

mean "continue to have software only", right ?

kas1e:
Ok tested !

-- automatic mode detection works. when i have ogles2.library installed, it use ogles2, if i remove that library, then it use minigl.
-- ```opengl_implementation``` in scummvm.ini still works as expected and can be that "backend" tab kind, so yeah, removing this tab was ok IMHO.

Thanks! Looks good enough IMHO.

lephilousophe:
I just created #3798 to hide the unavailable renderers.

kas1e:
> I just created https://github.com/scummvm/scummvm/pull/3798 to hide the unavailable renderers.

Yes, that one also works fine. When we on ogles2 we have only "opengl with shaders", when we on minigl we have "opengl".

lephilousophe:
And I just created #3799 to fix the version parsing.
This PR includes both #3798 and #3799 to ease testing.

kas1e:
> And I just created https://github.com/scummvm/scummvm/pull/3799 to fix the version parsing.

Yes, works fine as well :)

Btw, i noticed another moment on amigaos4 side, but this time it's about GUI of launcher. Not sure through to what it related.. For first, on running i have ``"WARNING: You are missing a valid 'translations.dat' file. GUI translation will not be available!```, through i do have ```translation.dat``` in both ```themes``` and ```extras``` dirs (i just copy it to both places just in case).

But then, i also have some visuall differences in gui. For first, as i can see, on Win32 we have button "Options" , while on amigaos4 we have button on the same place with words "Global Options...". For me, "Global Options..." looks better, just strange why on Win32 it's pure "Options".

Next, on win32 button for game's option called "Edit Game...", while on amigaos4 it's called "Game Options...", which also fine, just strange why different..

But my main concert and about suboptions in "Graphics" tab be it game options or global options : Instead of words "Game 3D Renderer:" (to choose opengl or opengl with shaders or software), on amigaos4 we have "Game 3...enderer:" , i.e. 3 dots insted of ```D```, then space, then ```r```. 

We also didn't have there competely entry ```3D Anti-Aliasing``` , but i assume that can be because of something missing in ogles2 ?

I assume that differences between EditGame/GameOptions , Global Options/Options are about this warning about missing translation.dat and some fallback to defaults ?  but issue with missing ```D r``` and ```3D Anti-Aliasing``` - dunno.

kas1e:
Hm, i noticed we broke something with new changes:

We fail to set properly things we have, and instead it return me now for GRIM:

{{{
OpenGL Red Bits: 0
OpenGL Green Bits: 0
OpenGL Blues Bits: 0
OpenGL Alpha Bits: 0
Z buffer depth bits: 16
Double Buffer: 0
Stencul buffer bits: 8
}}}

While before it was:

{{{
INFO: OpenGL Red bits: 8
INFO: OpenGL Green bits: 8
INFO: OpenGL Blue bits: 8
INFO: OpenGL Alpha bits: 8
INFO: OpenGL Z buffer depth bits: 24
INFO: OpenGL Double Buffer: 1
INFO: OpenGL Stencil buffer bits: 8
}}}

For Myst3 , it is now have "Double Buffer: 0", while before it was "Double Buffer :1".

It's not "gl parsing" change which cause this, but some previous change...

I will try to rollback changes to see which one start causing this

lephilousophe:
> > And I just created #3799 to fix the version parsing.
> 
> Yes, works fine as well :)
> 
> Btw, i noticed another moment on amigaos4 side, but this time it's about GUI of launcher. Not sure through to what it related.. For first, on running i have ``"WARNING: You are missing a valid 'translations.dat' file. GUI translation will not be available!`, through i do have `translation.dat``` in both ```themes``` and ```extras``` dirs (i just copy it to both places just in case).

Did you set up the themes and extras path in Global Options > Paths?

> 
> But then, i also have some visuall differences in gui. For first, as i can see, on Win32 we have button "Options" , while on amigaos4 we have button on the same place with words "Global Options...". For me, "Global Options..." looks better, just strange why on Win32 it's pure "Options".
> 
> Next, on win32 button for game's option called "Edit Game...", while on amigaos4 it's called "Game Options...", which also fine, just strange why different..

These two changes are because you are on master where we reworded the buttons. You must use a 2.5 version on Win32 and that's why it's different.

> 
> But my main concert and about suboptions in "Graphics" tab be it game options or global options : Instead of words "Game 3D Renderer:" (to choose opengl or opengl with shaders or software), on amigaos4 we have "Game 3...enderer:" , i.e. 3 dots insted of `D`, then space, then `r`.

Which graphical mode do you use in Global Options > Graphics? Is it OpenGL or SDL Surface?
You should use OpenGL (or default which is OpenGL).

> 
> We also didn't have there competely entry `3D Anti-Aliasing` , but i assume that can be because of something missing in ogles2 ?

This is indeed because OGLES2 doesn't seem to support anti-aliasing.

lephilousophe:
> Btw, i noticed that with pure OGLES2 we have that in GRIM
> 
> ```
> OpenGL Red Bits: 0
> OpenGL Green Bits: 0
> OpenGL Blues Bits: 0
> OpenGL Alpha Bits: 0
> Z buffer depth bits: 16
> Double Buffer: 0
> Stencul buffer bits: 8
> ```
> 
> While before with GL4ES it was:
> 
> ```
> INFO: OpenGL Red bits: 8
> INFO: OpenGL Green bits: 8
> INFO: OpenGL Blue bits: 8
> INFO: OpenGL Alpha bits: 8
> INFO: OpenGL Z buffer depth bits: 24
> INFO: OpenGL Double Buffer: 1
> INFO: OpenGL Stencil buffer bits: 8
> ```

We get these values directly from SDL. I don't know why it fails to read.
It's either a problem in SDL or OGLES2.

> For Myst3 , it is now have "Double Buffer: 0", while before it was "Double Buffer :1".
> 
> Not sure if those things are correct ? What you have on your setup ?

I am not sure it's really relevant as many settings are at 0.

> 
> And i found that GRIM have back some issue : when i out on the street, there is no visibly main hero. And in console i have : WARNING: SetActorShadowValid(-1) unknown purpose!
> 
> I test few different ogles2 versions (old ones, next ones, new ones), and bug seems there. Can it be because of those options being set like that ?

I have the same bug as well using OpenGL with shaders.
It's not something OGLES2 related. (pffiuh!)
The warning is present in OpenGL classic too although everything works.

GRIM with shaders support is still a bit unstable.

kas1e:
> Did you set up the themes and extras path in Global Options > Paths?
> Which graphical mode do you use in Global Options > Graphics? Is it OpenGL or SDL Surface?
> You should use OpenGL (or default which is OpenGL).

Oh, i was enough to just set Global Options -> Paths -> Extras, hit "Apply" and then font somehow refreshes, and issue with 3 dots disappear. See how it was while i didn't set neessary extra's path:

https://kas1e.mikendezign.com/aos4/scummvm/scummvm_dots.jpg

Through, that didn't help with "translations.dat" file, it still bring a warning.

I then checked if i have "translations.dat" file in the extras dir: yes, it here. But still it bring such a warning.

> I have the same bug as well using OpenGL with shaders.
> It's not something OGLES2 related. (pffiuh!)
> The warning is present in OpenGL classic too although everything works.

Good that it's on ogles2 only : ) 

> GRIM with shaders support is still a bit unstable.

Well i am 100% sure before when we test things with you when i out on the street it was working (maybe with disabled FBO? i need to check that). Now i just build everything as it, without any changes from my side, and so have that strange issue, which i have before when there were "depth" issue because of flippingY issue, but then it's gone all fine, but when i out on street, the hero disappear. 

I will try to recheck with FBO disabled, to see if it related. Maybe another (i hope last) bug :)

kas1e:
@[lephilousophe](https://github.com/lephilousophe)
Btw, i asked our SDL2 developer about why things is set to 0,0,0 etc,etc and that what he says:

ScummVM doesn't seem to check return values : https://github.com/scummvm/scummvm/blob/f11295eddaeffeaa16124a192e5687537524b70b/backends/graphics3d/openglsdl/openglsdl-graphics3d.cpp#L301

https://wiki.libsdl.org/SDL_GL_GetAttribute

Also setters are not checked: https://github.com/scummvm/scummvm/blob/90f2ff2532ca71033b4393b9ce604c9b0e6cafa0/backends/graphics/openglsdl/openglsdl-graphics.cpp#L48

lephilousophe:
Yes, the getters are not really important: it's only debug stuff.
For the setters, we try to set things up and if we get a usable context, then, we make use of it.

It's not that clean but it works, so except that you get strange debug information, it's not really a problem. :)

kas1e:
What worry me, is why when choose myst3 for example, we do have 24 for depth, but for grim we do have 16. Is those are set inside of the engines of those games itself ? I.e. that not common, but per-game settings in engine's code itself ? 

I just triying to find, what the differenes can be, that we didn't have hero rendered when we go out on street. 

So what i do now, is again build scumm vm without FBO and know what ? It return me now:

```
INFO: OpenGL Red bits: 8
INFO: OpenGL Green bits: 8
INFO: OpenGL Blue bits: 8
INFO: OpenGL Alpha bits: 8
INFO: OpenGL Z buffer depth bits: 24
INFO: OpenGL Double Buffer: 0
INFO: OpenGL Stencil buffer bits: 8
```

Without FBO, its just all bits 0, and depth 16. Very interesting ..

kas1e:
As we discuss with [lephilousophe](https://github.com/lephilousophe) turns out that issue with GRIM happens on Linux as well, so it's something general in the GRIM's opengl-shader code when FBO is used. The return values of the bits/depth/etc there have no meaning.

So by far all can be commited from this discussion to main trunk imho.

lephilousophe:
The GRIM bug has just been fixed. It was a flaw in engine logic and not in OpenGL stuff.
@BeWorld2018 did you try this branch?
I would like to merge it.

BeWorld2018:
I will try

BeWorld2018:
1. I build amigaos4 branch.
- OpenGL (Game 3D renderer) is detect/present, so is good :-)

image

2. Other little thing with GRIM/MONKEY4, i found a little bug with FPS draw (seem affect only MorphOS)

https://github.com/scummvm/scummvm/blob/master/engines/grim/gfx_opengl.cpp#L1837

if i put -1 (nearVal param), all working, but dont know if important for other system :

```
#ifdef __MORPHOS__
	glOrtho(0, _screenWidth, _screenHeight, 0, -1, 1);
#else
	glOrtho(0, _screenWidth, _screenHeight, 0, 0, 1);
#endif
```

lephilousophe:
For this, I propose you see with @aquadran. I am not an OpenGL expert and even less about projections.

As this PR works for you, I merge.

PR #3681, opened 2022-01-25
GUI: rename options buttons

Die4Ever:
Rename Edit Game to Game Options and Options to Global Options to make it less confusing.

Remove ellipsis on launcher buttons.

### DIFF
--- a/gui/launcher.cpp
+++ b/gui/launcher.cpp
@@ -227,7 +227,7 @@ void LauncherDialog::build() {
 #endif
 	if (!g_system->hasFeature(OSystem::kFeatureNoQuit))
 		new ButtonWidget(this, _title + ".QuitButton", _("~Q~uit"), _("Quit ScummVM"), kQuitCmd);
-	new ButtonWidget(this, _title + ".AboutButton", _("A~b~out..."), _("About ScummVM"), kAboutCmd);
+	new ButtonWidget(this, _title + ".AboutButton", _("A~b~out"), _("About ScummVM"), kAboutCmd);
 	if (g_system->getOverlayWidth() > 320)
 		new ButtonWidget(this, _title + ".OptionsButton", _("Global ~O~ptions..."), _("Change global ScummVM options"), kOptionsCmd);
 	else
--- a/gui/launcher.cpp
+++ b/gui/launcher.cpp
@@ -228,7 +228,7 @@ void LauncherDialog::build() {
 	if (!g_system->hasFeature(OSystem::kFeatureNoQuit))
 		new ButtonWidget(this, _title + ".QuitButton", _("~Q~uit"), _("Quit ScummVM"), kQuitCmd);
 	new ButtonWidget(this, _title + ".AboutButton", _("A~b~out..."), _("About ScummVM"), kAboutCmd);
-	if (g_system->getOverlayWidth() > 440)
+	if (g_system->getOverlayWidth() > 320)
 		new ButtonWidget(this, _title + ".OptionsButton", _("Global ~O~ptions..."), _("Change global ScummVM options"), kOptionsCmd);
 	else
 		new ButtonWidget(this, _title + ".OptionsButton", _c("Global ~O~pts...", "lowres"), _("Change global ScummVM options"), kOptionsCmd);
@@ -982,7 +982,7 @@ void LauncherSimple::build() {
 	_loadButton = loadButton;
 
 	// Add edit button
-	if (g_system->getOverlayWidth() > 410) {
+	if (g_system->getOverlayWidth() > 320) {
 		_editButton =
 			new ButtonWidget(this, "Launcher.EditGameButton", _("~G~ame Options..."), _("Change game options"), kEditGameCmd);
 	} else {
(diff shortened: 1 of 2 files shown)


Die4Ever:
image

Thunderforge:
I don't think that this is a good change. Using ellipses to indicate that additional information will pop up (as opposed to an action being performed) has been a standard of UI design since Mac OS 1.0 (technically "System 1.0") in 1984 and Windows 1.0 in 1985 ([source](https://uxdesign.cc/dot-dot-dot-7ce6170bfc7f)). It is still used to this day.

Having ScummVM behave differently than other contexts can be confusing for the user. I think it should stay as-is.

Die4Ever:
> I don't think that this is a good change. Using ellipses to indicate that additional information will pop up (as opposed to an action being performed) has been a standard of UI design since Mac OS 1.0 (technically "System 1.0") in 1984 and Windows 1.0 in 1985 ([source](https://uxdesign.cc/dot-dot-dot-7ce6170bfc7f)). It is still used to this day.
> 
> Having ScummVM behave differently than other contexts can be confusing for the user. I think it should stay as-is.

what about the wording changes? I've asked people to try out scummvm and stream it on twitch, I've watched them, and 100% of my testers have mixed up the game setttings vs global settings because they weren't clearly labeled

Thunderforge:
I am actually okay with the wording changes, but I think that should be its own PR.

A good rule of thumb I've learned: if you ever have to use the word "and" in a PR (or in this case, two paragraphs), it should be split up.

Die4Ever:
> I am actually okay with the wording changes, but I think that should be its own PR.
> 
> A good rule of thumb I've learned: if you ever have to use the word "and" in a PR (or in this case, two paragraphs), it should be split up.

if this was 2 separate PRs then they would've conflicted with each other lol

Thunderforge:
In that case, I think it should have either been presumed that you would have to resolve a merge conflict, or you submit one, wait for it to be declined or accepted, then submit the other one. 

As-is, you're asking us to evaluate a package deal of two changes. I like one of the changes but not the other one, so I am opposed to merging this PR, meaning we get neither. That's why I think they should have been considered separately. (But that's my 2¢).

Thunderforge:
Also worth noting that when the window is small, the buttons move to the bottom and take 1/4th of the width regardless. No space is reclaimed by removing the ellipses. So this proposed change would only affect the buttons on the side.

image

Thunderforge:
Although speaking of which, how does this shrunken down view react to your new "Global Options" button?

Die4Ever:
image

Die4Ever:
> Although speaking of which, how does this shrunken down view react to your new "Global Options" button? Nine buttons instead of eight.

it's still 8 buttons

Thunderforge:
Oh, I see, you got rid of "Edit Game…". I guess I had in my mind that it was staying and it was a new "Game Options…".

At first, I was unsure about this, until I noticed that 2.5.1's tooltip for "Edit Game…" says "Change game options". In that case, I'm okay with the rename.

Die4Ever:
image

raziel-:
Not that I had a say in this, but I like the change :-)

gui/launcher.cpp
```diff
@@ -979,12 +982,12 @@ void LauncherSimple::build() {
 	_loadButton = loadButton;
 
 	// Add edit button
-	if (g_system->getOverlayWidth() > 320) {
+	if (g_system->getOverlayWidth() > 410) {
```
bluegr, line 985: Why?

Die4Ever: that's around the point where Game Options was too long

criezy:
I agree on both points with @Thunderforge 
 - I don't think we should remove the ellipses for all buttons. It makes sense for the About button (and I have just merged the PR that does that), but it is a widely used and accepted UI design element to have ellipses for labels of buttons and menu items that require more inputs from the user.
 - I think renaming the `Edit Game...` button to `Game Options...`and `Options...` to `Global Options...` is a good idea.

We could also consider alternatives, such as `Game Settings...` and `Global Options...`, which would allow using `Settings...` and `Options...` in lowres mode. However I don't really like the idea of having two different names (`settings` and `options`) for what is basically the same thing applied to two different scopes.

gui/launcher.cpp
criezy: I don't like this particular change. There is a consistent understanding currently that lowres is for width of 320 pixels or less. With that change it becomes inconsistent. Also that is only based on the English labels and may not necessarily play well for the translated ones.
We could consider changing the definition of lowres consistently everywhere (for example anything below 640 in width).

criezy:
Also we will probably need to update the doc (in `doc/docportal`) to reflect the changes.

gui/launcher.cpp
Die4Ever: I put it back to 320

criezy (approved):
This needs to be squashed, but this can be done when merging the pull request.
We will however also need to remember to update the documentation in doc/docportal/
And we may want to consider increasing the width up to which we use the lowers mode (consistently for all widget), but this is something that is separate from this pull request.

sev-:
This feature went half-baked. Please fix 320x200 too:
Screenshot 2022-02-18 at 01 34 15

Die4Ever:
> This feature went half-baked. Please fix 320x200 too: Screenshot 2022-02-18 at 01 34 15

how did you generate that screenshot? looking at your screenshot it's attempting to spell out the whole word "Options", but look at the code and it should be shortened to "Opts"

the code only works for the initial window size and it doesn't work if you just resize the window (this is old behavior not something I changed), so if you're testing on desktop you need to edit the scummvm.ini file

this is what mine looks like:
image

sev-:
I launched scummvm in 2x and then switched to 1x. You have to implement this logic also in `::reflowLayout()`.

PR #3682, opened 2022-01-26
GUI: Changing "About..." button to "About"

Thunderforge:
The current recommended and de factor behavior from both Microsoft and Apple is to have "About" menus be without trailing ellipses. Therefore, I think we should follow suit.

----

Traditionally, About menus and buttons had a trailing ellipsis. [This article](https://uxdesign.cc/dot-dot-dot-7ce6170bfc7f) demonstrates this behavior for  both Mac OS 1.0 (technically "System 1.0") and Windows 1.0.

However, current versions of macOS and Windows do not have a trailing ellipsis in About menus. Moreover, their design guidelines recommend not using them. From [Apple](https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-anatomy/):

> Use an ellipsis whenever choosing a menu item requires additional input from the user. The ellipsis character (…) means a dialog or separate window will open and prompt the user for additional information or to make a choice.

Since no additional input is required from the user for the About menu, it is inappropriate to use an ellipsis according to these guidlelines.

Microsoft [explicitly says](https://docs.microsoft.com/en-us/windows/win32/uxguide/ctrl-command-buttons) that About menus should not have an ellipsis (emphasis mine):

> This doesn't mean you should use an ellipsis whenever an action displays another window only when additional information is required to perform the action. Consequently, *any command button whose implicit verb is to "show another window" doesn't take an ellipsis, such as with the commands About*, Advanced, Help (or any other command linking to a Help topic), Options, Properties, or Settings.

Given all this, I think we should be in line with current best practices and remove it from the menu.

### DIFF
--- a/gui/launcher.cpp
+++ b/gui/launcher.cpp
@@ -227,7 +227,7 @@ void LauncherDialog::build() {
 #endif
 	if (!g_system->hasFeature(OSystem::kFeatureNoQuit))
 		new ButtonWidget(this, _title + ".QuitButton", _("~Q~uit"), _("Quit ScummVM"), kQuitCmd);
-	new ButtonWidget(this, _title + ".AboutButton", _("A~b~out..."), _("About ScummVM"), kAboutCmd);
+	new ButtonWidget(this, _title + ".AboutButton", _("A~b~out"), _("About ScummVM"), kAboutCmd);
 	new ButtonWidget(this, _title + ".OptionsButton", _("~O~ptions..."), _("Change global ScummVM options"), kOptionsCmd);
 
 	// Above the lowest button rows: two more buttons (directly below the list box)


Kawa-oneechan:
If I recall correctly, the ellipsisises denote an action that requires further input to actually do something. "Save" doesn't need anything, it already has a file name to save to, but "Save as..." needs you to tell it the new name before it gets to save anything.

"About" doesn't need anything to do what it should: to show the credits.

Thunderforge:
Yes, that matches what the current Microsoft and Apple guidelines say. I think in the past, the ellipsis was sometimes used to mean that a new window would appear, but now it is used to indicate that more input is needed.

lephilousophe:
[Gnome](https://developer.gnome.org/hig/guidelines/writing-style.html#ellipses) and [KDE](https://develop.kde.org/hig/components/navigation/pushbutton/#appearance) all agree with this.
Use ellipsis when more input or confirmation will be needed for the action.

bluegr:
Also agree with this change

bluegr (approved):

aquadran (approved):

criezy:
This is a simple change that makes sense. I will merge it now as I don't think we need to wait for more developers to approve it 😉

criezy:
It might be good to update some of the screenshots in the doc as well, but we might want to wait a bit (see PR #3681).

Thunderforge:
> It might be good to update some of the screenshots in the doc as well, but we might want to wait a bit (see PR #3681).

I agree, we should wait. We will also need to redo some of them given that there is now sorting and the grid view, which are still being tweaked.

PR #3671, opened 2022-01-17
SCUMM: Allow replacing Loom's music with audio tracks

eriktorbjorn:
NOTE: This pull request makes changes to the themes. I have not rebuilt the theme ZIP files, nor have I updated the THEMERC files. The themes have changed since I branched, and I want to avoid conflicts.

~~For various reasons, I probably won't have much time to finish this in the coming weeks, so I'm posting this here for feedback, and in case someone wants to take it over.~~

The purpose of this pull request is to allow certain versions of Loom to replace the synthesized music with standalone audio tracks. These tracks should be named the same way as you would normally name them for games that use CD audio. Obviously, we can't distribute such tracks ourselves (unless someone takes it upon themselves to create new ones from the original Swan Lake score, which is surely in the public domain now). But we can write detailed instructions for how to do so, given the appropriate source material. My own notes, which are subject to corrections, so far:

| Track | Usage | FM Towns Track | TurboGrafx-16 Track | Ballet |
|-----|-----|-----|-----|-----|
| 1 | Overture[^1] | 1 or 9 | 14 | Act II, the swans swim on the lake<br>Act II, Odette and the Swans return to the lake |
| 2 | Loom | 2 or 10 | 15 | Act I, Pas de trois, Intrada |
| 3 | Clerics | 3 or 11 | 16 | Act I, Pas de trois, Andante sostenuto |
| 4 | Blacksmiths | 4 or 12 | 17 | Act I, Pas de trois, Moderato |
| 5 | Shepherds / Dragon caves[^2] | 5 or 13 | 18 | Act I, Pas d'action, Dance of the Drunken Courtier Wolfgang |
| 6 | Dragon (abduction)[^2] | 5 or 13 | 18 | Act I, Pas d'action, Dance of the Drunken Courtier Wolfgang |
| 7 | Glassmakers | 6 or 14 | 19 | Act II, Dance of the Swans |
| 8 | Weavers | 7 or 15 | 20 | Act IV, Dance of the Cygnets |
| 9 | Chaos[^3] | 8 or 16 | 21 | Act IV, Finale |
| 10 | Opening the sky | 1 or 9 | 14 | Near the end of the Overture |
| 11 | Chaos Appears / Unmaking the Loom | 8 or 16 | 21 | Near the end of Chaos |

When extracting from the ballet recording, it's probably a good idea to keep the lead-in short ~~but leave a few seconds of silence at the end for looping tracks~~. The music never loops. In fact, some tracks are unlikely to ever play all the way to the end.

[...]

PR #3684, opened 2022-01-26
SCUMM: Don't require Roland update for Loom demos

eriktorbjorn:
The Loom demos already have the Roland music included, so they don't need the Roland update. But to check for this, we need to set the GF_DEMO flag on the Loom demo. I don't know if this is the best way of doing so, but it's _one_ way at least.

With this done, we can now check if all of the Roland update files are present.

### DIFF
--- a/engines/scumm/scumm.cpp
+++ b/engines/scumm/scumm.cpp
@@ -1316,6 +1316,13 @@ Common::Error ScummEngine::init() {
 
 	Common::String macResourceFile;
 
+	if (_game.id == GID_LOOM && _game.version == 3 && _game.platform == Common::kPlatformDOS) {
+		// There are several files in the full game that don't exist
+		// in either demo, but we look for just one of them.
+		if (!Common::File::exists("06.LFL"))
+			_game.features |= GF_DEMO;
+	}
+
 	if (_game.platform == Common::kPlatformMacintosh) {
 		Common::MacResManager resource;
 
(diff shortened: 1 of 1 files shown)


henke37:
Shouldn't the demo flag be set in the detection code?

eriktorbjorn:
> Shouldn't the demo flag be set in the detection code?

Quite possibly, but I don't know how.

eriktorbjorn:
I suppose one way would be to fix computeGameSettingsFromMD5() instead.

Right now, it will get the following `md5Entry` when launching the demos:

```
(gdb) print *md5Entry
$9 = {md5 = 0x555555e402b8 "a00554c31d623fdb9fcb0f924b89b42b", 
  gameid = 0x555555e3b8a1 "loom", variant = 0x555555e3bd02 "EGA", 
  extra = 0x555555e3e349 "EGA Demo", filesize = -1, language = Common::EN_ANY, 
  platform = Common::kPlatformDOS}
```

This is matched to the following `GameSettings` entry, since they have the same `gameid` ("loom") and the same `variant` ("EGA"):

```
(gdb) print *g
$10 = {gameid = 0x555555e3b8a1 "loom", variant = 0x555555e3bd02 "EGA", 
  preferredTag = 0x555555e4356a "ega", id = 5 '\005', version = 3 '\003', 
  heversion = 0 '\000', midi = 3087, features = 0, 
  platform = Common::kPlatformUnknown, guioptions = 0x555555e43578 "\003"}
```

This does not have the `GF_DEMO` flag, and nothing is done about the `extra` field ("EGA Demo") of the MD5 entry.

criezy:
I think what you could do is:

1. Update `devtools/scumm-md5.txt` to use `EGA Demo` instead of just `EGA` as the variant (5th column) for the loom demos. Also regenerate the engines/scumm/scumm-md5.h
2. Update the `gameVariantsTable` in `engines/scumm/detection_tables.h` to add this loom variant with the `GF_DEMO` flag.

Hopefully that would be sufficient, but I am not very familiar with the SCUMM detection code.

eriktorbjorn:
I'm not familiar with it either, and I can't see any clear pattern. I'll have to ask someone who is familiar.

eriktorbjorn:
I suppose I could rely on the "extra" field in scummvm.ini, but that seems inconsistent.

eriktorbjorn:
This pull request has been properly - I hope - implemented in https://github.com/scummvm/scummvm/pull/3671 so I'm closing it.

PR #3685, opened 2022-01-27
DOCS: Adding Windows 11 to bug report page

Thunderforge:
[Report a bug](https://docs.scummvm.org/en/v2.5.1/help/report_bugs.html#the-scummvm-log-file) lists Windows Vista, 7, 8, and 10. Adding 11.

## Before

image

## After

image

### DIFF
--- a/doc/docportal/help/report_bugs.rst
+++ b/doc/docportal/help/report_bugs.rst
@@ -47,7 +47,7 @@ To help you report a bug, you can find error messages in the ScummVM log file. T
 	.. panels::
 		:column: col-lg-12 mb-2
 
-	        Window Vista/7/8/10
+	        Window Vista/7/8/10/11
 		^^^^^^^^^^^^^^^^^^^^^^
 
                 ``%APPDATA%\ScummVM\Logs\scummvm.log``


criezy:
Do we need to specify the versions?
It would make sense if there was one location for some Windows versions, and another location for other Windows versions. But here there is only one location specified.

Thunderforge:
@criezy The paths are different for Windows 95/98 and XP, which are supported builds. Whether we care or not is a different matter, as they are a very small minority of users.

That said, I’m okay removing them entirely

criezy:
Shouldn't we add the path for Windows 95/98 and XP then? Currently we only have the path for the later versions, which is where my question was coming from.

Also then that may allow finding a wording that would avoid listing all the versions and having it become obsolete when Windows 12 is released (something like `Windows 95/98 and XP ` and `Newer Windows versions` or `Other Windows versions` or `Windows Vista and above` or similar).

Also I see that we have the same issue in other places, for example https://docs.scummvm.org/en/latest/advanced_topics/configuration_file.html, and it might be good to fix it in a consistent way there as well.

Thunderforge:
This PR is superseded by #3687

